PR #471, opened 2016-02-09
nettle: add missing cmd:find as build prerequisite

fbrosson:


fbrosson:
FYI, with this PR the recipe of nettle 3.2 builds on __x86__ :
```
-rw-r--r-- 1 1987394 Feb  9 22:01 nettle_devel-3.2-1-x86.hpkg
-rw-r--r-- 1 3667237 Feb  9 22:00 nettle-3.2-1-x86.hpkg
```
and __x86_64__ :
```
-rw-r--r-- 1 2176317 Feb  9 21:42 nettle_devel-3.2-1-x86_64.hpkg
-rw-r--r-- 1 3939146 Feb  9 21:42 nettle-3.2-1-x86_64.hpkg
```
I tried to build it on x86_gcc2 but, as expectet, the build fails. So I marked it as !x86-gcc2, making the declaration of architectures  for nettle 3.2 identical to that of nettle 3.1.1.

Both the nettle and nettle_devel packages install correctly on both x86 and x86-64:
```
-> uname -a && cd /boot/system/lib && file libnettle.so* libhogweed.so*
Haiku dual 1 hrev50075 Feb  8 2016 01:09:28 BePC x86 Haiku
libnettle.so.6:    symbolic link to libnettle.so.6.2
libnettle.so.6.2:  ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV), dynamically linked, not stripped
libhogweed.so:     symbolic link to libhogweed.so.4.2
libhogweed.so.4:   symbolic link to libhogweed.so.4.2
libhogweed.so.4.2: ELF 32-bit LSB shared object, Intel 80386, version 1 (SYSV), dynamically linked, not stripped
```
```
-> uname -a && cd /boot/system/lib && file libnettle.so* libhogweed.so*
Haiku dual 1 hrev50075 Feb  7 2016 17:09:57 x86_64 x86_64 Haiku
libnettle.so.6:    symbolic link to libnettle.so.6.2
libnettle.so.6.2:  ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
libhogweed.so:     symbolic link to libhogweed.so.4.2
libhogweed.so.4:   symbolic link to libhogweed.so.4.2
libhogweed.so.4.2: ELF 64-bit LSB shared object, x86-64, version 1 (SYSV), dynamically linked, not stripped
```
BTW I'm playing with an 8GB tripple-boot USB key where each of the 3 "main" architectures (x86_gcc2, x86 and x86_64) has it's own partition. I use a separate partition for the haikuporter tree... with a symlink for the packages directory because the system does dot filter the hpkgs which do not match the native arch.
Reminder for new readers  only x86_gcc2 is supported.

fbrosson:
Sorry, there was a typo ("> 6" instead of ">= 6"). It's fixed now.

net-libs/nettle/nettle-3.2.recipe
```diff
@@ -10,12 +10,13 @@ REVISION="1"
 SOURCE_URI="ftp://ftp.gnu.org/gnu/nettle/nettle-$portVersion.tar.gz"
 CHECKSUM_SHA256="ea4283def236413edab5a4cf9cf32adf540c8df1b9b67641cfc2302fca849d97"
 
-ARCHITECTURES="?x86_gcc2 x86 x86_64 ?arm ?ppc"
-SECONDARY_ARCHITECTURES="?x86_gcc2 x86 ?x86_64 ?arm ?ppc"
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86 ?arm ?ppc"
```
korli: Please drop arm, ppc as secondary architectures.

korli:
nettle build normally install a few commands like nettle-hash or nettle-lfib-stream.
Can you investigate why they aren't built? Or they are only missing in PROVIDES?

fbrosson:
Well, it did build nettle-hash and nettle-lfib-stream.
```
~> nettle-hash --version
nettle-hash (nettle 3.2)
~> nettle-hash --list
      name digestsize (internal block size), in units of octets
       md2 16 (16)
       md4 16 (64)
       md5 16 (64)
 ripemd160 20 (64)
      sha1 20 (64)
    sha224 28 (64)
    sha256 32 (64)
    sha384 48 (128)
    sha512 64 (128)
~> nettle-lfib-stream --version
Usage: lfib-stream [SEED]
~> cd /bin && ls -go nettle-hash nettle-lfib-stream
-r-xr-xr-x 1 81688 Feb 10 12:07 nettle-hash
-r-xr-xr-x 1 36088 Feb 10 12:07 nettle-lfib-stream
```

korli:
Ok, then please add missing commands in the PROVIDES section.

fbrosson:
OK. Done. I've just added the declaration of the 5 commands provided by nettle:
* nettle-hash
* nettle-lfib-stream
* nettle-pbkdf2
* pkcs1-conv
* sexp-conv

i guess it is OK to declare them all in the recipe.

fbrosson:
haikuporter is complaining with the cmd: lines:
> Error: dashes are not allowed in provides- or requires declarations

I've tried a few workarounds (escaping the dash, using simple quotes) but none worked.
Is there a way to specify provided commands that include dashes in their names? Or should I just drop the cmd: lines from PROVIDES?

mmuman:
Replace them with an underscore.
Usually the policy warning suggests the missing provides.

fbrosson:
Done (I found scons had a command with a dash)

korli:
OK. Next thing is building with a secondary arch, and check these cmd:*.
You could check for instance the recipe for xzutils which provides cmds as a secondary arch.

fbrosson:
I'm having a problem with my new nettle since I added the cmd: : these commands are missing now.
For cross-building, I'll have to dig the question because I haven't read the docs yet. I have HaikuPorter working fine for native builds for x86_64, x86 and x86_gcc2. If its just a matter of running x86_gcc2  and adding some packages to it, then call haikuporter (or the hp alias) with some additional options, then I guess I can do it as soon as I find which packages and options I need.

korli:
You should be able to build on x86_gcc2 with: haikuporter --no-dependencies nettle_x86

fbrosson:
OK, thanks a lot. I'll switch to x86_gcc2 within 5 minutes. I think I found the problem with the new cmd: declarations: I had forgotten to use the $secondaryArchSuffix. Maybe haikuporter is picky even for native builds.

fbrosson:
Hmm, it complains:
```
haikuporter --no-dependencies nettle_x86
Error: nettle_x86 not found in repository
```
Shouldn't I be using at least one of these options:
* --cross-devel-package=CROSSDEVELPACKAGE
* --secondary-cross-devel-packages=SECONDARYCROSSDEVELPACKAGES
* --cross-tools=CROSSTOOLS

korli:
In haikuporter.conf there is a parameter to activate secondary arch.

fbrosson:
Of course! Thanks!

fbrosson:
I'm back! Sorry for the delay. (I was busy with a couple of higher priority tasks.)
Good news: everything looks ok. I was able to cross-build nettle_x86 while running x86_gcc2:
```
uname -a && ls -go nettle*
Haiku dual 1 hrev50075 Feb  8 2016 01:26:37 BePC x86 Haiku
-rw-r--r-- 1 1988123 Feb 11 13:47 nettle_x86_devel-3.2-1-x86_gcc2.hpkg
-rw-r--r-- 1 3670243 Feb 11 13:47 nettle_x86-3.2-1-x86_gcc2.hpkg
```

fbrosson:
I've installed nettle_x86 (and nettle_devel_x86) on the same system I used to build them (i.e. my x86_gcc2 hybrid) and the commands are all available. And as we can see with nettle-hash-x86 it looks like everything is OK:
```
nettle-hash-x86 --list
      name digestsize (internal block size), in units of octets
       md2 16 (16)
       md4 16 (64)
       md5 16 (64)
 ripemd160 20 (64)
      sha1 20 (64)
    sha224 28 (64)
    sha256 32 (64)
    sha384 48 (128)
    sha512 64 (128)

nettle-hash-x86 -a md5 haikuporter-1.0.0-1-any.hpkg 
haikuporter-1.0.0-1-any.hpkg: 682976d530ea2de7 d4b2ae88f090645f md5

md5sum haikuporter-1.0.0-1-any.hpkg 
682976d530ea2de7d4b2ae88f090645f  haikuporter-1.0.0-1-any.hpkg
```

BTW, in my previous build I used --no-source-packages, so I kicked another build without that option and the source package gets "built" as expected:
```
-rw-r--r-- 1 1988128 Feb 11 14:31 nettle_x86_devel-3.2-1-x86_gcc2.hpkg
-rw-r--r-- 1 1863892 Feb 11 14:32 nettle_x86_source-3.2-1-source.hpkg
-rw-r--r-- 1 3670246 Feb 11 14:31 nettle_x86-3.2-1-x86_gcc2.hpkg
```
So I guess the recipe is OK.
Is there something to do with the "?arm" and "?ppc" which are currently in ARCHITECTURES? Should I remove them completely, or move or add them to SECONDARY_ARCHITECTURES ? (I have an old ppc iBook G4 on which I might try Haiku some day, maybe.)

pulkomandy:
arm and ppc are never secondary architectures. Haiku doesn't run yet on arm or ppc anyway, so there is not much point in declaring support for these in the normal repository. They are used only in haikuports-cross at the moment (for preparing the build packages for these architectures).

fbrosson:
OK, thanks. So if I understand correctly they are not forbidden but they are not interesting unless it is in a package that might be copied to haikuports-cross. But they could be interesting to developers that manage to build and install haiku_arm_devel or haiku_ppc_devel, so it does not harm to keep "?arm" and/or "?ppc"  in ARCHITECTURES. Is this correct?

pulkomandy:
You can't install these packages and cross-compile this way, yet, so there is no way to test this. The only way to build ARM and PPC packages is running Haikuporter on Linux and using the haikuports-cross repo.
This means these architectures will only become useful in the main repo when we get Haiku running on these architectures (well enough to run Haikuporter natively), and we aren't quite there yet.

fbrosson:
OK, thanks.for your reply. (On some of the recipes I have made changes to I had added these architectures but now I understand it was really too early for that, so I won't add them anymore, I promisse :)

korli:
nettle_x86 shouldn't contain commands if they're already provided by nettle.

fbrosson:
Is there something wrong in the recipe? Should I make a conditional provides for the cmd: lines? If yes, do you know where I can look for a recipe with a similar issue? (or could you guide me for the fix) Thanks

mmuman:
You can look at dev-libs/openssl/openssl-1.0.0t.recipe for example.

fbrosson:
Important: I still need to test it.

korli:
This [part](https://github.com/haikuports/haikuports/blob/master/dev-libs/openssl/openssl-1.0.2f.recipe#L98) is missing.

fbrosson:
Fixed, thanks. BTW, I also added `libhogweed` after `prepareInstalledDevelLibs libnettle`. Regarding the man pages and docs, the recipe might still need some editing . I'll try to dig that. (hints are always welcome, as usual :-)

korli:
$prefix/bin isn't valid for secondary arch, please use $binDir.

fbrosson:
fixed

PR #475, opened 2016-02-17
Fix ordering

DarkmatterVale:
This is a massive commit, I apologize in advance :D. I have fixed all recipes (that adhere to the content within a recipe put forth on the wiki) whose order was not correct.

As you can probably guess, I did not edit each of these files manually. I created a python script that goes through all of the recipes in the tree and only edits ones that have only the correct content (it is hard to parse files that have additional content) but in the wrong order.


DarkmatterVale:
Please note the lint issues are not cause by my addition. Rather, they are caused by other peoples' recipes not having the correct values (which I am not at a stage to correct).

korli:
Thanks!

DarkmatterVale:
Glad I was able to save the devs some time by doing this automatically :)

PR #477, opened 2016-02-20
libburn: bump version

fbrosson:
Also add PATCH() to fix links to the cdrtools project.


dev-libs/libburn/libburn-1.4.2.pl01.recipe
```diff
@@ -3,11 +3,12 @@ DESCRIPTION="libburn is a library that can record preformatted data onto CD \
...
 a facility for reading data blocks from optical media."
 HOMEPAGE="http://www.libburnia-project.org/"
-COPYRIGHT="2006-2014 Mario Danic and Thomas Schmitt"
+COPYRIGHT="2006-2016 Mario Danic and Thomas Schmitt"
 LICENSE="GNU GPL v2"
 REVISION="1"
 SOURCE_URI="http://files.libburnia-project.org/releases/libburn-$portVersion.tar.gz"
-CHECKSUM_SHA256="46266bdd4b37ca9700a44bbd8dbfbf3526423ede25751fd2f0211b2b5ce29283"
+CHECKSUM_SHA256="33cfcd7f5deb3c6f325ca9eaef36c59b4fcb384bfcddaaa16d9a89da88a80edb"
+SOURCE_DIR="libburn-1.4.2"
```
Max-Might: You could use `$portVersion` here too.

fbrosson: Yep, I know. I tried
```Shell
SOURCE_DIR="libburn-${portVersion/.pl*/}"
```
and noticed that it worked fine to get rid of the **`.pl01`** suffix, but I did not dare using it in the recipe because I thought it could be seen as not necessary. But I can use it if you like it. (I think it's a good idea.)

PR #480, opened 2016-02-22
lynx: bump version & update homepage

fbrosson:


fbrosson:
Lynx 2.8.9dev.8 builds and works fine on Haiku hrev50107 (x86_gcc2, x86, x86_64):
```
package info lynx-2.8.9dev.8-1-x86_gcc2.hpkg
name: lynx  version: 2.8.9dev.8-1

package info lynx_x86-2.8.9dev.8-1-x86_gcc2.hpkg
name: lynx_x86  version: 2.8.9dev.8-1

package info lynx-2.8.9dev.8-1-x86.hpkg
name: lynx  version: 2.8.9dev.8-1

package info lynx-2.8.9dev.8-1-x86_64.hpkg 
name: lynx  version: 2.8.9dev.8-1
```

mmuman:
Looks ok.
Why did you change the commit author in the patch? Using git am or git rebase should keep those.

fbrosson:
>  Why did you change the commit author in the patch? Using git am or git rebase should keep those.

Well, because I used **`haikuporter --nobuild`**, applyed in turn each of the 3 parts of the old patch (with the **`patch`** command, not manually),  used git commit after each part, then used **`haikuporter --extract-patchset`** to regenerate the new patch. Then I made a test without the third part of the patch (and noticed it was still needed), then restored that third part, and finally removed the noise (random data such as local git hashes are useless and timestamps are redundant with those of the logs available from the Git repo on GitHub). So after that it seemed to me that it would have been bad practice to leave someone else's mail associated with the picky cleaning I did…
But I think it is fixed now, as I have overwritten the "From" lines and added back the "Date" lines from the old patch.
Ah, I just noticed that you are the author of the patch for 2.8.9dev.1. So here's a question for you: Is the new patch OK now? I mean, with the "From:" and "Date:" restored, but still with the other cleaning (that is, without git hashes and git index lines) ? Or would you prefer to see the first version of the patch for dev.8 with my "e-mail"? I've got a copy of that version of the patch and I can restore it in the commit if you wish. (BTW, I feel uncomfortable with refreshing the patch with the automatic method because that would only add the local random data which just consumes space and clutters git diffs).
Anyway, thanks for the hints about git and for having reviewed the recipe!

korli:
Your workflow seems over-complicated to me. Next time, just let haikuporter apply the patchset as is. When it cannot be applied automatically, git will ask you to fix things the usual way (apply the rebase patch manually, add modified files, then git am --continue). The patchset will stay in good shape and haikuporter --extract-patchset will regenerate the new patch

fbrosson:
OK, I'll try the simple workflow next time. (I was more comfortable with using the patch command than letting haikuporter apply the patch from the previous release.) Thanks for the hints!

PR #482, opened 2016-02-23
uif2iso: add x86 as 2nd arch, mark !x86_gcc2, rewrite description

fbrosson:

PR #483, opened 2016-02-24
daa2iso: reorder sections and mark x86_{gcc2,x64} as tested

fbrosson:
Also install documentation/packages/daa2iso/daa2iso.txt


fbrosson:
I've just added a tiny change for uif2iso. I would have included it in #482 if I had thought about it before that PR got merged. (Sorry)
Is it OK to "use" this same PR for that? If not I can easily revert the change and create a separate PR, but I guess it will be OK because both packages are related (same author and topic) and the change is really small.
BTW, I did not increment the revision of uif2iso. Should I do so? Next time I'll use a checklist before I submit a PR. Thanks.

korli:
The resulting packages are different, revision should be bumped.

fbrosson:
ok, done. Thanks for the merge!

PR #484, opened 2016-02-25
ed: bump version & add x86_64.

fbrosson:


fbrosson:
Upstream has switched from **`tar.gz`** to **`tar.lz`** when they released 1.10.
https://fossies.org/ keeps a _perfect_ **`tar.gz`**  "translation" of the original **`tar.lz`**, but they only keep the latest stable.
This means that we should probably keep both 1.9 and the latest stable (1.13), for situations where the recipe for the latest stable becomes outdated and a user of a __non hybrid__ x86_gcc2 choses to build from source. Since those users can't have **`lzip`** on non hybrid x86_gcc2 their only options would be to build ed-1.9 or to grab the **`tar.gz`** for 1.13 from some other mirror.
Let me know if having a conditional SOURCES_URI in the recipe for version 1.13 (in order to be able to use the lziped tarball from upstream instead of the gziped one from foosies.org) is acceptable.
BTW, if there is a way, with`$effectiveTargetArchitecture`or some other test,  to distinguish x86_gcc2 hybrid from non hybrid in a recipe then we could improve the recipe to make it usable with **`lzip`** on hybrid x86_gcc2 boxes.
Thanks for any feedback!

korli:
I don't quite understand. We only need the 1.13 recipe with the tar.gz SOURCE_URI. Or do I miss something? Please add a comment with the tar.lz SOURCE_URI as well.

fbrosson:
The ed-1.13.tag.gz tarball is not available from upstream. It was generated by fossies.org (from the official ed-1.13.tar.lz tarball).
It is safe to use ed-1.13.tag.gz but when/if ed-1.14 becomes available, fossies.org will delete their old tarball.
So it would be better to rely on the ed-1.13.tar.lz tarball.
Regarding the recipe for ed-1.9, I just added it in case a user on a non-hybrid x86_gcc2 system fails to find ed-1.13.tar.gz from an alternate mirror or is unable to use ed-1.13.tar.lz.
But if you prefer to only keep ed-1.13.recipe then I'll remve ed-1.9.recipe.
Regarding the tarball of ed-1.13, I think it would be nice to use the lziped tarball on platforms that have lzip, unless you prefer to use a single tarball for all architecures, in which case I'll switch to ed-1.13.tar.gz from fossies.org.
What comment would you like to see about the tar.lz?
Would something like the following line be OK for the comment?
"Upstream only ships lziped tarballs since release 1.10, but cmd:lzip is not available for non-hybrid x86_gcc."

fbrosson:
That's it. I got rid of **`ed-1.9.recipe`** and added a couple of comments in **`ed-1.13.recipe`** to document the reason why the recipe uses different tarball formats. Please don't hesitate to ask for extra changes.
BTW, is there a way to test whether we are on a hybrid x86_gcc2 versus a non-hybrid x86_gcc2? Thanks!

fbrosson:
Oops, I lost my comments. I'll add them again.

korli:
Please use only the tar.gz file SOURCE_URI and leave the other in comment.

fbrosson:
Done.

korli:
Thanks

PR #494, opened 2016-02-29
lcab: create recipe for version 1.0b12.

fbrosson:


korli:
Patch doesn't seem to work:
https://hpkg.mlotz.ch/x86_64/output/builds/29135.log

fbrosson:
Oops, I know what happend. I also get the same error... but _only_ with /bin/haikuporter.
With the haikuporter from the repo it works fine.
I'll fix the patch asap to make it work for all versions of haikuporter.
And I won't forget to test my future edits with the default haikuporter. (Sorry for this time.)

korli:
BTW I wonder whether we should explicit require git patches with author information. Simple patches will generally not be accepted upstream, so we have to live with the burden of maintaining them...

fbrosson:
Well, the new haikuporter (I mean, the one in the Git repo) does not have this requirement.
That's it. I've fixed the patch and made a couple of tiny changes to the recipe. It's in #495.

PR #490, opened 2016-02-28
dvdbackup: create recipe.

fbrosson:

PR #495, opened 2016-03-01
lcab: fix patch to meet /bin/haikuporter requirements.

fbrosson:
Although this recipe was OK for the haikuporter from the haikuport**_er_** repo, it was failling to pass the patch phase with the default haikuporter, i.e. the one that is available as a "pre-built" package from the haikuport**_s_** repo and which gets installed in /bin.
This edit fixes that issue by switching the patch format from ".patch" to ".patchset" and, most important, avoiding the use of patches created with "diff -U0" since haikuporter does not accept contexts of 0 lines.


fbrosson:
Hmm, I though the problem about the man pages not being stored in the package when the _default_ haikuporter was being used was due to haikuporter itself, but I've just tried re-buioding dvdbackup (https://github.com/haikuports/haikuports/pull/490) with the _default_ haikuporter (instead of the haikuporter from the repo) and noticed that it was successfully storing the man pages.
So the issue about the man page of lcab not being stored without explicit steps might not be due to haikuporter.
Unless, off course, this is due to the **`--omit-dirs`** switch of **`runConfigure`** that I absolutely had to provide for lcab (due to its configure script not accepting **`docDir`** nor **`dataRootDir`**), while dvdbackup's configure does not complain when given these Dir options.

waddlesplash:
> Unless, off course, this is due to the --omit-dirs switch of runConfigure that I absolutely had to provide for lcab (due to its configure script not accepting docDir nor dataRootDir), while dvdbackup's configure does not complain when given these Dir options.

It is. The manpages go in docDir, and if configure doesn't know where that is it's probably putting them in the wrong place.

fbrosson:
Thanks @waddlesplash for your hint about the docDir, and sorry for this late reply. The recipe now builds with all versions of haikuporter.
Except for the patch issue (now fixed), the other problems were not due to haikuporter as I initialy though, but by a filesystem full! That's why I've had to resize my partitions...
BTW, I was "playing" with an 8 GB thumb key with 3 partitions of 600 MB and a big data partition of around 6,2 GB. I have changed my partition layout and now I have 3 bigger system partitions and a comfortable data partition for the haikuports tree. In case some newcomer would like reproduce the triple-boot layout, here's the partition table I have:
```
#  Start      End Sectors  Size Id Type
1   8192  2465791 2457600  1.2G eb BeOS fs [x86 GCC 2 Hybrid]
2  65792  4923391 2457600  1.2G eb BeOS fs [x86 GCC 4 Hybrid]
3  23392  7380991 2457600  1.2G eb BeOS fs [x86_64]
4  80992 15646719 8265728    4G eb BeOS fs [data (haikuports)]
```
(Off couse, I used **`unzip -p current_anyboot.zip nightly-anyboot.image`** piped through **`dd`** to skip the first 4 MB of the haiku-nightly-anyboot.image and put it in a 600 MB partition, then installed it over the 1.2 GB partition.)
This is cool because with a tiny USB thump key I can check that a recipe builds on _any_ architecture, inluding both x86_gcc2 and x86 as secondary architectures.
BTW, the packagefs concept (among others) in Haiku is awsome!

waddlesplash:
Er, need $docDir, but that's still ommitted?

fbrosson:
Yep, we need to pass both docDir and dataRootDir as omit-dirs.
If we call **`runConfigure  ./configure`** or **`runConfigure --omit-dirs dataRootDir ./configure`** then we get this error:
```
configure: error: unrecognized option: --docdir=/packages/lcab-1.0b12-1/.self/documentation/packages/lcab
```
And if we call **`runConfigure --omit-dirs docDir ./configure`** then we get this other error:
```
configure: error: unrecognized option: --datarootdir=/packages/lcab-1.0b12-1/.self/data
```
So my conclusion is that the current commit is required _as is_.
Reminder: the recipe that is already in master already works fine with the haikuporter from the haikuport**_er_** repo. It only fails with the haikuporter from the haikuport**_s_** repo. (This commit will make the recipe work with the haikuporter from _any_ repo.)

waddlesplash:
> Reminder: the recipe that is already in master already works fine with the haikuporter from the haikuporter repo. It only fails with the haikuporter from the haikuports repo. (This commit will make the recipe work with the haikuporter from any repo.)

Oh, in that case, this isn't needed -- the haikuporter recipe is just an older version of HaikuPorter, we probably should update it.

fbrosson:
Well, I think @korli might be willing to have the fix merged, as he reported in # https://github.com/haikuports/haikuports/pull/494#issuecomment-190620793 that the patch was not passing.
BTW, I've just simplified the fix. I was not necessary to switch to the patchset format. The only problem was the length of the context for patching one of the files (the one that contained the wrong version string.)
So now the delta in the commit of this PR is reduced to the minimum. Maybe it would be worth merging.
Hmm, the refreshed commit (https://github.com/haikuports/haikuports/commit/40ef0b53527147165bdd9f333be685d167324c23) does not popup in this PR because it is closed. If you re-open it then maybe it will become available for merging without a new PR, in case you decide to merge. (I promisse I won't use any strange context lengths in patches anymore ;-)

PR #500, opened 2016-03-08
mksh: bump version, add x86{,_64} with a patch.

fbrosson:
* Drop the "r" prefix in version.
* Bump version to 52c and use portVersion where appropriate.
* Fix wrong LICENSE declaration and add missing licenses.
* Move haiku_devel from BUILD_PREREQUIRES to BUILD_REQUIRES to be consistent with most other recipes.
* Add missing cmd:awk to BUILD_PREREQUIRES.
* Add x86{,_64} to ARCHITECTURES with a tiny conditional patch to make the recipe build on these architectures.
* Do not create the «**`sh -> mksh`**» symbolic link anymore as changing the default «**`sh -> bash`**» symbolic link might create problems. Users willing to set /bin/mksh as their default shell just need to edit /etc/passwd.

### DIFF
--- a/app-shells/mksh/licenses/ISC
+++ b/app-shells/mksh/licenses/ISC
@@ -1,0 +1,21 @@
+ISC Licence (OSI approved)
+========================
+
+strlcpy.c is
+Copyright (c) 2006, 2008, 2009, 2013
+	mirabilos <m@mirbsd.org>
+Copyright (c) 1998
+	Todd C. Miller <Todd.Miller@courtesan.com>
+
+
+Permission to use, copy, modify, and distribute this software for any
+purpose with or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+
+THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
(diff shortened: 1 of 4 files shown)


fbrosson:
After trying some source patches for **`sh.h`** to fix an error about **`getrusage()`** being redeclared, I got another error which seemed to be related to stack protection. So I patched the build script to skip **`-fstack-protector-strong`** and **`-fstack-protector-all`** and noticed that without the stack protection switches everying seems to work both with and without the initial source patch. So I just kept the patch that drops the stack protection switches.
And since the recipe builds and works fine on x86_gcc2 without any patch, probably because the patch is neutral with gcc2, I chose to not apply any patch on x86_gcc2.
I don't know if disabling stack protection is acceptable, but since x86 and x86_64 were both missing in ARCHITECTURES I thought we could give it a try.
BTW, I don't use mksh but I guess some other Haiku users do. This will allow them to have it on more arches.

korli:
The version should probably be just 52c instead of r52c.

korli:
Licensing seems off the reality. This file seems to describe the licensing terms of mksh: https://www.mirbsd.org/TaC-mksh.txt

fbrosson:
> The version should probably be just 52c instead of r52c.

Good idea, thanks!

> Licensing seems off the reality. This file seems to describe the licensing terms of mksh: https://www.mirbsd.org/TaC-mksh.txt

I've fixed the LICENSE declaration and created licenses/MirOS (with the contents of appendix A) and licenses/ISC (with those of appendix C).
I did not add the license for the icon because we're not shipping it in the package.
I think everything is OK now. (Is it?)

Hmm, regarding the **`sh -> mksh`** symbolic link, maybe we should avoid creating it to allow occasional users to keep their default shell. What do you think?

pulkomandy:
the sh->mksh link sounds a bit dangerous as default. It should be possible to install multiple shells side by side and the default package shouldn't make the shell default.

What we could do is, have a separate package with just the link (mksh_systemshell ?), for people who want to use mksh as the system shell. We're not removing bash from Haiku in the near future because it was the shell used in BeOS and we want to stay compatible. In the R2 future, switchign to a MIT-licensed shell (if there is one) would be nice.

For now it is enough that the shell for users can be changed in /etc/passwd, no need to replace the system shell (which is not much used since the launch scripts have been replaced by launch daemon anyway)

fbrosson:
I moved the symbolic link to a separate package, mksh_systemshell, as you suggested, but the latter also contains copies of the license files, even though they are already in the base pacakge. I understand this is a sane default, but since the separate package only holds a symbolic link it would be nice to teach haikuporter to handle something like **`LICENSE_systemshell=""`**.
Well, I guess for now it is OK to have the license files embedded in both packages.

fbrosson:
Minor cosmetic fix for the LICENSE definition: spread across 3 lines instead of 5.

fbrosson:
After all, let's keep things simple. I've reverted the package split and removed the creation of the **`sh -> mksh`** symbolic link. Users willing to have mksh as their default shell just need to edit **`/etc/passwd`** if they want. There are very few recipes that split packages. We don't really need a package with just a symbolic link for mksh.

PR #501, opened 2016-03-08
musepack: add x86{,_64} to libcuefile, libreplaygain and musepack_tools.

fbrosson:
Also use $portVersion instead of the current version (r475) in SOURCE_URI and SOURCE_DIR.


fbrosson:
Shall I remove the "r" prefix in the version (r475) of these 3 recipes?

pulkomandy:
Does it create a problem with version comparison?

fbrosson:
No, I guess not.
There seems to be very few dependencies on these 3 ports, and none of them are versioned deps.
And if I'm not mistaken, haikuporter does not support versioned deps that contain non-digts. So getting rid of the "r" prefix will make it possible to add **`>=`** style deps in case such a versioned dependency (on one of these musepack components) becomes necessary someday.

pulkomandy:
I wrote these recipes for use in VLC 0.8.6i, so that probably is the only thing that needs to be adjusted (unless there were more added since then?).
Ok then, it should be fixed to make package versionning work.

fbrosson:
**`lib:libmpcdec`** (provided by musepack_tools) is required by:
* media-video/vlc/vlc-0.8.6i.recipe
* media-video/vlc/vlc-2.2.1.recipe
* media-sound/cmus/cmus-2.7.1.recipe

It also appears in **`media-video/mplayer/mplayer-1.1.1.recipe`** but it is currently commented out.

**`lib:libcuefile`** and **`lib:libreplaygain`** are only needed by musepack_tools.
There are currently no other recipes depending on any of the 3 recipes of this PR, so I think we can proceed with the changes.
Regarding the versioned lib provides, the current so versions are in sync with the PROVIDES, we have:
* **`0.0.0 compat = 0`** for lib:libcuefile (which perfecly matches libcuefile.so.0.0.0)
* **`1.0.0 compat = 1`** for lib:libreplaygain (which also perfecly matches libreplaygain.so.1.0.0)

Moreover, none of the REQUIRES are versioned, so this means we won't even need to bump any revision (since we aren't chaging the contents of the packages but only their declared versions.)

korli:
Provides for lib:libmpcdec should declare a version though. 6.1.0 compat >= 6 if I'm not mistaken. 

Please make sure you also rename patches files, or fix their names in recipes.

fbrosson:
> Provides for lib:libmpcdec should declare a version though. 6.1.0 compat >= 6 if I'm not mistaken. 

Upstream does not provide any versioning for libmpcdec. We only get libmpcdec.so. (Should we add something???)

> Please make sure you also rename patches files, or fix their names in recipes.

I've already renamed the patch files (dropping the "r") and reference them with
```Shell
PATCHES="musepack_tools-$portVersion.patchset"
```
for **`media-sound/musepack_tools/musepack_tools-475.recipe`** and with
```Shell
PATCHES="libcuefile-$portVersion.patchset"
```
for **`media-libs/libcuefile/libcuefile-475.recipe`** so I think everything is already OK in the current commit.
I can, however, rebase it against master if it makes merging easier.

korli:
Ok for patchsets.

For libmpcdec versioning, I refer to https://www.archlinux.org/packages/extra/i686/libmpcdec/ (click on View the file list for libmpcdec). But other distributions have versioning too.

korli:
OK it seems we use CMake, whereas distributions use autotools, which happen to produce versioned artefacts. Let's drop this item for when upstream has cleared the way.

fbrosson:
OK

PR #502, opened 2016-03-10
openssl: bump version (1.0.2g & 1.0.1s)

fbrosson:
* bump 1.0.2 series to 1.0.2g with **`enable-ssl2`** to prevent API breakage.

* bump 1.0.1 series to 1.0.1s with **`enable-ssl2`** to prevent API breakage.

* Fix the patchset for the 1.0.1 series:
 1. by using a copy of patch «**`Do not use __INTEL__ to detect x86_64`**»
from the 1.0.2 series for **`crypto/evp/e_aes_cbc_hmac_sha1.c`**.
 2. by using a copy of the 2nd chunk of patch «**`more __intel__ fixes...`**»
from the 1.0.2 series for **`crypto/evp/e_rc4_hmac_md5.c`**.

### DIFF
--- a/dev-libs/openssl/openssl-1.0.1s.recipe
+++ b/dev-libs/openssl/openssl-1.0.1s.recipe
@@ -15,9 +15,9 @@ HOMEPAGE="https://www.openssl.org/"
 COPYRIGHT="1995-1998 Eric Young
 	1998-2016 The OpenSSL Project."
 LICENSE="OpenSSL"
-REVISION="2"
+REVISION="1"
 SOURCE_URI="https://www.openssl.org/source/openssl-$portVersion.tar.gz"
-CHECKSUM_SHA256="932b4ee4def2b434f85435d9e3e19ca8ba99ce9a065a61524b429a9d5e9b2e9c"
+CHECKSUM_SHA256="e7e81d82f3cd538ab0cdba494006d44aab9dd96b7f6233ce9971fb7c7916d511"
 PATCHES="openssl-$portVersion.patchset"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64 ?arm ?ppc"
@@ -73,7 +73,7 @@ BUILD()
 	PERL="/bin/env perl" \
 	./config --prefix=$prefix --libdir=$relativeLibDir \
 		--openssldir=$dataRootDir/ssl \
-		zlib shared
+		enable-ssl2 zlib shared
 	make
 		# multi-job builds don't work correctly
 }
--- a/dev-libs/openssl/openssl-1.0.2g.recipe
+++ b/dev-libs/openssl/openssl-1.0.2g.recipe
@@ -17,7 +17,7 @@ COPYRIGHT="1995-1998 Eric Young
 LICENSE="OpenSSL"
 REVISION="1"
 SOURCE_URI="https://www.openssl.org/source/openssl-$portVersion.tar.gz"
-CHECKSUM_SHA256="784bd8d355ed01ce98b812f873f8b2313da61df7c7b5677fcf2e57b0863a3346"
+CHECKSUM_SHA256="b784b1b3907ce39abf4098702dade6365522a253ad1552e267a9a0e89594aa33"
 PATCHES="openssl-$portVersion.patchset"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64 ?arm ?ppc"
@@ -73,7 +73,7 @@ BUILD()
 	PERL="/bin/env perl" \
 	./config --prefix=$prefix --libdir=$relativeLibDir \
 		--openssldir=$dataRootDir/ssl \
-		zlib shared
+		enable-ssl2 zlib shared
 	make
 		# multi-job builds don't work correctly
 }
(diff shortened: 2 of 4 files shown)


korli:
Could you split the PR? 
There seems to be a possible ABI breakage in 1.0.2g, which needs a thorough check. To anyone please wait before merging, the output of "nm -g" for the lib in 1.0.2f and 1.0.2g should be checked.

fbrosson:
Yes, of course. But I'm not sure how you would like the split to be made.
Do you mean "keep this PR for 1.0.1s only" and "create a separate PR for 1.0.2g" ?

fbrosson:
I've made an initial split in case that could be usefull. I'll create a separate PR if this is what you were asking.
BTW, I also found another **`__INTEL__`** fix that had also been un-noticed in 1.0.1h (released in June, 2014) I've just updated the patchtset for 1.0.1s to fix that.

Reminder: **WORK IN PROGRESS** (I'll compare the output of **`nm -g`** for the libs of 1.0.2g vs 1.0.2f as suggested by @korli )

fbrosson:
Using **`nm -g`** piped through a **`cut -c10-`** for each of the libs in 1.0.2f and 1.0.2g I can say that:
* libcrypto from 1.0.2g has _all_ external symbols 1.0.2f has, **_plus_** the following:
```
T SRP_user_pwd_free
T SRP_VBASE_get1_by_user
```
* libssl from 1.0.2g has _all_ external symbols 1.0.2f has, **_except_** the following:
```
U EVP_CIPHER_CTX_block_size
U EVP_CIPHER_CTX_key_length
U sk_delete
T ssl2_accept
T ssl2_callback_ctrl
D ssl2_ciphers
T ssl2_clear
T ssl2_connect
T ssl2_ctrl
T ssl2_ctx_callback_ctrl
T ssl2_ctx_ctrl
T ssl2_default_timeout
T ssl2_do_write
T ssl2_enc
T ssl2_enc_init
T ssl2_free
T ssl2_generate_key_material
T ssl2_get_cipher
T ssl2_get_cipher_by_char
T ssl2_mac
T ssl2_new
T ssl2_num_ciphers
T ssl2_part_read
T ssl2_peek
T ssl2_pending
T ssl2_put_cipher_by_char
T ssl2_read
T ssl2_return_error
T ssl2_set_certificate
T ssl2_shutdown
R ssl2_version_str
T ssl2_write
T ssl2_write_error
T SSLv2_client_method
T SSLv2_method
T SSLv2_server_method
```
This is not surprising since 1.0.2g (as well as 1.0.1h) have completely disabled SSLv2 as stated in https://openssl.org/news/secadv/20160301.txt.
I guess we'll need to bump the revisions of **all** recipes that use openssl, and that we should probably bimp both 1.0.1 and 1.0.2.
Let me know if I should squash the 2 commits or if you would like to have a separate PR for one of them. Thanks!

korli:
For 1.0.2g we need to enable sslv2 at configure time (add "enable-ssl2").

fbrosson:
OK. I'll do this for 1.0.2g. Shall I do the same for 1.0.1s ?
(I'll check the output of **`nm -g`** for 1.0.1s vs 1.0.1r but I'm pretty sure the set of missing symbols will be the same as for 1.0.2g vs 1.0.2f.)
BTW, may I squash the commits?

pulkomandy:
Do we really want to keep SSLv2 however? The protocol is now known to be broken and enable some attacks just by being enabled.
Should we do this starting with 1.1.0 only and leave it in 1.0.2 and hope we can drop 1.0.2 fast enough? (they plan to release it next month according to https://www.openssl.org/policies/releasestrat.html)

korli:
@pulkomandy we don't really need another API breakage, moreover Gentoo, Redhat have keep SSLv2 enabled. It's well described in the release notes: 

> Even if "enable-ssl2" is used,
users who want to negotiate SSLv2 via the version-flexible SSLv23_method() will
need to explicitly call either of:

>    SSL_CTX_clear_options(ctx, SSL_OP_NO_SSLv2);
>    or
>    SSL_clear_options(ssl, SSL_OP_NO_SSLv2);

> as appropriate.  Even if either of those is used, or the application explicitly
uses the version-specific SSLv2_method() or its client or server variants,
SSLv2 ciphers vulnerable to exhaustive search key recovery have been removed.
Specifically, the SSLv2 40-bit EXPORT ciphers, and SSLv2 56-bit DES are no
longer available.

pulkomandy:
Ok, let's delay this to the 1.1.0 release then. Thanks for clarification.

fbrosson:
I think everything is in order.
BTW, I've also added the **`enable-ssl2`** for **`1.0.1s`** because it was also needed to avoid ABI breakage.

fbrosson:
Oops, there was a typo in the commit message. It's fixed now. (I replaced ABI by API).

fbrosson:
Wait! I lost some headers in the patchset for 1.0.2g. Will fix it right now.

fbrosson:
Done. Everything is OK now.

korli:
"enable-ssl2" is misplaced, it should be on the same line as zlib.

fbrosson:
Fixed.

PR #503, opened 2016-03-10
Generation of .pc files for pkg-config

Sylvain78:
ncurses:Minimal changes to generate .pc files.

### DIFF
--- a/sys-libs/ncurses/ncurses-5.9.recipe
+++ b/sys-libs/ncurses/ncurses-5.9.recipe
@@ -55,6 +55,7 @@ fi
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
+
 if [ -n "$secondaryArchSuffix" ]; then
 	REQUIRES="$REQUIRES
 		ncurses == $portVersion base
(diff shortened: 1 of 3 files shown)


mmuman:
Please use English for commit logs.
Also, as a convention, we prefix the first line with the name of the concerned recipe, like "ncurses: generate .pc files for pkg-config"
You'll also want to use git rebase origin and force push to remove the useless merge commit.

korli:
ncurses 6.0 is the newest recipe, please change this one. ncurses 5.9 is deprecated in Haiku.
I'm pretty sure your patch is broken for secondary architecture. Have you tried instead to define the variable PKG_CONFIG_LIBDIR?

mmuman:
I'm not sure, doesn't haikuporter use setarch now?

mmuman:
Ok, just need to make sure gcc4 .pc are fine now.

korli:
If you could just drop your first commit, only keep the 6.0 commit, that would be cool.
Also use an english comment for your patch.

mmuman:
Oh, we'll want to bump the REVISION, too :-)

korli:
Applied, thanks!

PR #504, opened 2016-03-11
check: add recipe for version 0.10.0.

fbrosson:


dev-libs/check/check-0.10.0.recipe
```diff
@@ -0,0 +1,65 @@
+SUMMARY="Unit Testing Framework for C"
```
waddlesplash: Sentence casing, please

dev-libs/check/patches/check-0.10.0.patch
```diff
@@ -0,0 +1,27 @@
...
+     /* Windows does not like tmpfile(). This is likely because tmpfile()
+      * call unlink() on the file before returning it, to make sure the
+      * file is deleted when it is closed. The unlink() call also fails
+      * on Windows if the file is still open. */
+     /* also note that mkstemp is apparently a C90 replacement for tmpfile */
+     /* perhaps all we need to do on Windows is set TMPDIR to whatever is
+        stored in TEMP for tmpfile to work */
+     /* and finally, the "b" from "w+b" is ignored on OS X, not sure about WIN32 */
+ 
++    *name = NULL;
```
waddlesplash: Are you sure this is correct?

fbrosson: Yep. **`*name`** also gets initialized later even for the **`!HAVE_MKSTRING`** case.

BTW, I noticed that I need to make some changes to the dependencies to allow packages for the  primary and secondary architectures to be co-installed. I'm testing the changes for that.
So there is a little bit more work in progress.

waddlesplash: But why is this needed? I'd prefer to not include patches that fix bugs in the original software rather than bugs specific to Haiku.

fbrosson: It's required for gcc2 because right after the **`#else`** we have:
```C
    int fd = -1;
    const char *tmp_dir = getenv ("TEMP");
```

So if we don't move the **`*name = NULL;`** we cannot build with gcc2.

fbrosson:
I just made the changes that allow to co-install the packages for both the primary and secondary architectures. On x86_gcc2 hybrid, for instance, we have:
* check-0.10.0-1-x86_gcc2.hpkg
* check_devel-0.10.0-1-x86_gcc2.hpkg
* check_x86-0.10.0-1-x86_gcc2.hpkg
* check_x86_devel-0.10.0-1-x86_gcc2.hpkg

In the base package we have, among others:
**`bin/checkmk`** (which is just an awk script).

In the package for the secondary architecture (e.g. x86) we have, among others:
bin/checkmk-x86 -> x86/checkmk
bin/x86/checkmk
But instead of keeping 2 copies of the script, we only keep one in the base package and replace the second copy by a symbolic link to the first copy: ../checkmk. This is done in **`INSTALL()`**.

All scripts and docs are in the base package for the primary architecture. The only files that appear twice are **`check.h`** and **`check_stdint.h`**. They are in **`develop/headers/`** for the primary-arch check_devel package, and in **`develop/headers/$effectiveTargetArchitecture`** for the devel package (e.g. check_x86_devel) of the secondary architecture.

I guess everything is in order now.
BTW, shall I remove (in this PR) the recipes for the old versions, **`0.9.13`** and **`0.9.14`**? (I haven't played with check yet.) Thanks for your review/feedback!

fbrosson:
> Sentence casing, please

@waddlesplash Sorry, I did not notice your very first message. I though about fixing the sentence casing before creating the PR, but when I saw that it was written this way in their own website I chose to leave it unchanged. Well, I'll fix it next time there is an upgrade.

PR #505, opened 2016-03-12
gnutls: bump version to 3.4.10.

fbrosson:
* Bump version to 3.4.10.
* ~~Install **`systemkey`** as **`systemkey-tool`** (to follow the man page name) as neither the command nor its manual page are being installed by default.~~
* Drop the **`--disable-doc`** switch during configuration and put all docs in a separate package, gnutls_doc.
* Add «**`= $portVersion compat >= 3.4`**» for each «**`cmd:`**» line in PROVIDES.
* Swap the lines for «**`cmd:psktool`**» and «**`cmd:ocsptool`**» as they were not sorted.
* Drop the «**`cmd:systemkey`**» line in PROVIDES as it is not installed by default.

### DIFF
--- a/net-libs/gnutls/gnutls-3.4.10.recipe
+++ b/net-libs/gnutls/gnutls-3.4.10.recipe
@@ -12,7 +12,7 @@ COPYRIGHT="2000-2016 Free Software Fundation Inc.
 LICENSE="GNU LGPL v2.1"
 REVISION="1"
 SOURCE_URI="ftp://ftp.gnutls.org/gcrypt/gnutls/v3.4/gnutls-$portVersion.tar.xz"
-CHECKSUM_SHA256="48594fadba33d450f796ec69526cf2bce6ff9bc3dc90fbd7bf38dc3601f57c3f"
+CHECKSUM_SHA256="6a32c2b4acbd33ff7eefcbd1357009da04c94c60146ef61320b6c076b1bdf59f"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
 SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
@@ -23,12 +23,11 @@ PROVIDES="
 	cmd:gnutls_cli$secondaryArchSuffix
 	cmd:gnutls_cli_debug$secondaryArchSuffix
 	cmd:gnutls_serv$secondaryArchSuffix
-	cmd:psktool$secondaryArchSuffix
 	cmd:ocsptool$secondaryArchSuffix
+	cmd:psktool$secondaryArchSuffix
 	cmd:srptool$secondaryArchSuffix
-	cmd:systemkey$secondaryArchSuffix
 	lib:libgnutls_openssl$secondaryArchSuffix = 27.0.2 compat >= 27
-	lib:libgnutls$secondaryArchSuffix = 30.6.1 compat >= 30
+	lib:libgnutls$secondaryArchSuffix = 30.6.2 compat >= 30
 	lib:libgnutlsxx$secondaryArchSuffix = 28.1.0 compat >= 28
 	"
 REQUIRES="
@@ -42,9 +41,9 @@ REQUIRES="
 	"
 
 PROVIDES_devel="
-	gnutls_devel$secondaryArchSuffix = $portVersion
+	gnutls${secondaryArchSuffix}_devel = $portVersion
 	devel:libgnutls_openssl$secondaryArchSuffix = 27.0.2 compat >= 27
-	devel:libgnutls$secondaryArchSuffix = 30.6.1 compat >= 30
+	devel:libgnutls$secondaryArchSuffix = 30.6.2 compat >= 30
 	devel:libgnutlsxx$secondaryArchSuffix = 28.1.0 compat >= 28
 	"
 REQUIRES_devel="
@@ -52,6 +51,20 @@ REQUIRES_devel="
 	devel:libnettle$secondaryArchSuffix
 	devel:libtasn1$secondaryArchSuffix
 	"
+
+if [ -z "$secondaryArchSuffix" ]; then
+# HaikuPorter does not yet support creating architecture-independent packages
+# for secondary-arch builds. So build it only if we're on a primary arch.
+	SUMMARY_doc="Documentation for GnuTLS"
+	ARCHITECTURES_doc="any"
+
+	PROVIDES_doc="
+		gnutls_doc = $portVersion
+		"
+	REQUIRES_doc="
+		haiku
+		"
+fi
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
@@ -75,8 +88,9 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	runConfigure ./configure --without-p11-kit --disable-doc --disable-nls \
-		--enable-openssl-compatibility
+	runConfigure ./configure --without-p11-kit --disable-nls \
+		--enable-openssl-compatibility \
+		${secondaryArchSuffix+--disable-doc}
 	make $jobArgs
 }
 
@@ -88,4 +102,8 @@ INSTALL()
 	fixPkgconfig
 
 	packageEntries devel $developDir
+
+	if [ -z "$secondaryArchSuffix" ]; then
+		packageEntries doc $documentationDir
+	fi
 }


fbrosson:
If tried to make the gnutls_doc package architecture-independent (using **`any`** at some places and some other changes)... but I did not manage to find the right syntax for that. Help, hints or pointers are all  welcome. Thanks!
BTW, I guess we should not be splitting ordinary packages, but this one is special because it has a lot of docs and we were not providing them at all. I hope it's OK, _for gnutls_, to provide the docs in a separate package.

waddlesplash:
> Install systemkey as systemkey-tool (to follow the man page name) as neither the command nor its manual page are being installed by default.

Is this what Linux does?

fbrosson:
According to https://www.archlinux.org/packages/core/x86_64/gnutls/files/ ArchLinux does not include  systemkey-tool in their package.  I though it was safer to install it (rather than just drop it) because PROVIDES  had a line with **`cmd:systemkey`**. Oops, I forgot to append the **`-tool`** suffix for it in thatline so this means the recipe is not yet 100% correct.
Well, shall I simply remove that systemkey tool (and get rid of the special install steps)?
Or would you prefer to have it in the package? I can fix the recipe easily. The only drawback is that the recipe is a little bit uggly with it.

fbrosson:
I added the **`_tool`** suffix to **`cmd:systemkey `** but there is still a tiny problem:
```
systemkey-tool-x86 --version
/bin/systemkey-tool-x86: error: `/bin/x86/.libs/systemkey' does not exist
This script is just a wrapper for systemkey.
See the libtool documentation for more information.
```

It looks like I can't simply rename a runtime after compilation. I need to change its name in the Makefile, or rename (and fix) the man page, but that second choice is not desirable because systemkey is not a very good name name.
I'm trying to fix this issue right now. So if you choose to keep that tool in the package then please don't merge yet.
If you choose to not include that tool, then I can update the recipe accordingly very easily. Just let me know your choice.

waddlesplash:
> I'm trying to fix this issue right now. So if you choose to keep that tool in the package then please don't merge yet.

We have no intention of renaming binaries for no reason other than "it's a bad name"; so unless any other Linux distro also renames it, please install it with the original name.

fbrosson:
Oops, I pushed an update right after your reply. I had removed the systemkey tool, but I can add it back with its original name.
I'm pretty sure, however, that nobody is including that tool simply because **`make install`** does not install it.
Shall I add it back with its original name, systemkey? Warning: in that case it might be appropriate to rename the manual page because it is available as **`systemkey-tool.1`** in the source tarball. I'll also need to remplace each occurrence of "**`syetmkey-tools`** by "**`systemkey`**, taking care of special escape sequences. Let me know what you would like to have.

pulkomandy:
What you copied is not the actual binary, but a libtool wrapper. With libtool messing with your buildsystem, it will be nearly impossible to install the binaries by copying manually (wrong library paths will be used, etc).
If the upstream decided to not install the tool with make install, I don't see why we should. If the tool is useful, make install should be fixed to do the right thing and the patch sent upstream.

fbrosson:
@pulkomandy Thanks for your message. I tried to figure out what should be changed to have **`make install`** install systemkey and its manual page, but it does not seem easy to do. So if someone can guide me I'll be happy to add systemkey to the package.

fbrosson:
I found a recipe that creates (or tries to create) both an architecture-dependent package and an architecture-**in**dependent one: **`games-rpg/freedroidrpg/freedroidrpg-0.15.1.recipe`**
BTW, it is the only one that tries to do so. So I started doing tests in a playground with **`app-cdr/uif2iso/uif2iso-0.1.7c.recipe`** and **`app-cdr/daa2iso/daa2iso-0.1.7e.recipe`** which are both very small and therefore do not take long to build. Adding:
```Shell
SUMMARY_doc="Doc for DAA2ISO"
ARCHITECTURES_doc="any"

PROVIDES_doc="
	daa2iso_doc = $portVersion
	"
REQUIRES_doc="
	haiku
	"
```
to the recipe of daa2iso (and calling haikuporter daa2iso) creates **`daa2iso-0.1.7e-2-x86_gcc2.hpkg`** + **`daa2iso_doc-0.1.7e-2-any.hpkg`** as expected.

Unfortunately, adding:
```Shell
SUMMARY_doc="Doc for UIF2ISO"
ARCHITECTURES_doc="any"
#SECONDARY_ARCHITECTURES_doc=""

PROVIDES_doc="
	uif2iso${secondaryArchSuffix}_doc = $portVersion
	"
REQUIRES_doc="
	haiku
	"
```
to the recipe of uif2iso (and calling haikuporter uif2iso_x86) builds architecture-dependent packages, **`uif2iso_x86-0.1.7c-2-x86_gcc2.hpkg`** and **`uif2iso_x86_doc-0.1.7c-2-x86_gcc2.hpkg`**, instead of makeing the doc package architecture-**in**dependent.
I did add **`packageEntries doc $documentationDir`** in the **`INSTALL()`** of both recipes. The only difference is that daa2iso has no secondary architectures, while uif2iso has.
I've also tried several variations in uif2iso to try and make it build a real architecture-independent package for the doc package but haven't found yet the correct method.
I did not try, however, to build freedroidRPG-0.15.1 because the source tarball is 200 MB.
Maybe haikuporter only supports creating architecture-independent packages for primary architectures. Does anyone know what I am missing for making architecture-independent doc packages when building for a secondary architecture? Thanks!
Warning: **`gnutls-3.4.10`** is currently **WORK IN PROGRESS** because I'm trying to fix the doc package.

pulkomandy:
Is that needed? If the doc package is useable for any architecture, then there is no need to build it twice, for primary and secondary. So, we only need to install the docs to the package for the primary arch, and the secondary arch should come with no docs at all (you can see how to handle this in the ffmpeg recipes, IIRC).

fbrosson:
I know we should only build the doc package once. The problem is that it is currently being marked as arch-dependent. So we would not be able to install a doc fox x86 on a x86_64 until we get rid of the wrong arch-dep in the doc package.
Hmm, your comment gives me an idea. Can I hack **`BUILD()`** and **`INSTALL()`**  to pretend the package accepts a primary-arch build? And in that case only create the doc package and a meta main package?
The problem

pulkomandy:
The trick would be that on at least one architecture, the package can be built on the primary arch. We get the doc package from there, and don't generate it in the secondary arch case.

Then we use that single doc package for all repos.

fbrosson:
What do you think of the new recipe? It's a bit hacky because I wanted to make it possible to create the architecture-independent gnutls_doc on any platform, event on x86_gcc2 (hybrid).
It does not require to copy the doc package built on an x86 or x86_64 arch to populate the repository of an x86_gcc2 hybrid repo. All it needs is a way to tell that, on x86_gcc2, we need to build the package for both the primary x86_gcc2 arch and the secondary x86 arch. The recipe states that on x86_gcc2 the gnutls package requires gnutls_x86, so a smart build farm will be able to figure out that it needs to build both of them.
I do not know, however, if the recipe is politicaly acceptable. Well, there it is. It might look a little bit complicated but I have included a few comments in the recipe to explain what it does when it is not trivial. Comments are welcome as usual. Thanks!

korli:
I find this recipe very messy now, we shouldn't workaround things like this (plus for a doc package nobody will probably install)
Please open a bug report for the problem on haikuporter. As Pulkomandy writes, add a comment in the recipe that the doc package is only buildable on a primary arch ATM.
"Add «= $portVersion compat >= 3.4» for each «cmd:» line in PROVIDES." what tells you that commands are only compatible with 3.4.x?

fbrosson:
> I find this recipe very messy now, we shouldn't workaround things like this (plus for a doc package nobody will probably install)

OK. Fixed now.
> Please open a bug report for the problem on haikuporter. As Pulkomandy writes, add a comment in the recipe that the doc package is only buildable on a primary arch ATM.

OK, comment added. For the bug report I'll open an issue in https://github.com/haikuports/haikuports/issues soon.

Regarding the versioning for the **`cmd:`** lines I though it was sane to add **compat >= 3.4`** because we don't have any other recipes depending on any of these commands. But I guess I missed something, so I reverted my changes. But please let me know if I should add back something.

Regarding the systemkey command, I had not noticed, until @pulkomandy spotted it, that it was a libtool wrapper. Since then I have found, however, that after compilation we get an ELF executable for the systemkey tool in **`src/.libs/`**. So I added it to the package with its original name, **`systemkey`**. I did not restore, however, the **`cmd:systemkey`** line in PROVIDES. Should I do so?
BTW, the recipe currently installs **`systemkey-tool.1`** and **`systemkey`**, without renaming any of them. The output of **`systemkey --help`** shows that it is what **`systemkey-tool.1`** calls systemkey-tool. Should we either rename **`systemkey`** as **`systemke-tool`** or  **`systemkey-tool.1`** as **`systemkey.1`** ? Or is it OK to leave both of them as is, with mismatching names?

pulkomandy:
I will repeat myself here: DO NOT copy binaries manually to install them.

If you think the binaries should be installed, cleanly patch the software build system to do so, in a way that makes it possible to send your patch upstream to gnutls developers (so edit Makefile.am to add it to the list of things to be installed, most likely).

Otherwise, do not install the binaries. There probably is a reason for the makefile to not install them.

fbrosson:
Sorry, I though it was OK to just copy the files because the way the build system is configured shows that upstream has no intention to make systemkey part of the default installation. And as you say, upstream might even have a good reason for that.
I have therefore fixed the recipe to not include it anymore.
In case someone knows how to install the systemkey tool with a clean method (and if there is concensus for doing so) then please let me know.

fbrosson:
Minor improvement: use **`--disable-doc`** (for secondary arch builds) instead of removing **`$documentationDir`**.
I think everything is OK now.

fbrosson:
Regarding the tests to check if we are building for a primary vs. secondary arch, may I replace, for readability purposes only,
```Shell
[ -n "$secondaryArchSuffix" ]
```
by
```Shell
[ "$targetArchitecture" != "$effectiveTargetArchitecture" ]
```
and
```Shell
[ -z "$secondaryArchSuffix" ]
```
 by
```Shell
[ "$targetArchitecture" = "$effectiveTargetArchitecture" ]
```
?
I know both pairs of expressions are currently equivalent. I was just wondering whether there was a best pratice on this topic. Thanks and sorry for being so talkative ;-)

korli:
What "readability purposes" actually means? If you need to check two variables for such checks, then it's not more readable.

fbrosson:
You are right, it's not more readable. In fact I was thinking about changing the tests to make them a little bit more explicit, in particular for novice packagers. But it's fine for me the way it is now, as I'm already used to the compact style.

net-libs/gnutls/gnutls-3.4.10.recipe
```diff
@@ -44,7 +43,7 @@ REQUIRES="
 PROVIDES_devel="
 	gnutls_devel$secondaryArchSuffix = $portVersion
```
korli: This should be gnutls${secondaryArchSuffix}_devel

net-libs/gnutls/gnutls-3.4.10.recipe
```diff
@@ -75,8 +88,14 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	runConfigure ./configure --without-p11-kit --disable-doc --disable-nls \
-		--enable-openssl-compatibility
+	if [ -n "$secondaryArchSuffix" ]
+	then
```
korli: Keep "then" on the previous line, to be consistent with the style of other ifs.

net-libs/gnutls/gnutls-3.4.10.recipe
```diff
@@ -75,8 +88,14 @@ BUILD_PREREQUIRES="
...
+		runConfigure ./configure --without-p11-kit --disable-nls \
+			--enable-openssl-compatibility --disable-doc
+	else
+		runConfigure ./configure --without-p11-kit --disable-nls \
+			--enable-openssl-compatibility
+	fi
```
korli: You could define a variable with "--disable-doc" for secondary Arch, and use this variable with configure.
This way you don't have two copies of runConfigure.

fbrosson: Instead of having to chose a variable name, would it be acceptable to use **`${secondaryArchSuffix+--disable-doc}`** ?
```

fbrosson:
Oops, my previous https://github.com/haikuports/haikuports/pull/505#discussion_r56123592 is hidden  since I've updated the commit with a change in **`BUILD()`** about the duplicate runConfigure.
BTW, the recipe is ready for a review.

PR #509, opened 2016-03-16
Fix recipe errors & convert all old recipes

DarkmatterVale:
Hi,

This is a pretty big commit, and I'd like to apologize about that. In this PR, I have done a few things:

1) Add dummy components for parts of recipes that are required but not in the recipe
2) Convert old recipes to the new format

This is a work-in-progress, so thoughts and suggestions are greatly appreciated!


pulkomandy:
I would not remove the MD5 checksums. Ideally, a tool would:
- Download the file at the given URL
- Compute the MD5 and see if it still matches
- Compute the SHA256
- Replace the MD5 with the SHA256.
This will need a lot of time to download everything, so maybe no need to do it now. However, if the MD5 are removed we have no way to know which file the recipe was built from (and sometimes we have to discover a mirror, because the file is not hosted at the original location, especially for old recipes).

I would maybe only add the "warnings" as a comment and not include the empty entries, as they don't help much in fixing the recipe, but still make it look as if it was fixed.

STATUS_HAIKU: this should be converted to ARCHITECTURES="x86_gcc2" (stable), !x86_gcc2 (broken), ?x86_gcc2 (unknown). It seems you have just removed it.

app-editors/vim/vim-7.4.recipe looks wrong. Fix that bug and come back! :)

DarkmatterVale:
Thanks for the feedback @pulkomandy. I will work on getting these errors fixed...

DarkmatterVale:
I think I have fixed all of the errors that you identified @pulkomandy

app-misc/googlemaps_mini/googlemaps_mini-0.1.recipe
```diff
@@ -15,6 +16,9 @@ PROVIDES="
 	app:googlemaps_mini = $portVersion
 	"
 REQUIRES="
+	"
+	haiku$secondaryArchSuffix
+	
 	haiku${secondaryArchSuffix}
```
diversys: Should probably be removed everywhere

DarkmatterVale: Should what I added be removed or the other one?

diversys: I guess the other one along with an extra line above it

app-arch/zoo/zoo-2.10.recipe
```diff
@@ -5,14 +5,31 @@ which compresses files to 20-80% of their original size depending on the data ty
...
+LICENSE="zoo"
+REVISION="1"
 SOURCE_URI="ftp://ftp.kiarchive.ru/pub/unix/arcers/zoo-2.10pl1.tar.gz"
 CHECKSUM_MD5="70a0a785ecc8f7f747be1421e495d837"
-REVISION="1"
-STATUS_HAIKU="stable"
-DEPEND=""
+# WARNING: CHECKSUM_SHA256 must be added to recipe here
+
+ARCHITECTURES="x86_gcc2"
```
korli: This recipe isn't actually tested, right? It should be ?x86_gcc2 then.

app-emulation/oricutron/oricutron-1.1_svn.recipe
```diff
@@ -18,7 +18,9 @@ PROVIDES="
 REQUIRES="
 	haiku
 	libsdl
+	libsdl >= 1.2.10
```
korli: This looks wrong anyway. It's probably not worth reusing DEPEND, just commenting it, or dropping it.

app-misc/figlet/figlet-222.recipe
```diff
@@ -11,8 +11,23 @@ LICENSE="AFL v2.1"
 REVISION="1"
 SOURCE_URI="ftp://ftp.figlet.org/pub/figlet/program/unix/figlet222.tar.gz"
 CHECKSUM_MD5="5f84ad52d092d5db4ad5916df767646b"
-STATUS_HAIKU="stable"
-DEPEND=""
+# WARNING: CHECKSUM_SHA256 must be added to recipe here
+
+ARCHITECTURES="x86_gcc2"
```
korli: Untested.

app-office/abiword/abiword-2.4.recipe
```diff
@@ -19,6 +20,9 @@ PROVIDES="
 	app:AbiWord
 	"
 REQUIRES="
+	"
```
korli: Looks wrong.

app-text/convmv/convmv-1.15.recipe
```diff
@@ -1,10 +1,30 @@
...
+COPYRIGHT="2003-2011 Bjoern JACKE"
+LICENSE="GNU GPL v2"
 REVISION="1"
-STATUS_HAIKU="stable"
-DEPEND="dev-lang/perl >= 5.8"
+SOURCE_URI="https://www.j3e.de/linux/convmv/convmv-1.15.tar.gz"
 CHECKSUM_MD5="b1bb703c08c6355868d15890ff193f7d"
+# WARNING: CHECKSUM_SHA256 must be added to recipe here
+
+ARCHITECTURES="x86_gcc2"
```
korli: Untested

dev-cpp/hugen/hugen-1.recipe
```diff
@@ -21,6 +21,9 @@ PROVIDES="
 	cmd:hugen$secondaryArchSuffix = $portVersion
 	"
 REQUIRES="
+	"
```
korli: Looks wrong

net-misc/wput/wput-0.6.1.recipe
```diff
@@ -1,10 +1,29 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
 SOURCE_URI="http://voxel.dl.sourceforge.net/sourceforge/wput/wput-0.6.1.tgz"
 CHECKSUM_MD5="92b41efed4db8eb4f3443c23bf7ceecf"
-REVISION="1"
-STATUS_HAIKU="stable"
-DEPEND=""
+# WARNING: CHECKSUM_SHA256 must be added to recipe here
+
+ARCHITECTURES="x86_gcc2"
```
korli: Untested

net-misc/tor/tor-0.2.3.25.recipe
```diff
@@ -1,11 +1,32 @@
...
+LICENSE="Zlib"
 REVISION="1"
-STATUS_HAIKU="stable"
-DEPEND="dev-libs/openssl >= 0.9.8
-		dev-libs/libevent >= 2.0.10"
+SOURCE_URI="https://www.torproject.org/dist/tor-0.2.3.25.tar.gz"
 CHECKSUM_MD5="a1c364189a9a66ed9daa8e6436489daf"
+# WARNING: CHECKSUM_SHA256 must be added to recipe here
+
+ARCHITECTURES="x86_gcc2"
```
korli: Untested

dev-vcs/mercurial/mercurial-3.2.1.recipe
```diff
@@ -16,6 +16,9 @@ PROVIDES="
 	cmd:hg = $portVersion compat >= 3
 	"
 REQUIRES="
+	"
```
korli: Looks wrong

dev-db/redis/redis-2.2.11.recipe
```diff
@@ -1,15 +1,32 @@
...
-"
-SOURCE_URI="http://redis.googlecode.com/files/redis-2.2.11.tar.gz"
+and sorted sets."
+HOMEPAGE="http://redis.io/"
+COPYRIGHT="2006-2010 Salvatore Sanfilippo"
+LICENSE="BSD (3-clause)"
 REVISION="1"
-DEPEND=""
+SOURCE_URI="http://redis.googlecode.com/files/redis-2.2.11.tar.gz"
 CHECKSUM_MD5="5501c90ca02ae7ea93c34f067b3e0619"
```
diversys: Should be converted to SHA if possible

pulkomandy: Not easily: it would require downloading the source file, checking that the MD5 checksum still matches, then computing the SHA256 checksum from the downloaded data.
Maybe it would be better to have the tool integrated in haikuporter or distributed separately ("haikuporter --modernize recipe" or similar) instead of trying to batch-fix everything at once. This way, it would be possible to make such more complex fixes.

DarkmatterVale: Yes I agree. We should leave checksums and more complicated fixes out of this.

net-misc/wput/wput-0.6.1.recipe
DarkmatterVale: @pulkomandy What do you think? Should we change it to untested?

pulkomandy: As I mentionned in my other comment, I now think it may not be a good idea to batch-apply the script this way. I think it would be more useful to:
1) Have a list of files that need changes (this pull request does that for now)
2) Have a way to run the script on a single recipe at a time, so we can then manually adjust the result and submit it.

Unless you can make the script good enough that it can test the recipe itself by running Haikuporter and making sure there are no errors left, in that case an automated run may be considered.

DarkmatterVale:
@pulkomandy Would you like me to close this then? Further, would you like me to work on integrating my tool into HaikuPorter?

DarkmatterVale:
bump...

waddlesplash:
Yeah, let's close this, and just continue doing it manually.

PR #510, opened 2016-03-17
parted: bump to version 3.2.

fbrosson:
* Bump to version 3.2.
* Drop patch «**`replace deprecated CPPFunction by rl_completion_func_t`**» as it has been applied upstream.
* Add patch to insert «**`AC_CONFIG_MACRO_DIR([m4])`**» in configure.ac.
* Add patch to use «**`-lroot`**» instead of «**`-ldl`**».
* Don't pass «**`--disable-dynamic-loading`**» to configure script as it is now able to find dlopen in libroot.
* Patch lib/gnulib.mk to skip the installtion of lib/charset.alias.
* Pass «**`--enable-debug`**» to configure script to ease troubleshooting.
* Add debuginfo package and a dirty hack to workaround a haikuporter bug that makes defineDebugInfoPackage fail if given a file that has dashes in its name.
* Change broken to untested for x86{,_64} as it builds fine on these architectures and even seems to work fine on x86 primary arch.


fbrosson:
On x86 I can list the partition table of the USB key on which I boot Haiku:
```
parted /dev/disk/usb/0/0/raw 
GNU Parted 3.2
Using /dev/disk/usb/0/0/raw
Welcome to GNU Parted! Type 'help' to view a list of commands.
(parted) unit MiB
(parted) print                                                            
Model:  (ide)
Disk /dev/disk/usb/0/0/raw: -2275109032850.20MiB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags: 

Number  Start    End      Size     Type     File system  Flags
 1      4.00MiB  1204MiB  1200MiB  primary
 2      1204MiB  2404MiB  1200MiB  primary
 3      2404MiB  3604MiB  1200MiB  primary               boot
 4      3604MiB  7640MiB  4036MiB  primary
```
**But be carefull:** there are some issues with arithmetics, as you can see in the negative size reported (eventhough the real size is only about 8 GB).
And when I list the partitions of a 16 GB microSD with an extended partition, I get a funny warning:
```
parted /dev/disk/usb/1/2/raw 
GNU Parted 3.2
Using /dev/disk/usb/1/2/raw
Welcome to GNU Parted! Type 'help' to view a list of commands.
(parted) unit MiB
(parted) p                                                                
Error: Can't have a logical partition outside of the extended partition on /dev/disk/usb/1/2/raw.
Ignore/Cancel? i                                                          
Model:  (ide)
Disk /dev/disk/usb/1/2/raw: -2275109032850.20MiB
Sector size (logical/physical): 512B/512B
Partition Table: msdos
Disk Flags: 

Number  Start    End       Size      Type      File system  Flags
 1      1.00MiB  32.0MiB   31.0MiB   primary   fat16
 2      32.0MiB  63.0MiB   31.0MiB   primary   fat16
 3      63.0MiB  287MiB    224MiB    primary
 4      287MiB   15279MiB  14992MiB  extended
 5      288MiB   290MiB    2.00MiB   logical   ext4
 6      291MiB   307MiB    16.0MiB   logical   fat16
 7      308MiB   1076MiB   768MiB    logical   ext4
```
So do not try to edit the partition table of a device if it contains valuable data.
I'll try on an empty USB key when I find one that is available for testing.

mmuman:
Nice to see it's finally building.
The negative size is probably some overflow somewhere or bad code doing sign extension.

fbrosson:
Thanks! Yep, that's cool. I've started looking at the code that prints the partition table but haven't been very far yet. But I guess you are right about the possible causes. Let's hope someone will find what exactly needs to be fixed.
BTW, if/when we manage to make parted work on all architecuters it could be interesting to enhance DriveSetup by making it able to use libparted and libparted-fs-resize.

sys-block/parted/parted-3.2.recipe
```diff
@@ -0,0 +1,144 @@
...
+		defineDebugInfoPackage parted \
+			$binDir/parted \
+			$binDir/partprobe \
+			$libDir/libparted.so.2.0.1 \
+			$LIBPARTED_FS_RESIZE_FILENAME
+fi
+
+if [ "$USE_DEBUG_INFO_FOR_PARTED" = all_with_workaround ]; then
+# defineDebugInfoPackage does not produce correct {PROVIDES,REQUIRES}_debuginfo
+# when a target has dashes in its name. So we need to override these.
```
waddlesplash: We probably should fix this instead of having a workaround -- cc @DarkmatterVale, can you look into this?

DarkmatterVale: Sure, I will look into this. I'm on a school trip this weekend, but I will have plenty of time all of next week.

sys-block/parted/parted-3.2.recipe
```diff
@@ -0,0 +1,144 @@
...
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autom4te
+	cmd:automake
+	cmd:autoreconf
+	cmd:awk
+	cmd:gawk
+	cmd:grep
+	cmd:iconv
+	cmd:libtool
```
korli: Should be cmd:libtoolize$secondaryArchSuffix

fbrosson: Fixed, thanks.

korli:
Could you drop the debugInfoPackage for the time being? Add a comment that it needs the bug number xxx fixed in Haikuporter first. This way we don't overbloat the recipe and can go forward with this PR.

fbrosson:
I found a much smaller hack (3 lines) which allows to keep the debuginfo package. I have updated the recipe for that. Is it acceptable?
BTW, I haven't created a report for the issue about defineDebugInfoPackage not handling files with dashes in their names. I'll do so.
BTW, I think I found another bug: **`defineDebugInfoPackage parted ...`** generates a **`REQUIRES_debuginfo="parted == $portVersion base"`** and this is OK for primary arch builds. Unfortunately, for secondary arch builds it does not define **`REQUIRES_debuginfo="parted$secondaryArchSuffix == $portVersion base"`** even if we call **`defineDebugInfoPackage parted$secondaryArchSuffix ...`**.
In case having the hack in the recipe is not an option, may I still define the debuginfo package without **`libparted-fs-resize.so`** ? Thanks!

fbrosson:
After all, let's just drop the debuginfo package completely. There is no need to add noise to the repo with a partial debuginfo package or with temporary hacks . It will be much better to add back a full and clean debuginfo package once the issues on haikuporter are fixed.

PR #511, opened 2016-03-18
pkgconfig: fix 0.27.1, 0.28 and 0.29, add 0.29.1.

fbrosson:
* Add recipe for version 0.29.1.
* Use the runConfigure wrapper.
* Use exact copy of the patchset for 0.29 as there are no offset mismatches.

For 0.27.1:
* Call configure wuth «**`--with-pc-path`**» instead of «**`--with-pc_path`**».

For 0.28:
    * Several changes to make it build and look almost like 0.29.1.

For 0.29:
* Switch to runConfigure in order to prevent pkg-config on secondary arch from shadowing that of the primary arch.

For 0.27.1, 0.28, 0.29 and 0.29.1:
* Fix COPYRIGHT.
* Update HOMEPAGE and SOURCE_URI to avoid «301 Moved Permanently».

### DIFF
--- a/dev-util/pkgconfig/patches/pkgconfig-0.29.1.patchset
+++ b/dev-util/pkgconfig/patches/pkgconfig-0.29.1.patchset
@@ -1,0 +1,78 @@
+From 6e759ba387f142a59edb1b3f70f1a2b3c710b3ab Mon Sep 17 00:00:00 2001
+From: Jerome Duval <jerome.duval@gmail.com>
+Date: Sun, 29 Nov 2015 15:49:57 +0000
+Subject: Haiku patch
+
+* Actually use pc_path when building pkgconfig.
+* skip network libs in glib.
+
+diff --git a/Makefile.am b/Makefile.am
+index e46c798..b431bcb 100644
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -22,7 +22,7 @@ AM_CPPFLAGS = \
+ 	-DPKG_CONFIG_SYSTEM_LIBRARY_PATH="\"$(system_library_path)\""
+ endif
+ 
+-AM_CFLAGS = \
++AM_CFLAGS = $(AM_CPPFLAGS)\
+ 	$(WARN_CFLAGS) \
+ 	$(GCOV_CFLAGS) \
+ 	$(GLIB_CFLAGS)
+diff --git a/glib/configure.ac b/glib/configure.ac
+index ffa7869..48f7537 100644
+--- a/glib/configure.ac
++++ b/glib/configure.ac
+@@ -1094,49 +1094,6 @@ AS_IF([test $glib_native_win32 = yes], [
+                                          [AC_MSG_ERROR([could not compile test program either way])])])])
+   AC_SUBST(NAMESER_COMPAT_INCLUDE)
+ 
+-  # We can't just use AC_CHECK_FUNC/AC_CHECK_LIB here. Bug 586150
+-  NETWORK_LIBS=""
+-  AC_MSG_CHECKING([for res_query])
+-  AC_TRY_LINK([#include <sys/types.h>
+-	       #include <netinet/in.h>
+-	       #include <arpa/nameser.h>
+-	       #include <resolv.h>
+-              ],[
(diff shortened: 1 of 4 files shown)


mmuman:
Btw, the current recipe for pkg-config seems buggy, it doesn't pass the binDir so the gcc4 version gets installed in /bin as well :^)

fbrosson:
Hmm, runConfigure takes care of passing all *Dir parameters, so the pkgconfig_x86 package on x86_gcc2 
has:
* **`bin/pkg-config-x86 -> x86/pkg-config`**
* **`bin/x86/pkg-config`**

which is OK and does not conflict with the package for the primary arch.
There was a problem, however, with some files:
* **`data/aclocal/pkg.m4`**
* **`documentation/man/man1/pkg-config.1`**

They were being installed for the secondary arch at the same places as the package for the primary arch, so I  improved the recipe in order to drop both **`$dataDir`** and **`$manDir`** for the secondary arch.
There was a third file:
* **`documentation/packages/pkgconfig_x86/pkg-config-guide.html`**

which was a duplicate of
* **`documentation/packages/pkgconfig/pkg-config-guide.html`**

So I also dropped **`$docDir`** for the secondary arch. Obviously, I added the required dependency (secondary arch requires primary arch).
So thanks @mmuman for your message because you made me notice there were duplicate files I had forgotten to remove from the secondary arch package.
I think the recipe is OK now.
BTW, @korli and @waddlesplash , if you wish I can add the **`rm -rf $docDir $dataDir $manDir`** for the secondary arch in the recipes for the older versions with the required dependency.

fbrosson:
@mmuman You were right about the pkg-config for gcc4. When built on x86_gcc2 for x86, the recipe for version 0.29 was installing pkg-config in bin instead of bin/x86, shadowing the pkg-config for the primary arch. I have therefore fixed this by switching 0.29 to runConfigure.
Regarding the issues about the duplicate installation of docDir, dataDir and manDir, I fixed them in 0.27.1 and 0.29 using the same method I used for 0.29.1.
BTW, version 0.28 was broken, so I dropped it.

To all: With this PR all recipes of pkgconfig will be fixed. 
BTW, pkg-config 0.29.1 fixes some issues in 0.29,
Please review. Thanks!

fbrosson:
I've just rebased the commit against master to speed-up the merge (if someone dares merging ;-)
BTW, as a discussion about package spliting showed that we should not force version synchronisation between primary and secondary architectures (except for ports like [dev-libs/check](https://github.com/haikuports/haikuports/tree/master/dev-libs/check) for which such version synchronisation enforcing is not a problem), I reverted the changes in the split layout for pkgconfig.
Please do not hesitate to tell if I missed something. Thanks!

dev-util/pkgconfig/pkgconfig-0.28.recipe
```diff
@@ -24,6 +24,11 @@ REQUIRES="
 	haiku$secondaryArchSuffix
 	lib:libiconv$secondaryArchSuffix
 	"
+if [ -n "$secondaryArchSuffix" ]; then
+	REQUIRES="$REQUIRES
+		pkgconfig == $portVersion base
+		"
+fi
```
korli: Why adding this in 0.28?

fbrosson: Oops, my mistake. Fixed, thanks!

fbrosson:
Thanks @korli for the merge.
BTW, the recent changes in the way PRs are merged is really cool, since it does not need the "Merge pull request ... from ..." empty commits anymore.
If the inspiration for it came from your observation of the "copy" I did of a commit TwoFX had made for PR 429 in a repository he has since deleted, then I'm very pleased to have contributed to the new "merge style" even my "copying/forging" of the commit was just an experimentation and I had no idea of how this could be nicely integrated in a process. So congratulations for having dug the subject and improved things!

korli:
It's nice of you, but no, Github now offers the choice between squash and merge. I didn't do anything except merging with this new option.

fbrosson:
Hmm, I was not aware of that new option. Thanks for the info!

PR #513, opened 2016-03-21
vala: fix missing $secondaryArchSuffix in PROVIDES{,_common}.

fbrosson:
The **`$secondaryArchSuffix`** suffix was missing for vala in PROVIDES and for vala_common in PROVIDES_common. This is required on x86_gcc2 because vala_x86 requires vala_common_x86.


fbrosson:
Oops, I forgot to bump **`REVISION`** but I guess this was not necessary because:
1. on x86 and x86_64 the fix is neutral.
2. on x86_gcc2 **`vala_x86-0.30.1-1-x86_gcc2.hpkg`** could not be installed without the fix.
3. There are no pre-built packages of vala on the depot servers.

PR #514, opened 2016-03-21
expat: drop 2.0.1, fix 2.1.0, add 2.1.1.

fbrosson:
* Drop version 2.0.1.
* $manDir was both in primary and secondary arch packages, causing problems when uninstalling only one of them. Fixed by discarding $documentationDir for secondary arch.
* Reorder sections for 2.1.0.
* configure.in already has «**`AC_CONFIG_MACRO_DIR([m4])`**» so there is no need to append it in PATCH().
* Drop PATCH() and, instead, use a real patch to create Makefile.am otherwise /bin/haikuporter complains about untracked files.
* Add «**`$secondaryArchSuffix`**» to cmd:libtoolize in BUILD_PREREQUIRES.
* Add recipe for 2.1.1 with patch for gcc2.


korli:
What "problems when uninstalling" is this about?

fbrosson:
I'm not absolutely sure of it, but I think that if I have two packages installed, say foo and foo_devel, and if they overlap on one or more files then **`pkgman uninstall foo_devel`** is not enough to get rid of the foo_devel package; I have to manually delete the hpkg file in **`/boot/system/packages/`**. I did not double check this but it seems that the problem goes away if I change foo_devel to not include any file already in foo. Next time I hit the problem I'll keep track of it.

korli:
Never heard of it. Usually the reason we don't include docs in secondary arch packages is more to reduce the size of packages than anything else.

PR #515, opened 2016-03-22
gcab: create recipe.

fbrosson:


app-arch/gcab/gcab-0.7.recipe
```diff
@@ -0,0 +1,97 @@
...
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	aclocal
+	autoconf
+	automake --add-missing
```
korli: Have you considered reducing these with autoreconf?

app-arch/gcab/gcab-0.7.recipe
```diff
@@ -0,0 +1,97 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libintl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libglib_2.0$secondaryArchSuffix >= 0.4600.2
+	devel:libgettextlib$secondaryArchSuffix >= 0.14.1
+	intltool >= 0.40
+	"
+BUILD_PREREQUIRES="
+	intltool
+	gettext >= 0.19.6
```
korli: These should be avoided when possible, they provide packages, not commands required by the build.

fbrosson:
Thanks @korli for your feedback. I've made the changes.
Regarding the **`cmd:libtoolize$secondaryArchSuffix`** line, are you sure we need the suffix? Because I have noticed that most recipes which have support for secondary arch do not use the suffix for the **`cmd:libtoolize`** line.

korli:
cmd:libtoolize brings the package libtool, whereas you might want libtool_x86, you could end up linking against libltdl instead of libltdl_x86. It's about the same with cmd:gettext ending up linking against libintl.

fbrosson:
OK, thanks @korli for the explanation. Well, in that case, may I append the **`$secondaryArchSuffix`** suffix to the **`cmd:libtoolize`** of each recipe that supports at least one secondary architecture? Of course, I would use a single commit for all recipes.

PR #516, opened 2016-03-23
curl: bump to version 7.48.0

fbrosson:
* Bump version.
* Move BUILD_REQUIRES to the recommended place.
* Use **`cmd:libtoolize$secondaryArchSuffix`** instead of **`cmd:libtoolize`** in BUILD_PREREQUIRES to make sure we don't accidentally pull libintl from the primary arch when building for a secondary arch.

### DIFF
--- a/net-misc/curl/curl-7.48.0.recipe
+++ b/net-misc/curl/curl-7.48.0.recipe
@@ -1,16 +1,17 @@
 SUMMARY="A commandline-tool and library for downloading data from URLs"
 DESCRIPTION="Curl is a command line tool for transferring data with URL \
 syntax, supporting DICT, FILE, FTP, FTPS, Gopher, HTTP, HTTPS, IMAP, IMAPS, \
-LDAP, LDAPS, POP3, POP3S, RTMP, RTSP, SCP, SFTP, SMTP, SMTPS, Telnet and TFTP. curl supports SSL certificates, HTTP POST, HTTP PUT, FTP uploading, HTTP form \
+LDAP, LDAPS, POP3, POP3S, RTMP, RTSP, SCP, SFTP, SMTP, SMTPS, Telnet and TFTP.
+curl supports SSL certificates, HTTP POST, HTTP PUT, FTP uploading, HTTP form \
 based upload, proxies, cookies, user+password authentication (Basic, Digest, \
 NTLM, Negotiate, kerberos...), file transfer resume, proxy tunneling and a \
 busload of other useful tricks."
-HOMEPAGE="http://curl.haxx.se"
+HOMEPAGE="https://curl.haxx.se/"
 COPYRIGHT="1996-2016, Daniel Stenberg"
 LICENSE="Curl"
 REVISION="1"
-SOURCE_URI="http://curl.haxx.se/download/curl-$portVersion.tar.bz2"
-CHECKSUM_SHA256="ddc643ab9382e24bbe4747d43df189a0a6ce38fcb33df041b9cb0b3cd47ae98f"
+SOURCE_URI="https://curl.haxx.se/download/curl-$portVersion.tar.bz2"
+CHECKSUM_SHA256="864e7819210b586d42c674a1fdd577ce75a78b3dda64c63565abe5aefd72c753"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
 SECONDARY_ARCHITECTURES="x86_gcc2 x86"
@@ -31,11 +32,6 @@ REQUIRES="
 	lib:libssl$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
-BUILD_REQUIRES="
-	haiku${secondaryArchSuffix}_devel
-	devel:libssl$secondaryArchSuffix
-	devel:libz$secondaryArchSuffix
-	"
 
 PROVIDES_devel="
 	curl${secondaryArchSuffix}_devel = $portVersion
@@ -48,11 +44,16 @@ REQUIRES_devel="
 	devel:libz$secondaryArchSuffix
 	"
 
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libssl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
 BUILD_PREREQUIRES="
 	cmd:autoconf
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
-	cmd:libtoolize
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	cmd:nroff
 	cmd:python


net-misc/curl/curl-7.48.0.recipe
```diff
@@ -1,16 +1,18 @@
...
 busload of other useful tricks."
-HOMEPAGE="http://curl.haxx.se"
+HOMEPAGE="https://curl.haxx.se/"
 COPYRIGHT="1996-2016, Daniel Stenberg"
 LICENSE="Curl"
 REVISION="1"
-SOURCE_URI="http://curl.haxx.se/download/curl-$portVersion.tar.bz2"
-CHECKSUM_SHA256="ddc643ab9382e24bbe4747d43df189a0a6ce38fcb33df041b9cb0b3cd47ae98f"
+SOURCE_URI="https://curl.haxx.se/download/curl-$portVersion.tar.bz2"
+#CHECKSUM_SHA256="ddc643ab9382e24bbe4747d43df189a0a6ce38fcb33df041b9cb0b3cd47ae98f"
```
diversys: Forgot to remove?

fbrosson: No, it's just that I first want to rebuild 7.47.1 locally with the new package split (which includes 2 arch-independent "common" and "common_devel" packages) to make sure future upgrades won't be affected by the new split layout.
So I'll install locally all 7.47.1 packages, then play with pkgman install and uninstall, and then try to upgrade all or a subset. Once I see that upgrades are OK with the new split layout I'll drop the SHA256 for 7.47.1 and put a nice commit message for directly 7.48.0. I'll also check that the upgrade from 7.45.0 to 7.48.0 works fine.
(The reason why I pushed it in WIP state was to save others a duplicate job ;-)

korli: What's your intention actually? We had a discussion some time ago, some people wanted a single package per application or library, we settled on two packages devel/non-devel for reasons. Splitting packages should really be justified..

fbrosson: There are a few problems with the current layout because there are a lot of html files in **`develop/documentation/curl/`** and they are both in curl_devel and curl_x86_devel.

The split layout I'm proposing is described in the following table, where:
* 1=primary arch
* 2=secondary arch
* 1dev=primary arch devel
* 2dev=secondary arch devel

pkg | files | moved to
--- | ----- | -----------
1 |	bin/curl
1 |	data/aclocal/libcurl.m4			| common
1 |	data/licenses/Curl			| common
1 |	documentation/man/man1/curl-config.1	| common
1 |	documentation/man/man1/curl.1		| common
1 |	documentation/packages/curl/curl.html	| common
1 |	lib/libcurl.so.4
1 |	lib/libcurl.so.4.4.0
||
2 |	data/aclocal/libcurl.m4		| common
2 |	data/licenses/Curl		| common
2 |	lib/x86/libcurl.so.4
2 |	lib/x86/libcurl.so.4.4.0
||
1dev |	bin/curl-config
1dev |	data/licenses/Curl			| common
1dev |	develop/documentation/curl/*.html	| common_devel
1dev |	develop/headers/curl/*.h
1dev |	develop/lib/libcurl.*
1dev |	develop/lib/pkgconfig/libcurl.pc
1dev |	documentation/man/man3/*.3		| common_devel
||
2dev |	bin/curl-config-x86
2dev |	bin/x86/curl-config
2dev |	data/licenses/Curl			| common
2dev |	develop/documentation/curl_x86/*.html	| common_devel
2dev |	develop/headers/x86/curl/*.h
2dev |	develop/lib/x86/libcurl.*
2dev |	develop/lib/x86/pkgconfig/libcurl.pc
2dev |	documentation/man/man3/*.3		| common_devel

With this layout we save:
* about 286 kiB which would otherwise be duplicated in curl_devel and curl_x86_devel;
* about 84 kiB which would otherwise be duplicated in curl and curl_x86.

Moreover, this layout allows to install only what you need. If you don't want to waste 270kiB with curl_devel  but still be able to install curl_x86_devel, you can do that.

Here are the sizes for curl-7.47.1 with the new layout:
```
-rw-r--r-- 1 305159 Mar 23 19:02 curl-7.47.1-1-x86_gcc2.hpkg
-rw-r--r-- 1  83885 Mar 23 19:02 curl_common-7.47.1-1-any.hpkg
-rw-r--r-- 1 290785 Mar 23 19:02 curl_common_devel-7.47.1-1-any.hpkg
-rw-r--r-- 1 274328 Mar 23 19:02 curl_devel-7.47.1-1-x86_gcc2.hpkg
-rw-r--r-- 1 204318 Mar 23 19:30 curl_x86-7.47.1-1-x86_gcc2.hpkg
-rw-r--r-- 1 308317 Mar 23 19:30 curl_x86_devel-7.47.1-1-x86_gcc2.hpkg
```

If we keep the old layout, then we should at least remove the duplicate files. But there is a consequence: curl_x86_devel will need to depend on curl_devel, otherwise you need to ship all html files in each devel package.

If you prefer the old layout, then I can switch back, but I think it would be much better with the new layout because we have real architecture-independent packages, we avoid package overlaps, and we save space.

korli:
Keep in mind one might install different versions for primary and secondary arch. There is no common package to think about. The docs should be with the devel package anyway.

fbrosson:
Oops, you are perfectly right. I'll switch to the old layout.
But wait, we have compat options that should allow to keep the versions of primary and secondary architectures in sync, haven't we?
Moreover, if we allow different versions for curl and curl_x86 then we won't even know which package the **`data/aclocal/libcurl.m4`** file comes from.  This could cause headaches for users with different versions. The same is true for the man pages: the man pages of one ot the versions will be shadowed by the man pages of the other version. So is it really safe to allow a version mismatch between primary and secondary architectures? Is it really worth allowing this, compared to the risk of using the wrong libcurl.m4 ?
Well, I know we could workaround the libcurl.m4 problem by installing it somewhere else for the secondary arch, but that would cause quite some changes.
To conclude, I don't want to force any decision. Just let me know what your choice is and I'll fix the recipe accordingly. Thanks!

fbrosson:
@korli In case my previous edit did not produce a GitHub notification, I'm adding this message to confirm that the recipe is ready and it uses the traditional split layout (curl and curl_devel) as you asked.
BTW, the alternate split (with common and common_devel) was causing problems in some situations (like building 4.48.0 over a 4.47.1 pre-installed with the new split layout), probably because of my wrong using of **`$portVersion base`** in common and common_devel or somewhere else. So I have absolutely no regrets in switching back to the traditional layout ;-)

korli:
Please leave the license file in curl${secondaryArchSuffix}_devel. No reason to single out this particular package.

korli:
Regarding --with-ca-fallback, this seems related to verify path. Not sure why you need to activate it though.
https://github.com/curl/curl/pull/569

fbrosson:
I've reverted the removal of the license file in curl_devel, as you asked.
Regarding the «**`--with-ca-fallback`**» switch, which just allows to have a fallback pool of root certificates in case curl needs to check a server's certitiface that is signed by a root certificate that is not in our ca_root_certificates package, I guess the only risk is that if one of the fallback root certificates gets compromised then I'm not sure wjether or not curl has an internal certificate revocation feature for such compromised root certificates. But if I'm not mistaken we already have this same risk with ca_root_certificates: If one of its root certificates gets compromised and we don't update that package then we'll have the same risks.
So I don't know if we should keep the «**`--with-ca-fallback`**» switch or not. Let me know if you prefer to drop it and I'll update the recipe.

fbrosson:
I read the discussion again and what I understand is that before 7.48.0 the swich did not exist but the feature it triggers was automaticaly enabled. The switch was added to allow admins to not include the CA fallback in case they know their system has up-to-date root certificates. So if we're not sure our ca_root_certificates package is up-to-date then we should enable the «**`--with-ca-fallback`**» switch otherwise we might hit some problems that we did not have in previous releases of curl.
If you agree with this then the recipe is OK.

fbrosson:
I've just removed the «**`--with-ca-fallback`**» configure option.
Afer all, we can always add it back if we find out that it is absolutely required.

korli:
Seems fair to me. I noticed Gentoo explicitly used --without-ca-fallback. 
Thanks.

PR #519, opened 2016-03-25
smtube: bump to version 16.3.0.

fbrosson:


fbrosson:
smtube comes with 2 files that we could probably process and install: **`smtube.rc`** and **`smtube.ico`**.
I'll try to find out how to do that.

pulkomandy:
rc and ico files are specific to Windows. If you want the executable to have an icon in Haiku, it should be in HVIF format and added to the executable (HaikuPorter has a rule to do that easily from the recipe).

fbrosson:
Thx. guess it's **`addResourcesToBinaries`**. I'm examining a few recipes that call it to see if I can embed file **`smtube-16.3.0/src/smtube.ico`**
```
smtube-16.3.0/src/smtube.ico: MS Windows icon resource - 5 icons, 256x256
```

pulkomandy:
Embedding a windows ICO file will not work: Haiku Tracker does not know how to display it. You need an HVIF file (Haiku Vector Icon Format) which can be designed in Icon-O-Matic.

fbrosson:
OK, thanks. I'll try Icon-O-Matic.
BTW, I was looking for a command to convert the icon to a **`resource vector_icon { ...  }`** text block in order to create a rdef file in INSTALL() and pass it to addResourcesToBinaries.

pulkomandy:
Our icons are in a vector format so you can't convert ICO (which is a bitmap) automatically.

fbrosson:
OK. Well, since I don't know Icon-O-Matic enough to create the HVIF I can't add an rdef file for smtube yet. But I guess this is OK.

Regarding the supported architectures, I've marked x86_gcc2 as not supported because smtube needs libqt4 which can only be built on x86_gcc2 using the secondary arch, x86.
Is it OK to mark a recipe as not supported on an arch just because it depends on another port which is not supported on that arch?

pulkomandy:
Yes, this is fine. We can always revisit it if someone ever needs to get it working.

fbrosson:
I found an online tool that converts images to SVG, gave it the ico file and got a black and white SVG which I exported to HVIF with Icon-O-Matic. I changed color to a blue close to the original icon, but the resulting icon is very far from the original. Would it be OK to have such an ugly icon? Or is it better to drop this custom icon for this PR?

pulkomandy:
It's better to not have any icon and wait for someone with the required drawing skills to offer one.

fbrosson:
OK. Icon removed (as well as the extra steps in the recipe).
(I think the recipe is ready now.)

mmuman:
Well, you should still be able to convert the .ico into a bitmap icon which we still do support, but the process is a bit involved.
It's also possible to use Inkscape to vectorize bitmaps but only some settings work correctly.

fbrosson:
Hmm, is there a mini how-to explaining how to embed a bitmap icon into an executables?
Well, that's not important anyway. We can wait until some graphist creates a nice HVIF for smtube.

pulkomandy:
You can use the command line tool "translate" to convert the icon to a BBitmap, and then add that as a bitmap icon resource. But a vector icon in the native style is preferred anyway.

mmuman:
Hmm flattened bitmaps aren't supported as icons, are they? We'd really need a howto :)

fbrosson:
OK, thanks for the explanation. I might try the process some day to see how it works, but I don't feel like adding a bitmap icon for smtube. (I agree with you that we can merge this PR and then wait until someone creates a HVIF for smtube.)

humdingerb:
Hey, fbrosson! Kudos for working so diligently on all those non-trivial recipes. I love seeing your latest PRs/merges almost every morning. :)

WRT icons or other things you may need, I suggest to post to the Haiku forums. Let's people with non-porting skills to get involved. IRC is also a nice way to get more immediate feedback.

fbrosson:
Thanks @humdingerb for your kind and warm message!
Regarding the (Haiku) forums, I'm too shy to use them. But I'll try to ;-)

fbrosson:
I've included in the commit of this PR a version bump for smplayer. Is this OK? Or should I move the version bump for smplayer to a separate PR or to separate commit in this PR?
BTW, is it good practive to rebase a commit against master if there are no other changes?
Thanks for your feedback.

fbrosson:
Thanks @korli for the merge!

PR #524, opened 2016-03-29
Git updated to 2.8.0

miqlas:
Just to be up to date.

### DIFF
--- a/dev-vcs/git/git-2.8.0.recipe
+++ b/dev-vcs/git/git-2.8.0.recipe
@@ -1,0 +1,223 @@
+SUMMARY="Fast, scalable, distributed revision control system"
+DESCRIPTION="Git is a free and open source distributed version control system \
+designed to handle everything from small to very large projects with speed \
+and efficiency.
+
+Git is easy to learn and has a tiny footprint with lightning fast performance. \
+It outclasses SCM tools like Subversion, CVS, Perforce, and ClearCase with \
+features like cheap local branching, convenient staging areas, and multiple \
+workflows."
+HOMEPAGE="http://git-scm.com/"
+LICENSE="GNU GPL v2"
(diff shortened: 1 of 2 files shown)


korli:
Please split your PR

miqlas:
What do you mean, Jerome? I don't understand you, sorry.

-- Z.

2016.03.29. dátummal, 17:17 időpontban Jérôme Duval <notifications@github.com> írta:

korli:
Well, the initial PR email had changes for fish too. Since they disappeared in between, you can ignore my comment.

miqlas:
Yep, i noticed and fixed that ;)
I'm not so good yet with git and make the same mistakes again and again, but trying hard to do my best!

-- Z.

2016.03.29. dátummal, 22:17 időpontban Jérôme Duval <notifications@github.com> írta:

TwoFX:
git 2.8.0 introduces a new option `core.untrackedCache` which speeds up tracking untracked files e.g. in `git status`. This option only works for certain file systems, but BeFS is one of them (tested with this recipe doing `git update-index --test-untracked-cache`). At least on my virtual machine, `git status` sometimes prints a message like "Enumerating untracked files took x seconds", where x > 15, so I suggest enabling this option by default.

miqlas:
Is it possible to enable core.untrackedCache in build-time?
As i see, it is a runtime configuration option. grepping the configure script and the whole source tree came out without any results about build-time option.
Light me up, if you got more info.

TwoFX:
I see two options: Either running `git config --global core.untrackedCache true` at install time (this didn't work right away when I tested it, it complained that $HOME was unset, but I think it can be done), so that haikuporter captures the created settings file, or shipping a config file with the fixed contents
```
[core]
	untrackedCache = true
```
with the package (this can also be done in INSTALL).

Both have the same result.

PR #525, opened 2016-03-29
Recipe for STFL

miqlas:


miqlas:
Somehow the fixPkgconfig can't find the required files:

> mv: cannot stat '/packages/stfl_x86-0.24-1/.self/lib/x86/pkgconfig': No such file or directory

pulkomandy:
Not all projects comme with a pkg-config file. If there isn't one, then there is nothing to fix.

sys-libs/stfl/stfl-0.24.recipe
```diff
@@ -0,0 +1,83 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	mkdir -p $libDir/pkgconfig
+	mkdir -p $includeDir
+
+	mv /usr/local/lib/libstfl.a $libDir
```
pulkomandy: This is a bit ugly. Is there no way to tell "make install" where to put the file? Usually you can do make install DESTDIR=$prefix or something similar.

sys-libs/stfl/stfl-0.24.recipe
```diff
@@ -0,0 +1,83 @@
...
+	mv /usr/local/lib/pkgconfig/stfl.pc $libDir/pkgconfig
+	mv /usr/local/include/stfl.h $includeDir
+
+	prepareInstalledDevelLibs libstfl
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+		$libDir/pkgconfig
```
pulkomandy: fixPkgConfig will move the pkg-config files to $developLibDir, and there should not be any left in $libDir. So, you don't need this.

sys-libs/stfl/stfl-0.24.recipe
miqlas: Sadly it is hardcoded in Makefile.cfg:

`export libdir ?= lib
export prefix ?= /usr/local
export DESTDIR ?= /`

I can actually "sed" to fix it, but would it better portable so? Idk.

miqlas: Ofc, the stfl.pc contains these paths too:

> prefix=/usr/local
> exec_prefix=${prefix}
> libdir=${exec_prefix}/lib
> includedir=${prefix}/include

So i need to do something with the Makefile.cfg.

pulkomandy: On Wed, Mar 30, 2016 at 10:34:15AM -0700, miqlas wrote:
> `export libdir ?= lib
> export prefix ?= /usr/local
> export DESTDIR ?= /`
> 
> I can actually "sed" to fix it, but would it better portable so? Idk.

This is not hardcoded, the ?= means "set if not alredy set, so you can
override them from the command line:

    make prefix=$prefix libdir=$relativelibdir

miqlas: Oh, cool! Lemme' fix this thing!

miqlas:
Hopefully better now.

sys-libs/stfl/stfl-0.24.recipe
```diff
@@ -0,0 +1,80 @@
...
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	mkdir -p $includeDir
+	mv $prefix/include/* $includeDir
```
pulkomandy: It would be better to patch this directly in the makefile so it can be upstreamed

sys-libs/stfl/stfl-0.24.recipe
```diff
@@ -0,0 +1,80 @@
...
+REQUIRES_devel="
+	stfl$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libncurses$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	cmd:python
+	cmd:perl
```
pulkomandy: the cmd: provides should be on the PREREQUIRES list.

sys-libs/stfl/stfl-0.24.recipe
```diff
@@ -0,0 +1,80 @@
...
+SOURCE_DIR="stfl-$portVersion"
+PATCHES="stfl-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	stfl$secondaryArchSuffix = $portVersion compat >= 0.24
+	lib:libstfl$secondaryArchSuffix = $portVersion compat >= 0.24
+	lib:libstfl.so.$portVersion$secondaryArchSuffix = $portVersion compat >= 0.24
```
pulkomandy: this looks wrong. The .so shouldn't be part of the provide entry (so the one above is correct).
also make sure the version and compat are correct (is there no libfstl.so.0 or other files?)

(same for the devel entry)

PR #529, opened 2016-03-31
smplayer: bump to version 16.4.0.

fbrosson:


media-video/smplayer/smplayer-16.4.0.recipe
```diff
@@ -0,0 +1,74 @@
...
+
+BUILD()
+{
+	# TODO: fix paths
+	#		DATA_PATH
+	#		DOC_PATH
+	#		TRANSLATION_PATH
+	#		THEMES_PATH
+	#		SHORTCUTS_PATH
+	echo "SVN-r6966" > svn_revision
```
miqlas: Is it really required? Why is it the same revision as the old one?

fbrosson: Good catch, thanks. It builds and works fine even if we skip the creation of file svn_revision.
I have therefore removed that line and updated the commit.
I did not dare to remove, however, the commented lines of the "TODO: fix paths" block.
(Should I remove them too?)

fbrosson:
Thanks @korli for the merge!

PR #530, opened 2016-04-01
qbittorrent: bump to version 3.3.4.

fbrosson:
* Upgrade 3.3.1 to 3.3.4, switching back to Qt4. The reason is that although the application builds fine against Qt5 and even seems to work, it cannot save its configuration. The Qt5-based version of qbittorrent is therefore not usable. A Qt5-based variant is added with all architectures marked as untested in case someone would like to troubleshoot the issue.
* Drop 3.2.3 (which used Qt4) as we now have 3.3.4 (also Qt4-based).
* Replace hard coded info in qbittorrent.rdef by @FIELD@-style vars to make it generic with regards to the $portVersion.
* Added missing lib:libexecinfo and lib:libssl in REQUIRES to match their devel: counterparts in BUILD_REQUIRES.
* Added dependency on libGL, as suggested by diversys. It seems to me that it is not required but I'm probably missing something.
    
**Warning:** If building against Qt4 fails then uninstall Qt5 and retry.


net-p2p/qbittorrent/qbittorrent-3.3.4.recipe
```diff
@@ -6,20 +6,28 @@ OS/2, FreeBSD).
 
 qBittorrent is based on Qt4 toolkit and libtorrent-rasterbar."
```
diversys: Qt4->Qt since we use devel:libqt5

net-p2p/qbittorrent/qbittorrent-3.3.4.recipe
```diff
@@ -68,8 +76,10 @@ INSTALL()
 	rm -r $prefix/{bin,data,documentation}
 	strip $appsDir/qBittorrent/qBittorrent
 
-	addResourcesToBinaries $portDir/additional-files/qbittorrent.rdef \
+	addResourcesToBinaries $portDir/additional-files/qbittorrent-$portVersion.rdef \
 		$appsDir/qBittorrent/qBittorrent
 
 	addAppDeskbarSymlink $appsDir/qBittorrent/qBittorrent
+	mkdir -p $commandBinDir
+	symlinkRelative  -s  $appsDir/qBittorrent/qBittorrent $commandBinDir/qbittorrent$commandSuffix
```
diversys: Extra white space after symlinkRelative :)

diversys:
POLICY WARNING: "bin/qbittorrent" needs library "libGL.so", but the package doesn't seem to declare that as a requirement

net-p2p/qbittorrent/qbittorrent-3.3.4.recipe
fbrosson: Good catch, thanks!

net-p2p/qbittorrent/qbittorrent-3.3.4.recipe
fbrosson: fixed ;-)

fbrosson:
I've added **`lib:libGL$secondaryArchSuffix`** to PROVIDES and **`devel:libGL$secondaryArchSuffix`** to BUILD_REQUIRES but I haven't tested since. On x86_gcc2 it works fine without it. Did it complain and/or refuse to build/run on your system?
BTW, since pre-built packages of Qt5 for x86{,_64} are not yet available I made the recipe depend on libqt4 (instead of qt5) for x86{,_64} because I don't have enough space to build them. I don't know if this is acceptable... and I still need to test this.

diversys:
I don't think it's a goo idea to depend on libqt4 for x86_64. Builders should build Qt5 dependency automatically.

fbrosson:
Good news: I noticed that 3.2.3 was Qt4 based, so I:
* reverted the Qt4/Qt5 "mix" in 3.3.4 (which, I agree, was not a very good idea) and made **`qbittorrent-3.3.4.recipe`** 100% Qt5 for all architectures as it was before.
* added a Qt4 variant of 3.3.4 in **`qbittorrent_qt4-3.3.4.recipe`**;
* dropped 3.2.3 (which was Qt4 only).

I think the commit is now both clean and ready ;-)
Thanks in advance for reviewing!

korli:
I would drop qbittorrent_qt4.

fbrosson:
Pre-built packages of Qt5 are not yet available for x86{,_64} and since I removed 3.2.3 (which was based on Qt4), I think it would be nice to have the Qt4 variant.
But if you prefer to have only the Qt5-based version then that's OK for me. Just confirm and I'll do the chenage.
BTW, I marked the Qt4 variant as tested on x86 and x86_64 since I have built and tested it on both of these architectures. If anyone has built 3.3.4 with Qt5 on x86{,_64} then I can mark these tested.
I also tested on x86_gcc2 with both versions of Qt and both work fine.

net-p2p/qbittorrent/qbittorrent-3.3.4.recipe
```diff
@@ -52,7 +57,7 @@ BUILD_PREREQUIRES="
 BUILD()
 {
 	bootstrap.sh
-	BOOST_CPPFLAGS=-I/system/$relativeHearersDir runConfigure \
+	BOOST_CPPFLAGS=-I/system/$relativeHeadersDir runConfigure \
```
diversys: I wonder how did it work before? Maybe passing BOOST_CPPFLAGS is not needed anymore?

fbrosson: That's a godd question. Well, I did not dare to remove it because I can't test the build against Qt5 (since I don(t have pre-build packages of Qt5 and building them requires a lot of free space and takes a very long time). But if you have tried without the BOOST_CPPFLAGS environment variable and found out that it still builds and runs fine then, of course, I can remove it in my commit ;-)

korli:
Well, it would be OK to have a Qt4 recipe if that actually keeps something working. ATM the 3.2.3 recipe has x86 and x86_64 marked as untested. So not supporting Qt4 on these architectures isn't probably a big problem.
The right thing to do here is to request for Qt5 packages on x86 and x86_64, which will enable you to test the 3.3.4 recipe on these architectures. Hope you understand my point.

fbrosson:
Cool, thanks for your OK about the Qt4 variant.
Regarding the request for Qt5 packages on x86 and x86_64, I guess someone would need to build them them and then upload the hpkg files to the depot server with HaikuDepot.
But since I did not notice any difference between qbittorrent on x86_gcc2 and qbittorrent_qt4 on x86 or x86_64, I think nobody would bother building Qt5 just for qbittorrent.
My rationale is that with the Qt4 variant of qbittorrent we're sure that anyone interested in running qbittorrent on Haiku will be be able to build it in less than 10 minutes. Conversely, if we did not have the Qt4 variant then most users would either switch to Haiku x86_gcc (if they have it) or to another OS. IMHO flexibility (with Q4 variants) makes Haiku more attractive.
BTW, I still have a few changes to push in order to fix a couple of issues.
Cheers,

korli:
It's not what I meant: When nobody worried about qbittorrent on x86 and x86_64 until now, then why the headache? Having Qt5 packages for all architectures is what's needed here, not multiple variants to maintain. This would also set a precedent for other Qt recipes, which we don't want.
I repeat myself: The right thing to do here is to request for Qt5 packages on x86 and x86_64, which will enable you to test the 3.3.4 recipe on these architectures.

fbrosson:
@korli, sorry for not havng replied earlier. While I was trying to fix the issue about the user settings not being saved I noticed that it was only affecting the Qt5-based build. So I swapped the main recipe with the variant in order to have the Q4-based recipe in qbittorrent and the Qt5-variant in qbittorrent_qt5.
There is probably a way to fix the issue on the Qt5 variant, but I haven't found it yet.
If you think it would be OK to merge what is in this PR then let me know and I'll do the merge (if you agree) since I'm now member of HaikuPorts!
Of course, we can also leave the PR open until someone finds out how to fix the issue.
Thanks!

fbrosson:
Cool, thanks!

PR #549, opened 2016-04-04
geoip: create recipe.

fbrosson:


fbrosson:
@diversys: Thanks again for the merge!

PR #550, opened 2016-04-05
geoipupdate: create recipe.

fbrosson:
Created recipe for version 2.2.2 with nice polishing for the man pages, the runtime and the default configuration:
* Template config in **`system/data/geoipupdate/GeoIP.conf.default`**.
* Active config in **`system/settings/GeoIP.conf`**.
* Download location for all GeoIP/GeoIP2 databases: **`/var/lib/geoip/`**.
* Download location for all GeoIP Legacy and GeoIP2 databases: B_SYSTEM_CACHE_DIRECTORY/GeoIP/ (i.e. /boot/system/cache/GeoIP/).
* Patch for the default config to allow **`geoipupdate`** to download the GeoLite2-City and GeoLite2-Country databases out-of-the-box.

Updated the recipe for geoip-1.6.9 to make it use the same directory used by geoipupdate instead of **`/var/lib/geoip/`** previously used when that recipe was created. Also fixed missing mkdir in geoipupdate.sh and added lost patch with miscellaneous enhancements for it and some cleaning in the recipe.


diversys:
It's probably better to use B_SYSTEM_CACHE_DIRECTORY **`system/cache`** instead of **`/var/lib/geoip/`**.

fbrosson:
Thanks @diversys for your suggestion. I made the changes for it in both the geoipupdate and geoip packages.
It was a bit tricky because of the default config for geoupupdate: It needs to be installed in system/data/ but the configuration scripts of geoipupdate were using $dataDir to define the location of the database files, so I had to switch to **`$sharedStateDir`** and pass **`"`finddir B_SYSTEM_CACHE_DIRECTORY`"`** as **`--sharedstatedir`** instead of letting runConfigure pass the default $dataDir. Well, that was tricky but both packages are now working fine with the new directory.
BTW,  for some reason **`dev-libs/geoip/patches/geoipupdate-r5.patchset`** did not get merged in #549 (probably because it was not in my first push but only in the second or third push, only 10 minutes before the merge). Anyway, since I'm ajusting geoip for the directory change, I've added in this commit what was missing in the previous plus some minor polishing or cleaning.
Sorry for the hiccups and thanks in advance for reviewing and merging!

PR #553, opened 2016-04-08
irssi: add recipe for version 0.8.19

pisculichi:


korli:
Hi,
should the PR for irssi 0.8.17 be dismissed?

pisculichi:
Yes, this is for the last versión.

net-irc/irssi/irssi-0.8.19.recipe
```diff
@@ -0,0 +1,66 @@
...
+HOMEPAGE="http://www.irssi.org/"
+SOURCE_URI="https://github.com/irssi-import/irssi/archive/0.8.19.tar.gz"
+CHECKSUM_SHA256="07630dcc4b8ba1a0bc7d56810ea6a4e0521de9768ea05bdb2f3d6b772903c3f6"
+REVISION="1"
+ARCHITECTURES="x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
+BUILD_PACKAGED_ACTIVATION_PHASE="INSTALL"
+STATUS_HAIKU="stable"
+PROVIDES="
+	botti$secondaryArchSuffix = $portVersion
```
korli: why providing botti? this line should be dropped.

ailin-nemui:
being totally ignorant about haiku, I would like you to double check if those "Move variable declarations" are all required. Because ANSI C does not demand variables only at function top at all

fbrosson:
@pisculichi, I can confirm what @ailin-nemui wrote because the gcc2 we have in Haiku for x86_gcc2 does not require that variables be declared in the begining of a function. It only requires that all declarations in a «**`{`** … **`}`**» bloc appear before the very first statement of that bloc. So you don't need to move declarations that appear inside «**`{`** … **`}`**» blocs if these declarations are already grouped in the begining of its deepest bloc. This means that if you see any declaration mixed with code then you only need to move them up near the closest «**`{`**» that contains it.
What happens very often is that you have to split the declaration and its initialisation if that initialisation is done with a function call.
```C
char *foo = strdup("foo");
char *bar = strdup("bar");
```
would have to be changed to:
```C
char *foo;
char *bar;
foo = strdup("foo");
bar = strdup("bar");
```

or, even better:
```C
char *foo, *bar;
foo = strdup("foo");
bar = strdup("bar");
```

BTW, thanks @ailin-nemui for spotting this.

ailin-nemui:
you're welcome to submit genuine ANSI C fixes upstream

fbrosson:
@ailin-nemui: That's cool, thanks!
@pisculichi: I'll let you take care of it, of course.

pisculichi:
@ailin-nemui @fbrosson you are right I moved all declarations when I saw errors at compiling. I update this commit with your recommendations.

net-irc/irssi/irssi-0.8.19.recipe
pisculichi: Droped!

mmuman:
@ailin-nemui @fbrosson as I already said, those are C89 fixes, because that's all gcc2 supports. But it's not strictly required for irssi since it doesn't provide any library so we don't really have any dependency on it, so we can build with gcc4 (5 now).

fbrosson:
@mmuman: I though it was good practice to try and make _all_ recipes build with gcc2 when that was either easy or essential. But now I just understood your remark regarding those packages that do not provide any library.
There is, however, a small reason to also convert those packages to x86_gcc2: fewer dependencies. And, if I'm not mistaken, such packages may also be used on BeOS if converted to gcc2.
Hence this question: Now that a gcc2 patch is ready for this package, would it be OK, if it works perfectly, to accept it?
@pisculichi: I read your gcc2 patches and everything seems OK to me. I was planning to build it but did not try it yet. I will do so and I hope your gcc2 patches will be accepted for a merge. I hope you won't be angry against me if the gcc2 patches do not get merged, because I was not aware of the rule when I wrote my first message in this PR...

pisculichi:
@fbrosson no problem. I did it to learn haikuports, git, and contribute in someway. I understand that support ANSI C in 2016 maybe has not sense.
Anyway, thanks all for the comments =)

fbrosson:
Thanks @pisculichi for your kind answer.
Hmm, the patch @pisculichi sent to upstream has just been accepted and merged by @ailin-nemui. That's cool. Thanks @ailin-nemui!

fbrosson:
The following bloc allows to reference a tarball of the snapshot just after the commit that adds the gcc2 patches:
```Shell
srcGitRev="1cfec5f63d4adf32117f68638fb35802146e2784"
SOURCE_URI="https://github.com/irssi/irssi/archive/$srcGitRev.tar.gz"
CHECKSUM_SHA256="e4118fedeee70133b4aab70406c30be107cc0078c4d5c509e2b068277284d88a"
SOURCE_FILENAME="irssi-$srcGitRev.tar.gz"
SOURCE_DIR="irssi-$srcGitRev"
```
I don't know, however, if it's OK to do so without renaming the recipe to, say, irssi-0.8.19~git.recipe (or maybe irssi-0.8.20~git.recipe or something similar) to make it clear that it's not exactly 0.8.19 but a snapshot between 0.8.19 and 0.8.20.
(BTW, I'm not sure I used the recomended order for CHECKSUM_SHA256, SOURCE_FILENAME and SOURCE_DIR. So please correct me if that's not the most used order.)

mmuman:
@fbrosson it's ok to keep the release + gcc2 patch until the next one is out. If you want a git recipe you'd rather use 0.8.19_git as naming I think.

fbrosson:
OK, thanks. BTW, I noticed that there are indeed several recipes with "_git", and only 2 with "~git", but these two are quite recent, so I though there was a little change in the naming conventions.
Anyway, I'll let @pisculichi drive this PR cause I don't plan to add a recipe with a snapshot.

waddlesplash:
Bump.

pisculichi:
I forgot about this, sorry.
Should I improve the changes that @fbrosson told me (modify and rename irssi-0.8.19_git)?

waddlesplash:
Yes, please.

net-irc/irssi/irssi-0.8.19_git.recipe
```diff
@@ -0,0 +1,68 @@
+LICENSE="GNU GPL v2"
+COPYRIGHT="1999-2015 Timo Sirainen"
```
waddlesplash: Please follow the order outlined in https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

net-irc/irssi/irssi-0.8.19_git.recipe
```diff
@@ -0,0 +1,68 @@
...
+srcGitRev="1cfec5f63d4adf32117f68638fb35802146e2784"
+SOURCE_URI="https://github.com/irssi/irssi/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="e4118fedeee70133b4aab70406c30be107cc0078c4d5c509e2b068277284d88a"
+SOURCE_FILENAME="irssi-$srcGitRev.tar.gz"
+SOURCE_DIR="irssi-$srcGitRev"
+REVISION="1"
+ARCHITECTURES="x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
+BUILD_PACKAGED_ACTIVATION_PHASE="INSTALL"
+STATUS_HAIKU="stable"
```
waddlesplash: This and BUILD_PACKAGED_ACTIVATION_PHASE are no longer used.

net-irc/irssi/irssi-0.8.19_git.recipe
```diff
@@ -0,0 +1,68 @@
...
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+PATCHES="
+	irssi-ncurses.patch
+	"
+
+BUILD() {
```
waddlesplash: Braces go on next line.

waddlesplash:
Manually merged. Sorry that took so long!

PR #554, opened 2016-04-08
gptfdisk: bump to version 1.0.1.

fbrosson:
* Bump version.
* Add **`{lib,devel}:libpopt`** to **`{,BUILD_}REQUIRES`**.
* Set **`ARCHITECTURES`** to **`x86_gcc2 x86 x86_64`** instead of **`!x86_gcc2`**.
* Add **`cmd:{cgdisk,fixparts,gdisk,sgdisk}`** to **`PROVIDES`**.
* Do not **`make install`** in **`INSTALL()`** because there is no such target in the makefile and, instead, call **`cmd:install`**.
* Fix patch to avoid including a missing **`sys/disk.h`** in **`support.h`**.
* Add patch to make **`diskio-unix.cc`** accept character devices on Haiku.
* Add patch to use the correct CRC32 polynom on big-endian machines in case someone wants to build and try Haiku on ppc.


fbrosson:
Thanks @diversys for the merge!
BTW, I just noticed I had an invitation to join the HaikuPorts organization... Shame on me! I should have read my e-mail more often.
This means I won't need to bother anyone from now on with all my pull requests :-)
Being a member of HaikuPorts is so cool!
Thanks to all members who have decided to invite me (and sorry for not having joined earlier.)

PR #557, opened 2016-04-15
Update DriveEncryption homepage to upstream repo

richienyhus:
Change homepage back to github repo of Axel Dörfler

### DIFF
--- a/haiku-apps/drive_encryption/drive_encryption-1.0.recipe
+++ b/haiku-apps/drive_encryption/drive_encryption-1.0.recipe
@@ -5,7 +5,7 @@ and protected via a password.
 It can also use devices created with TrueCrypt (http://www.truecrypt.org/), \
 available for Windows and Linux, as long as you've created them to use the \
 AES encryption algorithm and a RIPEMD-160 hash algorithm protected password."
-HOMEPAGE="https://github.com/chirayudesai/driveencryption"
+HOMEPAGE="https://github.com/axeld/driveencryption"
 COPYRIGHT="2007-2014 pinc Software"
 LICENSE="TrueCrypt
 	MIT"


fbrosson:
Thanks @richienyhus!

@haikuports/owners & @haikuports/developers: I'm **_so_** sorry for the double commit. I tried doing the merge from command line, but then I noticed that the status of the PR had not been chenged to Merged. So I switched to the web-based method, thinking GitHub would notice the commit had already been merged.
If I had tried the web-based merge from the beginning then it would have been OK, with no duplicate commit log. I feel so bad with this mistake!
I promisse I won't merge other PRs anymore, unless I'm forgiven for this mistake and I get green lights for future merges. But I guess I won't even think about merging PRs anymore because there are several other contributors who have far more experience with Git than myself. Thanks!

waddlesplash:
Using GitHub to merge is OK by me. Small changes like this are just fine for anyone to merge, but since you're still new I'd prefer if you didn't merge anything substantial without someone else signing off on it first at least.

fbrosson:
I knew it would have been OK because it was just to update a metadata that can be easily checked by anyone (I visited both the old and the new homepage and understood that the change was OK).
But I should have refrained from merging the PR myself. I can assure everyone that I won't merge any other PRs. (I would be too affraid of making another mistake...)

Regarding a similar topic, what is the best way to handle commits I would like to make but for which I'm not 100% sure of their correctness?
Until now all the "direct commits" I made (since I have commit access) were commits that I was 100% sure of.
But if I wish to get feed-back for a commit before merging it, is it better to push it into a branch of my clone repo and create a PR (as I've been doing until recently) or only push it to a branch of my clone repo and then chat on IRC?
BTW, I guess I'm being off topic here and should already have switched to IRC... shoudn't I? Thanks!

PR #559, opened 2016-04-15
ginac & cln: fix cln 1.3.4, update ginac 1.6.5, add ginac 1.6.7.

fbrosson:
sci-libs/cln:

* REQUIRES_devel was set to **`libcln$secondaryArchSuffix`** instead of **`cln$secondaryArchSuffix`**, therefore preventing both the installation of cln_devel and its use in chrooted builds of ginac.
* Drop the recipe for version 1.3.2 (which was really old.)
* Change cmd:gcc to cmd:g++ in BUILD_PREREQUIRES.
* On x86_gcc2 using x86 as secondary arch, install pi in bin/ instead of bin/x86/.
* Do not use $portVersion as version of lib:libcln. Instead, use the libs' real version. This change is safe since the only package wich needs libcln is ginac and it is being updated in this very commit.

sci-mathematics/ginac:

* Fix versioned dependency on libcln for ginac-1.6.5.
* On x86_gcc2 using x86 2nd arch, do not install runtimes in bin/x86/ but in bin/.
* Add ginac-1.6.7.


fbrosson:
The patch file is a patchset (and it is correct), and yet, its name is not **`cln-1.3.4.patchset`** but **`cln-1.3.4.patch`**. Hence this question:
When a recipe is changed but the contents of its associated patch do not need to be changed , should such a misnamed patchset file be renamed to "fix" its extension? Or is it OK to keep its name until a new version becomes available?
If everything is already OK then, if you agree, I can do the merge myself. Of course, if you prefer to merge it yourself then that's also fine for me ;-)
BTW, thanks @pulkomandy and @waddlesplash for your replies to the questions I had attached to my previous commit (https://github.com/haikuports/haikuports/commit/2456fe1a80c500b623907151e240d245983574ef) before I created this PR.

fbrosson:
Hi all!
I've added some changes to ginac which are required since I'm fixing the version of libcln.
Please review. Thanks!

fbrosson:
Both ginac 1.6.5 and 1.6.7 build and run OK with lib:libcln from **`sci-libs/cln/cln-1.3.4.recipe`**. I've checked that on all possible architectures: x86, x86_64, and, finaly x86_gcc2 using x86 secondary arch. 
I even played a little with ginsh to try and see if there was no regression between 1.6.5 and 1.6.7. I'm leaving the recipes for both versions, however, in case someone finds a bug in 1.6.7 and wishes to check against 1.6.5.

Regarding the name of the patch file, I do understand that patchsets should have the ".patchset" extension. I still don't know, however, if it is good practice to rename a patchset file if the only problem with it is its name. That's the case here for sci-libs/cln. So if someone could confirm that it should be renamed (from **`cln-1.3.4.patch`** to **`cln-1.3.4.patchset`**) then I'll do so. @pulkomandy, I read your previous answer but I could not come to a conclusion about our case here, sorry, so I just assumed, for the moment, that the right answer was the one which reduces the history log. Please let me know what's the right thing to do here. Thanks!

The only package which depends on cln is ginac and there are no packages depending on ginac. So it is 100% safe to merge. I've already updated the title of the commit message with a reference to this PR.
So if there is no objection I can merge, using the web-based interface ;-)
Thanks in advance for any reply!

pulkomandy:
I'll try to be clearer then.
A commit which only change is renaming a single file from patch to patchset adds more noise than value to the commit history. So, I think the best thing to do is do it whenever you make another change to the same recipe. This way, you don't need a specific commit for that anymore and there isn't as much "noise".
An alternative would be to write a script to detect and rename all misnamed patchset files automatically and make one big commit fixing all of them.

fbrosson:
Thanks for your reply.
This commit is required because there is an error in the REQUIRES_devel of cln whch makes it impossible to build ginac. It also fixes the incorrect versions of {lib,devel}:libcln in both cln and ginac.

In other words, this commit:
* fixes the wrong REQUIRES_devel in cln;
* fixes the wrong version of libcln in the PROVIDES{,_devel} of cln;
* fises the wrong version of libcln in the {,BUILD_}REQUIRES of ginac 1.6.5.

Moreover, it fixes a minor issue of cln on x86_gcc2: pi was being installed in bin/x86/ instead of bin/.
And finaly, this commit adds ginac 1.6.7.
So, as you can see, this commit won't be just noise ;-)

Now, regarding the question of whether or not the patch file should be renamed, although I did understand that a patch file should not be renamed when _no_ changes are required in the recipe, I still can't tell what the right answer is when other changes in the recipe _are_ required.
I'm therefore assuming it is not worth renaming it, even if other changes are required in the recipe.
If this is the correct answer, then the commit is ready for merging. (If not, I can update the commit to include the renaming of the patch file.)

(May I do the merge with the web-based interface?)
Thanks!

PR #567, opened 2016-04-18
ccache: bump to version 3.2.5 using Haiku-style directories.

fbrosson:
* Add **`PATCH()`** to use Haiku-style directories instead of the defaults:
 * **`/usr/local/etc/ccache.conf`** → **`/boot/system/settings/ccache.conf`**
 * **`$HOME/.ccache/ccache.conf`** → **`/home/config/cache/ccache/ccache.conf`**
 * **`$HOME/.ccache/`** → **`/home/config/cache/ccache/`**
 * **`/usr/local/bin/`** → **`/boot/home/config/non-packaged/bin/`**
* Create **`[/boot/system/]settings/ccache.conf`** as the system-wide configuration file with the **`keep-old`** label.
* Pass **`finddir B_SYSTEM_SETTINGS_DIRECTORY`** as sysconfdir in order to have the short path displayed instead of the long one, in case the system-wide configuration file contains errors.


fbrosson:
Should I remove the recipe for version 3.2.2? 3.2.5 builds and seems to work fine on all supported architectures. Well, I "tested" it on _hello world_ with both the symlink method and the "prepend" method and both work ;-)

PR #568, opened 2016-04-18
cmake: bump to version 3.5.2.

fbrosson:


fbrosson:
Sucessfully built on x86_gcc2 using x86 as secondary architecture. The package installs fine and I think 
it is OK. I did not test it yet, however, nor did I try to build it on x86 and x86_64. I just created the PR in case someone would like to test it more carefully.
Of course I won't merge it because I don't know if everything is OK on all architectures.

korli:
If cmake.org now works OK, you can delete the fossies SOURCE_URI. 
Otherwise +1

fbrosson:
Done.
BTW, I've also added **`!x86_gcc2`** to SECONDARY_ARCHITECTURES.
Is this a good idea to do so? I mean, for recipes which define SECONDARY_ARCHITECTURES and do not build on x86 primary arch using x86_gcc2 as secondary arch.
Or is it better to not add the **`!x86_gcc2`** to SECONDARY_ARCHITECTURES?
Thanks!

korli:
For recipes which don't provide a library, x86_gcc2 as secondary arch isn't recommended, I think there is a ticket about this.

fbrosson:
OK, fixed ;-)
(I tthough that specifying **`!x86_gcc2`** in SECONDARY_ARCHITECTURES could be a good thing because it carried some information... but after all, if we have **`!x86_gcc2`** in ARCHITECTURES then it is obvious that it is also not supported as secondary arch.)

PR #569, opened 2016-04-20
boost: bump version to 1.60.0

diversys:


korli:
Is it binary compatible with 1.55?

diversys:
I don't really know. There's something about fixed ABI regression introduced in Boost 1.55 at http://www.boost.org/doc/libs/1_60_0/doc/html/intrusive/release_notes.html

fbrosson:
@diversys: You might wish to use **`PATCHES="boost-$portVersion.patchset"`** instead of **`PATCHES="boost-1.60.0.patchset"`**.

And you might try using **`boost_${portVersion//./_}`** instead of **`boost_1_60_0`**. It works on bash 4.3.42. I'm not currently on Haiku so I can't test it right now, but I think it would also work.

pulkomandy:
According to the release note changes, there was an ABI change in 1.56 but "mostly" for MSVC (whatever that means), and also a source compatibility change (removal of deprecated functions). I would say this is compatible enough for us to go ahead without recipe renaming here.

It could possibly solve some build problems with some recipes (I think Wesnoth is one which was failing somewhere in Boost headers, and a newer version of Boost could help).

dev-libs/boost/patches/boost-1.60.0.patchset
```diff
@@ -0,0 +1,136 @@
...
++
+ # ifdef OS2
+ 
+ ptr_t GC_get_main_stack_base(void)
+@@ -1106,7 +1116,7 @@ ptr_t GC_get_main_stack_base(void)
+ 
+ #endif /* FREEBSD_STACKBOTTOM */
+ 
+-#if !defined(BEOS) && !defined(AMIGA) && !defined(MSWIN32) \
++#if !defined(BEOS) && !defined(HAIKU) && !defined(AMIGA) && !defined(MSWIN32) \
```
pulkomandy: Duplicate, it was already added later in the condition

dev-libs/boost/patches/boost-1.60.0.patchset
```diff
@@ -0,0 +1,136 @@
...
+index 323ba15..f432952 100644
+--- a/tools/build/src/engine/jambase.c
++++ b/tools/build/src/engine/jambase.c
+@@ -548,6 +548,18 @@ char *jambase[] = {
+ "NOARSCAN    ?= true ;\n",
+ "STDHDRS     ?= /boot/develop/headers/posix ;\n",
+ "}\n",
++"else if $(OS) = HAIKU\n",
++"{\n",
++"BINDIR      ?= /boot/common/bin ;\n",
```
pulkomandy: This directory is deprecated

korli:
IMO there is more breakage than this. When no renaming is used, this implies a rebuild of affected packages.

pulkomandy:
Judging from http://upstream.rosalinux.ru/versions/boost.html the changelog seems reasonably accurate, but still there are some changes detected (and latest versions are not listed since the page is sadly not updated anymore).

waddlesplash:
Any update?

diversys:
Superseded by 218dfd4205020520d3877b1084471e48ab0cfd69

PR #570, opened 2016-04-20
sqlite: bump to 3.12.2, reorder sections, add TEST().

fbrosson:


fbrosson:
I've noticed that we can use aclocal and automake instead of aclocal-1.13 and automake-1.13. Shall I clean the recipe or revert that change?
I also added **`$jobArgs`** to **`make`**. But I'm not sure this is a good idea because sqlite only has 2 source files:
```Shell
ls -go *.c
-rw-r--r-- 1  159755 Apr 18 18:56 shell.c
-rw-r--r-- 1 6640394 Apr 18 18:56 sqlite3.c
```
It probably does not make much difference. What do you think?

fbrosson:
I've cleaned the recipe by removing the lines which I had commented out.
Please let me know if I should revert the changes regarding aclocal, automake and $jobArgs.
BTW, I won't merge this PR myself because I might be missing something.
Thanks in advance for your feedback and/or for the merge.

mmlr:
If you take a look at the commit logs you'll see that I explicitly removed jobArgs for an earlier recipe as the build does not support parallel builds. To prevent any confusion in the future please add a comment explaining this in the recipe (as I should have done...).

fbrosson:
Removed $jobArgs as requested. I'll check the logs next time. Sorry, and thanks for spotting it.

fbrosson:
Oops, I had forgotten to have a look at the logs...
I dropped the $jobArgs I had added.
As you suggested I added a comment in BUILD() to explain $jobArgs should not be used.
Thanks for spotting the issue ;-)

PR #572, opened 2016-04-25
new recipes for TuxRacer

Begasus:
the library tcl had to be altered to use portVersion >= 8.0 instead of 8.5
best to be played with hardware accelaration :)

### DIFF
--- a/games-action/tuxracer/patches/tuxracer-0.61.patch
+++ b/games-action/tuxracer/patches/tuxracer-0.61.patch
@@ -1,0 +1,33 @@
+--- tuxracer-0.61/src/game_config.c.org	2016-04-26 17:44:17.903086080 +0200
++++ tuxracer-0.61/src/game_config.c	2016-04-26 17:44:05.612892672 +0200
+@@ -505,7 +505,7 @@
+         "# decreasing this number, at the cost of lower image quality." );
+ 
+     INIT_PARAM_BOOL( 
+-	fullscreen, True,
++	fullscreen, False,
+ 	"# If true then the game will run in full-screen mode." );
+ 
+     INIT_PARAM_INT( 
+--- tuxracer-0.61/src/main.c.org	2016-04-26 22:09:04.763887616 +0200
++++ tuxracer-0.61/src/main.c	2016-04-26 22:08:49.627572736 +0200
+@@ -56,6 +56,11 @@
+ #include "loading.h"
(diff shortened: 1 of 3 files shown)


games-action/tuxracer/patches/tuxracer-0.61.patch
```diff
@@ -0,0 +1,20 @@
+--- tuxracer-0.61/src/game_config.c.org	2016-04-25 18:59:37.147587072 +0200
++++ tuxracer-0.61/src/game_config.c	2016-04-25 18:59:23.122945536 +0200
+@@ -79,7 +79,7 @@
+ #  if defined( WIN32 )
+ #    define DATA_DIR "."
+ #  else
+-#    define DATA_DIR "/usr/local/share/tuxracer"
++#    define DATA_DIR "/boot/home/config/apps/TuxRacer/data"
```
korli: You know that this depends on where the data package is installed (ie this doesn't work when data are in /system/packages)

korli: Something like this should work: $portPackageLinksDir/tuxracer_data/apps/TuxRacer/data
$portPackageLinksDir is to be replaced with a macro at build time.

Begasus: working on the configure option datadir, that works so far, but I can't seem to install the data files to the /boot/system/apps .. folder, I can drop it, don't get an error but it doesn't show up, neither does TuxRacer after dropping it in /boot/system/packages .. wip

Begasus: Not familiar with macros, but with the latest changes you can even install them in /boot/system/packages ...

scottmc: how about a patch you can upstream?  maybe try addin an "if defined( HAIKU
)" instead of breaking it on systems that use
/usr/local/share/putmyprogramhere

-scott

Begasus: With the latest changes there is no need to a new define, the DATA_DIR patch has been removed and the changes needed were added to configure ...
+	./configure --prefix=$appsDir/TuxRacer --datadir=$appsDir/TuxRacer/data \
+	 --with-data-dir=$appsDir/TuxRacer/data --with-tcl-lib-name=tcl8.5

PR #575, opened 2016-04-26
libcaca: shrink 2nd arch packages to avoid duplicate files.

fbrosson:
* Drop the mirror for SOURCE_URI, since the official website is back.
* Add **`TEST()`** with **`make check`** since this target exists.
* No changes in primary arch packages, except for the SOURCE_URI.
* On secondary arch there is no need to ship man pages, data files and runtimes (except for caca-config-*), since they are already in the primary arch packages. All we need for 2nd arch packages is:
 * in the base package: the shared libs;
 * in the devel package: **`caca-config-*`** and **`develop/{headers,lib}`**.
  The only file we need to keep in every package is the licence.
* Cosmetic edits: duplicate white spaces, fake tab, **`-e`** for sed.

### DIFF
--- a/media-libs/libcaca/libcaca-0.99.beta19.recipe
+++ b/media-libs/libcaca/libcaca-0.99.beta19.recipe
@@ -2,8 +2,6 @@ SUMMARY="An ASCII art rendering library"
 DESCRIPTION="libcaca is a software library which allows applications to \
 automatically convert still and moving images into colored ASCII art."
 HOMEPAGE="http://caca.zoy.org/wiki/libcaca"
-# FIXME: the offical website is down right now. Using mirror provided by the
-#        MXE project.
 COPYRIGHT="1998, 1999, 2001 Jan Hubicka
 	2003-2008 Sam Hocevar
 	2003-2008 Jean-Yves Lamoureux
@@ -12,9 +10,8 @@ COPYRIGHT="1998, 1999, 2001 Jan Hubicka
 	2007-2008 Pascal Terjan"
 LICENSE="WTFPL
 	GNU GPL v2"
-REVISION="2"
-SOURCE_URI="http://caca.zoy.org/raw-attachment/wiki/libcaca/libcaca-$portVersion.tar.gz
-	http://s3.amazonaws.com/mxe-pkg/libcaca-$portVersion.tar.gz"
+REVISION="3"
+SOURCE_URI="http://caca.zoy.org/raw-attachment/wiki/libcaca/libcaca-$portVersion.tar.gz"
 CHECKSUM_SHA256="128b467c4ed03264c187405172a4e83049342cc8cc2f655f53a2d0ee9d3772f4"
 SOURCE_DIR="libcaca-$portVersion"
 PATCHES="libcaca-${portVersion}.patchset"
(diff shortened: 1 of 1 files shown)


fbrosson:
I've carefully compared the differences before and after this commit, in the base primary arch, devel primary arch, base secondary arch and devel secondary arch. (Tests done in x86_gcc2 primary arch with x86 as secondary arch.)
The results are that primary arch packages are identical before and after this commit, and that:
* the following files no longer exist in the new base secondary arch package:
```
bin/x86/cacaclock
bin/x86/cacademo
bin/x86/cacafire
bin/x86/cacaplay
bin/x86/cacaserver
bin/x86/cacaview
bin/x86/img2txt
data/libcaca/caca.txt
documentation/man/man1/caca-config.1
documentation/man/man1/cacafire.1
documentation/man/man1/cacaplay.1
documentation/man/man1/cacaserver.1
documentation/man/man1/cacaview.1
documentation/man/man1/img2txt.1
```

* the following files no longer exist in the new devel secondary arch package:
```
develop/documentation/libcaca_x86/*.html
develop/documentation/libcaca_x86/*.png
develop/documentation/libcaca_x86/*.css
documentation/man/man3/*.3caca
```

I have taken care of not discarding caca-config-*.

I'd like to have your thougths (if any ;-) about the name of a variable I introduced in the recipe: **`maybe_manDir_man3`**. It is set to $manDir/man3 if when building for a primary arch, or left empty when building for a secondary arch.
I think it is quite self-explanatory, but I'm open to any alternate suggestion.

Regarding the general changes proposed in this PR, I think it improves the recipe because it gets rid of all overlapping files (except for the licence which we need to keep in order to satisfy haikuports.)
I know there aren't many recipes that depend on libcaca but this allows to have some sort of template for what we can do with recipes that build in all primary and secondary architectures.
Thanks in advance for your feedback!

korli:
Isn't libcaca.txt the license text file?
I fail to see the need for "documentation/man/man3/*.3caca". There is already the html documentation in the package.

fbrosson:
Nope, the license file is WTFPL (and it's in data/licenses/). BTW, there is no libcaca.txt but a **`caca.txt`** (in **`data/libcaca/`**) and it's just a small ascii art picture.
Regarding the **`documentation/man/man3/*.3caca`** files, although the information they carry is probably also available in html format, I think it is better to not drop them , I mean, the *.3caca files. If keeping them in the primary arch devel package does not seem a good idea then I can move them to an alternate sub-package like libcaca_man or libcaca_doc if you agree. Because if we remove that stutff then it might upset developers who prefer reading real man pages rather then html files. And we might even miss some information if we only ship the documentation in html format.

korli:
At least arch completely removes the doc for this package, so it must not upset many people... another _doc package might be a good idea.

fbrosson:
Cool, thanks! I've moved the man3 pages (and the HTML documentation) to a **`_doc`** arch-independent sub-package as discussed. I think it makes sense to also move the HTML docs as I did. What do you think?
Regarding the static libs, is there any recommendation to pass **`--disable-static`** to configure when static libs are enabled by default? (libcaca builds both static and shared libs by default). In case there is no general recommendation, do you think we should drop the static libs for this particular recipe?
If you think the recipe is ready for merge then please feel free to either merge it yourself or, if you prefer, give me green lights to do so. Of course, if you would like to see other changes, please do not hesitate! Thanks!

korli:
I have no idea concerning static libs. Why dropping it for this recipe?

fbrosson:
Cause I guess nobody will ever link anything against libcaca's static libs. But of course it's OK to keep them, since they were already in the previous revision.
BTW, thanks for the merge!

PR #577, opened 2016-04-27
PenguinCommand new recipe

Begasus:


korli:
Have you tried patching configure.in and regenerating configure, instead of patching configure?

Begasus:
Not yet, I only got the error on -lm after the first attempt (for configure.ac) I'll try to look into it later or tomorrow

Begasus:
hmm not configure.ac ... wip

Begasus:
Had no luck with configure.in, but updated the recipe a bit

games-arcade/penguin_command/penguin_command-1.6.11.recipe
```diff
@@ -0,0 +1,55 @@
+SUMMARY="A clone of the classic Missile Command game"
+DESCRIPTION="Penguin Command is a clone of the classic Missile Command game, \
+but has better graphics and sound. The gameplay has been slightly modified."
+HOMEPAGE="http://www.linux-games.com/penguin-command/"
+COPYRIGHT="2007 Karl Bartel, Skaven"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://sourceforge.net/projects/penguin-command/files/penguin-command/$portVersion/penguin-command-$portVersion.tar.gz"
```
fbrosson: May I suggest using an alternate SOURCE_URI? With the one we have (in the current commit candidate),
```Shell
SOURCE_URI="https://sourceforge.net/projects/penguin-command/files/penguin-command/$portVersion/penguin-command-$portVersion.tar.gz"
```
we get _three_ "302 Found" before the final "200 OK", whereas with:
```Shell
SOURCE_URI="https://downloads.sf.net/penguin-command/penguin-command-$portVersion.tar.gz"
```
we get a "301 Moved Permanently", then a "302 Found" and then the final "200 OK", thus saving one redirect compared to the long address ;-)
Moreover, with the short address the recipe is safe even if the tarball gets moved to another directory of the same project. (I don't know if that's the case for this SF project but some do move old tarballs to subdirectories.)
(Of course, I used several calls to **`curl --head `** in order to count the number of redirects.)

Regarding the https vs http choice, SourceForge currently redirects all https download traffic to http, so I don't know if SF likes to receive https traffic but I prefer using https (like you did) just in case SF would change their rules and accept to serve downloads through https.
Cheers!

Begasus: At your service ;)

PR #578, opened 2016-04-28
update yab and yab-ide

bbjimmy:
yab Version 1.7.5.3, yab-IDE version 2.2.6
- gcc4 now compiles
- fix system shutdown bug
- remove un-needed BuildFactory files
- add https:// to texturl
- BuildFactory use declared appsig
- yab-IDE use declared appsig, or use filename for appsig


humdingerb:
Looks good to me. No real changes really to the older version. It does replace that older version instead of adding a new recipe for the new version. I guess that is OK for a minor release like this. For "bigger" releases it'd be useful to keep the old recipe around for people to be able to downgrade.

PR #579, opened 2016-04-30
ffmpeg: drop 2.6 & 2.7, bump to 2.8.7 & 3.0.2, minor fixes. (WIP)

fbrosson:
* Drop 2.6.3 & 2.7.2, bump 2.8.6 to 2.8.7 and 3.0.1 to 3.0.2.
* ffmpeg, ffprobe and ffserver were being shipped in secondary arch builds (due to **`rm $binDir`** instead of **`rm $prefix/bin`** where they were being installed) but did not have a **`cmd:`** in PROVIDES. Fixed this by moving these to a **`_bin`** subpackage on x86_gcc2 using 2nd arch x86.
* Fix CONFLICTS_devel in 3.0.2 (missing **`_legacy`**),
* Add ffmpeg2_bin to CONFLICTS in 3.0.2.
* Add ffmpeg and ffmpeg_legacy to CONFLICTS_bin in 2.8.7.
* Reorder some sections.
* Switch all HOMEPAGE and SOURCE_URI to https.
* Switch from tar.bz2 to tar.xz for 2.8.7.
* 2.8.7 and 3.0.2 build OK (on x86_gcc2 using x86 as 2nd arch) but **`make check`** fails with **`runtime_loader: Cannot open file libavutil.so.55: No such file or directory.`** Failure happens on **`TEST api-threadmessage`** for 3.0.2 and on **`TEST api-flac`** for 2.8.7.


korli:
About 2.6.8 and 2.7.6, I think it needs the patch from 2.8 since the gcc5 upgrade. We could also drop these two recipes.

fbrosson:
Thanks for the hint about the patch.  Since upstream maintains all these versions it might be useful to have them all too. Of course, if you think it's better to just drop 2.6.8 and 2.7.6 then it's even easier. But I'd be glad to try and make these work too.

korli:
ffmpeg maintains various versions because some Linux distributions stick to a ffmpeg branch (security patches). We don't need that, when we have and use 2.8 and 3.0.

fbrosson:
OK. Well, I guess I can remove 2.7.6 and 2.6.8 then. (if you change your mind please don't hesitate to ask.)

Regarding 2.8.x, I noticed that 2.8.6 does not declare any cmd: in its PROVIDES.
But it passes **`--bindir=$prefix/bin`** to configure and later deletes $binDir... so when building on x86_gcc2 for x86 as secondary arch we ended up with ffmpeg, ffprobe and ffserver in bin (because $binDir != $prefix/bin for 2nd arch builds). I noticed this because I call haikuporter with -S.
Do we want these cmd: for 2.8.7? Or do we just want the libs for 2.8.7 and everything for 3.0.2?

fbrosson:
I've dropped 2.6 and 2.7 as suggested. I also made a few other changes (e.g. with the package splitting and all the CONFLICTS) but I haven't tested them on all architectures yet and I don't know if they are politically compatible with haikuports' guidelines. BTW, there are quite some changes so please feel free to postpone the review until after I drop the "WIP" tag. Do not hesitate, however, to comment at any time ;-)

waddlesplash:
Yes, the package splitting is just fine and compatible with guidelines. Most packages that explicitly want cmd:ffmpeg should ask for it specifically.

fbrosson:
Thanks @waddlesplash!

PR #581, opened 2016-05-01
libcddb: do not ship the command line tool for 2nd arch.

fbrosson:


korli:
"Runtimes" is kind of misleading. "Command line tools" fit better with what was removed.

fbrosson:
I've just tried to improve the commit message, but I don't know if it"s 100% ok now.
English is not my native language so thanks a lot for your remark and suggestion. BTW, please do not hesitate to spot any other mistakes, even small, that I may make from time to time.
Regarding the removal of the static lib, do you think it's a good idea? (If not I'll revert it, of course.)

korli:
Regarding the static libs, I don't know. It's certainly a problem for projects which don't want to link against a shared lib, but given that libcddb is a LGPL project, this is even more unlikely.

fbrosson:
OK. Thanks!

PR #584, opened 2016-05-03
openssl: bump to 1.0.1t and 1.0.2h.

fbrosson:
* Bump versions.
* Use **`-e`** for **`sed`** to make the recipe more portable, since some systems have sed implementations that require it.
* Add **`cmd:cmp`** to BUILD_PREREQUIRES since **`make test`** needs it.

### DIFF
--- a/dev-libs/openssl/openssl-1.0.1t.recipe
+++ b/dev-libs/openssl/openssl-1.0.1t.recipe
@@ -17,7 +17,7 @@ COPYRIGHT="1995-1998 Eric Young
 LICENSE="OpenSSL"
 REVISION="1"
 SOURCE_URI="https://www.openssl.org/source/openssl-$portVersion.tar.gz"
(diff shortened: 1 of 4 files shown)


korli:
"since some systems have sed implementations that require it.", we don't intend to port any recipe to such a system, so why the trouble?

fbrosson:
If I'm not mistaken, it is possible to cross-build Haiku on FreeBSD. And, if I'm not mistaken again, sed on FreeBSD requires the **`-e`**. So if someone manages to use **`haikuports/haikuports`** from FreeBSD just like **`haikuports/haikuports.cross`** then there will be no missing **`-e`** for this recipe.
I undestarnd that making the change if there is no other change on a recipe is not a good idea. But since we're bumping a recipe I though it could be a good idea.
Of course, if you prefer to not have this change then I can remove it. Just let me know and I'll do so.

korli:
It certainly makes sense at haikuports.cross, but the recipes at haikuports aren't meant to build on FreeBSD (and won't ever).

fbrosson:
OK, fixed.

korli:
Thanks!

PR #585, opened 2016-05-06
abe, updated recipe, no fullscreen on startup

Begasus:


korli:
Please bump the revision.

fbrosson:
Haven't tried yet, but I was wondering if this would work:
```
runConfigure --omit-dirs binDir ./configure --bindir=$appsDir/Abe
```
If I'm not mistaken it would allow to drop the call to **`addAppDeskbarSymlink $appsDir/Abe/Abe Abe`** and would produce a cleaner $sppsDir. What do you think?

Begasus:
Bumpt the revision, changed the runConfigure line also (thanks fbrosson!)

PR #588, opened 2016-05-06
sdl2_mixer, updated recipe

Begasus:
Build doesn't work for gcc2 (SDL2_mixer-2.0.1/wavestream.c:94: field `loops' has incomplete type), I don't have the skills to patch this one
Missing libgl as a requirement

### DIFF
--- a/media-libs/sdl2_mixer/sdl2_mixer-2.0.1.recipe
+++ b/media-libs/sdl2_mixer/sdl2_mixer-2.0.1.recipe
@@ -7,14 +7,14 @@ Ogg Vorbis, and SMPEG MP3 libraries."
 HOMEPAGE="http://www.libsdl.org/projects/SDL_mixer/"
 COPYRIGHT="1997-2012 Sam Lantinga"
 LICENSE="Zlib"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="http://www.libsdl.org/projects/SDL_mixer/release/SDL2_mixer-$portVersion.tar.gz"
 CHECKSUM_SHA256="5a24f62a610249d744cbd8d28ee399d8905db7222bf3bdbc8a8b4a76e597695f"
 SOURCE_DIR="SDL2_mixer-$portVersion"
 PATCHES="sdl2_mixer-${portVersion}.patch"
 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	sdl2_mixer$secondaryArchSuffix = $portVersion compat >= 2.0
@@ -23,9 +23,10 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libSDL2_2.0$secondaryArchSuffix
+	lib:libgl$secondaryArchSuffix
+	lib:libsdl2_2.0$secondaryArchSuffix
 	lib:libogg$secondaryArchSuffix
-	lib:libFLAC$secondaryArchSuffix
+	lib:libflac$secondaryArchSuffix
 	lib:libfluidsynth$secondaryArchSuffix
 	lib:libmad$secondaryArchSuffix
 	lib:libvorbis$secondaryArchSuffix
@@ -66,7 +67,7 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	libtoolize --force --copy --install
+	libtoolize -fci
 	aclocal --force --install -I acinclude
 	autoconf
 	runConfigure ./configure


media-libs/sdl2_mixer/sdl2_mixer-2.0.1.recipe
```diff
@@ -23,9 +23,10 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libSDL2_2.0$secondaryArchSuffix
+	lib:libgl$secondaryArchSuffix
+	lib:libsdl2_2.0$secondaryArchSuffix
 	lib:libogg$secondaryArchSuffix
-	lib:libFLAC$secondaryArchSuffix
+	lib:libflac$secondaryArchSuffix
```
fbrosson: Good catch! I had not noticed there was an unaesthetic-only case inconsistency in sdl2_mixer-2.0.1.recipe, **`lib:libFLAC`** in REQUIRES and **`devel:libflac`** in BUILD_REQUIRES.
I noticed, however, that there are aroud 7 recipes with **`lib:libflac`** and around 12 with **`lib:libFLAC`**. Since the name of the lib is libFLAC, wouldn't it be _sligthly_ better to change **`devel:libflac`** into **`devel:libFLAC`** (instead of changing **`lib:libFLAC`** to **`lib:libflac`**) ?
I do understand that all this does not make much difference, but I was wondering if there was a recommended or prefered spelling for **`lib:`** and **`devel:`** items for libs that have uppercase letters in their names.

Begasus: IIRC there was an issue to use lowercase when possible, mixing them up wouldn't be a good idea, the provides is another thing as using the requirements .. or if I'm wrong please correct me (but I've been corrected in the past for this also -- using uppercase, having a clear understanding on what is prefered would be nice :)

fbrosson: Well, in that case I guess you are right (but I can't confirm). Can someone confirm?

scottmc:
I haven't checked lately but years ago there was an issue getting flac to work, and only solution then was to upcase it to libFLAC.  Maybe things have changed since then.  Try testing it.

fbrosson:
OK, thanks @scottmc! I think it works both ways, otherwise we would have noticed something, because among the recipes that need libFLAC 60% have lib:libFLAC and 40% have lib:libflac. But I'll dig this in a few days.

Begasus:
Years ago there was an issue with using uppercase/lowercase, I noticed that when building for instance flac you need to use libFLAC, as a requirement it's ok to use libflac ...

media-libs/sdl2_mixer/sdl2_mixer-2.0.1.recipe
KapiX: I can confirm that there was an issue in Haikuporter that entries in PROVIDES did not get converted to lowercase and those in REQUIRES did (or the other way around, I don't remember), resulting in unresolved dependency to libFLAC. This should be fixed now I think, but I'd prefer if all recipes used lowercase everywhere.

fbrosson:
OK, I see. Thanks!

media-libs/sdl2_mixer/sdl2_mixer-2.0.1.recipe
waddlesplash: > I do understand that all this does not make much difference, but I was wondering if there was a recommended or prefered spelling for lib: and devel: items for libs that have uppercase letters in their names.

HaikuPorter uses lowercase internally, but it's preferable to have it in the recipe the way it is on disk (so libgl is actually -lGL when being passed to GCC, for example, so use libGL.)

waddlesplash:
If there is any difference in PROVIDES/REQUIRES/etc. for FLAC vs flac, that's a bug in HaikuPorter. I patched the first set of case-consistency errors some time ago, if there are more, please report them.

media-libs/sdl2_mixer/sdl2_mixer-2.0.1.recipe
fbrosson: OK, thanks. That's cool!

Begasus:
it builds with either flac or FLAC, so if it's ok to use FLAC as a requirement please give a notice :)

waddlesplash:
Is the library in the filesystem "libFLAC.so" or "libflac.so"? Go with whichever it is.

fbrosson:
In the filesystem it is libFLAC.so. But this uppercase/lowercase issue about the REQUIRES and PROVIDES seems complex.
I've been preparing a few fixes/changes for libxau, libxcb, libx11 and libxpm, and I noticed that using lib:libXau in the PROVIDES of libxau and the same in the REQUIRES of libxcb does not work because haikuporter converts lib:libXau to lib:libxau, so it seems we still need to stick to lowercase names.
I can't explain, however, why lib:libGL only works like that. Well, maybe this is due to the fact that once you remove the "lib" prefix the rest is _all_ uppercase. This is just a speculation, though, because I haven't dug haikuporter enough.

waddlesplash:
> lib:libXau in the PROVIDES of libxau and the same in the REQUIRES of libxcb does not work because haikuporter converts lib:libXau to lib:libxau

It should convert both of those to lowercase. If it doesn't, as stated before, please file a bug report to HaikuPorter with a reproducible testcase.

PR #593, opened 2016-05-09
Fix 2nd arch for live555 & aplayer; cosmetic edits elsewhere.

fbrosson:
* live555: Use **`$developLibDir`** instead of **`$developDir/lib`** since the latter is not appropriate for 2nd arch builds.
* aplayer: Drop **`$secondaryArchSuffix`** for **`cmd:nasm`** in BUILD_PREREQUIRES.
* re2c: use tabs (instead of white spaces) in BUILD_REQUIRES.
* Remove x86_64 from SECONDARY_ARCHITECTURES for sdl2_net, edgar, paladin-git and mediainfo-0.7.61.


fbrosson:
I did not dare to bump REVISION for live555 because the error is very recent and only affects 2nd arch. But I can bump it if someone asks.
Sometimes is is 100% OK to keep the same revision, and sometimes it is not. For situations like this one, where it is _not_ 100% safe to keep the revision unchanged, is it good practive to refrain from bumping it? Thanks!

waddlesplash:
Generally, if your uncertain if you should or shouldn't bump revision, bump it anyway.

fbrosson:
OK, done.
BTW, the rule is cool!

haiku-apps/aplayer/aplayer-4.0_svn.recipe
```diff
@@ -31,7 +31,7 @@ BUILD_PREREQUIRES="
 	cmd:g++$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
 	cmd:make
-	cmd:nasm$secondaryArchSuffix
+	cmd:nasm
```
waddlesplash: ?

fbrosson: If someone tries to build aplayer for a 2nd arch, **`cmd:nasm$secondaryArchSuffix`** won't be available because SECOND_ARCHITECTURES is not defined in nasm's recipe. BTW, all the other recipes that need **`cmd:nasm`** do not have the $secondaryArchSuffix for it.
(I guess you have found the explation soon after you posted the question mark ;-)

waddlesplash: Ah, makes sense

PR #594, opened 2016-05-11
check: fix PROVIDES for secondary arch & improve pkg splitting.

fbrosson:
* 0.9.13, 0.9.14 and 0.10.0 were missing a **`cmd:gawk`** in REQUIRES.
* Fixed 2nd arch for 0.10.0 and improved the package splitting.
* Switch 0.9.13 and 0.9.14 to the same layout as 0.10.0.

### DIFF
--- a/dev-libs/check/check-0.10.0.recipe
+++ b/dev-libs/check/check-0.10.0.recipe
@@ -9,7 +9,7 @@ HOMEPAGE="https://libcheck.github.io/check/"
 COPYRIGHT="2001-2015 Arien Malec, Branden Archer, Chris Pickett, Fredrik \
 Hugosson, and Robert Lemmen."
 LICENSE="GNU LGPL v2.1"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/libcheck/check/files/71408/check-$portVersion.tar.gz"
 CHECKSUM_SHA256="f5f50766aa6f8fe5a2df752666ca01a950add45079aa06416b83765b1cf71052"
 PATCHES="check-$portVersion.patch"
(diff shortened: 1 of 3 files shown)


waddlesplash:
Why do we have 3 recipes anyway? Can we dump all but the latest version? I don't see any reason why not...

fbrosson:
I also though about dropping the old recipes, but after some tests I came to the conclusion that it could be worth keeping them all. What I noticed is that the results of:
* **`haikuporter -E check-0.9.13`** follwed by **`make check`**
* **`haikuporter -E check_x86-0.9.13`** follwed by **`make check`**
* **`haikuporter -E check-0.9.14`** follwed by **`make check`**
* **`haikuporter -E check_x86-0.9.14`** follwed by **`make check`**
* **`haikuporter -E check-0.10.0`** follwed by **`make check`**
* **`haikuporter -E check_x86-0.10.0`** follwed by **`make check`**

are quite different.
Hopefully, **`make check`** works for all versions (on both primary and secondary architectures), but sometimes the tests do not work in the chrooted environment. The strange thing is that I managed to have 0.9.13 pass all 7 tests in a chrooted env (using the -E option of haikuporter), but for 0.9.14 and 0.10.0 I don't get 100% success in a chrooted envs. Here's what I get:

version | env | test in non-chrooted env | test in chrooted env
---------- | ----- | --------------------------------- | ---------------------------
0.9.13 | x86_gcc2 primary | All 7 tests passed | All 7 tests passed
0.9.13 | x86 secondary | All 7 tests passed | 1 of 7 tests failed
0.9.14 | x86_gcc2 primary | All 7 tests passed | 1 of 7 tests failed
0.9.14 | x86 secondary | All 7 tests passed | 2 of 7 tests failed
0.10.0 | x86_gcc2 primary | All 8 tests passed | (forgot)
0.10.0 | x86 secondary | All 8 tests passed | (forgot)

Since the chrooted env haikuporter creates for 0.9.13 seems better than the one created for 0.9.14 and 0.10.0, I though it could be interesting to keep all 3 versions until we find out where the problem comes from. I know I should create an issue for that, but I first wanted to dig a little bit more to make sure the problem is not just on my machine.
BTW, when I tested without a chrooted env, I did take care of uninstalling all other versions and only have the appropriate hpkgs installed.
BTW, I've been looking for an option in haikuporter to make it call **`TEST()`** but did not find such a feature. Does such an option already exist?

waddlesplash:
> Since the chrooted env haikuporter creates for 0.9.13 seems better than the one created for 0.9.14 and 0.10.0, I though it could be interesting to keep all 3 versions until we find out where the problem comes from.

Interesting. What are the error messages from the later versions? I'd be more inclined to merge anyway, since it works outside chroot.

> BTW, I've been looking for an option in haikuporter to make it call TEST() but did not find such a feature. Does such an option already exist?

`--test`?

fbrosson:
Thanks for the merge and sorry for this late reply.

> **`--test`**

Thanks! It works like a charm. BTW, I don't know how I managed to miss this option!

> Interesting. What are the error messages from the later versions? I'd be more inclined to merge anyway, since it works outside chroot

With 0.10.0 using **`--test`** I get **`2 of 8 tests failed`** on both primary and secondary architectures.

PR #595, opened 2016-05-16
pipepanic, no need for $secondaryArchSuffix

Begasus:


korli:
This is kind of a useless change, isn't it?

Begasus:
Maybe, but iirc there has been a conversation to not use X86 architectures if not needed ...

korli:
Hmm yeah as secondary arch, it's fine as primary arch.
-1 from me.

Begasus:
Even if it builds with gcc2_x86?

waddlesplash:
The recipe does not (and will never) need to be built on anything but primary arch; so there's no need to have the machinery for that in the recipe file.

korli:
Augustin, IMO you should really think twice before merging any patch that comes up. Just have a look at the commit message... This commit also changes supported primary architectures for no good reason.
This kind of cleanup can be done through a script. Why should we accept such a change for one recipe only, this is more pollution. Just my thought.

waddlesplash:
> This kind of cleanup can be done through a script. Why should we accept such a change for one recipe only, this is more pollution. Just my thought.

Ah, true, that's a good point. Makes sense; I'll reject changes like this in the future.

PR #596, opened 2016-05-16
gettext: bump to 0.19.8.1, keep & fix lib*.la, sync patch.

fbrosson:
Some recipes, e.g. mc (Midnight Commander), need libintl.la, but we were previously dropping all **`develop/lib/lib*.la`** files in gettext.
We now ship them and also fix them to make sure they don't become unusable in case devel:libiconv or devel:libncurses get updated.
On x86 and x86_64 we also fix **`$developLibDir/libasprintf.la`** to use a path to libstdc++.la that does not depend on the revision of the gcc package.

### DIFF
--- a/sys-devel/gettext/gettext-0.19.8.1.recipe
+++ b/sys-devel/gettext/gettext-0.19.8.1.recipe
@@ -13,30 +13,30 @@ massage in various ways the sets of translatable strings, or already \
 translated strings. A special GNU Emacs mode also helps interested parties in \
 preparing these sets, or bringing them up to date.
 "
-HOMEPAGE="http://www.gnu.org/software/gettext/"
+HOMEPAGE="https://www.gnu.org/software/gettext/"
+COPYRIGHT="1998-2016 Free Software Foundation, Inc."
 LICENSE="GNU GPL v3"
-COPYRIGHT="1998, 2010 Free Software Foundation, Inc."
-SOURCE_URI="http://ftp.gnu.org/pub/gnu/gettext/gettext-$portVersion.tar.gz"
-CHECKSUM_SHA256="5386d2a40500295783c6a52121adcf42a25519e2d23675950619c9e69558c23f"
 REVISION="1"
+SOURCE_URI="http://ftpmirror.gnu.org/gettext/gettext-$portVersion.tar.gz"
+CHECKSUM_SHA256="ff942af0e438ced4a8b0ea4b0b6e0d6d657157c5e2364de57baa279c1c125c43"
+PATCHES="gettext-$portVersion.patchset"
+
 ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
 SECONDARY_ARCHITECTURES="x86_gcc2 x86"
-
-PATCHES="gettext-$portVersion.patchset"
 
 PROVIDES="
 	gettext$secondaryArchSuffix = $portVersion compat >= 0.18
 	devel:libasprintf$secondaryArchSuffix = 0.0.0 compat >= 0
 	devel:libgettextlib$secondaryArchSuffix = $portVersion
-	devel:libgettextlib_0.19.7$secondaryArchSuffix = $portVersion
-	devel:libgettextpo$secondaryArchSuffix = 0.5.3 compat >= 0
+	devel:libgettextlib_$portVersion$secondaryArchSuffix
+	devel:libgettextpo$secondaryArchSuffix = 0.5.4 compat >= 0
 	devel:libgettextsrc$secondaryArchSuffix = $portVersion
-	devel:libgettextsrc_0.19.7$secondaryArchSuffix = $portVersion
-	devel:libintl$secondaryArchSuffix = 8.1.4 compat >= 8
+	devel:libgettextsrc_$portVersion$secondaryArchSuffix
+	devel:libintl$secondaryArchSuffix = 8.1.5 compat >= 8
 	lib:libasprintf$secondaryArchSuffix = 0.0.0 compat >= 0
-	lib:libgettextlib_0.19.7$secondaryArchSuffix = $portVersion
-	lib:libgettextpo$secondaryArchSuffix = 0.5.3 compat >= 0
-	lib:libgettextsrc_0.19.7$secondaryArchSuffix = $portVersion
+	lib:libgettextlib_$portVersion$secondaryArchSuffix
+	lib:libgettextpo$secondaryArchSuffix = 0.5.4 compat >= 0
+	lib:libgettextsrc_$portVersion$secondaryArchSuffix
 	cmd:autopoint$secondaryArchSuffix = $portVersion compat >= 0
 	cmd:envsubst$secondaryArchSuffix = $portVersion compat >= 0
 	cmd:gettext$secondaryArchSuffix = $portVersion compat >= 0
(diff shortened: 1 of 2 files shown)


sys-devel/gettext/gettext-0.19.7.recipe
```diff
@@ -97,7 +107,6 @@ INSTALL()
 	make install AUTOCONF=: AUTOMAKE=: AUTOHEADER=: ACLOCAL=:
 
 	rm $libDir/charset.alias
-	rm $libDir/lib*.la
```
waddlesplash: CC @korli -- weren't we getting rid of the `.la`s now?

fbrosson:
**`rm $libDir/lib*.la`** in **`gettext-0.19.7.recipe`** was making it impossible to build Midnight Commander (mc, see #597) because it looks for libintl.la. It could only build with **`gettext-0.18.1.1`**. This PR fixes that issue.

Regarding the tarball format switch, the tar.gz is 18 GB while the tar.xz is only 7 GB, i.e. 63% smaller. If, however, there is any reason to continue with the tar.gz then I can revert this change.

I changed **`cmd:libtoolize`** into **`cmd:libtoolize$secondaryArchSuffix`** but did not mention it in the commit message because I felt it was not worth adding an extra line in the commit message to mention such an insignificant detail. But I can do so if needed.

Regarding the DESCRIPTION, I did not get rid of the 2 extra newlines. Should I do so?
If the answer is yes then I'll do so, moving the newlines to respect the 80-column limit, of course.

sys-devel/gettext/gettext-0.19.7.recipe
fbrosson: Oopis,, I only saw your comment after I submitted the comment below.

fbrosson: IIRC, I even tried using **`--disable-static`** together with **`--with-pic=libintl`** in mc before dropping the **`rm $libDir/lib*.la`** in gettext, but I did not manage to make mc build without libintl.la. I guess I missed something.
Anyway, since **`lib*.la`** files are really small it is probably worth keeping a given **`lib*.la`** if we know that at least one recipe needs that particular file.
BTW, is there a tool to dump the list of shared libs that a runtime needs? On Linux I use ldd, but I did not find it on Haiku. Thanks!

diversys: objdump -x /system/lib/libbe.so | grep NEEDED

korli:
Regarding tar.xz we tried to avoid it for core components because x86_gcc2 (not hybrid) doesn't provide xz_utils. Also remember that xz_utils isn't provided by a bootstrapped environment, so by changing to tar.xz, this actually adds on gettext a dependency on xz_utils (which might not be a problem if xz_utils doesn't depend on gettext direcly or not).
If we had a buildbot running the bootstrapping process, it would be easier to pinpoint such problems.

sys-devel/gettext/gettext-0.19.7.recipe
korli: We can keep *.la when they are needed. The only way to know that was to remove them first :=)
BTW when keeping *.la, make sure the paths they might contain are correct (no packages references for instance).

fbrosson: @diversys: Thanks! objdump works like a charm!

@korli: **`grep ^dependency_libs /system/develop/lib/libintl.la`** yelds:
```
dependency_libs=' /packages/libiconv-1.13.1-6/.self/develop/lib/libiconv.la'
```
Should the INSTALL() in gettext install a modified libintl.la to have this?
```
dependency_libs=' /system/develop/lib/libiconv.la'
```
Would calling fixPkgconfig do the job?

fbrosson:
* I switched back to tar.gz to avoid possible issues on x86_gcc non hybrid.
* I appended **`$secondaryArchSuffix`** to **`cmd:libtoolize`** in BUILD_PREREQUIRES for 0.19.7. Should I also do so for 0.18.1.1 ? Or is it OK to leave it like it is? Hmm, I can look for all recipes that are missing the **`$secondaryArchSuffix`** after **`cmd:libtoolize`** (among those that have a SECONDARY_ARCHITECTURES, of course) and create a separate PR to fix them all in a single commit.
* 0.18.1.1 has a different TEST than 0.19.7. Should I change **`make check`** to **`make -C gettext-runtime/tests check`** in 0.19.7 ?
* Although using mirrors is recommended, I wonder if it is good practice to change the SOURCE_URI when a previous revision of a recipe already had a different SOURCE_URI for the exact same archive. I did it for 0.19.7, assuming it is OK, but I can revert that, of course.

sys-devel/gettext/gettext-0.19.7.recipe
korli: Have a look in https://github.com/haikuports/haikuports/blob/master/media-libs/freetype/freetype-2.6.3.recipe.

fbrosson: OK, thanks for the template. I'm not running Haiku at the moment (and the PC I run Haiku on is currently stuck with a biiiiiig update of openSUSE) but I'll try to update my commit to include a fix for the **`lib*.la`** files.

fbrosson: Sorry for this late reply. I've just updated the commit candidate. I need to test.

fbrosson:
Wait, I have commented out the **`find . -name configure | xargs touch`** in BUILD() and noticed that it is not required on x86_gcc2, but I forgot to check if that's also true on x86 and x86_64.
In case it is not required in any arch, should I remove it completely?

korli:
Well it's probably safe to remove now that AUTOCONF variable is set to true (if timestamps are reversed, configure will try to regenerate).

fbrosson:
That's it. I've even checked on earch arch (x86_gcc2, x86 and x86_64) that this recipe builds OK and mc builds builds fine (and works OK) with it.
There is only one strange thing: on some of my latest tests I got an unexpected line in **`develop/lib/libasprintf.la`**:
**`dependency_libs=' /packages/gcc-5.3.0_2016_02_29-4/.self/develop/tools/lib/gcc/x86_64-unknown-haiku/5.3.0/libstdc++.la'`**
But maybe this was due to my building gettext on a branch without gcc-5.4 although I had just  pulled the latest commit with gcc-5.4 into master (and forgot to merge it into my gettext branch).
The funny thing is that I got these /packages/gcc-5.3.0_2016_02_29-4 in some of gettext's lib*.la only on x86 and x86_gcc2 but not on x86_gcc2. Gonna rebuild it on each arch again to see if something is wrong. But I guess that if there is something wrong then it's not in this commit.

fbrosson:
I've added some extra steps to fix the **`devel/lib/libasprintf.la`** on x86 and x86_64. I don't know, however, if the fix is clean enough for a merge.
Regarding the version bump, I did not remove 0.19.7 yet. May I do so?
I also need to double check if the following bloc works as expected on secondary architecures (as it does on primary):
```
		local libstdcpp="$portPackageLinksDir/gcc$secondaryArchSuffix"
		sed -i \
			-e "s,/packages/gcc-[^\ ]*/\(develop/tools/[^\ ]*libstdc++.la\),$libstdcpp/\1," \
			$developLibDir/libasprintf.la
	fi
```

fbrosson:
According to the comparison of my recipes for 0.19.7 and 0.19.8 it looks like upstream has incorrectly bumped several SO versions in 0.19.8:
```Diff
diff gettext-0.19.7.recipe gettext-0.19.8.recipe |tail -n +13
32c32
< 	devel:libgettextpo$secondaryArchSuffix = 0.5.3 compat >= 0
---
> 	devel:libgettextpo$secondaryArchSuffix = 1.5.3 compat >= 0
35c35
< 	devel:libintl$secondaryArchSuffix = 8.1.4 compat >= 8
---
> 	devel:libintl$secondaryArchSuffix = 9.1.4 compat >= 8
38c38
< 	lib:libgettextpo$secondaryArchSuffix = 0.5.3 compat >= 0
---
> 	lib:libgettextpo$secondaryArchSuffix = 1.5.3 compat >= 0
81c81
< 	lib:libintl${secondaryArchSuffix} = 8.1.4 compat >= 8
---
> 	lib:libintl${secondaryArchSuffix} = 9.1.4 compat >= 8
```
Hopefully, they have fixed it in 0.19.8.1 (which was released yesterday), as this comparison between 0.19.7 and 0.19.8.1 shows:
```Diff
diff gettext-0.19.7.recipe gettext-0.19.8.1.recipe |tail -n +13
32c32
< 	devel:libgettextpo$secondaryArchSuffix = 0.5.3 compat >= 0
---
> 	devel:libgettextpo$secondaryArchSuffix = 0.5.4 compat >= 0
35c35
< 	devel:libintl$secondaryArchSuffix = 8.1.4 compat >= 8
---
> 	devel:libintl$secondaryArchSuffix = 8.1.5 compat >= 8
38c38
< 	lib:libgettextpo$secondaryArchSuffix = 0.5.3 compat >= 0
---
> 	lib:libgettextpo$secondaryArchSuffix = 0.5.4 compat >= 0
81c81
< 	lib:libintl${secondaryArchSuffix} = 8.1.4 compat >= 8
---
> 	lib:libintl${secondaryArchSuffix} = 8.1.5 compat >= 8
```
So I guess I should drop 0.19.8 and only keep 0.19.8.1. I'll drop 0.19.7 once I'm ready to drop the "WIP" label.

waddlesplash:
Any update?

fbrosson:
Sorry for this late reply. The recipe looks OK since my last update, Sun Jun 12 21:54:13, including the fixes in the **`lib*.la`** files. So I'm just dropping the **`(WIP)`** from the commit message header.

On some architectures I have some tests that are failing:
* gettext-tools/gnulib-tests/test-math.c
* gettext-tools/gnulib-tests/test-mbsrtowcs3.sh
* gettext-tools/gnulib-tests/test-mbsrtowcs4.sh

fbrosson:
BTW, I forgot to drop 0.19.7. I only dropped 0.19.8. Shall I drop 0.19.7?
That would leave us with only 0.18.1.1 and 0.19.8.1. Thanks!

waddlesplash:
Sure.

fbrosson:
That's it.

sys-devel/gettext/gettext-0.19.8.1.recipe
```diff
@@ -60,32 +60,51 @@ PROVIDES="
...
 	gettext${secondaryArchSuffix}_libintl == $portVersion base
 	lib:libiconv$secondaryArchSuffix
 	lib:libncurses$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	"
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	REQUIRES="$REQUIRES
+		lib:libgomp$secondaryArchSuffix
+		gcc$secondaryArchSuffix
```
diversys: Do you remember why this was introduced?

PR #598, opened 2016-05-17
tar: bump to 1.29, rework patch, mark x86 as tested, add TEST().

fbrosson:


app-arch/tar/tar-1.29.recipe
```diff
@@ -26,17 +26,33 @@ PROVIDES="
...
 	cmd:automake
 	cmd:autoreconf
+	cmd:find
 	cmd:gcc
 	cmd:ld
 	cmd:make
 	cmd:sed
 	"
+if [ "$recipeAction" = TEST ]; then
+	BUILD_PREREQUIRES="$BUILD_PREREQUIRES
```
korli: It makes me think that it would be cool to have TEST_PREREQUIRES in Haikuporter...

app-arch/tar/patches/tar-1.29.patchset
```diff
@@ -5,26 +5,23 @@ Subject: gcc2 patch
...
--                   __attribute__ ((__format__ (__printf__, 1, 2)));
-+;
+@@ -22,7 +22,7 @@
  
-   const char **ws_env;
-   const char *(*ws_getvar) (const char *, size_t, void *);
+ #include <stddef.h>
+ 
+-#if 2 < __GNUC__ + (7 <= __GNUC_MINOR__)
++#if 2 < __GNUC__ + (7 <= __GNUC_MINOR__) - __HAIKU__
```
korli: Could you explain a bit? __HAIKU__ isn't meant to be used that way.

fbrosson: This is a tiny hack. We need the **`#if`** expression to be false for gcc2 and true for gcc4 (or later).
* On x86_gcc2, `__GNU__`=2 and `__GNUC_MINOR__`=95 so without the patch we would be evaluating  (2 < 2 + 1). This is true, but we want it to be false. Hence the - __HAIKU__, which only has an impact on x86_gcc2.
* On x86 and x97_64, i.e. with gcc4, and even with the patch, we evaluate (2 < 4 - 1) or (2 < 4 + 1 - 1), which are both true. If we were using gcc3 on then it would not have been possible to use the hack.

I know, this is a strange hack, but it is not so dirty and it works like a charm when __HAIKU__ is defined ;-)

korli: What isn't clear is that "__attribute__ ((__format__ spec))" is supposed to be available for GCC 2.95, so are you saying it isn't on Haiku?

fbrosson: Sorry, I just realized that you were not asking for an explanation __about__ the hack but for an explanation __in__ the patch header.
BTW, your comment makes me think it would be even better to replace the **`(7 <= __GNUC_MINOR__)`** by **`(95 < __GNUC_MINOR__)`** and not use `__HAIKU__` at all.
This would probably be upstreamable because what I noticed is that the **`__attribute__ ((__format__ spec))`** does not work with our gcc 2.95. Well, we would need to double check why upstream thinks gcc >= 2.7 supports **`__attribute__`**.
Shall I change the patch in that sense? I mean replacing "**`7 <=`**" by "**`95 < `**".
May I let you dig that question, since you know gcc **much** better than me?
Thanks!

app-arch/tar/tar-1.29.recipe
fbrosson: I also though about having the exact same TEST_PREREQUIRES ;-)
There would be, however, a small question to decide: should haikuporter consider that any item in BUILD_PREREQUIRES should be repeated in TEST_PREREQUIRES if it is also needed during the TEST() ?
If we say that items should be repeated, then it might be worth using something like:
```
PREREQUIRES="
	cmd:find
	cmd:make
	cmd:sed
	"
BUILD_PREREQUIRES="$PREREQUIRES
	cmd:automake
	cmd:autoreconf
	cmd:gcc
	cmd:ld
	"
TEST_PREREQUIRES="$PREREQUIRES
	cmd:compress
	cmd:gzip
	cmd:bzip2
	"
```

But this is a big change. Maybe we should think about it for while before we change anything.
Hmm, I think we also need gcc for the tests, so maybe the best solution is to simply append items to BUILD_PREREQUIRES if "$recipeAction" = TEST, like I did for the recipe. I guess this would be the first recipe to have such a conditional BUILD_PREREQUIRES. I think it's a good idea to keep that variable clean from test prerequisites, even if it introduces a small change in the layout of recipes that have a TEST. The nice thing is that it does not break anything and that we can "migrate" smoothly to the new layout everytime a recipe needs an update.

app-arch/tar/patches/tar-1.29.patchset
fbrosson: I've just retried building tar 1.29 on x86_gcc2 without the first chunk of the patch, i.e. the chunk with the tiny hack. The result is:
```
  CC       wordsplit.o
In file included from /sources/tar-1.29/lib/wordsplit.c:40:
/sources/tar-1.29/lib/wordsplit.h:43: argument format specified for non-function `ws_error'
/sources/tar-1.29/lib/wordsplit.h:45: argument format specified for non-function `ws_debug'
Makefile:1287: recipe for target 'wordsplit.o' failed
```
So I reworked that chunk to make it work, but avoiding the use of **`__HAIKU__`**.
I don't know if gcc > 2.95 is the correct criteria, but it is sufficient to discriminate between the version of gcc2 we have in x86_gcc2 and the one we have in x86 and x86_64.

fbrosson:
> What isn't clear is that "**`__attribute__ ((__format__ spec))`**" is supposed to be available for GCC 2.95, so are you saying it isn't on Haiku?

Exact. Upstream thinks it is supported for gcc >= 2.7 but it is actually **not** supported by the gcc 2.95 we have in Haiku/x86_gcc2. I have therefore updated the patch to fix **`tar-1.29/lib/wordsplit.h`**.

fbrosson:
Hmm, maybe upstream wanted to compare against 2.95.7 (instead of 2.7) and forgot the 95. Anyway, the recipe works for us ;-)

app-arch/tar/tar-1.29.recipe
korli: Well, I'd rather have a default TEST(), and TEST_PREREQUIRES inheriting BUILD_PREREQUIRES.
Conditional generally doesn't make a recipe more readable, on the contrary.
I also prefer to make a change in mass, and double check it, instead of going through tens of PR :)

fbrosson: OK. Shall I drop the conditional definition and put **`cmd:compress`**, **`cmd:gzip`** and **`cmd:bzip2`** in BUILD_PREREQUIRES and drop the rest: **`cmd:lzip`**, **`cmd:lzma`** and **`cmd:xz`**.
Or is it OK to keep the the small condition on "$effectiveTargetArchitecture" != x86_gcc2 ?

fbrosson:
Cool, thanks (again) for the merge!

korli:
FYI I had a second look, in tar 1.28, we had a patch to delete this attribute __format__ only in wordsplit.h, wordsplit.c wasn't affected as it is now.

fbrosson:
I've fixed the patch to revert the effect on **`wordsplit.c`**: https://github.com/haikuports/haikuports/compare/master...fbrosson:tar
tar-1.29 builds fine with the new patch. Shall I merge it? Should I create a new PR or is it OK to add a reference to this one in the commit message?

fbrosson:
I've just created #627.

PR #600, opened 2016-05-18
live555: Fix devel package and policies

Barrett17:


fbrosson:
Grouping a few remarks and questions:
* You forgot to bump REVISION to 3.
* Are you sure you don't want to PROVIDE **`lib:live555$secondaryArchSuffix`** ?
* I think you need to remove **`$portVersion`** from  **`devel:live555$secondaryArchSuffix = $portVersion`**  in PROVIDE_devel or replace it with the x.y.z digits of the version of the lib, because 2016.01.29 is not usual for versions of libraries.

Barrett17:
live555 is a particular package...not a canonic lib.

- No need to bump revisions on a early recipe IMO
- Yes there's not any lib provided by the base package just a set of tools
- There's not any official version by the devs in fact we are basing our revs in on a 3rd party repo, and the date is OK for the purpose as it will be always incremental

fbrosson:
OK. Well, I'm not totaly convinced that it is OK to have **`devel:live555$secondaryArchSuffix = $portVersion`** in PROVIDES_devel but no **`lib:live555$secondaryArchSuffix = $portVersion`** in PROVIDES, unless you change the recipe to remove liblive555.so.
My understanding is that we should declare all items in such a way that **`haikuporter -S`** does not complain. But maybe using the _strict mode_ of haikuporter is only a recommendatiion, in which case I guess the recipe is OK.
(I hope I'm not being _too_ picky. Sorry if that's the case ;-)

Barrett17:
There's no live555.so at all. And the base package doesn't provide libs as said...

fbrosson:
OK, I think I got it. The only libs that are provided are all static libs and are in the _devel package. Am I right?

Barrett17:
Yes exactly :-)

korli:
Please bump revision anyway.

Barrett17:
DONE

PR #601, opened 2016-05-18
sqlite: add recipe for version 3.13.0.0, but w/o libsqlite3.a.

fbrosson:


korli:
I'd check if any other recipe requires devel:libsqlite3 but not lib:libsqlite3, thus meaning that they use the static lib.

fbrosson:
Excellent idea. If just found (using an automated and error-free method) that there is only one recipe which needs devel:libsqlite3 but no dev:libsqlite3: [dev-perl/alien_svn/alien_svn-1.8.11.0.recipe](https://github.com/haikuports/haikuports/blob/56a7c925628a95e4fed51239c6f160ecad4b0248/dev-perl/alien_svn/alien_svn-1.8.11.0.recipe)
I'll check if it still builds OK without libsqlite3.a.

fbrosson:
alien_svn-1.8.11.0.recipe does not need libsqlite3.a to build. It builds fine (on x86_gcc2) with the devel:libsqlite3 from sqlit3 3.13.0.0.
I had checked this a couple of hours ago, and wanted to make sure alien_svn could also build on x86 and x86_64... but these are not in ARCHITECTURES.

For the record, I tryed building it anyway on x86 but:
```
checking for apr_int64_t Python/C API format string... 
configure: error: failed to recognize APR_INT64_T_FMT on this platform
configuring SVN failed at inc/My/SVN/Builder.pm line 118.
```
Anyway, as we can see, adding **`--disable-static`** in squite-3.13.0.0 is OK.

fbrosson:
Thanks!

PR #603, opened 2016-05-19
exiftool: move from media-gfx to media-libs, drop old, add new.

fbrosson:
* Drop 10.13 and 10.14, which were in media-gfx/exiftool.
* Add 10.15 and 10.16 in media-libs/exiftool to follow Gentoo.
* Switch to **`ARCHITECTURES="any"`**.
* Use **`cmd:perl`** (in REQUIRES) instead of just **`perl`**.


korli:
This recipe shouldn't require perl, but cmd:perl.

fbrosson:
Fixed REQUIRES. Thanks.

PR #604, opened 2016-05-19
mc: add missing cmd:checkmk to BUILD_PREREQ and drop 2nd arch.

fbrosson:


fbrosson:
BTW, I did not bump the revision here because there are no changes in the resulting package. We're just dropping SECONDARY_ARCHITECTURES and adding a missing item in BUILD_PREREQUIRES.
If anyone wishes, I can, however, bump REVISION ;-)

PR #607, opened 2016-05-22
color-lines, new recipe

Begasus:


korli:
Naming is incorrect: should be color-lines/color_lines-0.6.recipe

fbrosson:
You also need to add a missing **`$secondaryArchSuffix`** in PROVIDES to have:
```
PROVIDES="
	color_lines$secondaryArchSuffix = $portVersion
	app:color_lines = $portVersion
	"
```
I don't know, however, if you also need to add the same suffix to **`app:color_lines`**.

Regarding the order of some lines, we generally put PATCHES right after CHECKSUM_SHA256 and before ARCHITECTURES. The usual layout would be:
```
CHECKSUM_SHA256="e20c87dfcc8f940c8af7784dad12e66fa1cd10d0d97766979fef00eb17c0b0b3"
PATCHES="lines-$portVersion.patch"

ARCHITECTURES="x86_gcc2 x86 ?x86_64"
```
Hmm, I'm quite sure it can build on x86_64. I'll try and let you know.
Ah, maybe it would be a good idea to rename the patch file to color_lines-0.6.patch (and change the PATCH line to **`PATCHES="color_lines-$portVersion.patch"`**

fbrosson:
@Begasus I've pushed (to my GitHub clone) an updated branch of color_lines. Here's the diff against master: https://github.com/haikuports/haikuports/compare/master...fbrosson:colorlines
If you want to copy and push it to the PR, here are the steps:

1) Not required, but recommended: refresh you master and update your remote master:
```
git checkout master
git pull real_origin master
git push
```
 2) Then create a temporary branch to grab my commit:
```
git checkout -b fbrosson-colorlines master
git pull https://github.com/fbrosson/haikuports.git colorlines
```
3) Check with **`git log`** that everything is OK, then rename your original color_lines branch as color_lines_old and then rename the new fbrosson-colorlines branch as color_lines and finally push it:
```
git branch -M color_lines color_lines_old
git branch -M fbrosson-colorlines color_lines
git push -f --set-upstream origin color_lines
```
4) If anything goes wrong, you can undo everything with:
```
git checkout color_lines_old
git branch -M color_lines fbrosson-colorlines
git branch -M color_lines_old color_lines
git branch -D fbrosson-colorlines
```

In case GitHub allows to attach another branch to the PR, then I might be able to update the PR with fewer steps. I don't know if this would work, and I guess this is not recommended, but if you want and if we have green lights for it (and for this time only), then I can try it.
BTW, I kept your signature in the commit header, so if the merge is done with the "Squash" button then the commit will appear as yours.

Begasus:
Thanks so far, I'll see if I can make some time during this week or in the
weekend to look things over.

2016-05-23 15:44 GMT+02:00 fbrosson <notifications@github.com>:

Begasus:
I'm guessing that something probably went wrong with pushing ... 2 commits in it now 
I used git push -f --set-upstream github color_lines last ...

Begasus:
reverted the changes, but updated with your changes, everything should be in place now

fbrosson:
Looks good to me.

Begasus:
PS, maybe it would be a good idea to put the order of other stuff like for instance PATCHES, ADDITIONAL_FILES ... also in the generic file for apps in the haikuporter folder, that way we know what to follow :)

fbrosson:
Good news: PATCHES and ADDITIONAL_FILES are already listed in:
* https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

BTW, there are 3 other pages related to the subject:
* https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes
* https://github.com/haikuports/haikuports/wiki/A-Gentle-Introduction-to-HaikuPorter-%231
* https://github.com/haikuports/haikuports/wiki/A-Gentle-Introduction-to-HaikuPorter-%232

Begasus:
Too late for this evening, looks like something went wrong when updating the recipe :/ will try to look at it somewhere this week ...

fbrosson:
No problem, there are no deadlines for us ;-)

Begasus:
Looks like this one should be ok :)

games-puzzle/color-lines/color_lines-0.6.recipe
```diff
@@ -0,0 +1,59 @@
...
+		GAMEDATADIR=$appsDir/ColorLines/
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/ColorLines
+	cp -r gfx/ icon/ sounds/ $appsDir/ColorLines
+	cp color-lines $appsDir/ColorLines/ColorLines
+
+	addAppDeskbarSymlink $appsDir/ColorLines/color-lines ColorLines
```
fbrosson: You should either change this to **`addAppDeskbarSymlink $appsDir/ColorLines/ColorLines`** or change the previous command to **`cp color-lines $appsDir/ColorLines/`**, otherwise the symlink will point to an inexistent runtime.

Regarding the TEST function, although having a white space between "TEST" and "()" is not incorrect, we generally write **`TEST()`** without any white space. But for this recipe, since the Makefile does not have a **`check`** target (nor any similar target), you will want to drop the whole TEST function. I tried replacing **`make check`** by **`color-lines --version`** and then by **`color-lines -h`** but none worked as expected because it just tries to launch the app.
Having TEST functions is a good thing, but if there is no way to test the program from a script then that's perfectly OK to have no TEST function at all.

When updating your commit , remember that you can use **`git commit --amend color_lines-0.6.recipe`** (and then **`git push -f`**). That's equivalent to a normal commit followed by a squash.

fbrosson:
I was wondering whether it would be better to put the runtime directly under $appsDir and the datafiles in a separate directory, **`$dataDir/ColorLines/`**.
I'm asking this because when I **`ls -l /boot/system/apps`** I see that most apps don't have any subdirectory in there. There are a few exceptions, though, like Vision and SMPlayer (and a few more, I guess).
Anyway, this is not very important and it does not make much difference.

fbrosson:
I've noticed that the project's Makefile does have an install target, so if you wish to simplify INSTALL, you can use this:
```Shell
BUILD()
{
	make PREFIX=$prefix BINDIR=$appsDir GAMEDATADIR=$dataDir/color-lines/
}

INSTALL()
{
	make PREFIX=$prefix BINDIR=$appsDir GAMEDATADIR=$dataDir/color-lines/ \
		install
	rm -rf $prefix/share
	addAppDeskbarSymlink $appsDir/color-lines ColorLines
}
```
It works fine, but it also uses a different GAMEDATADIR.

Begasus:
Looks fine to me, thanks!

Begasus:
Thanks for the --amend tip! I always pushed then did --amend and pushed again, saves a few steps ;)

games-puzzle/color-lines/color_lines-0.6.recipe
```diff
@@ -0,0 +1,52 @@
...
+	devel:libsdl_mixer$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+    make PREFIX=$prefix BINDIR=$appsDir GAMEDATADIR=$dataDir/color-lines/
```
fbrosson: Almost perfect!
You just forgot to use real tabs for the indentation inside BUILD. (Same remark for INSTALL.)

fbrosson:
You're welcome!
There's another nice option that you can combine with **`--amend`** to change the time stamp of the commit: **`--date`**.
I use it all the time, but from within a **`gcad`** script I "wrote". (gcad = short for git commit amend)
cat /boot/system/non-packaged/bin/gcad
```
#!/bin/sh
```

And if I don't want to change the time stamp, e.g. when I'm just updating the commit message, then I use another script: gca.
cat /boot/system/non-packaged/bin/gca
```
#!/bin/sh
exec git commit --amend "$@"
```

Begasus:
good catch for the ident(s) :)

fbrosson:
@Begasus: The recipe is OK for me. I've tested it on x86_gcc2, x86 and x86_64. So thanks for your PR!
To all: Shall I merge it myself?
If yes, should I use the squash feature (if that is available) on the web-based interface of GitHub?
BTW, would it be OK to change (very slightly) the commit message to «**`color-lines: new recipe.`**» in case we use the squash feature?

Begasus:
Nice to see the comments are kept in place, this could be usefull for future stuff ;)

fbrosson:
If you're talking about the hidden comments, yep, they are;-)
(I don't know, however, if one's comments are kept if it's author removes it with the X button.)
If you're talking about the striked text above, I used the double tilde trick: **`~~text you want to strike~~`**

PR #608, opened 2016-05-23
pciutils: add recipe for 3.5.1, enable shared lib, split data.

fbrosson:
* Add recipe for 3.5.1, also installing libpci.so and moving pci.ids.gz to an arch-independent pciutils_data sub-package.
* In 3.4.1 BUILD() was passing **`LIBDIR=$headersDir`** (to make) instead of **`INCDIR=$includeDir`**. This is fixed in 3.5.1.
* Move pci.ids.gz from /system/data/ to /system/data/pciutils/.
* Add required steps to make **`pkg-config --variable=idsdir libpci`** yeld the correct path to the directory where pci.ids.gz is.
* Edit **`man8/{ls,set}pci.8`** to replace **`$dataDir/`** by **`/system/data/`** and **`/usr/include/{linux,pci}/`** by **`/system/develop/headers/pci/`**.

**WIP:**
Shall we add the possibility to use update-pciids?
If not, shall we use pci.ids instead of pci.ids.gz?
Shall we get rid of all old recipes?
Shall we enable the DNS method to query pci.id.ucw.cz? (That's a big question with potential privacy issues.)


fbrosson:
I did not dare to drop 3.2.1, 3.2.1_git and 3.4.1. Should I remove some or all of these old recipes?

fbrosson:
It might be intersting to either:
* use ZLIB=no (instead of ZLIB=yes) which will result in **`pci.ids`** (instead of **`pci.ids.gz`**) being installed uncompressed, without any increase of the package size since hpkgs are compressed anyway);
* or install pci.ids.gz in **`/system/cache/pciutils/`** instead of **`/system/data/pciutils/`** and patch the sources to make them first look for **`pci.ids.gz`** in the writable directory **`/boot/system/cache/pciutils/`** (using **`finddir B_SYSTEM_CACHE_DIRECTORY`**, either computed at build time, or computed at runtime) and fall back to **`/system/data/pciutils/`** if the file cannot be found in the writable directory.

What do you think?

waddlesplash:
> I did not dare to drop 3.2.1, 3.2.1_git and 3.4.1. Should I remove some or all of these old recipes?

Drop all but 3.2.1; it looks like that's the "current version" on master.

> Shall we add the possibility to use update-pciids?
> If not, shall we use pci.ids instead of pci.ids.gz?
> Shall we get rid of all old recipes?
> Shall we enable the DNS method to query pci.id.ucw.cz? (That's a big question with potential privacy issues.)

No, yes, see above, no.

fbrosson:
OK, thanks. I've just made the changes to switch from **`pci.ids.gz`** to **`pci.ids`** as discussed.
Given that we now pas **`ZLIB=no`**, we could probably remove **`devel:libz$secondaryArchSuffix`** from BUILD_REQUIRES. I did not do so because I plan to propose someday a patch to enable support for update-pciids, using an alternate directory, **`/system/cache/pciutils/`**. Should I remove **`devel:libz$secondaryArchSuffix`** anyway, and only bring it back later, if/when I manage to do the suggested change?

waddlesplash:
Sure.

fbrosson:
Updated. Works fine, as expected.

PR #609, opened 2016-05-23
make: add recipe for 4.2, add TEST(), switch to mirror.

fbrosson:
Also _revert an upstream commit until a real fix is available_, just like Gentoo did with 4.2-r1.
See https://bugs.gentoo.org/583812
(This was blocking **`haikuporter --test make-4.2`**.)
Finally, remove unneeded call to libtoolize.


fbrosson:
I did not dare to drop 4.1. Should I?

waddlesplash:
Leave it for now.

fbrosson:
OK, thanks!

PR #613, opened 2016-05-26
libbluray: bump to 0.9.3, fix libbluray.pc and REQUIRES_devel.

fbrosson:
* Bump to 0.9.3 ~~with a big gcc2 patch~~, add TEST() with **`make check`**.
* Fix **`Libs.private`** paths in **`pkgconfig/libbluray.pc`**.
* Add **`devel:lib{fontconfig,freetype,xml2}`** to REQUIRES_devel.
* Drop explicit call to autoreconf as configure works without it.
* ~~Drop~~ **Keep** the static lib ~~as~~ **although** handbrake, mpv and vlc do not need it.

### DIFF
--- a/media-libs/libbluray/libbluray-0.9.3.recipe
+++ b/media-libs/libbluray/libbluray-0.9.3.recipe
@@ -1,18 +1,18 @@
 SUMMARY="Library for playback of Blu-ray movies"
 DESCRIPTION="This library is simply a tool for playback of Blu-ray movies."
 HOMEPAGE="https://www.videolan.org/developers/libbluray.html"
-COPYRIGHT="2010-2015 VideoLAN"
+COPYRIGHT="2010-2016 VideoLAN"
 LICENSE="GNU LGPL v2.1"
 REVISION="1"
 SOURCE_URI="https://download.videolan.org/pub/videolan/libbluray/$portVersion/libbluray-$portVersion.tar.bz2"
-CHECKSUM_SHA256="efc994f42d2bce6af2ce69d05ba89dbbd88bcec7aca065de094fb3a7880ce7ea"
+CHECKSUM_SHA256="a6366614ec45484b51fe94fcd1975b3b8716f90f038a33b24d59978de3863ce0"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
-SECONDARY_ARCHITECTURES="x86"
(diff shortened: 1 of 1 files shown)


korli:
Is gcc2 support really needed? It's just a pain to bump version with such a patch to update. Or do you intend to have it upstreamed?

fbrosson:
Of course, it is not needed since most users have x86_gcc2 hybrid. But it can be interesting for users of x86_gcc2 non hybrid to be able to run vlc, mpv and handbrake, as well as users of x86_gcc2 hybrid to avoid having to install lots of x86 secondary arch packages.
BTW, if I'm not mistaken, apps built (with bluray support) on x86_gcc2 primary arch would also run on BeOS (and Haiku/x86_gcc2 non hybrid, of course).
Now, regarding the patch, I think it is unlikely that the patch gets accepted by upstream.
Do you think upstream would accept it? What would be the roadmap?
Merging this PR and making binary hpkgs available and then submitting the patch to upstream, with some links and screen captures?
Or do we need to have the patch merged upstream _before_ we enable x86_gcc2?
If you prefer to not include the patch then I can remove it and only keep the version bump.

pulkomandy:
It is unlikely that apps built on Haiku would run on BeOS. We have too much API extensions for this to work well. And, not much people still care about BeOS, anyway.

For gcc2, our goal is to provide something similar to BeOS in terms of available libraries. For everything else, gcc5 can be used. The gcc2-only Haiku is not meant for production use. On that platform you have no recent web browser (even NetSurf is now gcc4/5 only), etc.

So, if the upstream is ok with accepting the patch, it is not too much work to have the recipe for gcc2 as a "bonus". But if they don't want it, just stick with gcc5.

fbrosson:
OK, thanks. I'll try to get in touch with upstream and see what I can do.

waddlesplash:
Any update? If you haven't contacted upstream yet, don't bother.

fbrosson:
I wanted to build VLC 0.8.6i with my patched libbluray before contacting upstream... but this version of VLC does not even try to use libbluray. So I just dropped my big fcc2 patch.

So, besides the version bump for libbluray, the main changes are in **`develop/lib/pkgconfig/libbluray.pc`** which I fixed, as well as the REQUIRES_devel which was missing **`devel:lib{fontconfig,freetype,xml2}`**.

PR #614, opened 2016-05-27
New recipe file for Sanity-0.6

Vidrep:
sanity-0.6.patchset no longer required since changes implemented in master branch.
New recipe file to reflect these changes.
Build has been tested and binary is working.

### DIFF
--- a/sanity-0.6.recipe
+++ b/sanity-0.6.recipe
@@ -1,0 +1,48 @@
+SUMMARY="A GUI frontend for SANE"
+DESCRIPTION="A graphical Haiku scanner application, using SANE.
+SanityTranslator allows applications to use scanner devices through the \
+translation kit."
+HOMEPAGE="http://philippe.houdoin.free.fr/phil/beos/sanity/index-en.html"
+COPYRIGHT="2001-2015 Philippe Houdoin"
+LICENSE="MIT"
+REVISION="3"
+srcGitRev="e5cdc0623eaca2baabc12e2de2c9827302a60834"
+SOURCE_URI="https://github.com/diversys/sanity/archive/$srcGitRev.tar.gz"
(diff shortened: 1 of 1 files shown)


Barrett17:
The file need to be placed in the right directory, under haiku-apps/sanity/

Barrett17:
To be more clear, you have to modify the original recipe.

fbrosson:
I noticed that it fails on x86. We need a **`#include <string.h>`** in **`SaneUtils.cpp`**, but this is not the only error.
@diversys , shall we wait for a fix for x86 before someone merges this PR?

Barrett17:
The recipe is marked untested in x86, so that's a known issue.

fbrosson:
Since it does not currently  build on x86, should'nt we say **`!x86`** instead of **`?x86`** ?
Or is it better to just drop the **`?x86`** ?
I'm sure @diversys will find what needs to be changed to make it build on x86 (and x86_64).
(I haven't investigated enough, but it think it can be fixed.)

waddlesplash:
> Since it does not currently build on x86, should'nt we say !x86 instead of ?x86 ?

Yes.

fbrosson:
I guess it might be worth bumping REVISION, since SOURCE_URI is being changed, although this is just meta-data and the resulting runtime is unchanged.

Barrett17:
@Vidrep as I said yesterday don't worry about people commenting, there's nothing bad at all with your work and we are just helping you ;-) You have to just edit the file in the tree instead to add a new one, except that the process is identical to what you've done with this pull request. Changing also ?x86 to !x86 isn't that difficult at all, and if you need help just let us know.

fbrosson:
Yep, we are just trying to help. (It took me a few months to learn git and feel comfortable with it.)
And you should not be affraid of pushing anything. The Haiku community will help you fix your commits before they get merged. BTW, as you may have noticed, there aren't many active contributors, so I'm sure we are all hoping you will enjoy contributing!

Barrett17:
So closing this one.

PR #616, opened 2016-05-27
Create youtube_dl-2016.05.21.2.recipe

Vidrep:
New recipe for youtube-dl


waddlesplash:
Please squash your commits.

jessicah:
@waddlesplash perhaps we should enable squash merge for HaikuPorts, so this can be done automatically?

waddlesplash:
There is squash merge on most PRs. I wonder why I don't see the option here?

fbrosson:
Hmm, maybe it was disabled, or maybe it's only available for PRs which only contain commits from users who already have write permission.

PR #619, opened 2016-05-29
scummvm 1.8.1 new recipe

Begasus:


games-engines/scummvm/scummvm-1.8.1.recipe
```diff
@@ -0,0 +1,96 @@
...
+	devel:libtheora$secondaryArchSuffix
+	devel:libvorbis$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+"
+
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize
```
fbrosson: It's safer to use **`cmd:libtoolize$secondaryArchSuffix`** (instead of just **`cmd:libtoolize`**). BTW, on some recipes you really need to add the **`$secondaryArchSuffix`** suffix for **`cmd:libtoolize`**, so, IMHO, it is better to use it for all recipes that have a SECONDARY_ARCHITECTURES.

Begasus:
Ok, I'll change it for libtoolize, haven't had any problems so far not using the $secondaryArchSuffix for that one (yet)

fbrosson:
@korli explained it to me some time ago, but my Haiku box is not running so I can't remember which lib (libintl or libtool_ltdl) could cause trouble if pulled from the primary arch instead of the secondary arch.

games-engines/scummvm/scummvm-1.8.1.recipe
```diff
@@ -0,0 +1,96 @@
...
+SOURCE_URI="http://scummvm.org/frs/scummvm/$portVersion/scummvm-$portVersion.tar.xz"
+CHECKSUM_SHA256="935f53d7e4c460fbc22c58fee5007d4858891e4961831f53f575a3594b7f612a"
+REVISION="1"
+LICENSE="GNU GPL v2"
+COPYRIGHT="2001-2014 ScummVM Team"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+ADDITIONAL_FILES="scummvm.rdef"
```
fbrosson: Not very important, but according to https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering ADDITIONAL_FILES should appear before ARCHITECTURES.
And COPYRIGHT, LICENSE and REVISION should appear between HOMEPAGE and SOURCE_URI.
But I think there are still many recipes whith the _old_ sorting.

Begasus:
Updated with the new style .. hope this one is ok :)

games-engines/scummvm/scummvm-1.8.1.recipe
```diff
@@ -0,0 +1,95 @@
...
+CHECKSUM_SHA256="935f53d7e4c460fbc22c58fee5007d4858891e4961831f53f575a3594b7f612a"
+ADDITIONAL_FILES="scummvm.rdef"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	scummvm$secondaryArchSuffix = $portVersion
+	app:scummvm$secondaryArchSuffix = $portVersion
+"
```
fbrosson: We like to add a tab before the closing double quotes, like this:
```Shell
PROVIDES="
	scummvm$secondaryArchSuffix = $portVersion
	app:scummvm$secondaryArchSuffix = $portVersion
	"
```

Of couse, the same applies to REQUIRES, BUILD_REQUIRES and BUILD_PREREQUIRES.

games-engines/scummvm/scummvm-1.8.1.recipe
```diff
@@ -0,0 +1,95 @@
...
+	devel:libpng$secondaryArchSuffix
+	devel:libSDL$secondaryArchSuffix
+	devel:libtheora$secondaryArchSuffix
+	devel:libvorbis$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+"
+
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
```
fbrosson: I don't know why so many recipes have **`cmd:git`** in their BUILD_PREREQUIRES. Perhaps this is recommended on recipes that grab their archive from a snapshot. But since this recipe is for a stable release, I would remove it if it builds OK without the cmd:git.
But maybe there is a real reason to keep it. Feel free to keep it (or remove it), as you wish.

games-engines/scummvm/scummvm-1.8.1.recipe
```diff
@@ -0,0 +1,95 @@
...
+
+INSTALL()
+{
+	make install
+	mv $appsDir/scummvm $appsDir/ScummVM
+	addResourcesToBinaries $portDir/additional-files/scummvm.rdef \
+		$appsDir/ScummVM
+
+	addAppDeskbarSymlink $appsDir/ScummVM ScummVM
+}
```
fbrosson: If **`grep "^check:" work-x86-1.8.1-1/source/scumm-1.8.1/Makefile`** yelds something then you can (and probably should) add this at the end of the recipe:
```Shell

TEST()
{
	make check
}
```

With the TEST() function you can launch the test (after it has been built) by calling **`haikuporter --test scummvm_x86-1.8.1`** on x86_gcc2 or **`haikuporter --test scummvm-1.8.1`** on x86 or x86_64.
(I don't know if you have the infinite loop bug from time to time, but if haikuporter loops with **`waiting for build package scummvm_x86-1.8.1-1 to be activated`** , then you just need to press Control-C and retry.)

Begasus:
Haven't used or encountered the TEST() function yet, what does it do exactly?

fbrosson:
The TEST() function tells haikuporter what command it should launch to run the tests. Many projects, if not most, have configure scripts that generate a Makefile with a "check" target which runs the tests on each directory.
You can try it with libpng16-1.6.22:
```Shell
cd haikuporter
git checkout master
git pull real_origin master
haikuporter -S libpng16-1.6.22
haikuporter --test libpng16-1.6.22
```
You will see that libpng16 1.6.22 passes all the tests ;-)

Begasus:
for libpng I get this:

~> hp --test libpng
----------------------------------------------------------------------
TESTING media-libs::libpng-1.5.26
----------------------------------------------------------------------
heap size:                      791
TOC size:                        41
package attributes size:        790
total size:                     871
heap size:                      854
TOC size:                        41
package attributes size:        939
total size:                     934
Testing ...
keeping chroot folder /boot/home/haikuports/media-libs/libpng/work-1.5.26 intact for inspection

for scummvm I get this:

~> hp --test scummvm_x86
----------------------------------------------------------------------
TESTING games-engines::scummvm_x86-1.8.1
----------------------------------------------------------------------
Cleaning up temporary directories ...
heap size:                     1018
TOC size:                        41
package attributes size:       1483
total size:                    1098
Testing ...
fatal: No names found, cannot describe anything.
make: *** No rule to make target 'check'.  Stop.
Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 2
keeping chroot folder /boot/home/haikuports/games-engines/scummvm/work-x86-1.8.1 intact for inspection

Not sure what to make out of it, if I run it from a Terminal with ScummVM --version I get this:
~/config/apps> ScummVM --version
Signature in rsrc doesn't match constructor arg. (application/x-SDL-executable, application/x-vnd.scummvm)
ScummVM 1.8.1 (May 21 2016 19:40:41)
Features compiled in: Vorbis FLAC MP3 TiMidity RGB zLib MPEG2 FluidSynth Theora AAC FreeType2 JPEG PNG 
...

Begasus:
PS, grep didn't show anything for "check" (also looked into the Makefile where I didn't find any "make check" ...

fbrosson:
If the Makefile does not have a line starting with "check:" then **`make check`** will fail; that's normal. But maybe the target is not called "check" but "tests", in which case you would replace "make check" by "make tests" in the TEST() function. Or maybe there isn't any test suite at all for scummvm.
If you find something that looks like the name of a target that launches a test suite, then you can try it and use it in the TEST() function. Otherwise you can either use **`scummvm --version`** (instead of **`make check`**) in the TEST() function as you suggested, or completely remove the TEST() function. Just chose what you think is best ;-)

Begasus:
For "scummvm --version" in the TEST() function I get this with hp --test scummvm_x87:
Testing ...
WARNING: FSNode::createReadStream: 'scummvm-1.8.1' is a directory!
Default configuration file missing, creating a new one
WARNING: FSNode::createWriteStream: 'scummvm-1.8.1' is a directory!
ScummVM 1.8.1 (May 29 2016 20:13:25)
Features compiled in: Vorbis FLAC MP3 TiMidity RGB zLib MPEG2 FluidSynth Theora AAC FreeType2 JPEG PNG

Begasus:
make test works, but running it in the Terminal gives me a fail ...
http://pastebin.com/ifaFA30f (works for a month)

Begasus:
./test/runner
Running 216 tests........................................................................................................................................................................................................................OK!

got it ;)

Begasus:
One question, wouldn't it be better to have the prefix set to $appsDir/Scummvm instead of using the several option for configure atm? I know it has come up before, but having an app like ScummVM (or others) sounds reasonable to me to have in their own folder ...

fbrosson:
Using a prefix other then the default one is not forbidden but I think it might be considered bad practice.
Anyway, you don't need to give a "private" prefix, since each package has it's isolated environment. Haiku has indeed a very special filesystem called packageFS. I think it rocks. If you haven't read about it yet, you can read this article: https://www.haiku-os.org/blog/zooey/2011-01-08_package_management_first_draft which explains the concepts. Warning: some paths have changed, e.g. /boot/common, if I'm not mistaken).

Back to scummvm, did you try using **`runConfigure ./configure`** instead of the plain **`./configure`** ? It takes care of all default values, so you can use:
```
CPPFLAGS=`freetype-config --cflags` runConfigure ./configure --disable-debug --enable-release
```

I've learned a lot by simple reading existing recipes. I wrote myself a tiny script called findrecipe (in /boot/system/non-packaged/bin/ IIRC):
```Shell
#!/bin/sh
exec find . -type d -name "work-*" ! -prune -o -type f -name "*.recipe" "$@"
```
It is really flexible. You can search, for instance, all recipes that call freetype-config but do not have a **`cmd:freetype_config`** in their BUILD_PREREQUIRES:
```Shell
findrecipe -exec grep -q freetype-config {} \; ! -exec grep -q cmd:freetype_config {} \; -print
```
You can also look for all recipes that need **`devel:libcurl`** but don't need **`lib:libcurl`** with:
```Shell
findrecipe -exec grep -q devel:libcurl {} \; ! -exec grep -q lib:libcurl {} \; -print
```
Reading existing recipes is really instructive ;-)

Begasus:
I already use a terminal command to find things in the recipes: 
find . -maxdepth 3 -and -name '*.recipe' | xargs grep -e *
checking runConfigure ...

fbrosson:
I also use **è-maxdepth `** from time to time, but I prefer to use **`-prune`** when looking for recipes, because I can cd anywhere. I'm not very fond of xargs on the command line, however ;-)

Begasus:
CPPFLAGS=`freetype-config --cflags` runConfigure --omit-dirs "libExecDir infoDir localStateDir includeDir sysconfDir sharedStateDir sbinDir oldIncludeDir" ./configure \ ... 
This one is giving me if I try with runConfigure:

Running ScummVM configure...
error: unrecognized option: --build=i586-pc-haiku

Guess it would be better to go with the one that I have now ...

fbrosson:
OK, I see. It looks like the configure script from scummvm is not as standard as those found on most projects. So I guess you are right ;-)

Begasus:
While checking the PR for icebreaker the isseu of the app-signature came up, so I checked the one for this PR and saw that the rdef file still has an old signature, is there an easy way to reset the versioning in the file (think I came by some sed structure in a recipe a while back) or would it be wiser to have a appname-$portVersion.rdef file for each  update?

fbrosson:
I use patterns in the rdef file and then process them in INSTALL. So if you replace:
```
	addResourcesToBinaries $portDir/additional-files/scummvm.rdef \
		$appsDir/ScummVM
```
by
```
	local MAJOR="`echo "$portVersion" | cut -d. -f1`"
	local MIDDLE="`echo "$portVersion" | cut -d. -f2`"
	local MINOR="`echo "$portVersion" | cut -d. -f3`"
	sed \
		-e "s|@MAJOR@|$MAJOR|" \
		-e "s|@MIDDLE@|$MIDDLE|" \
		-e "s|@MINOR@|$MINOR|" \
		$portDir/additional-files/scummvm.rdef > scummvm.rdef

	addResourcesToBinaries scummvm.rdef \
		$appsDir/ScummVM
```
and edit the beginning of the rdef file to make it look like the one in [bafx/additional-files/bafx.rdef](https://github.com/haikuports/haikuports/blob/master/haiku-apps/bafx/additional-files/bafx.rdef) then you won't need to update the rdef for each release.
You have to be careful, though, with recipes that have non-digits in their version.
But that's easy to handle, if you agree to drop the non-digit from, say, the minor digit:
```
	local MINOR="`echo "$portVersion" | cut -d. -f3 | tr -d '[:alpha:]'`"
```

Begasus:
thanks, I was in favor also for using sed (as I had seen in an example in the tree somewhere), saves me the hunt for that file ;)

Begasus:
@fbrosson ... updated with the changes you handed, looks ok now :)

fbrosson:
I haven't tested it yet, but it looks OK to me too ;-)
Ah, I noticed that several recipes have **`cmd:git`** in their BUILD_PREREQUIRES. I think it is not required but maybe I'm missing something. I usually comment out the **`cmd:git`** and try a fresh rebuild to see if it still builds OK without Git.
BTW, I wish we had a recipe for https://diffoscope.org/ with a plugin to understand/compare hpkg. That could simplify package comparisons. Arch Linux has a package for it: https://www.archlinux.org/packages/community/x86_64/diffoscope/. Gentoo does not.

pulkomandy:
package list a.hpkg > a.txt
    package list b.hpkg > b.txt
    diff a.txt b.txt # or use whatever "diff" tool here, ponpoko diff is a graphical one

fbrosson:
Thanks @pulkomandy. I use **`package list ...`** a lot. But sometimes I also need to compare the contents of text files (e.g. some **`lib*.la`** files), so I use **`package extract -C path/to/empty/or/clean/subirectory  a_package.hpkg develop/lib`**, especially when the package is very big, like the one for gettext. That's why diffoscope would be great for such cases.
Hmm, I guess it is possible to mount a package in a user-defined mount point. I promised myself I would search that in the Haiku manual but haven't done so yet!!!

Begasus:
@pulkomandy that would mean you'd still have the 'older' hpkg file around I presume? ps, ponpoko works in Haiku (not in Haiku atm)?
@fbrosson ... I'm only adding "cmd:git" if I see a mention of it when configure is running (don't think scummvm does this) ... checking later :)

Begasus:
Thanks :) PS ... not checking this evening ... football :)

Begasus:
Any news on this PR? Is it ok or does it require some changes?

waddlesplash:
As said above, the Git BUILD_REQUIRES should be removed.

Begasus:
Ah, ok, didn't know this one was the dealbreaker ;) removed git :)

PR #620, opened 2016-05-30
curl: bump to 7.49.1, use _BSD_SOURCE (required for make check).

fbrosson:
* Bump to 7.49.1 and call **`runConfigure`** with **`CFLAGS=_BSD_SOURCE`** and **`CXXFLAGS=-D_BSD_SOURCE`** otherwise **`make check`** fails to build.
* Drop **`$manDir/man3`** on 2nd arch devel packages.
* Move the manual page of **`curl-config(1)`** to the devel package.
* Use **`install`** (instead of mkdir and cp) to simplify **`INSTALL()`**.
* Keep $binDir for 2nd arch since x86_gcc2 fails many tests.
* No need to call libtoolize. Keep autoconf (needed on x86_64).
* Add version info in rdef file to have "Get Info" show it.

### DIFF
--- a/net-misc/curl/curl-7.49.1.recipe
+++ b/net-misc/curl/curl-7.49.1.recipe
@@ -7,11 +7,12 @@ based upload, proxies, cookies, user+password authentication (Basic, Digest, \
 NTLM, Negotiate, kerberos...), file transfer resume, proxy tunneling and a \
 busload of other useful tricks."
 HOMEPAGE="https://curl.haxx.se/"
-COPYRIGHT="1996-2016, Daniel Stenberg"
+COPYRIGHT="1996-2016 Daniel Stenberg"
 LICENSE="Curl"
 REVISION="1"
 SOURCE_URI="https://curl.haxx.se/download/curl-$portVersion.tar.bz2"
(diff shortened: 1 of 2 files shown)


net-misc/curl/curl-7.49.1.recipe
```diff
@@ -52,6 +48,7 @@ BUILD_REQUIRES="
 BUILD_PREREQUIRES="
 	cmd:autoconf
 	cmd:gcc$secondaryArchSuffix
+	cmd:install
```
korli: It comes as default (as part of corruption), no need to add it.

korli: Sorry meant part of coreutils.

fbrosson: no problem ;-)

fbrosson: For the record, here's what we get on Haiku x86_gcc2 hybrid hrev50332:
```
haikuporter --test curl_x86-7.49.1
TESTDONE: 775 tests out of 777 reported OK: 99%
TESTFAIL: These test cases failed: 116 1094
TESTDONE: 1026 tests were considered during 683 seconds.

haikuporter --test curl-7.49.1
TESTDONE: 680 tests out of 794 reported OK: 85%
...
TESTDONE: 1026 tests were considered during 738 seconds.
```

PR #621, opened 2016-05-31
unrar: bump to 5.4.2, move unrar.rdef to additional-files.

fbrosson:


fbrosson:
I've just added a tiny commit (about the short and long info in unrar.rdef).
If it's OK then I'll squash it with the first one.

PR #625, opened 2016-06-04
bafx: add rdef with version info, add SHA256, mark x86_64 ok.

fbrosson:


humdingerb:
Can I ask anyone that adds/touches rdefs of applications to add an app-signature, too. Like:
`resource app_signature "application/x-vnd.BAfx";`
That way, those apps can be found by QuickLaunch.

fbrosson:
OK. I did not dare to add anything other than app_version in rdefs, but if that's OK then I can, of course.
So shall I fix this one? If yes, should I also bump REVISION?

humdingerb:
Yes, please. :) Bumping the revision never hurts...

PR #329, opened 2016-06-07
OpenSSH global-writeable settings with "keep-old" flag

fbrosson:
I updated my Haiku install, and then tried to log in through sftp. It worked before, but not anymore. I was able to log in through SSH, but not with SFTP.

I located the problem, it is the keep-old flag of the /boot/system/settings/ssh/sshd_config file.

It had the following line:
Subsystem	sftp	/packages/openssh-6.6p1-1/.self/lib/openssh/sftp-server
but the correct path is:
Subsystem	sftp	/packages/openssh-7.1p1-1/.self/lib/openssh/sftp-server

We should somehow update this file after each openssh update.


fbrosson:
I've just converted this issue into a PR by attaching a commit that fixes the bug.

PR #630, opened 2016-06-09
libpng16: bump to 1.6.23.

fbrosson:
Also add back lib:libpng to PROVIDES to allow upgrades from 1.6.20-1.

### DIFF
--- a/media-libs/libpng/libpng16-1.6.23.recipe
+++ b/media-libs/libpng/libpng16-1.6.23.recipe
@@ -7,9 +7,9 @@ COPYRIGHT="1998-2016 Glenn Randers-Pehrson
 	1996-1997 Andreas Dilger
 	1995-1996 Guy Eric Schalnat, Group 42, Inc."
 LICENSE="LibPNG"
-REVISION="2"
+REVISION="1"
 SOURCE_URI="http://downloads.sf.net/libpng/libpng-$portVersion.tar.gz"
-CHECKSUM_SHA256="7944056621477eca2eba31cf1e99fa484721216a8af03c124426fccb37665df9"
+CHECKSUM_SHA256="74a9db9ddd66caa64be82bfcb3cbec4fc4394ceb1053cb7e94b2a0aa9b78eb88"
 SOURCE_DIR="libpng-$portVersion"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64 arm ?ppc"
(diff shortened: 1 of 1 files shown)


fbrosson:
For the record, this is what we get with **`haikuports --test`** for libpng-1.6.23 for all architecures, both primary (x86_gcc2 x86 x86_64) and secondady (x86_gcc2 x86):
```
Testsuite summary for libpng 1.6.23
# TOTAL: 32
# PASS:  32
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
```

Now let me explain a dependency issue I've hit:
Since https://github.com/haikuports/haikuports/commit/4df6b5d6cc4b40849aabac45f86bc8b5a6ea632e has removed **`lib:libpng`** from the PROVIDES in libpng16-1.6.20.recipe, it is not possible to upgrade from **`libpng16-1.6.20-1`** to any newer version of libpng16 unless you accept to remove many packages:
```
problem 1: package **`fontconfig-2.11.94-1`** requires **`lib:libpng>=16.18.0`**, but none of the providers can be installed
```
And if you uninstall **`fontconfig-2.11.94-1`** then you also have to uninstall many other packages:
* fontconfig_devel-2.11.94-1
* fontconfig-2.11.94-1
* libass_devel-0.13.0-1
* libass-0.13.0-1

Note that all these packages come from the binary HaikuPorts depot.

So I have added «**`lib:libpng$secondaryArchSuffix = 16.23.0 compat >= 16`**» to the PROVIDES and I noticed that, as expected, it solves the dependency issue when upgrading. BTW, **`devel:libpng`** was not removed from PROVIDES_devel in the commit mentioned above.

**Hence this question:** May I update the commit in this PR to have the **`lib:libpng`** in PROVIDES ?
**`libpng12-1.2.56.recipe`** has both **`lib:libpng`** and **`lib:libpng12`** in its PROVIDES, so I guess having both **`lib:libpng`** and **`lib:libpng16`** in **`libpng16-1.6.23.recipe`** would be OK, but I wan't to be sure everyone agrees. Thanks!

korli:
I guess you can lib:libpng back. Never understood the reason to remove it on libpng16, it can only be removed on a major version bump.

fbrosson:
Cool, thanks! I just added back the lib:libpng.
Regarding the support for animated PNGs, I did not check that the packages that require lib:libpng16 still build OK, but I guess all should. Since it's easy to drop the support for animated PNGs, I think it's safe to try it and see what happens ;-)

korli:
I'd drop the APNG patch. No recipe requires it.

fbrosson:
OK. Shell I remove it completely or just comment it out?

fbrosson:
I've just made the APNG support optional, but I can remove it completely or comment it out, if you prefer.

fbrosson:
I've just rephrased the commit message.

PR #632, opened 2016-06-10
harfbuzz: bump to 1.2.7, add TEST() with "make check". (WIP)

fbrosson:
* Bump to version 1.2.7, add **`TEST()`** with **`make check`**.
* Add patch for 3 tests that would otherwise fail or be skipped.
* Add **`cmd:{python,which}`** to BUILD_PREREQ, as **`make check`** needs them.
* Drop explicit call to autoreconf as it is no longer needed.

### DIFF
--- a/media-libs/harfbuzz/harfbuzz-1.2.7.recipe
+++ b/media-libs/harfbuzz/harfbuzz-1.2.7.recipe
@@ -1,33 +1,43 @@
 SUMMARY="An OpenType text shaping engine"
 DESCRIPTION="HarfBuzz is an OpenType text shaping engine. It is used to \
 layout complex text such as the Indic and Arabic alphabets."
-HOMEPAGE="http://www.freedesktop.org/wiki/Software/HarfBuzz/"
-COPYRIGHT="2010,2011,2012 Google, Inc.
+HOMEPAGE="https://www.freedesktop.org/wiki/Software/HarfBuzz/"
+COPYRIGHT="2004-2016 Behdad Esfahbod
+	2000-2005 Owen Taylor
+	2013, 2015-2016 Khaled Hosny
+	2010,2011,2012 Google, Inc.
 	2012 Mozilla Foundation
 	2011 Codethink Limited
 	2008,2010 Nokia Corporation and/or its subsidiary(-ies)
 	2009 Keith Stribley
 	2009 Martin Hosken and SIL International
 	2007 Chris Wilson
-	2006 Behdad Esfahbod
 	2005 David Turner
 	2004,2007,2008,2009,2010 Red Hat, Inc.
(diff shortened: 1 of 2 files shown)


waddlesplash:
Looks good to me. Merge?

fbrosson:
Some tests were failling, but I figured out how to fix them ;-)
Now all tests pass, except one, **`src/check-libstdc++.sh`**, which yelds: **`check-libstdc++.sh: 'ldd' not found; skipping test`**
I'd like to patch that script to make it use **`objdump -x`** and **`grep NEEDED`**. Would it be OK to merge a recipe with such a patch?

Regarding the REQUIRES_devel, I added **`devel:libglib_2.0`** and **`devel:libgraphite2`** to it, based on the **`Requires.private`** of harfbuzz.pc: **`glib-2.0 >= 2.16 graphite2`**. Am I right for the REQUIRES_devel about that? Thanks!

waddlesplash:
>  I'd like to patch that script to make it use objdump -x and grep NEEDED. Would it be OK to merge a recipe with such a patch?

Eh, don't bother.

> Am I right for the REQUIRES_devel about that? Thanks!

That *sounds* correct, I suppose.

fbrosson:
OK, thanks for your reply.
BTW, would it be worth adding to REQUIRES_devel all the **`devel:lib*`** we have in BUILD_REQUIRES?


Switching back to x86_gcc2, when I test for x86 secondary arch, **`src/test-ot-tag`** fails and sends the following line to the log:
```
runtime_loader: Cannot open file libharfbuzz.so.0: No such file or directory
```
I noticed that launching **`src/.libs/test-ot-tag`** from the chrooted environment yelds the same error, but I can't find how to change the wrapper script, src/test-ot-tag, to fix this error.
Shall I ignore this error?
All tests are OK on x86 and x86_64 (with the exception of the skipped test), so I guess it might be OK to merge (once I remove the "WIP" tag, of course).
Shall I try to build mupdf 1.7 (and/or 1.9a) against harfbuzz 1.2.7 before I merge this PR?

korli:
"libraries from Requires.private will only be included when static linking."
Thus they shouldn't be added to REQUIRES_devel.

fbrosson:
OK, thanks. I've just fixed REQUIRES_devel.

waddlesplash:
@fbrosson btw: most HaikuPorts developers also hang out in the #haiku channel on Freenode, if you'd care to join us.

fbrosson:
Yep, I know. But I read, from time to time, the logs that echelog.com creates with what goes to the #haiku channel. I find it more confortable, since I'm always in read-only mode ;-)

waddlesplash:
Ah yes, but what's the fun in that :)

PR #633, opened 2016-06-11
icebreaker 1.9.7 new recipe

Begasus:
The app is working very nice, one this is still strange, if I copy the file into  /config/packages the score file is not installed, if I copy the file to /system/packages the score file is installed ...


Begasus:
~/destdir> tree 
.
├── apps
│   └── IceBreaker
├── data
│   ├── deskbar
│   │   └── menu
│   │       └── Applications
│   │           └── IceBreaker -> ../../../../apps/IceBreaker
│   └── icebreaker
│       ├── atom.bmp
│       ├── bgeveryone.wav
│       ├── bgvvint.wav
│       ├── billg.bmp
│       ├── billg.ibt
│       ├── blocks.ibt
│       ├── bwop.wav
│       ├── cheese.ibt
│       ├── cow.bmp
│       ├── cows.ibt
│       ├── crash.wav
│       ├── drum.wav
│       ├── ewww.wav
│       ├── icebreaker_48.bmp
│       ├── jz-like.ibt
│       ├── kitty.bmp
│       ├── kitty.ibt
│       ├── linux.ibt
│       ├── moo.wav
│       ├── mouse-r.bmp
│       ├── mouse.bmp
│       ├── nbwoanp.wav
│       ├── ouch.wav
│       ├── penguin.bmp
│       ├── penguinicon_32.bmp
│       ├── phase.wav
│       ├── pop.wav
│       ├── scrunch.wav
│       ├── splat.wav
│       ├── squeak.wav
│       ├── star.bmp
│       ├── stars.ibt
│       ├── tap.wav
│       ├── turtle.bmp
│       ├── turtles.ibt
│       └── woodblock.bmp
├── documentation
│   └── man
│       └── man6
│           └── icebreaker.6
└── settings
    └── icebreaker.scores

10 directories, 40 files

Begasus:
~/destdir> tree 
.
├── apps
│   └── IceBreaker
├── data
│   ├── deskbar
│   │   └── menu
│   │       └── Applications
│   │           └── IceBreaker -> ../../../../apps/IceBreaker
│   └── icebreaker
│       ├── atom.bmp
│       ├── bgeveryone.wav
│       ├── bgvvint.wav
│       ├── billg.bmp
│       ├── billg.ibt
│       ├── blocks.ibt
│       ├── bwop.wav
│       ├── cheese.ibt
│       ├── cow.bmp
│       ├── cows.ibt
│       ├── crash.wav
│       ├── drum.wav
│       ├── ewww.wav
│       ├── icebreaker_48.bmp
│       ├── jz-like.ibt
│       ├── kitty.bmp
│       ├── kitty.ibt
│       ├── linux.ibt
│       ├── moo.wav
│       ├── mouse-r.bmp
│       ├── mouse.bmp
│       ├── nbwoanp.wav
│       ├── ouch.wav
│       ├── penguin.bmp
│       ├── penguinicon_32.bmp
│       ├── phase.wav
│       ├── pop.wav
│       ├── scrunch.wav
│       ├── splat.wav
│       ├── squeak.wav
│       ├── star.bmp
│       ├── stars.ibt
│       ├── tap.wav
│       ├── turtle.bmp
│       ├── turtles.ibt
│       └── woodblock.bmp
├── documentation
│   └── man
│       └── man6
│           └── icebreaker.6
└── settings
    └── icebreaker.scores

10 directories, 40 files
 (listing of the package)

games-puzzle/icebreaker/icebreaker-1.9.7.recipe
```diff
@@ -0,0 +1,49 @@
...
+		highscoredir=$settingsDir mandir=$manDir
+}
+
+INSTALL()
+{
+	make prefix=$prefix bindir=$appsDir datadir=$dataDir \
+		highscoredir=$settingsDir mandir=$manDir install
+	mv $appsDir/icebreaker $appsDir/IceBreaker
+
+	addAppDeskbarSymlink $appsDir/IceBreaker IceBreaker
```
fbrosson: Since **`/boot/system/apps`** is in the default PATH, I think it would be slightly better to keep $appsDir/icebreaker lowercase and only use IceBreaker in the AppDeskbar.
It would also allow to drop the **`mv $appsDir/icebreaker $appsDir/IceBreaker`**.
You would just need to change the next line to **`addAppDeskbarSymlink $appsDir/icebreaker IceBreaker`**.
But that's just my personal taste, and I don't even know what the best practices are for these kinds of questions. My rationale is that commands we type on a terminal are always lowercase. Yep, I know this is not a unix command ;-)
Anyway, please feel free to ignore this remark. I won't be cry if you do ;-)
To all: do not hesitate to tell what your prefer between $appsDir/icebraker and $appsDir/IceBraker, keeping in mind that in the Applications menu you'll see it as IceBreaker anyway.

fbrosson:
I haven't tried the recipe yet, but the **`man6`** directory should go under **`data/documentation/man/`** (instead of **`data/documentation/`**. Maybe this can be fixed by passing **`mandir=$manDir/man`** (instead of **`mandir=$manDir`**).

fbrosson:
> if I copy the file to /system/packages the score file is installed ...

You can also install one or more packages with **`pkgman install ...`** instead of copying them to /system/packages/.
I think there are no differences except when you upgrade: if you use pkgman then it tells you that it is going to uninstall the previous version. Whereas if you use cp then you get a popup window.
There is one case, however, for which I use cp: if I want to replace a package by another one with the exact same name, version and revision. Using pkgman in such cases produces an error (saying some directory already exists) and fails to install the updated package. With **`cp`** the updated package only becomes available, however, after a reboot. So sometimes I just re-install an older version and then re-install the new one again. This saves a reboot since pkgman tells the packagefs daemon what to do.

Begasus:
The man is in the correct directory (only cp'ng the 'tree' here didn't put it in the correct order) :)
When installing it with HaikuDepot it is installed correctly (just the same as dropping/cp'ng it to system/packages) not sure why it doesn't install the score file when dropping it into config/packages ...

games-puzzle/icebreaker/icebreaker-1.9.7.recipe
Begasus: Would be nice if there was a mutual understanding in this, for the moment ScummVM also resides as "ScummVM" in the $appsDir folder (instead of "scummvm") ...

Begasus: PS ... having icebreaker vs IceBreaker in the $appsDir doesn't change much, you can't launch if from a Terminal from another location ...

Begasus:
PS, also noticed, the man file is installed when dropping (or installing with HaikuDepot), when dropping it in config/packages it isn't ... maybe someone could check this?

fbrosson:
Funny thing that **`tree`** gave you:
```
├── documentation
│   └── man
│   └── man6
```
instead of
```
├── documentation
│   └── man
│       └── man6
```
Regarding the other issue you mentioned, maybe packagefs is just picky with the contents of the package.

Begasus:
with tree it does give me the correct ident, just copying it here somehow didn't catch up with that ... (qupzilla)*

Begasus:
├── documentation
│   └── man
│       └── man6
│           └── icebreaker.6
└── settings
    └── icebreaker.scores

10 directories, 40 files

Begasus:
hmm same with Web+ ... ident doesn't show correctly ...

Begasus:
"Regarding the other issue you mentioned, maybe packagefs is just picky with the contents of the package."
Should it matter if it's installed in system/packages or config/packages ... in either way it 'should' install the files correctly ...

Begasus:
├── documentation
│   └── man
│       └── man6
│           └── icebreaker.6
├── icebreaker
└── settings
    └── icebreaker.scores

humdingerb:
I think, the issue is that in haikuporter's chroot, when executing the BUILD() and INSTALL(), $settingsDir is always in the /system hierarchy. That pretty much "hard-codes" the package to use the system's settings folder.   
Icebreaker should always use the B_USER_SETTINGS_DIRECTORY and the recipe a post-install script that copies the highscore file (which is a 0 byte file, so a "touch icebreaker.scores" in ~/config/settings should suffice). Alternatively, and maybe more elegant: patch icebreaker to create  icebreaker.scores" in ~/config/settings if it doesn't exist.

It'd be nice to have an icon and app-signature added with addResourcesToBinaries.

Personally, I'm a fan of having GUI apps bein self-contained in a folder under $appsDir. No spreading into various subfolders. $dataDir looks quite messy in my system already... I'd even drop the man file and go with a proper html in the app's folder...

Begasus:
http://pastebin.com/aYPwU0MY

fbrosson:
@humdingerb: I see more $appsDir as an alternate $binDir dedicated to GUI apps. As a consequence, if some command or app needs data files, IMHO they should go to a subdirectory in $dataDir. But that's just my point of view and I don't want to push into one direction. I'll adhere to whatever concensus emerges. But for the moment the only apps which have their data in $appsDir are Vision and SMPlayer (and maybe a few others). The vast majority do not have data files under $appsDir (or do not have any data files at all.)

games-puzzle/icebreaker/icebreaker-1.9.7.recipe
Begasus: "Since /boot/system/apps is in the default PATH" ok, this one I didn't know ;) I always installed the packages in config/packages (to try to keep the system apps dir not to big)
having "icebreaker" instead of "IceBreaker" in there looks a bit strang amongst the other apps there with Capitols ;)

fbrosson: Oops, my mistake. You are perfectly right. Believe me or not, I had not noticed that all the apps in **`/boot/system/apps/`** have capital letters!.

This means it is perfectly fine to keep the *`mv $appsDir/icebreaker $appsDir/IceBreaker`**.

BTW, when you symlink something with **`addAppDeskbarSymlink`** you can ommit the second parameter if the name of the symlink is the same as that of the target. So, if you want, you can replace:
```
addAppDeskbarSymlink $appsDir/IceBreaker IceBreaker
```
by
```
addAppDeskbarSymlink $appsDir/IceBreaker
```
(I discovered that once when I had a look at a file called **`wrapper-script`** that haikuporter puts in the work-* subdirectory, **`work-1.9.7-1`** when you build icebreaker-1.9.7.  BTW, I browse this **`wrapper-script`** file very often.)

Begasus:
added app-signature :)

humdingerb:
@Begasus , I have thrown together an icon that you may want to add to the rdef at https://gist.github.com/humdingerb/02ce782437b570e92c3bd15e55991e50
I simply placed zuMi's penguin icon into an ice cube.

Should we add the IOM file as "additional file" so people can improve it?
It's here: https://dl.dropboxusercontent.com/u/21023348/Icebreaker.iom

Begasus:
@humdingerb thanks! I've added and it looks nice :)

Begasus:
If you are ok with it I can add it to the PR (the IOM file)

games-puzzle/icebreaker/additional-files/icebreaker-1.9.7.rdef
```diff
@@ -0,0 +1,74 @@
...
+resource app_version {
+       major  = 1,
+       middle = 9,
+       minor  = 7,
+
+       variety = B_APPV_FINAL,
+       internal = 0,
+
+       short_info = "IceBreaker",
+       long_info = "An addictive action-puzzle game"
```
fbrosson: @Begasus, I've just noticed the indentation in the app_version block block is made of white spaces. I guess you will want to use tabs instead.

Begasus:
Fixed the ident, thanks @fbrosson :) (ps I copied the text over from scummvm's rdef, so that probably suffers from the same isseu)

humdingerb:
@Begasus , sure, add the IOM. The icon could be improved by making the ice block more transparent and showing Tuxi's frozen flippers...

Begasus:
@humdingerb, added it to the "additional files" folder ...

waddlesplash:
> @humdingerb: I see more $appsDir as an alternate $binDir dedicated to GUI apps.

Not exactly (see below).

> As a consequence, if some command or app needs data files, IMHO they should go to a subdirectory in $dataDir.

No -- the correct way to do things is that if something lives in $appsDir, it lives *exclusively* in $appsDir, and is in no other part of the system. By contrast, apps like VLC 2.0+ and the like which want users to be able to start VLC from the command line with flags and the like should not be in $appsDir at all, but should be spread around bin, lib, etc.

humdingerb:
@waddlesplash: You're quoting fbrosson...
That said, I feel the same about GUI apps belonging into $appsDir. Though if it's more than a single executable, I prefer its files all under $appsDir/appname.

> No -- the correct way to do things is that if something lives in $appsDir, it lives exclusively in $appsDir, and is in no other part of the system. By contrast, apps like VLC 2.0+ and the like which want users to be able to start VLC from the command line with flags and the like should not be in $appsDir at all, but should be spread around bin, lib, etc.

Is this "the correct way" because you say so, or was there a discussion with that conclusion?
There's a thread on the haikuports mailinglist that's a more proper venue to discuss this. Though nobody seems interested. Which is a shame because we have control over our OS and its guidelines, but we don't use that power to create a consistent environment. :(

fbrosson:
@waddlesplash: OK, thanks for the explanation about the $appsDir vs $binDir issue. I got it now (and I buy it ;-)
@humdingerb: Sorry for having triggered an implicit quotting ;-) Anyway, I think that anyone who followed this discussion has understood that **I** was the one who had a wrong understanding of the $appsDir vs $binDir issue. Now I have joined what seems to be the consensus :wink:

PR #635, opened 2016-06-12
vlc: use additional-files and fix duplicate vlc.rdef in patch.

fbrosson:
* Remove the duplicate **`vlc.rdef`** that was embedded in the patchset for vlc-0.8.6i and, instead, make it use the same **`vlc.rdef`** that vlc-2.2.1 already uses and which is in **`additional-files`**.
* Tweak **`vlc.rdef`** to make it usable with any version of vlc.
* Add **`TEST()`** with **`make check`** (to both 0.8.6i and 2.2.1).
* Fix the recipe of VLC 2.2.1 to make it build on x86 secondary arch.
* ~~Build vlc-0.8.6i without libdvbpsi because the last upgrade of libdvbpsi (from 1.2.0 to 1.3.0) has broken vlc-0.8.6i.~~
* ~~TODO **before** merge: Make VLC 0.8.6i use libdvbpsi 1.2.0 instead of 1.3.0 because I broke our vlc-0.8.6i.recipe when I upgraded libdvbpsi from 1.2.0 to 1.3.0 a few months ago. Shame on me!~~
* Force VLC 2.2.1 to use **`lib:libswscale`** from ffmpeg2. BTW, it seems this version of VLC cannot compile against ffmpeg >= 3.
* Comment out the **`lib:libdevice`** in REQUIRES because the haiku hpkg is missing a PROVIDES for it.
* ~~Force VLC 0.8.6i to use libdvbpsi 1.2.0 instead of 1.3.0, until we find out what is preventing VLC from using libdvbpsi 1.3.0. I had suspected that it was the upgrade to 1.3.0, but it turns out that it does not currently build with 1.2.0 either. I suspect that there is something else that is breaking VLC with libdvbpsi because I had successfully built VLC 0.8.6i on March 10 with hrev50139, that is, 4 days after I had bumped libdvbpsi to 1.3.0.~~
* VLC 0.8.6i was previously using libdvbpsi 1.2.0 but the compilation now fails for some reason (unknown to me). The compilation also fails when building against libdvbpsi 1.3.0, so drop support for it in VLC 0.8.6i. VLC 2.2.1 on x86 secondary arch builds fine with any of the mentioned releases of libdvbpsi, though.


pulkomandy:
I would appreciate less aggressive removing of old recipes. When I did the vlc 0.8.6 recipe, I took care of either porting matching versions of all the dependencies, or updating the VLC code to match current versions, when possible. If people just remove the old versions when adding new ones, my efforts are a bit useless.

As long as we don't have a complete native port of VLC 2.x or later, we should try to keep 0.8.6 working correctly, and at least as well as the BeOS builds did. This means renaming libdvbpsi old version to libdvbpsi02 or something similar so it can be installed next to the newer version.

fbrosson:
I'm not removing vlc-0.8.6 at all ;-)
I just commented out the dependency on libdvbpsi cause I noticed that vlc-0.8.6 only supports libdvbpsi 1.2.0, but not libdvbpsi 1.3.0. When I had upgraded libdvbpsi a few months ago, I made a mistake. I plan to add libdvbpsi 1.2.0 back as libdvbpsi12, and then uncomment the dependency on it in vlc-0.8.6i.
So don't worry, I know we need both versions.
BTW, I've even been trying to build VLC 2.2.2, 2.2.3 and 2.2.4.

pulkomandy:
Sorry, my mistake, libdvbpsi12 is the right name.
For later versions, unfortunately getting them to build is not enough. IIRC, we can use the Qt GUI (it's not native, but building a new native GUI needs a lot of work), but we still need audio and video output backends with some specific work.

fbrosson:
OK, thanks.

waddlesplash:
Is this still WIP?

fbrosson:
> Is this still WIP?

@waddlesplash: Yes it is, because I'm not sure my changes to 0.8.6i are all OK.
Moreover, something is preventing VLC from compiling when it is configured with support for libdvbpsi. But there are no real changes in the recipe. So merging this PR won't make things worse.
I plan to grab a snapshot of hrev50139 and try building VLC 0.8.6i with libdvbpsi 1.2.0 or 1.3.0, but that will take _some_ time.
In case you would agree to merge this PR before I find out what is breaking VLC here, then please let me know so that I can drop the "(WIP)" label and, if you wish, remove some commented-out lines I have inserted in REQUIRES_devel and BUILD_PREREQUIRES. Thanks!

fbrosson:
> * Force VLC 2.2.1 to use **`lib:libswscale`** from ffmpeg2. BTW, it seems this version of VLC cannot compile against ffmpeg >= 3.

I have finally managed to build VLC 2.2.1 (against ffmpeg2-2.8.7, libbluray-0.9.3 and libdvbpsi-1.3.0) on x86 primary arch. Will try on x86 secondary arch.

waddlesplash:
> I have finally managed to build VLC 2.2.1 (against ffmpeg2-2.8.7, libbluray-0.9.3 and libdvbpsi-1.3.0) on x86 primary arch. Will try on x86 secondary arch.

Any update? Can this be merged?

fbrosson:
I did not manage to build VLC 0.8.6i with libdvbpsi, no matter which version of libdvbpsi I tried, so I commented out the lines with libdvbpsi. VLC 0.8.6i only builds and runs fine without libdvbpsi. I don't know what is causing the build to fail if building with libdvbpsi. BTW, I don't know why **`lib:libdevice`** is no longer declared as provided by haiku.

I fixed **`vlc-2.2.1.recipe`** by adding a few missing **`$secondaryArchSuffix`**, so VLC 2.2.1 builds and runs on x86 secondary arch.
So yes, I would say that this PR can be merged. I'm not merging it myself in case someone would like to troubbleshoot the libdvbpsi issue in VLC 0.8.6i.

Should I remove (completely) the lines mentioning **`{lib,devel}:libdvbpsi`** in **`vlc-0.8.6i.recipe`** ?

waddlesplash:
No, leaving the comments is fine.

fbrosson:
OK, thanks!

diversys:
Is there a point in $secondaryArchSuffix for vlc 0.86? IIRC I removed it at some point.

fbrosson:
If you want vlc 0.8.6i to build on x86_gcc2 (primary) and x86 (primary **and** secondary) arch then the **`$secondaryArchSuffix`** I've added **_are all_** required. (The recipe was not correct for x86 secondary arch and the best way to fix it was to add the missing **`$secondaryArchSuffix`**).

If you want vlc 0.8.6i to build on x86_gcc2 (primary) and x86 primary (but **_not_** secondary), then you can drop all **`$secondaryArchSuffix`** **_if_** you **_also_** drop the whole **`SECONDARY_ARCHITECTURES="x86"`** line.

If you want vlc 0.8.6i to build on x86 primary **_and_** secondary but **_not_** on x86_gcc2 primary, then you should replace «**`runConfigure ./configure …`**» by «**`runConfigure --omit-dirs binDir ./configure --bindir=$prefix/bin …`**» **_and_**  remove **_only_** the **`$secondaryArchSuffix`** that appear in **`cmd:`** lines in **`PROVIDES{,_devel}`** otherwise the recipe won't be valid for x86 secondary arch. Of course, you should also change ARCHITECTURES to **`ARCHITECTURES="!x86_gcc2 x86"`**.

Obviously, given that VLC 0.8.6i is broken since you upgraded libebml to 1.3.4 (in https://github.com/haikuports/haikuports/commit/dcc0fda83ae91837c9c2565233e544b485afa24d), none of the actions described above will fix #740 or #741.

What I can tell is that right after this PR got merged the recipe for VLC 0.8.6i **did build fine** against the version of libebml that was in the tree at that time, 1.3.3. The only bad news with this PR is that I have had to remove the dependency on libdvbpsi (for VLC 0.8.6i only) because something unrelated to VLC made building VLC 0.8.6i impossible with libdvbpsi. But this was a minor issue since VLC 2.2.1 builds OK with libdvbpsi and can be built for x86 secondary arch on x86_gcc2.

PR #636, opened 2016-06-13
libdvbpsi: exhume 1.2.0, fix 1.3.0 versioning, drop static lib.

fbrosson:


fbrosson:
I did not manage to rename the lib to libdvbpsi12 so I had to add a CONFLICTS_devel.
Would it be OK to remove the CONFLICTS_devel and:
* rename **`develop/headers/dvbpsi`** to **`develop/headers/dvbpsi12`**
* rename **`lib/livdvbpsi.*`** to **`lib/livdvbpsi12.*`**
* rename **`lib/pkgconfig/libdvbpsi.pc`** to **`lib/pkgconfig/libdvbpsi12.pc`**
* update the paths and filenames in **`lib/libdvbpsi12.la`** and **`lib/pkgconfig/libdvbpsi12.pc`**

Or is it worthless?
Is there a standard way of doing this? I tried to see how **`libpng{12,16}`** handle this but could not figure out how it does.
Also, I tried to configure using **`--disable-static`** (or **`--enable-static=no`**) but none worked, so I just kept it. Would it be OK, though, to simply remove **`libdvbpsi.a`** ?

pulkomandy:
No, it is expected that we get conflicts on the devel package in this case.
In libpng case, the library is already renamed in each version so there are not too much problems.

In libdvbpsi, both versions of the lib have the same name, so you can have only one version of the dev package installed at the same time (this is safer, because we could also run into conflicts with the .h files, etc).

As long as the non-devel packages can be installed side by side, there is no problems. The devel package for the old version of the lib will only be used with haikuporter, and in that case there shouldn't be too much conflicts because the chroot used for building only has a small set of packages.

If you rename all the libs and .pc files, all projects using the lib would have to be modified. Since this is mostly dealing with legacy code (old version of VLC, etc), it would be a lot of work that no one would want to merge upstream.

fbrosson:
OK, thanks a lot for your detailed reply.
I noticed I had made a second mistake! I had forgotten to update the versioning of **`{lib,devel}:libdvbpsi`** in the **`PROVIDES{,_devel}`** of libdvbpsi-1.3.0, so I'm fixing it in this PR.

PR #666, opened 2016-06-23
games-arcade/alex4: create recipe for Alex the Allegator 4. (WIP)

fbrosson:
**TODO:** Check if it would be OK to also import the patches in:
http://data.gpo.zugaina.org/gamerlay/games-arcade/alex4/files/
and sync with alex4-1.1 in http://gpo.zugaina.org/games-arcade/alex4

Maybe we should use a (modified) rdef instead of the **`alex4_BeOS_src/alex4.rsrc`** in alex4_beos_src.zip, because when Alex4 is launched from command line we see this warning:
```
Signature in rsrc doesn't match constructor arg. (application/x-vnd.Allegro-Alex4, application/x-vnd.freelunchdesign-alex4)
```

### DIFF
--- a/games-arcade/alex4/alex4-1.0.recipe
+++ b/games-arcade/alex4/alex4-1.0.recipe
@@ -1,0 +1,67 @@
+SUMMARY="A small jump'n'run platform game"
+DESCRIPTION="Alex the Allegator 4 is a true jump'n'run, just like the old \
+classic ones. Guide Alex through the jungle in order to save his girlfriend \
+Lola from evil humans who want to make a bag of her."
+HOMEPAGE="https://sourceforge.net/projects/allegator/files/Alex4/"
+COPYRIGHT="2003 Johan Peitz"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/allegator/alex4src_data.zip"
+CHECKSUM_SHA256="d266d7fba64fbfedf13240d3d0eb21b8bacbedeaa5f22b26a27d472c8d23f103"
+SOURCE_URI_2="http://downloads.sourceforge.net/allegator/alex4_beos_src.zip"
+CHECKSUM_SHA256_2="b5333034638c19643a7bd8f6e54033b10b354bbcd4e8771b4ac7cd4a33c9e7cf"
+SOURCE_DIR="alex4src"
+SOURCE_DIR_2="alex4_BeOS_src"
+PATCHES_2="alex4-$portVersion.patch"
+# TODO: Check if it would be OK to also import the patches in:
+# http://data.gpo.zugaina.org/gamerlay/games-arcade/alex4/files/
+# and sync with alex4-1.1 in http://gpo.zugaina.org/games-arcade/alex4
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	alex4 = $portVersion
+	app:Alex4 = $portVersion
+	"
+REQUIRES="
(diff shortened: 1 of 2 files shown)


fbrosson:
@waddlesplash, @Begasus: Since the recipe is called alex4beos-1.0.recipe, I changed the PROVIDES to **`alex4beos`** instead of just **`alex4`**. But maybe I should drop the "beos" suffix from the recipe file name and keep the PROVIDES unchanged, i.e. without the "beos" suffix. What do you think?

Does anyone have sound working with Alex4?

fbrosson:
I've just  dropped the "beos" suffix in the name of the recipe to match the Gentoo naming conventions.

Does anyone have sound working with Alex4?

Begasus:
Thanks for taking over, I haven't been on Haiku for a while now (busy with
the dogs), I'll try to take a look at the new PR later this evening :)

2016-06-23 19:36 GMT+02:00 fbrosson <notifications@github.com>:

fbrosson:
No problem ;-)

Begasus:
@fbrosson, running into this atm:

Parse error in .PackageInfo(35:5) -> invalid url
Command '['/bin/package', 'create', '-bi', u'/build-packages/alex4-1.0-1-build.PackageInfo', '-I', u'/packaging/alex4', u'/build-packages/alex4-1.0-1-build.hpkg']' returned non-zero exit status 1
keeping chroot folder /boot/home/haikuports/games-arcade/alex/work-1.0 intact for inspection

I'm suspecting it has something to do with the patch function?

Begasus:
Same thing for Terminal here, maye an updated rdef would be better :) and no sound also ...
PS, it still remains in the Deskbar after closing

Begasus:
If allegro is not installed I get a "nothing provides lib:liballeg_x86>=4.4.2 needed by ..."
Maybe there is something not right for allegro? (also the same as I try to install the devel package for allegro if the lib isn't installed already - although both are available in the packages folder for haikuports)

fbrosson:
Both allegro and alex4 build OK on x86_gcc2 primary arch. So it should not be asking for any **`lib:lib*_x86`**. If you have built on one primary arch (e.g. Haiku/x86), then switched to another arch (e.g. Haiku/x86_gcc2) to build for e.g. x86_gcc2 then you need to clean the **`work-*`** build dir before you build for that other primary arch.
**`haikuporter -c alex4`** will get rid of the right **`work-*`** directory.

Begasus:
I know they build ok (allegro and alex - and in this case it was atanks which is x86), the haikuporter -c option I also know of (I usualy do this to net clutter the system to much with unnescecary files:) )
I just ment, if allegro is not installed and I try to install the devel file for allegro I'm getting the message that there is nothing providing allegro (be it x86_gcc2 or x86)

fbrosson:
Sorry, I though you were having the error message about allegro while installing alex4. I could not guess you had built allegro_x86 and were trying to install atanks_x86 ;-)
BTW, I guess you have found, in the meanwhile, that you needed to install allegro_x86 before (or with) atank_x86.

waddlesplash:
Can this be merged?

fbrosson:
The patches mentioned in the TODO part of the commit comment allow to save some prefs. I was willing to integrate them but it needs some time because we would need to patch a patch. That's not difficult but it needs some time because it's almost like solving a puzzle...
Since I currently don't have enough free time for that, I'm OK with merging the recipe like as is.
Should I make some changes in the commit message and/or the title of this PR?

waddlesplash:
Nah, that's OK.

PR #668, opened 2016-06-25
smpeg2 new recipe

Begasus:


media-libs/smpeg2/smpeg2-2.0.0.recipe
```diff
@@ -0,0 +1,88 @@
...
+CHECKSUM_SHA256="979a65b211744a44fa641a9b6e4d64e64a12ff703ae776bafe3c4c4cd85494b3"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	smpeg2$secondaryArchSuffix = 2.0.0 compat >= 2
+	cmd:smpeg2 = 2.0.0 compat >= 2
+	cmd:plaympeg$secondaryArchSuffix = 2.0.0 compat >= 2
+	lib:libsmpeg2$secondaryArchSuffix = 2.0_0.0.0 compat >= 2.0_0
```
fbrosson: I think underscores are not allowed for library versioning. All those I have seen so far only have digits and dots.

Begasus: It's a rework on the smpeg recipe that's in the tree atm, if it needs corrections I would like to hear :)

Begasus: The created lib is called "libsmpeg2-2.0.so.0.0.0", should I alter this as?:
lib:libsmpeg2$secondaryArchSuffix = 0.0.0 compat >= 0
PS, in the sdl libs I see more "_" in the recipes?

Begasus: Or should be go with lib:libsmpeg2$secondaryArchSuffix = $portVersion compat >= 2.0.0 ?

fbrosson: I think it should be just **` lib:libsmpeg2_2.0$secondaryArchSuffix`** without any lib versioning because a "2.0" versioning is already in the name of the lib before the ".so". BTW, you might wish to replace the **`_2.0`** by **`_${portVersion%\.*}`** to obtain: **` lib:libsmpeg2_${portVersion%\.*}$secondaryArchSuffix`** but I don't know if there is a wide acceptance for that. (I've used it in a couple of recipes and it was accepted ;-)

haikuporter has an excellent option that I always use: **`-S`**. It enables a _strict_ mode which reports many errors, like, for example, when you forget items for the **`PROVIDES*`** or **`REQUIRES*`**. So if you forget a dependency on an external lib it will spot it. And if a libname is wrong in the provides, it will also tell you that you don't have a provides for it. I really recommend always using the strict mode. BTW, if you use it on a recipe that has any of the errors it can spot then it will not create the hpkgs, and I like this behaviour ;-)

Begasus: Using the _S option for a while already, no problems so farVerzonden vanaf mijn Huawei mobiele telefoon-------- Oorspronkelijk bericht --------Onderwerp: Re: [haikuports/haikuports] smpeg2 new recipe (#668)Van: fbrosson Aan: haikuports/haikuports Cc: Begasus ,Author In media-libs/smpeg2/smpeg2-2.0.0.recipe:
> +software patents in the United States."
> +HOMEPAGE="http://www.icculus.org/smpeg"
> +COPYRIGHT="1999-2004 Sam Lantinga, Joe Tennies."
> +LICENSE="GNU LGPL v2"
> +REVISION="1"
> +SOURCE_URI="https://www.libsdl.org/projects/smpeg/release/smpeg2-$portVersion.tar.gz"
> +CHECKSUM_SHA256="979a65b211744a44fa641a9b6e4d64e64a12ff703ae776bafe3c4c4cd85494b3"
> +
> +ARCHITECTURES="x86_gcc2 x86 x86_64"
> +SECONDARY_ARCHITECTURES="x86"
> +
> +PROVIDES="
> +	smpeg2$secondaryArchSuffix = 2.0.0 compat >= 2
> +	cmd:smpeg2 = 2.0.0 compat >= 2
> +	cmd:plaympeg$secondaryArchSuffix = 2.0.0 compat >= 2
> +	lib:libsmpeg2$secondaryArchSuffix = 2.0_0.0.0 compat >= 2.0_0

I think it should be just lib:libsmpeg2_2.0$secondaryArchSuffix without any lib versioning because a "2.0" versioning is already in the name of the lib before the ".so". BTW, you might wish to replace the _2.0 by _${portVersion%\.*} to obtain: lib:libsmpeg2_${portVersion%\.*}$secondaryArchSuffix but I don't know if there is a wide acceptance for that. (I've used it in a couple of recipes and it was accepted ;-)

haikuporter has an excellent option that I always use: -S. It enables a strict mode which reports many errors, like, for example, when you forget items for the PROVIDES* or REQUIRES*. So if you forget a dependency on an external lib it will spot it. And if a libname is wrong in the provides, it will also tell you that you don't have a provides for it. I really recommend always using the strict mode. BTW, if you use it on a recipe that has any of the errors it can spot then it will not create the hpkgs, and I like this behaviour ;-)

fbrosson: Sorry, I though you were not using the **`-S`** option, but you are.
I though haikuporter would have complained about the «**` = 2.0_0.0.0 compat >= 2.0_0`**» but it looks like it does not. So maybe this is not an error after all.

Begasus: There are a few nice options that can be used with haikuporter :)
If the general feeling about this to change the lib versioning please let me know :)

Begasus: Any thoughts on this if it needs changing or not?

fbrosson:
Sorry, I was busy with other recipes and did not test this one yet. I'm trying to build vlc_x86-2.2.1 on x86_gcc2. But I'll give smpeg2 a try asap.

media-libs/smpeg2/smpeg2-2.0.0.recipe
waddlesplash: Since that's what the .so is using, let's keep it.

PR #675, opened 2016-07-02
xerces-c: bump to 3.1.4, drop libxerces-c.la.

fbrosson:


korli:
Please evaluate whether the libtool file can be removed (it's usually not needed).

fbrosson:
There are currently no packages depending on **`devel:xerces_c`** or **`lib:xerces_c`**, so dropping the libtool file **`$developLibDir/libxerces-c.la`** won't break anything.
Although I dropped the static lib, I did not dare to drop the libtool file because it could be usefull, if I'm not mistaken, to users willing to link some custom code against xserces-c's shared lib without having to use **`pkg-config`**.
But if you prefer to have a smaller recipe without the libtool file (and the steps to fix it), then I can change the PR for that.

pulkomandy:
In all cases I've seen, libtool does just fine when the la file is missing. So, it is not needed at all. a simple -l$(libname) in your linker command line is all you need to link a shared library. I think you found one exception to that, but it should stay just an exception, and I'd even say we should fix it on the other side (patch the app that fails to link libintl, instead of putting the .la file back).

Also, I don't understand why you are trying to remove all static libraries. Providing them is a choice of the lib packager, and using them is a choice of app developers (if they want their binaries to be more standalone, but larger). Why remove them? Also, what's your plan for packages which don't provide a shared library at all?

Note that the .a files are only in the "devel" packages and they won't get in the way for users anyway.

fbrosson:
OK, I got it. From now on I'll suggest removing the libtool files.
Regarding the removal of the static libs, I'm _just_ suggesting because I think it would be better to not ship them, but I do understand that it gives less choice. So I'm fine with keeping them if there is interest for that.
So, shall I keep the static lib for this recipe?

fbrosson:
OK, keeping the static lib.

korli:
Thanks!

PR #676, opened 2016-07-05
atanks, new recipe (WIP)

Begasus:
The game shows the first screen but halts from there with the message it can't create the config folder, that however is created but the config file isn't created, not sure if that is why it is stopping from going further ...

### DIFF
--- a/games-action/atanks/atanks-6.4.recipe
+++ b/games-action/atanks/atanks-6.4.recipe
@@ -1,0 +1,49 @@
+SUMMARY="A simple tank game"
+DESCRIPTION="Atomic Tanks is a game in which you control an overly-powerful \
+tank and attempt to blow up other highly powerful tanks. Players \
+get to select a number of weapons and defensive items and then \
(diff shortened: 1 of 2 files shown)


scottmc:
See the section on making files writable on the common issues wiki page: https://github.com/haikuports/haikuports/wiki/Common-Issues-when-Porting-Games
It might also require patching to the code so that it points to the settings directory if they hard coded it.

scottmc:
Have you tried the suggestion above yet?

Begasus:
It's been a while since I last checked, iirc I did some checking with the global files and user settings files ... not sure though :) Will check

Begasus:
Checked it this morning, maybe it's not related to creating the config files, I just can't seem to launch the app properly, it shows up but rendering is a problem I think, if so you can discard this PR

Begasus:
Haven't made any progress so far, this one could be closed (no need to stick around if there is no progress) ...

waddlesplash:
Makes sense.

PR #683, opened 2016-07-08
LuaJit recipe

miqlas:
Just for fun

### DIFF
--- a/dev-lang/luajit/luajit-2.1.0~beta2.recipe
+++ b/dev-lang/luajit/luajit-2.1.0~beta2.recipe
@@ -1,0 +1,64 @@
+SUMMARY="A Just-In-Time Compiler (JIT) for the Lua programming language"
+DESCRIPTION="Lua is a powerful, dynamic and light-weight programming \
+language. It may be embedded or used as a general-purpose, stand-alone \
+language. LuaJIT is a Just-In-Time Compiler (JIT) for it."
+HOMEPAGE="http://luajit.org/luajit.html"
+COPYRIGHT="2005-2015 Mike Pall"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="http://luajit.org/download/LuaJIT-2.1.0-beta2.tar.gz"
+CHECKSUM_SHA256="713924ca034b9d99c84a0b7b701794c359ffb54f8e3aa2b84fad52d98384cf47"
+SOURCE_DIR="LuaJIT-2.1.0-beta2"
+
+ARCHITECTURES="x86 !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	luajit$secondaryArchSuffix = $portVersion compat >= 5.1
+	lib:libluajit_5.1$secondaryArchSuffix = $portVersion compat >= 2
+	cmd:luajit_$portVersion$secondaryArchSuffix
+	cmd:luajit$secondaryArchSuffix
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	luajit$secondaryArchSuffix_devel = $portVersion compat >= 5.1
+	devel:libluajit_5.1$secondaryArchSuffix = $portVersion compat >= 2
+	"
+REQUIRES_devel="
+	luajit$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:gcc$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	# Use amalg target as recommended on luajit homepage to get better performance
+	make $jobArgs amalg PREFIX=$prefix
+}
+
+INSTALL()
+{
+	make install PREFIX=$prefix INSTALL_LIB="$libDir" INSTALL_BIN="$binDir" \
+		INSTALL_INC="$includeDir" INSTALL_SHARE="$dataDir" \
+		INSTALL_MAN="$manDir/man1" INSTALL_LMOD="$dataDir/lua/5.1"
+
+	ln -s $libDir/libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
+
+	# the pkg-config is for 5.1 compat
+	mkdir -p $libDir/pkgconfig
+	cp etc/luajit.pc $libDir/pkgconfig/luajit.pc
+
+	prepareInstalledDevelLib libluajit-5.1
+	fixPkgconfig
+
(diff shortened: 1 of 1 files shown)


dev-lang/luajit/luajit-2.1.0_beta2.recipe
```diff
@@ -0,0 +1,65 @@
...
+	make $jobArgs amalg PREFIX=$prefix
+}
+
+INSTALL()
+{
+	make install PREFIX=$prefix INSTALL_LIB="$libDir" INSTALL_BIN="$binDir" \
+		INSTALL_INC="$includeDir" INSTALL_SHARE="$dataDir" \
+		INSTALL_MAN="$manDir/man1" INSTALL_LMOD="$dataDir/lua/5.1"
+
+	ln -s $libDir/libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
```
fbrosson: If you need to create that symlink, then it should be:
```
	ln -s libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
```

dev-lang/luajit/luajit-2.1.0_beta2.recipe
```diff
@@ -0,0 +1,65 @@
...
+
+	# the pkg-config is for 5.1 comapt
```
fbrosson: s/comapt/compat/

dev-lang/luajit/luajit-2.1.0_beta2.recipe
```diff
@@ -0,0 +1,65 @@
...
+	make $jobArgs amalg PREFIX=$prefix
+}
+
+INSTALL()
+{
+	make install PREFIX=$prefix INSTALL_LIB="$libDir" INSTALL_BIN="$binDir" \
+		INSTALL_INC="$includeDir" INSTALL_SHARE="$dataDir" \
+		INSTALL_MAN="$manDir/man1" INSTALL_LMOD="$dataDir/lua/5.1"
+
+	ln -s libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
```
waddlesplash: This doesn't look right. You're symlinking a file in the local directory to `$libDir`?

korli:
Should be dev-lang/luajit/luajit-2.1.0~beta2.recipe

dev-lang/luajit/luajit-2.1.0_beta2.recipe
fbrosson: @waddlesplash, I do agree that it does not **_look_** right, but if you are creating a symlink, then it is OK to use:
```
ln -s relative/path/to/target $directory/symlink-name
```
because **`relative/path/to/target`** is _not_ relative to the **_current_** directory but to the _directory where the symlink is created_. That's why the following command:
```
ln -s libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
```
will create a «**`libluajit-5.1.so.2 -> libluajit-5.1.so.2.1.0`**» symlink in **`$libDir`**, no matter which current directory you are in when the command is executed.

(Of course, most users go to the directory where the symlink is to be created (before creating it) because this allows to use the sehll auto-complete for the **`relative/path/to/`**. This is why very few people know that "local" symlinks (I mean symlinks to files in the same directory) can be created even if you're not in the target dir.)

So @miqlas, you can safely use the command I had suggested earlier:
```
ln -s libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
```
For the record, you can get the same result with:
```
ln -s -r $libDir/libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
```
So you have the choice ;-)

fbrosson:
@miqlas, I don't know comfortable you are with git, but in case you need help for renaming the recipe, here's what you can do:
```
cd haikuports/dev-lang/luajit
git checkout luajit21 # in case you are not in that branch
git mv luajit-2.1.0_beta2.recipe luajit-2.1.0~beta2.recipe
git commit --amend luajit-2.1.0_beta2.recipe luajit-2.1.0~beta2.recipe
git push -f
```
(Sorry if that was obvious.)

dev-lang/luajit/luajit-2.1.0~beta2.recipe
```diff
@@ -0,0 +1,65 @@
...
+	lib:libluajit_5.1$secondaryArchSuffix = $portVersion compat >= 2
+	cmd:luajit_$portVersion$secondaryArchSuffix
+	cmd:luajit$secondaryArchSuffix
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	luajit${secondaryArchSuffix}_devel = $portVersion compat >= 5.1
```
diversys: Lose braces

dev-lang/luajit/patches/luajit-2.1.0~beta2.patchset
```diff
@@ -0,0 +1,34 @@
...
+diff --git a/src/Makefile b/src/Makefile
+index 3f25192..4f2b3fb 100644
+--- a/src/Makefile
++++ b/src/Makefile
+@@ -224,7 +224,7 @@ TARGET_DYNXLDOPTS=
+ TARGET_LFSFLAGS= -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE
+ TARGET_XCFLAGS= $(TARGET_LFSFLAGS) -U_FORTIFY_SOURCE
+ TARGET_XLDFLAGS=
+-TARGET_XLIBS= -lm
++TARGET_XLIBS= -lroot
```
diversys: This patch shouldn't be needed anymore

dev-lang/luajit/luajit-2.1.0~beta2.recipe
miqlas: I don't know what do you mean.

dev-lang/luajit/patches/luajit-2.1.0~beta2.patchset
miqlas: Patch made before -lm changes in Haiku. Do i really need to redo my patch now?

diversys: I think you can just drop it

dev-lang/luajit/luajit-2.1.0~beta2.recipe
diversys: You can remove braces here

dev-lang/luajit/patches/luajit-2.1.0~beta2.patchset
miqlas: Then the diff hash will change. One just cannot change the patch...

diversys: You can drop the patch altogether

PR #684, opened 2016-07-08
LibPCRE2 recipe

miqlas:
Just for fun

### DIFF
--- a/dev-libs/libpcre2/libpcre2-10.21.recipe
+++ b/dev-libs/libpcre2/libpcre2-10.21.recipe
@@ -1,0 +1,112 @@
+SUMMARY="Perl5 Compatible Regular Expressions"
+DESCRIPTION="The PCRE2 library is a set of functions that implement regular expression \
+pattern matching using the same syntax and semantics as Perl 5. PCRE2 has its \
+own native API, as well as a set of wrapper functions that correspond to the \
+POSIX regular expression API. The PCRE2 library is free, even for building \
+proprietary software.
+PCRE2 is a re-working of the original PCRE library to provide an entirely new \
+API."
+HOMEPAGE="http://www.pcre.org/"
+COPYRIGHT="1997-2016 University of Cambridge"
+LICENSE="PCRE"
+REVISION="1"
+SOURCE_URI="https://ftp.csx.cam.ac.uk/pub/software/programming/pcre/pcre2-$portVersion.tar.bz2"
+CHECKSUM_SHA256="c66a17509328a7251782691093e75ede7484a203ebc6bed3c08122b092ccd4e0"
+SOURCE_DIR="pcre2-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86 x86_gcc2"
+
+PROVIDES="
+	libpcre2${secondaryArchSuffix} = $portVersion compat >= 10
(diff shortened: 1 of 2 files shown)


dev-libs/libpcre2/libpcre2-10.21.recipe
```diff
@@ -0,0 +1,114 @@
+SUMMARY="Perl5 Compatible Regular Expressions"
+DESCRIPTION="The PCRE2 library is a set of functions that implement regular expression \
+pattern matching using the same syntax and semantics as Perl 5. PCRE2 has its \
+own native API, as well as a set of wrapper functions that correspond to the \
+POSIX regular expression API. The PCRE2 library is free, even for building \
+proprietary software.
+PCRE2 is a re-working of the original PCRE library to provide an entirely new \
+API."
+HOMEPAGE="http://www.pcre.org/"
+COPYRIGHT="	1997-2016 University of Cambridge"
```
fbrosson: No tab required for the first line of COPYRIGHT.

dev-libs/libpcre2/libpcre2-10.21.recipe
```diff
@@ -0,0 +1,114 @@
...
+	"
+
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc${secondaryArchSuffix}
+	cmd:make
+	cmd:ld${secondaryArchSuffix}
+	cmd:libtoolize
```
fbrosson: Safer with:
```
```
BTW, I think it would be better to use **`$secondaryArchSuffix`** instead of **`${secondaryArchSuffix}`** at every place where it is OK. But feel free to ignore this remark if you prefer using **`${secondaryArchSuffix}`**.

dev-libs/libpcre2/libpcre2-10.21.recipe
```diff
@@ -0,0 +1,114 @@
...
+	autoconf
+	automake --add-missing
+	runConfigure ./configure \
+		--with-pic \
+		--enable-pcre2-16 \
+		--enable-pcre2-32 \
+		--enable-jit \
+		--enable-pcre2grep-libz \
+		--enable-pcre2grep-libbz2 \
+		--enable-pcre2test-libedit \
```
fbrosson: Having a **`\`** here is risky because if someone drops the 2 following lines then the **`make $jobArgs`** wil be attached to the runConfigure command as extra arguments.

dev-libs/libpcre2/libpcre2-10.21.recipe
```diff
@@ -0,0 +1,114 @@
...
+		$developDir \
+		$manDir/man3
+}
+
+TEST()
+{
+	make check
+}
+
+# ----- devel package -------------------------------------------------------
```
fbrosson: The recommended place for PROVIDES_devel and REQUIRES_devel has been moved up between REQURIES and BUILD_REQUIRES.

dev-libs/libpcre2/libpcre2-10.21.recipe
miqlas: Then this someone needs to fix it :)

waddlesplash: @fbrosson I trust anyone modifying the recipe will know enough about Bash to not do that :)

PR #685, opened 2016-07-08
GNU Fortran enabled in GCC

miqlas:
For mathematical things.


korli:
I hope you understand this won't be merged as is: 
1/ gcc-fortran should be a sub package (this shouldn't be installed by default for users which don't need it).
2/ There should be at least one recipe which needs gcc with fortran enabled, to be able to test a build

miqlas:
Hi korli,
can you point me a recipe what i can use as an example?
BTW: everybody needs fortran :)

korli:
Gcc recipe has already a sub package for syslibs. Dev files for Fortran should go in the sub package Fortran, runtime should go in syslibs.

sys-devel/gcc/gcc-5.4.0_2016_07_08.recipe
```diff
@@ -0,0 +1,347 @@
+SUMMARY="C/C++ compiler for target ${effectiveTargetMachineTriple}"
+DESCRIPTION="The standard compiler for non-legacy Haiku (i.e. for all \
+architectures other than x86_gcc2)."
+HOMEPAGE="http://gcc.gnu.org"
+
+SOURCE_URI="https://ftp.gnu.org/gnu/gcc/gcc-5.4.0/gcc-5.4.0.tar.bz2"
```
fbrosson: You might wish to use this:
```
SOURCE_URI="http://ftpmirror.gnu.org/gcc/gcc-5.4.0/gcc-5.4.0.tar.bz2"
```
or this:
```
SOURCE_URI="http://ftpmirror.gnu.org/gcc/gcc-${portVersion%%_*}/gcc-${portVersion%%_*}.tar.bz2"
```
And for HOMEPAGE you might prefer to use this:
```
HOMEPAGE="https://gcc.gnu.org/"
```

sys-devel/gcc/gcc-5.4.0_2016_07_08.recipe
```diff
@@ -0,0 +1,347 @@
...
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:makeinfo
+	cmd:sed
+	cmd:strip
+	cmd:tar
+	cmd:xargs
+	"
+
+SOURCE_DIR="gcc-5.4.0"
```
fbrosson: This also works:
```
SOURCE_DIR="gcc-${portVersion%%_*}"
```

sys-devel/gcc/gcc-5.4.0_2016_07_08.recipe
korli: @fbrosson this has the inconvenience to be http only. What if we want secure sources?

fbrosson: Well, since we have CHECKSUM_SHA256 we don't really need https for SOURCE_URI.
Of course, if the mirror rediretor was available in https I would use it, but it is not.
IMHO using the mirror redirector, even on http, is much more interesting because it both speeds up the download time and makes it more resilient in case the main GNU site becomes unavailable.
The only real (but small) problem with the mirror redirector is that in very rare cases the (random) mirror that we could be redirected to could be missing the archive. That happened to me twice in almost 6 months. And the workaround is easy: relaunch haikuporter ;-)

korli: It's just inconsistent, it's secure or not. I guess we should factorise the mirror URL for most used hosting services, this way, one could switch from secure to non secure, based on the flag in Haikuporter config.

fbrosson: The added value of using https for SOURCE_URI is _just_ that it _looks_ consistent with using https for HOMEPAGE.
The added value of using the mirror redictor for SOURCE_URI is that the archives load faster and nobody will ever need to fetch it manually from a mirror if the main site becomes unavailable for e.g. a maintainance downtime.
What brings security is CHECKSUM_SHA256. Having https for SOURCE_URI does not bring anything more than _perceived consistency_.
So the choice is between _perceived consistency_ and _efficiency_.
The administrator of the GNU download site made the effort to build a mirror redirector service in the hope that it would be used. Not using it is just bad practice. Sorry if this reply looks rude.

korli: It seems everyone that comes to this project has his own set of rules. It reminds me Jonathan Schleifer, who ironically has the opposite stance to yours (more secure is better).
Moreover I think the builder at hpkg.mlotz.ch is configured to only build recipes with safe sources. So I don't see any bad practice at using HTTPS FWIW.

fbrosson: * Do we have any rules that forbid to use a redirector that would not use https?
* Are we allowed to say that some existing (or implicit) rule is wrong and change it?
* Who are we to force GNU to request that their mirrors accept https?
* If everybody refused to use download redirector services just because they use http instead of https, then most mirrors would be useless.
* What's the value added of using https if we already have a safe way to check that the archives are OK?
* Even if the majority thinks that we should use https for SOURCE_URI, this does not mean it is right.

I'm not going to stop contributing if you decide that we should not use "unsafe" SOURCE_URI, but I really think the project should avoid that mistake. If I did not manage to convince you, would you agree to ask others what they think about this question?
I'm not trying to fight against you. And I wish I had found a better way to explain my point of view.
Last but not least: I won't **ever** start saying "ha ha, I won!" if other contributors say I'm right.

fbrosson: > So I don't see any bad practice at using HTTPS FWIW.

I'm not saying that using https is bad practice. I'm saying that what is bad practice is to ignore download redirector services when they are available. That's quite different.

pulkomandy: There is the "safe source" problem, but HTTP is considered "safe" if there is a checksum (we can check we got the right file). What's "unsafe" is downloading from eg. a git repo, where you can't checksum anything (there isn't a single file, but several ones).

I'm not sure https on mirrors is very useful, because:
1) It would redirect to somewhere else anyway
2) You are redirected to a random website, and get a matching certificate. That does not mean you trust whatever you have been redirected to?

Using the official mirror from GNU for the initial writing of the recipe is a good idea, because it makes sure we get the official binary and the correct SHA256. But once the recipe is written, switching to the mirrors is ok, we just need to be careful when updating the recipe, that the SHA256 always matches what's on GNU main website (so we don't get a corrupted/tainted file from a mirror and think it is the "official" source).

korli: I drop it, wasted my time really.

fbrosson: Thanks!
BTW, you should not be upset with this thread. You have _much more_ experience than me in _many_ areas!

mmuman:
I'm porting SuperLU currently (need it for OpenToonz) and it wants a fortran compiler...
I disabled this part for now, but it'd be nice to have.

miqlas:
It won't merged, will try later to do a proper recipe. Now closing it as useless.

sys-devel/gcc/gcc-5.4.0_2016_07_08.recipe
mmuman: HTTPS is not only about avoiding tampering with the source, but also hiding what specific file you wanted. HTTP sends plain requests, including the vhost and filename. Not something most of us really care about right now, but in theory some people might, which is the reason for projects like [HTTPS-Everywhere](https://www.eff.org/https-everywhere).

PR #687, opened 2016-07-09
libmng build issues fix

return:
This commit fixes a few errors encountered whilst building libmng on haiku.


waddlesplash:
Thanks!

PR #689, opened 2016-07-09
fizmo, playinteractive fiction, also known as textadventures, new recipe

Begasus:


games-engines/fizmo/fizmo-0.7.10.recipe
```diff
@@ -0,0 +1,67 @@
...
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	fizmo$secondaryArchSuffix = $portVersion
+	cmd:fizmo_console$secondaryArchSuffix = $portVersion
+	cmd:fizmo_ncursesw$secondaryArchSuffix = $portVersion
+	"
+
```
fbrosson: The [guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering) say we don't need a blank line between PROVIDES and REQUIRES.
Same rule between PROVIDES_devel and REQUIRES_devel..
Same rule between BUILD_REQUIRES and  BUILD_PREREQUIRES.

There seems to be an exception, however, when there are conditional additions to PROVIDES or PROVIDES_devel, in which cases the conditional addition is followed by a blank line.
Well, this is what I have noticed, but I don't know if it is written somewhere.

Begasus: My bad, should have seen that, fixed now :)

PR #691, opened 2016-07-10
libxpm: enable 2nd arch, drop libXpm.la.

fbrosson:
* Enable 2nd arch and drop documentation when building for them.
* Drop the libtool library file since we don't need it.
* ~~Remove unneeded **`lib:lib{xau,xcb}`** from REQUIRES.~~
* Add **`devel:lib{x11,xcb,x11}`** to REQUIRES_devel.
* Add **`$secondaryArchSuffix`** to **`cmd:libtoolize`** in BUILD_PREREQUIRES.
* HOMEPAGE: http://xorg.freedesktop.org/ moved to https://www.x.org/.
* ARCHITECTURES: Mark as tested on x86 and x86_64.

### DIFF
--- a/x11-libs/libxpm/libxpm-3.5.11.recipe
+++ b/x11-libs/libxpm/libxpm-3.5.11.recipe
@@ -5,27 +5,30 @@ The development work is being done in conjunction with the freedesktop.org \
 community. \
 The X.Org Foundation is the educational non-profit corporation whose \
 Board serves this effort, and whose Members lead this work."
-HOMEPAGE="http://xorg.freedesktop.org/"
+HOMEPAGE="https://www.x.org/"
 COPYRIGHT="1989-95 GROUPE BULL
 	1998 Arnaud LE HORS
-	1996 Lorens Younes
-	"
+	1996 Lorens Younes"
 # TODO: check all the licenses in the COPYING file in the next century.
 LICENSE="MIT"
-REVISION="2"
-SOURCE_URI="http://xorg.freedesktop.org/releases/individual/lib/libXpm-3.5.11.tar.gz"
+REVISION="3"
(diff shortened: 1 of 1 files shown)


waddlesplash:
Looks good to me. Merge?

fbrosson:
The recipes builds fine, but I'd like to improve it a little.
A program which only calls libXpm should not need to say it depends on libX11. Without libtool files such a program would need to say it needs both libXpm and libX11. Whereas with an appropriately fixed libXpm.la that program would only need to say it needs libXpm.
Yes, I know that any program that calls functions in libXpm would probably also call functions in libX11, but if we chose to keep libtool files then keeping only some of them (instead of all of them) would not be an improvement.
Indeed, if we are working on a recipe that has many direct or indirect dependencies, the we would need to list all indirect dependencies which did not get included in the libtool files of intermediate dependencies.
Another argument in favor of keeping and fixing libtool files even for simple cases like libXpm is that it can serve as a template for other packages that have more dependencies.

If you read the recipe for tesseract you'll see that it needs to list indirect dependencies. This is not smart but we can't avoid that if we don't keep and fix libtool files.

mmuman:
Actually no, the only reason we have libx11 and all the other crap in is because libxpm needs it, and I need libxpm in my XEmacs port to display the toolbar buttons, even with the native GUI.

waddlesplash:
>  Yes, I know that any program that calls functions in libXpm would probably also call functions in libX11, but if we chose to keep libtool files then keeping only some of them (instead of all of them) would not be an improvement.

We're already having that argument in another thread, and the consensus seems to be there isn't a lot of reason for keeping them, so just drop them. Same for the libx11 PR.

PR #699, opened 2016-07-13
mutt recipe

miqlas:
For the sake of IMAP.
Sadly gmail requires SASL.

### DIFF
--- a/mail-client/mutt/mutt-1.7.2.recipe
+++ b/mail-client/mutt/mutt-1.7.2.recipe
@@ -1,0 +1,97 @@
+SUMMARY="Small but very powerful text-based mail client"
+DESCRIPTION="A text-based email client for Unix-like systems. \
+It was originally written by Michael Elkins in 1995 and released \
+under the GNU General Public License version 2 or any later version.
+
+The Mutt slogan is All mail clients suck. This one just sucks less."
+HOMEPAGE="http://www.mutt.org/"
+COPYRIGHT="1996-2016 Michael R. Elkins
+	1996-2002 Brandon Long
+	1997-2009 Thomas Roessler
+	1998-2005 Werner Koch
+	1999-2014 Brendan Cully
+	1999-2002 Tommi Komulainen
+	2000-2004 Edmund Grimley Evans
+	2006-2009 Rocco Rutte
+	2014-2015 Kevin J. McCarthy"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="ftp://ftp.mutt.org/pub/mutt/mutt-$portVersion.tar.gz"
+CHECKSUM_SHA256="1553501687cd22d5b8aaee4dc5a7d9dcf6cc61d7956f6aabaadd252d10cd5ff9"
+PATCHES="mutt-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	mutt$secondaryArchSuffix = $portVersion
+	cmd:mutt$secondaryArchSuffix = $portVersion
+	cmd:flea$secondaryArchSuffix = $portVersion
+	cmd:muttbug$secondaryArchSuffix = $portVersion
+	cmd:pgpewrap$secondaryArchSuffix = $portVersion
+	cmd:pgpring$secondaryArchSuffix = $portVersion
+	cmd:smime_keys$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgpgme$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libgpgme$secondaryArchSuffix
+	lib:libgpg_error$secondaryArch >= 0
+	lib:libgettextpo$secondaryArch
+	lib:libassuan$secondaryArch
+	lib:libintl$secondaryArch
+	lib:libassuan$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	devel:libncurses$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libgpgme$secondaryArchSuffix
+	devel:libgpg_error$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,91 @@
...
+COPYRIGHT="1996-2014 Michael R. Elkins
+1996-2002 Brandon Long
+1997-2009 Thomas Roessler
+1998-2005 Werner Koch
+1999-2014 Brendan Cully
+1999-2002 Tommi Komulainen
+2000-2004 Edmund Grimley Evans
+2006-2009 Rocco Rutte
+2014-2015 Kevin J. McCarthy
+"
```
waddlesplash: Quote goes on previous line. And all the continuation COPYRIGHT lines should be indented by 1 tab.

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,91 @@
...
+	cmd:gawk
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:m4
+	cmd:gawk
+	"
+
+BUILD()
+{
+	libtoolize --force
```
waddlesplash: Shouldn't this be `--force --copy --install`?

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,90 @@
+SUMMARY="Small but very powerful text-based mail client"
+DESCRIPTION="A text-based email client for Unix-like systems. \
+It was originally written by Michael Elkins in 1995 and released \
+under the GNU General Public License version 2 or any later \
+version.
```
diversys: Put it into previous line

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+	2006-2009 Rocco Rutte
+	2014-2015 Kevin J. McCarthy"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="ftp://ftp.mutt.org/pub/mutt/mutt-$portVersion.tar.gz"
+CHECKSUM_SHA256="98b26cecc6b1713082fc880344fa345c20bd7ded6459abe18c84429c7cf8ed20"
+PATCHES="mutt-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86 ?arm"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86 ?arm"
```
fbrosson: I think you can drop the **`?arm`** in **`SECONDARY_ARCHITECTURES`**.

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+	"
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:autoreconf
+	cmd:bison
+	cmd:gawk
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:m4
+	cmd:gawk
```
fbrosson: You can drop the second **`cmd:gawk`** ;-)

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+	cmd:smime_keys$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgpgme$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libgpgme$secondaryArchSuffix
+	lib:libgpg_error$secondaryArch
+	lib:libgpg_error.so.0$secondaryArch
```
fbrosson: I think you should replace
```
	lib:libgpg_error$secondaryArch
	lib:libgpg_error.so.0$secondaryArch
```
by just
```
	lib:libgpg_error$secondaryArch
```
(if you don't need a specific version of the lib),
or by
```
	lib:libgpg_error$secondaryArch >= 0
```
(if you need libgpg_error.so.0.)

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,87 @@
...
+	devel:libgettextlib$secondaryArchSuffix
+	devel:libassuan$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:autoreconf
+	cmd:bison
+	cmd:gawk
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
```
fbrosson: You need to add a **`cmd:libtoolize$secondaryArchSuffix`** to BUILD_PREREQUIRES. The best place for it is just before **`cmd:make`**.
BTW, the **`cmd:gpgme_config$secondaryArchSuffix`** currently in BUILD_REQUIRES would be better in BUILD_PREREQUIRES, between the **`cmd:gcc$secondaryArchSuffix`** and the **`cmd:libtoolize$secondaryArchSuffix`**.

mail-client/mutt/mutt-1.6.1.recipe
```diff
@@ -0,0 +1,87 @@
+SUMMARY="Small but very powerful text-based mail client"
+DESCRIPTION="A text-based email client for Unix-like systems. \
+It was originally written by Michael Elkins in 1995 and released \
+under the GNU General Public License version 2 or any later version.
+
+The Mutt slogan is All mail clients suck. This one just sucks less."
+HOMEPAGE="http://www.mutt.org"
```
fbrosson: IMHO it would be better to have
```
HOMEPAGE="http://www.mutt.org/"
```
I know most browsers hide the trailing slash when there is nothing after it, but, to be correct, a URL needs either a path or a trailing slash.
I also know that this is a detail and that some recipes are missing that trailing slash.
But I think we should have correct URLs in our recipes.
A practical reason for keeping the slash is that when wrong URLs are copy/pasted to forums that make hyperlinks with them, these hyperlinks do not always work because the wrapper that creates the hyperlinks is not able to correctly detect the end of the domain name when it is followed by a dot. I've already seen that in some places. This never happens with correct URLs ;-)

PR #700, opened 2016-07-13
Ruby recipe, FIXME

miqlas:
FIXME! :)


waddlesplash:
Why did you open a PR if you don't want comments/review?

miqlas:
I DO want comments, review. I have no idea, why you think the opposite.
I do my PR’s, because so the other peoples can see, on which field am i working and maybe they can give tips, ideas. Or i’m just stuck, and i’m in need of help.

waddlesplash:
What's in this recipe? Is it just a rename of the older Ruby recipe we already have? Or does it work better?

korli:
@kallisti5 pushed Haiku changes upstream, maybe let him review.

miqlas:
I tried my best, but i get this all the time if i try to compile ruby with haikuporter: 
http://termbin.com/gl3s
It works without HP,

miqlas:
I checked the missing symbols:
* _ZTIN8BPrivate7Libroot13LocaleBackendE : it is in libroot
* ___tls_get_addr : it should be in libsupc++ (but i'm not sure)

korli:
This builds fine on x86_64, so my suggestion would be to disable architectures for which the build fails, and document the failure on _x86. This PR could then be merged at least.

dev-lang/ruby/ruby-2.3.1.recipe
```diff
@@ -0,0 +1,107 @@
...
+programming language. It was designed and developed in the mid-1990s \
+by Yukihiro 'Matz' Matsumoto in Japan.
+
+Ruby embodies syntax inspired by Perl with Smalltalk-like features and \
+was also influenced by Eiffel and Lisp.  It supports multiple \
+programming paradigms, including functional, object oriented, and \
+imperative. It also has a dynamic type system and automatic memory \
+management. Therefore, it is similar in varying degrees to, Smalltalk, \
+Python, Perl, Lisp, Dylan, and CLU."
+HOMEPAGE="http://www.ruby-lang.org"
```
fbrosson: You might wish to switch to https and add a trailing slash (to have a really valid URI):
```
HOMEPAGE="https://www.ruby-lang.org/"
```

dev-lang/ruby/ruby-2.3.1.recipe
```diff
@@ -0,0 +1,107 @@
...
+	lib:libcrypto$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libncursesw$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	ruby${secondaryArchSuffix}_devel = $portVersion
+	devel:libruby_static$secondaryArchSuffix = $portVersion compat >= 2.2
+	lib:libruby_static$secondaryArchSuffix = $portVersion compat >= 2.2
```
fbrosson: Shouldn't **`lib:libruby_static`** go to PROVDES (instead of PROVIDES_devel)?
In other words, if any other recipe ever needs it, would it be desirable to make it depend on ruby_devel?

Does haikuporter complain if you build in strict mode? (**`haikuporter -S ruby-2.3.1`**)

dev-lang/ruby/ruby-2.3.1.recipe
```diff
@@ -0,0 +1,107 @@
...
+	devel:libiconv$secondaryArchSuffix
+	devel:libncursesw$secondaryArchSuffix
+	devel:libffi$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
+	cmd:autoconf
+	cmd:bison
+	cmd:git
```
fbrosson: I think we should avoid depending on **`cmd:git`**. Does the build fail without it?

dev-lang/ruby/ruby-2.3.1.recipe
KapiX: If libruby_static is a static library (.a file) then it shouldn't have `lib:` entry at all.

miqlas:
Version bump to 2.4.0.
Any news about merge?

miqlas:
Please, do not merge yet, it is horribly broken.
See?

> ~ ❯❯❯ uname -a
> Haiku shredder 1 hrev50788 Dec 24 2016 19:53:03 x86_64 x86_64 Haiku
> ~ ❯❯❯ ruby --version
> ruby 2.4.0p0 (2016-12-24 revision 57164) [x86_64-haiku]
> ~ ❯❯❯ ruby gem
> <internal:gem_prelude>:4:in `require': Operation not supported -- /packages/ruby-2.4.0-1/.self/lib/ruby/2.4.0/rubygems.rb (LoadError)
>         from <internal:gem_prelude>:4:in `<internal:gem_prelude>'
> ~ ❯❯❯ ruby -d -w gem
> Exception `LoadError' at <internal:enc/prelude>:4 - Operation not supported -- /packages/ruby-2.4.0-1/.self/lib/ruby/2.4.0/unicode_normalize.rb
> Exception `LoadError' at <internal:enc/prelude>:4 - Operation not supported -- /packages/ruby-2.4.0-1/.self/lib/ruby/2.4.0/unicode_normalize.rb
> Exception `LoadError' at <internal:gem_prelude>:4 - Operation not supported -- /packages/ruby-2.4.0-1/.self/lib/ruby/2.4.0/rubygems.rb
> Exception `LoadError' at <internal:gem_prelude>:4 - Operation not supported -- /packages/ruby-2.4.0-1/.self/lib/ruby/2.4.0/rubygems.rb
> <internal:gem_prelude>:4:in `require': Operation not supported -- /packages/ruby-2.4.0-1/.self/lib/ruby/2.4.0/rubygems.rb (LoadError)
>         from <internal:gem_prelude>:4:in `<internal:gem_prelude>'
> ~ ❯❯❯

miqlas:
See for more info: https://dev.haiku-os.org/ticket/1791#comment:8

mmuman:
This is not a fix. It's just hiding another issue: the fact that Ruby has to maintain a list of installed gems in a single file.

Either it has to create this file somewhere in non-packaged/ and hope it won't get out of sync when pkgman updates Ruby packages, or it has to be fixed to enumerate them directly by examining the directories.
The fact that this file is located in a packaged tree also hints that Ruby will probably want to also use packaged location for manual gems installs, which obviously won't work either.

waddlesplash:
Superseded by #1321.

PR #701, opened 2016-07-15
asciidoc: bump to 8.6.9.

fbrosson:
~~asciidoc: bump to 8.6.9, move config files to settings/asciidoc/.~~

~~Also switch to ARCHITECTURES="any"~~

### DIFF
--- a/app-text/asciidoc/asciidoc-8.6.9.recipe
+++ b/app-text/asciidoc/asciidoc-8.6.9.recipe
@@ -6,14 +6,14 @@ be translated to many formats including HTML, PDF, EPUB, man page.
 AsciiDoc is highly configurable: both the AsciiDoc source file syntax and the \
 backend output markups (which can be almost any type of SGML/XML markup) can \
 be customized and extended by the user."
-HOMEPAGE="http://asciidoc.org"
-COPYRIGHT="2002-2011 Stuart Rackham"
+HOMEPAGE="http://asciidoc.org/"
+COPYRIGHT="2002-2013 Stuart Rackham"
 LICENSE="GNU GPL v2"
-REVISION="2"
-SOURCE_URI="http://sourceforge.net/projects/asciidoc/files/asciidoc/8.6.8/asciidoc-8.6.8.tar.gz"
-CHECKSUM_SHA256="ffb67f59dccaf6f15db72fcd04fdf21a2f9b703d31f94fcd0c49a424a9fcfbc4"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/asciidoc/asciidoc-$portVersion.tar.gz"
+CHECKSUM_SHA256="78db9d0567c8ab6570a6eff7ffdf84eadd91f2dfc0a92a2d0105d323cab4e1f0"
 
-ARCHITECTURES="x86_gcc2 ?x86"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
 
 PROVIDES="
 	asciidoc = $portVersion compat >= 8.6.8
@@ -28,9 +28,9 @@ REQUIRES="
 	"
 
 BUILD_REQUIRES="
+	haiku_devel
 	"
 BUILD_PREREQUIRES="
-	haiku_devel
 	cmd:awk
 	cmd:make
 	cmd:sed
@@ -40,11 +40,16 @@ BUILD()
 {
 	# TODO: separate data and configuration files.
 	runConfigure ./configure \
-		--sysconfdir=$dataDir/asciidoc
+		--sysconfdir=$dataDir
 	make
 }
 
 INSTALL()
 {
 	make install
+}
+
+TEST()
+{
+	make test
 }


fbrosson:
Should I simplify GLOBAL_WRITABLE_FILES and drop the **`*.default`** _shadow_ symlinks?

waddlesplash:
Does the package have .pyc files in it? If so, it's arch-specific, not arch-independent.

Nah, don't remove those.

fbrosson:
Nope, there are no .pyc files. BTW, Gentoo and Arch Linux have also chosen to keep that package arch-independent by not compiling the Python scripts.
Regarding the **`.default`** symlinks I created, I could not find a way to tag them "replace" instead of "keep-old". All I found was "auto-merge" and "manual", so I took the latter.

Docs I read:
https://dev.haiku-os.org/wiki/PackageManagement/BuildingPackages
https://dev.haiku-os.org/wiki/PackageManagement/PackagingPolicy

waddlesplash:
Erm. Those aren't settings files, they're templates and the like. Configure knows where `$settingsDir` is, if it wanted them there, it would have put them there. So put them back.

fbrosson:
This is what we have in 8.6.**8**:
```
BUILD()
{
	# TODO: separate data and configuration files.
	runConfigure ./configure \
		--sysconfdir=$dataDir/asciidoc
	make
}

INSTALL()
{
	make install
}
```
a) Are you suggesting that we should keep that for 8.6.**9** ?
Because in 8.6.**9** I did not use **`--sysconfdir=$dataDir/asciidoc`** at all. I'm just suggesting an implementation for the **`TODO: separate data and configuration files.`**
I might have misunderstood your comment, though.
b) Do you agree that **`asciidoc.conf`** (and all its friends) should be in the **`$settingsDir/asciidoc/`** directory tree (instead of **`$dataDir/asciidoc/`**) ?
c) Do you agree that the recipe should provide a template for them in the **`$dataDir/asciidoc/defaults/`** directory tree?
d) Should I keep just one symlink per config file, instead of two?
I'm suggesting two symlinks per config file:
```
settings/asciidoc/asciidoc.conf  -> asciidoc.conf.default
settings/asciidoc/asciidoc.conf.default -> ../../data/asciidoc/defaults/asciidoc.conf
```
but, if you prefer, I can create only one per config file:
```
settings/asciidoc/asciidoc.conf  -> ../../data/asciidoc/defaults/asciidoc.conf
```
I can also duplicate each config file to have the editable copy in **`$settingsDir/asciidoc/`** and the template in **`$dataDir/asciidoc/defaults/`**.
The reason why I'm suggesting two symlinks is because most users won't probably need to change any of these config files. And if they wanted to do so, they would just need to delete the config file (e.g.asciidoc.conf) and copy the template by using the symlink to it, e.g.:
```
cd /systems/settings/asciidoc
rm  asciidoc.conf && cp asciidoc.conf.default asciidoc.conf
```
Providing only one symlink might be a better practice, but it has the downside of complicating a little bit the steps a user would have to follow to customize a config file.

If we use ordinary files instead of symlinks then each config file would appear twice in the package, so installing it would take more space on disk.
e) Should I populate the **`$settingsDir/asciidoc/`** directory tree with ordinary files  instead of symlinks?
Thanks!

waddlesplash:
According to AsciiDoc documentation, (I think) one can have their own configuration files on a per directory basis. In this case, the system ones are just defaults, and so if they are readonly it doesn't matter. I think that makes the most sense.

fbrosson:
After all, let's keep it simple, with read-only settings, as you suggested.
BTW, should I keep the settings in **`/system/data/asciidoc/asciidoc/`** or should I move them to **`/system/data/asciidoc/`** ?

waddlesplash:
The latter. Why was there an extra subdirectory of the same name, anyway?

fbrosson:
I don't know. It was already like that in 8.6.8.
Updated right now to have the settings in **`/system/data/asciidoc/`**.

fbrosson:
Oops, browsing the log for asciidoc I found this:
> commit ab9940f435a86d67a7e55e239bc8eff67390066e
> Revert "asciidoc: Switch to architecture "any""

> This reverts commit 1606e09674521d6da3db5bde203b546229d44576.

>  As Oliver said we might want to support different python version on different architectures, so we should stick to arch-dependent for python stuff anyway.

Should I drop the switch to "any" arch?

waddlesplash:
Yes, please. :)

fbrosson:
OK, I'll do it.
Oh, wait, I think it's not necessary because the only python scripts in this package are a2x.py and asciidoc.py, and both have a sane shebang: **`#!/usr/bin/env python`**
What do you think? Shall we keep the "any" arch?

waddlesplash:
>  Oh, wait, I think it's not necessary because the only python scripts in this package are a2x.py and asciidoc.py, and both have a sane shebang: #!/usr/bin/env python

It really doesn't have anything to do with the shebang. ".pyc" and ".pyo" files are arch-specific, and even if none are generated presently, that can easily change in the future if they decided to pre-compile them at install time (as other projects do.)

fbrosson:
OK, updated.
BTW, I've added **`cmd:gcc`** and **`cmd:ld`** to BUILD_PREREQUIRES hoping I would get compiled python scripts, but I got no **`.pyc`** files. Is there something else we need to obtain these?
Or should I just drop (or comment out) the **`cmd:gcc`** and **`cmd:ld`** I've added? Thanks!

waddlesplash:
> BTW, I've added cmd:gcc and cmd:ld to BUILD_PREREQUIRES hoping I would get compiled python scripts, but I got no .pyc files. Is there something else we need to obtain these?

... That is not how compiled Python works, even remotely. Even *remotely*. The Python interpreter will automatically create the .pyc files when it's run and it has permissions to save them. So just run the files once and it should work, if the interpreter wants to.

waddlesplash:
> Or should I just drop (or comment out) the cmd:gcc and cmd:ld I've added? Thanks!

Of course you should, they're totally unrelated.

fbrosson:
OK. Done.

PR #718, opened 2016-07-16
git: bump to 2.9.2.

fbrosson:


waddlesplash:
Does it work?

fbrosson:
Yep, it does.

I forgot to read the [release notes](https://raw.githubusercontent.com/git/git/master/Documentation/RelNotes/2.9.2.txt) before creating this PR... and it says the only changes are in a few tests, so I guess that merging without a fix for the compilation of the tests would not make much sense.
By changing line 445 of **`git-2.9.2/Makefile`** to:
```
PTHREAD_LIBS = -lroot -lnetwork -liconv -lbsd -lposix_error_mapper
```
I've been able to make a few steps forward, but now I'm hitting the **`memmem`** issue: **`undefined reference to memmem.`** Yep, I know this is a dirty hack and that it does not even fix the build. That's why I did not include it in this PR.

waddlesplash:
OK, so should this be closed then?

fbrosson:
After all, maybe we should merge it anyway, to make it easier for other contributors to dig into the test issue with the latest release.

PR #721, opened 2016-07-17
lighttpd: bump to 1.4.40, drop x86_gcc2, add x86 secondary arch.

fbrosson:
* Configure with more options: **`--with-libxml --with-ldap --with-lua --with-sqlite --with-uuid --with-webdav-locks`**
* Create a read-only copy of **`system/settings/lighttpd/`** in **`system/data/lighttpd/template-settings/`**.
* Rename **`index.shtml`** to **`index.lighttpd.shtml`** to make it easier to share the same document-root with other web servers.
* Add **`quickstart.txt`** in **`$docDir`** with some useful information.


waddlesplash:
Let's not bother with GCC2 patches (in general - there's no reason to have them for most software; GCC2 will go away after R1 anyway.) lighthttpd in particular probably benefits from optimizations that GCC4/5 have that GCC2 does not.

fbrosson:
I'm OK with dropping the gcc2 patches if that's OK to add x86 as secondary arch.
Otherwise I think we should keep the gcc2 patches until R1 is out, because annoyance of having the gcc2 patches in the repo is less than that of not having lighttpd on x86_gcc2.
What do you think? Shall I drop the gcc2 patches and add x86 as secondary arch?

waddlesplash:
>What do you think? Shall I drop the gcc2 patches and add x86 as secondary arch?

Yup, that's the way to go. No need to waste effort on GCC2.

waddlesplash:
Bump. Once the GCC2 patch is removed, this is ready to go...

fbrosson:
Yep, I also came to the conclusion that it was finaly better to not add x86 secondary arch and, instead, just merge the gcc2 patches into a single file that we'll remove once x86_gcc2 support is dropped.

waddlesplash:
Uh, that's not what I said at all. I said to drop the x86_gcc2 patch. So you should install the binaries in $prefix/bin unconditionally and add x86 as a secondary architecture. And unless there is a specific reason otherwise, always do that for both applications and libraries that do not work out of the box on gcc2.

fbrosson:
Sorry, I though you had changed your mind like I did.
Since the gcc2 patches are already done and are simple and small, converting the recipe to allow building for x86 secondary arch would be just a loss of time.
I have consolidated the 3 gcc2 patches into a single file. When x86_gcc2 becomes deprecated the only change in the recipe will be to drop the reference to the consolidated gcc2 patch.
If we converted the recipe for x86 secondary arch, then it would become useless when x86_gcc2 disappears. That would mean we would have to revert the changes for secondary_architectures.
Regarding the optimizations that recent versions of gcc have, anyone who chose x86_gcc2 can switch to x86 if such optimizations are sought.

waddlesplash:
> Since the gcc2 patches are already done and are simple and small, converting the recipe to allow building for x86 secondary arch would be just a loss of time.

Yes, a small one, but since there will be no GCC2 patch it reduces clutter. Please do it for this recipe, and for recipes in the future you can skip the GCC2 patch altogether.

> If we converted the recipe for x86 secondary arch, then it would become useless when x86_gcc2 disappears. That would mean we would have to revert the changes for secondary_architectures.

What? No. When GCC2 goes away, then "x86" (primary arch) will be the standard, and then "x86" (secondary arch) will just be redundant. The recipe won't break, assuming it already works on x86_gcc4(hybrid) builds.

>  Regarding the optimizations that recent versions of gcc have, anyone who chose x86_gcc2 can switch to x86 if such optimizations are sought.

That's not an argument for keeping GCC2...

waddlesplash:
(For future reference, in case we ever need to come back to this: the main libtool discussion happened in #665.)

fbrosson:
I've converted the recipe yesterday to make it build on x86 secondary arch, as requested.
I've also added **`template`** options in USER_SETTINGS_FILES.
I wrote a **`quickstart.txt`**. I don't know if that's OK. Feedback welcome (as usual).
I tested the recipe and everything works fine, so I guess it can be merged. I'm not merging it myself in case someone would like to suggest any changes, e.g. in the wording (or removal) of the **`quickstart.txt`** file.

waddlesplash:
Looks good to me.

fbrosson:
Cool, thanks!

PR #742, opened 2016-07-27
Add recipe for curlftpfs

fbrosson:
Recipe and patch attached. This goes in net-fs/curlftpfs

I know there are some small scale issues with the way dependencies are listed, but I don't have more time for this now so I wanted to get it submitted before it rots (it does work).

[curlftpfs-0.9.2.recipe.txt](https://github.com/haikuports/haikuports/files/384841/curlftpfs-0.9.2.recipe.txt)
[curlftpfs-0.9.2.patchset.txt](https://github.com/haikuports/haikuports/files/384840/curlftpfs-0.9.2.patchset.txt)

### DIFF
--- a/net-fs/curlftpfs/curlftpfs-0.9.2.recipe
+++ b/net-fs/curlftpfs/curlftpfs-0.9.2.recipe
@@ -1,0 +1,83 @@
+SUMMARY="Filesystem for accessing FTP hosts"
+DESCRIPTION="A file system for accessing ftp hosts based on FUSE and \
+libcurl.
+As the FTP protocol is not very feature rich, this filesystem does not \
+fulfill every constraint of a real filesystem, but it should be usable \
+for simple tasks like copying and editing files.
+
+CurlFtpFS differentiates itself from other FTP filesystems because it \
+features:
+
+* SSLv3 and TLSv1 support
+* connecting through tunneling HTTP proxies
+* automatically reconnection if the server times out
+* transform absolute symlinks to point back into the ftp file system
+
+Instructions for using CurlFtpFS on Haiku are available in \
+/system/documentation/packages/curlftpfs/readme.txt."
+HOMEPAGE="http://curlftpfs.sourceforge.net/"
+COPYRIGHT="2006-2008 Robson Braga Araujo"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/curlftpfs/curlftpfs-$portVersion.tar.gz"
+CHECKSUM_SHA256="4eb44739c7078ba0edde177bdd266c4cfb7c621075f47f64c85a06b12b3c6958"
+PATCHES="curlftpfs-$portVersion.patchset"
+ADDITIONAL_FILES="readme.txt"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	curlftpfs = $portVersion
+	addon:curlftpfs
+	cmd:curlftpfs
+	"
+REQUIRES="
+	haiku
+	userland_fs
+	lib:libglib_2.0
+	lib:libcurl
+	lib:libintl
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	userland_fs
+	devel:libglib_2.0
+	devel:libcurl
+	devel:libintl
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
+	cmd:gcc
+	cmd:gettext
+	cmd:ld
+	cmd:libtoolize
+	cmd:make
+	cmd:pkg_config
+	"
+
+BUILD()
+{
+	autoreconf -i
+	export FUSE_CFLAGS="-D_FILE_OFFSET_BITS=64 -I/system/develop/headers/userlandfs/fuse -DB_USE_POSITIVE_POSIX_ERRORS"
+	export FUSE_LIBS="-L/system/lib -luserlandfs_fuse -lposix_error_mapper"
+	export ac_cv_func_malloc_0_nonnull=yes
+	export ac_cv_func_realloc_0_nonnull=yes
+	runConfigure ./configure
+	make
+}
+
+INSTALL()
+{
+	install -d -m 755 \
+		$addOnsDir/userlandfs \
+		$manDir/man1 \
+		$docDir
+	install -t $addOnsDir/userlandfs \
+		curlftpfs
+	install -t $manDir/man1 \
+		doc/curlftpfs.1
+	install -t $docDir \
+		$portDir/additional-files/readme.txt
+}
--- a/net-fs/curlftpfs/additional-files/readme.txt
+++ b/net-fs/curlftpfs/additional-files/readme.txt
@@ -1,0 +1,16 @@
+CurlFtpFS needs a file system wrapper called UserlandFS which is available as an optional package on Haiku for most primary architectures from the corresponding Haiku repository. That package provides FUSE capability, which is needed by CurlFtpFS.
+
+If /system/servers/userlandfs_server is missing on your system then this means the userland_fs package is not installed. You can install that package using the HaikuDepot application or by typing the following command on a Terminal window:
+pkgman install userland_fs
+
+At the moment the integration of CurlFtpFS is either not finished or not fully tested.
+This means you might need to follow these steps to use it:
+
+First start the userland_fs server:
+	/system/servers/userlandfs_server curlftpfs &
+
+Then mount an ftp share:
+mount -t userlandfs -p 'curlftpfs -o option1,option2 \
+	ftp://someserver.somewhere/some/directory' /local/mount/directory
+
+curlftpfs has a manual page which lists all possible options, including support for SSLv3, TLSv1, proxies and many more.
--- a/net-fs/curlftpfs/patches/curlftpfs-0.9.2.patchset
+++ b/net-fs/curlftpfs/patches/curlftpfs-0.9.2.patchset
@@ -1,0 +1,634 @@
+From a814296060607ab57faaec270d92a0de926319ac Mon Sep 17 00:00:00 2001
+From: Ed Robbins <edd.robbins@gmail.com>
+Date: Thu, 21 Jul 2016 18:24:37 +0000
+Subject: Use off_t instead of __off_t
+
+__off_t is an internal C library type, which shouldn't be used by
+applications. It doesn't work on haiku, so use off_t instead
+
+diff --git a/ftpfs.c b/ftpfs.c
+index 0346354..ac95c55 100644
+--- a/ftpfs.c
++++ b/ftpfs.c
+@@ -687,7 +687,7 @@ static int test_exists(const char* path)
+ 	return ftpfs_getattr(path, &sbuf);
+ }
+ 
+-static __off_t test_size(const char* path)
++static off_t test_size(const char* path)
+ {
+ 	struct stat sbuf;
+ 	int err = ftpfs_getattr(path, &sbuf);
+@@ -950,7 +950,7 @@ static int ftpfs_truncate(const char* path, off_t offset) {
+ 
+   /* fix openoffice problem, truncating exactly to file length */
+   
+-  __off_t size = (long long int)test_size(path); 
++  off_t size = (long long int)test_size(path); 
+   DEBUG(1, "ftpfs_truncate: %s check filesize=%lld\n", path, (long long int)size);
+   
+   if (offset == size)  
+@@ -978,7 +978,7 @@ static int ftpfs_ftruncate(const char * path , off_t offset, struct fuse_file_in
+   }
+   /* fix openoffice problem, truncating exactly to file length */
+   
+-  __off_t size = test_size(path); 
++  off_t size = test_size(path); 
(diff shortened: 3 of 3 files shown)


net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+        gettext${secondaryArchSuffix}_libintl
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:autoreconf
+	cmd:make
+	cmd:aclocal
+	cmd:libtoolize$secondaryArchSuffix
+	pkgconfig$secondaryArchSuffix
```
korli: Should be cmd:pkg_config$secondaryArchSuffix

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+        glib2$secondaryArchSuffix >= 2
+        lib:libcurl$secondaryArchSuffix
+        gettext$secondaryArchSuffix
+        gettext${secondaryArchSuffix}_libintl
+        userland_fs
+        "
+
+BUILD_REQUIRES="
+        haiku${secondaryArchSuffix}_devel
+        glib2${secondaryArchSuffix}_devel
```
korli: Should be devel:libglib_2.0$secondaryArchSuffix

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+        curlftpfs$secondaryArchSuffix = $portVersion
+	addon:curlftpfs$secondaryArchSuffix
+        cmd:curlftpfs$secondaryArchSuffix
+        "
+REQUIRES="
+        haiku$secondaryArchSuffix
+        glib2$secondaryArchSuffix >= 2
```
korli: Should be lib:libglib_2.0$secondaryArchSuffix

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+        userland_fs
+        "
+
+BUILD_REQUIRES="
+        haiku${secondaryArchSuffix}_devel
+        glib2${secondaryArchSuffix}_devel
+        userland_fs
+        devel:libcurl$secondaryArchSuffix
+        gettext$secondaryArchSuffix
+        gettext${secondaryArchSuffix}_libintl
```
korli: Should be devel:libintl$secondaryArchSuffix

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+        curlftpfs$secondaryArchSuffix = $portVersion
+	addon:curlftpfs$secondaryArchSuffix
+        cmd:curlftpfs$secondaryArchSuffix
+        "
+REQUIRES="
+        haiku$secondaryArchSuffix
+        glib2$secondaryArchSuffix >= 2
+        lib:libcurl$secondaryArchSuffix
+        gettext$secondaryArchSuffix
+        gettext${secondaryArchSuffix}_libintl
```
korli: This is not needed.

fbrosson: Thanks @korli for your feedback.
I knew my initial changes were not enough so I was going to ask for feedback... and there it is, before I even start asking ;-)
So that's cool, because now I just need to apply your changes!

fbrosson:
Builds and works fine on x86_gcc2 ;-)
Gonna test on x86 and x86_64.
Any suggestions regarding the DESCRIPTION (or anything else) ?
BTW, I guess I should change COPYRIGHT to just "2007-2008 Robson Braga Araujo"

humdingerb:
Looks like development started in 2006. A short look at the website only shows v0.9.1 in 2007. If v0.9.2 was released 2008, I guess COPYRIGHT is "2006-2008 Robson Braga Araujo".

There's some more info on the site and the readme. I suggest:

SUMMARY="A filesystem for accessing FTP hosts"
DESCRIPTION="A file system for accessing ftp hosts based on FUSE and libcurl.
As the FTP protocol is not very feature rich, this filesystem does not \
fulfill every constraint of a real filesystem, but it should be usable \
for simple tasks like copying and editing files.

CurlFtpFS differentiates itself from other FTP filesystems because it features:
- SSLv3 and TLSv1 support
- connecting through tunneling HTTP proxies
- automatically reconnection if the server times out
- transform absolute symlinks to point back into the ftp file system"

ed-dismantle:
Can I suggest that the description keeps the usage information? The process of starting the userlandfs_server first, as well as exactly how to mount, is really not intuitive (or documented elsewhere) so otherwise people wont know how to use it.

Thanks for acting on this so fast and making the necessary fixes!

fbrosson:
Thanks @humdingerb.
I did not dare to remove the Usage paragraph. ~~Should I?~~

I haven't found **`userland_fs`** in  x86. Is it available on x86 and x86_64?

ed-dismantle:
BTW, it does build with gcc4, but I haven't tested it as I don't currently have a gcc4 or gcc4h install

fbrosson:
@ed-dismantle: I'm glad I took this PR, because curlftpfs is cool stuff! So thanks for the gcc2 patch and the initial recipe.
BTW, on x86 nothing provides userland_fs. I'll reboot on x86_gcc2 to run a **`pkgman search userland_fs`**

@korli, @humdingerb, @ed-dismantle: Should I drop the support for x86 as _secondary_ architecure and remove all occurences of **`$secondaryArchSuffix`** ?

ed-dismantle:
Thinking about it, maybe the man page should be patched to add the haiku specific usage as well?

Since the package does actually build fine with gcc4/5, perhaps we should leave the secondary arch (or change to ?) and raise an issue on haiku trac to get the userlandfs package built in gcc4 builds? (IIRC userlandfs is built as part of the main haiku image but put in a separate hpkg distributed with the main install).

humdingerb:
The recipe I downloaded didn't have a usage in the description... But by all means, do included if you have it.
If it builds with gcc2 and gcc4 there's no reason to keep the $secondaryArchSuffix AFAIK.

ed-dismantle:
Oh yeah, it isn't possible to install it as a secondary architecture because you only get a userland fs package for the primary architecture. Perhaps that's the confusion?

@humdingerb I changed the description shortly after posting last night (within 5-10 mins), as I realised no one would know how to use it. You must have downloaded it before that

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+SOURCE_URI="http://downloads.sourceforge.net/curlftpfs/curlftpfs-$portVersion.tar.gz"
+CHECKSUM_SHA256="4eb44739c7078ba0edde177bdd266c4cfb7c621075f47f64c85a06b12b3c6958"
+PATCHES="curlftpfs-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+        curlftpfs$secondaryArchSuffix = $portVersion
+	addon:curlftpfs$secondaryArchSuffix
```
korli: This line isn't correctly aligned.

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+        cmd:curlftpfs$secondaryArchSuffix
+        "
+REQUIRES="
+        haiku$secondaryArchSuffix
+        lib:libglib_2.0$secondaryArchSuffix
+        lib:libcurl$secondaryArchSuffix
+        gettext$secondaryArchSuffix
```
korli: gettext shouldn't needed.

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+        userland_fs
+        "
+
+BUILD_REQUIRES="
+        haiku${secondaryArchSuffix}_devel
+        devel:libglib_2.0$secondaryArchSuffix
+        userland_fs
+        devel:libcurl$secondaryArchSuffix
+        gettext$secondaryArchSuffix
+        gettext${secondaryArchSuffix}_libintl
```
korli: devel:libintl$secondaryArchSuffix should be enough, cmd:gettext$secondaryArchSuffix might go in BUILD_PREREQUIRES (if any).

fbrosson: Thanks again @korly for your feedback (and sorry for this late reply.)
I noticed that **`/system/servers/userlandfs_server`** is onlt available on x86_gcc2, so I marked x86 and x86_64 as not supported. Is it OK or would it be better to just have **`ARCHITECTURES="x86_gcc2"`** ?

BTW, I tried to build on x86 using x86_gcc2 as secondary arch but nothing provides userland_fs_x86_gcc2, so the only platform on which this recipe is usable is x86_gcc2 primary arch.

So shall I drop the **`SECONDARY_ARCHITECTURES`** and all the **`$secondaryArchSuffix`** ?

korli: The package UserlandFS is only available for the primary arch, yes. Dropping SECONDARY_ARCHITECTURES makes sense.

fbrosson: OK. Recipe converted to primary arch only.
BTW, I found the userland_fs package for x86 and x86_64 ;-)

net-fs/curlftpfs/curlftpfs-0.9.2.recipe
```diff
@@ -0,0 +1,82 @@
...
+CurlFtpFS differentiates itself from other FTP filesystems because it \
+features:
+
+* SSLv3 and TLSv1 support
+* connecting through tunneling HTTP proxies
+* automatically reconnection if the server times out
+* transform absolute symlinks to point back into the ftp file system
+
+Usage on Haiku: First start the userland_fs server:
+	/system/servers/userlandfs_server curlftpfs &
```
korli: IMO this should be a launch_daemon task, once curlftpfs is installed. Something as in https://github.com/haikuports/haikuports/blob/master/haiku-apps/lnlauncher/lnlauncher-1.1.2.recipe#L48

ed-dismantle: The intended behaviour is that the kernel automatically invokes the server (if not already running) when you mount (see https://www.haiku-os.org/documents/dev/running_userlandfs). I don't know what mechanism this uses, but it is broken (see https://dev.haiku-os.org/ticket/11750#comment:4 and http://lists.ports.haiku-files.org/pipermail/haikuports-devs-ports.haiku-files.org/2012-June/003271.html). Also, if you want to use any interactive options to your userlandfs plugin (like prompt for a password) you need to pass the options to uselandfs_server instead of when invoking.

pulkomandy: The kernel automatically runs the server as needed, and that works in the normal case (userlandfs package installed in /system). The workarounds detailed in that ticket are for handling the case where the userlandfs server is installed in /home.

What makes you think it is broken in the normal case? What are the errors given?
The best solution would be, of course, to fix the bugs, instead of hacking around them with launch daemon scripts or instructions in package descriptions.

ed-dismantle: Nothing happens, no errors, but nothing gets mounted! This workaround seems fine to me until it is fixed. I will try to investigate further at some point but cannot in the short term.

korli: I see a problem in describing this workaround as "Usage on Haiku", in fact I was myself misleaded to think it is the normal usecase of userlandfs_server.

ed-dismantle: Perhaps it should say "workaround on haiku". userlandfs is not very well documented (for a user or developer), and my point is only that, without the user knowing that they must do such a thing, they aren't going to be able to use the package at all. The usage is still different though, as on other OSs you can actually invoke curlftpfs directly, so the normal docs are actually wrong. Perhaps this recipe should be held until I have time to look into this further.

fbrosson: I've moved the "instructions" to a readme.txt in $docDir and added a small introduction for it. Please let me if that's OK and if anything can be improved (e.g. english style or grammar).
I've also converted the INSTALL to use **`install`** instead of **`mkdir`** and **`cp`**. It works fine, but I don't know if everyone prefers the "new" INSTALL ;-)
I think the recipe is good enough for being merged. IMO the readme.txt is both usable and accurate regarding the WIP or "workaround" status of the launch method.

fbrosson: I've just made minor changes to the last line in the readme.txt.

PR #745, opened 2016-07-29
rocksndiamonds new recipe for 4.0.0.1

Begasus:
didn't include the patch from previous version as they changed the source up to a point where I can't find the changes from the previous patchset, all config files are stored/created in ~/home/.rocksndiamonds

### DIFF
--- a/games-arcade/rocksndiamonds/rocksndiamonds-4.0.0.1.recipe
+++ b/games-arcade/rocksndiamonds/rocksndiamonds-4.0.0.1.recipe
@@ -1,0 +1,75 @@
+SUMMARY="Arcade style game in the style of Boulder Dash"
+DESCRIPTION="
+- network multiplayer games (upto 4 players)
+- local multiplayer games (upto 4 players)
+- soft scrolling with 50 frames per second
+- freely customizable keyboard and joystick support
+- stereo sound effects and music
+- music modules and fullscreen in SDL version
+- contains levels to play Boulder Dash, Emerald Mine and Sokoban
+- lots of additional levels available (over 10.000)"
+HOMEPAGE="http://www.artsoft.org/rocksndiamonds/"
+COPYRIGHT="2001-2017 Artsoft Entertainment"
+LICENSE="GNU GPL v2"
+REVISION="1"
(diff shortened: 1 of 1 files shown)


pulkomandy:
That's not really acceptable, the patch for fixing the settings dir should be redone.

Begasus:
Couldn't find the precense of the older commitment ...

games-arcade/rocksndiamonds/patches/rocksndiamonds-4.0.0.0_rc3.patchset
```diff
@@ -0,0 +1,34 @@
...
++
+ ifeq ($(PLATFORM),macosx)		# compiling on Mac OS X
+ EXTRA_LDFLAGS = -lstdc++
+ override PLATFORM = unix
+@@ -123,7 +128,7 @@ OPTIONS = -O3 -Wall
+ endif
+ 
+ CFLAGS = $(OPTIONS) $(SYS_CFLAGS)  $(EXTRA_CFLAGS) $(CONFIG)
+-LDFLAGS =           $(SYS_LDFLAGS) $(EXTRA_LDFLAGS) -lm
++LDFLAGS =           $(SYS_LDFLAGS) $(EXTRA_LDFLAGS)
```
korli: To be upstreamed, you should better define by default MATHLIBS to -lm, and set it to empty for Haiku. Then use MATHLIBS  in LDFLAGS.

Begasus: updated the patchset with your comment, does this look ok?

korli:
Please adjust the recipe name to rocksndiamonds-4.0.0.0~rc3.recipe

Begasus:
Adjusted :)

fbrosson:
@Begasus, I think it would be perfectly ok to "recycle" this PR and update the recipe for 4.0.0.0  ;-)
Easier to say than to do, I have to admit ;-)
Good luck, btw!

Begasus:
with some modifications (like the version in the recipe changing into $portVersion) and the name of the recipe it builds without patching, just need to create a patch so that the files will get stored in the correct folder ...

Begasus:
dismissed the patch (for now, untill there is a fix for the save folder), builds and runs ok :), also bumped to the latest version

PR #749, opened 2016-08-01
lcms: bump to 2.8, drop liblcms2.la, reorder sections, add TEST.

fbrosson:
* Don't use **`$secondaryArchSuffix`** for **`cmd:{jpg,link,ps,tif,trans}icc`** and install them in **`$prefix/bin`** (instead of **`$binDir`**) for x86 secondary arch. This is safe (and needs no external changes) because no other recipe references any **`cmd:`** provided by this one.
* Use **`{aclocal,automake}`** from automake (instead of automake113) since upstream now uses that version. Thanks @korli for the hint!


media-libs/lcms/lcms-2.8.recipe
```diff
@@ -1,31 +1,35 @@
...
 
 PROVIDES="
 	lcms$secondaryArchSuffix = $portVersion compat >= 2
-	lib:liblcms2$secondaryArchSuffix = 2.0.7 compat >= 2
-	cmd:jpgicc$secondaryArchSuffix
-	cmd:linkicc$secondaryArchSuffix
-	cmd:psicc$secondaryArchSuffix
-	cmd:tificc$secondaryArchSuffix
-	cmd:transicc$secondaryArchSuffix
+	lib:liblcms2$secondaryArchSuffix = $portVersion compat >= 2
```
korli: $portVersion isn't equal to the soname.

media-libs/lcms/lcms-2.8.recipe
```diff
@@ -34,6 +38,19 @@ REQUIRES="
 	lib:libtiff$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
+
+PROVIDES_devel="
+	lcms${secondaryArchSuffix}_devel = $portVersion
+	devel:liblcms2$secondaryArchSuffix = $portVersion compat >= 2
```
korli: Ditto

media-libs/lcms/lcms-2.8.recipe
```diff
@@ -42,32 +59,32 @@ BUILD_REQUIRES="
...
 BUILD_PREREQUIRES="
-	cmd:gcc${secondaryArchSuffix}
-	cmd:ld${secondaryArchSuffix}
-	cmd:libtoolize
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:aclocal_1.13
 	cmd:autoconf
 	cmd:automake_1.13
```
korli: It seems they updated to a newer automake, it might work with cmd:automake also.

media-libs/lcms/lcms-2.8.recipe
fbrosson: Good catch, thanks!

media-libs/lcms/lcms-2.8.recipe
fbrosson: Yep, it does ;-) Excellent idea. Updated recipe. Thanks!

PR #751, opened 2016-08-04
rubberband: fix x86 2nd arch and simplify install a little.

fbrosson:


fbrosson:
Regarding the **`devel:lib*`** I'm adding in REQUIRES_devel, I don't know if this is correct.
(This means I'll revert that part if we don't need them ;-)

korli:
Adding the devel:lib* is not needed, because the rubberband headers do no reference headers of ffw3, libsamplerate, etc. Also no external libs in the Libs attribute of the pkgconfig file.

fbrosson:
OK, thanks. I'm dropping the **`devel:lib*`** then.

PR #754, opened 2016-08-07
swars, new recips

Begasus:
Game crashes on exit, fullscreen was working but at some point frooze my system ... best to play it in windowed mode (default)


Begasus:
Credits to munchausen for providing the files :)

PR #755, opened 2016-08-07
libpng16: bring back libpng*.la because fontconfig needs it.

fbrosson:


fbrosson:
As explained in #756, we don't need to add back **`libpng*.la`** to make fontconfig build. It is much better to drop **`libfreetype.la`** in freetype.
Closing this PR, as the fix is now in #757.

PR #756, opened 2016-08-07
fontconfig: bump to 2.12.1, reorder sections, sync patch.

fbrosson:
* Convert the **`settings/fonts/conf.d/*.conf`** absolute symlinks into (relative) symlinks to **`../../data/fontconfig/conf.avail/`** instead of **`/packages/fontconfig-2.12.1-1/.self/data/fontconfig/conf.avail/`**.
* Switch HOMEPAGE and SOURCE_URI to https.

### DIFF
--- a/media-libs/fontconfig/fontconfig-2.12.1.recipe
+++ b/media-libs/fontconfig/fontconfig-2.12.1.recipe
@@ -1,19 +1,17 @@
 SUMMARY="A library for font customization and configuration"
-DESCRIPTION="
-Fontconfig is a library for font customization and configuration.
-"
-HOMEPAGE="http://www.freedesktop.org/wiki/Software/fontconfig"
-SOURCE_URI="http://www.freedesktop.org/software/fontconfig/release/fontconfig-$portVersion.tar.bz2"
-CHECKSUM_SHA256="b433e4efff1f68fdd8aac221ed1df3ff1e580ffedbada020a703fe64017d8224"
-LICENSE="MIT"
+DESCRIPTION="Fontconfig is a library for font customization and configuration."
+HOMEPAGE="https://www.freedesktop.org/wiki/Software/fontconfig"
 COPYRIGHT="2000-2005, 2006-2007 Keith Packard
 	2005 Patrick Lam
 	2009 Roozbeh Pournader
 	2008-2009 Red Hat, Inc.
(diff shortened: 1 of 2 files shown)


fbrosson:
I forgot to mention that IMO #755 should be merged before this one, otherwise fontconfig fails to build. Here's the error I get if I try to build fontconfig with a libpng16_devel that does not have the libpng16.la file:
```
  CCLD     libfontconfig.la
/bin/grep: /packages/freetype_devel-2.6.3-1/devel~libpng16/develop/lib/libpng16.la: No such file or directory
/bin/sed: can't read /packages/freetype_devel-2.6.3-1/devel~libpng16/develop/lib/libpng16.la: No such file or directory
libtool: link: `/packages/freetype_devel-2.6.3-1/devel~libpng16/develop/lib/libpng16.la' is not a valid libtool archive
make[3]: *** [Makefile:567: libfontconfig.la] Error 1
make[3]: Leaving directory '/sources/fontconfig-2.12.1/src'
make[2]: *** [Makefile:497: all] Error 2
make[2]: Leaving directory '/sources/fontconfig-2.12.1/src'
make[1]: *** [Makefile:580: all-recursive] Error 1
make[1]: Leaving directory '/sources/fontconfig-2.12.1'
make: *** [Makefile:464: all] Error 2
```
Since bringing back **`libpng16.la`** fixes the build, I guess it will be OK to merge #755, but I'll let that one open in case someone has a better workaround for building fontconfig without **`libpng16.la`**.

korli:
I guess this happens because freetype provides a libtool file with a reference to the libpng16.la. Removing the freetype libtool file might also work.

fbrosson:
Exact. I've just verified it. Dropping libfreetype.la also fixes the problem and allows to leave libpng16 as is, i.e. without its libpng*.la.

I could have replaced the commit for freetype into #755 using its current "libpng" branch ( and it would have worked), but it was more exciting to try and attach a "freetype" branch into that same PR, so I used curl to do so. But the API complained, telling it needed a title. I though it was a security to avoid unwanted renamngs of branchs in PRs, but it seems that once a branch name has been associated with a PR there is no way to change its name... I ended with a new PR, #757. I'll close #755.
(Maybe I could have tried renaming the "libpng" branch from within the web-based interface, and then force a push into it.)

PR #758, opened 2016-08-09
torsocks bump version, fix lib dependency

Begasus:
re-implemented the patches from previous version

### DIFF
--- a/net-proxy/torsocks/torsocks-1.3.recipe
+++ b/net-proxy/torsocks/torsocks-1.3.recipe
@@ -1,0 +1,70 @@
+SUMMARY="A transparent socks proxy for use with tor"
+DESCRIPTION="Torsocks allows you to use most socks-friendly applications \
+in a safe way with Tor. It ensures that DNS requests are handled safely \
+and explicitly rejects UDP traffic from the application you're using."
+HOMEPAGE="https://www.torproject.org"
+COPYRIGHT="2000-2008 Shaun Clowes
+	2008-2011 Robert Hogan
+	2005 Total Information Security Ltd.
+	2010 Alex Rosenberg
+	2004, 2006 Peter Palfrader
+	2000 Alessandro Iurlano
+	2004 Tomasz Kojm"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/dgoulet/torsocks/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="09debf9e35d70802fdb0e56335bacab65e84fad615f68d7006539713c1698123"
+PATCHES="torsocks-$portVersion.patchset"
+GLOBAL_WRITABLE_FILES="settings/torsocks.conf manual"
+
+ARCHITECTURES="x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	torsocks$secondaryArchSuffix = $portVersion
+	cmd:torsocks$secondaryArchSuffix
+	cmd:usewithtor$secondaryArchSuffix
+	lib:libtorsocks$secondaryArchSuffix = 1.0.0 compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	torsocks${secondaryArchSuffix}_devel = $portVersion compat >= 1.0
+	devel:libtorsocks$secondaryArchSuffix = 1.0.0 compat >= 0
+	"
+REQUIRES_devel="
+	torsocks$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:autoreconf
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf -vfi
+	runConfigure ./configure --datadir=$dataDir/torsocks/
+	make
+}
+
+INSTALL()
+{
+	make install
+
+	rm $libDir/*.la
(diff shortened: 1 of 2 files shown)


net-proxy/torsocks/torsocks-1.3.recipe
```diff
@@ -0,0 +1,74 @@
...
+	make install
+
+	rm $libDir/*.la
+	prepareInstalledDevelLibs libtorsocks
+
+	packageEntries devel \
+			$developDir
+
+	#link the .so lib so torsocks can find it in $libDir
+	ln -s $libDir/libtorsocks.so.1.0.0 $libDir/libtorsocks.so
```
fbrosson: I think it would be slightly better to _also_ pass **`-r`** (or **`--relative`**) to **`ln`** to create a relative symlink instead of an absolute one.

net-proxy/torsocks/torsocks-1.3.recipe
```diff
@@ -0,0 +1,74 @@
...
+	runConfigure ./configure --datadir=$dataDir/torsocks/
+	make
+}
+
+INSTALL()
+{
+	make install
+
+	rm $libDir/*.la
+	prepareInstalledDevelLibs libtorsocks
```
fbrosson: Since there is only one lib in this recipe, it _might_ be desirable to call prepareInstalledDevelLib (instead of prepareInstalledDevelLib**s**), although many other "single-lib" recipes call prepareInstalledDevelLib**s**.
I don't know, however, if there is any recommendation for using prepareInstalledDevelLib (instead of prepareInstalledDevelLib**s**) for "single-lib" recipes.
So please feel free to ignore this suggestion ;-)

net-proxy/torsocks/torsocks-1.3.recipe
```diff
@@ -0,0 +1,74 @@
...
+HOMEPAGE="https://www.torproject.org"
+COPYRIGHT="
+	2000-2008 Shaun Clowes
+	2008-2011 Robert Hogan
+	2005 Total Information Security Ltd.
+	2010 Alex Rosenberg
+	2004, 2006 Peter Palfrader
+	2000 Alessandro Iurlano
+	2004 Tomasz Kojm
+	"
```
fbrosson: You migh wish to drop 2 newline to have this:
```
COPYRIGHT="2000-2008 Shaun Clowes
	2008-2011 Robert Hogan
	2005 Total Information Security Ltd.
	2010 Alex Rosenberg
	2004, 2006 Peter Palfrader
	2000 Alessandro Iurlano
	2004 Tomasz Kojm"
```

net-proxy/torsocks/torsocks-1.3.recipe
Begasus: ln: cannot do --relative without --symbolic when using "ln -r" ?

Begasus: PS, thanks ... was looking for that one! :)

net-proxy/torsocks/torsocks-1.3.recipe
Begasus: Didn't know about this one ... changing ....

korli:
The latest version is 2.1. Any reason not to use that one?

Begasus:
I have been looking into it, but it failed (iirc) on not finding sys/sockets.h(?) which is not available in Haiku ... this was one that still could be build without hacking to much in the source I think ...

Begasus:
removed the check for syscall in configure.ac and when it starts building it fails out now on ... 

make[2]: Entering directory '/sources/torsocks-2.1.0/src/common'
  CC       log.lo
cpp0: config.h: No such file or directory
Makefile:389: recipe for target 'log.lo' failed
Haven't found out where to change this one (yet)

net-proxy/torsocks/torsocks-1.3.recipe
```diff
@@ -0,0 +1,72 @@
...
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	torsocks$secondaryArchSuffix = $portVersion
+	cmd:torsocks$secondaryArchSuffix
+	cmd:usewithtor$secondaryArchSuffix
+	lib:libtorsocks$secondaryArchSuffix = 1.0.0 compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix >= $haikuVersion
```
fbrosson: You can (or should?) drop the «**` >= $haikuVersion`**» for «**`haiku$secondaryArchSuffix`**» and «**`haiku${secondaryArchSuffix}_devel`**».

net-proxy/torsocks/torsocks-1.3.recipe
```diff
@@ -0,0 +1,72 @@
...
+REQUIRES_devel="
+	torsocks$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel >= $haikuVersion
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize
```
fbrosson: **`cmd:libtoolize`** is safer with a **`$secondaryArchSuffix`**.

net-proxy/torsocks/torsocks-1.3.recipe
```diff
@@ -0,0 +1,72 @@
...
+	cmd:aclocal
+	cmd:autoconf
```
fbrosson: Even if **`cmd:autoreconf`** and **`cmd:autoconf`** are provided by the same package, which means that having **`cmd:autoconf`** in BUILD_PREREQUIRES is enough to pull autoreconf, I think it is recommended to add a **`cmd:autoreconf`**, since BUILD calls autoreconf.
I don't know, however, if this point of view is widely shared ;-)

waddlesplash: Yes, adding cmd:autoreconf is recommended.

Begasus:
Updated

PR #759, opened 2016-08-09
ffmpeg: bump to 3.1.2.

fbrosson:


fbrosson:
**Work in progress**, so please do not merge yet (unless you finish testing before me, of course ;-)

fbrosson:
Yep, the package looks OK to me too, although on x86 secondary arch, when I run the tests, there is a runtime which does not find a library so the tests fail almost instantly. On x86 primary arch, however, the tests seem OK, although, at some point, they fail because I did not give them enough disk space... I should have attached an extra disk with more free space... but did not. Shame on me!
So please feel free to merge ;-)

PR #762, opened 2016-08-11
hunspell: bump to 1.4.0, mark ?x86_gcc2, reorder sections.

fbrosson:
* Drop old recipes (1.3.2 and 1.3.3) and update to 1.4.0.
* Mark x86_gcc2 as untested because it is failing 7 tests out of 111 whereas x86_64 and x86 (both primary and secondary) pass them all.
* Do not drop **`$binDir`** for secondary arch anymore.
* Remove unneeded **`touch po/t-hu.gmo`** and **`touch po/t-it.gmo`**.


korli:
Just a note: "convert recipe to new format" means for you reordering sections of the recipe, right? 
Because there was an old recipe format, for which there are still recipes in the tree. I understand these can be converted, others are simply not following a recommended layout. A commit message can be misleading :)

fbrosson:
What a coincicende! Yesterday i noticed that my commit message was misleading, but postponed the update (of the commit message) because I was wondering whether I should drop the **`touch po/t-hu.gmo`** and **`touch po/t-it.gmo`** calls in BUILD.
Now the commit message is fixed.

BTW, shall I dig the question about the **`touch po/t-{hu,it}.gmo`** ?

korli:
the **`touch po/t-{hu,it}.gmo`** is probably here to avoid needing to regenerate such files (which indeed would require gettext). The problem can appear because, by unpacking, files get slightly off timestamps, which may cause undetermistic behavior at build time.

fbrosson:
OK, thanks.
I'll make a few tests to see if it the touch are still required.
BTW, I've just fixed a typo (s/reoder/reorder/) in the commit message.

fbrosson:
I've noticed that the recipe builds OK without the 2 touch commands, and that the resulting packages are identical. Well, the static lib embeds some timestamp (or random data) which makes it change on every build, but this only happens on x86_gcc2.
I've also noticed that x86_gcc2 fails on 7 tests out of 111, so I changed the recipe a little, as detailed in the commit message.
I think the recipe is now ready for the merge.
I'm not merging it myself, hwoever, because I've made several changes that probably need to be validated by someone else.

fbrosson:
Cool, thanks!

Issue #178, opened 2015-06-30
fblend is broken


soakbot:
(Original issue 178 created by scottmc on 2015-02-11T08:12:23.473721+00:00)

fblend recipe is still in old .bep file format and is marked as broken.  It is unknown as to which if any old BeOS or Allegro games/apps would even require this still.

Begasus:
Is it possible to create only a "devel" package? I got v0.4 build but I have an empty fblend-0.4-1-x86_gcc2.hpkg (and a correct) fblend_devel-0.4-1-x86_gcc2.hpkg

Begasus:
just created a PR for this one (version 0.4)

PR #764, opened 2016-08-11
fblend 0.4 recipe, working

Begasus:
if there is a sollution to just create a devel package alone I would be glad to hear it, this one is working for x86_gcc2 (with the examples from allegro)

### DIFF
--- a/media-libs/fblend/fblend-0.4.recipe
+++ b/media-libs/fblend/fblend-0.4.recipe
@@ -1,31 +1,53 @@
+SUMMARY="Color blenders for Allegro"
 DESCRIPTION="FBlend is a series of super-fast color blenders for Allegro."
-HOMEPAGE="http://sourceforge.net/projects/fblend/"
-SOURCE_URI="http://sourceforge.net/projects/fblend/files/FBlend%20beta/fblend-0.5-beta01/fblend-0.5-beta01.zip/download"
-REVISION="1"
-STATUS_HAIKU="broken"
-DEPEND="media-libs/allegro >= 4.4
-	app-text/dos2unix >= 1.0"
-CHECKSUM_MD5="e252ae521e099793fe6caa0fb7d048f3"
+HOMEPAGE="https://sourceforge.net/projects/fblend/"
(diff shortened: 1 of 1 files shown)


waddlesplash:
It only creates static libs, not dynamic ones?

Begasus:
It only creates a *.a lib yes
this one is one for the issue reported at https://github.com/haikuports/haikuports/issues/178
there is only one lib and one header file ...

waddlesplash:
Eh, just make a non-devel package, but have it provide devel:libfblend.

Begasus:
Haven't been there yet ;) looking into it ...

Begasus:
updated, not sure if it's the right way, but working still with the examples ...

media-libs/fblend/fblend-0.4.recipe
```diff
@@ -1,31 +1,53 @@
...
-SOURCE_URI="http://sourceforge.net/projects/fblend/files/FBlend%20beta/fblend-0.5-beta01/fblend-0.5-beta01.zip/download"
-REVISION="1"
-STATUS_HAIKU="broken"
-DEPEND="media-libs/allegro >= 4.4
-	app-text/dos2unix >= 1.0"
-CHECKSUM_MD5="e252ae521e099793fe6caa0fb7d048f3"
+COPYRIGHT="2001-2002 Robert J Ohannessian"
+LICENSE="BSD (3-clause)"
+REVISION="2"
+SOURCE_URI="https://sf.net/projects/fblend/files/FBlend/0.4/fblend-0.4.zip/download"
```
fbrosson: You can use the redirector service:
```
SOURCE_URI="https://downloads.sf.net/fblend/fblend-$portVersion.zip"
```
Reminder: SourceForge redirects all https downloads to http, so using https is not safer than http. But maybe SF will start serving https for their downloads too, some day.

BTW, you can switch HOMEPAGE to https (because SF does serve sourceforge.net/projects/... through https).

Begasus:
Updated ...

PR #770, opened 2016-08-16
unrar: bump to 5.4.5.

fbrosson:


fbrosson:
I'm adding a tiny **`test.rar`** in additional-files although I don't like very much to add binary files to a repo.
On the other hand, the easiest workaround would be to add it encoded with base64 and then decode it in TEST(), but that would not be much better because it would add some extra steps and would also need a **`cmd:base64`** in BUILD_PREREQUIRES.
Is it OK as is, i.e. with the tiny **`test.rar`** in additional-files?

pulkomandy:
Usually we don't write tests ourselves, we use TEST only when there is a "make check" or similar already available.
Anyway, base64 is of no use. git can handle binary files just fine and, if you use base64, it can show you a diff without special settings, but it isn't particularly readable. However, if you use the binary file directly, you can configure git to use an external tool to obtain a text rendering (for example it could run "unrar l" to list the contents) and then diff that (see the --textconv option to git diff).

fbrosson:
> Usually we don't write tests ourselves, we use TEST only when there is a "make check" or similar already available.

Well, since there are no tests available in unrar, I though that it could be worth doing something similar to what https://github.com/Homebrew does in their unrar formula (https://github.com/Homebrew/homebrew-core/blob/master/Formula/unrar.rb) but I can remove it if you prefer.

> |...] git can handle binary files just fine [...]

Yep, I know, but I was just wondering whether or not it would have been better to convert it to plain text, given that some purists, if I'm not mistaken, claim that adding binary files to source repos is not recommended.
Well, I understand from your reply that it's OK to to add binary files to HaikuPorts :-)

Comming back to unrar, shall I drop the test.rar and TEST or do you think it could be worth having such a test in this particular recipe?

BTW, I had never seen the paragraph about the **`--textconv`** option in **`man git-diff`**, so thanks for the info ;-)

pulkomandy:
I would remove it, if the upstream developers of unrar decided not to have tests, and they have bugs, it's their problem :).

As for storing binary files: my opinion is, it is better when possible to store the sources and a way to compile them. But, this is not a strict rule, and generated files can be ASCII (for example a .c file generated by bison/yacc/flax things, or HTML generated from markdown), or the sources could be a non-ASCII file in some cases (for example, a tokenized BASIC file, or a smalltalk image, or even a word document or schematics file for an electronic design tool). As long as there is a suitable tool to use with --textconv to see what is going on, this is fine.

fbrosson:
OK, fine.

PR #772, opened 2016-08-19
gnupg: bump to 1.4.21, add missing aclocal to BUILD_PREREQUIRES.

fbrosson:


fbrosson:
FWIW, Gentoo has 2 packages with the same sub-directory name: [app-crypt/gnupg](https://packages.gentoo.org/packages/app-crypt/gnupg) and [app-vim/gnupg](https://packages.gentoo.org/packages/app-vim/gnupg), which means, if I'm not mistaken, that **_if_** we wanted to add the second one we would need to either rename its PROVIDES to e.g. vim_gnupg or to make some changes to haikuporter like embedding something that tells the package is under the app-vim vs app-crypt tree/category.
I'm not planing to add app-vim/gnupg any soon, though. I'm just bringing the subject because I suspect that renaming packages (to avoid name collisions) might not be the best solution.

mmuman:
It's not the first conflict we had. I suppose the 2nd one is a vim plugin, so prefixing with vim should be fine.

fbrosson:
@mmuman, I agree that vim_gnupg should be fine. However, as you also pointed, there are other packages with name collisions. I've seen a pair a couple of weeks ago but did not write it down... and can't remember what it was! All I remember is that the name was quite short (around 5 or 6 characters long) and that it did not start with "lib", although one of the packages was a library for decoding multimedia streams and the other one was in a completely different category.
I don't know how many packages with name collisions are likely to be added to the tree, but if there are many of them then it might be worth considering a few changes in haikuporter to allow it to continue to stick to the ports tree in Gentoo. OTOH, using slightly different names (when necessary) is much simpler than having to change haikuporter. So I guess there's no need to change haikuporter for the moment ;-)

Issue #440, opened 2016-01-26
add bencode-tools


scottmc:
From old site:

bencode-tools

Reported by:    nicholas    Owned by:   scottmc
Priority:   normal  Milestone:  
Component:  dev-util/awka   Version:  
Severity:   normal  Keywords:   bencode-tools
Cc:  
Description 

Reply
bencode-tools' is a collection of C and Python tools for manipulating bencoded data. This was formerly named 'type validator'.
The collection consists of a C library for reading and writing bencoded data, and a Python implementation of bencode included with 'typevalidator' that can be used to verify typing and contents of bencoded messages. 
Attachments

bencode-tools-2011-03-15.bep​ (629 bytes) - added by nicholas 3 years ago.
Download all attachments as: .zip
Attach file
Change History

Changed 3 years ago by nicholas

attachment bencode-tools-2011-03-15.bep​ added

[bencode-tools-2011-03-15.bep.txt](https://github.com/haikuports/haikuports/files/105682/bencode-tools-2011-03-15.bep.txt)

Begasus:
https://github.com/haikuports/haikuports/pull/773

Begasus:
I gather this one can be closed? ;)

PR #773, opened 2016-08-20
bencode-tools, new recipe

Begasus:
per request for issue: https://github.com/haikuports/haikuports/issues/440

### DIFF
--- a/dev-util/bencode/bencode_tools-20110315.recipe
+++ b/dev-util/bencode/bencode_tools-20110315.recipe
@@ -1,0 +1,68 @@
+SUMMARY="C and Python tools for manipulating bencoded data"
+DESCRIPTION="bencode-tools is a collection of tools for manipulating \
+bencoded data. See http://en.wikipedia.org/wiki/Bencode for information \
+on bencode format.
+
+This was formerly named typevalidator. With the introduction of \
+libbencodetools the project is named bencode-tools."
+HOMEPAGE="http://zakalwe.fi/~shd/foss/bencode-tools/"
+COPYRIGHT="2002-2005 Bram Cohen and Ross Cohen"
+LICENSE="Public Domain
+	BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://zakalwe.fi/~shd/foss/bencode-tools/bencode-tools-2011-03-15.tar.bz2"
+CHECKSUM_SHA256="b97e37a82593d38c44448c6f5571f14e47edbb6545391fab9ffb631bbdd2abec"
+SOURCE_DIR="bencode-tools-2011-03-15"
+PATCHES="bencode_tools-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+
+PROVIDES="
+	bencode = $portVersion
+	cmd:bencat = $portVersion
+	lib:libbencodetools = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libbencodetools
+	"
+
+PROVIDES_devel="
+	bencode_devel = $portVersion
+	devel:libbencodetools = $portVersion
+	"
+REQUIRES_devel="
+	bencode == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $includeDir $binDir $libDir
+	cp -R include/* $includeDir
+
+	make install
+
+	# devel package
+	prepareInstalledDevelLibs libbencodetools
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
+}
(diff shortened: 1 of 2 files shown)


dev-util/bencode/bencode_tools-20110315.recipe
```diff
@@ -0,0 +1,69 @@
+SUMMARY="C and Python tools for manipulating bencoded data"
+DESCRIPTION="bencode-tools is a collection of tools for manipulating \
+bencoded data. See http://en.wikipedia.org/wiki/Bencode for information \
+on bencode format.
+
+This was formely named typevalidator. With the introduction of \
```
waddlesplash: should be 'formerly'

dev-util/bencode/bencode_tools-20110315.recipe
```diff
@@ -0,0 +1,69 @@
...
+libbencodetools the project is named bencode-tools."
+HOMEPAGE="http://zakalwe.fi/~shd/foss/bencode-tools/"
+COPYRIGHT="2002-2005 Bram Cohen and Ross Cohen"
+LICENSE="LICENSE.bencode"
```
waddlesplash: Call this "Bencode License". Also, are you sure that it isn't just 3-clause BSD? It looks like that at first glance to me.

dev-util/bencode/bencode_tools-20110315.recipe
```diff
@@ -0,0 +1,69 @@
...
+	cp libbencodetools.so $libDir
+	cp -R include/* $developDir
+
+	#make install /* creates a include direcory */
+
+	# devel package
+	prepareInstalledDevelLibs libbencodetools
+	packageEntries devel \
+		$developDir
+
```
waddlesplash: Extraneous whitespace

dev-util/bencode/bencode_tools-20110315.recipe
```diff
@@ -0,0 +1,69 @@
...
+}
+
+INSTALL()
+{
+	mkdir -p $developDir $binDir $libDir
+	cp bencat $binDir
+	cp libbencodetools.so $libDir
+	cp -R include/* $developDir
+
+	#make install /* creates a include direcory */
```
waddlesplash: It would be preferable to use "make install" and then move the include directory to $includeDir.

dev-util/bencode/bencode_tools-20110315.recipe
Begasus: Will change (copied it over from the site) :)

dev-util/bencode/bencode_tools-20110315.recipe
Begasus: I'm not too familiar with Licenses (this one is the one that came with the source)

dev-util/bencode/bencode_tools-20110315.recipe
Begasus: Will look into this, have been fidling with that one for a while .. no luck so far (maybe moving could do the trick)

dev-util/bencode/bencode_tools-20110315.recipe
Begasus: fixing ...

Begasus:
Not sure where or why the check is failing atm ... haven't touched the "make install" yet, building fine here ...

Begasus:
found it ;) will look into the *make install* problem a bit more, only thing is that I don't really know how to move a "include" directory into the proper place ...

Begasus:
updated ...

dev-util/bencode/bencode_tools-20110315.recipe
```diff
@@ -0,0 +1,67 @@
...
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $developDir $binDir $libDir
+	cp -R include/* $developDir
+
+	make install
```
waddlesplash: $developDir is wrong above; it should be $includeDir. Also, isn't an extraneous "include" directory created because of this make install now?

dev-util/bencode/bencode_tools-20110315.recipe
```diff
@@ -0,0 +1,67 @@
+SUMMARY="C and Python tools for manipulating bencoded data"
+DESCRIPTION="bencode-tools is a collection of tools for manipulating \
+bencoded data. See http://en.wikipedia.org/wiki/Bencode for information \
+on bencode format.
+
+This was formerly named typevalidator. With the introduction of \
+libbencodetools the project is named bencode-tools."
+HOMEPAGE="http://zakalwe.fi/~shd/foss/bencode-tools/"
+COPYRIGHT="2002-2005 Bram Cohen and Ross Cohen"
+LICENSE="Bencode"
```
waddlesplash: OK, so it is indeed the BSD 3-clause, but there is also another LICENSE file that says everything but the python script is public domain. So this should have 2 entries: Public Domain, and BSD 3-clause, and you should delete the custom license.

dev-util/bencode/bencode_tools-20110315.recipe
Begasus: The extraneous "include" directory is removed with the patchfile in "Makefile.in" Will look into the other items when I get around to it, thanks for the feedback :)

dev-util/bencode/bencode_tools-20110315.recipe
Begasus: Ok, will remove the custom one and add as mentioned :)

Begasus:
Updated

PR #786, opened 2016-08-29
New recipe: libplist

return:
libplist is a library that handles the reading and writing of the property list XML binary format used in Apple iDevices.

### DIFF
--- a/app-pda/libplist/libplist-1.12.recipe
+++ b/app-pda/libplist/libplist-1.12.recipe
@@ -1,0 +1,97 @@
+SUMMARY="A library to handle Apple Property List format in binary or XML"
+DESCRIPTION="libplist is a library for reading and writing \
+the Apple binary and XML property list format. It is part of \
+the libimobiledevice stack providing access to iDevices."
+HOMEPAGE="http://www.libimobiledevice.org/"
+COPYRIGHT="2008-2010 Jonathan Beck
+	2009-2014 Nikias Bassen
+	2009-2014 Martin Szulecki
+	2008 Matt Colyer"
+LICENSE="GNU GPL v2
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/libplist-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	libplist$secondaryArchSuffix = $portVersion compat >= 1
+	cmd:plistutil = $portVersion compat >= 1
+	lib:libplist$secondaryArchSuffix = 3.0.0 compat >= 3
+	lib:libplist++$secondaryArchSuffix = 3.0.0 compat >= 3
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libplist${secondaryArchSuffix}_devel = $portVersion compat >= 1
+	devel:libplist$secondaryArchSuffix = 3.0.0 compat >= 3
+	devel:libplist++$secondaryArchSuffix = 3.0.0 compat >= 3
+	"
+REQUIRES_devel="
+	libplist$secondaryArchSuffix == $portVersion base
+	devel:libxml2$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libxml2$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:cmake
+	cmd:python
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	if [ -n "$secondaryArchSuffix" ]; then
+		maybe_without_cython=--without-cython
+	else
+		maybe_without_cython=
+	fi
+
+	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir \
+		$maybe_without_cython
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	# remove libtool files
+	rm $libDir/lib*.la
+
+	prepareInstalledDevelLibs libplist libplist++
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+
+}
+
+TEST()
+{
+	make check
+}


dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:cmake
+	cmd:python
+	cmd:pkg_config
```
fbrosson: Should be
```
```

mmuman:
Cool.
Could you fix this: "Trailing whitespace in 'dev-libs/libplist/libplist-1.12.recipe'."
Like the tabs alone before the BUILD phase and inside it?

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+{
+	runConfigure ./configure
+	make $jobArgs
+}
+INSTALL()
+{
+	make install
+
+	# remove libtool file
+	rm $libDir/$portName++.la
```
fbrosson: You **_could_** use «**`rm $libDir/$portName++.la`**» **_if_** the recipe was **_only_** for primary architectures.
But when you build for a secondary architecture, e.g. x86, «**`$portName`**» is not «**`libplist`**» but «**`libplist_x86`**», so this means you should use:
```
	rm $libDir/libplist++.la
```
or, if you prefer,
```
	rm $libDir/lib*.la
```

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	
+	prepareInstalledDevelLibs libplist libplist++
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+		
+
```
fbrosson: There are a few unneeded tabs here, in lines 73 and 74, as well as between blocks REQUIRES and PROVIDES_devel (line 27), and between blocks REQUIRES_devel and BUILD_REQUIRES (line 36).

BTW, adding libplist is really great! Thanks!

dev-libs/libplist/libplist-1.12.recipe
fbrosson: Oops, if there is also a **`libplist.la`** (and I guess this is the case), then you will probably want to use:
```
	rm $libDir/libplist*.la
```
or
```
	rm $libDir/lib*.la
```
(Both suggestions are OK, so you can choose the one you prefer ;-)

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:cmake
+	cmd:python
+	cmd:pkg_config
+	cmd:sh
+	"
+
```
fbrosson: Hmm, another trailling tab here, just above **`BUILD()`**.

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	lib:libz$secondaryArchSuffix
+	"
+	
+PROVIDES_devel="
+	libplist${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	devel:libplist$secondaryArchSuffix = $portVersion compat >= 0
+	devel:libplist++$secondaryArchSuffix = $portVersion compat >= 0
+	"
+REQUIRES_devel="
+	libxml2$secondaryArchSuffix
```
fbrosson: You forgot the dependency on the base package.
And the dependency on libxml2 should be prefixed with a **`devel:`**.
BTW, the devel package **_might_** also need libglib2.
So, if I'm not mistaken, **`REQUIRES_devel`** should be:
```
REQUIRES_devel="
	libplist$secondaryArchSuffix == $portVersion base
	devel:libglib_2.0$secondaryArchSuffix
	devel:libxml2$secondaryArchSuffix
	"
```

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	libplist++$secondaryArchSuffix = $portVersion compat >= 0
+	cmd:plistutil$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libplist$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libplist++$secondaryArchSuffix = $portVersion compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
```
fbrosson: On Arch Linux the [libplist package](https://www.archlinux.org/packages/extra/x86_64/libplist/) needs glib2, so maybe you should add this to REQUIRES:
```
```

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/$portName-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	libplist$secondaryArchSuffix = $portVersion compat >= 0
+	libplist++$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson: For the PROVIDES block I think you don't need the line that starts with «**`libplist++`**».
Having «**`libplist$secondaryArchSuffix = $portVersion compat >= 0`**» is enough.

Those recipes that will depend on **`libplist++`** just need to have a **`lib:libplist++$secondaryArchSuffix`** in their REQUIRES.

fbrosson:
On Gentoo the libplist package is not in the [dev-libs](https://packages.gentoo.org/categories/dev-libs) category but in [app-pda/libplist](https://packages.gentoo.org/packages/app-pda/libplist).
Since HaikuPorts tries to stick to the Gentoo ports categories, maybe we should use app-pda too.
Any thoughts?

@return, for your convenience, here are the commands to move the recipe:
```
cd haikuports/dev-libs/libplist
mkdir -p ../../app-pda/libplist
git mv libplist-1.12.recipe ../../app-pda/libplist
git commit --amend libplist-1.12.recipe ../../app-pda/libplist/libplist-1.12.recipe
```
Hmm, I looks like you are already an experienced Git user, so sorry if this is all obvious to you ;-)

dev-libs/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	cmd:pkg_config
+	cmd:sh
+	"
+	
+BUILD()
+{
+	runConfigure ./configure
+	make $jobArgs
+}
+INSTALL()
```
fbrosson: An empty line just above **`INSTALL()`** would be cool ;-)

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+	libplist$secondaryArchSuffix = $portVersion compat >= 0
+	cmd:plistutil$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libplist$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libplist++$secondaryArchSuffix = $portVersion compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libglib2.0$secondaryArchSuffix
```
fbrosson: You forgot the underscore. It should be:
```
```
(Haikuporter does not accept **`-`** in filenames, so when these have dashes we need to refer to them with underscores.)

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
...
+{
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	# remove libtool file
```
fbrosson: Cool, we're reaching perfection!
The only remaining changes that I see are:
* changing line 65 to:
```
	# remove libtool files
```

* checking if **`devel:libglib_2.0$secondaryArchSuffix`** should be added to **`REQUIRES_devel`**.

To see if **`devel:libglib_2.0`** should be added to **`REQUIRES_devel`** one needs to browse the header files that get packaged in **`libplist_devel`** and see if any of them refer to any header in the **`glib2_devel`** package.

That said, If you're not sure whether **`libplist_devel`** needs **`libglib2_devel`** then I think you can pretend it does (and, consequently, add **`devel:libglib_2.0$secondaryArchSuffix`**  to **`REQUIRES_devel`**) because it doesn't hurt ;-)

To all: if anyone knows that **`libplist_devel`** does not refer to any header files in **`libglib2_devel`** please let us know cause I'm not running Haiku at the moment so I can't check it right now. Thanks!

fbrosson: Oops, I forgot to mention that **`devel:libglib_2.0$secondaryArchSuffix`** should (also) be added to **`BUILD_REQUIRES`** in all cases, i.e. **_even_** if it is not needed in **`REQUIRES_devel`**.

return: The libplist maintainers have mentioned in the [NEWS](https://github.com/libimobiledevice/libplist/blob/060eda6b1846968aa40b0d38dcf92ac46f933417/NEWS) file that they have removed the glib dependency from libplist.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
+SUMMARY="A library to handle Apple Property List format in binary or XML"
+DESCRIPTION="libplist is a library for reading and writing \
+the Apple binary and XML property list format. It is part of \
+the libimobiledevice stack providing access to iDevices."
+HOMEPAGE="http://www.libimobiledevice.org"
```
fbrosson: May I suggest adding a trailing **`/`** to **`HOMEPAGE`**?
I know most browsers hide the trailing slash when there is nothing after it, but, to be correct, a URL needs either a path or a trailing slash.
I also know that this is a detail and that some recipes are missing that trailing slash.
But I think we should have correct URLs in our reicpes.
A practical reason for keeping the slash is that when wrong URLs are copy/pasted to forums that make hyperlinks with them, these hyperlinks do not always work because the wrapper that creates the hyperlinks is not able to correctly detect the end of the domain name when it is followed by a dot. I've already seen that in some places. This never happens with correct URLs ;-)
Well, please feel free to ignore my suggestion if you don't like trailing slashes ;-) Thanks!

app-pda/libplist/libplist-1.12.recipe
fbrosson: Hmm, I did not know that. Thanks for the info and pointer.
(I though libplist needed libglib2 because I had seen it in the packaging made by Arch Linux.)

We don't have libcnary but it seems libplist has a copy of it embedded.
So you can get rid of all libglib stuff in the recipe.
Sorry for my feedback based on outdated info ;-)

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,79 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:cmake
+	cmd:python
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sh
+	"
+BUILD()
```
fbrosson: Sorry, I just noticed there is no empty line above the **`BUILD()`** block. Could you add one please?
Once that is done I think the recipe will be ready for a merge.
(I haven't tested it yet, but I'll do so.)
Thanks for your contribution!

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,80 @@
+SUMMARY="A library to handle Apple Property List format in binary or XML"
+DESCRIPTION="libplist is a library for reading and writing \
+the Apple binary and XML property list format. It is part of \
+the libimobiledevice stack providing access to iDevices."
+HOMEPAGE="http://www.libimobiledevice.org/"
+COPYRIGHT="2009 Nikias Bassen"
+LICENSE="GNU GPL v3"
```
fbrosson: It seems libplist has 2 licenses: **`GNU GPL v2`** (according to file **`COPYING`**) and **`GNU LGPL v2.1`** (according to file **`COPYING.LESSER`**), so the recipe should have this:
```
LICENSE="GNU GPL v2
	GNU LGPL v2.1"
```

Regarding the COPYRIGHT variable, could you change it to:
```
COPYRIGHT="2008-2010 Jonathan Beck
	2010-2014 Nikias Bassen"
```
Thanks again!

fbrosson: Maybe we should also add:
```
	2009-2014 Martin Szulecki
	2008 Matt Colyer
```
They have each contributed almost 10% of all commits, compared to 41% for Jonathan Beck and 23% for Nikias Bassen, so maybe we should have all these 4 names in COPYRIGHT. What do you think?

return: I've added all the mentioned authors to the COPYRIGHT section, let me know if this recipe needs to be further refactored.

fbrosson: Cool, thanks.
I tryed to build it on x86_gcc2 (primary arch) but it looks like it does not build out of the box with gcc2. So it needs a gcc2 patch. I don't know yet how many changes need to be made.
It builds fine on x86, however. I was going to build on x86_64 but since there were a couple of incorrect library version numbers, I'll add comments about them in a few minutes.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,85 @@
...
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	libplist$secondaryArchSuffix = $portVersion compat >= 0
+	cmd:plistutil$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libplist$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libplist++$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson: Should be
```
	lib:libplist$secondaryArchSuffix = 3.0.0 compat >= 3
	lib:libplist++$secondaryArchSuffix = 3.0.0 compat >= 3
```
(The other lines in PROVIDES are all OK).

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,85 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libplist${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	devel:libplist$secondaryArchSuffix = $portVersion compat >= 0
+	devel:libplist++$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson: Should be
```
	devel:libplist$secondaryArchSuffix = 3.0.0 compat >= 3
	devel:libplist++$secondaryArchSuffix = 3.0.0 compat >= 3
```
(The other lines in PROVIDES_devel are all OK).

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,85 @@
...
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+
+}
+
+TEST()
+{
+	make test
```
fbrosson: Should be
```
	make check
```

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,85 @@
+SUMMARY="A library to handle Apple Property List format in binary or XML"
+DESCRIPTION="libplist is a library for reading and writing \
+the Apple binary and XML property list format. It is part of \
+the libimobiledevice stack providing access to iDevices."
+HOMEPAGE="http://www.libimobiledevice.org/"
+COPYRIGHT="2009-2014 Martin Szulecki
+		   2008-2010 Jonathan Beck
+		   2009 Nikias Bassen
+		   2008 Matt Colyer
```
fbrosson: For COPYRIGHT and LICENSE we don't put the closing double quotes in a separate line but at the last line. And we only put one tab starting at the second line. The result should look like this:
```
COPYRIGHT="2009-2014 Martin Szulecki
	   2008-2010 Jonathan Beck
	   2009 Nikias Bassen
	   2008 Matt Colyer"
```

Yep, these "rules" (about the closing double quotes) are not the same as for {PROVIDES,REQUIRES}{,_devel}...

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,85 @@
...
+		   "
+LICENSE="GNU GPL v2
+		 GNU LGPL v2.1"
```
fbrosson: There should be only one tab on the beginning of each line (for COPYRIGHT and LICENCE) except the very first line. So we should have this:
```
LICENSE="GNU GPL v2
	GNU LGPL v2.1"
```
Funny "rules", isn't it? Well, you'll get used to it after a couple of recipes ;-)

app-pda/libplist/libplist-1.12.recipe
fbrosson: Regaring the gcc2 patch, there are 2 options:
**_a_**) try to create a patch to make it build on x86_gcc2;
**_b_**) change **`x86_gcc2`** to **`!x86_gcc2`** in ARCHITECTURES and SECONDARY_ARCHITECTURES.
I'll dig option **_a_** to see if the patch is not too big. But maybe it's not worth doing **_a_**.

app-pda/libplist/libplist-1.12.recipe
fbrosson: BTW, when you list the contents of the devel package with **`package list /path/to/libplist_devel-*.hpkg`** you can see what the library names are. You can also have a look at the **`configure.ac`** and notice that there is a "3.0.0" somewhere.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+	2009 Nikias Bassen
+	2008 Matt Colyer"
+LICENSE="GNU GPL v2
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/$portName-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="x86 !x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
```
fbrosson: You forgot to change **`x86_gcc2`** to **`!x86_gcc2`** in SECONDARY_ARCHITECTURES.
You can do so now or you can wait until we find out if a gcc2 patch is desirable.
Shall we continue tomorrow? I'll dig the gcc2 patch option in the meanwhile.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+the libimobiledevice stack providing access to iDevices."
+HOMEPAGE="http://www.libimobiledevice.org/"
+COPYRIGHT="2009-2014 Martin Szulecki
+	2008-2010 Jonathan Beck
+	2009 Nikias Bassen
+	2008 Matt Colyer"
+LICENSE="GNU GPL v2
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/$portName-$portVersion.tar.bz2"
```
fbrosson: Oops, I had noticed a small error here but had forgotten to report it. We can't use $portName here because it's not always equal to "libplist". So we need to have:
```
SOURCE_URI="http://www.libimobiledevice.org/downloads/libplist-$portVersion.tar.bz2" 
```
BTW, when you fix this, maybe it's a good idea, after all, to change SECONDARY_ARCHITECTURES to:
```
```
This way you won't need to make another update in case we find out that the gcc2 patch is not desirable.

fbrosson:
I managed to create a patch to build libplist on x86_gcc2. It builds fine and passes all tests.
The problem is that the patch won't probably be accepted in haikuports because it is not a small patch: 445 lines, 8 files patched, 37 chunks.
So I decided to build libplist for x86 secondary arch. I've had to make a couple of changes to the recipe to make it build... but it fails all the tests with this error message:
```
runtime_loader: Cannot open file libplist.so.3: No such file or directory
```

So I'm a little bit stuck at the moment. Maybe someone can jump into this.
Anyway, I'll add comments with the changes I made so far.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/libplist-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="x86 !x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	libplist$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson: Every **`compat >= 0`** should be changed to **`compat >= 1`**

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+	2008 Matt Colyer"
+LICENSE="GNU GPL v2
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/libplist-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="x86 !x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
```
fbrosson: Can you add this after SECONDARY_ARCHITECTURES and before PROVIDES:
```

	commandBinDir=$binDir
else
	commandBinDir=$prefix/bin
fi

```

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+PROVIDES="
+	libplist$secondaryArchSuffix = $portVersion compat >= 0
+	cmd:plistutil$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson: Can you change this to:
```
	cmd:plistutil = $portVersion compat >= 1
```

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+	cmd:make
+	cmd:cmake
+	cmd:python
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sh
+	"
+
+BUILD()
+{
+	runConfigure ./configure
```
fbrosson: Can you change **`BUILD()`** to make it look like this:
```
BUILD()
{
		maybe_without_cython=--without-cython
	else
		maybe_without_cython=
	fi

	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir \
		$maybe_without_cython

	make $jobArgs
}
```

IMPORTANT: I don't know if passing **`--without-cython `** to configure is a good idea.
I had to do that because otherwise it would not build on x86 secondary arch.

But maybe we should depend on **`cmd:python$secondaryArchSuffix`** and avoid the **`--without-cython `**.
Let's wait until someone else can help us with this issue.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,84 @@
...
+	lib:libplist++$secondaryArchSuffix = 3.0.0 compat >= 3
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libplist${secondaryArchSuffix}_devel = $portVersion compat >= 0
```
fbrosson: Reminder: this shoud be changed to:
```
	libplist${secondaryArchSuffix}_devel = $portVersion compat >= 1
```

app-pda/libplist/libplist-1.12.recipe
pulkomandy: There is no particular reason Cython is not allowed to build for the secondary architectures. I was just too lazy to test if that could be made to work. Did someone try to get it building?
I don't know how it interacts with Python (which can't easily be built for the secondary arch, IIRC?).

fbrosson: I did not try to build cython, and I don't think I'll have time to do so in the near future...
And I guess you are right about python for the secondary arch.
So I think I'll wait until someone digs these issues ;-)

I've saved my gcc2 patch localy, and I can share it if turns out to be the easiest way to have libplist on x86_gcc2. BTW, the only thing it does is add  **`{`** and **`}`** at several places.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,99 @@
...
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.libimobiledevice.org/downloads/libplist-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0effdedcb3de128c4930d8c03a3854c74c426c16728b8ab5f0a5b6bdc0b644be"
+
+ARCHITECTURES="x86 !x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+    commandBinDir=$binDir
```
fbrosson: Oops, I did use tabs in my comments, but when you copy/paste them the tabs become spaces. It seems the only way preserve the tabs when you copy the blocs from my comments is to:
* click on the "edit" icon (i.e. the little pencil) to pretend you want to edit my comment;
* select and copy the block;
* click cancel;
* paste the clipboard to your editor.

Of course you can also replace each group of 4 spaces by a tab.

What I usually do before commiting is:
* look for 2 consecutive spaces anywhere;
* look for space+tab or tab+space anywhere;
* look spaces or tabs at end of lines.

Reminder: we try to always use tabs for the indentation. Of course, there may be exceptions from time to time, but here we prefer to have tabs. Sorry, BTW; I should have mentioned it earlier.

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,99 @@
+SUMMARY="A library to handle Apple Property List format in binary or XML"
+DESCRIPTION="libplist is a library for reading and writing \
+the Apple binary and XML property list format. It is part of \
+the libimobiledevice stack providing access to iDevices."
+HOMEPAGE="http://www.libimobiledevice.org/"
+COPYRIGHT="2009-2014 Martin Szulecki
+	2008-2010 Jonathan Beck
+	2009 Nikias Bassen
```
fbrosson: You forgot the **`-2014`** in **`2010-2014 Nikias Bassen`**. BTW, it might be a good idea to sort the 4 main contributors this way:
```
COPYRIGHT="2008-2010 Jonathan Beck
	2010-2014 Nikias Bassen
	2009-2014 Martin Szulecki
	2008 Matt Colyer"
```

app-pda/libplist/libplist-1.12.recipe
```diff
@@ -0,0 +1,99 @@
...
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	libplist$secondaryArchSuffix = $portVersion compat >= 1
+	cmd:plistutil = $portVersion compat >= 1
+	lib:libplist$secondaryArchSuffix = 3.0.0 compat >= 1
+	lib:libplist++$secondaryArchSuffix = 3.0.0 compat >= 1
```
fbrosson: Oops, you also changed **`compat >= 3`** into **`compat >= 1`**. Only the **`compat >= 0`** were to be changed to **`compat >= 1`**.
Well, since this is a bit confusing, I've uploaded the recipe to my remote branch:
https://github.com/fbrosson/haikuports/blob/libplist/app-pda/libplist/libplist-1.12.recipe
BTW, you also changed the indentation in BUILD and it is not very common. Could you copy the indentation I put in my branch? Thanks!

app-pda/libplist/libplist-1.12.recipe
fbrosson: I've uploaded my gcc2 patch, in case someone would like to have a look at it.
https://github.com/haikuports/haikuports/compare/master...fbrosson:libplist-gcc2

So if we don't find how to workaround the issues about building libplist for x86 secondary arch then maybe we can use the gcc2 patch if that is acceptable.

fbrosson:
@return: Thank you for your PR. libplist is a must have!

As discussed earlier (in comments that are now hidden by default), it works great for x86 primary arch and x86_64.
It also builds fine on x86_gcc2 hybrid using the x86 secondary arch... but it fails all tests, probably because of the **`--without-cython`** which seems to be needed given the current dependencies.
My understanding is that we need to either make it depend on python for the same (secondary) arch (instead of python for the primary arch), or on cython for the same (secondary) arch.
So we need to dig this or get help from someone who can solve this ;-)

The other option might be to add the gcc2 patch I have prepared: https://github.com/haikuports/haikuports/compare/master...fbrosson:libplist-gcc2
It builds fine on all architectures and passes all tests on all primary architectures.

So let's wait and see what other contributors think about all this ;-)

fbrosson:
There have been no new comments in this PR in almost 4 weeks, after my previous message so I guess @return is wondering what we will be doing with his PR...

Hence this question to all: Did anyone succeed in building python for x86 as secondary arch? If not, is anyone willing to do so soon?
If not, I really think we should switch to the variant that includes the gcc2 patch: https://github.com/haikuports/haikuports/compare/master...fbrosson:libplist_recipe

If anyone thinks we should **_not_** include a gcc2 patch in this recipe, but, instead, tag it as broken on x86_gcc2, please let us know.
I see two other options: leave this PR open until support for x86_gcc2 is dropped or until someone manages to build python for x86 secondary arch.
But IMHO libplist is stable enough to disserve a gcc2 patch for this recipe.
BTW, having libplist for all architectures will make it possible to add recipes for libimobiledevice and ideviceinstaller.

@return: If nobody is against the variant with the gcc2 patch, and if, like me, you think having the gcc2 patch is a good idea then you can either:
* grab my remote branch and push it to yours;
* **_or_** check the "[Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)" checkbox in this PR and I'll push (to your remote branch) an exact copy of what I have pushed to my remote branch.

(As you can see, I have taken care of keeping you as the author of the commit ;-)
Of course, once this is done I'll wait several days in order to allow other contributors to tell what they think about this recipe.

fbrosson:
@return, I just noticed you have deleted your remote branch. You can probably restore it from within GitHub's web-based interface. I can guide you if you need help.
BTW, it seems nobody disagrees with my gcc2 variant :-)

So if you restore your remote branch (or recreate it with the same name, libplist_recipe) and then push a copy of my remote branch then we might be able to continue with this PR.

Otherwise I can push directly to master my remote branch fbrosson:libplist_recipe and we'll get almost the same result as what we would get with a merge. The only difference is that this PR will turn red instead of purple.

return:
@fbrosson Since you have the additional gcc2 patches for libplist, you can close this PR and push your changes directly to master.

fbrosson:
OK, thanks.
Hmm, are you sure you don't want to try the "[Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)" ? All you need to do is one click.
I'm not sure it will work, but if it does then the PR will have its status changed to _Merged_ when I'll merge it.
If it does not work then I'll push the commit directly.
Reminder: the checkbox only grants write access to the remote branch you chose initially.

return:
This button doesn’t appear on PRs with deleted forks, but I have already ticked this box in #818. So I think your libplist commits should be pushed directly.

fbrosson:
OK, thanks.
Commit pushed directly 4 minutes ago.

PR #790, opened 2016-09-01
libpng: bump to 1.6.25.

fbrosson:


fbrosson:
I've added a "custom" **`libVersion`** variable and it works like a charm.
I don't know, however, if everyone agrees with its use.
Of course, if you don't like the idea I can revert to the hardcoded **`16.25.0`**.
I can also update the PR if you like the idea but would like to have another name or a hardcoded **`libVersion=16.25.0`** (instead of **`libVersion="${portVersion/\./}.0"`**).

PR #793, opened 2016-09-04
SDL2_mixer fix (libGL added, it require it, idk why...)

miqlas:


diversys:
This probably needs a better commit message

Begasus:
Iirc SDL2 requires libGL, hence all libs depending on SDL2 require libGL? Maybe it would be an idea to remove the dependency for libGL for SDL2, there have been people running SDL2 things without libGL (with own build recipes), haven't looked into that so far ...

waddlesplash:
If SDL2 requires libGL, then that should be the fix made, not this.

waddlesplash:
Oh, and it's in the recipe now: https://github.com/haikuports/haikuports/blob/master/media-libs/libsdl2/libsdl2-2.0.5.recipe#L25

PR #805, opened 2016-09-16
BeScreenCapture 2.1 recipe

jackburton79:
I just released BeScreenCapture 2.1 so I created a recipe for it.

### DIFF
--- a/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
+++ b/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
@@ -37,7 +37,7 @@ REQUIRES="
 	"
 REQUIRES_inputfilter="
 	haiku
-	bescreencapture == $portVersion
+	bescreencapture == $portVersion base
 	"
 
 BUILD_REQUIRES="
@@ -67,7 +67,7 @@ INSTALL()
 	cp -a inputfilter/objects/BeScreenCaptureInputFilter $addOnsDir/input_server/filters
 
 	packageEntries inputfilter \
-	$addOnsDir/input_server/filters/BeScreenCaptureInputFilter
+		$addOnsDir/input_server/filters/BeScreenCaptureInputFilter
 
 	addAppDeskbarSymlink $appsDir/BeScreenCapture
 }
--- a/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
+++ b/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
@@ -44,10 +44,6 @@ BUILD_REQUIRES="
 	haiku_devel
 	"
 
-BUILD_REQUIRES="
-	haiku_devel
-	"
-
 BUILD_PREREQUIRES="
 	makefile_engine
 	cmd:make
@@ -71,7 +67,7 @@ INSTALL()
 	cp -a inputfilter/objects/BeScreenCaptureInputFilter $addOnsDir/input_server/filters
 
 	packageEntries inputfilter \
-	add-ons/input_server/filters/BeScreenCaptureInputFilter
+	$addOnsDir/input_server/filters/BeScreenCaptureInputFilter
 
 	addAppDeskbarSymlink $appsDir/BeScreenCapture
 }
--- a/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
+++ b/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
@@ -13,8 +13,7 @@ combination (CTRL-COMMAND-SHIFT + R)."
 HOMEPAGE="https://github.com/jackburton79/bescreencapture"
 COPYRIGHT="2014-2016 Stefano Ceccherini"
 LICENSE="BSD (3-clause)
-	MIT
-	"
+	MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v2.1.tar.gz"
 CHECKSUM_SHA256="5b3ff575fcfe7519556fcd3d9487ef5491ead656f1cffd79b78dfcb3a52332f8"
--- a/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
+++ b/haiku-apps/bescreencapture/bescreencapture-2.1.recipe
@@ -11,22 +11,21 @@ user to launch BeScreenCapture and start/stop recording using a keyboard \
 combination (CTRL-COMMAND-SHIFT + R)."
 
 HOMEPAGE="https://github.com/jackburton79/bescreencapture"
-SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v2.1.tar.gz"
-CHECKSUM_SHA256="5b3ff575fcfe7519556fcd3d9487ef5491ead656f1cffd79b78dfcb3a52332f8"
-SOURCE_FILENAME="bescreencapture-2.1.tar.gz"
-LICENSE="
-	BSD (3-clause)
+COPYRIGHT="2014-2016 Stefano Ceccherini"
+LICENSE="BSD (3-clause)
 	MIT
 	"
-COPYRIGHT="2014-2016 Stefano Ceccherini"
 REVISION="1"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v2.1.tar.gz"
+CHECKSUM_SHA256="5b3ff575fcfe7519556fcd3d9487ef5491ead656f1cffd79b78dfcb3a52332f8"
+SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"
+
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 
 PROVIDES="
 	bescreencapture = $portVersion
 	app:BeScreenCapture = $portVersion
 	"
-
 PROVIDES_inputfilter="
 	bescreencapture_inputfilter = $portVersion
 	app:BeScreenCaptureInputFilter = $portVersion


haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,77 @@
...
+DESCRIPTION_inputfilter="Input Server Addon for BeScreenCapture. Allows the \
+user to launch BeScreenCapture and start/stop recording using a keyboard \
+combination (CTRL-COMMAND-SHIFT + R)."
+
+HOMEPAGE="https://github.com/jackburton79/bescreencapture"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v2.1.tar.gz"
+CHECKSUM_SHA256="5b3ff575fcfe7519556fcd3d9487ef5491ead656f1cffd79b78dfcb3a52332f8"
+SOURCE_FILENAME="bescreencapture-2.1.tar.gz"
+LICENSE="
+	BSD (3-clause)
```
fbrosson: A few lines are not following the [recommended order](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering). Could you replace lines 13 to 23 by the following?
```
HOMEPAGE="https://github.com/jackburton79/bescreencapture"
COPYRIGHT="2014-2016 Stefano Ceccherini"
LICENSE="BSD (3-clause)
	MIT"
REVISION="1"
SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v$portVersion.tar.gz"
CHECKSUM_SHA256="5b3ff575fcfe7519556fcd3d9487ef5491ead656f1cffd79b78dfcb3a52332f8"
SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"

```
Reminder: please make sure you put one tab (and no whitespaces) before **`MIT"`**.

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,77 @@
...
+	bescreencapture_inputfilter = $portVersion
+	app:BeScreenCaptureInputFilter = $portVersion
+	"
+
+REQUIRES="
+	haiku
+	"
+REQUIRES_inputfilter="
+	haiku
+	bescreencapture == $portVersion
```
fbrosson: I think a **`base`** keyword should be added here, like this:
```
	bescreencapture == $portVersion base
```

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,77 @@
...
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	cp -a objects/BeScreenCapture $appsDir
+	mkdir -p $addOnsDir/input_server/filters
+	cp -a inputfilter/objects/BeScreenCaptureInputFilter $addOnsDir/input_server/filters
+
+	packageEntries inputfilter \
+	add-ons/input_server/filters/BeScreenCaptureInputFilter
```
fbrosson: The indentation of the continuation line would look nicer with an extra tab.

BTW, it might also be a good idea to find which environment variable points to the add-ons dir and use it.
To find the name of that variable you can:
```Shell
grep add-ons /path/to/haikuports/haiku-apps/bescreencapture/work-2.1/wrapper-script
```
(I don't remeber what it is and I'm not running Haiku at the moment.)

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,77 @@
...
+REQUIRES_inputfilter="
+	haiku
+	bescreencapture == $portVersion
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+BUILD_REQUIRES="
```
fbrosson: Duplicate BUILD_REQUIRES.

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
fbrosson: Sorry, I just saw that you're already using the **`$addOnsDir`** variable.

So now I have a question: Shouldn't we also use it in the call to **`packageEntries`**, like this:
```
	packageEntries inputfilter \
		$addOnsDir/input_server/filters/BeScreenCaptureInputFilter
```

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,77 @@
...
+	"
+COPYRIGHT="2014-2016 Stefano Ceccherini"
+REVISION="1"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	bescreencapture = $portVersion
+	app:BeScreenCapture = $portVersion
+	"
+
```
fbrosson: I think the blocks should appear in this order:
```
PROVIDES
REQUIRES

PROVIDES_inputfilter
REQUIRES_inputfilter
```
But maybe the current order,
```
PROVIDES
PROVIDES_inputfilter
REQUIRES
REQUIRES_inputfilter
```
is also OK. Does someone know which one is recommended?

jackburton79:
Il 17/set/2016 09:25 AM, "fbrosson" <notifications@github.com> ha scritto:
>
> @fbrosson commented on this pull req " +
>
> I think the blocks should appear in this order:
>
> PROVIDES REQUIRES PROVIDES_inputfilter REQUIRES_inputfilter
>
> But maybe the current order,
>
> PROVIDES PROVIDES_inputfilter REQUIRES REQUIRES_inputfilter
>
> is also OK. Does someone know which one is recommended?
>

Actually I copied the recipe for bescreencapture 2.0 already in HaikuPorts
repo and just changed the references to the source code, so in case this is
not correct, that recipe should be fixed, too.

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
humdingerb: The first order is correct, i think. See recipes with _devel package.

fbrosson:
Oops, sorry. I had not noticed that the repo already had the recipes for 4 older releases.
Well, since the changes are only cosmetic, maybe it's not worth editing the old recipes.
I'm sure it's OK to "fix" only the new recipe. The older ones will probably be removed some day.

BTW, thanks for developing BeScreenCapture !

jackburton79:
I updated the recipe with the requested changes.

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
fbrosson: Yep, I was also thinking about the ordering of the _devel subpackages.
But I just noticed that the recipes for git are using the second style, so maybe _devel is a special case.
So I guess the recommended ordering for this recipe might be:

PROVIDES
PROVIDES_inputfilter

REQUIRES
REQUIRES_inputfilter

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
fbrosson: > I updated the recipe with the requested changes.

All the changes are OK, but you forgot to fix the ordering of a few variables:
* COPYRIGHT should go right after HOMEPAGE.
* LICENSE should go right after COPYRIGHT.
* REVISION should go right after LICENSE.

There are a couple of empty lines to add or remove:
* It is recommended to **add** a blank line between SOURCE_FILENAME and ARCHITECTURES.
* It is recommended to **remove** the blank line between PROVIDES and PROVIDES_inputfilter.

All the rest is OK ;-)

jackburton79:
Il 18/set/2016 22:24, "fbrosson" <notifications@github.com> ha scritto:
>
> @fbrosson commented on this pull request.
>

> All the changes are OK, but you forgot to fix the ordering of a few
variables:
>
> COPYRIGHT should go right after HOMEPAGE.
> LICENSE should go right after COPYRIGHT.
> REVISION should go right after LICENSE.
>
> There are a couple of empty lines to add or remove:
>
> It is recommended to add a blank line between SOURCE_FILENAME and
ARCHITECTURES.
> It is recommended to remove the blank line between PROVIDES and
PROVIDES_inputfilter.
>
> All the rest is OK ;-)
>
>
Oops! Will do, sorry.

jackburton79:
Done. Is it ok now ?

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,72 @@
...
+SUMMARY_inputfilter="Shortcut handler for BeScreenCapture"
+DESCRIPTION_inputfilter="Input Server Addon for BeScreenCapture. Allows the \
+user to launch BeScreenCapture and start/stop recording using a keyboard \
+combination (CTRL-COMMAND-SHIFT + R)."
+
+HOMEPAGE="https://github.com/jackburton79/bescreencapture"
+COPYRIGHT="2014-2016 Stefano Ceccherini"
+LICENSE="BSD (3-clause)
+	MIT
```
fbrosson: For LICENSE (and COPYRIGHT) it is recommended to put the closing quotes to the right of the last line, instead of putting it in a separate line.
You can have a look at [line 11 of bescreencapture-1.9.2.recipe](https://github.com/haikuports/haikuports/blob/56a7c925628a95e4fed51239c6f160ecad4b0248/haiku-apps/bescreencapture/bescreencapture-1.9.2.recipe#L11) to see the recommended layout for the LICENSE variable.

haiku-apps/bescreencapture/bescreencapture-2.1.recipe
```diff
@@ -0,0 +1,72 @@
...
+	"
+REVISION="1"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v2.1.tar.gz"
```
fbrosson: This is not required, but it can be handy to use the **`$portVersion`** variable for **`SOURCE_URI`**:
```
SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v${portVersion}.tar.gz"
```

jackburton79:
2016-09-19 15:06 GMT+02:00 fbrosson <notifications@github.com>:

> *@fbrosson* commented on this pull request.
> ------------------------------
> +SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v2.1.
> tar.gz"
>
> This is not required, but it can be handy to use the *$portVersion*
> variable for *SOURCE_URI*:
>
> SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v${portVersion}.tar.gz"
>
>
>
I wouldn't comfortable with that since I'm not sure I will use the same
tagging scheme in the future, so it would be confusing, at least for me.

fbrosson:
Hmm, that should not be a problem. Using $portVersion has no drawbacks other than forcing us to look at the name of the recipe to find out which release it is using.
You can safely use $portVersion in the recipe for the current release (2.1). Then, if you change the tagging scheme it will always be possible to update the SOURCE_URI if, for example, the prefix of the tarballs change from "v" to something else (e.g. "bescreencapture-").

But if you prefer to keep the SOURCE_URI as is, with the hard-coded "2.1", that's OK. Many recipes do the same ;-)

fbrosson:
Great app. Thanks @jackburton79 !

PR #808, opened 2016-09-20
GYP Hash fix

miqlas:


TwoFX:
I just downloaded the file multiple times and running `sha256sum` gave a different result each time. It appears that this way of obtaining the source does not give stable results. Is there a different way of getting the source for gyp?

TwoFX:
cc @pulkomandy @humdingerb 

I dug around some more and found out that the FreeBSD ports project has a mirror of the gyp source.

Their current version is `e679e688ba6db12336b79650c27fcf8a5ed21e60`. I have no idea how they came up with that version number. The version of the source that they distribute under that hash is what the gyp project itself files under commit [`02b145a1a4f4e1c62e8bae06045caf852d9ef17f`](https://chromium.googlesource.com/external/gyp/+/02b145a1a4f4e1c62e8bae06045caf852d9ef17f) from May 4 (which I determined by manually bisecting). This mirror is located at http://distcache.FreeBSD.org/local-distfiles/hrs/gyp-e679e688ba6db12336b79650c27fcf8a5ed21e60.tar.gz and has the sha256 `51fadf46a7a108b0c8f273fb726b113609d7be5394d286716cbf472d1327d18c`.

So if that version is new enough for everyone's needs we could use the FreeBSD mirror (though I don't know if there are some usage policies or such which we would break by doing that).

@humdingerb, you updated the gyp version from Jan 06 2016 to Aug 10 2016. Was there some special need for that or is the May 04 2016 version fine too?

humdingerb:
Weird that every downloaded archive has a different checksum! Of course, I didn't realize that when I changed the checksum...
When I recently updated the recipe, I originally only wanted to fix the checksum as that was the reason it failed at [Pulkomandy's build log](http://pulkomandy.tk/~pulkomandy/). Why not also update to the latest "release", I thought.

I think it's OK to just use the FreeBSD fork. Thanks for the research!

pulkomandy:
The checksum changes because the server doesn't store the archive, instead it generates it on-demand when it gets a download request. We hit the same problem with sourceforge archives when exporting a specific commit.
We don't get this problem with github, either they cache the files for a long enough time, or they are able to generate exactly the same archive each time.

TwoFX:
Switched to the FreeBSD mirror in cea5f2b.

PR #816, opened 2016-10-02
openjpeg: bump to 2.1.2.

fbrosson:
* Use **`binDir`** for secondary arch builds.
* Drop **`TEST()`** because it is useless without an additional data file.
* **`LIB_SUFFIX`** is not (or no longer) used so don't pass it to cmake.
* Depend on **`liblcms2`** on all architectures where it is available.
* Add «**`>= 5`**» for the dependency on **`libtiff`** to pull the latest.

### DIFF
--- a/media-libs/openjpeg/openjpeg-2.1.2.recipe
+++ b/media-libs/openjpeg/openjpeg-2.1.2.recipe
@@ -20,7 +20,7 @@ LICENSE="BSD (2-clause)"
 REVISION="1"
 SOURCE_URI="http://github.com/uclouvain/openjpeg/archive/v$portVersion.tar.gz"
 SOURCE_FILENAME="openjpeg-$portVersion.tar.gz"
-CHECKSUM_SHA256="82c27f47fc7219e2ed5537ac69545bf15ed8c6ba8e6e1e529f89f7356506dbaa"
+CHECKSUM_SHA256="4ce77b6ef538ef090d9bde1d5eeff8b3069ab56c4906f083475517c2c023dfa7"
 PATCHES="openjpeg-$portVersion.patchset"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64"
@@ -28,21 +28,26 @@ SECONDARY_ARCHITECTURES="x86_gcc2 x86"
 
 PROVIDES="
 	openjpeg$secondaryArchSuffix = $portVersion
-	cmd:opj_decompress = $portVersion
-	cmd:opj_compress = $portVersion
-	cmd:opj_dump = $portVersion
-	lib:libopenjp2$secondaryArchSuffix = 7.0.1 compat >= 7
+	cmd:opj_decompress$secondaryArchSuffix = $portVersion
+	cmd:opj_compress$secondaryArchSuffix = $portVersion
+	cmd:opj_dump$secondaryArchSuffix = $portVersion
+	lib:libopenjp2$secondaryArchSuffix = 7.0.2 compat >= 7
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
 	lib:libpng$secondaryArchSuffix
-	lib:libtiff$secondaryArchSuffix
+	lib:libtiff$secondaryArchSuffix >= 5
 	lib:libz$secondaryArchSuffix
 	"
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	REQUIRES="$REQUIRES
+		lib:liblcms2$secondaryArchSuffix
+		"
+fi
 
 PROVIDES_devel="
 	openjpeg${secondaryArchSuffix}_devel = $portVersion
-	devel:libopenjp2$secondaryArchSuffix = 7.0.1 compat >= 7
+	devel:libopenjp2$secondaryArchSuffix = 7.0.2 compat >= 7
 	"
 REQUIRES_devel="
 	openjpeg$secondaryArchSuffix == $portVersion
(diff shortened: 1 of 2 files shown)


media-libs/openjpeg/openjpeg-2.1.2.recipe
```diff
@@ -20,18 +20,18 @@ LICENSE="BSD (2-clause)"
...
 PROVIDES="
 	openjpeg$secondaryArchSuffix = $portVersion
-	cmd:opj_decompress = $portVersion
-	cmd:opj_compress = $portVersion
-	cmd:opj_dump = $portVersion
-	lib:libopenjp2$secondaryArchSuffix = 7.0.1 compat >= 7
+	cmd:opj_decompress$secondaryArchSuffix = $portVersion
+	cmd:opj_compress$secondaryArchSuffix = $portVersion
+	cmd:opj_dump$secondaryArchSuffix = $portVersion
+	lib:libopenjp2$secondaryArchSuffix = $portVersion compat >= 7
```
fbrosson: I changed **`7.0.1`** into **`$portVersion `** because the real soname is **`libopenjp2.so.2.1.2`**, although the symbolic link (to it) is **`libopenjp2.so.7`**.
Please correct me if I'm missing something. Thanks!

media-libs/openjpeg/openjpeg-2.1.2.recipe
```diff
@@ -85,5 +86,7 @@ INSTALL()
 
 TEST()
 {
-	make check
+	cd tests
+	cmake ..
+	make
```
fbrosson: **`make check`** did not work, so I changed it to something that does not fail... but it does not seem to launch any tests, unfortunately.
If anyone knows how to launch the tests on projects that use CMake that would be great. Thanks!

korli: https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/openjpeg/openjpeg-2.1.2.ebuild#n74
https://cmake.org/Wiki/CMake/Testing_With_CTest

Doesn't "make test" work?

media-libs/openjpeg/openjpeg-2.1.2.recipe
fbrosson: @korli: Thanks again for your review and sorry for this late reply...

I don't know what to use instead of **`7.0.1`**. Should I use **`7.0.2`** or **`7.1.0`**. Or should I use something else?

Hmm, the only recipes that need libopenjp2 are documentviewer, pillow and mupdf, and none of them require a specific version of libopenjp2. So maybe we could use **`$portVersion`** without the **`compat >= 7`** or with a **`compat >= 2`**.

Anyway, since I'm not sure of what should be used here, just tell me what's the best option and I'll update the recipe with your suggestion. Thanks!

media-libs/openjpeg/openjpeg-2.1.2.recipe
fbrosson: Nope, IIRC the generated Makefile does not have a "check" target.

But thanks a lot for your pointers. I think the project has a CTestTestfile.cmake file in a subdirectory, so I guess that calling ctest from within that directory will do the trick. I'll try that and update the recipe with an improved TEST function.

media-libs/openjpeg/openjpeg-2.1.2.recipe
korli: The already built packages of documentviewer, pillow and mupdf have the version requirement. So you can only change the major version number, if you also rebuild these packages.
I suppose 7.0.2 is OK for now.

fbrosson: OK, thanks. I've just updated the PR to use **`7.0.2`**.

BTW, I found that passing «**`-DOPENJPEG_SOVERSION:STRING=2.1`**» to the call to **`cmake`** in **`BUILD()`** changes the name of the symbolic link to **`libopenjp2.so.2.1.2`**:
With that value we get a «**`libopenjp2.so.2.1 -> libopenjp2.so.2.1.2`**» symlink instead of the default «**`libopenjp2.so.7 -> libopenjp2.so.2.1.2`**» symlink.
I did not use it, however, because the **`CMakeLists.txt`** file says that autotools does not support X.Y notation for SOVERSION.
But who knows, maybe we could use the **`OPENJPEG_SOVERSION`** parameter if we ever decide to get rid of the fake **`7.0.2`** SOVERSION.

media-libs/openjpeg/openjpeg-2.1.2.recipe
fbrosson: The tests require an extra file that is not in the source tarball, so I've removed the **`TEST()`** function.

BTW, I think that if we ever chose to add the data set used for the tests then we should also pass «**`-DBUILD_TESTING=ON`**» to the call to **`cmake`** in **`BUILD()`** and define **`TEST()`** with something like:

```
TEST()
{
	cd tests
	cmake \
		-DOPENJPEG_SOURCE_DIR=$sourceDir \
		-DOPENJPEG_BINARY_DIR=$sourceDir/build \
		-DZ_INCLUDE_DIRNAME=$portPackageLinksDir/devel~libz$secondaryArchSuffix/$relativeIncludeDir \
		-DPNG_INCLUDE_DIRNAME=$portPackageLinksDir/devel~libpng$secondaryArchSuffix/$relativeIncludeDir \
		-DTIFF_INCLUDE_DIRNAME=$portPackageLinksDir/devel~libtiff$secondaryArchSuffix/$relativeIncludeDir \
		.
	make
}
```

fbrosson:
Thanks!

PR #817, opened 2016-10-03
Replaced Qt5 5.6.1 SHA 256 checksum

return:
Fixes the SHA 256 checksum issue in the qt5-5.6.1 recipe.

### DIFF
--- a/dev-qt/qt5/qt5-5.6.1.recipe
+++ b/dev-qt/qt5/qt5-5.6.1.recipe
@@ -7,7 +7,7 @@ HOMEPAGE="https://qt.io/"
 COPYRIGHT="2015 The Qt Company Ltd."
 LICENSE="GNU LGPL v2"
 
-SOURCE_URI="https://download.qt.io/official_releases/qt/5.6/$portVersion/single/qt-everywhere-opensource-src-$portVersion.tar.xz"
+SOURCE_URI="https://download.qt.io/official_releases/qt/5.6/$portVersion/single/qt-everywhere-opensource-src-$portVersion.tar.gz"
 SOURCE_URI_2="https://download.qt.io/community_releases/5.6/$portVersion/qtwebkit-opensource-src-$portVersion.tar.xz"
 CHECKSUM_SHA256="2e5be28fcaaf95e179041a0eb95ccca69b17f50e39245e20974fad9469c10f2f"
 CHECKSUM_SHA256_2="f5cfbfa5fad2b65a9be907d426f99694b86a04c3bb5a8814b21bd0ade4c672aa"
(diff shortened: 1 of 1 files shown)


dev-qt/qt5/qt5-5.6.1.recipe
```diff
@@ -9,7 +9,7 @@ LICENSE="GNU LGPL v2"
 
 SOURCE_URI="https://download.qt.io/official_releases/qt/5.6/$portVersion/single/qt-everywhere-opensource-src-$portVersion.tar.xz"
```
fbrosson: Good catch!
Wait, the SHA256 was correct... for the qt-everywhere-opensource-src-5.6.1.tar.**gz** ;-)

So if you could revert the CHECKSUM_SHA256 and replace the .tar.**xz** by a .tar.**gz** in SOURCE_URI that would be great!

BTW, while at it, would you mind removing the whitespace at the end of line 327 (which starts with devCommands) ? (I know it was already there before your edit, of course.)
Thanks !

fbrosson: I forgot to explain why we try to avoid xz archives when possible: haikuporter is not (yet) able to extract *.tar.xz archives with on-the-fly decompression. It needs to unxz them first, and then it calls tar to extract the uncompressed tarball (before deleting it.) So for very big archives this can take a while...
And the other reason why we try to avoid xz archives is that xz is not available for non hybrid x86_gcc2.

fbrosson:
Great job, thanks!
BTW, don't worry about squashing your commits. I'll do that with GitHub when I merge.

dev-qt/qt5/qt5-5.6.1.recipe
```diff
@@ -322,10 +322,10 @@ INSTALL()
 	mv -f $libDir/libQt5QmlDevTools* $developLibDir/
 	mv -f $libDir/libQt5Bootstrap* $developLibDir/
 	mv -f $libDir/libQt5PlatformSupport* $developLibDir/
-	mv -f $libDir/libQt5OpenGLExtensions* $developLibDir/	
+	mv -f $libDir/libQt5OpenGLExtensions* $developLibDir/
 
-	devCommands="assistant designer linguist lconvert lrelease lupdate 
-		pixeltool qcollectiongenerator qdbusviewer qdbus qhelpconverter
+	devCommands="assistant designer linguist lconvert lrelease lupdate
+	pixeltool qcollectiongenerator qdbusviewer qdbus qhelpconverter
```
fbrosson: Oops, I've just noticed you ate one tab in line 328, which starts with the word pixeltool.
Could you please insert it back, so that this line starts with 2 tabs? Thanks again!

fbrosson:
Thanks!

dev-qt/qt5/qt5-5.6.1.recipe
korli: I don't get where this is going again. Xz is a significant download and space gain for a piece like Qt. And Qt is out of question on x86_gcc2, so this point is moot.

fbrosson: There were 2 ways to fix this recipe:
* change the 64 characters of the CHECKSUM_SHA256.
* **_or_** change a single character: tar.**x**z → tar.**g**z

Since you had created the recipe with the SHA256 of the **gz**ipped tarball, it was clear to me that you had downloaded that tarball.

Now, if you want to switch back to the tar.xz you know you don't need anyone's permission.

korli: Well, thanks for your help.

PR #818, opened 2016-10-04
discount: updated recipe to 2.2.1

return:


app-text/discount/discount-2.2.1.recipe
```diff
@@ -21,27 +21,27 @@ PROVIDES="
 	cmd:theme = $portVersion compat >= 2
 	"
 REQUIRES="
-	haiku
+	haiku$secondaryArchSuffix
```
fbrosson: **`$secondaryArchSuffix`** is only useful for recipes that define a SECONDARY_ARCHITECTURES.
Since both conditions are met:
* this recipe provides only commands (**`cmd:theme`**) but no library (**`lib:` or **`devel:`**);
* this recipe builds for all primary architectures;

we don't need to make it build for secondary architectures.
So having **`$secondaryArchSuffix`** is not needed (nor recommended) because it will always be the empty.

app-text/discount/discount-2.2.1.recipe
```diff
@@ -7,11 +7,11 @@ how to fully utilize the markdown library."
 HOMEPAGE="http://www.pell.portland.or.us/~orc/Code/discount/"
 COPYRIGHT="2007-2012 David Loren Parsons"
 LICENSE="BSD (4-clause)"
-REVISION="1"
-SOURCE_URI="http://www.pell.portland.or.us/~orc/Code/discount/discount-2.1.3.tar.bz2"
-CHECKSUM_SHA256="7280da7a27be42244311912eeee19695623f196817ec08f43b652fcc2ded975d"
+REVISION="2"
```
fbrosson: No need to bump REVISION to 2 since this is a new release. You can keep "1".

app-text/discount/discount-2.2.1.recipe
```diff
@@ -7,11 +7,11 @@ how to fully utilize the markdown library."
 HOMEPAGE="http://www.pell.portland.or.us/~orc/Code/discount/"
 COPYRIGHT="2007-2012 David Loren Parsons"
```
fbrosson: 2012 → 2016

humdingerb:
> To all: Is it worth adding support for secondary architectures for this recipe?
I don't think so. As I understand it, secondaryArch is mostly interesting if a software can't be build for the primaryArch (either x86 or x86_gcc2). There might be exceptions for add-ons...

fbrosson:
@humdingerb: Thanks for your answer!

@return: This means this recipe should not provide support for building on secondary architectures. So if you could remove all occurrences of **`$secondaryArchSuffix`** (and update the copyright year in COPYRIGHTS) that would be great! Thanks again :-)

Reminder: if you are busy then I can take care of everything. All you need to do is check the "[Allow edits from maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)" checkbox in this PR on the web-based interface of GitHub and let me know you did so so that I can later push the required changes to your remote branch and then click "Squash and merge" on this PR. That said, there is no urgency at all, so you can do as you wish.

fbrosson:
The "Allow edits from maintainers" feature works like a charm!
I've just updated this PR with a few cosmetic changes:
* Added cmd:diff to BUILD_PREREQUIRES;
* Sorted lines in BUILD_PREREQUIRES;
* Added a TEST() function.

fbrosson:
@return: Sorry, I just realized that, contrary to what I though,  this recipe **_does_** provide a shared library. This means that the **`$secondaryArchSuffix`** you had added in your initial commit were all OK. So sorry for having asked you to remove them...

The reason why I though this recipe was not providing any shared lib was that it did not have a **`lib:libmarkdown`** entry in PROVIDES.
And I guess the reason why you did not include a **`lib:libmarkdown`** entry in PROVIDES was that you were calling the **`./configure.sh`** script with **`--enable-shared`** while it was expecting just **`--shared`** in order to create the shared lib.

Well, since you have kindly checked the "allow edits" checkbox, I can fix everything!
So I'll restore your initial commit, the one from Tuesday evening, and add a few fixes in a separate commit (that will get squashed before the merge).
Of course, when I merge this PR I'll use the "Squash and merge" feature ;-)

fbrosson:
For the squash and merge, shall we use a one-line commit message (without any detailed message) ?
```
discount: bump to version 2.2.1 and enable secondary arch.
```

korli:
LGTM

fbrosson:
@korli: Thanks for your review and sorry for this late reply...

I found a minor issue but did not dig it yet: **`haikuporter --test discount`** worked fine until I fixed the configure arguments to really build the shared lib.
The problem is that when haikuporter launches the TEST function from within the chrooted environment, the system fails to find libmarkdown.so when ./markdown is called.
Hopefully this does not happen if **`make test`** is called normaly, i.e. without the chrooted environment.
So I guess the PR can be merged. But I'll try to fix the small issue mentioned above before the merge.
That said, if anyone wishes to merge this PR as is, that's OK for me. (Please don't forget to remove the commit's long description and to change the commit tittle to something like **`discount: bump to version 2.2.1, enable 2nd arch, add TEST.`** or **`discount: bump to version 2.2.1, enable 2nd arch.`**, in case you merge.)

PR #825, opened 2016-10-10
Recipe for BeScreenCapture 2.2.1

jackburton79:
Recipe for BeScreenCapture 2.2.1.
Reorganized PROVIDES and REQUIRES fields,
added $portVersion

### DIFF
--- a/haiku-apps/bescreencapture/bescreencapture-2.2.1.recipe
+++ b/haiku-apps/bescreencapture/bescreencapture-2.2.1.recipe
@@ -1,0 +1,71 @@
+SUMMARY="A screen recorder utility"
+DESCRIPTION="BeScreenCapture is a screen recorder utility for Haiku.
+It allows you to record what happens on your screen, then save it \
+to any media format supported in Haiku.
+BeScreenCapture can record either the entire screen, or just a section you \
+select."
+
+SUMMARY_inputfilter="Shortcut handler for BeScreenCapture"
+DESCRIPTION_inputfilter="Input Server Addon for BeScreenCapture. Allows the \
+user to launch BeScreenCapture and start/stop recording using a keyboard \
+combination (CTRL-COMMAND-SHIFT + R)."
+
+HOMEPAGE="https://github.com/jackburton79/bescreencapture/releases"
+COPYRIGHT="2014-2016 Stefano Ceccherini"
+LICENSE="BSD (3-clause)
+	MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="779a7ac4e4e7b935f31ff95b47ab1325a81a98d9fe5d76017fb8c4e01b3ed5ac"
+SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	bescreencapture = $portVersion
+	app:BeScreenCapture = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+PROVIDES_inputfilter="
+	bescreencapture_inputfilter = $portVersion
+	app:BeScreenCaptureInputFilter = $portVersion
+	"
+REQUIRES_inputfilter="
+	haiku
+	bescreencapture == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:make
+	cmd:gcc
+	"
+
+BUILD()
+{
+	make OBJ_DIR=objects \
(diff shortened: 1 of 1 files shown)


jackburton79:
After I created the release tag, I found out that the source code didn't compile on gcc5. So I had to create a new tag (2.2.1a). How can I reference this tag in the recipe, without changing version ?
If I add the "a" after $portVersion (like this: bescreencapture-v($portVersion)a ) the source is downloaded correctly, but then tar fails to extract the source. Any idea ?

humdingerb:
Try bescreencapture-v${portVersion}a.

pulkomandy:
you will probably need to set the SOURCE_DIR as well, as it is now not matching the default of $portName-$portVersion.

fbrosson:
You _can_ update the CHECKSUM_SHA256, change SOURCE_FILENAME and add a SOURCE_DIR:
```
SOURCE_FILENAME="bescreencapture-${portVersion}a.tar.gz"
SOURCE_DIR="bescreencapture-$[portVersion}a"
```
but I think it would be **_much better_** to only update the CHECKSUM_SHA256 (with the hash from the 2.2.1a tarball) and rename the recipe:
```
cd /path/to/haikuports/haiku-apps/bescreencapture
vi bescreencapture-2.2.1.recipe # to update the CHECKSUM_SHA256 to 2.2.1a
git mv bescreencapture-2.2.1.recipe bescreencapture-2.2.1a.recipe
git commit --amend bescreencapture-2.2.1.recipe bescreencapture-2.2.1a.recipe
```

fbrosson:
Hmm, I just noticed this pull request is for a commit you made to your master branch. Although doing so is OK, most PRs are created for commits made to [feature-branches](https://github.com/haikuports/haikuports/wiki/DevelopmentModel#what-are-feature-branches) because it gives more flexibility.

jackburton79:
If I rename the recipe, it will change the version number, too, right ?
I'd like the version to remain 2.2.1, not 2.2.1a.
So I just have to add SOURCE_DIR.
Thanks for the hint about feature-branches.

haiku-apps/bescreencapture/bescreencapture-2.2.1.recipe
```diff
@@ -15,9 +15,10 @@ COPYRIGHT="2014-2016 Stefano Ceccherini"
 LICENSE="BSD (3-clause)
 	MIT"
 REVISION="1"
-SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v$portVersion.tar.gz"
-CHECKSUM_SHA256="779a7ac4e4e7b935f31ff95b47ab1325a81a98d9fe5d76017fb8c4e01b3ed5ac"
-SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v"$portVersion"a.tar.gz"
+CHECKSUM_SHA256="d36f34d0dcf8bb87d58d782e649f29862fbbaed68a26182ccb695b0b0716ec87"
+SOURCE_FILENAME="bescreencapture-"$portVersion"a.tar.gz"
+SOURCE_DIR="bescreencapture-"$portVersion"a"
```
fbrosson: >  I'd like the version to remain 2.2.1, not 2.2.1a.

OK, no problem ;-)

BTW, would you mind making a small neutral change? I think this would be slightly better:
```
SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v${portVersion}a.tar.gz"
+CHECKSUM_SHA256="d36f34d0dcf8bb87d58d782e649f29862fbbaed68a26182ccb695b0b0716ec87"
+SOURCE_FILENAME="bescreencapture-${portVersion}a.tar.gz"
+SOURCE_DIR="bescreencapture-${portVersion}a"
```

I know that using «**`"$portVersion"a`**» works fine here, but IMHO «**`${portVersion}a`**» is more consistent with the style of all other recipes that already use «**`${portVersion}`**». Thanks!

fbrosson: Oops, I assumed everyone was OK with keeping «**`2.2.1`**» for the version (instead of «**`2.2.1a`**»), but actually I don't know if all other major contributors also agree with that. (So comments are welcome, as usual!)

Reminder: I'm OK with keeping «**`2.2.1`**» for the version, although I would have personally used «**`2.2.1a`**».

Regarding the  «**`"$portVersion"a`**» vs «**`${portVersion}a`**», however, I think the style of the latter preferable.

jackburton79:
Is it okay now ?

fbrosson:
> Is it okay now ?

Yes. BTW, sorry for this late reply and thanks for your update ;-)

I haven't tried to build it yet, but I'll try to do so. That said, if anyone already has and wants to merge, please do not hesitate!

jackburton79:
> Yes. BTW, sorry for this late reply and thanks for your update ;-)

Don't worry about that :-)

PR #832, opened 2016-10-21
geoip: drop libtool file, permalinkify SOURCE_URI_2, add TEST().

fbrosson:


fbrosson:
Oops, I just noticed that "make check" needs a gcc2 patch.
It looks like it is simple. Update comming soon.

fbrosson:
> Oops, I just noticed that "make check" needs a gcc2 patch.
> It looks like it is simple. Update comming soon.

Done.

Reminder: If you want to run **`haikuporter --test geoip`** then you can grab a **`GeoIP.dat`** with **`geoipupdate.sh`** (from the geoip recipe) or a **`GeoLiteCountry.dat`** with **`geoipupdate`** (from the geoipupdate recipe) and copy it as **`GeoIP.dat`** in **`/path/to/haikuports/dev-libs/geoip/work-*/sources/geoip-api-c-*/data/`**, but if you already have one of these then do not grab the latest because their DBInfo are truncated in the middle of the word "Copyright"... and this makes the test fail.

So if you have an old **`GeoIP.dat`** for which the following command yelds something:
```
geoiplookup -v localhost | grep "GeoIP Country Edition:" | grep Copyright
```

then the tests will succeed (provided you create a **`data`** subdirectory in **`$sourceDir`** and copy the file in there.) 
FWIW, I had an old copy of the free **`GeoIP.dat`** from January 2016 which has a DBInfo that makes the test happy:
```
GeoIP Country Edition: GEO-106FREE 20160105 Build 1 Copyright (c) 2016 MaxMind 
```
FWIW, the DBInfo of the current free **`GeoIP.dat`** is truncated (and makes the test fail):
```
GeoIP Country Edition: GEO-106FREE 20161004 Build 1 Copy
```

fbrosson:
Thanks!

PR #834, opened 2016-10-22
Libguess 1.2 recipe

miqlas:
New recipe

### DIFF
--- a/app-i18n/libguess/libguess-1.2.recipe
+++ b/app-i18n/libguess/libguess-1.2.recipe
@@ -1,0 +1,71 @@
+SUMMARY="Character set guessing library"
+DESCRIPTION="Libguess is a high speed, open source character set encoding \
+detector. Libguess employs discrete-finite automata to deduce the character \
+set of the input buffer. The advantage of this is that all character sets can \
+be checked in parallel, and quickly."
+HOMEPAGE="http://atheme.org/projects/libguess.html"
+COPYRIGHT="2000-2003 Shiro Kawai
+	2009-2013 Tony Vroon"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://rabbit.dereferenced.org/~nenolod/distfiles/libguess-$portVersion.tar.bz2"
+CHECKSUM_SHA256="8019a16bdc7ca9d2efbdcc1429d48d033d5053d42e45fccea10abf98074f05f8"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	libguess$secondaryArchSuffix = $portVersion compat >= 1.2
+	lib:libguess$secondaryArchSuffix = 1.0.0 compat >= 1
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libguess${secondaryArchSuffix}_devel = $portVersion compat >= 1.2
+	devel:libguess$secondaryArchSuffix = 1.0.0 compat >= 1
+	"
+REQUIRES_devel="
+	libguess$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libiconv$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:gettext
+	"
+
+BUILD()
+{
+	autoreconf -fi
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	prepareInstalledDevelLib libguess
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
+}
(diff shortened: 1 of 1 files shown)


app-i18n/libguess/libguess-1.2.recipe
```diff
@@ -0,0 +1,66 @@
...
+	autoreconf -fi
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	prepareInstalledDevelLibs libguess
```
fbrosson: You might wish to use prepareInstalledDevel**Lib** (instead of prepareInstalledDevel**Libs**).

app-i18n/libguess/libguess-1.2.recipe
```diff
@@ -0,0 +1,66 @@
...
+	lib:libguess$secondaryArchSuffix = 1.2 compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libguess${secondaryArchSuffix}_devel = $portVersion compat >= 1.2
+	devel:libguess$secondaryArchSuffix = 1.2 compat >= 0
```
fbrosson: I think it should be:
```
	devel:libguess$secondaryArchSuffix = 1.0.0 compat >= 1
```

app-i18n/libguess/libguess-1.2.recipe
```diff
@@ -0,0 +1,66 @@
...
+REVISION="1"
+SOURCE_URI="http://rabbit.dereferenced.org/~nenolod/distfiles/libguess-$portVersion.tar.bz2"
+CHECKSUM_SHA256="8019a16bdc7ca9d2efbdcc1429d48d033d5053d42e45fccea10abf98074f05f8"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	libguess$secondaryArchSuffix = $portVersion compat >= 1.2
+	lib:libguess$secondaryArchSuffix = 1.2 compat >= 0
```
fbrosson: I think it should be:
```
	lib:libguess$secondaryArchSuffix = 1.0.0 compat >= 1
```

app-i18n/libguess/libguess-1.2.recipe
```diff
@@ -0,0 +1,66 @@
+SUMMARY="Character set guessing library"
+DESCRIPTION="Libguess is a high speed, open source character set encoding detector. \
+Libguess employs discrete-finite automata to deduce the character set of \
+the input buffer. The advantage of this is that all character sets can be \
+checked in parallel, and quickly."
```
fbrosson: The first line is 85 chars long... You can rewrap DESCRIPTION like this (to have lines <= 80 chars):
```
DESCRIPTION="Libguess is a high speed, open source character set encoding \
detector. Libguess employs discrete-finite automata to deduce the character \
set of the input buffer. The advantage of this is that all character sets can \
be checked in parallel, and quickly."
```

miqlas:
Fixed everything.

fbrosson:
> Fixed everything.

Cool, thanks!
I even noticed that you have already squashed your commits. That's nice, but it could be even nicer if you removed the "Fixed recipe." in the detailed commit message ;-)
Since this is a **_new_** recipe, a simple commit message (i.e. in one line <= 64 chars) would be perfect.

I haven't tried to build the recipe yet but might try to do so.
@waddlesplash, in case you have already built it for all architectures and decide to merge this PR, would you mind editing the commit message to get rid of the "Fixed recipe." line (in case you merge before @miqlas updates the commit message) ? Thanks!

Hmm, maybe we should ask (or request) our contributors to provide nice commit messages (and wait until they do so). Or maybe we should take care of "cleaning the dust" in commit messages left by **_new_** contributors but request that more regular contributors do the cleaning themselves.

fbrosson:
Good news: libguess builds on x86_gcc2 !
Bad news: there is no **`check`** target in the main Makefile. But there is a Makefile in src/tests/testbench which is supposed to run some tests, but it tries to launch src/tests/guess_testharness which needs libguess.so.1 but the runtime_loader fails to find libguess.so.1. Still digging...

app-i18n/libguess/libguess-1.2.recipe
```diff
@@ -0,0 +1,71 @@
...
+HOMEPAGE="http://atheme.org/projects/libguess.html"
+COPYRIGHT="2000-2003 Shiro Kawai
+	2009-2013 Tony Vroon"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://rabbit.dereferenced.org/~nenolod/distfiles/libguess-$portVersion.tar.bz2"
+CHECKSUM_SHA256="8019a16bdc7ca9d2efbdcc1429d48d033d5053d42e45fccea10abf98074f05f8"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
fbrosson: Should be:
```
```

app-i18n/libguess/libguess-1.2.recipe
```diff
@@ -0,0 +1,71 @@
...
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
```
fbrosson: Should be:
```
	make -C src/tests/testbench
```
(the current indentation, with one tab, is OK)

fbrosson: > Bad news: there is no check target in the main Makefile. But there is a Makefile in src/tests/testbench which is supposed to run some tests, but it tries to launch src/tests/guess_testharness which needs libguess.so.1 but the runtime_loader fails to find libguess.so.1. Still digging..

I managed to pass all tests. What I did was:
```
haikuporter -S libguess
cp /path/to/haikuports/packages/libguess*.hpkgs /path/to/haikuports/app-i18n/libguess/work-*/sources
haikuporter -E libguess
cp /sources/libguess*.hpkg /system/packages
make -C src/tests/testbench
exit
```
The problem is that haikuporter leaves the empty build package in the chrooted /system/packages/ after the build is finished. Maybe we should make haikuporter install the generated packages in the chrooted environment whenever **`haikuporter --test`** is run. I did it by hand and it allowed the tests to run.
BTW, I think this would fix the tests in a lot of recipes.

PR #835, opened 2016-10-24
WIP unibilium recipe

miqlas:
Work in Progress.
Now i get : `undefined reference to`__stack_chk_fail_local'`

Help required.

### DIFF
--- a/dev-libs/unibilium/unibilium-1.2.0.recipe
+++ b/dev-libs/unibilium/unibilium-1.2.0.recipe
@@ -13,16 +13,16 @@ ARCHITECTURES="x86 ?x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
-	unibilium$secondaryArchSuffix = $portVersion compat >= 1.0
-	lib:libunibilium$secondaryArchSuffix = $portVersion compat >= 1.0
+	unibilium$secondaryArchSuffix = $portVersion compat >= 0.3.0
+	lib:libunibilium$secondaryArchSuffix = $portVersion compat >= 0.3.0
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
 
 PROVIDES_devel="
-	unibilium${secondaryArchSuffix}_devel = $portVersion compat >= 1.0
-	devel:libunibilium$secondaryArchSuffix = $portVersion compat >= 1.0
+	unibilium${secondaryArchSuffix}_devel = $portVersion compat >= 0.3.0
+	devel:libunibilium$secondaryArchSuffix = $portVersion compat >= 0.3.0
 	"
 REQUIRES_devel="
 	libunibilium$secondaryArchSuffix == $portVersion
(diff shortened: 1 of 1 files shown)


miqlas:
Have anybody idea, what am i doing wrong? It compiles without a hitch outside of chroot...

fbrosson:
Maybe the recipe is missing a dependency.
Does «**`haikuporter --nochroot unibilium`**» work ?
BTW, «**`--nochroot`**» has a one-letter equivalent, «**`-C`**», so you can use that instead of «**`--nochroot`**».

dev-libs/unibilium/unibilium-1.2.0.recipe
```diff
@@ -0,0 +1,80 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtool$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	pkgconfig$secondaryArchSuffix
```
fbrosson: Should be:
```
```

fbrosson:
Oops, «**`haikuporter --nochroot ...`**» is not working for me. Maybe I need to upgrade something on my system.
Anyway, here's how you might identify a missing dependency:
* Re-build unibilium without haikuporter (or skip this step if you still have your build directory around), since you said it compiles without a hitch when you build without haikuporter.
* Then use «**`objdump -x /path/to/libunibilium.so | grep NEEDED`**» to see which shared libraries it needs.
* Use «**`pkgman search lib:libNAME`**» for every **`libNAME.so.X`** that you get with the previous command.
* Add one «**`lib:libNAME`**» for each library that is being provided by a package.

BTW, I havent tried this myself, so I'm not sure it works. Good luck!

miqlas:
fbrosson, the only objdump output is:
`NEEDED      libroot.so`

Now, what?

fbrosson:
This means libunibilium.so does not depend on any lib provided by other recipes. So the runttime dependencies seem OK.
Maybe you could compare the output of the configure script in the chrooted environment versus the non-chrooted environment.

dev-libs/unibilium/unibilium-1.2.0.recipe
```diff
@@ -0,0 +1,80 @@
...
+	cmd:ld$secondaryArchSuffix
+	cmd:libtool$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+export LDFLAGS="-lnetwork -lbsd -lroot"
+export CFLAGS="-D_BSD_SOURCE"
```
waddlesplash: These should be in `BUILD`, no?

waddlesplash:
@miqlas Travis build failed; please check.

miqlas:
Strange, but it builds just fine outside of Haikuporter, but inside it cannot find -lubsan and -lasan
Grepped the whole tree, nothing.
Must be a bug in the matrix or something.

miqlas:
Ok, it actually works now.

miqlas:
Merge?

PR #836, opened 2016-10-24
MSGPACK-C-CPP recipe

miqlas:
neovim requires it.


dev-libs/msgpack_c_cpp/msgpack_c_cpp-2.0.0.recipe
```diff
@@ -0,0 +1,94 @@
+SUMMARY="A serialization library"
+DESCRIPTION="An efficient binary serialization format, which lets you exchange data \
+among multiple languages like JSON, except that it's faster and smaller. \
```
fbrosson: Lines 2 and 3 should be:
```
DESCRIPTION="An efficient binary serialization format, which lets you exchange \
data among multiple languages like JSON, except that it's faster and smaller. \
```

dev-libs/msgpack_c_cpp/msgpack_c_cpp-2.0.0.recipe
```diff
@@ -0,0 +1,94 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:doxygen
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	pkgconfig$secondaryArchSuffix
```
fbrosson: Should be:
```
```

dev-libs/msgpack_c_cpp/msgpack_c_cpp-2.0.0.recipe
```diff
@@ -0,0 +1,94 @@
...
+	"
+
+PATCH()
+{
+	sed -i 's/-pthread/-lroot/g' "example/cpp03/CMakeLists.txt"
```
fbrosson: Double quotes are not required here, so you can safely change this to:
```
	sed -i 's/-pthread/-lroot/g' example/cpp03/CMakeLists.txt
```

dev-libs/msgpack_c_cpp/msgpack_c_cpp-2.0.0.recipe
```diff
@@ -0,0 +1,94 @@
...
+{
+	mkdir -p $includeDir
+	mkdir -p $libDir
+
+	cd build
+	make install
+
+	mv $libDir/../libmsgpackc.* $libDir/.
+	mv $libDir/../pkgconfig $libDir/.
+	mv $prefix/include/* $includeDir/
```
fbrosson: No need to keep trailing «**`/.`**» or «**`/`**». You can replace these lines by:
```
	mv $libDir/../libmsgpackc.* $libDir
	mv $libDir/../pkgconfig $libDir
	mv $prefix/include/* $includeDir
```

That said, if you patch some files and/or pass extra parameters to cmake it might be possible to make «**`make install`**» install everything to the correct destinations.

dev-libs/msgpack_c_cpp/msgpack_c_cpp-2.0.0.recipe
```diff
@@ -0,0 +1,94 @@
...
+	rm -rf $prefix/include
+
+	rm -rf $libDir/*.la
```
fbrosson: No need for «**`-r`**» for the line that removes the libtool files. You can use:
```
	rm -f $libDir/lib*.la
```

dev-libs/msgpack_c_cpp/msgpack_c_cpp-2.0.0.recipe
```diff
@@ -0,0 +1,94 @@
...
+	"
+
+PATCH()
+{
+	sed -i 's/-pthread/-lroot/g' example/cpp03/CMakeLists.txt
+}
+
+BUILD()
+{
+	export LDFLAGS="-lroot -lnetwork -lbsd"
```
waddlesplash: -lroot is not needed.

miqlas:
Ready!

PR #838, opened 2016-10-24
Haiku mime support for Fish

miqlas:
Use /bin/open on Haiku to open files with the associated programs from cli.

Slowly enabling advanced features in fish.....

### DIFF
--- a/app-shells/fish/patches/fish-2.3.1.patchset
+++ b/app-shells/fish/patches/fish-2.3.1.patchset
@@ -72,3 +72,34 @@ index 290975c..bcbb65a 100644
                      ! strcmp((const char*)p->ifa_name, interface))
 -- 
 2.7.0
+
+From 4fc160f28d37f7cf2dff76da2dca53cba3c1a646 Mon Sep 17 00:00:00 2001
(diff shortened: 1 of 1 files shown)


mmuman:
I already have a port of XDG utils around but it's not upstreamed nor packaged yet, so meanwhile it's useful to have this fallback.

miqlas:
mmuman, it is nice to see, that you follow the fish port developement. And thanks for your xdg port, but i don't really want an another layer to get the same results. But it is my own preference.

Btw, i sent a PR for fish-shell to get it upstreamed: https://github.com/fish-shell/fish-shell/pull/3487

mmuman:
Well, when the xdg-utils package will be installed it will take precedence anyway, but that's fine.

miqlas:
Not surely, check the upstream link for more info.

-- Z.

2016.10.24. dátummal, 18:39 időpontban François Revol <notifications@github.com> írta:

mmuman:
Yes, it does a ```else if type -q -f xdg-open```, before the test you added, so it will use xdg-open first.
Btw, on linux open is an alias for openvt, which does something completely different. You might want to actually check for Haiku...

PR #840, opened 2016-10-26
Vim 8 recipe with updated patchset

miqlas:
The patchset is just a dirty update, i'll try to get it a bit cleaner later.
It builds and works.

http://chunk.io/f/73233ee286834b58b3868a7c9439b593

### DIFF
--- a/app-editors/vim/vim-8.0.0046.recipe
+++ b/app-editors/vim/vim-8.0.0046.recipe
@@ -1,0 +1,103 @@
+SUMMARY="A highly configurable text editor"
+DESCRIPTION="Vim is a highly configurable text editor built to enable \
+efficient text editing. It is an improved version of the vi editor \
+distributed with most UNIX systems.
+
+Vim is often called a \"programmer's editor,\" and so useful for programming \
+that many consider it an entire IDE. It's not just for programmers, though. \
+Vim is perfect for all kinds of text editing, from composing email to editing \
+configuration files.
+Despite this, Vim can be configured to work in a very simple (Notepad-like) \
+way, called evim or Easy Vim."
+HOMEPAGE="http://www.vim.org"
+COPYRIGHT="Bram Moleenar et al."
+LICENSE="Vim"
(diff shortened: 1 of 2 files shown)


app-editors/vim/vim-8.0.0046.recipe
```diff
@@ -0,0 +1,103 @@
...
+	cmd:rvim = $portVersion
+	cmd:vi = $portVersion
+	cmd:view = $portVersion
+	cmd:vim = $portVersion
+	cmd:vimdiff = $portVersion
+	cmd:vimtutor = $portVersion
+	cmd:xxd = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
```
korli: Please drop $secondaryArchSuffix

app-editors/vim/vim-8.0.0046.recipe
```diff
@@ -0,0 +1,103 @@
...
+	lib:libiconv
+	lib:libintl
+	lib:libncurses
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
```
korli: Ditto

app-editors/vim/vim-8.0.0046.recipe
fbrosson: @miqlas, the reason why we don't need **`$secondaryArchSuffix`** is because the following conditions are **_all_** met:
* the recipe builds on all primary architectures,
* and the recipe provides no **`lib:lib*`** nor **`devel:lib*`**.

So if you have successfully built the recipe on x86_gcc2 primary arch (and x86 primary arch) then you should proceed with the removal of all occurences of **`$secondaryArchSuffix`** (and **`${secondaryArchSuffix}`** of course.)

BTW, if a recipe does not have a non empty SECONDARY_ARCHITECTURES then **`$secondaryArchSuffix`**  is always empty. This is why we prefer to remove them when they are not useful.

PR #841, opened 2016-10-26
Updated SDL2 to use dlopen instead of load_add_on, as discussed on haikuports ML

ksterker:
Have added the strictly required changes to the SDL2.0.5 patchset, but haven't removed files that are no longer needed from SDL2. Will submit a more complete patch to upstream, however.


waddlesplash:
Thanks!

PR #846, opened 2016-10-30
Updated mpd recipe.

miqlas:
No functional change


fbrosson:
While at it, could you please add a **`$secondaryArchSuffix`** to the **`cmd:libtoolize`** in line 85? Thanks!

korli:
Could you please provide commit message?  port name: description of the change.
For update for instance: "mpd: bump version".

fbrosson:
@miqlas: I think you can drop the detailed commit message and use a short commit message instead.
What about:
```
cd /path/to/haikuports
git checkout mpd_update
git commit --amend -m "mpd: bump to 0.19.19, add 2 missing secondaryArchSuffix."
git push -f
```

BTW, if you like GitHub you should delte all your unused remote branches, that is, those which have already been merged. To do so you can point your browser to https://github.com/miqlas/haikuports/branches and click on the delete icon for most of your remote branches (other than master), for example: libuv, libguess, exiftool. Be careful, however, because if you delete a remote branch which has an active PR then I don't know if you can push it back from your local branch.

miqlas:
Guys, correct me if i'm wrong, but are you talking about the commit message?

fbrosson:
Yes. BTW, it is good practice to write good commit messages. A good commit message is something that is short enough to be understood, and which does not show information that is not necessary.
The best rule is to ask yourself what information you think needs to appear in the commit message for someone to have an overview of the changes.
The other rule is to have lines of less than 70 chars (or maybe even less than 65 chars).
And for the very first line of the commit message, if the commit is associated to a PR then you should make the first line fit in 64 chars (or less), otherwise it gets split.

fbrosson:
BTW, the commit message suggest by @korli ("mpd: bump version") is even better. No need, after all, to mention that 2 missing secondaryArchSuffix were added.
Personaly I would use "mpd: bump to 0.19.19." but some users prefer to have just "mpd: bump version". I think both are OK.
Chose the one you prefer, and please do not hesitate to take @korli's suggestion. He has far more contributions than me so you should probably follow his suggestion ;-)

PR #850, opened 2016-10-31
Updated libsigc++ recipe

miqlas:
No functional change.

### DIFF
--- a/dev-libs/libsigc++/libsigc++3-2.99.7.recipe
+++ b/dev-libs/libsigc++/libsigc++3-2.99.7.recipe
@@ -1,0 +1,66 @@
+SUMMARY="Callback Framework for C++"
+DESCRIPTION="A typesafe callback system for standard C++. It allows you to \
+define signals and to connect those signals to any callback function, either \
+global or a member function, regardless of whether it is static or virtual.
+
+It also contains adaptor classes for connection of dissimilar callbacks and \
+has an ease of use unmatched by other C++ callback libraries."
+HOMEPAGE="http://libsigc.sourceforge.net/stable.shtml"
+COPYRIGHT="2002-2015 The libsigc++ Development Team"
+LICENSE="GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="http://ftp.gnome.org/pub/GNOME/sources/libsigc++/2.99/libsigc++-$portVersion.tar.xz"
+CHECKSUM_SHA256="e819884fec16ee20f97773361258b9a911580d5b64fd9da981ef0c80356ff5f9"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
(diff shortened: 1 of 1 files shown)


miqlas:
I made a commit mistake and it is already rebased so i have no idea, how can i fix it. Help wanted.
Thanks!

fbrosson:
If the following command:
```
git remote -v | grep upstream
```
outputs this:
```
upstream        https://github.com/haikuports/haikuports.git (fetch)
upstream        https://github.com/haikuports/haikuports.git (push)
```
then the following commands should work:
```
git checkout master
git pull upstream master
git push
git checkout libsigc_update
git merge --no-edit upstream/master
git rebase upstream/master
git push -f
```

dev-libs/libsigc++/libsigc++-2.99.7.recipe
```diff
@@ -55,12 +54,13 @@ INSTALL()
 {
 	make install
 
-	prepareInstalledDevelLib libsigc-2.0
+	prepareInstalledDevelLib \
+		libsigc-3.0
```
fbrosson: Comment about style: For prepareInstalledDevel**Libs** it's a good idea to split the line, but for prepareInstalledDevel**Lib** I think it is better to keep the library name on the same line as the command.

korli:
This update breaks compatibility for dependencies of libsigc++-2.4.1. This shouldn't be an update but a new recipe, for instance libsigc++3 (the new API is libsigc++-3.0). Alternative is to provide the libsigc++-2.0 API in a libsigc++2-2.4.1 recipe.

dev-libs/libsigc++/libsigc++-2.99.7.recipe
fbrosson: >  i just tried your proposed fix, but it doesn't help much.

I guess you are talking about the git commands. Does the command with the **`git remote -v`** yeld what I was expecting? If not, could you run:
```
cd /path/to/haikuports
git remote -v | grep haikuports/haikuports
```
and copy/paste the result here? (I just need the two lines of output).
I think you'll get this:
```
real_origin        https://github.com/haikuports/haikuports.git (fetch)
real_origin        https://github.com/haikuports/haikuports.git (push)
```
It should be 2 lines which both start with the same word, e.g. "real_origin". That word is what you should use instead of the word "upstream" in the commands of my first message in this PR.

But don't worry if you are lost. We can always do as follows:
a) I grab your remote branch, update it localy, do the rebase against current master and then push it back to your remote branch;
b) then you delete your local branch and grab a fresh copy of your newly updated (by me) remote branch.
So if you want to try this, just let me know and as soon as I have finished step a) I'll send you the commands that you should type for step b).

Of course, you'll have to address @korli's comment. (Later I can try to help you with that too if you wish.)

miqlas:
Renamed to libsigc++3. Hopefully it is ok now.

PR #851, opened 2016-11-01
lighttpd: bump to 1.4.42, with mod_geoip and 6 other modules.

fbrosson:
* This release of lighttpd comes with 7 new modules: mod_deflate, mod_authn_{file,gssapi,ldap,mysql}, mod_uploadprogress and mod_geoip.
* The GeoIP module is being added in a separate sub-package to limit the dependency on the geoip package to that sub-package only.
* Add a quickstart-geoip.txt and a demo geoip page for novice users.
* In lighttpd.conf don't use paths that change when the package is updated.
* Keep lighttpd.pid in **`/var/run`** instead of **`/var`**.
* Switch SOURCE_URI from tar.gz to tar.xz ;-)

### DIFF
--- a/www-servers/lighttpd/additional-files/create-lighttpd-directories.sh
+++ b/www-servers/lighttpd/additional-files/create-lighttpd-directories.sh
@@ -11,3 +11,4 @@ if [ ! -e var/www/htdocs/index.lighttpd.html ]; then
 fi
 
 mkdir -p var/log/lighttpd var/lib/lighttpd var/cache/lighttpd var/tmp/lighttpd
+mkdir -p var/run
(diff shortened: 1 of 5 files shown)


fbrosson:
Comments and reviews welcome!
BTW, please do not hesitate to make suggestions for improving the wording in quickstart.txt and quickstart-geoip.txt, as well as in the comments in the recipe.

I found that creating a file (e.g. lighttpd_userlaunch) in $dataDir/user_launch/ with the following contents:
```
target desktop {
	service x-vnd.lighttpd {
		launch /system/bin/lighttpd -f /system/settings/lighttpd/lighttpd.conf
		no_safemode
		legacy
		ondemand
		initial_volumes_mounted
	}
}
```
allows to autostart the server at boot time, but I did not include it in the package because I did not find how to let the user enable or disable the autostart.
I guess it would need some glue code in Network to have lighttpd shown in the list along with the services that it can already control:
* FTP server
* SSH server
* Telnet server

If anyone has some info or pointers for that, I'd be happy to improve the recipe.

BTW, I wanted to bump to 1.4.43 (released yesterday) but I've hit an issue in src/Makefile when it tries to compile a new source file, lemon.c. I need to dig this. Help welcome, BTW ;-)

fbrosson:
Thanks!

PR #855, opened 2016-11-04
curl: bump to 7.51.0, drop html documentation.

fbrosson:
Upstream does not ship **`docs/libcurl/*.html`** anymore. Although these may be generated with "make html", we would need roffit or any other MAN2HTML conversion utility, so let's just drop the html doc.


fbrosson:
Thanks!

PR #859, opened 2016-11-06
qt5: switch to tar.xz, move rdef files to additional-files.

fbrosson:
The 6 rdef files are now used by both qt5-5.5.1 and qt5-5.6.1.
Also cleaned a few trailing white spaces or tabs.


fbrosson:
With this commit the 6 rdef files that were in the patches for both qt5-5.5.1 and qt5-5.6.1 are now in additional-files.

BTW, I only switched 5.6.1 (from tar.gz to tar.xz), since haikuporter is now able to decompress xz tarballs on the fly, thanks to @korli!
Both archives have exactly the same tar inside:
```
cd /path/to/haikuports/dev-qt/qt5/downloads
( gzip -cd qt-everywhere-opensource-src-5.6.1.tar.gz | sha256sum && \
  xz   -cd qt-everywhere-opensource-src-5.6.1.tar.xz | sha256sum )  \
| uniq -c
      2 579e9ed7d5a22814102c1f71179c515895f55f22222f9b7652f970800d0faa2a  -
```

Hmm, I did not switch 5.5.1 to tar.xz. Should I do so?

mmuman:
Ideally the rdef files would be upstreamed like the rest...

mmuman:
Just when I'm trying to build Qt for the 3rd night in a row...

fbrosson:
> Just when I'm trying to build Qt for the 3rd night in a row...

Oops, sorry. The objective was to save space, now that haikuporter is able to uncompress xz tarballs on the fly. The xz tarball (285MB) is 25% smaller than the gz tarball (383MB).

> Ideally the rdef files would be upstreamed like the rest...

I'm not sure upstream would be willing to import the rdef files...
But I think that moving them to additional-files will make it easier to maintain the patchsets.

BTW, when I started the tar.gz to tar.xz switch I had no intention to change anything else, except synchronising the patchset (with **`haikuporter -b`** + **`haikuporter -e`** + vim) in case some patch offsets would be outdated. So I had a look at the patchset and noticed that the same set of rdef files was burried in the two patchsets.

Now, since Qt5 is huge, shall we keep this PR open until you finish (and push) your changes?
That's OK for me. I can wait and then rebase this PR against the new recipe(s).

fbrosson:
Thanks!

mmuman:
Ok I'll be pushing some more patches soon :)

fbrosson:
Cool!
BTW, sorry for this early merge.
I swear I did not bribe anybody ;-)

mmuman:
Ok, done. If anyone wants to port them to 5.5 as well...

fbrosson:
I think I'll try, later, to port your new patches to 5.5.

PR #861, opened 2016-11-06
Adding new port SimpleBackup v2.0.1

Perelandra0x309:
Request to add SimpleBackup which I have updated to run on Haiku nightly image.

### DIFF
--- a/haiku-apps/simplebackup/simplebackup-2.0.1.recipe
+++ b/haiku-apps/simplebackup/simplebackup-2.0.1.recipe
@@ -1,0 +1,60 @@
+SUMMARY="Graphical interface to run the zip command"
+DESCRIPTION="This application makes it easy to create and update backups of \
+large amounts of data.  For example, you can create a backup of your home \
+directory and run this application regularly to make sure the backup is up to \
+date.  SimpleBackup also accepts command line options, so it is easily \
+scriptable and can be used with scheduling applications."
+HOMEPAGE="https://perelandra0x309.github.io/simplebackup"
+COPYRIGHT="2010-2016 Brian Hill"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/Perelandra0x309/simplebackup/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="463df905df35a74e81127d1c28d20a1f2c94ff1b89f6939abf081ecc70e60c62"
+
+ARCHITECTURES="x86_gcc2 !x86"
+
+PROVIDES="
+	simplebackup = $portVersion
+	app:SimpleBackup = $portVersion
+	cmd:SBQueue = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	"
+
+USER_SETTINGS_FILES="settings/SimpleBackup directory"
+
+BUILD()
+{
+	cd src
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/SimpleBackup
+	mkdir -p $binDir
+
+	# copy documentation
+	cp -af ReadMe License $appsDir/SimpleBackup
+
+	# copy executables
+	cp -af SimpleBackup $appsDir/SimpleBackup
+	cp -af SBQueue $binDir
+
+	# copy example files
+	cp -af home_backup_settings backup_example.sh $appsDir/SimpleBackup
+	chmod +x $appsDir/SimpleBackup/backup_example.sh
+
+	# deskbar link
+	addAppDeskbarSymlink $appsDir/SimpleBackup/SimpleBackup
+}


haiku-apps/simplebackup/simplebackup-2.0.1.recipe
```diff
@@ -0,0 +1,61 @@
...
+directory and run this application regularly to make sure the backup is up to \
+date.  SimpleBackup also accepts command line options, so it is easily \
+scriptable and can be used with scheduling applications."
+HOMEPAGE="https://perelandra0x309.github.io/simplebackup"
+COPYRIGHT="2010-2016 Brian Hill"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/Perelandra0x309/simplebackup/archive/v${portVersion}.tar.gz"
+CHECKSUM_SHA256="463df905df35a74e81127d1c28d20a1f2c94ff1b89f6939abf081ecc70e60c62"
+SOURCE_DIR="simplebackup-${portVersion}"
```
fbrosson: @Perelandra0x309: You can (or should?) drop the line that defines SOURCE_DIR because it is the same as the default value.
BTW, in SOURCE_URI you can use «**`v$portVersion.tar.gz`**» (instead of «**`v${portVersion}.tar.gz`**») because the curly braces are not required.

humdingerb: IMO, The SOURCE_DIR could really be dropped. The curlies in SOURCE_URI would only be needed if there were a space in the portVersion or some other special characters after them (?). If the {} aren't ususally use, we maybe shouldn't do so here either.

fbrosson: @humdingerb: That's great! Thanks for your approval of my 2 suggestions!
(BTW, I knew that my 2 suggestions were OK, and I was almost sure that you would approve them both ;-) The reason why I asked for your thoughts was to let @Perelandra0x309 know that you approved them ;-)

humdingerb: "I'm Humdinger, and I approve this message." ;)

Perelandra0x309: Sorry for the delay in response, I had meant to try it without the SOURCE_DIR last night because I thought when I had tried it without it did not work, but time just got away from me.  If it runs fine feel free to remove. I'll be able to try myself about 02:00 GMT tonight.
The curly brackets are probably left over from a different recipe I used as a template.

fbrosson: No problem ;-) Don't worry, you can make the changes anytime. BTW, you can even wait several days (or weeks) if you are busy.
We are very happy (and lucky) to have contributors like you, so we're not going to blame anyone for not updaing PRs quickly ;-)

Perelandra0x309: I do not have an x86 or 64 install yet but I might soon!  I just got a Core2 quad setup with a second HD that I will dedicate only to Haiku and maybe some old BeOS installs for nostalgia.

Perelandra0x309: Which x86 gcc4 build should I install, the non-hybrid?

fbrosson: Well, I would recommend using x86 (gcc4) hybrid. This allows to build for both:
* x86 as primary arch,
* and x86_gcc2 as secondary arch.

Whereas if you chose x86 (gcc4) **_non_** hybrid, then you will only be able to build for x86 (gcc4) as primary arch.

BTW, I have 3 bootable partitions:
* x86_gcc2 hubrid
* x86 hybrid (i.e. the one mentionned in the beginning of this message)
* x86_64/
and a "data" partition where I have the haikuports and haikuporter trees.

Then, with a few symbolic links, I managed to have separate directories for both:
/data/haikuports/packages
/data/haikuports/repository

humdingerb: It's a good idea to have a gcc4hybrid and 64bit Haiku for testing/building.
The first thing I do however, is to build the gcc4-as-2nd-arch in the standard gcc2hybrid with "hp simplebackup_x86". If that fails badly, i.e. not easily fixable, I don't have to bother with testing under primary gcc4, let alone 64bit, and can justify a "!x86" in the ARCHITECTURES.

fbrosson: In case anyone would like to use the same haikuports tree with all their "flavour" architectures, here's what I have setup:

* In my "data" partition I have 2 symbolic links:
/data//haikuports/packages -> /home/packages
/data//haikuports/repository -> /home/repository
as well as 6 directories:
/data//packages.x86_gcc2
/data//packages.x86
/data//packages.x86_64
/data//repository.x86_gcc2
/data//repository.x86
/data//repository.x86_64

* In my "Haiku-x86_gcc2" bootable partition I have 2 symbolic links:
/boot/home/packages -> /data/packages.x86_gcc2
/boot/home/repository -> /data/repository.x86_gcc2

* In my "Haiku-x86" bootable partition I have 2 symbolic links:
/boot/home/packages -> /data/packages.x86
/boot/home/repository -> /data/repository.x86

* In my "Haiku-x86_64" bootable partition I have 2 symbolic links:
/boot/home/packages -> /data/packages.x86_64
/boot/home/repository -> /data/repository.x86_64

All this allows to have separate packages and repository subdirectories, and yet, share the same haikuports tree. I came to the conclusion that this was the easiest way to have a single haikuports tree and keep things working. During my setup I understood that:
* Mixing the packages directories from the 3 platforms is not a good idea.
* Mixing the repository directories from the 3 platforms is a very bad idea.

All I have to do when I reboot into another architecture is remember to clean the work-* subirectories, especially when I build for primary architectures.

@Perelandra0x309: sorry for posting here what I should put in a wiki...

Perelandra0x309: Thanks for this info, I was just expecting a simple answer but got a lot of useful tips here.  Yes this should definitely be in a wiki.  I got my box all setup and ready to start playing with some other architectures.

fbrosson: Yep, I know. Will try to do so soon.

PR #867, opened 2016-11-12
hunspell: bump to 1.4.2, rename analyze, chmorph and {un,}munch.

fbrosson:
Gentoo renames four utilities (analyze, chmorph, munch, unmunch) by adding a **`hunspell-`** prefix to their names, to avoid having too generic names. So let's follow Gentoo, using their patch ;-)
Also remove any empty **`data/locale/*/LC_MESSAGES`** dirs (hu & it).


app-text/hunspell/hunspell-1.4.2.recipe
```diff
@@ -73,6 +83,10 @@ INSTALL()
 	make install
 
 	rm $libDir/charset.alias $libDir/*.la
+
+	# remove any existing (but empty) data/locale/*/LC_MESSAGES directory
+	rmdir $dataDir/locale/*/LC_MESSAGES $dataDir/locale/* || true
```
fbrosson: I did _not_ dare to also
```
	rm $dataDir/locale/locale.alias
```
Shall I do so, before the **`rmdir`**, and then also **`rmdir $dataDir`** ?
Because the only file in **`$dataDir`** is **`data/locale/locale.alias`**, but it either gets shadowed by the **`locale.alias`** from another package or shadows all **`locale.alias`** from the other packages.

Reminder: on some other packages we already drop **`data/locale/charset.alias`**, so I was wondering if we should also treat this **`locale.alias`** file same way we already treat the **`charset.alias`** file.
Thanks!

korli: No idea. I would say, if it's empty it can be removed.

fbrosson: It is not empty. It has some comments and then 47 lines of aliases for locales.
BTW, the last comment is «**`# Packages using this file: hunspell`**». 
I guess we should keep the **`locale.alias`** file, at least until we add a hunspell dictionary and test hunspell without the **`locale.alias`** file.

korli:
1.5.3 is out.

fbrosson:
Cool, thanks! I'll try 1.5.3.
BTW, a few days ago I noticed that 1.5.**2** was out so I was trying to build it, but I've hit an issue:
```
In file included from xsize.c:3:0:
xsize.h:31:3: error: #error "Please include config.h first."
  #error "Please include config.h first."
```

fbrosson:
I've added 1.5.2 and 1.5.3 in case someone would like to have a look at it. Both fail to build.

waddlesplash:
Should we merge the recipe anyway, as broken?

fbrosson:
As you wish. Le tme see if the latest 1.5.x release works better.
If not then I guess that bumping to the working 1.4.2 and adding the latest 1.5.x marked as broken is what we need. (I guess this is what you are suggesting.)
Just give me half an hour and I'll update my PR.

fbrosson:
1.6.0 was released 3 days ago. I'll try it.

fbrosson:
That's it.
I did not touch 1.4.2, which was already working when I created this PR.

I've included 1.5.4 and 1.6.0, which both fail to build, so I marked them as broken.
We can merge this PR as is or without the broken releases.
But I guess it might be worth having them, so that anyone willing to investigate on these newer releases won't need to redo the packaging.

waddlesplash:
Sounds good.

PR #868, opened 2016-11-12
ImageMagick: bumped version

miqlas:
The old source dissapeared.


media-gfx/imagemagick/imagemagick-6.9.6.4.recipe
```diff
@@ -8,10 +8,10 @@ draw text, lines, polygons, ellipses and Bézier curves."
 HOMEPAGE="http://www.imagemagick.org"
```
fbrosson: Could you please change this to:
```
HOMEPAGE="https://www.imagemagick.org/"
```
(Please don't forget the trailing slash ;-)
Thanks!

PR #870, opened 2016-11-13
QtAV : new recipe

miqlas:


media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
+SUMMARY="A cross-platform multimedia framework"
+DESCRIPTION="
+QtAV is a multimedia playback library based on Qt and FFmpeg. \
```
fbrosson: It is better to join DESCRIPTION and its first line:
```
DESCRIPTION="QtAV is a multimedia playback library based on Qt and FFmpeg. \
```
(For already existing recipes, sometimes we don't bother to respect that tiny rule, but for new recipes we should do so.)

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+SOURCE_DIR="QtAV-1.11.0"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qtav$secondaryArchSuffix = $portVersion
+	lib:libqtav$secondaryArchSuffix = 1.11.0 compat >= 1
+	lib:libqtavwidgets$secondaryArchSuffix = 1.11.0 compat >= 1
+
```
fbrosson: No need for an empty line here.

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+	lib:libgl$secondaryArchSuffix
+	lib:libopenal$secondaryArchSuffix
+	lib:libuchardet$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	qtav$secondaryArchSuffix = $portVersion
+	devel:libqtav$secondaryArchSuffix = 1.11.0 compat >= 1
+	devel:libqtavwidgets$secondaryArchSuffix = 1.11.0 compat >= 1
+
```
fbrosson: No need for an empty line here.

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+* Video eq(software and OpenGL): brightness, contrast, saturation, hue
+* QML support. Most playback APIs are compatible with QtMultimedia module
+* Compatiblity: QtAV can be built with both Qt4 and Qt5, FFmpeg(>=1.0) \
+and Libav (>=9.0). Latest FFmpeg release is recommended.
+
+Extensible Framework
+
+Some components in QtAV are designed to be extensible. For example, \
+you can write your decoder, audio output for particular platform."
+HOMEPAGE="http://qtav.org/"
```
fbrosson: It would be better to insert a "www." in order to have:
```
HOMEPAGE="http://www.qtav.org/"
```
because
```
curl -I http://qtav.org/
```
yelds
```
HTTP/1.1 301 Moved Permanently
Server: GitHub.com
Content-Type: text/html
Location: http://www.qtav.org/
```

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qtav$secondaryArchSuffix = $portVersion
+	lib:libqtav$secondaryArchSuffix = 1.11.0 compat >= 1
+	lib:libqtavwidgets$secondaryArchSuffix = 1.11.0 compat >= 1
+
+	"
+REQUIRES="
+	haiku${secondaryArchSuffix}
```
fbrosson: You might wish to replace this line by:
```
	haiku$secondaryArchSuffix
```

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+	"
+
+PROVIDES_devel="
+	qtav$secondaryArchSuffix = $portVersion
+	devel:libqtav$secondaryArchSuffix = 1.11.0 compat >= 1
+	devel:libqtavwidgets$secondaryArchSuffix = 1.11.0 compat >= 1
+
+	"
+REQUIRES_devel="
+	haiku${secondaryArchSuffix}
```
fbrosson: You might wish to replace this line by:
```
	haiku$secondaryArchSuffix
```

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+}
+
+INSTALL()
+{
+	cd build
+
+	mkdir -p $libDir
+	mkdir -p $includeDir/{QtAV/5.5.1/QtAV,QtAVWidgets}
+	mkdir -p $dataDir/Qt5/mkspecs/features
+	mkdir -p $dataDir/qml/QtAV/
```
fbrosson: You can drop the trailing slash for the mkdir.

(You could also drop the trailing slash for some of the cp commands below, but that's fine to keep them for these cases, since it makes it clear that the target is a directory ;-)

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+
+	cp -f lib_unix_x86/*Qt*AV*.so* $libDir/
+	cp -R tools/install_sdk/mkspecs $dataDir/Qt5/
+	cp -R ../src/QtAV/ $includeDir/
+	cp -R ../widgets/QtAVWidgets/ $includeDir/
+	cp -f -R bin/QtAV $dataDir/qml
+	cp -f ../qml/plugins.qmltypes $dataDir/qml/QtAV/
+
+	prepareInstalledDevelLibs libQtAV \
+		libQtAVWidgets
```
fbrosson: IMHO it would be nicer to spawn the prepareInstalledDevelLibs across 3 lines:
```
	prepareInstalledDevelLibs \
		libQtAV \
		libQtAVWidgets
```

media-video/qtav/qtav-1.11.0.recipe
```diff
@@ -0,0 +1,152 @@
...
+	qmake ..
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+
+	mkdir -p $libDir
+	mkdir -p $includeDir/{QtAV/5.5.1/QtAV,QtAVWidgets}
```
fbrosson: If "5.5.1" is the version of Qt you have in your system then I think it should not be hard-coded in this recipe.
Unfortunately I don't know how to obtain the version from the installed qt5 package.
Does someone know how this should be handled?

pulkomandy: Why is the INSTALL rule not using make install? If that doesn't work, our qmake should be fixed to do the right thing there.
Otherwise, maybe a hack with qtconfig or pkgconfig or similar is possible?

miqlas:
Tried my best to get it work without much hackery.

PR #873, opened 2016-11-15
bash: bump to 4.4.005, simplify patch and rename it to patchset.

fbrosson:


fbrosson:
Did someone succeed in building bash 4.4 ?
I'm asking this because I get the same errors for 4.4 than what I get for 4.4.005:
```
bashhist.c:324: undefined reference to history_lines_read_from_file
bashline.c:4103: undefined reference to rl_clear_visible_line
bashline.c:4151: undefined reference to rl_redraw_prompt_last_line
```
(4.3.48 builds fine, however)

korli:
lib:libreadline requirement should be >= 7. From https://lists.gnu.org/archive/html/info-gnu/2016-09/msg00012.html:

> Only readline-7.0 and later versions are able to provide all of the symbols that bash-4.4 requires;
earlier versions of the Readline library will not work correctly.

korli:
Otherwise +1

fbrosson:
> lib:libreadline requirement should be >= 7.

Cool, great! Fixed (but not tested yet.) Thanks a lot!
BTW, how come did I not think of using Google ?!!!

fbrosson:
Thanks!

PR #874, opened 2016-11-16
gettext: drop libtool files, update gettext compat to >= 0.19.

fbrosson:
Also drop devel:lib{iconv,ncurses} from REQUIRES.


fbrosson:
Thanks!

PR #880, opened 2016-11-23
libmtp: bump to 1.1.12, install to bin, comment out cmd:doxygen.

fbrosson:
Install **`cmd:*`** to **`$prefix/bin`** (instead of **`$binDir`**), drop libtool file, fix SUMMARY, add copyright years, sort BUILD_PREREQUIRES, add TEST.


fbrosson:
Thanks!

PR #881, opened 2016-11-25
libgpg_error: bump to 1.25.

fbrosson:


korli:
SONAME bump?

fbrosson:
> SONAME bump?

Oops, good catch. Fixed. Thanks!

PR #883, opened 2016-11-26
imagemagick: bump to 6.9.6.6. Improve layout. Minor changes.

fbrosson:
* Remove **`lib:libgomp`** from REQUIRES on x86_gcc2 primary arch, since gcc_syslibs_devel does not provide it on that arch.
* Remove unneeded cmd:svnversion from BUILD_PREREQUIRES.
* Drop libtool files.
* Add TEST with **`make check`**.
* Improve layout by moving some items from the base package to the _devel subpackage: **`lib/pkgconfig/*.pc`** (thanks to fixPkgconfig), **`bin/*-config`** (and their associated man1 manual pages), and some docs: **`data/doc/ImageMagick-6/www/api{.html,/}`**.
* Add two extra download locations to SOURCE_URI to make sure the tarball will still be available after a new version is released.

### DIFF
--- a/media-gfx/imagemagick/imagemagick-6.9.6.6.recipe
+++ b/media-gfx/imagemagick/imagemagick-6.9.6.6.recipe
@@ -9,9 +9,11 @@ HOMEPAGE="https://www.imagemagick.org/"
 COPYRIGHT="1999-2016 ImageMagick Studio LLC"
 LICENSE="IMAGEMAGICK"
 REVISION="1"
-SOURCE_URI="http://www.imagemagick.org/download/releases/ImageMagick-6.9.6-4.tar.xz"
-CHECKSUM_SHA256="d523c2c45dc2ff55f1bc27aa0cfa6e00e6f80dc182e8d6c27fcac99461c462d7"
-SOURCE_DIR="ImageMagick-6.9.6-4"
+SOURCE_URI="http://www.imagemagick.org/download/releases/ImageMagick-6.9.6-6.tar.xz
+	https://sourceforge.net/projects/imagemagick/files/old-sources/6.x/6.9/ImageMagick-6.9.6-6.tar.xz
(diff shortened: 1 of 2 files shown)


fbrosson:
Oops, I just noticed that 6.9.6.6 is out, so I'll update this PR.

korli:
I find problematic that for each update the download link becomes invalid. Is there an archived version somewhere?

fbrosson:
I don't know.

korli:
Maybe add another URL for instance: http://gentoo.mirrors.ovh.net/gentoo-distfiles/distfiles/ImageMagick-6.9.6-6.tar.xz

fbrosson:
Hmm, do you have Gento's URL before the redirection to gentoo.mirrors.**OVH**.net ?

BTW, I found the old tarballs in downloads.sourceforge.net, under /project/imagemagick/old-sources/6.x/6.9/.
But 6.9.6.6 will only be moved to that directory after 6.9.6.7 becomes available. And that directory only has tar.gz tarballs... But I guess switching from tar.xz to tar.gz is not desirable.

If you have a URL which does not rely on a specific _effective_ server then I think it is OK do add it, otherwise I would rather keep a single SOURCE_URI for this recipe. What do you think?

korli:
The URL comes from the download mirror list: https://www.gentoo.org/downloads/mirrors/. No redirect URL is available.

fbrosson:
OK, fine. I'll add your suggestion right now.
BTW, I was wondering if we should also add the URL from SF, right before the one from OVH, just to make sure we don't break any policy we might not be aware of regarding the usage of mirrors.

fbrosson:
@korli, is SOURCE_URI ok as it is? Shall we merge this?
(I've successfully built this on all architectures.)

korli:
Fine with me!

fbrosson:
Cool, thanks!

PR #891, opened 2016-11-30
libsodium: drop libtool file, configure with --disable-ssp.

fbrosson:
* **`configure --disable-ssp`** to avoid the dependency on libssp.
* Enable x86_64.
* Use 18.1.1 for the lib version, instead of **`$portVersion`**.
* Replace **`cmd:libtool`** by **`cmd:libtoolize$secondaryArchSuffix`**.
* Change prepareInstalledDevel**Libs** to prepareInstalledDevel**Lib**, since we have only one lib here.

### DIFF
--- a/dev-libs/libsodium/libsodium-1.0.11.recipe
+++ b/dev-libs/libsodium/libsodium-1.0.11.recipe
@@ -4,42 +4,45 @@ fork of NaCl, with a compatible API, and an extended API to improve usability \
 even further.
 Its goal is to provide all of the core operations needed to build higher-level \
 cryptographic tools.
-The design choices emphasize security, and \"magic constants\" have clear rationales.
+The design choices emphasize security, and \"magic constants\" have clear \
+rationales.
 Despite the emphasis on high security, primitives are faster across-the-board \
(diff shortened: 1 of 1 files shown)


fbrosson:
I'm getting this error when I call **`haikuporter -S libsodium`** on x86_64:
```
POLICY ERROR: can't find used library "libssp.so.0"
POLICY ERROR: can't find used library "libssp.so.0"
Error: packaging policy violation(s) in strict mode
```
BTW **`pkgman search lib:libssp`** does not know that libssp is in **`/system/develop/tools/lib/gcc/x86_64-unknown-haiku/5.4.0/`**.
Does someone know how this issue should be handled?
Thanks!

pulkomandy:
The gcc packages provides should be fixed.

fbrosson:
OK, thanks @pulkomandy!
(My x86_64 is still on hrev50701, so I'll update to the latest hrev, just in case this fixes the issue.)

korli:
@fbrosson  it should be configured with --disable-ssp.

fbrosson:
Thanks @korli!

PR #892, opened 2016-11-30
Fish shell: bump version, enabled system PCRE2.

miqlas:


korli:
Hi miqlas, just name the recipe fish-2.5.0~git.recipe because 2.4.0~git means sort of alpha of 2.4.0. Whereas  I think it's a beta for the next release. Also usually we don't enable the architectures on non stable recipes (when a stable version exists).

miqlas:
Mistakenly tagged on fish github page. Closing for now and waiting till the next stable release.

PR #893, opened 2016-11-30
Adding Einsteinium 1.3.2 recipe

Perelandra0x309:
Recipe for new version, also now builds for both x86_gcc2 and x86.


humdingerb:
Looked good aside from to 80char limit violations.
As I always wanted to try github's online editor, I corrected that and see if it all checks out with merge&squashing the commits...

PR #894, opened 2016-11-30
Gif2apng

R167:
Added in gif2apng recipe. Does not support gcc2.

### DIFF
--- a/media-gfx/gif2apng/gif2apng-1.9.recipe
+++ b/media-gfx/gif2apng/gif2apng-1.9.recipe
@@ -1,0 +1,43 @@
+SUMMARY="Converts GIF animations into APNG format"
+DESCRIPTION="This program converts GIF animations into animated PNG format. \
+Usually it makes the files smaller."
+HOMEPAGE="http://gif2apng.sourceforge.net/"
+COPYRIGHT="2009-2013 Max Stepin"
+LICENSE="Zlib"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/gif2apng/gif2apng-$portVersion-src.zip"
+CHECKSUM_SHA256="3b21308e935d799b3ffb4a86c6e00ffa4cb9b3f72f52d58d51c66eb0574ae7d2"
+SOURCE_DIR=""
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	gif2apng$secondaryArchSuffix = $portVersion
+	cmd:gif2apng = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:ld$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $prefix/bin
+	cp gif2apng $prefix/bin
+}


media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,44 @@
...
+SOURCE_URI="https://sourceforge.net/projects/gif2apng/files/$portVersion/gif2apng-$portVersion-src.zip"
+CHECKSUM_SHA256="3b21308e935d799b3ffb4a86c6e00ffa4cb9b3f72f52d58d51c66eb0574ae7d2"
+SOURCE_DIR=""
+
+ARCHITECTURES="x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	gif2apng$secondaryArchSuffix = $portVersion
+	cmd:gif2apng$secondaryArchSuffix = $portVersion
```
fbrosson: Then we drop the **`$secondaryArchSuffix`** from **`cmd:gif2apng$secondaryArchSuffix`** to let haikuporter know that it should not expect anything in $prefix/bin/x86/, since we're installing to $prefix/bin/.
You should therefore have:
```
	cmd:gif2apng = $portVersion
```

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,44 @@
...
+
+BUILD()
+{
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $binDir
+	mv gif2apng $binDir
```
fbrosson: Should be:
```
        mkdir -p $commandBinDir
        cp gif2apng $commandBinDir
```
The reason why we prefer using cp instead of mv is that mv would break **`haikuporter -F`**.

fbrosson:
There are a couple of things to change. I'm writing a review.

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,44 @@
...
+COPYRIGHT="2009-2013 Max Stepin"
+LICENSE="Zlib"
+REVISION="1"
+SOURCE_URI="https://sourceforge.net/projects/gif2apng/files/$portVersion/gif2apng-$portVersion-src.zip"
+CHECKSUM_SHA256="3b21308e935d799b3ffb4a86c6e00ffa4cb9b3f72f52d58d51c66eb0574ae7d2"
+SOURCE_DIR=""
+
+ARCHITECTURES="x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
```
fbrosson: If a recipe does not build on x86_gcc2 primary arch then we usually install to $prefix/bin instead of $binDir (which is $prefix/bin/x86 on x86 secondary arch).
For that purpose we define commandBinDir with these lines between SECONDARY_ARCHITECTURES and PROVIDES:
```

	commandBinDir=$binDir
else
	commandBinDir=$prefix/bin
fi

```
(You can copy that block from haikuports/media-video/dvdauthor/dvdauthor-0.7.1.recipe)

mmuman:
Doesn't it have a make install ?
I'll let @fbrosson list the rest.

R167:
@mmuman no, just generates the binary in the main directory. no install task defined in the makefile

media-gfx/gif2apng/gif2apng-1.9.recipe
R167: forgot to add the patch to the recipe file. Fixing.

mmuman:
Ok then.

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,44 @@
+SUMMARY="Converts GIF animations into APNG format"
+DESCRIPTION="This program converts GIF animations into animated PNG format. \
+Usually it makes the files smaller. CLI version is OS-independent."
```
fbrosson: We can probably remove the "**`CLI version is OS-independent.`**"

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,44 @@
...
+HOMEPAGE="http://gif2apng.sourceforge.net/"
+COPYRIGHT="2009-2013 Max Stepin"
+LICENSE="Zlib"
+REVISION="1"
+SOURCE_URI="https://sourceforge.net/projects/gif2apng/files/$portVersion/gif2apng-$portVersion-src.zip"
```
fbrosson: I forgot to suggest a slightly better SOURCE_URI:
```
SOURCE_URI="https://downloads.sourceforge.net/gif2apng/gif2apng-$portVersion-src.zip"
```

R167:
just pushed updates @fbrosson. Thoughts?

fbrosson:
Sorry for the delay. I was trying to build on x86_gcc2 primary arch with a variant of your patch.
The idea was to replace **`-lstdc++`** by **`$(LIB_STDCPP)`** in the Makefile with the patch, and then define 	LIB_STDCPP with:
```
	PATCHES="gif2apng-$portVersion-lstdc++.patchset"
	LIB_STDCPP=-lstdc++.r4
else
	LIB_STDCPP=-lstdc++
fi
```
in order to be able to have this in BUILD:
```
	LIB_STDCPP=$LIB_STDCPP make
```

Unfortunately it does not work: I get **`undefined reference to __builtin_clz`**:
```
obj/zopfli/util.o: In function `ZopfliGetDistExtraBits':
util.c:(.text+0x21): undefined reference to `__builtin_clz'
obj/zopfli/util.o: In function `ZopfliGetDistExtraBitsValue':
util.c:(.text+0x5c): undefined reference to `__builtin_clz'
obj/zopfli/util.o: In function `ZopfliGetDistSymbol':
util.c:(.text+0xac): undefined reference to `__builtin_clz'
```

So I guess the best option is to drop x86_gcc2 as primary arch. (That's what you did ;-)
I don't know, however, if we should have:
```
SECONDARY_ARCHITECTURES="x86"
```
or
```
ARCHITECTURES="x86 x86_64"
SECONDARY_ARCHITECTURES="x86"
```
(I _think_ we should have the **`!x86_gcc2`** but I'm not sure.)

R167:
@fbrosson yeah, that was the error I kept running into with GCC2. the issue is that zopfli requires later than GCC4. might just not include the x86_gcc2 for that reason, but up to you.

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,50 @@
...
+}
+
+INSTALL()
+{
+	if [ "$targetArchitecture" != x86_gcc2 ]; then
+		commandBinDir=$binDir
+	else
+		commandBinDir=$prefix/bin
+	fi
+	echo $commandBinDir
```
fbrosson: It would be better without the echo $commandBinDir ;-)

Everything else looks OK. I'll build it on each of my architectures and then merge.

R167: WHOOPS! completely forgot about that. doh!

fbrosson: While at it, can you also change ARCHITECTURES to:
```
```
Thanks!

R167: okay, fixed that now.

R167:
okay, now it's fixed.

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+INSTALL()
+{
+	if [ "$targetArchitecture" != x86_gcc2 ]; then
+		commandBinDir=$binDir
+	else
+		commandBinDir=$prefix/bin
+	fi
+	mkdir -p $commandBinDir
+	cp gif2apng $commandBinDir
```
fbrosson: Maybe we should simplify INSTALL to have only:
```
INSTALL()
{
	mkdir -p $prefix/bin
	cp gif2apng $prefix/bin
}
```
I had suggested defining commandBinDir because we often do that, but since we don't even use it in BUILD (for this recipe) it does not really make sense to define commandBinDir just for INSTALL. Moreover, we just use it twice. (So sorry for the bad suggestion...)

R167: it's all good.

media-gfx/gif2apng/gif2apng-1.9.recipe
```diff
@@ -0,0 +1,49 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:automake
```
fbrosson: It builds fine without **`cmd:automake`** so it would be nice to remove that dependency. Thanks!

R167: good now?

media-gfx/gif2apng/gif2apng-1.9.recipe
fbrosson: Great. Thanks! I'll validate your task asap.
Then I'll build the recipe on x86 and x86_64 before I merge.

media-gfx/gif2apng/gif2apng-1.9.recipe
fbrosson: Yes, it's perfect ;-)

To HaikuPorts members: If you merge, please don't forget to schrink the commit message.
But don't worry. I'll merge it myself as soon as I have built it on all architectures.

pulkomandy:
On Wed, Nov 30, 2016 at 05:41:01PM -0800, fbrosson wrote:
> (I _think_ we should have the **`!x86_gcc2`** but I'm not sure.)

Yes, it means "tested on x86_gcc2, does not work".

fbrosson:
OK, thanks.

fbrosson:

PR #897, opened 2016-12-01
Re-add patchset

ohnx:
Don't think it should have been removed, as a lot of changes were made in https://github.com/haikuports/haikuports/commit/1ac39b83d173defd7db80c12ff2f57f8845eb41f. A small change like https://github.com/haikuports/haikuports/commit/1ac39b83d173defd7db80c12ff2f57f8845eb41f#diff-833e49ab9e6069299e254050cccb7874L43 might have gone unnoticed.


ohnx:
Also gets rid of the linter warning https://gist.github.com/ohnx/676338c83bd396261dd294029387cf7c#file-haikuports-errors-txt-L48

korli:
Thanks! I let the revision as is because this package wasn't published on any arch. Though if someone needs it, just bump the rev.

PR #898, opened 2016-12-01
added xdelta recipe

R167:
Currently x86 only. Should I change the default binDir for where it generates? aka, not in bin/x86 to just bin?

### DIFF
--- a/dev-util/xdelta/xdelta-3.1.0.recipe
+++ b/dev-util/xdelta/xdelta-3.1.0.recipe
@@ -1,0 +1,59 @@
+SUMMARY="A binary diff tool implementing VCDIFF"
+DESCRIPTION="Xdelta is a tool and library for differential compression. Xdelta \
+release version 3 supports VCDIFF encoding and decoding."
+HOMEPAGE="http://xdelta.org/"
+COPYRIGHT="2001-2015 Josh MacDonald"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="https://github.com/jmacd/xdelta/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="7515cf5378fca287a57f4e2fee1094aabc79569cfe60d91e06021a8fd7bae29d"
+SOURCE_DIR="xdelta-$portVersion/xdelta3"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	xdelta$secondaryArchSuffix = $portVersion
+	cmd:xdelta3 = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:liblzma$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:liblzma$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf --install
+	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir
+	make
+}
+
+INSTALL()
+{
+	make install
+}
+
+TEST()
+{
+	make check
+	xdelta3 test || true
+}


pulkomandy:
Yes, move it to just bin if there is no gcc2 version.

Is there actually a libxdelta3 (matching lib:xdelta3$secondaryArchSuffix), and is the soname version the same as the package version?

Is there anything in the development package? (headers files and such)? If there is, does it include .la files (these should be removed) or a pc file for pkgconfig (it should be adjusted with fixpkgconfig)?

fbrosson:
It does not build any lib. But maybe if we add **`lib:liblzma$secondaryArchSuffix`** to REQUIRES and **`devel:liblzma$secondaryArchSuffix`** to BUILD_REQUIRES then we might get a lib.

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+CHECKSUM_SHA256="7515cf5378fca287a57f4e2fee1094aabc79569cfe60d91e06021a8fd7bae29d"
+SOURCE_DIR="xdelta-$portVersion/xdelta3"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	xdelta$secondaryArchSuffix = $portVersion
+	cmd:xdelta3$secondaryArchSuffix = $portVersion
+	lib:xdelta3$secondaryArchSuffix = $portVersion
```
fbrosson:
```
	lib:liblzma$secondaryArchSuffix = $portVersion
```
should be dropped, as well as the entire PROVIDES_devel and REQUIRES_devel.

fbrosson:
I've built xdelta with liblzma... but I did not get a lib, so the _devel pacakge is empty. It should therefore be dropped, as well as the **`devel:liblzma$secondaryArchSuffix = $portVersion`** in PROVIDES.

@R167, you can have a look at haikuports/media-video/dvdauthor/dvdauthor-0.7.1.recipe and copy the parts that mention commandBinDir to follow @pulkomandy's suggestion.

That said, if we can add an trivial or small patch for building on x86_gcc2 primary then it might be OK to drop support for building for x86 secondary arch.

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
```
fbrosson: You can add this to REQUIRES:
```
	lib:liblzma$secondaryArchSuffix
```

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+	xdelta${secondaryArchSuffix}_devel = $portVersion
+	devel:xdelta3$secondaryArchSuffix = $portVersion
+	"
+REQUIRES_devel="
+	xdelta$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
```
fbrosson: You can add this to BUILD_REQUIRES:
```
	devel:liblzma$secondaryArchSuffix
```

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:awk
```
fbrosson: I _think_ it would be better to have:
```
BUILD_PREREQUIRES="
	cmd:aclocal
	cmd:autoconf
	cmd:automake
	cmd:autoheader
	cmd:gcc$secondaryArchSuffix
	cmd:libtool$secondaryArchSuffix
	cmd:make
	"
```
Can an autotool expert comment about this? Thanks!

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtool$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf --install
```
fbrosson: It worked for me (on x86_64) without the call to autoreconf.
But I need to check that it still builds OK without the call to autoreconf when building for x86 secondary arch on x86_gcc2.

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+	runConfigure ./configure
+	make
+}
+
+INSTALL()
+{
+	make install
+}
```
fbrosson: You can add a TEST block with "make check", and then call **`hp --test xdelta`**

dev-util/xdelta/xdelta-3.1.0.recipe
fbrosson: **Oops,** sorry, I juist realized that the devel package was empty because this was missing in INSTALL:
```
```
We need to test with it, before we remove the _devel package.

dev-util/xdelta/xdelta-3.1.0.recipe
fbrosson: Sorry, unlike you, I had not noticed that configure is missing from the archive. You are perfectly right. We only have configure.ac, so we need the call to autoreconf. You are doing a great job!

dev-util/xdelta/xdelta-3.1.0.recipe
fbrosson: It is recommended to have a **`cmd:autoreconf`** even if it is pulled by the **`cmd:autoconf`**.

dev-util/xdelta/xdelta-3.1.0.recipe
R167: I added in test and made other suggested changes.

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+	xdelta$secondaryArchSuffix = $portVersion
+	cmd:xdelta3$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:liblzma$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
```
fbrosson: Should be
```
```

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf --install
+	runConfigure ./configure
```
fbrosson: Should be
```
	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir
```

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+HOMEPAGE="http://xdelta.org/"
+COPYRIGHT="2001-2015 Josh MacDonald"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="https://github.com/jmacd/xdelta/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="7515cf5378fca287a57f4e2fee1094aabc79569cfe60d91e06021a8fd7bae29d"
+SOURCE_DIR="xdelta-$portVersion/xdelta3"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
fbrosson: I'm almost sure we need to insert a **`!x86_gcc2`** in SECONDARY_ARCHITECTURES. Anyway, it should be followed by the definition of commandBinDir, as shown below:
```

	commandBinDir=$binDir
else
	commandBinDir=$prefix/bin
fi
```

dev-util/xdelta/xdelta-3.1.0.recipe
R167: currently fixing these things. Need to reconcile changes across dev envs... What is the reasoning behind rebasing the PRs? Just curious.

R167:
@fbrosson Should we keep secondary arch?

fbrosson:
Yes, we need to keep support for building for x86 _secondary_ arch (on x86_gcc2) because it does not build for x86_gcc2 _primary_ arch.
BTW, that's perfectly OK to have support for building for secondary architectures.

And not being able to build for x86_gcc2 primary arch a recipe which does not provide any library is not a problem at all, since a **`cmd:*`** item (e.g. xdelta3) provided by a secondary arch build of a recipe can be called from any other program, no matter what architecture it was built for, I mean, primary vs secondary.

R167:
@fbrosson Okay. Think I understand now. Just pushed one last change to deal with the naming better like it was done in dvdauthor.

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,58 @@
+SUMMARY="A binary diff tool that uses the VCDIFF (RFC 3284) format and compression"
```
fbrosson: Can you shrink this line to something <= 70 chars, like, for instance:
```
SUMMARY="A binary diff tool that uses the VCDIFF (RFC 3284) format"
```
This is because SUMMARY is used by HaikuDepot, which needs to show it in a big list.
(See https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#summary for more details.)
Thanks!

R167:
@fbrosson made those changes.

dev-util/xdelta/xdelta-3.1.0.recipe
```diff
@@ -0,0 +1,58 @@
...
+}
+
+INSTALL()
+{
+	make install
+}
+
+TEST()
+{
+	make check && xdelta3 test
```
fbrosson: The recipe is OK, but **`xdelta3 test`** is failling on its 23rd test:
```
xdelta3: testing no_output...xdelta3: command was xdelta3 -q -f -e /tmp/xdtest.16470.target.3 /tmp/xdtest.16470.noperm.3
 failed: command should have not succeeded: XD3_INTERNAL
```
So if you could change TEST to:
```
TEST()
{
	make check
	xdelta3 test || true
}
```
that would be great.
I'll validate you task asap. Thanks again for your great work!
(I'll keep the PR unmerged for some hours in case someone finds a fix for the internal tests.)

dev-util/xdelta/xdelta-3.1.0.recipe
fbrosson: I see two strategies for fixing a recipe that is being discussed in a PR.

The first strategy is to push incremental commits, generally one every time a change is requested. This helps the mentors to easily understand what changes were made at every step. And when everything looks OK, I would say that squashing all commits is a good practice.

The second strategy, which I use, is to squash my commits (and rebase them against master) everytime I have to fix something in a PR I have opened.
I also update the timestamp of my (squashed) commit to the current time just before I push it to my remote branch. This allows other contributors to easily notice that there has been a recent change and determine whether or not (s)he has already examined it.
That said, when I change only the commit message (or when I just want to rebase against master) then I don't update the timestamp.

I don't know if these practices are widely adopted. I came to the conclusion that this was a good strategy because it makes it easier for other contributors to pull remote branches (for testing purposes) without having to dive too far in the past. This is handy if, e.g. you want to push a tiny fix to the remote branch of a contributor who does not have time to update his PR.

fbrosson:
@R167, could you please update this PR with a tiny change to the TEST function?
```
TEST()
{
	make check
	xdelta3 test || true
}
```
Thanks!
BTW, if you are busy then I can do the change myself, preserving you as the author of the commit, of course.

R167:
@fbrosson updated like you said.

fbrosson:
Cool, thanks @R167 !

fbrosson:

PR #901, opened 2016-12-01
calc: new recipe for a C-style arbitrary precision calculator

raefaldhia:


sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,53 @@
...
+{
+         make -f Makefile.simple INCDIR=$includeDir \
+         BINDIR=$binDir \
+         LIBDIR=$libDir \
+         EXTRA_CFLAGS="-D_HAIKU_ -DUSE_TERMIOS -DHAVE_UID_T"  
+}
+
+INSTAL()
+{
+    make -f Makefile.simple install
```
fbrosson: Multiple whitespaces here too.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,53 @@
...
+         cmd:less
+         cmd:ln
+	"
+
+BUILD()
+{
+         make -f Makefile.simple INCDIR=$includeDir \
+         BINDIR=$binDir \
+         LIBDIR=$libDir \
+         EXTRA_CFLAGS="-D_HAIKU_ -DUSE_TERMIOS -DHAVE_UID_T"
```
fbrosson: We have a Ctrl-V Ctrl-V Ctrl-Z between **`make -f Makefile.simple`** and **`NCDIR=$includeDir`**, and a Ctrl-X after **`LIBDIR=$libDir`**.

BTW, I forgot to mention that we usually add one extra tab for continuation lines:
```
	make -f Makefile.simple INCDIR=$includeDir \
		BINDIR=$binDir \
		LIBDIR=$libDir \
		EXTRA_CFLAGS="-D_HAIKU_ -DUSE_TERMIOS -DHAVE_UID_T"
```
Hmm, **`__HAIKU__`** gets defined automatically. Maybe we don't need to define **`_HAIKU_`**, unless the sources test for it.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,53 @@
...
+SOURCE_URI="http://www.isthe.com/chongo/src/calc/calc-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0fcf60917efb10aab145f1180c278a3ab63d789aad6d5e5f066a4458b5f9b692"
+PATCHES="calc$secondaryArchSuffix-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	calc$secondaryArchSuffix = $portVersion
+	cmd:calc$secondaryArchSuffix
```
fbrosson: I noticed a few non-printable chars in 3 different lines.
The first ones are to the right of **`cmd:calc$secondaryArchSuffix`**: a Ctrl-X and a Ctrl-S.
I guess you love emacs ;-)

raefaldhia: What is non-printable chars?

sci-mathematics/calc/calc-2.12.5.3.recipe
raefaldhia: What is Ctrl-V Ctrl-Z do in Pe? look like that's my typo because in haiku to copy and paste use `ALT` sometimes it make me confused hehe

raefaldhia:
Desc updated.

sci-mathematics/calc/calc-2.12.5.3.recipe
fbrosson: Mainly, anything between Control-A to Control-Z.
Vim shows them as ^A to ^Z. You can also find them in a file by running:
```
cat -v myfile | diff - myfile
```

sci-mathematics/calc/calc-2.12.5.3.recipe
fbrosson: I don't use Pe. I use vi (or vim, to be precise) everywhere ;-) Sometimes I use Emacs, but vim loads much faster.
BTW, vim displays all non-printable chars.
If you are not used to vim then I can tell you that learning vim is a very good investment. Vim is available everywhere. Well, to tell the truth, you might find some systems without vi, but you'll always have ed (and probably pico or nano). And since learning ed is automatic if you learn vi, vi is the best choice if you want to use a single editor.
BTW, vi is the only editor which allows to save&quit with only 2 keys: ZZ
Isn't this great? ;-)

pulkomandy: You can change the keyboard mapping in Haiku preferences/keymap if you want ctrl to be used for copy and paste.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+PATCHES="calc$secondaryArchSuffix-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	calc$secondaryArchSuffix = $portVersion
+	cmd:calc$secondaryArchSuffix
+	"
+
```
fbrosson: According to the [guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering) we do not usually have an empty line between PROVIDES and REQUIRES.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+		LCC=gcc \
+		INCDIR=$includeDir \
+		LIBDIR=$libDir \
+		BINDIR=$binDir \
+		EXTRA_CFLAGS="-DUSE_TERMIOS -DHAVE_UID_T"
+}
+
+INSTAL()
+{
+	make Makefile install
```
fbrosson: I guess you meant:
```
	make -f Makefile install
```
In which case you could replace it by:
```
	make install
```

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+	"
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+
```
fbrosson: According to the [guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering) we do not usually have an empty line between BUILD_REQUIRES and BUILD_PREREQUIRES.

(This is the second empty line I was refering to in my last review... I forgot to clic on a button...)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,71 @@
...
+Otherwise, it enters interactive mode. In this mode, it accepts \
+commands one at a time, processes them, and displays the answers. \
+In the simplest case, commands are simply expressions which are evaluated."
+HOMEPAGE="http://www.isthe.com/chongo/tech/comp/calc/"
+COPYRIGHT="1999-2014  Landon Curt Noll"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.isthe.com/chongo/src/calc/calc-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0fcf60917efb10aab145f1180c278a3ab63d789aad6d5e5f066a4458b5f9b692"
+PATCHES="calc$secondaryArchSuffix-$portVersion.patchset"
```
fbrosson: Although **`haikuporter -e calc_x86`** generates **`calc_x86-2.12.5.3.patchset`**, you should rename that file to **`calc-2.12.5.3.patchset`**. For your convenience, here's how I would do so:
```
cd /path/to/haikuports/sci-mathematics/calc/patches
git mv calc_x86-2.12.5.3.patchset calc-2.12.5.3.patchset
git commit --amend calc_x86-2.12.5.3.patchset calc-2.12.5.3.patchset
```
Then you should change PATCHES to:
```
PATCHES="calc-$portVersion.patchset"
```
These changes will allow the package to be also built on x86 primary arch and x86_64.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,71 @@
...
+	cmd:rmdir
+	cmd:cp
+	cmd:mv
+	cmd:cat
+#	cmd:col
+	cmd:less
+	cmd:ln
+	cmd:strip
+	"
+
```
fbrosson: To fix the issue about the hard links, you can insert this block here:
```
PATCH()
{
	sed -i -e 's/^LN= ln$/& -s/' Makefile
}
```
You could also avoid this method and, instead, change the "**`LN= ln`**" into "**`LN= ln -s`**" in the Makefile and regenerate the patchset.
Feel free to chose whatever method you prefer. Both are OK ;-)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,71 @@
...
+{
+	make all BLD_TYPE=calc-static-only \
+		LCC=gcc \
+		INCDIR=$includeDir \
+		LIBDIR=$libDir \
+		BINDIR=$binDir \
+		EXTRA_CFLAGS="-DUSE_TERMIOS -DHAVE_UID_T"
+}
+
+INSTAL()
```
fbrosson: Should be **`INSTALL()`** (you forgot an L).

raefaldhia:
@fbrosson do you know what's the replacement for /usr/share?

fbrosson:
I think it is $dataDir (which will later be visible as /system/data/) but I need to check.

When you build calc_x86-2.12.5.3 you get in haikuports/sci-mathematics/calc/work-x86-2.12.5.3 a file called wrapper-script which has plenty of paths. I often look there. I know, I should refer to the docs instead...
wrapper-script

sci-mathematics/calc/calc-2.12.5.3.recipe
fbrosson: I've updated the PATCH function above. It sets COL to the empty string... and solves the issue about **`col`** not being available!

raefaldhia:
Is it done now?

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
+SUMMARY="C-style arbitrary precision calculator"
+DESCRIPTION="Calc is an interactive calculator which provides for \
+easy large numeric calculations, but which also can be easily programmed for \
+difficult or long calculations. It can accept a command line \
+argument, in which case it executes that single command and exits. \
+Otherwise, it enters interactive mode. In this mode, it accepts \
+commands one at a time, processes them, and displays the answers. \
+In the simplest case, commands are simply expressions which are evaluated."
```
fbrosson: In order to follow the [guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines), could you please change DESCRIPTION to:
```
DESCRIPTION="Calc is an interactive calculator which provides for easy large \
numeric calculations, but which also can be easily programmed for difficult or \
long calculations. It can accept a command line argument, in which case it \
executes that single command and exits. Otherwise, it enters interactive mode. \
In this mode, it accepts commands one at a time, processes them, and displays \
the answers. In the simplest case, commands are simply expressions which are \
evaluated."
```
(The text is unchanged. I've only moved the line breaks, to make all lines as long as possible, but still <= 80 chars long.)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
...
+HOMEPAGE="http://www.isthe.com/chongo/tech/comp/calc/"
+COPYRIGHT="1999-2014  Landon Curt Noll"
```
fbrosson: Should be:
```
COPYRIGHT="1999-2016 Landon Curt Noll"
```
(updated year and removed double white spaces)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	calc${secondaryArchSuffix}_devel = $portVersion
+	devel:libcalc$secondaryArchSuffix
+	devel:libcustcalc$secondaryArchSuffix
+"
```
fbrosson: There should be a tab before the closing double-quotes.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
...
+REQUIRES_devel="
+	calc$secondaryArchSuffix == $portVersion base
+"
```
fbrosson: There should be a tab before the closing double-quotes.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
...
+		LDCONFIG= \
+		INCDIR=$includeDir \
+		LIBDIR=$libDir \
+		BINDIR=$binDir \
+		CALC_SHAREDIR=$dataDir/calc \
+		MANDIR=$manDir/man1 \
+		EXTRA_CFLAGS="-DUSE_TERMIOS -DHAVE_UID_T"
+
+	prepareInstalledDevelLib libcalc
+	prepareInstalledDevelLib libcustcalc
```
fbrosson: The 2 calls to prepareInstalledDevel**Lib** should be combined into a single one to prepareInstalledDevel**Libs**:
```
	prepareInstalledDevelLibs \
		libcalc \
		libcustcalc
```

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
...
+HOMEPAGE="http://www.isthe.com/chongo/tech/comp/calc/"
+COPYRIGHT="1999-2014  Landon Curt Noll"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.isthe.com/chongo/src/calc/calc-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0fcf60917efb10aab145f1180c278a3ab63d789aad6d5e5f066a4458b5f9b692"
+PATCHES="calc-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
fbrosson: Good news: it builds fine on x86_gcc2 primary arch!
This means you can (and even should) remove the **`!`** in **`!x86_gcc2`** in both ARCHITECTURES and SECONDARY_ARCHITECTURES ;-)
```
```

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,107 @@
...
+
+	prepareInstalledDevelLib libcalc
+	prepareInstalledDevelLib libcustcalc
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
+}
+
+
```
fbrosson: Since the Makefile has a **`check`** target that launches some tests, you can (and probably should) add a TEST function:
```
TEST()
{
	make check
}
```
The TEST function allows to call **`hp --test calc`** (after it has been built) to run the tests.
BTW, I've launched the tests with a build for x86_gcc2 primary arch, and I got quite some output, but now it is stuck on:
```
...
8830: pi() ^ asin(-2) == power(pi(), asin(-2)
8831: (3+4i) ^ (2+3i) == power(3+4i, 2+3i)
8832: ln(-10) ^ (2+3i) == power(ln(-10), 2+3i)
8833: (pi()*1i) ^ asin(-2) == power(pi()*1i, asin(-2))
8834: (exp(1)+pi()*1i) ^ asin(-2) == power(exp(1)+pi()*1i, asin(-2))
8835: Ending test_somenew

8900: Starting test of calc resource functions by Christoph Zurnieden
8901: read -once "test8900"
8902: about to run test8900(1,,8903)
```

sci-mathematics/calc/calc-2.12.5.3.recipe
fbrosson: I realized that setting COL to the empty string was what was making the build fail when I tried to build the shared libs. (It was complaining about **`-b`** not being found... because it was calling **`$COL -b`**.)
So the way you handled this in the patchset is much better!
Your idea to pass **`LN="ln -s"`** to make (instead of patching the Makefile as I had suggested) is also great! Well done!

BTW, I'm now trying to build the package with the shared libs.

sci-mathematics/calc/calc-2.12.5.3.recipe
raefaldhia: @fbrosson After playing arround i found that your problem caused by permision try to grant write,exec,read perms to all .cal file

raefaldhia: but still there's another 4 error

fbrosson: Is your timezone UTC+7 ?
It looks like you have set your computer (or VM) to the UTC timezone and its clock is 7 hours in the future.

I think you should set your timezone to "GMT+07.00".
(Right click on the digital clock and choose "Time preferences...")

GitHub uses the timestamps of your commits to display them at the appropriate positions between our comments. So if your clock is in the future then your commits will allways appear on the bottom of this page (unless we wait 7 hours after your last commit update to add comments).
That said, this is a minor issue, of course!

fbrosson: Cool, you made all the changes and the recipe is almost good.
Can you move TEST _after_ INSTALL? (see the [guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering) ;-)

I think we don't need to list every **`cmd:*`** in BUILD_PREREQUIRES. The most used ones are always available.
You can remove chmod, mkdir, rm, touch, rmdir, cp, cat and ln.
Let's see if other contributors see other **`cmd:*`** that we can omit.

sci-mathematics/calc/calc-2.12.5.3.recipe
raefaldhia: "BTW, I'm now trying to build the package with the shared libs."

Do you mean about calc-dynamic-only?

sci-mathematics/calc/calc-2.12.5.3.recipe
raefaldhia: Ok done

fbrosson:
@raefaldhia: Cool, you have correctly set your timezone. Thanks!

I remember now the rule to decide which **`cmd:*`** (in BUILD_PREREQUIRES) should be considered as implicit: all commands available in coreutils. To list them:
```
package list -p /system/packages/coreutils-*-*.hpkg | sed -n -e 's|^bin/||p'
```

@korli, could you please confirm that we can/should skip in BUILD_PREREQUIRES any **`cmd:*`** that is in the coreutils package? Thanks!

raefaldhia:
@fbrosson ok recipe updated

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,102 @@
...
+		LDCONFIG= \
+		INCDIR=$includeDir \
+		LIBDIR=$libDir \
+		BINDIR=$binDir \
+		CALC_SHAREDIR=$dataDir/calc \
+		MANDIR=$manDir/man1 \
+		EXTRA_CFLAGS="-DUSE_TERMIOS -DHAVE_UID_T"
+
+	prepareInstalledDevelLibs \
+		libcalc libcustcalc
```
fbrosson: After all, maybe it is better to have a single line:
```
	prepareInstalledDevelLibs libcalc libcustcal
```
Leaving the command alone and then putting two libs together in the same line is not very common.
(The reason why we sometimes put one library per line is when we have too many or when their names match a nice pattern.)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,102 @@
...
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
+
+	cd $dataDir/calc
+	find *.cal | xargs chmod 755
```
fbrosson: It would be better to protect the **`*`** because if for some reason, during the build, someone does a "touch foobar.cal" in the directory where the find command is launched, then it will not visit any subdirectory. I would also recomment **`chmod +w`** instead of **`chmod 755`**. That's why I would use:
```
	find "*.cal" | xargs chmod +w
```
or
```
	find \*.cal | xargs chmod +w
```
Both work. Take the one you prefer!

sci-mathematics/calc/patches/calc-2.12.5.3.patchset
```diff
@@ -0,0 +1,44 @@
...
++all: ${BLD_TYPE} CHANGES
+ 
+ check_include:
+ 	$(Q) if [ ! -d /usr/include ]; then \
+@@ -2149,7 +2149,7 @@ calc.usage: calc.1 ${MAKE_FILE}
+ 	    LESSCHARSET=iso8859 ${CALCPAGER} calc.1; \
+ 	else \
+ 	    ${NROFF} -man calc.1; \
+-	fi | ${COL} -b > $@
++	fi | > $@
```
fbrosson: You forgot to remove the pipe. It should be:
```
++	fi > $@
```

sci-mathematics/calc/calc-2.12.5.3.recipe
raefaldhia: find "*.cal" no such file or directory . .

raefaldhia: same thing happened for find \*.cal

sci-mathematics/calc/patches/calc-2.12.5.3.patchset
raefaldhia: Actually i was unsure about that because i'm not too good at BASH

sci-mathematics/calc/calc-2.12.5.3.recipe
fbrosson: Sorry, it should be:
```
	find . -name \*.cal
```

raefaldhia: updated

fbrosson: Or 
```
	find -name \*.cal
```
The reason why I forgot the **`-name`** is that until recently I did not know that the path was not mandatory, so I had made a _mental_ patch in my brain: "Don't forget that find may also be called without the path". But I must have turned it into the incorrect idea that "find could be called without the **`-name`** parameter. I guess I need some sleep ;-)

raefaldhia: Ayylmao, updated

fbrosson: I don't know if it is worth, but I think it would look nicer with an additional **`-type f`**: 
```
	find -type f -name \*.cal | xargs chmod +w
```
(My rationale here is that we can instruct find to not try to compare any directory name against the **`*.cal`** pattern.)

raefaldhia:
updated

fbrosson:
Cool!

> > "BTW, I'm now trying to build the package with the shared libs."

> Do you mean about calc-dynamic-only?

Yes, exactly.
It would be nice to have the static libs (in the calc_devel hpkg) and the shared libs (in the calc hpkg).
As we know it builds fine with **`BLD_TYPE=calc-static-only`** but when I try **`BLD_TYPE=calc-dynamic-only`** I get several "implicit declaration of function" for:
memmove, memcmp, chdir, strdup, memchr, access, sleep, usleep, strstr, strncat, getpid, getppid, getuid, geteuid, getgid, getegid, getsid and getpgid.

Since it would be great to fix this, would you like to dig that topic?. You don't need to do it today. I can add several days to the task.

There is another nice feature in calc: it can link against libreadline, libhistory and libncurses !
The Makefile explains how to do. My understanding is that we just need to pass the following definitions to each of the "make" calls, like you did for LN and the other parameters:
```
USE_READLINE=-DUSE_READLINE
READLINE_LIB=-lreadline
READLINE_EXTRAS="-lhistory -lncurses"
```
Of course, these will only be available in the chroot for the compilation if:
* **`devel:libreadline`**, **`devel:libhistory`**, and **`devel:libncurses`** are added to **`BUILD_REQUIRES`**;
* _and_ **`lib:libreadline`**, **`lib:libhistory`**, and **`lib:libncurses`** are added to **`REQUIRES`**.

Would you like to try that?

Don't worry if you think this is too difficult. You have already done a great job and we can validate your task now (or later) if you want. But I though you might wish to dig these two topics. Thanks!

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,101 @@
...
+		LCC=gcc \
+		LN="ln -s" \
+		LDCONFIG= \
+		INCDIR=$includeDir \
+		LIBDIR=$libDir \
+		BINDIR=$binDir \
+		CALC_SHAREDIR=$dataDir/calc \
+		MANDIR=$manDir/man1 \
+		EXTRA_CFLAGS="-DUSE_TERMIOS -DHAVE_UID_T"
+
```
fbrosson: Could you insert these commands here:
```

	# At the moment, calc and calc-static are identical
	rm $binDir && ln -s calc-static $binDir/calc

```

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,101 @@
...
+	prepareInstalledDevelLibs libcalc libcustcalc
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
+
```
fbrosson: Could you please insert these lines here:
```

	# remove useless dangling symbolic links
	rm $developLibDir/libcalc.so $developLibDir/libcustcalc.so

```

raefaldhia:
@fbrosson i'd love to, but wait i'll thinking first

fbrosson:
Cool!
And don't worry. Even if you accept but later find out that the challenge is too hard, you have my word that your task will be validated!
Also, there is no urgency. You can enjoy the week-end and come back later to tell what you have decided.
I've just added one more day so that you will be able to think and/or get some rest during the whole week-end if you wish ;-)

raefaldhia:
@fbrosson so it's possible to accept this whenever ex: after i completed 2 more gci issue because that's min target i need to pass?

fbrosson:
Of course it is! All we need to do is validate the task so that you can claim another one.
The task will appear as completed, but you will still have the possibility to continue "outside" of the GCI.
Then, when you come back to this recipe you will have all the time you want.
And when the mentors of Haiku Inc. will chose the finalists, we will of course remember that if have improved the recipe after the task got validated. All I ask you is, first, that you do the tiny extra changes I've put in my last review and, secondly, don't advertize too much on our deal ;-)
This way we will be able to merge the recipe as it is now, or keep it open a few extra days, in case you would like to work on the libreadline+libhistory+libncurses challenge and leave the "dynamic-lib" challenge for later.

raefaldhia:
Look like libreadline+libhistory+libncurses are done.

raefaldhia:
@fbrosson Look like libreadline+libhistory+libncurses are done.

raefaldhia:
Messing around with `Makefile` and `calc-dynamic-only`, i got linker error, undefined reference then
when i changed the makefile line `ifeq ($(target),Linux)` to `ifeq ($(target),**Haiku**)` and build it with calc_x86 arch it successed, i don't even know what i've done ¯\_(ツ)_/¯ , i just thinking that because haiku have /posix directory that contains headers for linux-like then any software that could compile on linux also compile on haiku.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,117 @@
...
+	"
+REQUIRES_devel="
+	calc$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libreadline
+	devel:libhistory
+	devel:libncurses
```
fbrosson: Well, it looks like your calc_data is ready! You have become an expert!
I would like to apologize, however, because I forgot to mention two things that need to be checked for the calc_data:
1) Before we create such a sub-package, we need to make sure that, without the calc_data, the contents of the files we want to put in calc_data are the same, when we compare what we get for a primary arch build with what we get for a secondary arch build. When we think about this it seems an obvious thing to do, but I should however have mentioned it.
2) The second thing that needs to be checked is that haikuporter is happy with the calc_data stuff not only for a primary arch build, but **_also_** for a secondary arch build. And that's where the problems might come: I remember that I have hit this problem with a package, but I don't remember if it was with file (and then that I found how to fix it) or if it was with another recipe, for which I had to give up the creation of the arch-independent sub-package.

The hardest part of calc_data is to make sure that INSTALL does not fail with secondary arch builds, when it tries to copy the data files to the read-only directory that holds the files from the BUILD_REQUIRE~S~d calc_data. I'm almost sure that we will hit this problem, and that we will need to find a workaround.
So this means it is a bit more complicated to make sure calc_data work perfectly for secondary arch builds.
If your recipe is already working fine with calc_data on secondary arch builds, then we only need to choose a nice text for SUMMARY_data. Otherwise, what I suggest is that you keep your current calc_data changes locally on a safe place and restore, on this pull request, the version without calc_data, and then come back later (if you wish) after you have done as many other tasks as you wish.
Shall we do that?
BTW, as I have already said, your recipe is already great without calc_data, and you have done a great job!
Let me just check that the READLINE stuff is working fine and then we can merge this recipe and validate your task so that you can enjoy other tasks ;-)

fbrosson:
That's cool, it looks like you have found the right thing to do! Congratulations! I did not get close to that myself!
Regarding the cscript directory, yes, your suggestion looks good. But we need to make sure the contents are the same when we compare its files between a primary arch and a secondary arch. We might also hit the "read-only" problem on secondary arch builds, as explained in my previous message. So we need to be careful with this.
Hmm, in your screen capture I see some thing strange. You have:
* lib/x86/libcustcalc.so.2.12.5.3
* lib/x86/libcalc.so.2.12.5.3

which seems OK, but you also seem have:
* lib/calc/cscript/

That's a little bit strange, because if you are passing **`SCRIPTDIR=$libDir/calc/cscript`** (and building for x86 secondary arch) then you would have lib/x86/calc/cscript instead of lib/calc/cscript.
But I guess you have already made the required changes to teach calc_x86 that the cscript will be in lib/calc/cscript for both primary and secondary arch builds. You are an expert!

I need to test your recipe and I'll come back.
If the shared libs is working fine then this is a great step!
This task is definitely very interesting and exciting!

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,131 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libreadline$secondaryArchSuffix
+	lib:libhistory$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	calc_data == $portVersion
+	"
+
+if [ -z "$secondaryArchSuffix" ]; then
+	SUMMARY_data="CHANGE ME LATER"
```
fbrosson: Shall we use this string?
```
SUMMARY_data="Calc standard resource files"
```
(It's taken from the very first line of the README file in the calc directory.)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,131 @@
...
+	"
+
+BUILD()
+{
+	make all \
+		BLD_TYPE=calc-static-only \
+		LCC=gcc \
+		LN="ln -s" \
+		LDCONFIG= \
+		INCDIR="$includeDir" \
```
fbrosson: You can drop the double quotes here if you wish, because there are no spaces in the definition of $includeDir.
(If in doubt about any of these variables, you can look for their definition in work-2.12.5.3/wrapper-script or work-x86-2.12.5.3/wrapper-script.)
Alternatively, you can add double quotes around $libDir, $binDir and all other parameters, to have a uniform style, but I guess the best choice is to drop them where they are not required (i.e. for the INCDIR parameter.)
(Of course, in any case, we need to keep the double quotes for LN, READLINE_EXTRAS and EXTRA_CFLAGS.)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,131 @@
...
+}
+
+INSTALL()
+{
+	make install \
+		BLD_TYPE=calc-static-only \
+		LCC=gcc \
+		LN="ln -s" \
+		LDCONFIG= \
+		INCDIR="$includeDir" \
```
fbrosson: Same remark as above: You can drop the double quotes here if you wish.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,131 @@
...
+
+	prepareInstalledDevelLibs libcalc libcustcalc
+	fixPkgconfig
+
+	cd $dataDir/calc
+	find -type f -name \*.cal | xargs chmod +w
+	if [ -z "$secondaryArchSuffix" ]; then
+		packageEntries data \
+			$dataDir/calc
+	fi
```
fbrosson: Can you please replace this block:
```
		packageEntries data \
			$dataDir/calc
	fi
```
by these two blocks:
```

		# Some files in $dataDir/calc/help/ have code snippets with
		# long shebangs that we can shorten. Currenly, these files are
		# cscript, full, libcalc and usage.
		sed -i -e "1 s|$binDir/calc |/$relativeBinDir/calc |" \
			$dataDir/calc/help/*

		packageEntries data \
			$dataDir/calc
	else
		rm -rf $dataDir/calc $manDir
	fi

	# Shorten the paths to calc in the shebangs of the files in cscript.
	sed -i -e "1 s|$binDir/calc |/$relativeBinDir/calc |" \
		$libDir/calc/cscript/*
```
The changes in the first block will ensure that we don't ship the /system/data/calc/ in the base package of secondary arch builds. It also fixes the shebangs in a few files in /system/data/calc/data/help/.

The second block fixes the shebangs in the scripts that will go to /system/lib/calc/cscripts/ (on primary arch) or  /system/lib/x86/calc/cscripts/ (on x86 secondary arch).

fbrosson: The reason why we need to remove the $dataDir/calc directory when building for a secondary architecture is that during a secondary arch build haikuporter does not know we have a calc_data sub-package, so it _thinks_ that everything we did not move to the calc_devel package should go to the calc base package.
Removing that dir is therefore required otherwise /system/data/calc/ will also be in the calc_x86 hpkg.

raefaldhia:
@fbrosson, have you trying to install the package that contain the 'calc_data' ? I have some problem when installing the package, and today until friday i have some exam to do, i'am not sure i could fix this issue at that day, i'll do some easier task because i couldn't be productive at that day, so would you let me to do another tasks first? And I suggest don't merge this PR until i'am back

fbrosson:
OK, no problem. I'll validate you task on GCI and keep this PR open.
(I had no issues with the calc_data, but I guess this was because I added a couple of minor changes, as explained in my last review).

raefaldhia:
Updated, what actually this recipe needs, i'm forget, i'll post some TODO to PR desc. `calc-dynamic-only` now working.

fbrosson:
Sorry for this late reply. (I was AFK all day.)
Thanks for your update!

I've played with the recipe a week ago and I also found how to build the shared libs.
The first method requires patching the Makefile.
The second method, if I'm not mistalen, is easier: we only need to pass a parameter to make to make it think we are building on Linux. I need to retest this to make sure it works fine.

I guess you have also found how to build the shared lib.
So I suggest that we do everything in this PR.
Feel free to suggest your changes with updates to this PR.
I'll give you hints if you wish ;-)
The goal is to have a polished recipe with everything working, all in a single commit with you as the author ;-)

I need to review the work of anotther student and then I'll come back to this PR. Thanks!

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+		# Shorten some paths in the man page.
+		sed -i \
+			-e "s|#!$binDir/calc\>|#!/$relativeBinDir/calc|" \
+			-e "s|$prefix/|/system/|" \
+			$manDir/man1/calc.1
+
+		# Some files in $dataDir/calc/help/ have code snippets with
+		# long shebangs that we can shorten. Currently, these files are
+		# cscript, full, libcalc and usage.
+		sed -i -e "1 s|$binDir/calc |/$relativeBinDir/calc |" \
```
fbrosson: Can you please replace:
```
		sed -i -e "1 s|$binDir/calc |/$relativeBinDir/calc |" \
			$dataDir/calc/help/*
```
by
```
		sed -i \
			-e "s|$binDir/calc\>|/$relativeBinDir/calc|" \
			-e "s|$prefix/|/system/|g" \
			$dataDir/calc/help/*
```
(The last line is unchanged. I'm only copying it here to provide the full command.)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+			$dataDir/calc/help/*
+
+		packageEntries data \
+			$dataDir/calc
+	else
+		rm -rf $dataDir/calc $manDir
+	fi
+
```
fbrosson: Can you please insert the following block here? 
```

	# Shorten the paths to bin/calc in several shebangs.
	sed -i -e "1 s|$binDir/calc\>|/$relativeBinDir/calc|" \
		$libDir/calc/cscript/*

```

fbrosson: BTW, I think it might be possible to simplify the patch by passing the appropriate parameters to the make command, like it is already the case for HAVE_USTAT, instrad of patching the Makefile. I did not try that yet, however.

raefaldhia: @fbrosson Some of them that i patched in makefile ex: HAVE_STDLIB_H instead commandline because there's bug it wont apply to the makefile when i patch it with commandline.

fbrosson: OK, no problem.

With the changes in my last review we'll have a really polished recipe anyway ;-)

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,155 @@
...
+		# Shorten the paths to bin/calc in several shebangs.
+		sed -i -e "1 s|$binDir/calc\>|/$relativeBinDir/calc|" \
+			$libDir/calc/cscript/*
+
+		packageEntries data \
+			$dataDir/calc
+	else
+		rm -rf $dataDir/calc $manDir
+	fi
+
```
raefaldhia: Oops sorry

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,155 @@
...
+	cmd:cmp
+	cmd:touch
+	cmd:less
+	cmd:strip
+	"
+
+BUILD()
+{
+	make target=Haiku \
+		INCDIR="$includeDir" \
```
fbrosson: I think it might be a good idea to replace:
```
	make target=Haiku \
		INCDIR="$includeDir" \
```
by
```
	make \
		target=Haiku \
		INCDIR=$includeDir \
```
or maybe just drop the double quotes around $includeDir, or add double quotes around $libDir and $binDir to have a consistent style.

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,155 @@
...
+		LN="ln -s" \
+		LDCONFIG= \
+		WNO_ERROR_LONG_LONG=
+}
+
+INSTALL()
+{
+	make install \
+		target=Haiku \
+		INCDIR="$includeDir" \
```
fbrosson: I think it might be a good idea to replace:
```
		INCDIR="$includeDir" \
```
by
```
		INCDIR=$includeDir \
```
in order to have a consistent style with the way we're passing $libDir and $binDir.

fbrosson:
Thanks @raefaldhia for your great work!

To the project leaders: Shall we merge this PR?

raefaldhia:
@fbrosson Thank you too for all of your guide and correction xd.

pulkomandy:
> To the project leaders: Shall we merge this PR?

Why not?

@fbrosson: btw, there is not a very clear "project leaders" at HaikuPorts. With all the work you are doing here, you are part of it I think. So if you don't have any more improvements to suggest, let's merge it. We can always come back and fix things as needed.

fbrosson:
Thanks @pulkomandy for your kind reply ;-)

Well, I was just wondering if other contributors had noticed anything that could be improved before the merge, like, for instance, simplifying the patch by passing some extra parameters to the make command instead of changing them in the Makefile.

Well, I guess there is already so much great work in this PR (thanks to @raefaldhia) that, as you say, we can merge it now. It even has all the polishing we could identify, regarding the long /packaging/-style paths.

fbrosson:

PR #910, opened 2016-12-05
Created apng2gif recipe

R167:
Related to gif2apng recipe.

### DIFF
--- a/media-gfx/apng2gif/apng2gif-1.7.recipe
+++ b/media-gfx/apng2gif/apng2gif-1.7.recipe
@@ -1,0 +1,51 @@
+SUMMARY="Converts APNG animations into GIF format"
+DESCRIPTION="This program converts APNG animations into animated GIF format. \
+It performs the reverse of gif2apng."
+HOMEPAGE="http://apng2gif.sourceforge.net/"
+COPYRIGHT="2009-2015 Max Stepin"
+LICENSE="Zlib"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/apng2gif/apng2gif-$portVersion-src.zip"
+CHECKSUM_SHA256="6c7b2cce851378f2fc8f6b490ce098e9cf0038aa05da7099771c8be94620e6ca"
+SOURCE_DIR=""
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	LIBS="-lstdc++.r4 -lpng -lz"
+else
+	LIBS="-lstdc++ -lpng -lz"
+fi
+
+PROVIDES="
+	apng2gif = $portVersion
+	cmd:apng2gif = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libpng
+	lib:libz
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libpng
+	devel:libz
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:ld
+	cmd:make
+	"
+
+BUILD()
+{
+	make LIBS="$LIBS"
+}
+
+INSTALL()
+{
+	install -d -m 755 $binDir $docDir
+	install -t $binDir -c -m 555 -s apng2gif
+	install -t $docDir -c -m 444 readme.txt
+}
--- a/media-gfx/gif2apng/gif2apng-1.9.recipe
+++ b/media-gfx/gif2apng/gif2apng-1.9.recipe
@@ -4,13 +4,13 @@ Usually it makes the files smaller."
 HOMEPAGE="http://gif2apng.sourceforge.net/"
 COPYRIGHT="2009-2013 Max Stepin"
 LICENSE="Zlib"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://downloads.sourceforge.net/gif2apng/gif2apng-$portVersion-src.zip"
 CHECKSUM_SHA256="3b21308e935d799b3ffb4a86c6e00ffa4cb9b3f72f52d58d51c66eb0574ae7d2"
 SOURCE_DIR=""
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="x86"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
 
 PROVIDES="
 	gif2apng$secondaryArchSuffix = $portVersion
@@ -27,8 +27,8 @@ BUILD_REQUIRES="
 	"
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
-	cmd:make
 	cmd:ld$secondaryArchSuffix
+	cmd:make
 	"
 
 BUILD()
@@ -38,6 +38,7 @@ BUILD()
 
 INSTALL()
 {
-	mkdir -p $prefix/bin
-	cp gif2apng $prefix/bin
+	install -d -m 755 $prefix/bin $docDir
+	install -t $prefix/bin -c -m 555 -s gif2apng
+	install -t $docDir -c -m 444 readme.txt
 }


fbrosson:
Thanks!
I'm preparing a review right now, with some polishing ;-)

media-gfx/apng2gif/apng2gif-1.7.recipe
```diff
@@ -0,0 +1,45 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	apng2gif$secondaryArchSuffix = $portVersion
+	cmd:apng2gif = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
```
fbrosson: That's not required, by we like to sort alphabetically the **`lib:*`** lines in REQUIRES.

media-gfx/apng2gif/apng2gif-1.7.recipe
```diff
@@ -0,0 +1,45 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libz$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
```
fbrosson: That's not required, by we like to sort alphabetically the **`devel:*`** lines in BUILD_REQUIRES.

media-gfx/apng2gif/apng2gif-1.7.recipe
```diff
@@ -0,0 +1,45 @@
...
+
+BUILD()
+{
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $prefix/bin
+	cp apng2gif $prefix/bin
```
fbrosson: Could you please use the install command, instead of mkdir and cp?
We have a recipe which you can get inspiration from:
app-misc/hexcompare/hexcompare-1.0.4.recipe

R167:
Sorry about that, just tacked the deps onto the end of the list cause I was rushing. Also, updated where I had previously done the cp in gif2apng. Would that change require a bump to revision?

fbrosson:
LGTM.
I just need to build it and I'll merge this PR and validate your task.

Regarding the revision bump for cosmetic changes in other recipes, there is no need to bump if the resulting package is identical.
(But we usualy don't edit a single recipe with just cosmetic edits.)

If you wish, however, you might claim this task:
[Get the Lint out of 5 Haikuports recipes (1)](https://codein.withgoogle.com/tasks/5670956970278912/)

R167:
Sure thing. Only made that change because I had just copied the majority of that recipe over and thought it best to use a more standardized process than the mkdir && cp workflow. Will fix that latest one too. :-)

fbrosson:
Hmm, that's a very good idea to fix apng2gif with gif2apng! Well done!

fbrosson:
May I suggest an alternate commit message? What about this:
```
apng2gif: new recipe. gif2apng: a bit of polishing.
```

Hmm, you may need to add **`cmd:install`** to BUILD_PREREQUIRES. I need to figure this out.
If cmd:install is provided by the coreutils hpkg then it is not required. If not, then it should probably be added here.

R167:
just checked and everything went fine without `cmd:install`. Also, just squashed commits and updated primary commit message.

media-gfx/apng2gif/apng2gif-1.7.recipe
```diff
@@ -0,0 +1,54 @@
...
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandSuffix=$secondaryArchSuffix
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
```
fbrosson: Hmm, let's drop this block (in both apng2gif and gif2apng).
There are several reasons for this:
* We are not using commandSuffix.
* We are only using commandBinDir in one place, INSTALL, and we can replace **`commandBinDir`** by **`$prefix/bin`** for this recipe on all architectures, both primary and secondary.
* This recipe does not provide any lib, so after we drop x86_gcc2 (some time in the future) we will be able to completely drop support for secondary architectures for this recipe.

R167: Sure thing. Was initially going that route, but wasn't sure if you were going to end up having me add it.

fbrosson:
OK, since you have successfully tested without cmd:install, I guess we don't need it in BUILD_PREREQUIRES.

fbrosson:
Looks OK. May I ask you a final touch?
It would be nice to drop the detailed commit message, I mean, only keep a one-line commit title.

fbrosson:
Cool! Thanks for the update of the commit message.

This recipe was really easy, wasn't it?
BTW, if you claim all the easy ones there won't be any left for the other students!
Would you mind taking a few hard tasks, from time to time?
We've got a few that might be hard (but might turn out to be not very hard): libdmtx, libusbmuxd, iec16022. Thanks!

R167:
Sure thing. I just did this one cause I just worked on gif2apng recipe and knew about its weird quirks (Source_dir needs to be an empty string and all the fun oddities for make), so I just figured I would do it. Definitely want to look more into the library ones if only to better understand how they work for linking, etc. Will do! :)

fbrosson:
Oops, it would be better, after all, to bump REVISION (for gif2apng only, of course) because now the package includes the readme.txt file in $docDir.
(Sorry, I missed this, or maybe your question came before I had mentionned using cmd:install.)

R167:
done.

fbrosson:
Thanks!
I'm having a look at the Makefile and it seems that the reason why it does not build on x86_gcc2 primary arch out of the box is that it links against **`libstdc++`**, which we have on x86 and x86_64, but which is missing on x86_gcc2 (to be precise, it **_is_** available but with a slightly **_different_** name: **`libstdc++.r4`**.)

I did workaround this issue myself on another package (not yet in the repo) which has the same dependency, but did never submit it because it still needs some polishing.

Would you mind if I ask @korli or @pulkomandy if it would be desirable to try to make apng2gif build on x86_gcc2 primary arch?

R167:
I don't mind. We could set it up so that it uses the correct one depending on target architecture, but that sounds like excessive complexity if the intention is to EOL gcc2 soon (not like it was already EOL everywhere else :)). I remember when I found that before, but ended up not using it because of a C99 requirement or something like that. IDK: I am tired and rambly right now. Judgement call is up to you.

Also, where can I find docs for adding libs in haikuports? Didn't see anything specific in the wiki, and I am looking at tackling the libp11 task next. If not any docs, any particular recipe/example to look at? (Biggest thing I really don't know how to do is going to be how to test if I did everything right). Thanks again, @fbrosson!

fbrosson:
This is what we might try:
```
	PATCHES="$PATCHES
		apng2gif-$portVersion-gcc2.patchset
		"
	LIB_STDCPP=-lstdc++.r4
else
	LIB_STDCPP=-lstdc++
fi
```
And in the patch we would replace -lstdc++ by $(LIB_STDCPP).
And in BUILD we would replace **`make`** by
```
LIB_STDCPP=$LIB_STDCPP make
```

Reminder: I don't know if this would desirable, so you might wish to wait for @korli's or @pulkomandy's or @mmuman's opinion about this.

fbrosson:
Hmm, I wonder if this would work, without any patch:
```
	LIBS="-lstdc++.r4 -lpng -lz"
else
	LIBS="-lstdc++ -lpng -lz"
fi
```

pulkomandy:
The proper way would be to use g++ as a linker, instead of calling ld or gcc directly. g++ knows which compiler flags to add so there is no need for an explicit -lstdc++.
I don't know if it is possible with the buildsystem here, if it is not, using the effective target arch seems the only way.

fbrosson:
Since the Makefile defines CC=gcc, I _thought_ it would be possible, with **`CC=g++ make`** or with **`make CC=g++`**, but none of these work:
```
/sources/apng2gif.cpp: In function `int load_apng(char *, vector<APNGFrame,allocator<APNGFrame> > &, unsigned int *)':
/sources/apng2gif.cpp:349: warning: comparison between signed and unsigned
/packages/gcc-2.95.3_2014_10_14-3/.self/develop/tools/i586-pc-haiku/bin/ld: cannot find -lstdc++
collect2: ld returned 1 exit status
make: *** [Makefile:8: all] Error 1
```

I tried, however, to use this (on x86_gcc2 primary arch):
```
make LIBS="-lstdc++.r4 -lpng -lz"
```
and it works perfectly!
Shall we drop SECONDARY_ARCHITECTURES and $secondaryArchSuffix and, instead, use the little conditional setting for LIBS?

fbrosson:
BTW, I forgot to thank you for your reply!!!
(I was too excited with testing the LIBS= workaround ;-)

media-gfx/apng2gif/apng2gif-1.7.recipe
```diff
@@ -0,0 +1,46 @@
...
+HOMEPAGE="http://apng2gif.sourceforge.net/"
+COPYRIGHT="2009-2015 Max Stepin"
+LICENSE="Zlib"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/apng2gif/apng2gif-$portVersion-src.zip"
+CHECKSUM_SHA256="6c7b2cce851378f2fc8f6b490ce098e9cf0038aa05da7099771c8be94620e6ca"
+SOURCE_DIR=""
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
fbrosson: Shall use the LIBS= workaround?
If I'm not mistaken, this is what we need to do:

1) replace:
```
SECONDARY_ARCHITECTURES="x86"
```
by
```

	LIBS="-lstdc++.r4 -lpng -lz"
else
	LIBS="-lstdc++ -lpng -lz"
fi
```

2) drop all occurences off **`secondaryArchSuffix`**;

3) replace **`$prefix/bin`** by **`$binDir`** (in INSTALL)

(But let's see what our other contibutors thing about this.)

pulkomandy: yes, looks fine.

fbrosson: OK. fine.
Let me update the GCI task with a "Request More Work", otherwise we might forget and Google will think Haiku is not responding quickly enough ;-)

fbrosson: Oops, the commit title I suggested earlier is not very nice, after all.
So when you update this PR could you please replace
```
apng2gif: new recipe. gif2apng: a bit of polishing.
```
by
```
apng2gif: new recipe. gif2apng: polishing."
```
(My _"a bit of"_ was _a bit_ silly, wasn't it ;-)
Thanks!

R167:
these changes work for apng2gif, but the libs override doesn't work with gif2apng, erroring out on:
```
gcc -Wall -pedantic -s -o gif2apng obj/./gif2apng.o obj/7z/WindowIn.o obj/7z/LZMA.o obj/7z/CRC.o obj/7z/DeflateDecoder.o obj/7z/LZMAEncoder.o obj/7z/LSBFDecoder.o obj/7z/IInOutStreams.o obj/7z/7zdeflate.o obj/7z/HuffmanEncoder.o obj/7z/LZMADecoder.o obj/7z/7zlzma.o obj/7z/AriBitCoder.o obj/7z/LSBFEncoder.o obj/7z/LiteralCoder.o obj/7z/LenCoder.o obj/7z/WindowOut.o obj/7z/InByte.o obj/7z/DeflateEncoder.o obj/7z/OutByte.o obj/zopfli/zlib_container.o obj/zopfli/cache.o obj/zopfli/util.o obj/zopfli/hash.o obj/zopfli/katajainen.o obj/zopfli/tree.o obj/zopfli/gzip_container.o obj/zopfli/lz77.o obj/zopfli/squeeze.o obj/zopfli/zopfli_lib.o obj/zopfli/blocksplitter.o obj/zopfli/deflate.o -lstdc++.r4 -lm -lz
obj/zopfli/util.o: In function `ZopfliGetDistExtraBits':
util.c:(.text+0x21): undefined reference to `__builtin_clz'
obj/zopfli/util.o: In function `ZopfliGetDistExtraBitsValue':
util.c:(.text+0x5c): undefined reference to `__builtin_clz'
obj/zopfli/util.o: In function `ZopfliGetDistSymbol':
util.c:(.text+0xac): undefined reference to `__builtin_clz'
collect2: ld returned 1 exit status
Makefile:20: recipe for target 'gif2apng' failed
```

media-gfx/apng2gif/apng2gif-1.7.recipe
R167: LIBS works great with apng2gif, but gif2apng errors out on undefined references (i did make sure that it was using the right version). See error below. (added in patches for that, but it did NOT have the PATCHES var set. changes didn't succeed in fixing it anyways)

fbrosson:
Everything is OK.
Although I can squash your commits _"on the fly before the merge"_, but it would look much better if you do it yourself.
I guess you were waiting for the signal to squash. You got it ;-)
I've just validated your task on GCI, but don't forget to squash your commits here ;-)
Thanks!

R167:
Sure thing. Will squash for you now. Was just waiting for your signal in case I needed to revert anything. :)

fbrosson:
That's perfect! Thanks!

fbrosson:
@R167, I wrote [this 2 weeks ago](#issuecomment-264786772) because you were by far the most active student at that time.
> Would you mind taking a few hard tasks, from time to time?

But it was not fair (from my side) to suggest that you took harder tasks, even only from time to time.
So please ignore that old suggestion and, instead, **_do_** feel free to claim **_any_** task you wish!
BTW, I'd love to see more recipes claimed ;-)

PR #911, opened 2016-12-05
Fix broken recipe that caused by HOMEPAGE

raefaldhia:


raefaldhia:
omg what happened to the commit mesage: "terminus_ttf: any architecture, disable source package"

PR #912, opened 2016-12-05
Fix broken recipe that caused by HOMEPAGE

raefaldhia:
Added new version:
- [x] beshare

Fix broken thing & reformat the style to following guidelines:
- [x] belife
- [x] beshare
- [x] catkeyseditor
- [x] colors
- [x] copynametoclipboard
- [x] desknotes
- [x] genesis_commander

There's another recipe that produce same errors, but i couldn't fix that yet, i've tried but i got error something like couldn't find file `packagename.dependencyinfo` in `repository/`.

List that couldn't be fixed:
- [ ] beae
- [ ] befar
- [ ] bevexed


haiku-apps/beshare/beshare-2.31.recipe
```diff
@@ -49,19 +53,17 @@ BUILD_PREREQUIRES="
 	makefile_engine
 	"
 
-PATCHES="beshare-$portVersion.patchset"
-
 BUILD()
 {
 	ln -sf ../../sources-2/muscle muscle
```
fbrosson: Not required, but it could be nice to use
```
	ln -s -f -r $sourceDir2/muscle

File: haiku-apps/beshare/beshare-2.31.recipe [hunk 2]
```diff
@@ -13,16 +13,19 @@ have to click \"refresh\" or anything first)
...
 given host for efficiency."
+HOMEPAGE="https://github.com/HaikuArchives/BeShare"
 COPYRIGHT="1999-2012 Jeremy Friesner"
 LICENSE="
 	Public Domain
 	libsanta
 	"
-HOMEPAGE="https://github.com/HaikuArchives/BeShare"
-SOURCE_URI="$(HOMEPAGE)/archive/b721fcc.tar.gz"
+SOURCE_URI="${HOMEPAGE}/archive/b721fcc.tar.gz"
```
fbrosson: Wouldn't it be slightly better to use:
```
srcGitRev="b721fcc4a4297b32738994b9c427d7db26c2c8fc"
SOURCE_URI="$HOMEPAGE/archive/$srcGirRev.tar.gz"
```
BTW, the latest revision is:
```
srcGitRev="2345e48664fddfff1d9a968d1da2a185dbbdb50c"
```
but switching to that would require updating CHECKSUM_SHA256SUM and testing the build.

haiku-apps/beshare/beshare-2.31.recipe
```diff
@@ -13,16 +13,19 @@ have to click \"refresh\" or anything first)
 - BeShare supports any number of simultaneous uploads and downloads \
 in parallel, but will serialize simultaneous file requests to any \
 given host for efficiency."
+HOMEPAGE="https://github.com/HaikuArchives/BeShare"
 COPYRIGHT="1999-2012 Jeremy Friesner"
 LICENSE="
 	Public Domain
 	libsanta
 	"
```
fbrosson: While at it, shouldn't we change this to:
```
LICENSE="Public Domain
	libsanta"
```

haiku-apps/beshare/beshare-2.32.recipe
```diff
@@ -0,0 +1,68 @@
...
+- BeShare includes a rudimentary chat mechanism for you social types.
+- BeShare supports any number of simultaneous uploads and downloads \
+in parallel, but will serialize simultaneous file requests to any \
+given host for efficiency."
+HOMEPAGE="https://github.com/HaikuArchives/BeShare"
+COPYRIGHT="1999-2012 Jeremy Friesner"
+LICENSE="
+	Public Domain
+	libsanta
+	"
```
fbrosson: Same thing here: shouldn't we change this to:
```
LICENSE="Public Domain
	libsanta"
```

haiku-apps/beshare/beshare-2.32.recipe
```diff
@@ -0,0 +1,68 @@
...
+SOURCE_URI="${HOMEPAGE}/archive/v2.32.tar.gz"
+CHECKSUM_SHA256="f5634da15a59f1a822bb818fee698ea147b3817300660d0877ed066259a388f5"
+SOURCE_URI_2="https://public.msli.com/lcs/muscle/muscle6.11.zip"
+REVISION="1"
```
fbrosson: REVSION goes between LICENSE and SOURCE_URI.

haiku-apps/belife/belife-1.0.0.recipe
```diff
@@ -14,7 +14,9 @@ HOMEPAGE="https://github.com/HaikuArchives/BeLife"
 COPYRIGHT="2006 Studio-33"
 LICENSE="BeLife License"
 REVISION="2"
-SOURCE_URI="$(HOMEPAGE)/archive/96e31e8b4543723c7c8ea4a3d0c57c8cd52f82ca.tar.gz"
+SOURCE_URI="${HOMEPAGE}/archive/96e31e8b4543723c7c8ea4a3d0c57c8cd52f82ca.tar.gz"
+CHECKSUM_SHA256="2566dc527e336277173179c7cce0771579257e19bbcd853377cd6112547a08d1"
+SOURCE_DIR="BeLife-96e31e8b4543723c7c8ea4a3d0c57c8cd52f82ca"
 
 ARCHITECTURES="x86 x86_gcc2"
```
raefaldhia: Should x86_gcc2 come first?

haiku-apps/beshare/beshare-2.31.recipe
raefaldhia: `2345e48664fddfff1d9a968d1da2a185dbbdb50c` is still categorized beshare  2.31 isn it?

haiku-apps/belife/belife-1.0.0.recipe
fbrosson: If it's a new recipe then I put x86_gcc2 first. Otherwise I leave it the way it is.
In other words, there is no need to change this line if the only change is swapping the order the archictectures appear.

haiku-apps/beshare/beshare-2.31.recipe
fbrosson: I don't know, however, if everyone would like such a change ;-)
(I guess some contributors prefer merging one-letter options. IMHO merging one-letter options is OK for commands typed in _interactive_ shells, and should not be used in scripts. Scripts should be readable even to someone who does not know anything about the ln command. But that's only my personal point of view, of course.)

haiku-apps/beshare/beshare-2.31.recipe
fbrosson: Exact. How did I not notice that before?!!!
So shall we just change this to:
```
srcGitRev="b721fcc4a4297b32738994b9c427d7db26c2c8fc"
SOURCE_URI="$HOMEPAGE/archive/$srcGirRev.tar.gz"
```
(which is the srcGitRev already in the recipe, in its shortened version), and leave 2.32 for a separate commit? (by you or anyone else)
Comments welcome, as usual, from all contributors ;-)

pulkomandy: There is already a recipe for 2.32, so why don't we just delete this old one for 2.31?

fbrosson: I don't think so: [haiku-apps/beshare](https://github.com/haikuports/haikuports/tree/master/haiku-apps/beshare)
Maybe you have a local file.

fbrosson: Ah, ok. (That's what happens when I do 2 things at the same time ;-)))

So, in that case, shall we use:
```
srcGitRev="2345e48664fddfff1d9a968d1da2a185dbbdb50c"
SOURCE_URI="$HOMEPAGE/archive/$srcGirRev.tar.gz"
```
in 2.32 (and drop, as you suggested, 2.31) ?

haiku-apps/beshare/beshare-2.31.recipe
fbrosson: Sure, but not all programs use getopt. Some programs use long option names with a single dash, so not merging single-letter options together is imho a nice way to attract new contributors.
It does not cost much to have clearly separated options, and it reduces the risk of visitors saying to themselves: "ah, I definitely don't understand scripts" when reading all these merged options.
Making the effort of being novice-friendly in our recipes can only increase the audience.

Of course, I'm not telling we should change all scripts. I just think that when we change a line that has such merged options, we could, while at it, also switch them to separate options.

That said, I do admit that if this ever increases the audience, the increase will probably be very small.
I also would like to say that this is just a suggestion, and I personally don't care if we use merged options. I just think that a system which targets average users should try to use, when possible, novice-friendly scripts, even for things (e.g. recipes) that these users are probably not going to read.

haiku-apps/beshare/beshare-2.31.recipe
fbrosson: I would certainly use srcGitRev, since several other recipes which point to a specific commit already use this very same (custom) parameter.
I see two advantages in using this: First, when we switch to a newer commit revision, the changes are in a single line. Secondly, the length of the _short_ alias might change in the future, so if/when such changes occur all recipes that use short commit refs would require an update.

Regarding the removal of catkeyseditor `0.1.0`, I think it would be better to keep it.
But I'm not against its removal if the project leaders say we should.

raefaldhia:
Updated.

haiku-apps/colors/colors-2.3.recipe
```diff
@@ -11,10 +11,13 @@ The Features:
  * With an eye-dropper tool you can pick a color from anywhere on your desktop."
 HOMEPAGE="https://github.com/jscipione/Colors"
 COPYRIGHT="2001-2008 Werner Freytag
-	2009-2013 John Scipione"
+		2009-2013 John Scipione"
```
fbrosson: The third line of COPYRIGHT has 2 tabs, while it should only have one tab.

haiku-apps/colors/colors-2.3.recipe
```diff
@@ -23,8 +26,7 @@ PROVIDES="
 	colors$secondaryArchSuffix = $portVersion
 	app:Colors = $portVersion
 	"
-REQUIRES="
-	"
+
 BUILD_PREREQUIRES="
 	haiku${secondaryArchSuffix}_devel
```
fbrosson: It would be better to move the first line of BUILD_PREREQUIRES to BUILD_REQUIRES (instead of removing tha latter completely):
```
REQUIRES="
	haiku${secondaryArchSuffix}_devel"
	"
BUILD_PREREQUIRES="
	makefile_engine
	cmd:gcc$secondaryArchSuffix
	cmd:make
	"
```

haiku-apps/genesis_commander/genesis_commander-0.42.recipe
```diff
@@ -15,6 +18,7 @@ PROVIDES="
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
+
```
fbrosson: There should be no empty line between REQUIRES and BUILD_PREREQUIRES.

fbrosson: Source: [HaikuPorter-Guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering).

raefaldhia: omg i do always checking all files multiple times but yeah i can't still see that haha.

raefaldhia:
Updated.

haiku-apps/belife/belife-1.0.0.recipe
```diff
@@ -14,7 +14,10 @@ HOMEPAGE="https://github.com/HaikuArchives/BeLife"
 COPYRIGHT="2006 Studio-33"
 LICENSE="BeLife License"
 REVISION="2"
-SOURCE_URI="$(HOMEPAGE)/archive/96e31e8b4543723c7c8ea4a3d0c57c8cd52f82ca.tar.gz"
+srcGitRev="96e31e8b4543723c7c8ea4a3d0c57c8cd52f82ca"
+SOURCE_URI="${HOMEPAGE}/archive/${srcGitRev}.tar.gz"
```
raefaldhia: Yeah i've do this before, but i reverted it because that secondaryArchSuffix_devel give me a problem.

raefaldhia:
Updated.

haiku-apps/colors/colors-2.3.recipe
```diff
@@ -24,9 +27,13 @@ PROVIDES="
 	app:Colors = $portVersion
 	"
 REQUIRES="
+	haiku${secondaryArchSuffix}
```
fbrosson: Oops, I had missed this one. You can change it to:
```
	haiku$secondaryArchSuffix
```

haiku-apps/beshare/beshare-2.32.recipe
```diff
@@ -36,27 +35,26 @@ REQUIRES="
...
 	"
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
 	cmd:make
+	cmd:wget
 	makefile_engine
 	"
 
-PATCHES="beshare-$portVersion.patchset"
```
fbrosson: I just noticed this line is being removed.
Well, if `haiku-apps/beshare/patches/beshare-2.31.patchset` is no longer needed for 2.32 then it might be a good idea to remove from the repo:
```
cd haikuports/haiku-apps/beshare/patches
git rm beshare-2.31.patchset
git commit --amend --no-edit beshare-2.31.patchset
```
I guess the patch is not needed anymore, but I would wait for a confirmation from  @pulkomandy and/or @mmuman.
BTW, if you have tried to build beshare-2.32, did it work fine?

raefaldhia: @fbrosson yes, it works fine, and yeah you're right it should being removed, look like the patch just applied in another commit ex:
https://github.com/HaikuArchives/BeShare/blob/master/source/ChatWindow.cpp#L1175
https://github.com/HaikuArchives/BeShare/blob/master/source/ChatWindow.cpp#L14

But iam not sure too, since not all of them applied, just some.

PR #916, opened 2016-12-06
Vim patchset: vim GUI requires libstdc++

miqlas:
With this patch i was able to build it on x64 Haiku.


fbrosson:
This is great news!

To anyone who would like to merge this, please do **not** use the "Create a merge commit", as we have the **"Squash and merge"** feature that GitHub has added a few months ago.

**"Squash and merge"** is the **prefered** merge method. Thanks!

miqlas:
Please, do not merge yet, it is broken on x86.

mmuman:
Are you sure you just don't need to fix the test just the line below ? We don't use gcc**4** anymore, do we?

miqlas:
You have right, i'll fix it.

PR #923, opened 2016-12-11
Xerces-C: some supporting libs added.

miqlas:
Solves the missing symbol: "connect"


dev-libs/xerces_c/xerces_c-3.1.4.recipe
```diff
@@ -64,7 +74,10 @@ BUILD_PREREQUIRES="
 BUILD()
 {
 	autoreconf -fi
-	runConfigure ./configure
+	LDFLAGS="-lnetwork -lbsd" \
```
korli: You should patch m4/xerces_netaccessor_selection.m4 line 29 instead.

korli:
Please precise the need to enable curl, when a native "Sockets-based NetAccessor" option seems to be available (I'd think more features available) in your commit message.
Also the choice of iconv vs ICU.

miqlas:
curl removed.

dev-libs/xerces_c/xerces_c-3.1.4.recipe
miqlas: Already did it.

PR #924, opened 2016-12-11
LMMS: enabled x64 arch

miqlas:


miqlas:
I thought, that it is required.

-- Z.

2016.12.11. dátummal, 22:35 időpontban Jérôme Duval <notifications@github.com> írta:

korli:
Well x86_64 is only available as a primary arch. How would you be able to test such a change anyway?

PR #928, opened 2016-12-13
Fish shell: updated Requires for better compatibility...

miqlas:
Requires according to bash recipe.


miqlas:
You have actually right, fish doesn't depend on readline libs. But the curses update could be handy. Sadly the garbage ouput during command line edit still there :(
I will check it later without customisation and fresh env and clean settings. If it still there, then it should be lying in the ncurses Haiku port, or in the Haiku Terminal.
I'm not sure if it helps if i try to ssh from my osx box...

PR #931, opened 2016-12-16
Amend "calc: new recipe for a C-style arbitrary precision calculator"

raefaldhia:
Oppss.

In current recipe we removed ldconfig due to there's no ldconfig, but i just found out that ldconfig is exist with name ld in haiku.

I found that from:
https://github.com/haikuports/haikuports/blob/75a3cf5384f650242f4b8a1094ab305d33ed99e7/app-emulation/pcem/pcem-10.1.recipe#L39

should i revert or make new pr?

But look like there's no big impact either we add it or not.

Amend haikuports/haikuports#901

### DIFF
--- a/sci-mathematics/calc/calc-2.12.5.3.recipe
+++ b/sci-mathematics/calc/calc-2.12.5.3.recipe
@@ -1,0 +1,156 @@
+SUMMARY="C-style arbitrary precision calculator"
+DESCRIPTION="Calc is an interactive calculator which provides for easy large \
+numeric calculations, but which also can be easily programmed for difficult or \
+long calculations. It can accept a command line argument, in which case it \
+executes that single command and exits. Otherwise, it enters interactive mode. \
+In this mode, it accepts commands one at a time, processes them, and displays \
+the answers. In the simplest case, commands are simply expressions which are \
+evaluated."
+HOMEPAGE="http://www.isthe.com/chongo/tech/comp/calc/"
+COPYRIGHT="1999-2016 Landon Curt Noll"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://www.isthe.com/chongo/src/calc/calc-$portVersion.tar.bz2"
+CHECKSUM_SHA256="0fcf60917efb10aab145f1180c278a3ab63d789aad6d5e5f066a4458b5f9b692"
+PATCHES="calc-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	calc$secondaryArchSuffix = $portVersion
+	cmd:calc$secondaryArchSuffix = $portVersion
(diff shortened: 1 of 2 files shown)


raefaldhia:
Or maybe you could do `git --amend` for me to make commit message clear.
There's just little difference, about `LDCONFIG=ln` and add `cmd:ld$secondaryArchSuffix`

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,156 @@
...
+		HAVE_UNISTD_H=YES \
+		HAVE_FPOS_POS=1 \
+		HAVE_UID_T=1 \
+		HAVE_USTAT=1 \
+		EXTRA_CFLAGS="-DUSE_TERMIOS" \
+	 	USE_READLINE=-DUSE_READLINE \
+		READLINE_LIB=-lreadline \
+		READLINE_EXTRAS="-lhistory -lncurses" \
+		LN="ln -s" \
+		LDCONFIG=ld \
```
raefaldhia: Here

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,156 @@
...
+	cmd:sed
+	cmd:diff
+	cmd:egrep
+	cmd:ctags
+	cmd:xargs
+	cmd:cmp
+	cmd:touch
+	cmd:less
+	cmd:strip
+	cmd:ld$secondaryArchSuffix
```
raefaldhia: Here

sci-mathematics/calc/calc-2.12.5.3.recipe
```diff
@@ -0,0 +1,156 @@
...
+		HAVE_UNISTD_H=YES \
+		HAVE_FPOS_POS=1 \
+		HAVE_UID_T=1 \
+		HAVE_USTAT=1 \
+		EXTRA_CFLAGS="-DUSE_TERMIOS" \
+	 	USE_READLINE=-DUSE_READLINE \
+		READLINE_LIB=-lreadline \
+		READLINE_EXTRAS="-lhistory -lncurses" \
+		LN="ln -s" \
+		LDCONFIG=ld \
```
raefaldhia: Here

raefaldhia:
Closing this, i need to test.

pulkomandy:
ldconfig and ld are not  the same thing.

ld is the linker, it is used to take a bunch of .o files (compiled by gcc) and make a library or executable from them.
ldconfig is a Linux specific tool, it is used to scan all libraries in the system (/usr/lib, /lib, ...) and generate a database/cache to map their soname to complete file paths.

On Haiku, there is no database/cache and the lookup is performed by opening the directories and looking for the files. We always make sure the soname matches the filename, which makes this not cost too much. As a result, ldconfig is not needed on Haiku.

raefaldhia:
@pulkomandy ok then thanks, i thought it's same since when i typed --help ldconfig and ld look alike.

PR #933, opened 2016-12-17
libical: Add new recipe

raefaldhia:
Adding new recipe for libcal, an implementation of icalendar protocols and data formats.

### DIFF
--- a/dev-libs/libical/libical-2.0.0.recipe
+++ b/dev-libs/libical/libical-2.0.0.recipe
@@ -1,0 +1,87 @@
+SUMMARY="An implementation of iCalendar protocols and data formats"
+DESCRIPTION="Libical is an Open Source implementation of the iCalendar \
+protocols and protocol data units. The iCalendar specification describes how \
+calendar clients can communicate with calendar servers so users can store \
+their calendar data and arrange meetings with other users."
+HOMEPAGE="http://libical.github.io/libical/"
+COPYRIGHT="1999-2015 Eric Busboom"
+LICENSE="MPL v1.0
+	GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://github.com/libical/libical/releases/download/v$portVersion/libical-$portVersion.tar.gz"
+CHECKSUM_SHA256="654c11f759c19237be39f6ad401d917e5a05f36f1736385ed958e60cf21456da"
+PATCHES="libical-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	libical$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libical$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libical_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libicalss$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libicalss_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libicalvcal$secondaryArchSuffix = $portVersion compat >= 2
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libicui18n$secondaryArchSuffix
+	lib:libicuuc$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libical${secondaryArchSuffix}_devel = $portVersion compat >= 2
+	devel:libical$secondaryArchSuffix = $portVersion compat >= 2
+	devel:libical_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	devel:libicalss$secondaryArchSuffix = $portVersion compat >= 2
+	devel:libicalss_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	devel:libicalvcal$secondaryArchSuffix = $portVersion compat >= 2
+	"
+REQUIRES_devel="
+	libical$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libicui18n$secondaryArchSuffix
+	devel:libicuuc$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:perl
+	"
+
+BUILD()
+{
+	cmake . \
+		-DCMAKE_INSTALL_PREFIX=$prefix \
+		-DCMAKE_INSTALL_LIBDIR=$libDir \
+		-DCMAKE_INSTALL_INCLUDEDIR=$incDir \
+		-DCMAKE_INSTALL_DATAROOTDIR=$dataDir \
+ 		-DUSE_BUILTIN_TZDATA=true
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	prepareInstalledDevelLibs \
+		libical \
+		libical_cxx \
+		libicalss \
+		libicalss_cxx \
+		libicalvcal
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir \
+		$libDir/cmake
+}
+
+TEST()
+{
+	make test
+}
--- a/dev-libs/libical/licenses/MPL v1.0
+++ b/dev-libs/libical/licenses/MPL v1.0
@@ -1,0 +1,117 @@
+MOZILLA PUBLIC LICENSE
+Version 1.0 
+
+1. Definitions.
+
+1.1. ``Contributor'' means each entity that creates or contributes to the creation of Modifications.
+1.2. ``Contributor Version'' means the combination of the Original Code, prior Modifications used by a Contributor, and the Modifications made by that particular Contributor.
+
+1.3. ``Covered Code'' means the Original Code or Modifications or the combination of the Original Code and Modifications, in each case including portions thereof.
+
+1.4. ``Electronic Distribution Mechanism'' means a mechanism generally accepted in the software development community for the electronic transfer of data.
+
+1.5. ``Executable'' means Covered Code in any form other than Source Code.
+
+1.6. ``Initial Developer'' means the individual or entity identified as the Initial Developer in the Source Code notice required by Exhibit A.
+
+1.7. ``Larger Work'' means a work which combines Covered Code or portions thereof with code not governed by the terms of this License.
+
+1.8. ``License'' means this document.
+
+1.9. ``Modifications'' means any addition to or deletion from the substance or structure of either the Original Code or any previous Modifications. When Covered Code is released as a series of files, a Modification is:
+
+A. Any addition to or deletion from the contents of a file containing Original Code or previous Modifications.
+
+B. Any new file that contains any part of the Original Code or previous Modifications.
+
+1.10. ``Original Code'' means Source Code of computer software code which is described in the Source Code notice required by Exhibit A as Original Code, and which, at the time of its release under this License is not already Covered Code governed by this License.
+
+1.11. ``Source Code'' means the preferred form of the Covered Code for making modifications to it, including all modules it contains, plus any associated interface definition files, scripts used to control compilation and installation of an Executable, or a list of source code differential comparisons against either the Original Code or another well known, available Covered Code of the Contributor's choice. The Source Code can be in a compressed or archival form, provided the appropriate decompression or de-archiving software is widely available for no charge.
+
+1.12. ``You'' means an individual or a legal entity exercising rights under, and complying with all of the terms of, this License or a future version of this License issued under Section 6.1. For legal entities, ``You'' includes any entity which controls, is controlled by, or is under common control with You. For purposes of this definition, ``control'' means (a) the power, direct or indirect, to cause the direction or management of such entity, whether by contract or otherwise, or (b) ownership of fifty percent (50%) or more of the outstanding shares or beneficial ownership of such entity.
+
+2. Source Code License.
+2.1. The Initial Developer Grant. 
+The Initial Developer hereby grants You a world-wide, royalty-free, non-exclusive license, subject to third party intellectual property claims:
+(a) to use, reproduce, modify, display, perform, sublicense and distribute the Original Code (or portions thereof) with or without Modifications, or as part of a Larger Work; and
+
+(b) under patents now or hereafter owned or controlled by Initial Developer, to make, have made, use and sell (``Utilize'') the Original Code (or portions thereof), but solely to the extent that any such patent is reasonably necessary to enable You to Utilize the Original Code (or portions thereof) and not to any greater extent that may be necessary to Utilize further Modifications or combinations.
+
+2.2. Contributor Grant. 
+Each Contributor hereby grants You a world-wide, royalty-free, non-exclusive license, subject to third party intellectual property claims:
+
+(a) to use, reproduce, modify, display, perform, sublicense and distribute the Modifications created by such Contributor (or portions thereof) either on an unmodified basis, with other Modifications, as Covered Code or as part of a Larger Work; and
+
+(b) under patents now or hereafter owned or controlled by Contributor, to Utilize the Contributor Version (or portions thereof), but solely to the extent that any such patent is reasonably necessary to enable You to Utilize the Contributor Version (or portions thereof), and not to any greater extent that may be necessary to Utilize further Modifications or combinations.
+
+
+
+3. Distribution Obligations.
+3.1. Application of License. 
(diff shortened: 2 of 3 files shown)


dev-libs/libical/libical-2.0.0~.recipe
```diff
@@ -0,0 +1,79 @@
...
+		-DCMAKE_INSTALL_DATAROOTDIR=$dataDir
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	prepareInstalledDevelLibs libical libical_cxx libicalss libicalss_cxx libicalvcal
```
fbrosson: This line is longer than 80 chars:
```
	prepareInstalledDevelLibs libical libical_cxx libicalss libicalss_cxx libicalvcal
```
So we need to split in at least 2 lines. Shall we split it with one lib per line, like this?
```
	prepareInstalledDevelLibs \
		libical \
		libical_cxx \
		libicalss \
		libicalss_cxx \
		libicalvcal
```

raefaldhia:
Actually i don't understand the `tarball` part but i hope my lastest commit already fix that.

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+DESCRIPTION="Libical is an Open Source implementation of the iCalendar \
+protocols and protocol data units. The iCalendar specification describes how \
+calendar clients can communicate with calendar servers so users can store \
+their calendar data and arrange meetings with other users."
+HOMEPAGE="http://libical.github.io/libical/"
+COPYRIGHT="Eric Busboom"
+LICENSE="MPL v1.0
+	GNU LGPL v2.1"
+REVISION="1"
+scrGitRev="6c4af23b0a95fd105f38e879908cbc80390f3716"
```
pulkomandy: This is not needed anymore since you now use the release tarball.

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+SOURCE_URI="https://github.com/libical/libical/releases/download/v$portVersion/libical-$portVersion.tar.gz"
+CHECKSUM_SHA256="654c11f759c19237be39f6ad401d917e5a05f36f1736385ed958e60cf21456da"
+SOURCE_DIR="libical-$portVersion"
```
pulkomandy: This is the default SOURCE_DIR pattern, so this line can be removed

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	libical$secondaryArchSuffix = $portVersion
+	lib:libical$secondaryArchSuffix
```
pulkomandy: Check if the library has a soname (there is a file called libical.so.1.2.3 and libical.so.1, for example) and add the matching "= 1.2.3 compat >= 1" to the libraries.

dev-libs/libical/libical-2.1.0~.recipe
```diff
@@ -0,0 +1,83 @@
+SUMMARY="An implementation of iCalendar protocols and data formats"
```
pulkomandy: Is the ~ at the end of the package name intentional? It looks like a strange version pattern.

dev-libs/libical/libical-2.1.0~.recipe
```diff
@@ -0,0 +1,83 @@
...
+scrGitRev="1509e29de73044856271b18f1024acdee3d6f8ac"
+SOURCE_URI="https://github.com/libical/libical/archive/$scrGitRev.tar.gz"
+CHECKSUM_SHA256="3d4b02ff991916fee19badf6c0117627b6643fbdc1558e216e812108ab40fd9e"
+SOURCE_DIR="libical-$scrGitRev"
+
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 ?x86"
+
+PROVIDES="
+	libical$secondaryArchSuffix = $portVersion
```
pulkomandy: Are this and the libical 2 compatible at the ABI level?
If they are not, you need to have different package names (libical and libical2) and proper sonames/compat declarations (see my comments on the other recipe). Otherwise, the two packages will conflict and it will not be possible to install them side by side.
You can look at the libpng recipes for an example of this.

dev-libs/libical/libical-2.1.0~.recipe
raefaldhia: Oppss it should be libical-2.1.0~git missing the git

fbrosson: Yep, it is intentional, but @raefaldhia just forgot to rename the file to **`libical-2.1.0~git.recipe`** (which is, if I'm not mistaken, the convention to use when we know that 2.1.0 has not been released yet, and that the current git commit will be part of that release when it becomes available.)

dev-libs/libical/libical-2.0.0.recipe
fbrosson: @raefaldhia, the line @pulkomandy is implicitly requesting the removal is the one which defines srcGitRev, of course. -Sorry if this was obvious.)

dev-libs/libical/libical-2.0.0.recipe
fbrosson: According to Gentoo and Arch Linux the sonames for all these libs in release 2.0.0 are:
```
libical.so.2.0.0
libical_cxx.so.2.0.0
libicalss.so.2.0.0
libicalss_cxx.so.2.0.0
libicalvcal.so.2.0.0
```
I guess we could therefore use "= $portVersion compat >= 2" for all the **`lib:libical*`** lines in PROVIDES and **`devel:libical*`** lines in BUILD_REQUIRES. Would that be OK to use that? My understanding is that we should avoid this in the general case, but since upstream is using these sonames in sync with the version string, it would be cool to use $portVersion here.

BTW, maybe we should drop **`libical-2.1.0~git.recipe`** and only keep **`libical-2.0.0.recipe`**.

raefaldhia:
Updated.

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,81 @@
+SUMMARY="An implementation of iCalendar protocols and data formats"
+DESCRIPTION="Libical is an Open Source implementation of the iCalendar \
+protocols and protocol data units. The iCalendar specification describes how \
+calendar clients can communicate with calendar servers so users can store \
+their calendar data and arrange meetings with other users."
+HOMEPAGE="http://libical.github.io/libical/"
+COPYRIGHT="Eric Busboom"
```
fbrosson: It would be nice to add the copyright years, using this format:
```
COPYRIGHT="20??-201? Eric Busboom"
```

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,81 @@
+SUMMARY="An implementation of iCalendar protocols and data formats"
+DESCRIPTION="Libical is an Open Source implementation of the iCalendar \
+protocols and protocol data units. The iCalendar specification describes how \
+calendar clients can communicate with calendar servers so users can store \
+their calendar data and arrange meetings with other users."
+HOMEPAGE="http://libical.github.io/libical/"
+COPYRIGHT="1999-201? Eric Busboom"
```
fbrosson: Please replace "201?" by "2015" (since 2.0.0 was released last year.)

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+	prepareInstalledDevelLibs \
+		libical \
+		libical_cxx \
+		libicalss \
+		libicalss_cxx \
+		libicalvcal
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
```
fbrosson: Please replace
```
	packageEntries devel \
		$developDir
```
by
```
	packageEntries devel \
		$developDir \
		$libDir/cmake
```

raefaldhia:
@fbrosson 
> I also wish to figure out if we really need cmd:perl in BUILD_PREREQUIRES.

https://github.com/libical/libical/blob/master/Install.txt#L15

fbrosson:
OK, thanks for the info about perl being required. (I had skipped that!!!)

Regarding icu, this is what I did to find out which libs it can use:
1) I added this to BUILD_REQUIRES:
```
	devel:libicudata$secondaryArchSuffix
	devel:libicui18n$secondaryArchSuffix
	devel:libicuio$secondaryArchSuffix
	devel:libicule$secondaryArchSuffix
	devel:libiculx$secondaryArchSuffix
	devel:libicutest$secondaryArchSuffix
	devel:libicutu$secondaryArchSuffix
	devel:libicuuc$secondaryArchSuffix
```
2) I called "hp -S libicall_x86," which told me that libicui18n.so and libicuuc.so were required.
3) I then removed all the **`devel:libicu*`** lines I had previously added to BUILD_REQUIRES and only kept these 2::
```
	devel:libicui18n$secondaryArchSuffix
	devel:libicuuc$secondaryArchSuffix
```
4) I added this to REQUIRES:
```
	lib:libicui18n$secondaryArchSuffix
	lib:libicuuc$secondaryArchSuffix
```
This builds fine for x86 secondary arch on x86_gcc2 and should also build fine on x86 primary arch, of course.
(I need to check it on x86_64.)

Regarding the tests, I found an error message about a missing "bswap_32" symbol but did not google it yet.

fbrosson:
I forgot to mention that I also added **`cmd:icu_config$secondaryArchSuffix`** to BUILD_PREREQUIRES.
BTW, the tests still fail, even with libicu :-(

raefaldhia:
Here's refference linux build, some test actually success without libicu
https://gist.github.com/raefaldhia/fe638e77442df7f19d4b58badceff4a2

raefaldhia:
Maybe you're right, it caused by bswap, when i try to run the regression manually i got this result:
```sh
~/haiku/haikuports/dev-libs/libical/work-x86-2.0.0/sources/libical-2.0.0/src/test> ./regression
runtime_loader: /boot/system/lib/x86/libical.so.2.0.0: Could not resolve symbol 'bswap_32'
resolve symbol "bswap_32" returned: -2147478780
runtime_loader: /boot/system/lib/x86/libical.so.2.0.0: Troubles relocating: Symbol not found
~/haiku/haikuports/dev-libs/libical/work-x86-2.0.0/sources/libical-2.0.0/src/test> ./timezones
runtime_loader: /boot/system/lib/x86/libical.so.2.0.0: Could not resolve symbol 'bswap_32'
resolve symbol "bswap_32" returned: -2147478780
runtime_loader: /boot/system/lib/x86/libical.so.2.0.0: Troubles relocating: Symbol not found
~/haiku/haikuports/dev-libs/libical/work-x86-2.0.0/sources/libical-2.0.0/src/test>   
```

raefaldhia:
I found interesting stuff, replacing bswap_32 macro inside icaltz-util.c make some test passed but 
2 test, 2 test failed, 3 test passed.

I replaced the macro with: 
http://svn.openmoko.org/trunk/src/host/qemu-neo1973/bswap.h

fbrosson:
**`man bswap`** tells bswap_32 is in **`byteswap.h`**, so I looked for it in:/system/develop/headers and found it is shipped in.. calc_devel, produced by one of the recipes you wrote! Isn't this amazing?

Well, it defines several macros, but with different names. Moreover, it would not be appropriate to make libical build-depend on calc_devel.
So I guess your macro is the best option.
Can you include it in your PR, as well as the ICU stuff?

raefaldhia:
I couldn't understand why the 4 tests still fail, when i checked it's still aboult bswap_32.

raefaldhia:
Haiku should have byteswap.h ... http://cgit.haiku-os.org/haiku/plain/src/system/libroot/posix/glibc/string/byteswap.h

raefaldhia:
My lastest commit should work. do you still faced something about bswap_32 ?

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,86 @@
...
+	libical$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libical$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libical_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libicalss$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libicalss_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	lib:libicalvcal$secondaryArchSuffix = $portVersion compat >= 2
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	icu$secondaryArchSuffix
```
fbrosson: I think you should replace:
```
	icu$secondaryArchSuffix
```
by
```
	lib:libicui18n$secondaryArchSuffix
	lib:libicuuc$secondaryArchSuffix

```
But maybe we should wait for a confirmation from another contributor.

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,86 @@
...
+	devel:libicalss_cxx$secondaryArchSuffix = $portVersion compat >= 2
+	devel:libicalvcal$secondaryArchSuffix = $portVersion compat >= 2
+	"
+REQUIRES_devel="
+	libical$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	icu${secondaryArchSuffix}_devel
```
fbrosson: I think you should replace:
```
	icu${secondaryArchSuffix}_devel
```
by
```
	devel:libicui18n$secondaryArchSuffix
	devel:libicuuc$secondaryArchSuffix

```
But maybe we should wait for a confirmation from another contributor.

dev-libs/libical/libical-2.0.0.recipe
```diff
@@ -0,0 +1,86 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:perl
+	cmd:wget
```
fbrosson: I think we also need **`cmd:icu_config$secondaryArchSuffix`** in BUILD_PREREQUIRES.

BTW, do we really need **`cmd:wget`** ?
Ah, maybe it is required for the test...

raefaldhia: @fbrosson oopps forget to remove it. i used that before, but now the recipe doesn't need it.

fbrosson: Cooll.
BTW, congratulations for having solved the bswap_32 issue!

fbrosson:
Thanks for the update. I'll test it right now.

BTW, on Arch Linux they also pass "-DCMAKE_BUILD_TYPE=Release" to cmake.
Maybe we could do this too? ;-)

fbrosson:
The recipe looks OK, although 4 tests are failing on my system (like on yours):
```
The following tests FAILED:
	  1 - regression (Failed)
	  2 - recur (Failed)
	  4 - testvcal (OTHER_FAULT)
	  6 - builtin_timezones (Failed)
Errors while running CTest
```
(testvcal even causes a crash.)

I think we can merge this as is, and if anyone ever notices that libical is not working then we can always either try to fix it or prepend a "?" to each arch in this recipe.
(So let's just forget about the "-DCMAKE_BUILD_TYPE=Release" at the moment ;-)

raefaldhia:
Crash caused by exception maybe,`testvcal` give exception on linux, `recur` fail on linux. should i publish task in my gci task-instances?
actually https://github.com/haikuports/haikuports/pull/933/files#diff-2f6381c793bd9ea937a061556fbe38d7R63 affect `regression`(without it, regression crash) and `builtin_timezone`

fbrosson:
If you would like to dig the issue about the tests, that's OK for me, but I guess you should probably take some rest and look for other recipes to write, while the contest is open ;-)

So if you publish this task in your gci task-instances then I will validate it!

I'll wait for @pulkomandy's approval before I merge, but I guess there won't be any changes.

pulkomandy:
If the library is actually named libical.so-2.0.0 with a soname named libical.so-2, then yes, it is fine. Otherwise, you will need to remove the portversion/compat for all the lib: entries in provides. Can't check myself, I'm at work with no access to Haiku machine.

raefaldhia:
So when library named libical.so.2.0.0(with dot) i should remove the portversion/compat?

pulkomandy:
Oops, you're right, it is a dot. So if you have libical.so.2.0.0 and libical.so.2, you can keep it the way it is now.

fbrosson:
@pulkomandy: we get nice filenames:
```
package list -p /system/packages/libical_x86-2.0.0-1-x86_gcc2.hpkg  | grep ^lib/
lib/x86
lib/x86/libical.so.2
lib/x86/libical.so.2.0.0
lib/x86/libical_cxx.so.2
lib/x86/libical_cxx.so.2.0.0
lib/x86/libicalss.so.2
lib/x86/libicalss.so.2.0.0
lib/x86/libicalss_cxx.so.2
lib/x86/libicalss_cxx.so.2.0.0
lib/x86/libicalvcal.so.2
lib/x86/libicalvcal.so.2.0.0
```

@raefaldhia: the current declarations seem OK to me.

PR #938, opened 2016-12-20
(GCI) Added recipe for app-crypt/stegsnow

hermord:
Added a recipe for SNOW (Steganographic Nature Of Whitespace), a steganography tool for hiding information in plaintext files using trailing whitespace.

### DIFF
--- a/app-crypt/stegsnow/stegsnow-20130616.recipe
+++ b/app-crypt/stegsnow/stegsnow-20130616.recipe
@@ -1,0 +1,47 @@
+SUMMARY="Hides information in text files using trailing whitespace"
+DESCRIPTION="SNOW (Steganographic Nature Of Whitespace) is a steganography \
+tool which conceals messages in ASCII text by appending spaces to the end of \
+lines. Optional encryption is supported to further secure the message."
+HOMEPAGE="http://www.darkside.com.au/snow/"
+COPYRIGHT="1998-2013 Matthew Kwan"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="http://www.darkside.com.au/snow/snow-$portVersion.tar.gz"
+CHECKSUM_SHA256="c0b71aa74ed628d121f81b1cd4ae07c2842c41cfbdf639b50291fc527c213865"
+SOURCE_DIR="snow-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	stegsnow = $portVersion
+	cmd:stegsnow = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	"
+
+PATCH()
+{
+	sed -i 's/snow/steg&/g; s/SNOW/STEG&/g;' snow.1
+}
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	# Upstream Makefile has no install target
+	install -d -m 755 $binDir $manDir/man1
+	install -m 555 -s -T snow $binDir/stegsnow
+	install -m 444 -T snow.1 $manDir/man1/stegsnow.1
+}


app-crypt/snow/snow-20130616.recipe
```diff
@@ -0,0 +1,43 @@
+SUMMARY="Hides information in text files using trailing whitespace"
+DESCRIPTION="SNOW (Steganographic Nature Of Whitespace) is a steganography \
+tool which conceals messages in ASCII text by appending spaces to the end of \
+lines. Has an option to encrypt the message first to further secure it."
+HOMEPAGE="http://www.darkside.com.au/snow/"
+COPYRIGHT="Matthew Kwan"
```
fbrosson: We like to include the copyright years, if available. That would be something in the following format:
```
COPYRIGHT="YYYY-20YY Matthew Kwan"
```

app-crypt/snow/snow-20130616.recipe
```diff
@@ -0,0 +1,43 @@
...
+	"
+
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES=""
+
+BUILD_PREREQUIRES="
+	haiku_devel
```
fbrosson: There has been a small change some time ago. We now prefer to have haiku_devel in BUILD_REQUIRES instead of BUILD_PREREQUIRES:
```
BUILD_REQUIRES="
	haiku_devel
	"
BUILD_PREREQUIRES="
	cmd:gcc
	cmd:make
	"
```

app-crypt/snow/snow-20130616.recipe
```diff
@@ -0,0 +1,43 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	# Upstream Makefile has no install target
+	mkdir -p $binDir
+	install -t $binDir snow
+	mkdir -p $manDir/man1
+	install -t $manDir/man1 snow.1
```
fbrosson: I would recommend replacing:
```
	mkdir -p $binDir
	install -t $binDir snow
	mkdir -p $manDir/man1
	install -t $manDir/man1 snow.1
```
by
```
	install -d -m 755 $binDir $manDir/man1
	install -t $binDir -c -m 555 -s snow
	install -t $manDir/man1 -c -m 444 snow.1
```

hermord:
Should I also add a TEST()? That would be very easy to do for this package.

app-crypt/snow/snow-20130616.recipe
```diff
@@ -0,0 +1,43 @@
+SUMMARY="Hides information in text files using trailing whitespace"
+DESCRIPTION="SNOW (Steganographic Nature Of Whitespace) is a steganography \
+tool which conceals messages in ASCII text by appending spaces to the end of \
+lines. Has an option to encrypt the message first to further secure it."
+HOMEPAGE="http://www.darkside.com.au/snow/"
+COPYRIGHT="Matthew Kwan"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="http://www.darkside.com.au/snow/$portVersionedName.tar.gz"
```
fbrosson: Although
```
SOURCE_URI="http://www.darkside.com.au/snow/$portVersionedName.tar.gz"
```
is perfectly OK, we prefer to avoid $portVersionedName because when building for a secondary architecture it contains a _x86 os _x86_gcc2 suffix.
So even if this recipe does not have support for secondary architectures I would still recommend to use:
```
SOURCE_URI="http://www.darkside.com.au/snow/snow-$portVersion.tar.gz"
```

app-crypt/snow/snow-20130616.recipe
```diff
@@ -0,0 +1,43 @@
...
+CHECKSUM_SHA256="c0b71aa74ed628d121f81b1cd4ae07c2842c41cfbdf639b50291fc527c213865"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	snow = $portVersion
+	cmd:snow = $portVersion
+	"
+
```
fbrosson: Please remove the empty line between PROVIDES and REQUIRES,
(See our [HaikuPorter Guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering))

hermord:
Alright, thanks a lot for the feedback! I'll push the fixed version out in a minute.

fbrosson:
> Should I also add a TEST()? That would be very easy to do for this package.

It depends. If the source tarball already includes the files required for the test, then yes, it would be a very good idea.
Otherwise, it depends. I'm not against adding a TEST function, but you would need to convince a few other contributors ;-)

> Alright, thanks a lot for the feedback! I'll push the fixed version out in a minute.

You are very welcome! We are happy to see new contributors!

hermord:
OK, fixed. Yeah, on second thought, the TEST() I started to write looks a bit messy. I guess it'd be best not to include it :smile:

fbrosson:
Thanks for your update. I haven't tested it yet but it looks OK.

Hmm, Debian installs this package as stegsnow (instead of snow) because snow is too short for such a special program.

I think we should also rename it, even if this requires a few extra steps.
Debian users probably don't even know that the original name is just snow, so if they look for stegsnow on Haiku they won't find it... unless we also rename it ;-)
(And Gentoo does not have at all.)
What do you think?
Any thoughts from other contributors?

hermord:
Well, actually, the GCI task had it named `stegsnow`. The author refers to it as just `SNOW`, however.

hermord:
So I went the author's way. If `stegsnow` is a better option, I'll change the name to that.

hermord:
Does this look better?

hermord:
Oh, sorry. Forgot to change the name in PROVIDES

hermord:
There, fixed it.

fbrosson:
Thanks for the rename.
Well, we need to also rename the command and the man page.
Debian has a patch for replaceing each occurence of snwo by stegsnow in the man page.
I'll see if we can use it with a minimum overhead.

BTW, I found why Debian has chose to rename the package to stegsnow:
```
# Rename due to existing binary "snow" in package r-cran-snow
```

hermord:
Wouldn't a simple `sed -i 's/snow/stegsnow/g' snow.1` work?

fbrosson:
Yep, that's what I also realized. We probably don't need the Debian patch.
I'm trying this:
```
PATCH()
{
	sed -i -e "s/snow/steg&/g; s/SNOW/STEG&/g;" snow.1
}
```
with this:
```
INSTALL()
{
	# Upstream Makefile has no install target
	install -d -m 755 $binDir $manDir/man1
	install -C -m 555 -T snow $binDir/stegsnow
	install -C -m 444 -T snow.1 $manDir/man1/stegsnow.1
}
```
Then I'll compare the generated stegsnow.1 with what I have on my Debian system.

hermord:
Just a quick question, since I'm not that familiar with the syntax of `install`: why are you using the `-C` flag here? The destination never exists anyway, right?

fbrosson:
You are perfectly right. We don't need the -C.
(So sorry for my previous suggestion and thanks for having spotted it.)

hermord:
Alright, this seems to work for me

fbrosson:
Same for me ;-)
(I've already built it on x86_64)

app-crypt/stegsnow/stegsnow-20130616.recipe
```diff
@@ -0,0 +1,47 @@
+SUMMARY="Hides information in text files using trailing whitespace"
+DESCRIPTION="SNOW (Steganographic Nature Of Whitespace) is a steganography \
+tool which conceals messages in ASCII text by appending spaces to the end of \
+lines. Has an option to encrypt the message first to further secure it."
```
fbrosson: Shall we replace "Has an option to ..." by "It has an option to..." ?
(I'm just inserting "It" before "Has")

hermord: That's a possibility, yeah. Dropping the subject is a bit colloquial, I guess; it is often found in informal documentation, though. If you think the second version is more appropriate, I'll change it right away.

app-crypt/stegsnow/stegsnow-20130616.recipe
```diff
@@ -0,0 +1,47 @@
...
+HOMEPAGE="http://www.darkside.com.au/snow/"
+COPYRIGHT="1999-2013 Matthew Kwan"
```
fbrosson: The software already existed in 1998.
(Can you update COPYRIGHT to fix that? Thanks!)

BTW, CHECKSUM_SHA256 should come before SOURCE_DIR.

(See [HaikuPorter Guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering))

hermord: Hm, I can't seem to find a definite creation date anywhere on the project's website. Should I just put 1998 as the creation date?

app-crypt/stegsnow/stegsnow-20130616.recipe
hermord: "There is an option" could work as well.

app-crypt/stegsnow/stegsnow-20130616.recipe
fbrosson: Yep, 1998 will be fine ;-)

hermord:
I changed the wording of DESCRIPTION a bit to avoid the awkwardness in the last sentence.

fbrosson:
Great jiob! Thanks @hermord for your contribution!

hermord:
Thank you for your guidance!

PR #939, opened 2016-12-21
(GCI) Added recipe for app-text/odt2txt

hermord:
Added a recipe for odt2txt, a converter from OpenDocument Text to plain text.

### DIFF
--- a/app-text/odt2txt/odt2txt-0.5.recipe
+++ b/app-text/odt2txt/odt2txt-0.5.recipe
@@ -1,0 +1,49 @@
+SUMMARY="A simple converter from OpenDocument Text to plain text"
+DESCRIPTION="odt2txt is a command-line tool which extracts the text out of \
+OpenDocument Texts produced by LibreOffice, OpenOffice, StarOfficem, KOffice \
+and others."
+HOMEPAGE="https://github.com/dstosberg/odt2txt"
+COPYRIGHT="2002-2016 Dennis Stosberg"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/dstosberg/odt2txt/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="23a889109ca9087a719c638758f14cc3b867a5dcf30a6c90bf6a0985073556dd"
+PATCHES="odt2txt-$portVersion.patch"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	odt2txt = $portVersion
+	cmd:odt2txt = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libiconv
+	lib:libz
+	lib:libzip
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libiconv
+	devel:libz
+	devel:libzip
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:groff
+	cmd:make
+	cmd:pkg_config
+	"
+
+BUILD()
+{
+	make HAVE_LIBZIP=1 $jobArgs
+}
+
+INSTALL()
+{
+	# Upstream Makefile has DESTDIR set to /usr/local mistakenly,
+	# hence the need to unset it.
+	make install HAVE_LIBZIP=1 DESTDIR= PREFIX=$prefix MANDIR=$manDir
+}
--- a/app-text/odt2txt/patches/odt2txt-0.5.patch
+++ b/app-text/odt2txt/patches/odt2txt-0.5.patch
@@ -1,0 +1,14 @@
+--- a/Makefile
++++ b/Makefile
+@@ -54,6 +54,11 @@
+ ifeq ($(UNAME_S),NetBSD)
+ 	CFLAGS += -DICONV_CHAR="const char"
+ endif
++ifeq ($(UNAME_S),Haiku)
++	# libzip installs headers to a non-standard dir on Haiku
++	CFLAGS += $(shell pkg-config --cflags libzip)
++	LIBS += -liconv
++endif
+ ifeq ($(UNAME_S),SunOS)
+ 	ifeq ($(CC),cc)
+ 		ifdef DEBUG


app-text/odt2txt/odt2txt-0.5.recipe
```diff
@@ -0,0 +1,47 @@
...
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	# Upstream Makefile uses $(DESTDIR)$(BINDIR) and $(DESTDIR)$(MAN1DIR),
+	# hence the need to have a trailing slash in DESTDIR.
+	make install DESTDIR="$prefix/" BINDIR="$relativeBinDir" \
+	             MANDIR="$relativeManDir"
```
fbrosson: It looks like this _also_ works:
```
	make install DESTDIR= PREFIX="$prefix" MANDIR="$manDir"
```

BTW, I'm trying to make odt2txt also use libzip by adding this to make in BUILD():
```
CFLAGS=-I/system/$relativeOldIncludeDir HAVE_LIBZIP=1
```
but I'm not sure it is linking to libzip.

waddlesplash: The existing method looks better to me.

fbrosson: The error comes from the Makefile. It defines:
```
DESTDIR = /usr/local
PREFIX  =
```
instead of:
```
DESTDIR = 
PREFIX  = /usr/local
```
which would be much better.

BTW, using
```
	make install DESTDIR= PREFIX=$prefix MANDIR=$manDir
```
allows to avoid having to specify BINDIR and does not need any trailing slash either ;-)

hermord: Would it be better to patch the Makefile, then? That way, we'd keep the recipe itself as semantically clear as possible: `make install PREFIX=$prefix MANDIR=$manDir`

fbrosson: Please change INSTALL to:
```
INSTALL()
{
	make install DESTDIR= PREFIX=$prefix MANDIR=$manDir
}
```

fbrosson: No need to add a patch to unset DESTDIR. It can be unset when make install is called.

hermord:
OK, done. I'm trying to build using `libzip` as well.

hermord:
I'll post back when it's working.

fbrosson:
OK, thanks.
BTW, you can relax; there is no urgency ;-)

fbrosson:
I've approved the change in INSTALL, but I'm still very much interested in the libzip stuff.
Of course, you have all the time you wish to work on the libzip support.
Thanks!

hermord:
OK, so the deal with `libzip` is actually quite complicated. Simply passing `HAVE_LIBZIP=1` to `make` doesn't work: there is a missing header, `zipconf.h`, which `zip.h` requires, and `zip.h` is included from `odt2txt.c`. That missing header is actually located in `/system/lib/libzip/include` (!), which obviously shouldn't be the case. It is installed there as part of `libzip` (not `libzip_devel`!) up to version 1.1.3, at which this bug was fixed and the header from then on is supposed to be installed at `/system/develop/include/libzip`. Version 1.1.3 is not released as a binary backage yet, AFAIK, but even if it were, `/system/develop/headers/libzip` is _still_ not where `zip.h` expects `zipconf.h` to be: the include directive in `zip.h` has to be changed from `#include<zipconf.h>` to `#include<libzip/zipconf.h>` for things to work. You would think that all of this could be circumvented temporarily by passing an extra include directory in `CFLAGS` to `make`, but, the thing is, there is no way to do that cleanly either: if we pass it directly: `make CFLAGS=-I/system/lib/libzip/include HAVE_LIBZIP=1 $jobArgs`, then `CFLAGS` won't ever get set inside `Makefile`, so things will break because `-DHAVE_LIBZIP` won't get appended to them. If we just `export CFLAGS=-I/system/lib/libzip/include`, on the other hand, then `CFLAGS` will get overriden before it comes to compilation. and the extra include path won't be honored. All around, the only good solution would be to fix things on the `libzip` side: `zip.h` and `zipconf.h` have to be brought to agreement, one way to do so being replacing `#include<zipconf.h>` with `#include<libzip/zipconf.h>`.

hermord:
Either way, I'd say this warrants a separate PR.

hermord:
Another alterbative is moving `zipconf.h` one level up the hierarchy to `/system/develop/headers/`, which is easier in that we'd only have to change the `ziplib` recipe just a bit, as opposed to patching `zip.h`. But it also might violate some guidelines, as we'd be putting things that clesrly belong in a subdirectory in the root just because it's easier. I don't really know what's better.

fbrosson:
Thank you very much indeed for having digged the topic that far.

As I [wrote](##discussion_r93375775)  in the first comment of my first review, passing this:
```
	CFLAGS=-I/system/$relativeOldIncludeDir HAVE_LIBZIP=1
```
to make in BUILD allows to build the two sources that use libzip, but I did not investigate why the resulting runtime was not linking against libzip.

Well, this libzip part is really hard.
So I guess we should leave it for a separate task and PR, as you suggested.

I also tried to use libiconv without the patch but did not go very far yet.
So if you would like to submit your recipe as is, then I'll validate the task.

Wait, could you please do a couple of cosmetic changes?
It's about the alphabetical sorting in REQUIRES and BUILD_REQUIRES: it would be nice to have libiconv appear before libz.

But there is no urgency. All this can wait. We both need to rest ;-)

pulkomandy:
Doesn't ziplib come with a pkg-config file? If so, pkg-config --cflags ziplib should give the proper include paths to use. The output of this command is usually added to GCC command line. If that is the case, then we should patch the makefile here to try to use that command when available.

hermord:
Well, the problem, I guess, is that we would have to do that for every single package that uses `zip.h`, because it is effectively broken. I don't know about other systems, but on Arch Linux `zipconf.h` and `zip.h` are both installed in `/usr/include`, negating the need to include extra paths. I suspect that it's like that on most Linuxes and probably on most *NIXes, so we would patch all the software we port to Haiku to call `pkg-config`, because the authors wouldn't need to call it on other systems.

hermord:
And if we do that, we might as well move `zip.h` to the `libzip` directory to avoid cluttering  `/system/develop/headers`.

hermord:
But, well, the new recipe works. Notice the change in Makefile.

fbrosson:
Cool, that's good news!
I guess you have built libzip-1.1.3 from source.

BTW, thiese are the files the libzip_devel package has in /system/develop/headers/:
```
package list -p /system/packages/libzip_devel-1.1.3-1-x86_gcc2.hpkg  | grep ^develop/headers
develop/headers
develop/headers/libzip
develop/headers/libzip/zipconf.h
develop/headers/zip.h
```

With the attemps I made to integrate libzip, I did not manage to make odt2txt link against libzip:
```
objdump -x /bin/odt2txt |grep NEEDED
  NEEDED      libz.so.1
  NEEDED      libiconv.so.2
  NEEDED      libroot.so
```
I'll build your new recipe.

hermord:
Oh yeah, maybe I should set `libzip >= 1.1.3` as a requirement?

hermord:
Oh, so `zipconf.h` got moved to `libzip_devel`. That's good, maybe `zip.h` should be moved to `/system/develop/headers/libzip` as well?

fbrosson:
> Oh yeah, maybe I should set libzip >= 1.1.3 as a requirement?

When haikuporter starts, it shows all the packages that it thinks it needs to expose in the build environment. The list goes is displayed after this line:
```
chroot has these packages active
```
So if you also get this:
```
	/boot/system/packages/libzip-1.1.3-1-x86_gcc2.hpkg
	/boot/system/packages/libzip_devel-1.1.3-1-x86_gcc2.hpkg
```
then I guess we don't need to force a particular version of libzip.

> Oh, so `zipconf.h` got moved to `libzip_devel`. That's good, maybe `zip.h` should be moved to `/system/develop/headers/libzip` as well?

I think zip.h should remain in develop/headers/ (which is the exact equivalent of /usr/include) because on Debian-based distros the zip.h file is there.

(And the zipconf.h goes is fine in /system/develop/headers/libzip/.)

(I'm still trying to build with libzip, BTW...)

hermord:
I actually tried to build it with an older `libzip` version in chroot, and it worked equally well: `-I/system/lib/libzip/include` gets added to `CFLAGS` as required. So we definitely don't need to constraint the version of `libzip`.

hermord:
> I think zip.h should remain in develop/headers/ (which is the exact equivalent of /usr/include) because on Debian-based distros the zip.h file is there.

Well, I just thought, since we're breaking compatibility with Debian by not having `zipconf.h` in the root of the standard headers directory (`/usr/include <-> /system/develop/headers/`), we might as well move them both to `/system/develop/headers/libzip`, where they logically (IMHO) belong.

hermord:
> (I'm still trying to build with libzip, BTW...)

Is it not working on x86_64? Because it worked for me on x86_gcc2.

fbrosson:
On Debian zipconf.h is in a libzip subdirectory. When I fixed the libzip-1.1.3 package some time ago, I just moved what was being installed to /usr/include into /system/develop/headers, keeping the dubdirectories.
So our libzip-1.1.3 has a layout that is compatible with what Debian has.

> Is it not working on x86_64? Because it worked for me on x86_gcc2.

I have a tripple boot USB stick (x86_gcc2 hybrid + x86 hybrid + x86_64), which I can boot aither on real hardware, or on a VM.
The trick is that I have a symlink to /dev/sdb called haiku.hdd and I give that path when I create the VM.
Of course, I need to chgrp /dev/sdb to a group that I belong too.
This is handy because I can try any architecture, either on real hardware or in a VM.

But now I'm running x86_gcc2 in a VM.

hermord:
> On Debian zipconf.h is in a libzip subdirectory.

Welp, looks like a made a fool out of myself. I assumed, since there is no clause in the Makefile for different flavors of Linux, that it is a convention to store both `zipconf.h` and `zip.h` in `/usr/include`. My distro does that:
```fish
hermod@Festung ~> find /usr/include/ -name 'zipconf.h'
/usr/include/zipconf.h
```
If that is not the case in Debian, then replicating their layout does indeed make sense.

hermord:
Just checked on my other machine, in Fedora `zipconf.h` is installed in `/usr/include` as well.

fbrosson:
Actually, on Debian we have a big zip.h in /usr/include/, a small zipconf.h in a subdir, and a symbolic link to it in another subdir:
```
ls -gor $(dpkg -L libzip-dev | grep include/.*\.h)
-rw-r--r-- 1   992 May 16  2016 /usr/lib/i386-linux-gnu/libzip/include/zipconf.h
-rw-r--r-- 1 20945 May 16  2016 /usr/include/zip.h
lrwxrwxrwx 1    49 May 16  2016 /usr/include/i386-linux-gnu/zipconf.h -> ../../lib/i386-linux-gnu/libzip/include/zipconf.h
```
But Debian does that because they support multiarch, which we don't in Haiku, although we have secondary arch.
So, as you suggested, maybe we should change the libzip in order to move zipconf.h to the same directory where zip.h is.

Shall we forget about the libzip support for now?
We can always come back later for that (or not ;-)

Many thanks for having digged the topic.
(BTW, you can remove the chunks about libzip in the recipe, don't bother commenting them out.)

hermord:
Well, again, seeing as `CFLAGS` are already conditionally appended to in the `Makefile` depending on system (`-I/usr/local/include` for FreeBSD etc.), I guess we could merge things as they are right now. The extra two lines in the patch shouldn't be that big of a problem, and they will allow us to not have to change anything if we do move `zip.h` into the `libzip` subdir. Or is the current recipe not working for you?

hermord:
Because `libzip` support _is_ actually working for me right now with the current recipe.

fbrosson:
It's not building on my side :-(
On your side does libzip appear on the output of the following command?
```
objdump -x /bin/odt2txt | grep NEEDED
```

hermord:
It does

hermord:
Does the packaging stage succeed at all? Because I had a problem with it failing because I forgot to pass `HAVE_LIBZIP=1` to `make install`.

fbrosson:
BTW, I'm trying @pulkomandy's suggestion: Passing this:
```
CFLAGS="`pkg-config --cflags libzip`"
```
to make.
But I get an error that is unrelated to that:
```
uilding ...
cc -o odt2txt  odt2txt.o regex.o mem.o strbuf.o  -lzip -liconv
odt2txt.o: In function `read_from_zip':
odt2txt.c:(.text+0x72a): undefined reference to `kunzip_get_offset_by_name'
odt2txt.c:(.text+0x778): undefined reference to `kunzip_next_tobuf'
strbuf.o: In function `strbuf_append_inflate':
```
I guess I need to check the versions of my libzip packages.
Well, I'll try on x86 and x86_64 before cleaning my x86_gcc2.

hermord:
> CFLAGS="`pkg-config --cflags libzip`"

Oh yeah, that *does not* work. I've tried different variations of that multiple times. If you pass `CFLAGS` to `make`, nothing in `Makefile` will have any effect on them, so `-DHAVE_LIBZIP` will never get appended to them. I outlined that in one of my previous posts. There is actually no way to pass the include dir from outside, you pretty much have to do it inside the `Makefile`, which is what I did.

hermord:
Just try my [latest commit](hermord/haikuports@608d6657169cabad002c99e4d3b1199cad36b215).

fbrosson:
Sorry, I just understood why I was having an error: I had not grabbed your last patch!
Will try right now.

fbrosson:
**_Congratulations! Your libzip support is working!_** :tada: 

I've got a few cosmetic edits to suggest:
1) rename odt2txt-0.5-makefile-fix.patch to odt2txt-0.5.patch
2) use **`PATCHES="odt2txt-$portVersion.patch"`** (instead of **`PATCHES="odt2txt-0.5-makefile-fix.patch"`**
3) Replace the first 2 lines of the patch by:
```diff
--- a/Makefile
+++ b/Makefile
```
in order to reduce the entropy of that patch.

This is really a minor thing, so please feel free to keep your recipe as is if you prefer.

app-text/odt2txt/odt2txt-0.5.recipe
```diff
@@ -0,0 +1,49 @@
+SUMMARY="A simple converter from OpenDocument Text to plain text"
+DESCRIPTION="odt2txt is a command-line tool which extracts the text out of \
+OpenDocument Texts produced by LibreOffice, OpenOffice, StarOfficem, KOffice \
+and others."
+HOMEPAGE="https://github.com/dstosberg/odt2txt/"
```
fbrosson: You can change this to:
```
HOMEPAGE="https://github.com/dstosberg/odt2txt"
```
(I'm just dropping the trailing slash.)

app-text/odt2txt/odt2txt-0.5.recipe
```diff
@@ -0,0 +1,49 @@
...
+COPYRIGHT="2002-2016 Dennis Stosberg"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/dstosberg/odt2txt/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="23a889109ca9087a719c638758f14cc3b867a5dcf30a6c90bf6a0985073556dd"
+PATCHES="odt2txt-0.5-makefile-fix.patch"
```
fbrosson: Shall we change this to:
```
PATCHES="odt2txt-$portVersion.patch"
```
Of course, this assumes you also:
```
cd patches
git mv odt2txt-0.5-makefile-fix.patch odt2txt-0.5.patch
git commit --amend --no-edit odt2txt-0.5-makefile-fix.patch odt2txt-0.5.patch
git push -f
```

app-text/odt2txt/patches/odt2txt-0.5-makefile-fix.patch
```diff
@@ -0,0 +1,14 @@
+--- a/Makefile	2016-12-21 03:29:47.103546880 +0000
++++ b/Makefile	2016-12-21 07:43:15.831258624 +0000
```
fbrosson: Shall we drop the timestamps and just keep this, as the first 2 lines:
```diff
--- a/Makefile
+++ b/Makefile
```

hermord:
Alright, just give me a second ;)

hermord:
Oh, sorry, and the trailing slash

hermord:
✔️ Alright, all done

fbrosson:
Thanks @hermord for your really great work!
You rock!

hermord:
It's my pleasure!

PR #945, opened 2016-12-22
Fish: hide socket warning at start.

miqlas:
Not an ideal solution, but works and doesn't breaks anything.

### DIFF
--- a/app-shells/fish/patches/fish-2.4.0.patchset
+++ b/app-shells/fish/patches/fish-2.4.0.patchset
@@ -107,3 +107,30 @@ index eb5dc5c..097f384 100755
 -- 
 2.10.0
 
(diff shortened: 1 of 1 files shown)


waddlesplash:
The error is completely valid, no? So why should we hide it?

mmuman:
That's odd, we should already support AF_UNIX... maybe only for SOCK_STREAM sockets though?
Ah, indeed, cf. src/add-ons/kernel/network/protocols/unix/unix.cpp

miqlas:
This code is used to check for old fish version:
https://github.com/fish-shell/fish-shell/issues/3317#issuecomment-268859613

So we can freely disable it, as "old" means Fish 1.xx, our Fish 1.xx recipe long gone (i deleted it), and the package was instable, i don't expect any real world user in the wild.

waddlesplash:
Oh, very well, it can be merged I guess.

PR #948, opened 2016-12-23
(GCI) Added recipe for sys-fs/squashfs_tools

hermord:
Added a recipe for squashfs_tools, a suite of tools for creating and manipulating squashfs filesystems.

### DIFF
--- a/sys-fs/squashfs_tools/patches/squashfs_tools-4.3-gcc2.patch
+++ b/sys-fs/squashfs_tools/patches/squashfs_tools-4.3-gcc2.patch
@@ -1,0 +1,50 @@
+--- a/unsquash-2.c
++++ b/unsquash-2.c
+@@ -118,11 +118,11 @@ int read_fragment_table_2(long long *directory_table_end)
+ void read_fragment_2(unsigned int fragment, long long *start_block, int *size)
+ {
+ 	TRACE("read_fragment: reading fragment %d\n", fragment);
+-
++{
+ 	squashfs_fragment_entry_2 *fragment_entry = &fragment_table[fragment];
+ 	*start_block = fragment_entry->start_block;
+ 	*size = fragment_entry->size;
+-}
++}}
+ 
+ 
+ struct inode *read_inode_2(unsigned int start_block, unsigned int offset)
+--- a/unsquash-3.c
++++ b/unsquash-3.c
+@@ -105,11 +105,11 @@ int read_fragment_table_3(long long *directory_table_end)
+ void read_fragment_3(unsigned int fragment, long long *start_block, int *size)
+ {
+ 	TRACE("read_fragment: reading fragment %d\n", fragment);
+-
++{
+ 	squashfs_fragment_entry_3 *fragment_entry = &fragment_table[fragment];
+ 	*start_block = fragment_entry->start_block;
+ 	*size = fragment_entry->size;
+-}
++}}
+ 
+ 
+ struct inode *read_inode_3(unsigned int start_block, unsigned int offset)
+--- a/unsquash-4.c
++++ b/unsquash-4.c
+@@ -85,13 +85,13 @@ int read_fragment_table_4(long long *directory_table_end)
+ void read_fragment_4(unsigned int fragment, long long *start_block, int *size)
+ {
+ 	TRACE("read_fragment: reading fragment %d\n", fragment);
+-
++{
+ 	struct squashfs_fragment_entry *fragment_entry;
+ 
+ 	fragment_entry = &fragment_table[fragment];
+ 	*start_block = fragment_entry->start_block;
+ 	*size = fragment_entry->size;
+-}
++}}
+ 
+ 
+ struct inode *read_inode_4(unsigned int start_block, unsigned int offset)
--- a/sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
+++ b/sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
@@ -1,0 +1,107 @@
+SUMMARY="Provides tools for creating and manipulating squashfs filesystems"
+DESCRIPTION="squashfs-tools provides tools for creating and appending to \
+(mksquashfs) and extracting (unsquashfs) squashfs filesystems.
+
+Squashfs is a compressed read-only filesystem. It is intended for general \
+read-only filesystem use, archival use (i.e. in cases where a .tar.gz file \
+may be used, and in constrained block device/memory systems (e.g. embedded \
+systems) where low overhead is needed."
+HOMEPAGE="http://squashfs.sourceforge.net/"
+COPYRIGHT="2002-2014 Philip Lougher"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sf.net/squashfs/squashfs$portVersion.tar.gz"
+CHECKSUM_SHA256="0d605512437b1eb800b4736791559295ee5f60177e102e4d4ccd0ee241a5f3f6"
+SOURCE_DIR="squashfs$portVersion/squashfs-tools"
+
+debianRevision="3"
+SOURCE_FILENAME_2="squashfs-tools_$portVersion-$debianRevision.debian.tar.xz"
+SOURCE_URI_2="http://httpredir.debian.org/debian/pool/main/s/squashfs-tools/$SOURCE_FILENAME_2"
+CHECKSUM_SHA256_2="1c296cc147e322e7124bf23a3c242ac83f6a986e6d6f64829ad2424b33914fc4"
+
+commit="dceb729f0369d72f1d7820705fd12510b71446d2"
+patchsetBaseUri="https://gitweb.gentoo.org/repo/gentoo.git/plain/sys-fs/squashfs-tools/files"
+SOURCE_FILENAME_3="squashfs-tools-$portVersion-2gb.patch"
+SOURCE_FILENAME_4="squashfs-tools-$portVersion-aligned-data.patch"
+SOURCE_FILENAME_5="squashfs-tools-$portVersion-local-cve-fix.patch"
+SOURCE_FILENAME_6="squashfs-tools-$portVersion-mem-overflow.patch"
+SOURCE_URI_3="$patchsetBaseUri/$SOURCE_FILENAME_3?id=$commit#noarchive"
+SOURCE_URI_4="$patchsetBaseUri/$SOURCE_FILENAME_4?id=$commit#noarchive"
+SOURCE_URI_5="$patchsetBaseUri/$SOURCE_FILENAME_5?id=$commit#noarchive"
+SOURCE_URI_6="$patchsetBaseUri/$SOURCE_FILENAME_6?id=$commit#noarchive"
+CHECKSUM_SHA256_3="cdb9c85c2f733c0db879275d0ee2480f528aa3191f236bbf1ff681c75b14a024"
+CHECKSUM_SHA256_4="3030ddfa5ba9d0baea7f5e4761b33dea67fbc4d49c86e475830fc1bfbfddb145"
+CHECKSUM_SHA256_5="7edceedead4f5a57dc009afecd0f414d60a3878dc43f8de5224bff391ed32c76"
+CHECKSUM_SHA256_6="a94f9df3439bef373da31a734e6dad42f5fc5a6d91f2b9d15d30b02f1aac73b3"
+
+PATCHES="
+	squashfs_tools-$portVersion-haiku-compat.patch
+	squashfs_tools-$portVersion-gcc2.patch
+	"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	HAVE_XZ=1
+else
+	HAVE_XZ=
+fi
+
+PROVIDES="
+	squashfs_tools$secondaryArchSuffix = $portVersion
+	cmd:mksquashfs$secondaryArchSuffix = $portVersion
+	cmd:unsquashfs$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:liblz4$secondaryArchSuffix
+	${HAVE_XZ:+lib:liblzma$secondaryArchSuffix}
+	lib:liblzo2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:liblz4$secondaryArchSuffix
+	${HAVE_XZ:+devel:liblzma$secondaryArchSuffix}
+	devel:liblzo2$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+MAKE_ARGS="
+	LZ4_SUPPORT=1
+	LZMA_XZ_SUPPORT=${HAVE_XZ:+1}
+	LZO_SUPPORT=1
+	XATTR_SUPPORT=0
+	XZ_SUPPORT=${HAVE_XZ:+1}
+	"
+
+PATCH()
+{
+	patch -p2 -i $sourceDir2/debian/patches/0002-*.patch
+	patch -p2 -i $sourceDir2/debian/patches/0003-*.patch
+	patch -p2 -i $sourceDir2/debian/patches/0004-*.patch
+	patch -p2 -i $sourceDir3/$SOURCE_FILENAME_3
+	patch -p2 -i $sourceDir4/$SOURCE_FILENAME_4
+	patch -p2 -i $sourceDir6/$SOURCE_FILENAME_6
+	# This one patch requires -p1
+	patch -p1 -i $sourceDir5/$SOURCE_FILENAME_5
+}
+
+BUILD()
+{
+	make $MAKE_ARGS $jobArgs
+}
+
+INSTALL()
+{
+	make install $MAKE_ARGS INSTALL_DIR=$binDir
+
+	install -d -m 755 $manDir/man1
+	install -m 444 -t $manDir/man1 $sourceDir2/debian/manpages/*.1
+}
--- a/sys-fs/squashfs_tools/patches/squashfs_tools-4.3-haiku-compat.patch
+++ b/sys-fs/squashfs_tools/patches/squashfs_tools-4.3-haiku-compat.patch
@@ -1,0 +1,83 @@
+--- a/action.c
++++ b/action.c
+@@ -39,6 +39,7 @@
+ #include <limits.h>
+ #include <errno.h>
+ 
++#include "haiku_macros.h"
+ #include "squashfs_fs.h"
+ #include "mksquashfs.h"
+ #include "action.h"
+@@ -1953,9 +1957,12 @@ static char *get_start(char *s, int n)
+ 
+ static int subpathname_fn(struct atom *atom, struct action_data *action_data)
+ {
+-	return fnmatch(atom->argv[0], get_start(strdupa(action_data->subpath),
+-		count_components(atom->argv[0])),
+-		FNM_PATHNAME|FNM_PERIOD|FNM_EXTMATCH) == 0;
++	char *dup = strdup(action_data->subpath);
++	int ret = fnmatch(atom->argv[0], get_start(dup,
++			count_components(atom->argv[0])),
++			FNM_PATHNAME|FNM_PERIOD|FNM_EXTMATCH) == 0;
++	free(dup);
++	return ret;
+ }
+ 
+ TEST_VAR_FN(filesize, ACTION_REG, action_data->buf->st_size)
+--- /dev/null
++++ b/haiku_macros.h
+@@ -0,0 +1,14 @@
++#ifndef HAIKU_MACROS_H
++#define HAIKU_MACROS_H
++
++/* Haiku does not provide these */
++#define major(x) ((int)(0x00ff & ((x) >> 8)))
++#define minor(x) ((int)(0xffff00ff & (x)))
++#define makedev(maj,min) ((0xff00 & ((maj)<<8))|(0xffff00ff & (min)))
++
++/* Mirroring a patch accepted to upstream after the release */
++#ifndef FNM_EXTMATCH
++#define FNM_EXTMATCH 0
++#endif
++
++#endif /* HAIKU_MACROS_H */
+--- a/mksquashfs.c
++++ b/mksquashfs.c
+@@ -55,7 +55,7 @@
+ #define __BYTE_ORDER BYTE_ORDER
+ #define __BIG_ENDIAN BIG_ENDIAN
+ #define __LITTLE_ENDIAN LITTLE_ENDIAN
+-#include <sys/sysctl.h>
++#include "haiku_macros.h"
+ #else
+ #include <endian.h>
+ #include <sys/sysinfo.h>
+@@ -4105,7 +4109,7 @@ void initialise_threads(int readq, int fragq, int bwriteq, int fwriteq,
+ 		BAD_ERROR("Failed to set signal mask in intialise_threads\n");
+ 
+ 	if(processors == -1) {
+-#ifndef linux
++#if !defined(linux) && !defined(__HAIKU__)
+ 		int mib[2];
+ 		size_t len = sizeof(processors);
+
+--- a/unsquashfs.c
++++ b/unsquashfs.c
+@@ -31,7 +31,7 @@
+ #include "unsquashfs_info.h"
+ #include "stdarg.h"
++#include "haiku_macros.h"
+ 
+-#include <sys/sysinfo.h>
+ #include <sys/types.h>
+ #include <sys/time.h>
+ #include <sys/resource.h>
+@@ -2190,7 +2193,7 @@ void initialise_threads(int fragment_buffer_size, int data_buffer_size)
+ 			"\n");
+ 
+ 	if(processors == -1) {
+-#ifndef linux
++#if !defined(linux) && !defined(__HAIKU__)
+ 		int mib[2];
+ 		size_t len = sizeof(processors);
+ 


sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,78 @@
...
+	squashfs_tools-$portVersion-haiku-compat.patch
+	squashfs_tools-$portVersion-fix-mem-calculation.patch
+	squashfs_tools-$portVersion-local-cve-fix.patch
+	squashfs_tools-$portVersion-lzma-magic.patch
+	squashfs_tools-$portVersion-mem-overflow.patch
+	squashfs_tools-$portVersion-xattrs.patch
+	"
+
+DEBIAN_REV="3"
+SOURCE_URI_2="http://http.debian.net/debian/pool/main/s/squashfs-tools/squashfs-tools_$portVersion-$DEBIAN_REV.debian.tar.xz"
```
fbrosson: It would be slightly better to use the canonical name:(httpredir.debian.org instead of http.debian.net):
```
SOURCE_URI_2="http://httpredir.debian.org/debian/pool/main/s/squashfs-tools/squashfs-tools_$portVersion-$DEBIAN_REV.debian.tar.xz"
```

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,78 @@
...
+
+Squashfs is a compressed read-only filesystem. It is intended for general \
+read-only filesystem use, archival use (i.e. in cases where a .tar.gz file \
+may be used, and in constrained block device/memory systems (e.g. embedded \
+systems) where low overhead is needed."
+HOMEPAGE="http://squashfs.sourceforge.net"
+COPYRIGHT="2002-2014 Philip Lougher"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://prdownloads.sf.net/squashfs/squashfs$portVersion.tar.gz"
```
fbrosson: I think we should avoid the old "prdownloads.sf.net" name and, instead, use "downloads.sf.net" or "downloads.sourceforge.net":
```
SOURCE_URI="https://downloads.sf.net/squashfs/squashfs$portVersion.tar.gz"
```

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,78 @@
+SUMMARY="Provides tools for creating and manipulating squashfs filesystems"
+DESCRIPTION="squashfs-tools provides tools for creating and appending to \
+(mksquashfs) and extracting (unsquashfs) squashfs filesystems.
+
+Squashfs is a compressed read-only filesystem. It is intended for general \
+read-only filesystem use, archival use (i.e. in cases where a .tar.gz file \
+may be used, and in constrained block device/memory systems (e.g. embedded \
+systems) where low overhead is needed."
+HOMEPAGE="http://squashfs.sourceforge.net"
```
fbrosson: Please add a trailing slash:
```
HOMEPAGE="http://squashfs.sourceforge.net/"
```

hermord:
Alright, I'll fix it up right away!

hermord:
I pushed the amended commit out. For some reason, it still shows up in the same place as the old one. But it has been changed.

fbrosson:
Do you have lib:liblzma on x86_gcc2?

BTW, I noticed that several patches were already available on Gentoo, so we could do something similar to what we have for tiff-3.9.7.
```
SOURCE_URI="https://downloads.sf.net/squashfs/squashfs$portVersion.tar.gz"
CHECKSUM_SHA256="0d605512437b1eb800b4736791559295ee5f60177e102e4d4ccd0ee241a5f3f6"
SOURCE_DIR="squashfs$portVersion/squashfs-tools"

DEBIAN_REV="3"
SOURCE_FILENAME_2="squashfs-tools_$portVersion-$DEBIAN_REV.debian.tar.xz"
SOURCE_URI_2="http://http.debian.net/debian/pool/main/s/squashfs-tools/$FILENAME_2"
CHECKSUM_SHA256_2="1c296cc147e322e7124bf23a3c242ac83f6a986e6d6f64829ad2424b33914fc4"

commit="dceb729f0369d72f1d7820705fd12510b71446d2"
patchesBaseUri="https://gitweb.gentoo.org/repo/gentoo.git/plain/sys-fs/squashfs-tools/files"
SOURCE_FILENAME_3="squashfs-tools-$portVersion-2gb.patch"
SOURCE_FILENAME_4="squashfs-tools-$portVersion-aligned-data.patch"
SOURCE_FILENAME_5="squashfs-tools-$portVersion-local-cve-fix.patch"
SOURCE_FILENAME_6="squashfs-tools-$portVersion-mem-overflow.patch"
SOURCE_FILENAME_7="squashfs-tools-$portVersion-sysmacros.patch"
SOURCE_FILENAME_8="squashfs-tools-$portVersion-xattrs.patch"
SOURCE_URI_2="$patchesBaseUri/$SOURCE_FILENAME_2?id=$commit#noarchive"
SOURCE_URI_3="$patchesBaseUri/$SOURCE_FILENAME_3?id=$commit#noarchive"
SOURCE_URI_4="$patchesBaseUri/$SOURCE_FILENAME_4?id=$commit#noarchive"
SOURCE_URI_5="$patchesBaseUri/$SOURCE_FILENAME_5?id=$commit#noarchive"
SOURCE_URI_6="$patchesBaseUri/$SOURCE_FILENAME_6?id=$commit#noarchive"
SOURCE_URI_7="$patchesBaseUri/$SOURCE_FILENAME_7?id=$commit#noarchive"
SOURCE_URI_8="$patchesBaseUri/$SOURCE_FILENAME_8?id=$commit#noarchive"
CHECKSUM_SHA256_3="cdb9c85c2f733c0db879275d0ee2480f528aa3191f236bbf1ff681c75b14a024"
CHECKSUM_SHA256_4="3030ddfa5ba9d0baea7f5e4761b33dea67fbc4d49c86e475830fc1bfbfddb145"
CHECKSUM_SHA256_5="7edceedead4f5a57dc009afecd0f414d60a3878dc43f8de5224bff391ed32c76"
CHECKSUM_SHA256_6="a94f9df3439bef373da31a734e6dad42f5fc5a6d91f2b9d15d30b02f1aac73b3"
CHECKSUM_SHA256_7="b44afb79ba6108a657fe6c51cecdf0e9623cbc91234b297478a6c3308cf0eddf"
CHECKSUM_SHA256_8="f98fb6bee2f46ba57b9ab68dbc1ad15ed5ac8c7592cbdfead36d6642096acada"

PATCHES="
	squashfs_tools-$portVersion-cve2015-4645_4646.patch
	squashfs_tools-$portVersion-haiku-compat.patch
	squashfs_tools-$portVersion-fix-mem-calculation.patch
	squashfs_tools-$portVersion-lzma-magic.patch
	"

```
... and then:
```
PATCH()
{
	patch -p1 -i $sourceDir3/$SOURCE_FILENAME_3
	patch -p1 -i $sourceDir4/$SOURCE_FILENAME_4
	patch -p1 -i $sourceDir5/$SOURCE_FILENAME_5
	patch -p1 -i $sourceDir6/$SOURCE_FILENAME_6
	patch -p1 -i $sourceDir7/$SOURCE_FILENAME_7
	patch -p1 -i $sourceDir8/$SOURCE_FILENAME_8
}
```
The only problem is that the patches do not apply yet. Maybe I need to change "-p1".

hermord:
I actually grabbed three of the four other patches from the debian patchset. So we could apply them manually from `$sourceDir2` as well, and only keep the `haiku-compat` patch, which I wrote myself. You have to `cd` one level up for them to work, though, because they want to change files in a `squashfs-tools/` subdirectory.

hermord:
The reason I put them all in `patches` instead of applying them manually is to make the recipe less verbose. But if it's a bad practice to do so, I'll change the recipe.

fbrosson:
I think we should use as much external patches as possible.
That's OK if we have a PATCH function that talks too much ;-)

BYW, doesn't your recipe depend on lib:liblzma ? That lib is provided by xz_utils which does not build on x86_gcc2.

@waddlesplash, did you build the squashfs_tools on x86_gcc2?

hermord:
Oh, I didn't know `xz-utils` were unavailable on x86_gcc2. I'll set up a VM with that and try to get them to build. Worst case scenario, we could disable lzma and xz support.

fbrosson:
Since Debian has the remaining patches in squashfs-tools_4.3-3.debian.tar.xz we can probably add something like this in our PATCH function:
```
	for i in \
		0001-kfreebsd.patch \
		0002-fix_phys_mem_calculation.patch \
		0003-CVE-2015-4645_and_CVE-2015-4646.patch \
		0004-unsquashfs-add-support-for-LZMA-magics.patch
	do
		patch -p1 -i $sourceDir2/debian/patches/$i
	done
```

hermord:
Actually, I'd skip the first one (`0001-...`). It messes with my `haiku-compat` patch and, besides, isn't really useful on Haiku. But yes, we could do something like that. Gentoo packages actually snuck in a really insidious inconsistency in the `local-cve-fix` patch -- all other patches apply fine with `-p2`, but that one requires `-p1`.

fbrosson:
We could add support for x86 secondary arch and define LIBLZMA like this:
```
	LIBLZMA=liblzma
else
	LIBLZMA=
fi
```
Then we would use ${LIBLZMA+lib:$LIBLZMA} in REQUIRES and ${LIBLZMA+devel:$LIBLZMA} in BUILD_REQUIRES.

hermord:
Well, I'll try to get xz-utils to build on x86 first, since it'd be nice to have them there anyway.

fbrosson:
Do you mean "on x86_gcc2" ?
I did a patch 7 months ago for building xz_utils-5.2.2 on x86_gcc2 but I had some unsolved issues and the patch was quite big (2316 lines); so I had to give up.
I still have it locally, so if there is interest I can push it to a xz_utils branch of my remote repo, in case you (or someone else) would like to have a look at it.

hermord:
> Do you mean "on x86_gcc2" ?
Yes, sorry, I tend to conflate the two.
I understand that the issue is quite global, so I'd rather implement conditional xz/lzma support as you proposed and get this working on x86_gcc2, and then look into the `xz_utils` business.

hermord:
Wait, I think I messed up somewhere

hermord:
Oh yeah, I see it now.

hermord:
Just to give you an update, turns out, the sources don't quite build with GCC2. Investigating the issue.

hermord:
Alright, **now** everything seems to be in order.

hermord:
Oops, didn't mean to press that 😅

korli:
@fbrosson I'm personally not so happy with a ton of URL for patches which won't change forever, because the upstream project stalled. It's adding a dependency on Gentoo servers for no good reason, simply packing all patches in a gentoo patchset in patches directory would be a better and more maintainable.
Sorry I'm not a fan of verbosity and of PATCH() in general.

hermord:
Alright, I like this version a lot better. I do agree with @korli about verbosity, but I'm new here, so I don't really know what the common practice is.

fbrosson:
@hermord, I love your current recipe! (I'll test it asap.)

@korli, I'm sorry, I don't think that adding copies of patches that are already available somewhere would be a good idea. And imho the dependency on Gentoo and Debian is not a problem. Moreover, since we have checksums, we would detect any change or removal, should that ever happen.

korli:
Yes, I just expect your opinion to not become a rule.

waddlesplash:
I'm rather conflicted on the remote-patches idea myself; I'm tending towards @korli's stance on it but I haven't really thought about it. But we don't need to do it all that often, so I don't think it'll become a problem...

fbrosson:
@korli: Sure, this won't become a rule because there are very few recipes where this can happen.

@waddlesplash: I think you have found a nice consensus.

@hermord: squashfs_tools builds like a charm on x86 and x86_64. Will try it on x86_gcc2 right now.
This package was not easy and you did a great job!

All: Shall we validate the GCI task right now but keep the PR open for 1 or 2 days to do a little polishing (e.g. see if we can get less fuzzy patches by either sorting them or rebasing the patches) and maybe add support for building for x86 secondary arch?
(I can take care of investigating the polishing of the patches)

Thanks to all of you!

waddlesplash:
That sounds OK to me.

hermord:
Alright, so I got rid of two patches that are not needed on Haiku (they have to do with xattrs and linux-only includes). I also changed my own patches a bit so that they don't mess with offsets. Sadly, Debian patches still have wrong offsets and that doesn't seem to depend on ordering.

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,105 @@
...
+CHECKSUM_SHA256_4="3030ddfa5ba9d0baea7f5e4761b33dea67fbc4d49c86e475830fc1bfbfddb145"
+CHECKSUM_SHA256_5="7edceedead4f5a57dc009afecd0f414d60a3878dc43f8de5224bff391ed32c76"
+CHECKSUM_SHA256_6="a94f9df3439bef373da31a734e6dad42f5fc5a6d91f2b9d15d30b02f1aac73b3"
+
+PATCHES="
+	squashfs_tools-$portVersion-haiku-compat.patch
+	squashfs_tools-$portVersion-gcc2.patch"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
fbrosson: Can please replace
```
SECONDARY_ARCHITECTURES="x86"
```
by
```

```
(notice the blank line.)

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,105 @@
...
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	HAVE_XZ=1
+else
+	HAVE_XZ=
+fi
+
+PROVIDES="
+	squashfs_tools = $portVersion
+	cmd:mksquashfs = $portVersion
+	cmd:unsquashfs = $portVersion
```
fbrosson: Please change PROVIDES to:
```
PROVIDES="
	squashfs_tools$secondaryArchSuffix = $portVersion
	cmd:mksquashfs$secondaryArchSuffix = $portVersion
	cmd:unsquashfs$secondaryArchSuffix = $portVersion
	"
```

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,105 @@
...
+	"
+REQUIRES="
+	haiku
+	lib:liblz4
+	${HAVE_XZ:+lib:liblzma}
+	lib:liblzo2
+	lib:libz
```
fbrosson: Please change REQUIRES to:
```
REQUIRES="
	haiku$secondaryArchSuffix
	lib:liblz4$secondaryArchSuffix
	${HAVE_XZ:+lib:liblzma$secondaryArchSuffix}
	lib:liblzo2$secondaryArchSuffix
	lib:libz$secondaryArchSuffix
	"
```

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,105 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:liblz4
+	${HAVE_XZ:+devel:liblzma}
+	devel:liblzo2
+	devel:libz
```
fbrosson: Please change BUILD_REQUIRES to:
```
BUILD_REQUIRES="
	devel:liblz4$secondaryArchSuffix
	${HAVE_XZ:+devel:liblzma$secondaryArchSuffix}
	devel:liblzo2$secondaryArchSuffix
	"
```

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,105 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
```
fbrosson: Please change BUILD_PREREQUIRES to:
```
BUILD_PREREQUIRES="
	cmd:gcc$secondaryArchSuffix
	cmd:make
	"
```

waddlesplash:
> Sadly, Debian patches still have wrong offsets and that doesn't seem to depend on ordering.

OK; so we should fetch all of those and make a .patchset ourselves out of them.

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,105 @@
...
+CHECKSUM_SHA256="0d605512437b1eb800b4736791559295ee5f60177e102e4d4ccd0ee241a5f3f6"
+SOURCE_DIR="squashfs$portVersion/squashfs-tools"
+
+DEBIAN_REV="3"
+SOURCE_FILENAME_2="squashfs-tools_$portVersion-$DEBIAN_REV.debian.tar.xz"
+SOURCE_URI_2="http://httpredir.debian.org/debian/pool/main/s/squashfs-tools/$SOURCE_FILENAME_2"
+CHECKSUM_SHA256_2="1c296cc147e322e7124bf23a3c242ac83f6a986e6d6f64829ad2424b33914fc4"
+
+COMMIT="dceb729f0369d72f1d7820705fd12510b71446d2"
+PATCHES_BASE_URI="https://gitweb.gentoo.org/repo/gentoo.git/plain/sys-fs/squashfs-tools/files"
```
fbrosson: Can you please replace
```
COMMIT="dceb729f0369d72f1d7820705fd12510b71446d2"
PATCHES_BASE_URI="https://gitweb.gentoo.org/repo/gentoo.git/plain/sys-fs/squashfs-tools/files"
```
by
```
commit="dceb729f0369d72f1d7820705fd12510b71446d2"
patchesBaseUri="https://gitweb.gentoo.org/repo/gentoo.git/plain/sys-fs/squashfs-tools/files"
```
and change the SOURCE_URI_{3-8} to:
```
SOURCE_URI_3="$patchesBaseUri/$SOURCE_FILENAME_3?id=$commit#noarchive"
SOURCE_URI_4="$patchesBaseUri/$SOURCE_FILENAME_4?id=$commit#noarchive"
SOURCE_URI_5="$patchesBaseUri/$SOURCE_FILENAME_5?id=$commit#noarchive"
SOURCE_URI_6="$patchesBaseUri/$SOURCE_FILENAME_6?id=$commit#noarchive"
SOURCE_URI_7="$patchesBaseUri/$SOURCE_FILENAME_7?id=$commit#noarchive"
SOURCE_URI_8="$patchesBaseUri/$SOURCE_FILENAME_8?id=$commit#noarchive"
```

Thanks!

BTW, here are the reasons for these cosmetic changes:

* We currenly have "COMMIT" for 4 recipes, "commit" for 22 recipes and "srcGitRev" for 42 recipes.

When I started recommending "srcGitRev" sometime ago, I did so because it was already the most used custom variable, but I had not noticed that there were almost 20 "commit"s.
Since "srcGitRev" is a bit strange, I think we should stop using it and start using "commit" instead.

As said earlier, I think we should avoid all-caps in custom variables, so I would recommend to use commit instead of COMMIT.

* haikuporter complains about PATCHES_BASE_URI and says it will ignore it.
So I would recommend to use patchesBaseUri instead of PATCHES_BASE_URI.

fbrosson:
> OK; so we should fetch all of those and make a .patchset ourselves out of them.

@waddlesplash, I'm almost sure we can find an order (to apply the patches) that will make the offsets work fine. I'll dig this right now ;-)

fbrosson:
> Sadly, Debian patches still have wrong offsets and that doesn't seem to depend on ordering.

@hermord, There is necessarily an order which minimizes the number of offset mismatches. I'm trying to find that order.
And even if the best order still produces many offset mismatches, it still works! So I think we should keep the patches external.

BTW, +1 for having removed the sysmacros and xattrs patches ;-)

hermord:
> @hermord, There is necessarily an order which minimizes the number of offset mismatches. I'm trying to find that order.

Oh, definitely, it's just that `0002-fix-phys-mem-calculation.patch` misses by a whopping 267 line offset, which I don't think can be fixed by reordering.

fbrosson:
The ebuild Gentoo wrote for release 4.3 applies all the patches from Debian and then their own patches in this order:
* sysmacros.patch
* aligned-data.patch
* 2gb.patch
* local-cve-fix.patch
* mem-overflow.patch
* xattrs.patch

So I guess that if we also apply the exact same patches then should have no offset mismatches when not applying our own patches (i.e. the ones you wrote), and very few if we also apply them (because haikuporter applis our patches before calling the PATCH() function if I'm not mistaken.

fbrosson:
Hmm, maybe we could, some day, teach haikuporter to:
1) call **PRE**PATCH
2) apply the patches in PATCHES
3) call PATCH

hermord:
My own patches actually don't change the offsets, expect the second `action.c` hunk, but nothing else is patching `action.c`.

hermord:
Even when I manually apply the Debian patches one by one onto the original source tree, `0004-unsquashfs-add-support-for-LZMA-magics.patch` still misses by 267 lines (and that's the one, not `0002-...`, my mistake).

hermord:
I looked into Gentoo's `epatch` function that they use to patch things, and it silently ignores all non-fatal errors. So there's that.

fbrosson:
> My own patches actually don't change the offset

That's great. It makes the problem a little less complex ;-)
BTW, I apologize for the trouble I'm causing you with these patch issues.
So please don't bother with this if you would like to claim other tasks.
I'll see if I can reduce the number of offset mismatches and then I'll suggest a change, if I find something better.
I have all your patches and I can play with them for a while.
Of course, if you want to continue digging the topic you are welcome!

Hmm, before you switch to another recipe (if you chose to do so), could you please address the comments in my last review, to add support for building for secondary architectures?
I tried it (just before I wrote the review) and it works perfectly ;-)
Thanks!

> I looked into Gentoo's epatch function that they use to patch things, and it silently ignores all non-fatal errors. So there's that.

No regrets, then! Let's also ignore the offset mismatches!

sys-fs/squashfs_tools/patches/squashfs_tools-4.3-gcc2.patch
```diff
@@ -0,0 +1,50 @@
...
+ 
+ void read_fragment_2(unsigned int fragment, long long *start_block, int *size)
+ {
+-	TRACE("read_fragment: reading fragment %d\n", fragment);
+-
+ 	squashfs_fragment_entry_2 *fragment_entry = &fragment_table[fragment];
+ 	*start_block = fragment_entry->start_block;
+ 	*size = fragment_entry->size;
++
++	TRACE("read_fragment: reading fragment %d\n", fragment);
```
fbrosson: I think it would be better to insert a block, instead of moving TRACE down.
What about having this in the patch:
```Diff
 void read_fragment_3(unsigned int fragment, long long *start_block, int *size)
 {
 	TRACE("read_fragment: reading fragment %d\n", fragment);
-
+{
 	squashfs_fragment_entry_3 *fragment_entry = &fragment_table[fragment];
 	*start_block = fragment_entry->start_block;
 	*size = fragment_entry->size;
-}
+}}
```

hermord: That shouldn't make a difference, since `fragment` isn't modified in the function. But if that looks cleaner, then sure.

fbrosson: You are right, it does not make any difference because we are not building with TRACE enabled.
So yes, that's just to make things easier to change, should anyone wish to compile with TRACE enabled.

hermord:
I seem to be missing something, because `setarch x86` doesn't change anything.

sys-fs/squashfs_tools/patches/squashfs_tools-4.3-gcc2.patch
fbrosson: I've just generated a patchset with haikuporter. It starts with a mail-style header and then contains the patch:
```Diff
diff --git a/unsquash-2.c b/unsquash-2.c
index 0f2bfe8..5a3c39b 100644
--- a/unsquash-2.c
+++ b/unsquash-2.c
@@ -118,13 +118,13 @@ int read_fragment_table_2(long long *directory_table_end)
 void read_fragment_2(unsigned int fragment, long long *start_block, int *size)
 {
 	TRACE("read_fragment: reading fragment %d\n", fragment);
-
+	{
 	squashfs_fragment_entry_2 *fragment_entry = &fragment_table[fragment];
 	*start_block = fragment_entry->start_block;
 	*size = fragment_entry->size;
+	}
 }
 
-
 struct inode *read_inode_2(unsigned int start_block, unsigned int offset)
 {
 	static union squashfs_inode_header_2 header;
diff --git a/unsquash-3.c b/unsquash-3.c
index d5af57a..b5177c0 100644
--- a/unsquash-3.c
+++ b/unsquash-3.c
@@ -105,13 +105,13 @@ int read_fragment_table_3(long long *directory_table_end)
 void read_fragment_3(unsigned int fragment, long long *start_block, int *size)
 {
 	TRACE("read_fragment: reading fragment %d\n", fragment);
-
+	{
 	squashfs_fragment_entry_3 *fragment_entry = &fragment_table[fragment];
 	*start_block = fragment_entry->start_block;
 	*size = fragment_entry->size;
+	}
 }
 
-
 struct inode *read_inode_3(unsigned int start_block, unsigned int offset)
 {
 	static union squashfs_inode_header_3 header;
```

In the beginning it's a bit hard to understand what we need to do to have haikuporter generate pathsets.
But basically, if you start from scratch without any patch, you:
```
hp -b squashfs_tools
cd work-4.3/sources/squashfs4.3
vi unsquash-2.c unsquash-3.c
git add unsquash-2.c unsquash-3.c
git commit -m "gcc2 patch" unsquash-2.c unsquash-3.c
cd ../../..
hp -e squashfs_tools
```
(hp is an alias to "haikuporter --get-dependencies")
Warning: "hp -e" will delete any file called squashfs_tools-4.3.patch and overwrite squashfs_tools-4.3.patchset without any notice, so don't forget to make backups ;-)

Now if you already have a patchset with several distinct patches, you can change any of them by adding a commit and then doing an interactive squash/rebase against the git revision that corresponds to the import.
I can detail that if you need.

The other difficult topic is when you have a patch for an old release, and it does not apply to the new release. It took me a while to become confortable with how to use "git add" and "git am --continue" to refresh a patchset. If you need more info I can explain what I have learned ;-)

Using patchsets is recommended, especially if the patch is not trivial, because it simplifies the maintainance of the patchsets when we upgrade a recipe.

fbrosson:
You don't need to call setarch.
You just need to call:
hp -S squashfs_tools_x86
or
hp -S squashfs_tools_x86-4.3

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+SOURCE_URI_5="$patchsetBaseUri/$SOURCE_FILENAME_5?id=$commit#noarchive"
+SOURCE_URI_6="$patchsetBaseUri/$SOURCE_FILENAME_6?id=$commit#noarchive"
+CHECKSUM_SHA256_3="cdb9c85c2f733c0db879275d0ee2480f528aa3191f236bbf1ff681c75b14a024"
+CHECKSUM_SHA256_4="3030ddfa5ba9d0baea7f5e4761b33dea67fbc4d49c86e475830fc1bfbfddb145"
+CHECKSUM_SHA256_5="7edceedead4f5a57dc009afecd0f414d60a3878dc43f8de5224bff391ed32c76"
+CHECKSUM_SHA256_6="a94f9df3439bef373da31a734e6dad42f5fc5a6d91f2b9d15d30b02f1aac73b3"
+
+PATCHES="
+	squashfs_tools-$portVersion-haiku-compat.patch
+	squashfs_tools-$portVersion-gcc2.patch"
```
fbrosson: Should be:
```
PATCHES="
	squashfs_tools-$portVersion-haiku-compat.patch
	squashfs_tools-$portVersion-gcc2.patch
	"
```
(the closing double quotes goes to a separate line)

sys-fs/squashfs_tools/squashfs_tools-4.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	HAVE_XZ=1
+else
+	HAVE_XZ=
+fi
+
+PROVIDES="
+	squashfs_tools$secondaryArchSuffix = $portVersion
+	cmd:mksquashfs$secondaryArchSuffix= $portVersion
```
fbrosson: Missing space before = in
```
	cmd:mksquashfs$secondaryArchSuffix = $portVersion
```

hermord:
`devel:liblz4_x86: Name not found` is what I'm getting. Sorry, I'm not too familiar with building for secondary architectures. Are devel packages supposed to be arch-qualified?

fbrosson:
> devel:liblz4_x86: Name not found is what I'm getting. Sorry, I'm not too familiar with building for secondary architectures. Are devel packages supposed to be arch-qualified?

You just need to:
```
hp -S lz4_x86
```
(Well, the -S is not required, but I always use it.)

There's something I do since a few months: I have a dropliba script which gets rid of all static libs of a hpkg. (It basically unpacks the hpkg with "package extract", looks for all lib*.a files, deletes them, and recreates the package with "package create".) So everytime I build some package, I run dropliba on its devel package and then I install both the base and the devel package.
This way I am sure that my devel package and my base package always have the same revision.
The problem we see when you don't install devel packages is that when it becomes outdated then the devel package you had in haikuports/packages is moved to a haikuports/packages.obsolete/ so you don't have it anymore, unless you installed it ;-)

fbrosson:
When you build lz4 for x86 secondary arch, you should obtain these hpkgs in haikuports/packages/:
* lz4_x86-r131-1-x86_gcc2.hpkg
* lz4_x86_devel-r131-1-x86_gcc2.hpkg

You don't need to install any of them before building squashfs_tools_x86, but if you later wish to install squashfs_tools_x86 you will need to also install (at least) lz4_x86.
This is why I tend to install everything I need, including the devel packages (which I shrink with my dropliba script ;-) because any package that is in /system/packages/ has precedence over any other version of the same package in haikuports/packages/.

hermord:
Just in case, I updated the recipe to incorporate the costmetic edits.

Issue #651, opened 2016-06-19
Add a recipe for Csound


richienyhus:
https://github.com/csound/csound

Port of CSound 4.23 to BeOS :  [csound-4.23-beos-src.zip](https://github.com/haikuports/haikuports/files/322194/csound-4.23-beos-src.zip)

PR #951, opened 2016-12-24
Add Csound

dash102:


dash102:
Is it preferred to squash as I go and keep 1 commit, or would it be better to just keep committing as I go and squash in the end when everything is ready to merge?

fbrosson:
It doesn't really matter. Some contributors add commits, some other squash them everytime they push something.
I tend to squash mine because it gets rid of my wrong (or modified) snippets ;-)
Actually, I don't even squash. I use "git commit --amend", so it makes thnigs even easier, since I never need to squash, given that I use **`--amed`** (**except** for the very first commit in my new branch, and this is very important, of course.)

If you chose to also squash your commits everytime, or if you use **`--amed`**, then please make sure you update the timestamp of your suashed commit to the current time.
You can do so by using
```
git commit --amend --date="`date '+%a %b %-d %T %Y %z'`"
```

BTW, I have a script called gcad which has:
```
#!/bin/sh
```
and a variant, gca without the **`--date`**, which I use when I just want to update the commit message.

That said, if you find it easier to push separate commits, that's perfectly OK too, and many users do that BTW.

That's it! Now you know everythng ;-)

media-sound/csound/csound-6.06.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+	csound = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libsndfile
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	boost_devel
```
korli: Please be specific about which boost lib is used, same in REQUIRES.

media-sound/csound/csound-6.06.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+COPYRIGHT="1991 Barry Vercoe, John ffitch"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+srcGitRev="be2e7e36946ce767329d330c5cf926def3aa5851"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="6db8221e2074d9b9acb5eb0572da4406b11c5a896098a774f120b38927be4fb4"
+SOURCE_DIR="Csound-$srcGitRev"
+PATCHES="csound-$portVersion.patch"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
```
korli: Any chance to have a x86 secondary arch?

fbrosson: @dash102, adding support for secondary arch is not difficult. You may take the recipe of dvdauthor as a template ;-)

media-sound/csound/csound-6.06.0.recipe
dash102: How can I be more specific about the lib? Also, when you mention REQUIRES, do you mean I need to add boost_devel to REQUIRES since it's not there now?

korli: Never mind I thought boost would be a runtime dependency. Just let as is.

dash102:
Cool. Thanks for the help and happy holidays!

dash102:
Oops - this is meant for another pull request - was I supposed to wait for this to be merged before pushing something else?

richienyhus:
This PR should close enhancement #651 once accepted.

PR #954, opened 2016-12-24
Doxygen: bump version

miqlas:
No functional changes. The old (1.8.10) recipe not working on x86_64 (because too new flex), so here is a working one.


app-doc/doxygen/doxygen-1.8.12.recipe
```diff
@@ -17,20 +16,19 @@ Doxygen can also visualize the relations between the various elements \
 by means of include dependency graphs, inheritance diagrams, and \
 collaboration diagrams, which are all generated automatically."
 HOMEPAGE="http://www.doxygen.org"
```
fbrosson: Can you please add a trailing slash to have:
```
HOMEPAGE="http://www.doxygen.org/"
```
Thanks!

miqlas:
ready :)

PR #963, opened 2016-12-27
Add Recipe for HIDAPI (GCI 2016)

CodeforEvolution:
I have added a recipe and patchset for hidapi below, for some reason though, there are some weird merge commits added in though shouldn't affect anything I'd think.


dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
+SUMMARY="Library for Interfacing with HID-Class devices"
+DESCRIPTION="HIDAPI is a library working with multiple platforms for interfacing with HID-Class devices through USB and Bluetooth. (On Haiku, it only works with USB). "
```
korli: Please wrap at 80 characters.

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+HOMEPAGE="http://www.signal11.us/oss/hidapi/"
+COPYRIGHT="2010 Alan Ott
+          Signal 11 Software"
```
korli: A year is needed for Signal 11 Software

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+          GNU GPL v3"
+REVISION="1"
+srcGitRev="a6a622ffb680c55da0de787ff93b80280498330f"
+SOURCE_URI="https://github.com/signal11/hidapi/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="d94e39be127ab6ac6a2bbdf6d321d6492b1cdc6cc5c00b6901228eb6923e81a6"
+SOURCE_DIR="hidapi-$srcGitRev"
+PATCHES="hidapi-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
korli: x86 as secondary arch seems hardly supported in this recipe and patchset : missing $secondaryArch in PROVIDES and REQUIRES, on pkgconfig in BUILD_PREREQUIRES.

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+	cmd:ld$secondaryArchSuffix
+         cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+         cmd:pkg_config
+	"
+
+BUILD()
+{
+
+         ./bootstrap
```
korli: Please use one tab.

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+
+         runConfigure ./configure \
```
korli: Ditto

dev-libs/hidapi/patches/hidapi-20160919.patchset
```diff
@@ -0,0 +1,223 @@
...
++CFLAGS   ?= -Wall -g -fPIC
++
++CXX      ?= c++
++CXXFLAGS ?= -Wall -g
++
++COBJS     = hid.o
++CPPOBJS   = ../hidtest/hidtest.o
++OBJS      = $(COBJS) $(CPPOBJS)
++INCLUDES  = -I../hidapi -I/boot/system/develop/headers
++LDFLAGS   = -L/boot/system/develop/lib
```
korli: Shouldn't be hard coded.

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+CHECKSUM_SHA256="d94e39be127ab6ac6a2bbdf6d321d6492b1cdc6cc5c00b6901228eb6923e81a6"
+SOURCE_DIR="hidapi-$srcGitRev"
+PATCHES="hidapi-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	hidapi = $portVersion
+	lib:libhidapi = $portVersion
```
fbrosson: Please replace
```
	hidapi = $portVersion
	lib:libhidapi = $portVersion
```
by
```
	hidapi$secondaryArchSuffix = $portVersion
	lib:libhidapi$secondaryArchSuffix = $portVersion
```

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libusb_1.0$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	hidapi_devel = $portVersion
+	devel:libhidapi = $portVersion
```
fbrosson: Please replace
```
	hidapi_devel = $portVersion
	devel:libhidapi = $portVersion
```
by
```
	hidapi${secondaryArchSuffix}_devel = $portVersion
	devel:libhidapi$secondaryArchSuffix = $portVersion
```

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+	"
+REQUIRES_devel="
+	hidapi == $portVersion base
```
fbrosson: Please replace
```
	hidapi == $portVersion base
```
by
```
	hidapi$secondaryArchSuffix == $portVersion base
```

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+         cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+         cmd:pkg_config
```
fbrosson: Please use a tab in lines 45 and 47.

dev-libs/hidapi/hidapi-20160919.recipe
fbrosson: I might be wrong, but if the archive already has a configure script then you don't need to call ./bootstrap.

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,76 @@
...
+                  --with-gnu-ld
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	prepareInstalledDevelLibs libhidapi
```
fbrosson: You can use prepareInstalledDevel**Lib** instead of prepareInstalledDevel**Libs**.

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,80 @@
...
+CHECKSUM_SHA256="d94e39be127ab6ac6a2bbdf6d321d6492b1cdc6cc5c00b6901228eb6923e81a6"
+SOURCE_DIR="hidapi-$srcGitRev"
+PATCHES="hidapi-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	hidapi$secondaryArchSuffix = $portVersion
+	lib:libhidapi$secondaryArchSuffix = $portVersion
```
fbrosson: Please replace
```
	lib:libhidapi$secondaryArchSuffix = $portVersion
```
by
```
	lib:libhidapi$secondaryArchSuffix = 0.0.0 compat >= 0
```

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,80 @@
...
+		--disable-dependency-tracking \
+		--with-gnu-ld
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
```
fbrosson: Please insert this:
```

	rm $libDir/libhidapi.la

```
between the "make install" and the "prepareInstalledDevelLib libhidapi".
(Yes, it's nice to have one blank line above and another below.)

dev-libs/hidapi/hidapi-20160919.recipe
```diff
@@ -0,0 +1,80 @@
...
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make hidtest
+
+	cd hidtest/.libs
+	./hidtest
```
fbrosson: Please replace
```
	./hidtest
```
by
```
	LIBRARY_PATH=$sourceDir/libusb/.libs:$LIBRARY_PATH ./hidtest || true
```

CodeforEvolution:
Alright, I will be on to that, otherwise, I never knew why though why is the libtool file removed?

korli:
Libtool files generally hardcode paths which can end up broken when a dependency is updated. Some Linux distributions remove them too for instance.

pulkomandy:
The .la files were introduced by the libtool project. The goal was to allow automatic linking of transitive dependencies. For example, app needs sdl_image, which needs libpng. It would be nice if app only links against sdl_image, and libpng is pulled in automatically.
On some (old) systems, you need to link in libpng at link time. The .la file allows libtool to know that, and do the right thing.

On Haiku however, the runtime_loader is able to figure things out on its own at run time (when the application is loaded and run). This means the .la files are not needed. They have some downsides, too:
- When using a .la file for sdl_image in the above example, the app would end up linking directly with libpng. This means if we update sdl_image to change which version of libpng it links to, all apps now need to be rebuilt, instead of just the lib.
- As korli mentionned, the .la file tend to use absolute paths to other .la files, and reference specific versions of libraries, which is not needed on Haiku and tends to break packages.

Ideally we should tell libtool and the autotools to not create the files at all, but no one had the time to dig into how to do that. So for now, we just delete them after the fact.

CodeforEvolution:
Ah, I see, thanks for the explanation!

CodeforEvolution:
I have updated the recipe once again, hopefully that should be good to go.

Issue #964, opened 2016-12-27
OpenBLAS broken recipe


miqlas:
The OpenBLAS recipe doesn't make any libopenblas.so.0 in /boot/system/lib, just in /boot/system/develop/lib, but it isn't in stander LD PATH (fixme if i'm wrong).
So Octave cannot find the installed BLAS.

I made a symlink in non-packaged/lib, now Octan can find it, but gets another problem:

> runtime_loader: /boot/system/lib/libopenblasp-r0.2.19.so: Could not resolve symbol 'get_num_procs'
> resolve symbol "get_num_procs" returned: -2147478780
> runtime_loader: /boot/system/lib/libopenblasp-r0.2.19.so: Troubles relocating: Symbol not found

Maybe it requires some extra patching.
@mmuman, @korli : any idea?

miqlas:
Fixed.

PR #972, opened 2016-12-30
OpenBLAS: fixed port, enabled fortran code

miqlas:
Solves: https://github.com/haikuports/haikuports/issues/964


sci-libs/openblas/openblas-0.2.19.recipe
```diff
@@ -19,6 +20,7 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
+	gcc${secondaryArchSuffix}_syslibs
```
korli: You should specify which lib you need.

miqlas:
Ready.

PR #973, opened 2016-12-30
Fixed recipe for sqlitebrowser

xdizzaster:
This is a fix to the broken recipe that was merged in #969. Due to my own misunderstanding of resource files and how they work in Haiku, I did not notice that the wrong filename was being used and that would make the build fail. This PR is a fix to that.


korli:
Thanks!

PR #978, opened 2016-12-31
pango, bump version

Begasus:


x11-libs/pango/patches/pango-1.40.3.patchset
```diff
@@ -1,7 +1,7 @@
-From 9448ecaaebab4f4fed512101b27a2c697552d803 Mon Sep 17 00:00:00 2001
-From: Jerome Duval <jerome.duval@gmail.com>
-Date: Mon, 12 Oct 2015 22:19:59 +0000
-Subject: Haiku patch
+From 524b659f611461e0dd8dd8f8b47c99d0c258989c Mon Sep 17 00:00:00 2001
+From: begasus <begasus@gmail.com>
```
fbrosson: @Begasus, why don't you just keep the patch unchanged, intead of replacing the metadata?
When you noticed that the patch from 1.40.2 was ok for 1.40.3 you should have used:
```
git mv pango-1.40.2.patchset pango-1.40.3.patchset
```

Begasus: My bad, I was checking if the provided patch was really needed, it turns out that even without the patch the check work the same as without the patch, so maybe it shouldn't be needed at all?

Begasus:
@fbrosson  ... updated :)

PR #980, opened 2017-01-01
libpng12: bump to 1.2.57, drop libtool files.

fbrosson:


fbrosson:
For libpng12-1.2.56 we were using **3**.56.0 for libpng.so and libpng12.so, while we should probably be using **0**.56.0 for libpng12, and **3**.56.0 **_only_** for libpng.so.

This is being fixed here with two custom variables: libVersion and libVersion12.
Shall I keep this fix (and mention it in the commit message with a "fix soname for libpng12") or drop libVersion12 (to use libVersion for both libpng.so and libpng12.so) ?

korli:
This package should continue to provide "provides" compatible with packages requiring it. Fixing provides for the sake of being correct isn't fair when users have to cope with the effects...

PR #981, opened 2017-01-01
BePDF: reordered recipes, enabled x64 for 2.x, patch

miqlas:


waddlesplash:
If you could submit those patches as PRs to the upstream BePDF repo, that'd be great.

PR #988, opened 2017-01-05
binutils: also build PE targets for x86_64

jessicah:


sys-devel/binutils/binutils-2.26.1_2016_07_22.recipe
```diff
@@ -94,17 +95,25 @@ BUILD()
...
+	# Touch all *.y files, ensure they get generated; without this, the
+	# PE targets would fail to link
+	(cd $sourceDir; find . -name \*.y | xargs touch)
 
 	mkdir -p $objectsDir
 	cd $objectsDir
 
+	binUtilsTargets=$effectiveTargetMachineTriple
+
+	if [ $effectiveTargetArchitecture == x86_64 ]; then
```
fbrosson: @jessicah, I think it would be (slightly) better to use
```
[ $effectiveTargetArchitecture = x86_64 ]
```
instead of
```
[ $effectiveTargetArchitecture == x86_64 ]
```
because **`==`** only works with bash, while **`=`** works with both bash ans sh.

PR #990, opened 2017-01-06
cutecom: added recipe

xdizzaster:
Added recipe for CuteCom graphical serial terminal.

### DIFF
--- a/net-dialup/cutecom/cutecom-0.22.0.recipe
+++ b/net-dialup/cutecom/cutecom-0.22.0.recipe
@@ -1,0 +1,71 @@
+SUMMARY="Graphical serial terminal"
+DESCRIPTION="CuteCom is a graphical serial terminal, like minicom."
+HOMEPAGE="http://cutecom.sourceforge.net/"
+COPYRIGHT="2004-2009 Alexander Neundorf"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://cutecom.sourceforge.net/cutecom-$portVersion.tar.gz"
+CHECKSUM_SHA256="1b6620a6159cf3d50bb36cce544e91486817df7f1d553bf239d6db6108dd2ea5"
+PATCHES="cutecom-$portVersion.patchset"
+ADDITIONAL_FILES="cutecom.rdef.in"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	cutecom$secondaryArchSuffix = $portVersion
+	app:cutecom$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libQtCore$secondaryArchSuffix
+	lib:libQtGui$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libQtCore$secondaryArchSuffix
+	devel:libQtGui$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	"
+
+PATCH()
+{
+	sed -i \
+		-e "s|share/man\(/man1\)|$relativeManDir\1|" \
+		-e "/binInstallDir bin/ s|bin|$relativeAppsDir|2" \
+		CMakeLists.txt
+}
+
+BUILD()
+{
+	mkdir -p build
+	cd build
+	cmake -DCMAKE_INSTALL_PREFIX=$prefix ..
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	local MAJOR="`echo "$portVersion" | cut -d. -f1`"
+	local MIDDLE="`echo "$portVersion" | cut -d. -f2`"
+	local MINOR="`echo "$portVersion" | cut -d. -f3`"
+
+	sed \
+		-e "s|@MAJOR@|$MAJOR|" \
+		-e "s|@MIDDLE@|$MIDDLE|" \
+		-e "s|@MINOR@|$MINOR|" \
+		$portDir/additional-files/cutecom.rdef.in > cutecom.rdef
+	addResourcesToBinaries cutecom.rdef $appsDir/cutecom
+
+	addAppDeskbarSymlink $appsDir/cutecom CuteCom
+}
--- a/net-dialup/cutecom/patches/cutecom-0.22.0.patchset
+++ b/net-dialup/cutecom/patches/cutecom-0.22.0.patchset
@@ -1,0 +1,22 @@
+From b0dbd64f4f75b01be2b8ad8dee3b14b655b72c82 Mon Sep 17 00:00:00 2001
+From: Tudor Nazarie <nazarietudor@gmail.com>
+Date: Fri, 6 Jan 2017 00:11:55 +0200
+Subject: Fix termios.h error
+
+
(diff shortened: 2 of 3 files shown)


net-dialup/cutecom/patches/cutecom-0.22.0.patchset
```diff
@@ -0,0 +1,35 @@
...
+diff --git a/qcppdialogimpl.cpp b/qcppdialogimpl.cpp
+index c419d0f..44a9f95 100644
+--- a/qcppdialogimpl.cpp
++++ b/qcppdialogimpl.cpp
+@@ -51,7 +51,7 @@ using namespace std;
+ #include <unistd.h>
+ #include <ctype.h>
+ #include <sys/ioctl.h>
+-#include <sys/termios.h>
++#include <posix/termios.h>
```
korli: Should be #include <termios.h>

net-dialup/cutecom/patches/cutecom-0.22.0.patchset
```diff
@@ -0,0 +1,35 @@
...
+diff --git a/qcppdialogimpl.h b/qcppdialogimpl.h
+index a70c506..38a8a22 100644
+--- a/qcppdialogimpl.h
++++ b/qcppdialogimpl.h
+@@ -21,7 +21,7 @@
+ 
+ #include "ui_cutecommdlg.h"
+ 
+-#include <termios.h>
++#include <posix/termios.h>
```
korli: This was correct before.

xdizzaster:
Updated the patch like @korli suggested.

fbrosson:
@xdizzaster, a few days ago I tried to find out whether it would be possible to pass a few more parameters to cmake in order to have the man pages installed to the correct destination instead of having to:
```
	mkdir -p $manDir
	mv $appsDir/share/man/* $manDir
	rm -r $appsDir/share
```
but I seems it can only be done by patching CMakeLists.txt.

If you wish to investigate on this you can
```
cd haikuports/net-dialup/cutecom/work-0.22.0
grep -E '^manDir|^relativeManDir|^prefix' wrapper-script
```
and find out what sed commands you can use to patch CMakeLists.txt using $relativeManDir.
I can guide you if you wish.

Well, it seems I've got the answer. You just need to insert a PATCH function before BUILD:
```Bash
PATCH()
{
	sed -i -e "s|share/man\(/man1\)|$relativeManDir\1|" CMakeLists.txt
}
```

I haven't tested this. I'll let you try it if you wish.

xdizzaster:
I'm generally a fan of "Don't fix it if it ain't broken", but as the solution with a `PATCH()` function is much more elegant than moving man pages around, I tried your function. Unfortunately it does not put them in the correct folder, under `documentation`, but under `apps/documentation`. I'll attach a screenshot to show you what I mean:
virtualbox_haiku x86 gcc 5 hybrid_07_01_2017_22_26_14

fbrosson:
Oops, you are right. I forgot that the recipe is passing $appsDir insted of $prefix for the PREFIX.
But I think it can be fixed with:
```
	cmake -DCMAKE_INSTALL_PREFIX=$prefix binInstallDir=$appsDir ..
```
(but I haven't tried that yet.)

xdizzaster:
Updated to include @fbrosson's `PATCH()` function instead of moving the man pages around.

fbrosson:
Hmm, I though that with binInstallDir=$appsDir (or -DbinInstallDir=$appsDir) we would not need to
```
	mkdir -p $appsDir
	mv $prefix/bin/cutecom $appsDir/CuteCom
	rm -r $prefix/bin
```
There must be a nice way to avoid this, and I think we won't need to add another patch to CMakeLists.txt.
I'll have a closer look at it later.

xdizzaster:
I think that another patch might be the cleanest solution. I tried multiple ways of passing `binInstallDir` and they all failed.

fbrosson:
Look what I found in line 17 of CMakeLists.txt:
```
set(binInstallDir bin )
```
So I think this is what we need:
```Bash
PATCH()
{
	sed -i \
		-e "s|share/man\(/man1\)|$relativeManDir\1|" \
		-e "/binInstallDir bin/ s|bin|$relativeAppsDir|" \
		CMakeLists.txt
}
```

xdizzaster:
Unfortunately your command is broken. It replaces the bin in binInstallDir, making it appsInstallDir. I would love to fix it myself, but I don't know enough about sed and regexes (never got around to learning them properly, but I keep telling myself I will someday) and it would just be quicker to signal the mistake to you.

xdizzaster:
Nevermind, I did it myself. The correct function is: 
```
PATCH()
{
	sed -i \
		-e "s|share/man\(/man1\)|$relativeManDir\1|" \
		-e "/binInstallDir bin/ s|bin|$relativeAppsDir|2" \
		CMakeLists.txt
}
```

net-dialup/cutecom/cutecom-0.22.0.recipe
```diff
@@ -0,0 +1,73 @@
...
+	cmake -DCMAKE_INSTALL_PREFIX=$prefix ..
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	mv $appsDir/cutecom $appsDir/CuteCom
```
fbrosson: I would drop the mv (i.e. keep the runtime all lowercase) and only use CuteCom for the AppDeskbarSymlink.
On GNU/Linux, for instance, the runtime is called cutecom, so I think we should stick to that (and only use CuteCom on the DeskBar.)

net-dialup/cutecom/cutecom-0.22.0.recipe
```diff
@@ -0,0 +1,73 @@
...
+
+	local MAJOR="`echo "$portVersion" | cut -d. -f1`"
+	local MIDDLE="`echo "$portVersion" | cut -d. -f2`"
+	local MINOR="`echo "$portVersion" | cut -d. -f3`"
+
+	sed \
+			-e "s|@MAJOR@|$MAJOR|" \
+			-e "s|@MIDDLE@|$MIDDLE|" \
+			-e "s|@MINOR@|$MINOR|" \
+			$portDir/additional-files/cutecom.rdef.in > cutecom.rdef
```
fbrosson: You can reduce the indentation by one tab in lines 66 to 69.

net-dialup/cutecom/cutecom-0.22.0.recipe
```diff
@@ -0,0 +1,73 @@
...
+	addResourcesToBinaries cutecom.rdef $appsDir/CuteCom
+
+	addAppDeskbarSymlink $appsDir/CuteCom CuteCom
```
fbrosson: If you drop the mv command above, don't forget to change this to:
```
	addAppDeskbarSymlink $appsDir/cutecom CuteCom
```

xdizzaster:
Cosmetic edits as @fbrosson suggested.

net-dialup/cutecom/cutecom-0.22.0.recipe
```diff
@@ -0,0 +1,71 @@
+SUMMARY="Graphical serial terminal"
+DESCRIPTION="CuteCom is a graphical serial terminal, like minicom."
+HOMEPAGE="http://cutecom.sourceforge.net"
```
fbrosson: Please change HOMEPAGE to:
```
HOMEPAGE="http://cutecom.sourceforge.net/"
```
(trailing slash added)

fbrosson: Although some (or many) browsers and many websites don't display the trailing slashes, URIs which do not have a path part (i.e. which only have the protocol and a domain name) **need** a trailing slash in order to be valid.

Now, if a tool or webpage which displays URIs wants to hide the trailing slashes, that's perfectly OK.

xdizzaster:
Added trailing slash to `HOMEPAGE`, as @fbrosson suggested.

PR #991, opened 2017-01-06
unrtf: added recipe

xdizzaster:
Added recipe for UnRTF.

### DIFF
--- a/app-text/unrtf/unrtf-0.21.9.recipe
+++ b/app-text/unrtf/unrtf-0.21.9.recipe
@@ -1,0 +1,63 @@
+SUMMARY="Rich Text Format command line converter"
+DESCRIPTION="UnRTF is a command-line program written in C which converts \
+documents in Rich Text Format (.rtf) to HTML, LaTeX, troff macros, and RTF \
+itself."
+HOMEPAGE="https://www.gnu.org/software/unrtf/"
+COPYRIGHT="2001-2015 Free Software Foundation"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://www.gnu.org/software/unrtf/unrtf-$portVersion.tar.gz"
+CHECKSUM_SHA256="22a37826f96d754e335fb69f8036c068c00dd01ee9edd9461a36df0085fb8ddd"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	unrtf$secondaryArchSuffix = $portVersion
+	cmd:unrtf = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libiconv$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autom4te
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	aclocal
+	automake --add-missing
+	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir\
+		LIBS=-liconv
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	sed -i \
+		-e "s|$dataDir/|/system/$relativeDataDir/|" \
+		$manDir/man1/unrtf.1
+}
+
+TEST()
+{
+	make check
+}


app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+itself."
+HOMEPAGE="https://www.gnu.org/software/unrtf/"
+COPYRIGHT="2001-2015 Free Software Foundation"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://www.gnu.org/software/unrtf/unrtf-$portVersion.tar.gz"
+CHECKSUM_SHA256="22a37826f96d754e335fb69f8036c068c00dd01ee9edd9461a36df0085fb8ddd"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
```
fbrosson: Please replace
```
```
by
```

	commandBinDir=$binDir
else
	commandBinDir=$prefix/bin
fi
```

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+
+PROVIDES="
+	unrtf$secondaryArchSuffix = $portVersion
+	cmd:unrtf$secondaryArchSuffix = $portVersion
```
fbrosson: Please replace
```
	cmd:unrtf$secondaryArchSuffix = $portVersion
```
by
```
	cmd:unrtf = $portVersion
```

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	automake$secondaryArchSuffix == 1.15.0
+	devel:libiconv$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:ln
```
fbrosson: No need for cmd:ln since it is provided by coreutils and iis therefore implicitely available.

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+	cmd:make
+	"
+
+BUILD()
+{
+	rm -rf config
+	ln -s /boot/system/data/automake-1.15/ config
+	runConfigure ./configure LIBS=-liconv
```
fbrosson: Please replace
```
	runConfigure ./configure LIBS=-liconv
```
by
```
	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir \
		LIBS=-liconv
```

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
```
fbrosson: Please add
```

	sed -i \
		-e "s|$dataDir/|/system/$relativeDataDir/|" \
		$manDir/man1/unrtf.1
```
after "make install"

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+	cmd:unrtf$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	automake$secondaryArchSuffix == 1.15.0
```
fbrosson: Please remove
```
	automake$secondaryArchSuffix == 1.15.0
```
from  BUILD_REQUIRES and add:
```
	cmd:automake == 1.15.0
```
to BUILD_**PRE**REQUIRES (and sort alphabetically.)

app-text/unrtf/unrtf-0.21.9.recipe
xdizzaster: I confess that I skipped testing it on x86_gcc2. Could you explain what this change does? Why change the binary directory?

fbrosson:
I did not investigate on why automake 1.15 was required, but I guess you did ;-)

Regarding the configuration, the man page Isuggests that the user can customize /system/data/unrtf/html.conf but this is not possible because this is a read-only directory.
This means we could add some changes to move that file to a directory with read-write permission: /system/settings/unrtf.
But I think this will be quite tricky, so maybe we can keep this file where it is.
That said, if you want to try to learn how setting files are handled on Haiky, you are welcome ;-)
You can have a look at www(servers/lighttpd/ but that recipe is quite complex, so you migjht find better examples, if you wish to improve the recipe.

xdizzaster:
Updated as @fbrosson suggested. I commented on one of the changes as

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,61 @@
...
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	rm -rf config
+	ln -s /boot/system/data/automake-1.15/ config
+	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir\
+			LIBS=-liconv
```
fbrosson: 2 tabs (instead of 3) are fine for line 46.

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,61 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	sed -i \
+			-e "s|$dataDir/|/system/$relativeDataDir/|" \
+			$manDir/man1/unrtf.1
```
fbrosson: 2 tabs (instead of 3) are fine for lines 54 and 55.

xdizzaster:
It does not need specifically automake 1.15, but unrtf uses a bunch of symlinks to some tools provided by automake (install-sh, config.guess, etc.) that are broken on Haiku because of different filesystem layout. I found that the easiest way to fix this was to symlink to the local automake tools and, since they are located in a directory like `/boot/system/data/automake-${automake_version}/` and I didn't know how to reliably get the path of that directory, I hardcoded automake 1.15 as a dependency.

pulkomandy:
6 janvier 2017 17:25 "Tudor Nazarie" <notifications@github.com> a écrit:
> @xdizzaster commented on this pull request.

> +ARCHITECTURES="x86_gcc2 x86 x86_64"
> +SECONDARY_ARCHITECTURES="x86_gcc2 x86"

> I confess that I skipped testing it on x86_gcc2. Could you
> explain what this change does? Why change the binary directory?
> 

The default path for the secondary architecture is bin/x86/ for binaries, so they are not in the path and a symlink is created in bin with the -x86 suffix (so you leed to run prog-x86 instead of prog).
This only makes sense where two versions of a binary need to coexist, one for gcc5, and one for gcc2. Examples are gcc itself, pkg-config, and a few other build tools.
In all other cases, we try to move the binaries to the main "bin" dir, either by using the primary architecture version, or by moving the secondary one into the "main" bin directory.

xdizzaster:
Updated with the indentation changes @fbrosson suggested.

fbrosson:
Let me add that we only use the commandBinDir trick when a recipe cannot build for x86_gcc2. So we know that it's OK to build it for x86 secondary arch and have it installed to /system/bin/ instead of /system/bin/x86/.
This is mainly to avoid having a /system/bin/unrtf-x86 -> x86/unrtf symlink, and it allows (on x86_gcc2) to call unrtf instead of unrtf-x86.

xdizzaster:
Now that I think about it, I could copy the required tools from my local install and add them as a patch to avoid having automake 1.15 as a dependency.

xdizzaster:
Updated adding a patch with the required automake tools and removing the hardcoded dependency for automake 1.15

fbrosson:
Hmm, it looks like we won't need a patch. I'll write a review with the changes for that.

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+	haiku$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libiconv$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
```
fbrosson: If you replace
```
	cmd:awk
```
by
```
	cmd:aclocal
	cmd:autom4te
	cmd:automake
```
and ...

app-text/unrtf/unrtf-0.21.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir\
```
fbrosson: ... and insert these 2 line in BUILD before calling runConfigure:
```
	aclocal
	automake --add-missing
```
then the recipe builds fine ;-)

xdizzaster:
Updated the recipe and removed the patch, thanks to @fbrosson's suggestion to use aclocal and automake. Thanks!

fbrosson:
Thanks for your PR and don't forget to click submit on your GCI task ;-)

BTW, I noticed that your updates appear at the top of this PR. This is probably because you used "git commit --amend" which does not update the timestamps.
This is perfecly OK to use the --amend option. I would recommend, however, to also use the --date option (when you use the --amend option) so that your updates get correct timestamps.
You can do so from the command line:
```
git commit --amend --date="`date '+%a %b %-d %T %Y %z'`" myfile(s)
```
You might wish to create yourself a script, e.g. /system/non-packaged/bin/gcad with:
```
#!/bin/sh
```

Of course, you don't need to update this PR since the recipe is OK.
I'm jjust giving you this trick for your next recipes because it is easier to follow the commit updates when they appear at the right chronological place, between the comments ;-)

xdizzaster:
Thanks for the tip and for your time!

Issue #734, opened 2016-07-24
Add a recipe for FairTrade


richienyhus:
https://github.com/czeidler/fairtrade

korli:
Recipe added in 86ba526193456737ab0bb0ec04651b8ecd998bb0

PR #992, opened 2017-01-06
Recipe for fairtrade

dacianf:


haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
+SUMMARY="The fair trade application supports sellers during the trade"
```
fbrosson:
```
SUMMARY="A keyboard only app for fair trade sellers"
```
might be a slightly summary.

haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+DESCRIPTION="This application is designed for our local fair trade group (also see http://www.transfair.org/ (German link) for more info about the fair trade philosophy). The fair trade application supports sellers during the trade. Products can be scanned using a bar code scanner. All scanned products are showed in a shopping basket. For example, the seller scans a customer's product and the product appears in the shopping basket. If a customer wants to buy multiple products of the same kind, the product only has to be scanned once and the amount of products can be entered using the keypad. /
+Furthermore, it helps managing products on stock and provides an easy way to add new products. Sold products are automatically removed from the stock when completing a bargain. Product prices, descriptions, names and quantities can be edit as well. Old products can be disabled from the product view."
```
fbrosson: DESCRIPTION should be wrapped to lines <= 80 chars (as you can see on most recipes.)
BTW, the current description is probably too long and needs some changes: "for our local fair group" shoud probablly be replaced by "for the faire trade group run by the owners of www.transfair.org.

haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+HOMEPAGE="https://github.com/czeidler/fairtrade"
+COPYRIGHT="2006-2013 Clemens Zeidler"
+LICENSE="MIT"
+REVISION="1"
+commit="2b4ff648e2118021b092c4f0d9d0b61bc51c1374"
+SOURCE_URI="${HOMEPAGE}/archive/$commit.tar.gz"
+CHECKSUM_SHA256="8c4f411c008519c43fb5d4fa77867e883332fcea35c36fe9522223d3ec40695a"
```
fbrosson: The chsckum we currently have in the recipe corresponds to the zip archive.
Since we swtiched to the tar.gz archive you have to update the checksum:
```
CHECKSUM_SHA256="3508410755c1bdf9c90403fd5bbf622f5f6036fe7a787cfe570c6542fd9f4186"
```

BTW, the build fails for me. I've tried so far fairtgrade and fairtrade_x86.

dacianf:
I've fixed it

haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+BUILD()
+{
+	make $jobArgs OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	cp -a objects/fairtrade $appsDir
+}
```
fbrosson: You can add the following line, after the cp:
```
	addAppDeskbarSymlink $appsDir/fairtrade
```

haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	fairtrade$secondaryArchSuffix = $portVersion
+	app:fairtrade$secondaryArchSuffix = $portVersion
+	"
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:sqlite3$secondaryArchSuffix
```
fbrosson: Should be
```
	lib:libsqlite3$secondaryArchSuffix
```

haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
...
+LICENSE="MIT"
+REVISION="1"
+commit="2b4ff648e2118021b092c4f0d9d0b61bc51c1374"
+SOURCE_URI="${HOMEPAGE}/archive/$commit.tar.gz"
+CHECKSUM_SHA256="3508410755c1bdf9c90403fd5bbf622f5f6036fe7a787cfe570c6542fd9f4186"
+SOURCE_DIR="fairtrade-$commit"
+PATCHES="fairtrade-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
fbrosson: It looks like you will need to replace
```
ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
SECONDARY_ARCHITECTURES="x86"
```
by
```
ARCHITECTURES="x86_gcc2 !x86 !x86_64"
SECONDARY_ARCHITECTURES="x86_gcc2 !x86"
```

haiku-apps/fairtrade/fairtrade-1.0.0.recipe
```diff
@@ -0,0 +1,51 @@
+SUMMARY="A keyboard only app for fair trade sellers"
```
fbrosson: Hmm, I found a better summary compared to the one I had suggested.
It's taken from the [project homepage](https://github.com/czeidler/fairtrade).
```
SUMMARY="A simple to use point of sale system and product management"
```

BTW, you forgot to split DESCRIPTION into lines <= 80 chars.

Regarding the name of the recipe, maybe it should be changed to fairtrade-0.0.0~git_2014.06.22.recipe or something similar.
@korli, what name should we use?

dacianf:
@fbrosson is it ok now?

fbrosson:
Yep. It's OK. Thanks for the update!

korli:
1.0.0~git is fine actually. Any reason to keep SECONDARY_ARCHITECTURES?

fbrosson:
> 1.0.0~git is fine actually.

OK, thanks!

> Any reason to keep SECONDARY_ARCHITECTURES?

It fails to build on x86.

Full answer for newbies: having x86_gcc2 in SECONDARY_ARCHITECTURES (and some $secondaryArchSuffix here and there) allows to build and install fairtrade_x86_gcc2 on _x86 hybrid_. Without support for secondary architectures it would only be possible to install fairtrade on x86_gcc2 (and x86_gcc2 hybrid).

@dacianf, to rename the recipe (and patch file) with the name suggested by @korli, you can:
```
cd haikuports/haiku-apps/fairtrade

git mv fairtrade-1.0.0.recipe fairtrade-1.0.0~git.recipe
git commit --amend --no-edit fairtrade-1.0.0.recipe fairtrade-1.0.0~git.recipe

cd patches

git mv fairtrade-1.0.0.patchset fairtrade-1.0.0~git.patchset
git commit --amend --no-edit fairtrade-1.0.0.patchset fairtrade-1.0.0~git.patchset
```
Could you please run these commands (and do a push) next time you run Haiku ?
No need to do this right now, of course, especially if you're not running Haiku at the moment.
Thanks!

korli:
Well, there is no point in keeping "!x86" in SECONDARY_ARCHITECTURES. I'd argue that x86_gcc2 is also pointless: the x86 build should be fixed at some time in the future and left disabled for now.

fbrosson:
Wouldn't it be handy to keep x86_gcc2 in SECONDARY_ARCHITECTURES, while the PR is open, and only remove it:
* before the merge
* and after the project gets a fix for building on x86?

richienyhus:
Will fix #734 once merged.

korli:
@fbrosson you wrote that the runtime crashes. If so, this makes little sense to have it building on x86 hybrid too. Nonetheless this recipe can be merged as is :)

fbrosson:
Before my answer, here's a short summary of the situation:

* On x86_gcc2 (hybrid or not) the recipe builds but the runtime crashes.

* On x86, it starts building fine (using x86 primary arch) but fails when it tries to compile fairtrade.cpp:
```
g++ -c fairtrade.cpp -iquote./  -iquote./  -iquotegui/  -iquotegui/column/  -iquotegui/layout/  -iquotesqlite/  -isystemgui  -isystemgui/column  -isystemsqlite  -isystemgui/layout  -isystemConfigApp  -O3   -DDEBUG=1  -o "objects/fairtrade.o"
In file included from ./stock.h:14:0,
                 from gui/productsview.h:18,
                 from gui/mainwindow.h:17,
                 from fairtrade.h:12,
                 from fairtrade.cpp:6:
sqlite/SQLConnection.h:33:25: error: 'string' has not been declared
   virtual status_t Exec(string query);
                         ^
sqlite/SQLConnection.h:39:11: error: 'string' does not name a type
   virtual string GetLastError();
           ^
sqlite/SQLConnection.h:53:11: error: 'string' does not name a type
   virtual string GetValue(int row, int field);
           ^
sqlite/SQLConnection.h:61:11: error: 'string' does not name a type
   virtual string GetValue(int row, string fieldname);
```
So I guess this could be fiixed (but I don't have time to do that, unfortunately).

* On x86 hybrid, I can build fairtrade_x86_gcc2 but it crashes, just like it crashes on x86_gcc2 primary arch.

Now comes my answer ;-)

Those users who do not have a x86_gcc2 system but just a x86 hybrid will be able to more easily investigate why fairtrade_x86_gcc2 is crashing and why fairtrade (x86 primary arch) is not building.

BTW, do we really want to merge a recipe which does not build on most architectures and which crashes on those where it builds?

That said, I'm OK with having WIP recipes merged ;-)

korli:
We could mark the recipe broken for all arch then.
 The build problem seems to relate to a missing 'using namespace std'.

fbrosson:
I don't know what is best here.
My only wish is that we keep support for secondary architectures until a patch is written allowing to build fairtrade (and run it without crashing) on each primary arch.

BTW, would it be acceptable to use "?" (instead of "!") for x86 and x86_64, even if we know that the recipe is missing a patch to make the build succeed?

Anyway, I'll let you do what you think is best ;-)

scottmc:
>
>
> BTW, would it be acceptable to use "?" (instead of "!") for x86 and
> x86_64, even if we know that the recipe is missing a patch to make the
> build succeed?
>

I'd mark it with the ! as we know it needs work, but also open a matching
issue explaining what we know or suspect it needs to have done, so that
whoever picks it up with have a starting point.

dacianf:
@scottmc What do i have to do for this recipe?

richienyhus:
@dacianf You need to change 'x86_gcc2' to '!x86_gcc2'. Also you should try to [upstream](https://github.com/czeidler/fairtrade/pulls) your patch.

dacianf:
@richienyhus  I've made a pr for patch https://github.com/czeidler/fairtrade/pull/1 and nobody said anything.

korli:
I pushed my changes so the app works on x86_64. I didn't check the gcc2 build.
For me it's good enough to be merged, except for the secondary architecture.

dacianf:
@korli Is something that i can do?

korli:
@dacianf please build the app on x86_gcc2, and eventually test the package. Thanks!

korli:
I went ahead and merged as is. Thanks!

PR #994, opened 2017-01-07
beecrypt: added recipe

xdizzaster:
Added recipe for beecrypt crypto library.

### DIFF
--- a/dev-libs/beecrypt/beecrypt-4.2.1.recipe
+++ b/dev-libs/beecrypt/beecrypt-4.2.1.recipe
@@ -1,0 +1,78 @@
+SUMMARY="A strong and fast cryptography toolkit"
+DESCRIPTION="BeeCrypt is an ongoing project to provide a strong and fast \
+cryptography toolkit. It includes entropy sources, random generators, block \
+ciphers, hash functions, message authentication codes, multiprecision integer \
+routines, and public key primitives."
+HOMEPAGE="http://beecrypt.sourceforge.net/"
+COPYRIGHT="2009 Bob Deblier"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/beecrypt/beecrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="286f1f56080d1a6b1d024003a5fa2158f4ff82cae0c6829d3c476a4b5898c55d"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	beecrypt$secondaryArchSuffix = $portVersion
+	lib:libbeecrypt$secondaryArchSuffix = 7.0.0 compat >= 7
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	beecrypt${secondaryArchSuffix}_devel = $portVersion
+	devel:libbeecrypt$secondaryArchSuffix = 7.0.0 compat >= 7
+	"
+REQUIRES_devel="
+	beecrypt$secondaryArchSuffix == $portVersion base
+	devel:libgomp$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	aclocal
+	automake --add-missing
+	runConfigure ./configure --disable-threads --with-java=no --with-python=no
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	install -d $docDir
+	install -t $docDir BENCHMARKS BUGS CONTRIBUTORS NEWS README
+	rm $libDir/libbeecrypt.la
+
+	prepareInstalledDevelLib libbeecrypt
+	fixPkgconfig
+
+	packageEntries devel \
+		$docDir/BENCHMARKS \
+		$docDir/BUGS \
+		$docDir/NEWS \
+		$developDir
+}
+
+TEST()
+{
+	make check
+}


dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,63 @@
+SUMMARY="A strong and fast cryptography toolkit"
+DESCRIPTION="BeeCrypt is an ongoing project to provide a strong and fast \
+cryptography toolkit. It includes entropy sources, random generators, block \
+ciphers, hash functions, message authentication codes, multiprecision integer \
+routines, and public key primitives."
+HOMEPAGE="http://beecrypt.sourceforge.net"
```
fbrosson: Please add a trailing slash::
```
HOMEPAGE="http://beecrypt.sourceforge.net/"
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,63 @@
...
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/beecrypt/beecrypt-4.2.1.tar.gz"
+CHECKSUM_SHA256="286f1f56080d1a6b1d024003a5fa2158f4ff82cae0c6829d3c476a4b5898c55d"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86 x86_64"
+
+PROVIDES="
+	beecrypt$secondaryArchSuffix = $portVersion
+	lib:libbeecrypt$secondaryArchSuffix = 0.0.0 compat >= 0
```
fbrosson: If
```
package list -p /path/to/beecrypt-4.2.1-1-*.hpkg | grep lib/libbecrypt.so
```
yelds
```
lib/libbecrypt.so.7.0.0
```
(among other lines) then you should change this line of the recipe to:
```
	lib:libbeecrypt$secondaryArchSuffix = 7.0.0 compat >= 7
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,63 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	beecrypt${secondaryArchSuffix}_devel = $portVersion
+	devel:libbeecrypt$secondaryArchSuffix = 0.0.0 compat >= 0
```
fbrosson: Same as above, I _think_ it should be:
```
	devel:libbeecrypt$secondaryArchSuffix = 7.0.0 compat >= 7
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,63 @@
...
+	"
+REQUIRES_devel="
+	beecrypt$secondaryArchSuffix == $portVersion
```
fbrosson: Should be
```
	beecrypt$secondaryArchSuffix == $portVersion base
```
(The "base" keyword tells the beecrypt_devel sub-package that its _base_ package is beecrypt.)

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,63 @@
+SUMMARY="A strong and fast cryptography toolkit"
+DESCRIPTION="BeeCrypt is an ongoing project to provide a strong and fast \
+cryptography toolkit. It includes entropy sources, random generators, block \
+ciphers, hash functions, message authentication codes, multiprecision integer \
+routines, and public key primitives."
+HOMEPAGE="http://beecrypt.sourceforge.net"
+COPYRIGHT="2009 Bob Deblier"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/beecrypt/beecrypt-4.2.1.tar.gz"
```
fbrosson: Please change this to
```
SOURCE_URI="https://downloads.sourceforge.net/beecrypt/beecrypt-$portVersion.tar.gz"
```
(Well,  I know this project did not get new releases for a long time, but it could some day ;-)

dev-libs/beecrypt/beecrypt-4.2.1.recipe
xdizzaster: I might be doing something wrong, but for me it only yields `.PackageInfo` and no other file. I think this library might not provide a dynamic library to link against.

xdizzaster:
Updated a few things as @fbrosson suggested.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
xdizzaster: I tried compiling on my Linux machine and it seems that there is a `.so`, it just doesn't get packaged on Haiku.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,63 @@
...
+routines, and public key primitives."
+HOMEPAGE="http://beecrypt.sourceforge.net/"
+COPYRIGHT="2009 Bob Deblier"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/beecrypt/beecrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="286f1f56080d1a6b1d024003a5fa2158f4ff82cae0c6829d3c476a4b5898c55d"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86 x86_64"
```
fbrosson: Oops, I missed this one. It should be
```
```
Well, I did not test yet on x86_gcc2, so I don't know if we need to have "!x86_gcc2" or "x86_gcc2" (for both ARCHITECTURES and SECONDARY_ARCHITECTURES).

xdizzaster:
Changed the `SECONDARY_ARCHITECTURE` variable.

xdizzaster:
It seems that the make process does not generate shared libraries, although the tests pass just fine. I'm not sure what is going wrong or whether or not this is a problem at all.

fbrosson:
> I tried compiling on my Linux machine and it seems that there is a .so, it just doesn't get packaged on Haiku.

I guess you are right about the shared libs: the configure and/or the Makfile probably don't enable shared libs on Haiku.
This means you'll need to dig ;-)
I'll try to investigate this later.
(If other mentors see what the problem is they will certainly guide you.)

xdizzaster:
Updated the recipe to regenerate `configure` script. Turns out this was the problem; 8 year old scripts don't work too well on modern systems. Who would have thought?

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/beecrypt/beecrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="286f1f56080d1a6b1d024003a5fa2158f4ff82cae0c6829d3c476a4b5898c55d"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	beecrypt$secondaryArchSuffix = $portVersion
+	lib:libbeecrypt$secondaryArchSuffix = 7.0.0 compat >= 0
```
fbrosson: Should be "7.0.0 compat >= **7**" (not "7.0.0 compat >= **0**")

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	beecrypt${secondaryArchSuffix}_devel = $portVersion
+	devel:libbeecrypt$secondaryArchSuffix = 7.0.0 compat >= 0
```
fbrosson: Should be "7.0.0 compat >= **7**" (not "7.0.0 compat >= **0**")

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:awk
+	cmd:diff
```
fbrosson: You can remove
```
	cmd:awk
	cmd:diff
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize
+	cmd:make
+	cmd:m4
```
fbrosson: You can remove
```
	cmd:m4
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:awk
+	cmd:diff
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize
```
fbrosson: Should be
```
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:awk
+	cmd:diff
+	cmd:gcc$secondaryArchSuffix
```
fbrosson: I would recommend inserting
```
	cmd:g++$secondaryArchSuffix
```
just before
```
	cmd:gcc$secondaryArchSuffix
```
because there is a file named "cppglue.cxx".
BTW, it might be nice to find out how to generate the library libbeecrypt_cxx.so.7.0.0.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+	"
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	aclocal
+	autoheader
+	automake -a -c --foreign
+	autoconf
+	runConfigure ./configure --with-cplusplus=no --with-java=no --with-python=no
```
fbrosson: You can safely replace:
```
	autoheader
	automake -a -c --foreign
	autoconf
	runConfigure ./configure --with-cplusplus=no --with-java=no --with-python=no
```
by
```
	automake --add-missing
	runConfigure ./configure --disable-threads
```
I've checked that these tiny changes do not affect the resulting packages.

xdizzaster:
Added `cmd:libtoolize$secondaryArchSuffix`

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:awk
+	cmd:diff
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:m4
```
fbrosson: You can remove
```
	cmd:m4
```

dev-libs/beecrypt/beecrypt-4.2.1.recipe
xdizzaster: If I enable C++ in configure, the compilation fails with an error. It might be something with the ICU library which it needs or it might be something else, I haven't looked into it too much.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
xdizzaster: Missed that

dev-libs/beecrypt/beecrypt-4.2.1.recipe
xdizzaster: Are you sure? I just checked and it will not generate the shared library as it did before I added them.

xdizzaster: Oh, nevermind, I was tricked by GitHub into thinking you were telling me to remove the `libtoolize` and `aclocal` commands too.

fbrosson: Yep, something strange happened. I had included in my review a comment about fixing the cmd:libtoolize line, but while I was adding more comments to my review, the "Refresh" button appeared, and my comment about fixing the cmd:libtoolize line disappeared when I clicked refresh. But it was still there after I finished my review, so I deleted the comment afterwards. We might have hit a minor corner-case bug in GitHub ;-)
Well, GitHub rocks anyway!

dev-libs/beecrypt/beecrypt-4.2.1.recipe
fbrosson: You are right.
BTW, I get this during the configure:
```
checking for IBM's ICU library version >= 2.8... no
configure: WARNING: disabling cplusplus
```
(but it does not stop.)
I'll try adding icu_devel to BUILD_REQUIRES. If it works then I'll replace the icu_devel by the correct devel:lib\*.

xdizzaster:
Updated the recipe as @fbrosson suggested. I looked into the C++ thing, but the compilation fails because of some undefined typedefs. I have found them in `aclocal.m4`, but for some reason they don't get included in the sources I don't know how to get them there. If someone can figure it out, I'll fix it, otherwise I might leave C++ support off.

xdizzaster:
The close pull requrest button is way too close to the comment button. I have said it and I will say it again.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,72 @@
...
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	aclocal
+	automake --add-missing
+	runConfigure ./configure --disable-threads --with-cplusplus=no --with-java=no --with-python=no
```
fbrosson: I would recommend changing this to:
```
	runConfigure ./configure --disable-threads --with-java=no --with-python=no
```
If we don't then we need to split the line into 2 lines, otherwise we have a line that is > 80 chars long.
The other advantage is that anyone reading the output of the configure will have a hint about how to add support for C++.

xdizzaster: The recipe complains about missing ICU in the output, so they will have to edit the recipe and add that as well for C++, but alright.

xdizzaster:
Removed `--with-cplusplus=no` from `runConfigure ./configure`.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,72 @@
...
+	aclocal
+	automake --add-missing
+	runConfigure ./configure --disable-threads --with-java=no --with-python=no
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	rm $libDir/libbeecrypt.la
```
fbrosson: I think it would be nice to add these lines after the "make install":
```Bash
	install -d $docDir
	install -t $docDir BENCHMARKS BUGS CONTRIBUTORS NEWS README

```
and lthen move some of these files to the _devel package, as shown below.

dev-libs/beecrypt/beecrypt-4.2.1.recipe
```diff
@@ -0,0 +1,72 @@
...
+
+	prepareInstalledDevelLib libbeecrypt
+	fixPkgconfig
+
+	packageEntries devel $developDir
```
fbrosson: Can you please replace
```
```
by
```
	packageEntries devel \
		$docDir/BENCHMARKS \
		$docDir/BUGS \
		$docDir/NEWS \
		$developDir
```

xdizzaster:
Added some more files in `$docDir` as @fbrosson suggested.

PR #998, opened 2017-01-07
paperkey: added recipe

xdizzaster:
Added recipe for Paperkey secret key data extraction tool.

### DIFF
--- a/app-crypt/paperkey/paperkey-1.4.recipe
+++ b/app-crypt/paperkey/paperkey-1.4.recipe
@@ -1,0 +1,66 @@
+SUMMARY="A tool to backup/restore OpenPGP keys to/from printed paper"
+DESCRIPTION="Paperkey extracts the secret bits from OpenPGP keys, to be printed \
+on paper as a form of backup and allows it to be reintroduced to reconstruct the \
+secret key if needed."
+HOMEPAGE="http://www.jabberwocky.com/software/paperkey/"
+COPYRIGHT="2007-2016 David Shaw"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://www.jabberwocky.com/software/paperkey/paperkey-$portVersion.tar.gz"
+CHECKSUM_SHA256="e12bb0ec835127d12a922a8d60b3dfdb3ca8ee60bb5b4d15ae4cea85bbcf336f"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	paperkey$secondaryArchSuffix = $portVersion
+	cmd:paperkey$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	REQUIRES="$REQUIRES
+		lib:libcrypto$secondaryArchSuffix
+		"
+fi
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	BUILD_REQUIRES="$BUILD_REQUIRES
+		devel:libcrypto$secondaryArchSuffix
+		"
+fi
+
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+		runConfigure ./configure --with-openssl
+		make $jobArgs LIBS="-lcrypto"
+	else
+		runConfigure ./configure
+		make $jobArgs
+	fi
+}
+
+INSTALL()
+{
+	make install
+	install -d $docDir
+	install -t $docDir AUTHORS NEWS README
+}
+
+TEST()
+{
+	make check
+}


app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,48 @@
...
+BUILD()
+{
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
```
fbrosson: Can you please add these lines after the "make install" ?
```
	install -d $docDir
	install -t $docDir AUTHORS NEWS README
```

app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,48 @@
...
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure ./configure
```
fbrosson: It might be worth calling configure with `--with-openssl `.

This will require a few other changes in:
* REQUIRES: add lib:libssl and/or lib:libcrypto
* BUILD_REQUIRES: add deve:libssl and/or devel:libcrypto

To find out which are needed, you can start by adding just openssl${secondaryArchSuffix}_devel to BUILD_REQUIRES and then use haikuporter -S to find which dependencies are missing in REQUIRES.

fbrosson: It looks like beecrypt only needs libcrypto. This means we just need to add:
* in REQUIRES:
```
```
* and in BUILD_REQUIRES:
```
	devel:libcrypto$secondaryArchSuffix
```

But you also need to change the call to make in BUILD to:
```
	make $jobArgs LIBS="-lcrypto"
```

xdizzaster:
Added some files to `$docDir` as @fbrosson suggested.

app-crypt/paperkey/paperkey-1.4.recipe
xdizzaster: Unfortunately, if I enable OpenSSL, the build fails because of some weird undefined references that won't go away even if I include the header file right before them.

xdizzaster: Oh, I didn't think about adding `-lcrypto`, will try

xdizzaster:
Added compiling with OpenSSL, as @fbrosson suggested.

app-crypt/paperkey/paperkey-1.4.recipe
fbrosson: Hmm, it looks like paperkey won't build with x86_gcc2 if we pass --with-openssl.
Here's what I get:
```
DEPDIR=.deps depmode=gcc /bin/sh ../depcomp \
gcc -DHAVE_CONFIG_H -I. -I..     -g -O2 -c sha1.c
In file included from /sources/paperkey-1.4/gl/sha1.h:37,
                 from /sources/paperkey-1.4/gl/sha1.c:29:
/sources/paperkey-1.4/gl/gl_openssl.h:98: macro `OPENSSL_FN' used without args
make[4]: *** [Makefile:539: sha1.o] Error 1
```
I've prepared a workaround that I'll put in a review.

app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,52 @@
...
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	paperkey$secondaryArchSuffix = $portVersion
+	cmd:paperkey$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libcrypto$secondaryArchSuffix
```
fbrosson: You can define REQUIRES with:
```
REQUIRES="
	haiku$secondaryArchSuffix
	"
	REQUIRES="$REQUIRES
		"
fi
```

app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,52 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libcrypto$secondaryArchSuffix
```
fbrosson: You can define BUILD_REQUIRES with:
```
BUILD_REQUIRES="
	"
	BUILD_REQUIRES="$BUILD_REQUIRES
		devel:libcrypto$secondaryArchSuffix
		"
fi
```

app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,52 @@
...
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure ./configure --with-openssl
+	make $jobArgs LIBS="-lcrypto"
```
fbrosson: You cand define BUILD with:
```
BUILD()
{
		runConfigure ./configure --with-openssl
		make $jobArgs LIBS="-lcrypto"
	else
		runConfigure ./configure
		make $jobArgs
	fi
}
```

xdizzaster:
Added `if` clause when building on x86_gcc2 to disable OpenSSL.

app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,65 @@
+SUMMARY="Tool to extract secret bits from a secret key to be printed on paper"
```
fbrosson: Sorry, I just noticed that this line is > 70 chars, so it breaks the [recommendations for SUMMARY](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#summary).
Maybe we could use this:
```
SUMMARY="A tool to backup/restore OpenPGP keys to/from printed paper"
```
If you have something better or with "secret bits", please feel free to suggest ;-)

app-crypt/paperkey/paperkey-1.4.recipe
```diff
@@ -0,0 +1,65 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	BUILD_REQUIRES="$BUILD_REQUIRES
+		devel:libcrypto$secondaryArchSuffix
+		"
+fi
+BUILD_PREREQUIRES="
```
fbrosson: Shall we add a blank line before BUILD_PREREQUIRES ?
There isn't any official rule for it, but we usually have one such blank line between the last line of BUILD_REQUIRES and BUILD_PREREQUIRES **_when_** the last part of BUILD_REQUIRES is a conditional block (i.e. an if then .. fi).
(Sorry, I forgot to mention this in my previous review.)

app-crypt/paperkey/paperkey-1.4.recipe
xdizzaster: No, I can't come up with anything better, but I think this is fine.

xdizzaster:
Cosmetic touches, as @fbrosson suggested.

PR #1005, opened 2017-01-08
irssi: bump version, removed old recipes and patchfiles.

miqlas:
Enabled x86_64 arch, enabled dynamic perl support.


net-irc/irssi/irssi-1.0.0.recipe
```diff
@@ -54,13 +52,18 @@ GLOBAL_WRITABLE_FILES="
 
 BUILD()
 {
+	export CFLAGS="-D_BSD_SOURCE"
+	export LDFLAGS="-lnetwork -lbsd"
```
korli: -lnetwork should be unnecessary (should be detected in configure).  What's needed from libbsd?

mmuman:
Indeed I [upstreamed](https://github.com/irssi/irssi/commit/0d7b3877e46cda8e960cab30397f15003e9c7527) the libnetwork patch long ago.
There's probably an optional call in libbsd.

miqlas:
Fixed, thanks!

mmuman:
I just tested and I saw no difference between the generated irssi-config.h with and without -lbsd and the define. The configure output doesn't change either, it doesn't seem to need anything from it.

net-irc/irssi/irssi-1.0.0.recipe
```diff
@@ -7,31 +7,29 @@ HOMEPAGE="https://irssi.org/"
...
+CHECKSUM_SHA256="790eb3b62d8889d1fb3973845d8ca5d7c2578620632abc671ffdff3da13aa487"
+SOURCE_DIR="irssi-$portVersion"
 
-ARCHITECTURES="x86_gcc2 x86"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
 
 PROVIDES="
-  	irssi$secondaryArchSuffix = $portVersion
-  	cmd:irssi$secondaryArchSuffix
+	irssi$secondaryArchSuffix = $portVersion
```
fbrosson: Since we don't have any SECONDARY_ARCHITECTURES we should drop every occurrence of $secondaryArchSuffix and ${secondaryArchSuffix}.

BTW, It's funny that nobody (including myself) has spotted this until now ;-)

BTW, you can also drop `cmd:chmod` from BUILD_PREREQUIRES, as chmod is in coreutils:
```
package list -p /system/packages/coreutils-*-*.hpkg | grep ^bin/chmod
```

miqlas: Already fixed.

PR #1007, opened 2017-01-08
libopenmpt: added recipe

xdizzaster:
Added recipe for libopenmpt module player library.

### DIFF
--- a/media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
+++ b/media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
@@ -1,0 +1,96 @@
+SUMMARY="OpenMPT based module player library and command-line player"
+DESCRIPTION="libopenmpt is a cross platform C++ and C libary to decode tracked \
+music files (modules) int a raw PCM audio stream. openmpt123 is a \
+cross-platform command line or terminal based module file player."
+HOMEPAGE="https://lib.openmpt.org/libopenmpt/"
+COPYRIGHT="2004-2016 OpenMPT contributors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://lib.openmpt.org/files/libopenmpt/src/libopenmpt-${portVersion/\~/-}-autotools.tar.gz"
+CHECKSUM_SHA256="a6a7e6da1ae66e1cf46985ee92c182e50652d71b96135e9fa6048e132d844753"
+SOURCE_DIR="libopenmpt-${portVersion%~*}-autotools"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	libopenmpt$secondaryArchSuffix = $portVersion
+	lib:libopenmpt$secondaryArchSuffix = 0.0.20 compat >= 0
+	cmd:openmpt123 = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libFLAC$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libmpg123$secondaryArchSuffix
+	lib:libogg$secondaryArchSuffix
+	lib:libsndfile$secondaryArchSuffix
+	lib:libvorbis$secondaryArchSuffix
+	lib:libvorbisfile$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libopenmpt${secondaryArchSuffix}_devel
+	devel:libopenmpt$secondaryArchSuffix = 0.0.20 compat >= 0
+	"
+REQUIRES_devel="
+	libopenmpt$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libFLAC$secondaryArchSuffix
+	devel:libltdl$secondaryArchSuffix
+	devel:libmpg123$secondaryArchSuffix
+	devel:libogg$secondaryArchSuffix
+	devel:libsndfile$secondaryArchSuffix
+	devel:libvorbis$secondaryArchSuffix
+	devel:libvorbisfile$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	# portaudio and portaudiocpp disabled as they are untested on all arches
+	runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir \
+		--without-pulseaudio \
+		--without-portaudio \
+		--without-portaudiocpp \
+		--disable-doxygen-doc
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	rm $libDir/libopenmpt.la
+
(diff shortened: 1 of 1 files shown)


media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+DESCRIPTION="libopenmpt is a cross platform C++ and C libary to decode tracked \
+music files (modules) int a raw PCM audio stream. openmpt123 is a \
+cross-platform command line or terminal based module file player."
+HOMEPAGE="https://lib.openmpt.org/libopenmpt/"
+COPYRIGHT="2004-2016 OpenMPT contributors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://lib.openmpt.org/files/libopenmpt/src/libopenmpt-${portVersion/\~/-}-autotools.tar.gz"
+CHECKSUM_SHA256="a6a7e6da1ae66e1cf46985ee92c182e50652d71b96135e9fa6048e132d844753"
+SOURCE_FILENAME="libopenmpt-${portVersion/\~/-}.tar.gz"
```
fbrosson: It would be better to remove the definition of SOURCE_FILENAME.
The original name is OK, and it makes it clear that we are using the autotools-based tarball.

media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+SOURCE_DIR="libopenmpt-0.2.7386-autotools"
```
fbrosson: You can use:
```
SOURCE_DIR="libopenmpt-${portVersion%~*}-autotools"
```
(handy, isn't it?)

media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	libopenmpt$secondaryArchSuffix = $portVersion
+	lib:libopenmpt$secondaryArchSuffix = 0.0.20
```
fbrosson: I think you should use a compat version:
```
	lib:libopenmpt$secondaryArchSuffix = 0.0.20 compat >= 0
```

media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+	lib:libogg$secondaryArchSuffix
+	lib:libvorbis$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libsndfile$secondaryArchSuffix
+	lib:libFLAC$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libopenmpt${secondaryArchSuffix}_devel
+	devel:libopenmpt$secondaryArchSuffix = 0.0.20
```
fbrosson: I think you should use a compat version:
```
	devel:libopenmpt$secondaryArchSuffix = 0.0.20 compat >= 0
```

Hmm, you forgot to put a REQUIRES_devel ;-)

And in BUILD_PREREQUIRES I think you should insert:
```
	cmd:aclocal
	cmd:autoconf
	cmd:autoheader
	cmd:automake
```
(sorted alphabeticallly with what is already in there).

xdizzaster:
Added the suggestions made by @fbrosson above.

xdizzaster:
Marked `x86_gcc2` arch as broken since the package requires at least GCC 4.1.

media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,87 @@
...
+CHECKSUM_SHA256="a6a7e6da1ae66e1cf46985ee92c182e50652d71b96135e9fa6048e132d844753"
+SOURCE_DIR="libopenmpt-${portVersion%~*}-autotools"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	libopenmpt$secondaryArchSuffix = $portVersion
+	lib:libopenmpt$secondaryArchSuffix = 0.0.20 compat >= 0
+	cmd:openmpt123$secondaryArchSuffix = $portVersion
```
fbrosson: > Marked x86_gcc2 arch as broken since the package requires at least GCC 4.1.

OK. This means we can use the commandBinDir "trick" (like you did for app-text/unrtf/unrtf-0.21.9.recipe):
You need to drop the $secondaryArchSuffix after cmd:openmpt123, insert the conditional definition for commandBinDir and change the runConfigure command (as you did for unrtf.)

media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,87 @@
+SUMMARY="OpenMPT based module player libary and command-line player"
```
fbrosson: Small typo: libary → library

xdizzaster:
Fixed a typo and changed the binary directory as @fbrosson suggested.

media-sound/libopenmpt/libopenmpt-0.2.7386~beta20.3.recipe
```diff
@@ -0,0 +1,94 @@
...
+	devel:libvorbis$secondaryArchSuffix
+	devel:libltdl$secondaryArchSuffix
+	devel:libsndfile$secondaryArchSuffix
+	devel:libFLAC$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
```
korli: are autoconf/automake needed?

fbrosson: I guess the recipe will also build without these, but since we are using the autotools-based tarball, I thought it could be a good idea to have them in the recipe.
This can be handy if someone want's to manually patch configure.ac and/or Makefile.am and then continue the build.
But if you think it is better to remove them that's OK for me.
Actually, I though it was good practice to add them because, if I'm not mistaken, configure looks for them even if it does not run them when the configure script is newer than configure.ac (and Makefile is newer than Makefile.am).

xdizzaster: @fbrosson is correct, the configure script checks for the autotools even if they are not needed and only complains about them if they are needed. I think there's no harm in leaving them there.

fbrosson: Thanks @xdizzaster for the confirmation.
I guess @korli's question was more:
_"Why not drop the dependency on autoconf and automake since we know that the configure script will still work without them?"_

As I wrote above, I think that keeping the dependencies on the autotools has more advantages than disadvantages.
So I share @xdizzaster's conclusion.
(BTW, we already have many recipes which have these dependencies and yet do not explicitely call any of the autotools.)

fbrosson: I have 2 questions regarding the REQUIRES and BUILD_REQUIRES:
1) When I sort and merge the outputs of
```
objdump -x /lib/x86/libopenmpt.so.0.0.20  | grep NEEDED
```
or
```
objdump -x /bin/openmpt123 | grep NEEDED 
```
I find that the current REQUIRES (and BUILD_REQUIRES) are only missing one lib: lib:libvorbisfile. But this one is provided by the same package which provides lib:libvorbis.
So my first question is: should'nt we be also adding a lib:libvorbisfile$secondaryArchSuffix to REQUIRES and a devel:libvorbisfile$secondaryArchSuffix to BUILD_REQUIRES ?

2)  I noticed that the order of these lines in REQUIRES (and BUILD_REQUIRES) is the same as the order of these libs when we use the objdump command.
So my second question is: is it OK to keep the current order? Or shouldn't we sort them alphabetically, like we (try) to do for all our recipes?

I guess the right answers are: "add the libvorbisfile" & "sort alphabetically".
@korli, am I correct?
Thanks!

xdizzaster:
Added `lib:libvorbisfile` as a dependency and sorted the library deps alphabetically.

PR #1009, opened 2017-01-09
libass: bump to 0.13.6, add TEST().

fbrosson:


fbrosson:
handbrake, mpv, qmp and qtav all depend on lib:libass.
Should I bump REVISION for these?
If yes, should I do so in this PR or a separate commit?
Thanks!

korli:
Yes, you can bump revision on a separate commit if all build correctly with the new version. 
That said, I see that ArchLinux hasn't yet picked up this upgrade. Maybe temporize a bit to avoid being beta tester? 
Anyway, I don't plan to sync x86_64 packages on the HEAD anymore, but for now on the 2017Q1 branch (mostly build fixes and security updates).

fbrosson:
OK, thanks for the info.
I'll keep this PR open until I have both:
* managed to build all these packages against the new version of libass,
* prepared a separate _"bump revision for handbrake, mpv, qmp and qta"_ that I'll push right after the merge, **_if_** I can merge this PR, that is, **_if_** all packages build fine against the new libass.

fbrosson:
Thanks @waddlesplash for the approval.
I haven't built any of the packages _|handbrake, mpv, qmp and qta]_ which depend on libass against this new release of libass, but I can prepare a separate PR to bump the revision of these recipes, in case someone has tested al these packages and wishes to merge this PR.
Sorry, btw, because I though I would have had the time to test all these packages.
(The only package which I have tested is libass itself.)

PR #1010, opened 2017-01-09
iec16022: added recipe

xdizzaster:
Added recipe for iec16022 DataMatrix generation tool and library.

### DIFF
--- a/media-gfx/iec16022/iec16022-0.2.7.recipe
+++ b/media-gfx/iec16022/iec16022-0.2.7.recipe
@@ -1,0 +1,81 @@
+SUMMARY="2D barcode/DataMatrix generator"
+DESCRIPTION="With iec16022 you can produce 2D barcodes, also known as \
+DataMatrix, as defined in ISO/IEC 16022."
+HOMEPAGE="https://datenfreihafen.org/projects/iec16022.html"
+COPYRIGHT="2006 Stefan Schmidt"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/rdoeffinger/iec16022/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="34f00da5f6c83dd14b2803e5ed0f5310310e452023fe989c59fc5489b78cd33a"
+SOURCE_FILENAME="iec16022-$portVersion.tar.gz"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	iec16022$secondaryArchSuffix = $portVersion
+	lib:libiec16022$secondaryArchSuffix = 0.1.0 compat >= 0
+	cmd:iec16022 = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libpopt$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	iec16022${secondaryArchSuffix}_devel
+	devel:libiec16022$secondaryArchSuffix = 0.1.0 compat >= 0
+	"
+REQUIRES_devel="
+	iec16022$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libpopt$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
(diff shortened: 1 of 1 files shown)


media-gfx/iec16022/iec16022-0.2.7.recipe
```diff
@@ -0,0 +1,82 @@
...
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandBinDir=$binDir
+else
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	iec16022$secondaryArchSuffix = $portVersion
+	lib:libiec16022$secondaryArchSuffix = 0.1.0 compat >= 0
+	cmd:iec16022$secondaryArchSuffix = $portVersion
```
fbrosson: Should be
```
	cmd:iec16022 = $portVersion
```

Remember, we are using the commandBinDir "trick", so we should not expect to have the bin/iec16022-x86 symbolic link to bin/x86/iec16022 but just one runtime: bin/iec16022.

media-gfx/iec16022/iec16022-0.2.7.recipe
```diff
@@ -0,0 +1,82 @@
+SUMMARY="2D barcode/DataMatrix generator"
+DESCRIPTION="With iec16022 you can produce 2D barcodes, also known as \
+DataMatrix, as defines in ISO/IEC 16022."
```
fbrosson: typo?
"as define**s** in" → "as define**d** in"

media-gfx/iec16022/iec16022-0.2.7.recipe
```diff
@@ -0,0 +1,82 @@
...
+	"
+
+BUILD()
+{
+	aclocal
+	libtoolize --force
+	autoheader
+	automake --add-missing
+	autoconf configure.ac > configure
+	chmod +x configure
```
fbrosson: I think we can replace
```
	aclocal
	libtoolize --force
	autoheader
	automake --add-missing
	autoconf configure.ac > configure
	chmod +x configure
```
by something better, although this is what upstream has put in autogen.sh ;-)

media-gfx/iec16022/iec16022-0.2.7.recipe
xdizzaster: Yep, it's a typo

media-gfx/iec16022/iec16022-0.2.7.recipe
xdizzaster: If that is how the author recommends building the program, should we change it?

xdizzaster:
Fixed a typo.

media-gfx/iec16022/iec16022-0.2.7.recipe
fbrosson: I just found that this block works just as fine as what the author uses.
```
	aclocal
	autoheader
	autoconf
	automake --add-missing
```
Isn't it much better than what we have in autogen.sh ?

xdizzaster: I guess so.

fbrosson: This also works:
```
	aclocal
	libtoolize --force
	autoheader
	automake --add-missing
	autoconf
```
And it is almost what the author recommends.
The only difference is that we got rid of the ugly part of the autogen.sh ;-)

xdizzaster:
Updated the `BUILD()` function as @fbrosson suggested.

media-gfx/iec16022/iec16022-0.2.7.recipe
```diff
@@ -0,0 +1,81 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libpopt$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	iec16022${secondaryArchSuffix}_devel
+	devel:libiec16022$secondaryArchSuffix = 0.1.0 compat >=0
```
korli: Please add a space after >=

korli:
One format to fix, otherwise good.

media-gfx/iec16022/iec16022-0.2.7.recipe
fbrosson: Good catch!

@xdizzaster, before you push the fix with the space, if you would like to replace
```
	aclocal
```
by
```
	aclocal
	libtoolize --force
```
and
```
	autoconf
	automake --add-missing
```
by
```
	automake --add-missing
	autoconf
```
please do not hesitate.
I know it won't make much difference. The resulting packages are even identical.
It's just that it will be closer to what the author suggests in autogen.sh.
After all, he is the author of a great utility, and I feel a little bit guilty of having asked you to put the block in a different order than that recommended by the author.

@korli, do you think these cosmetic changes are welcome?

Thanks!

xdizzaster:
Added a missing space and reshuffled the commands in `BUILD()`.

PR #1020, opened 2017-01-09
icoutils: added recipe

xdizzaster:
Added recipe for icoutils icon and cursor file convertor.

### DIFF
--- a/media-gfx/icoutils/icoutils-0.31.1.recipe
+++ b/media-gfx/icoutils/icoutils-0.31.1.recipe
@@ -1,0 +1,63 @@
+# FIXME: extresso and genresscript won't work because of missing libwww-perl
+# and that won't compile because reasons
+SUMMARY="Tools for extracting and converting Windows icon and cursor files"
+DESCRIPTION="A set of command-line programs for extracting and converting \
+images in Microsoft Windows icon and cursor files (usually with the extension \
+.ico or .cur, but they can also be embedded in executables and DLL libraries."
+HOMEPAGE="http://nongnu.org/icoutils/"
+COPYRIGHT="1998-2017 Frank Richter"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="http://savannah.nongnu.org/download/icoutils/icoutils-$portVersion.tar.bz2"
+CHECKSUM_SHA256="26e29d3c78f25d4cdf402501ac0414c51a9a092daebf6c9dee3b837dee693093"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	icoutils = $portVersion
+	cmd:extresso = $portVersion
+	cmd:genresscript = $portVersion
+	cmd:icotool = $portVersion
+	cmd:wrestool = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libpng16
+	lib:libz
+	cmd:perl
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libpng16
+	devel:libz
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:gcc
+	cmd:ld
+	cmd:make
+	"
+
+PATCH()
+{
+	sed -i -e "1 s|/usr/bin/perl|/bin/env perl|" \
+		extresso/extresso.in \
+		extresso/genresscript.in
+}
+
+BUILD()
+{
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
+
+TEST()
+{
+	make check
+}


media-gfx/icoutils/icoutils-0.31.1.recipe
```diff
@@ -0,0 +1,73 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="http://savannah.nongnu.org/download/icoutils/icoutils-$portVersion.tar.bz2"
+CHECKSUM_SHA256="26e29d3c78f25d4cdf402501ac0414c51a9a092daebf6c9dee3b837dee693093"
+PATCHES="icoutils-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	icoutil$secondaryArchSuffix = $portVersion
```
fbrosson: If the recipes builds for x86_gcc2, x86 and x86_64, then you don't need any $secondaryArchSuffix.
This is because this package does not provide any library.

If, on the other side, t does not build on one (or more) of these primary architectures, then you should keep the $secondaryArchSuffix and add a SECONDARY_ARCHITECTURES with the appropriate archies, andfix ARCHITECTURES (with a ! where appropriate.)

Regarding the libpng16 vs libpng choice, this is a bit strange because both build fine on all architectures.
I think the problem is somewhere else.
I'll see if I can build it with libpng16 on x86_gcc2.

All the rest looks OK at first glance ;-)

xdizzaster: It builds on x86_gcc2 with libpng, it's just that at the end, before compressing the package, haikuporter complains about libpng16 for some reason, so I added that to fix it.

xdizzaster:
Removed `$secondaryArchSuffix` from binaries.

media-gfx/icoutils/icoutils-0.31.1.recipe
fbrosson: Can you retry, after building the latest release of libpng16 ?

xdizzaster: I'm not so sure about the `PATCH()` function. I already generated a patchfile and the scripts don't work anyway because of missing LWP, so I don't think the way we change the hashbang matters all that much.

xdizzaster: If I build the latest libpng16, it works without throwing any error. For reference, this is the error I was getting, it wasn't anything compile related:

virtualbox_haiku x86 gcc 2 hybrid_09_01_2017_20_58_43

xdizzaster:
Removed the `if`'s and replaced them with `lib:libpng` on all arches.

media-gfx/icoutils/icoutils-0.31.1.recipe
fbrosson: When I get these "No such file or directory:  ... /work-xxx" then I just remove the work-* directories.

Regarding the PATCH function, I do agree that it does not make any difference at the end.
But I tend to prefer a simple PATCH function.
BTW, if you had a patchset and the remove it, you need to remove the work-* directory.

media-gfx/icoutils/icoutils-0.31.1.recipe
```diff
@@ -0,0 +1,57 @@
...
+	icoutils = $portVersion
+	cmd:extresso = $portVersion
+	cmd:genresscript = $portVersion
+	cmd:icotool = $portVersion
+	cmd:wrestool = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
```
fbrosson: I would recommend using lingpng16 instead of libpng.
(Debian builds icoutils against the 1.6.x branch of libpng, btw)

media-gfx/icoutils/icoutils-0.31.1.recipe
xdizzaster: It's not the work directory, it's the warning in blue that causes the problem. It gave me this error on a fresh work directory.

xdizzaster:
Changed libpng to libpng16

fbrosson:
Regarding the PATCH function, it would be nice to use it to replace the patchsets.
These patchsets are recommended when the changes occur at places that can change. Here we know that these changes will always be on the very first line. So a PATCH function is the best choice for this, imho.

xdizzaster:
Removed the patchfile, added a `PATCH()` function.

media-gfx/icoutils/icoutils-0.31.1.recipe
```diff
@@ -0,0 +1,63 @@
...
+
+PROVIDES="
+	icoutils = $portVersion
+	cmd:extresso = $portVersion
+	cmd:genresscript = $portVersion
+	cmd:icotool = $portVersion
+	cmd:wrestool = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
```
fbrosson: I've just checked that the recipe builds fine on x86_gcc2.
I had already checked that it also builds fine on x86.

So this means we ~~can~~ should get rid of every occurence of $secondaryArchSuffix and ${secondaryArchSuffix}.

This is great news, isn't it ?

Hmm, if you could sort alphabetically the lib:lib\* lines in REQUIRES and the devel:lib\* lines in BUILD_REQUIRES that would be great!
Thanks!

xdizzaster:
Removed every instance of `$secondaryArchSuffix` since the package builds just fine on x86_gcc2.

korli:
Please open an issue for the identified problems. We can then merge as is.

xdizzaster:
Opened https://github.com/haikuports/haikuports/issues/1031. Hope it is alright, it's my first issue.

korli:
Thanks!

PR #1032, opened 2017-01-10
Recipe and patch for AMC

dacianf:
Build fails in makefile on line 60 recipe for target amc.o failed

### DIFF
--- a/media-sound/amc/amc-1.14.0.recipe
+++ b/media-sound/amc/amc-1.14.0.recipe
@@ -1,0 +1,47 @@
+SUMMARY="A'dam Music Composer"
+DESCRIPTION="This is a great app for creating music with, it supports \
+keyboard and it creates MP3s and WAVE files"
+HOMEPAGE="https://github.com/HaikuArchives/AMC"
+COPYRIGHT="1989 Atsushi Moriwaki"
+LICENSE="Public Domain"
+REVISION="1"
+srcGitRev="451087418bc91f626ba7cb1b112816e3bac898f9"
+SOURCE_URI="https://github.com/HaikuArchives/AMC/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="8e90759b128c66f3a3443fecf0b9ffda112a911be77fc14d7666d7bb051b98f7"
+SOURCE_DIR="AMC-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2"
+
+PROVIDES="
+	amc = $portVersion
+	app:AMC = $portVersion
+	"
+
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:make
+	cmd:gcc
+	"
+
+BUILD()
+{
+	cd amc/src
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd amc/src
+	mkdir -p $appsDir/AMC
+	cp -a amc $appsDir/AMC/A\'dams\ Music\ Composer
+	cp -r ../doc $appsDir/AMC
+	cp -r ../tunes $appsDir/AMC
+}


fbrosson:
```
c++  -c -O -fno-default-inline amc.cpp
In file included from /sources/AMC-.../amc/src/amc.cpp:30:
/sources/AMC-.../amc/src/midi-keyb.h:1: usb/USBKit.h: No such file or directory
```
We have a USB**KIT**.h , but it is in os/device/: 
[headers/os/device/USBKit.h](https://github.com/haiku/haiku/blob/master/headers/os/device/USBKit.h)
I don't know, however, it it would satisfy AMC.

korli:
That should be it: just drop the "usb/" prefix.

dacianf:
where should i make those changes?

humdingerb:
I would suggest, you fork and clone the amc repo from HaikuArchives and make the changes there. Then create a PR when it works. No use to create a patch with the recipe, when we can change the source directly.

You have to keep "making" and solve one issue that pops up after another. E.g. it's BUSBDevice, BUSBRoster ...etc. now, instead of <sys/dir.h> you have to include <dirent.h> etc.

korli:
@humdingerb to be honest, only the recipe and patch are part of the GCI task. The fact that AMC doesn't even build on Haiku is already an unexpected issue IMO.

in amc/src/midi-keyb.h: replace #include "usb/USBKit.h" by #include "USBKit.h"

humdingerb:
We could create a task to "Fix build of AMC" first. Would you like that, @dacianf ?
Then you at least got 2 tasks out of it. :)

dacianf:
@humdingerb  we can do this, but i cannot take 2 tasks at the same time...

humdingerb:
I'd say keep working on getting it to build and make a PR to HaikuArchives. When that's done, we close the recipe task and you take the "make-it-work" task to update the recipe (which should be very easy). Kinda switch the actual task you're doing...

dacianf:
@humdingerb Ok

fbrosson:
Since the current task is almost completed, it would be better, imho, to:
1. create a "Fix build of AMC" task.
2. validate the current "Create a recipe for A'dam Music Composer" task.
3. publish the "Fix build of AMC" task.

dacianf:
@fbrosson what do you say about line 18? If i let it there without comment line build does't work.

fbrosson:
I'm quite sure that what is wrong in
```
	app:A\'dams\ Music\ Composer = $portVersion
```
is the quote and the white spaces.

dacianf:
@fbrosson now it works

media-sound/amc/amc-1.14.0.recipe
```diff
@@ -0,0 +1,48 @@
+SUMMARY="A'dam Music Composer"
+DESCRIPTION="This Mini-Scheme Interpreter is based on SCHEME Interpreter in \
+Common Lisp in Appendix of T.Matsuda & K.Saigo, Programming of LISP, \
+archive No5 (1987) p6 - p42 (published in Japan)."
```
fbrosson: I guess this description is from another recipe ;-)

media-sound/amc/amc-1.14.0.recipe
```diff
@@ -0,0 +1,48 @@
...
+{
+	cd amc/src
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd amc/src
+	mkdir -p $appsDir/AMC
+	cp -a amc $appsDir/AMC/A\'dams\ Music\ Composer
```
fbrosson: There was a discussion a long time ago about what should go to the $appsDir.
My opinion was/is that we should not put subdirectories in $appsDir, but only runtimes.
The rarionale for this is that $appsDir is in $PATH, so putting any app in a subdirectory of $appsDir means that you cannot launch that app from the command line (unless you give its full path.)
The problem with my approach is that some apps require that their data reside on the same directory as the runtime itself, but enforcing this would lead to a crowded $appsDir.
The ideal solution would be to put the data in $dataDir/AMC/ and the runtime in $appsDir, using AMC for the name of the runtime. But I don't know if this works for AMC. (I did not try to build it yet.)
I had given up trying to convince the other contributors about this, and I don't plan to try it again.

I just started writing this because I think the runtime should be called AMC, not "A'dams Music Composer", and here am I, bringing back an old topic I don't really care about.
So please feel free to ignore this message.

humdingerb: I don't want to open that can of worms, either. :)
So, not talking in general, but only for this case, here's my rational for putting the docs and tunes with the binary into a subfolder in $appsDir: users easily find the app in its folder in $appsDir and thereby also find its docs and the example tunes.  Since there's no help menu in the app itself, the users otherwise probably won't know that AMC comes with docs and example tunes and where to look for them.

I assumed the declaration of "app:" under  PROVIDES had to be the exact name of the app under that it will be installed. Guess that was wrong. :)
To me the app's filename should be the name that appears in the app's window title. In this case it's "A'dams Music Composer".
Hmm... I stand corrected, here it's "the A'dams Music Composer". Man, this app is sucha mess... I dare you to look at its source... :)

dacianf:
@humdingerb  or @fbrosson  what is wrong? what do i have to change?

fbrosson:
DESCRIPTION ;-)

dacianf:
@fbrosson Isn't it ok now?

scottmc:
What's the status on this one?  Does it build ok on gcc2?  Can it be merged as is and then be improved on later?

dacianf:
It builds fine, you can build it if you want, to be sure it's ok :).

PR #1030, opened 2017-01-10
beohms: added recipe

xdizzaster:
Added recipe for BeOhms Ohm's Law calculator.

PR #1034, opened 2017-01-11
bebuilder: added recipe

xdizzaster:
Added recipe for BeBuilder BeOS GUI builder. 

Since the source code was not building properly because some parts of it were not standard compliant (using `string` instead of `std::string`, some missing libraries in the makefile, etc.), I had to write a patch to fix this. I think it would be best if I opened a PR on the BeBuilder repo and submit the patch there, like I did in #1030.

### DIFF
--- a/haiku-apps/bebuilder/bebuilder-0.5.2~git.recipe
+++ b/haiku-apps/bebuilder/bebuilder-0.5.2~git.recipe
@@ -1,0 +1,46 @@
+SUMMARY="GUI designer for BeOS"
+DESCRIPTION="BeBuilder is a GUI designer for BeOS."
+HOMEPAGE="https://github.com/HaikuArchives/BeBuilder"
+COPYRIGHT="1999 Jerome Fillmon"
+LICENSE="Public Domain"
+REVISION="1"
+commit="a488ce807404481b74f989eac179398b5a48c1d5"
+SOURCE_URI="https://github.com/HaikuArchives/BeBuilder/archive/$commit.tar.gz"
+CHECKSUM_SHA256="a42815e7672c7c393850ee282f1e66acd195d4faaea33f1884a4418f4853a82f"
+SOURCE_FILENAME="BeBuilder-$commit.tar.gz"
+SOURCE_DIR="BeBuilder-$commit"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	bebuilder = $portVersion
+	app:BeBuilder = $portVersion
+	"
+
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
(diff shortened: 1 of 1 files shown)


korli:
@xdizzaster a PR at the Bebuilder repo is indeed nice, but not required. Given its size, it's possible that the patch might need more work though, I dunno.

xdizzaster:
As I said, it is mostly changing iostream.h to iostream, including things like stdlib for malloc and removing the default arguments in functions (from what I found, in modern C++, this is only permitted in the function declaration in the header files, not where the actual function code is written).

fbrosson:
@korli, @waddlesplash, if the recipe (with @xdizzaster's patch) builds fine (as expected) on each arch, would this be enough for merging the patch on the upstream repo?

pulkomandy:
Please submit the changes as a PR to BeBuilder, as it will be much easier to review than here. If there are further changes needed we can always discuss it there?

xdizzaster:
Opened https://github.com/HaikuArchives/BeBuilder/pull/2

xdizzaster:
Updated with the latest commit and removed the patchfile.

xdizzaster:
I think this PR can be merged and closed. Is there anything else to be done? Pinging @fbrosson

fbrosson:
Yes, I was thinking about the rsrc to rdef conversion.
If you wish to do the conversion, that would be great!
But there is no urgency. We can keep this PR open so that you can come back to it (and do the conversion) **after** the GCI contest ends.
This will allow you to focus on other tasks if you want, instead of doing some polishing.
Of course, if you wish, you can do the conversion before the GCI ends.
Needless to say, we can also merge this recipe without the conversion if you prefer.
You decide what is best! After all, you did all the work!

xdizzaster:
Opened https://github.com/HaikuArchives/BeBuilder/pull/3 'cause I don't like leaving unfinished business.

xdizzaster:
Updated to download and use the latest commit. I think it can be merged now.

fbrosson:
Shall we first see what's the best way to update the middle and minor parameters in the rdef?

xdizzaster:
I'll open another PR then

scottmc:
In that case, can you rename the file:
BeBuilder/Doc/art/BeBuilderToolBar,jpg
BeBuilder/Doc/art/BeBuilderToolBar.jpg

that typo (comma) in there makes it so you can't view it online.  When fixed it could we added to the README.md file such that the image could be seen online.

xdizzaster:
Yeah, sure. I'll open a PR with the requested changes.

xdizzaster:
Opened https://github.com/HaikuArchives/BeBuilder/pull/4

xdizzaster:
Updated the commit and the checksum.

PR #1036, opened 2017-01-11
Zsh : bump version, updated recipe and patchset

miqlas:
Disabled /bin/sh provides, makes problems, needs more testing.
The recipe works hovewer, ok to merge.


app-shells/zsh/zsh-5.3.1.recipe
```diff
@@ -3,13 +3,13 @@ DESCRIPTION="Zsh is a shell designed for interactive use, although it \
...
 were added."
-HOMEPAGE="http://zsh.sourceforge.net"
-COPYRIGHT="1992-2014 Paul Falstad, Richard Coleman, Zoltán Hidvégi, \
+HOMEPAGE="https://zsh.org/"
+COPYRIGHT="1992-2017 Paul Falstad, Richard Coleman, Zoltán Hidvégi, \
 Andrew Main, Peter Stephenson, Sven Wishnowsky, and others"
 LICENSE="ZSH"
-REVISION="2"
+REVISION="1"
 SOURCE_URI="http://sourceforge.net/projects/zsh/files/zsh/$portVersion/zsh-$portVersion.tar.xz"
```
fbrosson: You can change this to:
```
SOURCE_URI="https://downloads.sourceforge.net/zsh/zsh-$portVersion.tar.xz"
```
or
```
SOURCE_URI="https://downloads.sf.net/zsh/zsh-$portVersion.tar.xz"
```

app-shells/zsh/zsh-5.3.1.recipe
```diff
@@ -3,13 +3,13 @@ DESCRIPTION="Zsh is a shell designed for interactive use, although it \
...
-CHECKSUM_SHA256="74e9453b5470b3c0970f9f93cfd603d241c3d7b1968adc0e4b3951073e8d3dec"
+CHECKSUM_SHA256="fc886cb2ade032d006da8322c09a7e92b2309177811428b121192d44832920da"
 PATCHES="zsh-$portVersion.patchset"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64"
```
fbrosson: You are using $secondaryArchSuffix and
```
SECONDARY_ARCHITECTURES="x86"
```
but the recipe provides no libs, so you should either:
* remove support for secondary architectures,
* **_or_** mark x86_gcc2 as broken (if that's the case).

app-shells/zsh/zsh-5.3.1.recipe
```diff
@@ -0,0 +1,79 @@
...
+		--enable-pcre
+	sed --in-place '/HAVE_WAIT3/d' config.h
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	#zsh isn't bash compatible enough
+	#ln -s zsh $binDir/sh
```
waddlesplash: Why did you disable this just because Zsh is "not bash compatible"? All previous versions of the recipe left it enabled.

miqlas: If zsh  enabled on my machine with cmd:sh it generates serious problems during configuring thngs with automake (autmake scripts using /bin/sh almost all the times). It is a bug, it shouldn't let break anything.
Without zsh the works correctly. We should let /bin/sh for bash right now.

mmuman:
sh should be a POSIX compatible shell. Pissibly zsh doesn't detect properly it's started as sh (maybe argv[0] contains the path or something) and doesn't enable POSIX compatibility...
Anyway, the proper way to change the interactive user shell is with chsh.

PR #1040, opened 2017-01-12
flac123: new recipe for a commandline FLAC player

raefaldhia:
Adding new recipe for flac123, a commandline program for playing FLAC audio files.

### DIFF
--- a/media-sound/flac123/flac123-0.0.12.recipe
+++ b/media-sound/flac123/flac123-0.0.12.recipe
@@ -1,0 +1,60 @@
+SUMMARY="A command-line program for playing FLAC audio files"
+DESCRIPTION="flac-tools aims to be a complete set of tools for playing and \
+manipulating audio files stored in the FLAC format. flac123 is the only \
+component of flac-tools at this time."
+HOMEPAGE="http://flac-tools.sourceforge.net/"
+COPYRIGHT="2003-2012 Jake Angerman, Dan Johnson, Hans Oesterholt"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/flac-tools/flac123-$portVersion-release.tar.gz"
+CHECKSUM_SHA256="1976efd54a918eadd3cb10b34c77cee009e21ae56274148afa01edf32654e47d"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	flac123 = $portVersion
+	cmd:flac123 = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libFLAC
+	lib:libao
+	lib:libogg
+	lib:libpopt
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libFLAC
+	devel:libao
+	devel:libogg
+	devel:libpopt
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:awk
+	cmd:gcc
+	cmd:make
+	"
+
+BUILD()
+{
+	aclocal
+	automake --add-missing
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
+
+TEST()
+{
+	# No real test at the moment
+	make check
+}


media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,52 @@
...
+component of flac-tools at this time."
+HOMEPAGE="http://flac-tools.sourceforge.net/"
+COPYRIGHT="2003-2017 Jake Angerman"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/flac-tools/flac123-$portVersion-release.tar.gz"
+CHECKSUM_SHA256="1976efd54a918eadd3cb10b34c77cee009e21ae56274148afa01edf32654e47d"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
```
korli: It seems there is no need for SECONDARY_ARCHITECTURES, as this package doesn't provide any lib.

fbrosson: @raefaldhia, this means you just need to:
* remove the line that is defining SECONDARY_ARCHITECTURES, as @korli already wrote.
* remove each occurrence of $secondaryArchSuffix and ${secondaryArchSuffix} because these are unneeded when haikuporter is building for a primary arch. BTW, the recipe works with them because haikuporter simply leaves them undefined, so they count for nothing. But we like to get rid of them when they are not needed.

(Sorry if that was obvious.)

Hmm, I haven't tested the recipe on any arch yet. If you tried on x86 (but not on x86_gcc2) then we first need to make sure it works on x86_gcc2, otherwise you might have to add back the SECONDARY_ARCHITECTURES and the $secondaryArchSuffix and ${secondaryArchSuffix}.

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,52 @@
+SUMMARY="A command-line program for playing FLAC audio files"
+DESCRIPTION="flac-tools aims to be a complete set of tools for playing and \
+manipulating audio files stored the the FLAC format.  flac123 is the only \
```
fbrosson: "stored **the** the FLAC format" -> "stored **in** the FLAC format"

raefaldhia: Lol that's actualy copied from readme inside the flac123 if i'm not mistaken XD.

media-sound/flac123/flac123-0.0.12.recipe
raefaldhia: Edited, thank you!

raefaldhia: Nah, i tried the recipe on x86_gcc2.

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,51 @@
+SUMMARY="A command-line program for playing FLAC audio files"
+DESCRIPTION="flac-tools aims to be a complete set of tools for playing and \
+manipulating audio files stored in the FLAC format.  flac123 is the only \
```
fbrosson: You can squeeze (into 1 white space) the 2 white spaces between "format." and "flac123".

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,51 @@
...
+component of flac-tools at this time."
+HOMEPAGE="http://flac-tools.sourceforge.net/"
+COPYRIGHT="2003-2017 Jake Angerman"
```
fbrosson: Jake Angerman is one of the original authors. Maybe we should add the other one, and the current maintainer too.
The only problem is that the AUTHORS file does not provide any copyright years.

So I don't know if
```
COPYRIGHT="2003-2012 Jake Angerman, Dan Johnson, Hans Oesterholt"
```
would be acceptable.

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,51 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/flac-tools/flac123-$portVersion-release.tar.gz"
```
fbrosson: You can change this to
```
SOURCE_URI="https://downloads.sourceforge.net/flac-tools/flac123-$portVersion-release.tar.gz" 
```
or
```
SOURCE_URI="https://downloads.sf.net/flac-tools/flac123-$portVersion-release.tar.gz" 
```
since SourceForge is accepting https traffic and, for flac123, it is not even redirecting it to http ;-)
(Well, maybe I was lucky and got an https-enabled mirror ;-)
Anyway, https will work for everyone. And if SF redirects to http that is not a problem. You can use https for the SOURCE_URI.

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,51 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libFLAC
+	devel:libao
+	devel:libogg
+	devel:libpopt
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
```
fbrosson: I would replace
```
	cmd:autoreconf
```
by
```
	cmd:autoconf
        cmd:autoheader
        cmd:automake
```

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,51 @@
...
+	cmd:awk
+	cmd:gcc
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf -fi
```
fbrosson: I would replace
```
	autoreconf -fi
```
by
```
	aclocal
	automake --add-missing
```
It works fine on x86_gcc2. I'll check this on x86 and x86_64 but there is not reason it would not work there too.

I think this will speed things a little bit. But if @korli tells you autoreconf is better than that's OK for me ;-)
(In that case, you should keep the cmd:autoreconf)

Anyway, you can also add a TEST() block with "make check".

raefaldhia: Actually the `make check` doesn't do the "test" :/

media-sound/flac123/flac123-0.0.12.recipe
pulkomandy: This is standard english typography: https://en.wikipedia.org/wiki/Sentence_spacing
But yes, many people (even english) consider it strange and we usually don't use it.

media-sound/flac123/flac123-0.0.12.recipe
pulkomandy: Well, not everyone can use https unfortunately. We had at least one GCI student this year whose ISP has a "transparent" proxy for HTTPS, serving its own certificate so it can spy on HTTPS connections. This is a problem because Haiku does not recognize the certificate (for good reasons) and there is no easy way to add it.
However, we should support the use of SSL secure connections whenever possible.

media-sound/flac123/flac123-0.0.12.recipe
pulkomandy: But that's flac's fault. The result will be that the test always passes.
Alternatively you can do this:

    TEST()
    {
        true # Sorry, this project has no tests!
    }

This allows us to remember that the project has no support. It can be useful, as in next years GCI, we may have tasks about researching which recipes would benefit from having a TEST block added, and we need to remember which one were already checked.

media-sound/flac123/flac123-0.0.12.recipe
raefaldhia: Should i use 2012 instead?

raefaldhia:
Oopps forgot to update copyright wait a sec.

media-sound/flac123/flac123-0.0.12.recipe
fbrosson: I've just checked that the recipe also works on x86 and x86_64 with my suggestions.
BTW, I also droped the cmd:awk.
So you can go ahead, and replace
```
	cmd:autoreconf 
	cmd:awk
```
by
```
	cmd:autoconf
        cmd:autoheader
        cmd:automake
```
and

```
	autoreconf -fi
```
by
```
	aclocal
	automake --add-missing
```
(as suggested above.)

Regarding the tests, can't we have this?
```
TEST()
{
	# no real test at the moment
	make check
}
```
It's better than a simple "true" because it lets us know that the Makefile has a check target that might be used some day, should the upstream authors add some tests.

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,56 @@
+SUMMARY="A command-line program for playing FLAC audio files"
+DESCRIPTION="flac-tools aims to be a complete set of tools for playing and \
+manipulating audio files stored in the FLAC format. flac123 is the only \
+component of flac-tools at this time."
+HOMEPAGE="https://flac-tools.sourceforge.net/"
```
fbrosson: Oops, the https is for SOURCE_URI, not for HOMEPAGE.

BTW, don'y you want to drop the autoreconf in favor of the aclocal + automake --missing ?
autoreconf calls other tools which are not needed here.

raefaldhia: lol wait, xd, ok i'll do that i was confused when reading that part of sentence.

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,59 @@
...
+	devel:libao
+	devel:libogg
+	devel:libpopt
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
+	cmd:awk
```
fbrosson: You can drop the cmd:awk.

You can also add a comment before the "make check" in TEST(). I think @pulkomandy will like it ;-)
```
	# no real tests at the moment
```

media-sound/flac123/flac123-0.0.12.recipe
```diff
@@ -0,0 +1,59 @@
...
+}
+
+INSTALL()
+{
+	make install
+}
+
+TEST()
+{
+	make check
```
fbrosson: ~~You can drop the cmd:awk.~~
Sorry, I had not noticed that the configure needed awk. So let's keep it ;-)

You can also add a comment before the "make check" in TEST(). I think @pulkomandy will like it ;-)
```
	# no real tests at the moment
```

media-sound/flac123/flac123-0.0.12.recipe
fbrosson: I agree that we should use https whenever it is possible.
And I would even add that we should never give up using https just because some users have "unfriendly" proxys. BTW, As we know, doing so would help these bad proxys to remain unnoticed... and we don't want that ;-)

BTW, even on countries where people think they live in free democracies, when you get free wifi in some malls, what you have is spying proxys which do exactly what you mentioned: self-signed  certificates generated "on-the-fly". Stupid popular smartphones don't even notice it. Shame on them!

PR #1047, opened 2017-01-12
lrzsz: added recipe

xdizzaster:
Added recipe for lrzsz communication package.

I am aware that editing `Makefile.in` is not an optimal solution, but because of the package's age, it does not respond well when attempting to regenerate the configuration scripts with a modern version of the autotools. If someone is able to do it, I will be happy to update the commit, but until then, this is the solution I've come up with.

### DIFF
--- a/net-dialup/lrzsz/lrzsz-0.12.20.recipe
+++ b/net-dialup/lrzsz/lrzsz-0.12.20.recipe
@@ -1,0 +1,58 @@
+SUMMARY="Communication package providing the XMODEM, YMODEM & ZMODEM protocols"
+DESCRIPTION="lrzsz is a Unix package providing the XMODEM, YMODEM AND ZMODEM \
+file transfer protocols."
+HOMEPAGE="https://ohse.de/uwe/software/lrzsz.html"
+COPYRIGHT="1996-1998 Uwe Ohse"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://ohse.de/uwe/releases/lrzsz-$portVersion.tar.gz"
+CHECKSUM_SHA256="c28b36b14bddb014d9e9c97c52459852f97bd405f89113f30bee45ed92728ff1"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	lrzsz = $portVersion
+	cmd:lrz = $portVersion
+	cmd:lsz = $portVersion
+	cmd:lrb = $portVersion
+	cmd:lrx = $portVersion
+	cmd:lsb = $portVersion
+	cmd:lsx = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:ld
+	cmd:make
+	cmd:sed
+	"
+
+PATCH()
+{
+	sed -i -e "s|ln|ln -r -s|" src/Makefile.in
+}
+
+BUILD()
+{
+	runConfigure --omit-dirs "docDir dataRootDir" ./configure
+	make $jobArgs LIBS="-lnetwork"
+}
+
+INSTALL()
+{
+	make install
+	mkdir -p $dataDir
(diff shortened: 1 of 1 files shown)


net-dialup/lrzsz/lrzsz-0.12.20.recipe
```diff
@@ -0,0 +1,57 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:ld
+	cmd:make
+	"
+
+PATCH()
+{
+	sed -i -e "s|ln|ln -s|" src/Makefile.in
```
fbrosson: Can you please change this to:
```
	sed -i -e "s|ln|ln -r -s|" src/Makefile.in
```
The "-r" (or --relative) will produce nice relative symbolic links (instead of the absolute ones) for the files we install in bin.

net-dialup/lrzsz/lrzsz-0.12.20.recipe
```diff
@@ -0,0 +1,57 @@
...
+	runConfigure --omit-dirs "docDir dataRootDir" ./configure
+	make $jobArgs LIBS="-lnetwork"
+}
+
+INSTALL()
+{
+	make install
+	mkdir -p $prefix/data/
+	mv $prefix/share/* $prefix/data
+	rmdir $prefix/share/
```
xdizzaster: Yes, I tried to avoid the mv. It seems that the configure scripts does not support being passed the `--datarootdir` argument the way modern configure scripts do, so it must be disabled when calling runConfigure and done manually.

xdizzaster:
Changed some things, as @fbrosson suggested.

net-dialup/lrzsz/lrzsz-0.12.20.recipe
```diff
@@ -0,0 +1,57 @@
...
+	runConfigure --omit-dirs "docDir dataRootDir" ./configure
+	make $jobArgs LIBS="-lnetwork"
+}
+
+INSTALL()
+{
+	make install
+	mkdir -p $dataDir
+	mv $prefix/share/* $dataDir
+	rmdir $prefix/share/
```
fbrosson: The trailing slash in "share/" does not harm, but it is not needed either, and the reader already knows it is a directory, since the command we're calling is a simple rmdir.
So imho it would be better to drop that trailing slash.
(That said, you can keep it if you wish ;-)
Hmm, maybe you did not notice I had dropped it, or maybe you forgot to remove it.
Anyway, you can do as you wish ;-)

xdizzaster:
Removed a slash.

xdizzaster:
One way of simplifying would be to regenerate the configure script, but, because of its age, autotools fail to do so. The configuration files probably need some serious patching, if not to be rewritten from the ground up and I don't know if it's worth the trouble.

net-dialup/lrzsz/lrzsz-0.12.20.recipe
```diff
@@ -0,0 +1,57 @@
...
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:ld
+	cmd:make
```
korli: Please add "cmd:sed"

fbrosson: @xdizzaster, you can go ahead with adding cmd:sed.
@korli started contributing to Haiku at least two years before me (and has pushed much more commits than I did) so you can trust him.
Don't worry, you can tell you were busy and could not make the change earlier. He won't be angry. We are all very happy to see you making great contributions!

xdizzaster: Umm, I really was busy and didn't see the GitHub notification. It's not really a matter of trust.

xdizzaster:
Added `sed` as a build dependency

net-dialup/lrzsz/lrzsz-0.12.20.recipe
fbrosson: No problem ;-)
Let me do some magic with our two messages, then!

PR #1048, opened 2017-01-12
libksba: added recipe

xdizzaster:
Added recipe for liksba X.509 library.

### DIFF
--- a/dev-libs/libksba/libksba-1.3.5.recipe
+++ b/dev-libs/libksba/libksba-1.3.5.recipe
@@ -1,0 +1,75 @@
+SUMMARY="Library for working with X.509 certificates, CMS data and related"
+DESCRIPTION="Libksba is a library to make the task of working with X.509, CMS \
+data and related objects easier. It provides a highlevel interface to \
+implemented protocols and presents the data in a consistent way."
+HOMEPAGE="https://www.gnupg.org/related_software/libksba/"
+COPYRIGHT="2001-2015 g10 Code GmbH"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://www.gnupg.org/ftp/gcrypt/libksba/libksba-$portVersion.tar.bz2"
+CHECKSUM_SHA256="41444fd7a6ff73a79ad9728f985e71c9ba8cd3e5e53358e70d5f066d35c1a340"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	libksba$secondaryArchSuffix = $portVersion
+	lib:libksba$secondaryArchSuffix = 8.11.6 compat >= 8
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgpg_error$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libksba${secondaryArchSuffix}_devel = $portVersion
+	devel:libksba$secondaryArchSuffix = 8.11.6 compat >= 8
+	cmd:ksba_config$secondaryArchSuffix = $portVersion
+	"
+REQUIRES_devel="
+	libksba$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libgpg_error$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	./autogen.sh
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	rm $libDir/libksba.la
+
+	prepareInstalledDevelLib libksba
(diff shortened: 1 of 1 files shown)


dev-libs/libksba/libksba-1.3.5.recipe
```diff
@@ -0,0 +1,67 @@
...
+data and related objects easier. It provides a highlevel interface to \
+implemented protocols and presents the data in a consistent way."
+HOMEPAGE="https://www.gnupg.org/related_software/libksba/"
+COPYRIGHT="2001-2015 g10 Code GmbH"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://www.gnupg.org/ftp/gcrypt/libksba/libksba-$portVersion.tar.bz2"
+CHECKSUM_SHA256="41444fd7a6ff73a79ad9728f985e71c9ba8cd3e5e53358e70d5f066d35c1a340"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
korli: This is a library, SECONDARY_ARCHITECTURES is needed :)  Otherwise perfect.

fbrosson: @xdizzaster, this means you just need to add
```
```
and insert the famous $secondaryArchSuffix at the usual places.

Great job indeed!

xdizzaster:
Added `SECONDARY_ARCHITECTURES`

dev-libs/libksba/libksba-1.3.5.recipe
```diff
@@ -0,0 +1,68 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libgpg_error$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
```
fbrosson: We prefer to not have cmd:git as a dependency, unless the build fails without it.
But it never fails ;-)

dev-libs/libksba/libksba-1.3.5.recipe
```diff
@@ -0,0 +1,68 @@
...
+
+INSTALL()
+{
+	make install
+	rm $libDir/libksba.la
+
+	prepareInstalledDevelLib libksba
+	fixPkgconfig
+
+	packageEntries devel $developDir
```
fbrosson: Please replace
```
```
by
```
	packageEntries devel \
		$binDir/ksba-config \
		$developDir
```

xdizzaster:
Moved `ksba_config` to the devel package

xdizzaster:
Removed `git` as a build dep

dev-libs/libksba/libksba-1.3.5.recipe
```diff
@@ -0,0 +1,70 @@
...
+{
+	make install
+	rm $libDir/libksba.la
+
+	prepareInstalledDevelLib libksba
+	fixPkgconfig
+
+	packageEntries devel \
+		$binDir/ksba-config \
+		$developDir
```
fbrosson: I think we should also move data/aclocal/ksba.m4 to the _devel package.
(I wrote it in my EDIT2 but it is in a comment that is now hidden by default, so you did not notice it.)
```
	packageEntries devel \
		$binDir/ksba-config \
		$dataDir/aclocal/ksba.m4 \
		$developDir
```
@korli, I'm not sure for the data/aclocal/ksba.m4. Could you please confirm that it is OK to move it the _devel package? Thanks!

korli: I confirm.

fbrosson: @xdizzaster, couly you please also add this block:
```

		rm -fr $infoDir
	fi

```
before the "packageEntries devel"?
(with blank lines before and after.)
Thanks!

xdizzaster:
Moved ksba.m4 to the devel package

dev-libs/libksba/libksba-1.3.5.recipe
```diff
@@ -0,0 +1,71 @@
...
+
+INSTALL()
+{
+	make install
+	rm $libDir/libksba.la
+
+	prepareInstalledDevelLib libksba
+	fixPkgconfig
+
+	packageEntries devel \
```
fbrosson: @xdizzaster, could you please also add this block:
```

		rm -fr $infoDir
	fi

```
before the "packageEntries devel"?
(with blank lines before and after.)
Thanks!

BTW, what it does is that it gets rid of the documentation/info/ksba.info when we build for a secondary arch.

xdizzaster: Why would we want to get rid of the documentation? It seems like the thing you wouldn't want to do

fbrosson: The rm is only executed if $secondaryArchSuffix is not an empty string.
And since that variable is empty for any primary arch (and not empty otherwise) we have a simple method to keep the doc only once, i.e. in the hpkgs created for primary architectures.

fbrosson: Guess what? With libksba we will be able to add a recipe for gnupg-2.x.y !
(We're currently stuck with gnupg-1.4.21)

xdizzaster: Oh, I get it now, thanks

xdizzaster:
Removed info if building on a secondary arch

PR #1052, opened 2017-01-13
stockfish: new recipe for a chess engine

raefaldhia:
Adding new recipe for stockfish, a chess engine claimed to be strongest in the world.

### DIFF
--- a/games-board/stockfish/stockfish-8.recipe
+++ b/games-board/stockfish/stockfish-8.recipe
@@ -1,0 +1,67 @@
+SUMMARY="A chess engine claimed to be strongest in the world"
+DESCRIPTION="Stockfish is a free UCI chess engine derived from Glaurung 2.1. It \
+is not a complete chess program and requires some UCI-compatible GUI (e.g. \
+XBoard with PolyGlot, eboard, Arena, Sigma Chess, Shredder, ChessPartner or \
+Fritz) in order to be used comfortably. Read the documentation for your GUI of \
+choice for information about how to use Stockfish with it."
+HOMEPAGE="https://www.stockfishchess.org/"
+COPYRIGHT="2004-2008 Tord Romstad (Glaurung author)
+	2008-2016 Marco Costalba, Joona Kiiski, Gary Linscott, Tord Romstad"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/official-stockfish/Stockfish/archive/sf_$portVersion.tar.gz"
+CHECKSUM_SHA256="e2b77d801d8837c22c2391b8cae08622c1434a2bb5947ddb08b081e1ad35c76c"
+SOURCE_DIR="Stockfish-sf_$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandSuffix=$secondaryArchSuffix
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+case "$effectiveTargetArchitecture" in
+	x86) commandArch=x86-32;;
+	x86_64) commandArch=x86-64;;
+	arm) commandArch=armv7;;
+	ppc) commandArch=ppc-32;;
+esac
+
+PROVIDES="
+	stockfish$secondaryArchSuffix = $portVersion
+	cmd:stockfish$commandSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	make -C src build ARCH=$commandArch
+}
+
+INSTALL()
+{
+	install -d $docDir
+	install -t $docDir AUTHORS Readme.md
+
+	make -C src install BINDIR=$commandBinDir
+}
+
+TEST()
+{
+	test `src/stockfish bench 2>&1 1>/dev/null \
+		| sed -n -e "s/^Nodes searched *: *//p"` \
+		-eq 5926706
+}


games-board/stockfish/stockfish-sf_8.recipe
```diff
@@ -0,0 +1,72 @@
...
+COPYRIGHT="2004-2008 Tord Romstad (Glaurung author)
+	2008-2017 Marco Costalba, Joona Kiiski, Gary Linscott, Tord Romstad"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/official-stockfish/Stockfish/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="e2b77d801d8837c22c2391b8cae08622c1434a2bb5947ddb08b081e1ad35c76c"
+SOURCE_DIR="Stockfish-sf_8"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
raefaldhia: Should i remove this SECONDARY_ARCHITECTURES?

games-board/stockfish/stockfish-sf_8.recipe
```diff
@@ -0,0 +1,72 @@
...
+		commandBinDir=$prefix/bin
+		commandArch=x86-32
+	elif [ "$targetArchitecture" == x86_64 ]; then
+		commandBinDir=$binDir
+		commandArch=x86-64
+	fi
+else
+	commandBinDir=$prefix/bin
+	commandArch=x86-32
+fi
```
raefaldhia: I'm not sure, is it right?

fbrosson: The rule is that we set commandBinDir to $binDir for all primary architectures where it builds fine.

If it does not build for x86_gcc2, then, for that arch only, we set commandBinDir to $prefixBin so that we can build stockfish_x86, that is, we build it on x86_gcc2 but using the x86 secondary arch.
The result of the commandBinDir trick is that instead of installing the runtimes in /system/bin/x86/ we install then in /system/bin/, where you know we won't ever clash with the runtimes built for x86_gcc2 primary arch since we know it does not build on x86_gcc2 primary arch.

fbrosson: So you only need to set commandBinDir to $prefix/bin if both of these conditions are met:
* the software fails to build on x86_gcc2 primary arch
* haikuporter is building for the x86 secondary arch.

So I guess you should use the same block as you did for haikuports/sci-libs/primesieve/primesieve-5.7.3.
Alternatively, you can use the following variant, which we have on many recipes, except that they also define commandSuffix. (BTW, I guess that also using commandSuffix would be usefull in the future if we create a x86_64 hybrid, i.e. with 64 bits for the primary arch and 32 for a x86 secondary arch.)
```
commandBinDir=$binDir
	commandBinDir=$prefix/bin
fi
```

Regarding the commandArch, I think you shoud use something like:
```
case "$effectiveTargetArchitecture" in
	x86) commandArch=x86-32;;
	x86_64) commandArch=x86-64;;
	arm) commandArch=armv7;;
	ppc) commandArch=ppc-32;;
esac
```
But I don't know if it's worth adding ppc and arm right now, since our nightly builds for these architectures are not currently disabled (due to missing devs or bootstrap issue, I don't know exactly).
So let's see what the other mentors say about that, i.e. if you can keep arm and ppc in this recipe.
Since you have investigated about the ARCH parameter to pass for them, they might agree with keeping them here, although it won't be possible to build the recipe for arm or ppc right now ;-)

games-board/stockfish/stockfish-sf_8.recipe
fbrosson: If you tried to build on x86_gcc2 using the primary arch, that is, building stockfish (and not stockfish_x86) and noticed that the source is not compatible with gcc2, then you should keep SECONDARY_ARCHITECTURES. Your current definition for it looks OK, btw.

If, on the other side, it builds fine on x86_gcc2 primary arch, then there is no need to keep SECONDARY_ARCHITECTURES and you should, in that case, remove it.

games-board/stockfish/stockfish-sf_8.recipe
raefaldhia: Ok, updated.

games-board/stockfish/stockfish-sf_8.recipe
```diff
@@ -0,0 +1,65 @@
...
+INSTALL()
+{
+	cd src/
+	make install \
+		BINDIR=$commandBinDir
+}
+
+TEST()
+{
+	true # Sorry, this project has no tests!
```
raefaldhia: From my previous pr, @pulkomandy told me that i could do something like this for reasons
>This allows us to remember that the project has no support. It can be useful, as in next years GCI, we may have tasks about researching which recipes would benefit from having a TEST block added, and we need to remember which one were already checked.

games-board/stockfish/stockfish-sf_8.recipe
```diff
@@ -0,0 +1,65 @@
...
+case "$effectiveTargetArchitecture" in
+	x86) commandArch=x86-32;;
+	x86_64) commandArch=x86-64;;
+	arm) commandArch=armv7;;
+	ppc) commandArch=ppc-32;;
+esac
+
+PROVIDES="
+	stockfish$secondaryArchSuffix = $portVersion
+	cmd:stockfish$secondaryArchSuffix = $portVersion
```
fbrosson: Should be either
```
	cmd:stockfish = $portVersion
```
or
```
	cmd:stockfish$commandSuffix = $portVersion
```

In the second case, you need to define commandSuffix the same way it is being defined in haikuports/app-arch/gcab/gcab-0.7.recipe.

raefaldhia: ok updated.

games-board/stockfish/stockfish-sf_8.recipe
```diff
@@ -0,0 +1,67 @@
+SUMMARY="A chess engine claimed to be strongest in the world"
+DESCRIPTION="Stockfish is a free UCI chess engine derived from Glaurung 2.1. It \
+is not a complete chess program and requires some UCI-compatible GUI (e.g. \
+XBoard with PolyGlot, eboard, Arena, Sigma Chess, Shredder, ChessPartner or \
+Fritz) in order to be used comfortably. Read the documentation for your GUI of \
+choice for information about how to use Stockfish with it."
+HOMEPAGE="https://www.stockfishchess.org/"
+COPYRIGHT="2004-2008 Tord Romstad (Glaurung author)
+	2008-2017 Marco Costalba, Joona Kiiski, Gary Linscott, Tord Romstad"
```
fbrosson: sf_8 was released on November 1st, 2016, so you should replace 2017 by 2016.

BTW, we should probably rename the recipe to stockfish-8.recipe and use "sf_$portVersion" where we are using "$portVersion".
You would then change SOURCE_DIR to
```
SOURCE_DIR="Stockfish-sf_$portVersion"
```

To rename:
```
cd haikuports/games-board/stockfish
git mv stockfish-sf_8.recipe stockfish-8.recipe
git commit --amend --no-edit stockfish-sf_8.recipe stockfish-8.recipe
```

raefaldhia: ok done.

games-board/stockfish/stockfish-8.recipe
```diff
@@ -0,0 +1,67 @@
...
+{
+	cd src/
+	make build \
+		ARCH=$commandArch
+		COMP=gcc
+}
+
+INSTALL()
+{
+	cd src/
```
fbrosson: Please insert
```
	install -d $docDir
	install -t $docDir AUTHORS Readme.md
```
before the "cd src".
BTW, I would also drop (in BUILD and INSTALL) the "/" in "cd src/" since it is not needed.
Thanks!

raefaldhia: anyway about the comment in gci task, you could try to run 
./stockfish bench 2>&1 >/dev/null | grep 'Nodes searched' | tee bench1
this code available inside travis.ci on the package but unfortunatelly to test if this works correctly we need to build the package both in x86 and x86-64 then compare the result(see more in travis.ci file in scripts section, i'm not quite sure).

fbrosson: That's interesting!
BTW, on x86 I get:
```
Nodes searched  : 5926706
```

raefaldhia: I have same result on x86: 5926706

fbrosson: Same thing on x86_64: 5926706

games-board/stockfish/stockfish-8.recipe
```diff
@@ -0,0 +1,69 @@
...
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
```
fbrosson: Shall we add this to BUILD_PREREQUIRES?
```
	cmd:sed
```

games-board/stockfish/stockfish-8.recipe
```diff
@@ -0,0 +1,69 @@
...
+	install -t $docDir AUTHORS Readme.md
+
+	cd src
+	make install \
+		BINDIR=$commandBinDir
+}
+
+TEST()
+{
+	true # Sorry, this project has no tests!
```
fbrosson: This works fine:
```
TEST()
{
	test `src/stockfish bench 2>&1 1>/dev/null \
		| sed -n -e "s/^Nodes searched *: *//p"` \
		-eq 5926706
}
```

games-board/stockfish/stockfish-8.recipe
```diff
@@ -0,0 +1,69 @@
...
+}
+
+INSTALL()
+{
+	install -d $docDir
+	install -t $docDir AUTHORS Readme.md
+
+	cd src
+	make install \
+		BINDIR=$commandBinDir
```
fbrosson: Shall we replace
```
	cd src
	make install \
		BINDIR=$commandBinDir
```
by
```
	make -C src install BINDIR=$commandBinDir
```
and move it to the top of INSTALL ?

games-board/stockfish/stockfish-8.recipe
```diff
@@ -0,0 +1,69 @@
...
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	cd src
+	make build \
+		ARCH=$commandArch
```
fbrosson: Shall we replace
```
	cd src
	make build \
		ARCH=$commandArch
```
by
```
	make -C src build ARCH=$commandArch
```

games-board/stockfish/stockfish-8.recipe
raefaldhia: Nice!

games-board/stockfish/stockfish-8.recipe
waddlesplash: `TEST` runs outside of chroot presently, so, that'd be pointless I think.

PR #1076, opened 2017-01-15
tox: new recipe for messaging platform (WIP, do not merge yet)

raefaldhia:
Adding new recipe for tox, an encrypted, P2P, messaging, audio/video calling platform.

### DIFF
--- a/net-libs/tox/patches/tox-0.1.4.patchset
+++ b/net-libs/tox/patches/tox-0.1.4.patchset
@@ -1,0 +1,23 @@
+From c00ce44468e0e913c9cb2738dd17dea27c6fd11c Mon Sep 17 00:00:00 2001
+From: raefaldhia <raefaldhiamartya@gmail.com>
+Date: Sat, 14 Jan 2017 18:59:20 +0700
+Subject: Add compatibility to build on haikuporter
+
+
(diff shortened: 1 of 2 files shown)


raefaldhia:
Are you sure not facing kernel panic with --test? i still get it atm.

fbrosson:
OMG, I also get the kernel crash when I run the tests!
I thought you were getting the crash when lauching the runtime. I had not tried the test before you mentioned it.

raefaldhia:
Yes the crash caused by encryptsave_test, tox_test, union_test, and dhf_test. still have no idea what caused it.

For encryptsave_test that's caused by something inside `new_network_ex`.

raefaldhia:
After some hours spending time to track the code flow, `setsocketopt` above `return NULL;` causes crash for `encryptsave_test` still have no idea why this is happened.

screenshot from 2017-01-15 17-24-32

Ignore the `return NULL;` that's just my code to track the code flow.

mmlr:
KDLs are never the fault of the port (or porter) but always a bug in Haiku itself that should be taken up in the Haiku bug tracker at https://dev.haiku-os.org/ . Your last comment makes for a nice bug report as you've already narrowed the cause down significantly.

PR #1081, opened 2017-01-16
cmake: bump version

miqlas:
Libuv: broken, see here: https://github.com/haikuports/haikuports/issues/1082 (disabled for now)
Libjsoncpp: needs to be ported


dev-util/cmake/cmake-3.7.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+	#devel:libuv$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	haiku_devel
+	cmd:gcc$secondaryArchSuffix
+	cmd:gfortran$secondaryArchSuffix
+	cmd:grep
+	cmd:ld$secondaryArchSuffix
+	cmd:libtool
```
fbrosson: Should be
```
```

dev-util/cmake/cmake-3.7.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+	devel:libexpat$secondaryArchSuffix
+	devel:libform$secondaryArchSuffix
+	#devel:jsoncpp$secondaryArchSuffix
+	devel:liblzma$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	#devel:libuv$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	haiku_devel
```
fbrosson: I guess you should remove "haiku_devel" from BUILD_**PRE**REQUIRES because you already have
```
```
in BUILD_REQUIRES.

korli:
Can you make the bump without system libs first? 
It's going to be painful to maintain cmake if additional dependencies are not stable.
In this case, whenever a problem appears, this version will be disabled :(

miqlas:
I can do that!

I tought some days about using syslibs for cmake, but as it provides in-tree libs, we could never be sure if they containing our latest Haiku-supporting patches or not, so it would be better to have them dynamically linked.
On the other side it can compicate the buildstrap process on new platforms.
I had to check the syslibs, most of them doesn't require cmake to get compiled, the only exception is zlib (is zlib and libzip comaptible?). I'm not sure if it a good thing, because it can create a circular dependecy hell for the buildbots. Any thoughts, @korli?

miqlas:
@korli: cmake have a new feature in the 3.7.x branch: the server-mode. It is currently disabled, theoretically it requires libuv. Do we want to have it on Haiku?

korli:
The bootstrap process is still possible because zlib exists in the cross_devel repository, but others don't.
I don't care much about server-mode, because it won't be required by another recipe. 
cmake 3.7 uses pthread_barrier which is now available in Haiku, but not widely tested. Thus it's a good idea to do the bump, then later enable features. Otherwise it will probably rust until someone really needs this version.

miqlas:
Aye-aye sir!

dev-util/cmake/cmake-3.7.2.recipe
```diff
@@ -0,0 +1,77 @@
+SUMMARY="A cross-platform build system"
+DESCRIPTION="CMake is a family of tools designed to build, test and package \
+software. It is used to control the software compilation process using \
+simple platform and compiler independent configuration files. CMake generates \
+native makefiles and workspaces that can be used in the compiler environment \
+of your choice."
+HOMEPAGE="https://cmake.org/"
+COPYRIGHT="2002-2016 Kitware, Inc., Insight Consortium"
```
fbrosson: 2016 -> 2017
(3.7.2 was released 5 days ago)

dev-util/cmake/cmake-3.7.2.recipe
```diff
@@ -0,0 +1,77 @@
...
+	cmd:sed
+	"
+
+BUILD()
+{
+	# not an autotools configure
+	./configure --prefix=$prefix \
+		--datadir=/$relativeDataDir/cmake \
+		--docdir=/$relativeDocDir \
+		--mandir=/$relativeManDir \
```
fbrosson: I _think_ this is incorrect:
```
		--datadir=/$relativeDataDir/cmake \
		--docdir=/$relativeDocDir \
		--mandir=/$relativeManDir \
```
Shoudn't we be using
```
		--datadir=$dataDir/cmake \
		--docdir=$docDir \
		--mandir=$manDir \
```
instead ?

dev-util/cmake/cmake-3.7.2.recipe
miqlas: Did it, thx. I live in 2016 yet.

dev-util/cmake/cmake-3.7.2.recipe
```diff
@@ -0,0 +1,77 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libcurl$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:gfortran$secondaryArchSuffix
```
korli: cmd:gfortran shouldn't be required

miqlas: I'm sure it is just optional, but cmake actually checks for it. Should i delete then?

korli: Yes

miqlas: Ready

dev-util/cmake/cmake-3.7.2.recipe
pulkomandy: As is noted in the recipe, this is "not an autotools configure" and the arguments are slightly different. IIRC the configure script will convert them to absolute paths and then call the bootstrap cmake with them.

PR #1084, opened 2017-01-18
Setuptools : bump version, unify Python2 and 3 recipes, drop old

miqlas:


dev-python/setuptools/python_setuptools-33.1.0.recipe
```diff
@@ -9,13 +9,15 @@ LICENSE="Python"
...
 SOURCE_FILENAME="setuptools-$portVersion.tar.gz"
-CHECKSUM_SHA256="90a40d54d21d4cfe55ae2c3597ed3f2a63d9e8fe768f03ff16f2a58ed97bcc32"
+CHECKSUM_SHA256="efa8538fec784c6d33af572abbc6662ae5f3e3bd6d7bbc42124080492215cb45"
 SOURCE_DIR="setuptools-$portVersion"
 
 ARCHITECTURES="x86 x86_gcc2 x86_64"
 
 PROVIDES="
 	python_setuptools = $portVersion
+	cmd:easy_install
```
pulkomandy: Now both recipe provide that command, which means they can't be installed at the same time. Can we avoid that?

PR #1085, opened 2017-01-18
Numpy : bump version

miqlas:
Enabled gfortran and libopenblas for science and justice


dev-python/numpy/python3_numpy-1.11.2.recipe
```diff
@@ -23,22 +23,33 @@ PROVIDES="
...
 
 BUILD()
 {
+	rm -rf site.cfg || true
+	# Check at every update in site.cfg.example
+	echo "[openblas]" >> site.cfg
+	echo "libraries = openblas" >> site.cfg
+	echo "library_dirs = /system/`echo ${relativeDevelopLibDir}`" >> site.cfg
+	echo "include_dirs = /system/`echo ${relativeIncludeDir}`" >> site.cfg
+	echo "runtime_library_dirs = /system/`echo ${relativeLibDir}`" >> site.cfg
```
pulkomandy: Shouldn't this be done in the PATCH() function rather than BUILD() ? It would also make the hack with rm unneeded.

miqlas: Fixed

PR #1087, opened 2017-01-19
Mimetic: new recipe

miqlas:


dev-libs/mimetic/mimetic-0.9.8.recipe
```diff
@@ -0,0 +1,66 @@
...
+PROVIDES_devel="
+	libmimetic${secondaryArchSuffix}_devel = $portVersion
+	devel:libmimetic$secondaryArchSuffix = 0.0.0
+	"
+REQUIRES_devel="
+	libmimetic$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
```
fbrosson: Not really needed, but I think ot would be better to move
```
```
to BUILD_REQUIRES.

dev-libs/mimetic/mimetic-0.9.8.recipe
```diff
@@ -0,0 +1,66 @@
...
+	libmimetic$secondaryArchSuffix = $portVersion
+	lib:libmimetic$secondaryArchSuffix = 0.0.0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libmimetic${secondaryArchSuffix}_devel = $portVersion
+	devel:libmimetic$secondaryArchSuffix = 0.0.0
```
fbrosson: I don't know if this is required, but it looks like we always add a compat.
So maybe you could add a "compat >= 0" here too?

dev-libs/mimetic/mimetic-0.9.8.recipe
```diff
@@ -0,0 +1,66 @@
...
+SOURCE_URI="http://www.codesink.org/download/mimetic-$portVersion.tar.gz"
+CHECKSUM_SHA256="3a07d68d125f5e132949b078c7275d5eb0078dd649079bd510dd12b969096700"
+PATCHES="mimetic-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+PROVIDES="
+	libmimetic$secondaryArchSuffix = $portVersion
+	lib:libmimetic$secondaryArchSuffix = 0.0.0
```
fbrosson: I don't know if this is required, but it looks like we always add a compat.
So maybe you could add a "compat >= 0" here too?

dev-libs/mimetic/mimetic-0.9.8.recipe
```diff
@@ -0,0 +1,66 @@
...
+	cmd:grep
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+BUILD()
+{
+	autoreconf -vfi
```
fbrosson: I guess we don't need --verbose.

dev-libs/mimetic/mimetic-0.9.8.recipe
pulkomandy: No, not always!
You have to check what is actually provided in the package.
If you have only libmimetic.so.0.0.0, then there is no compat.
If you have both libmimetic.so.0.0.0 AND libmimetic.so.0, then you should put a "compat >= 0". It allows to say "anything that links to libmimetic.so.0 will run with this package". Most of the time, other packages will depend on the .0 and be flexible, but sometimes they will need a more specific version (for example, because a bug was fixed), and they can depend on >= 0.0.1, for example.

fbrosson: Thanks @pulkomandy!
(This is obvious, BTW, now that I read your answer ;-)

PR #1089, opened 2017-01-20
Remove old agar recipe, add WIP agar-1.5.0

scottmc:


scottmc:
This is a WIP recipe, but it's getting pretty far.  Having issues trying to get it to detect libjpeg, libfontconfig and a few others.  Also can't figure out where to add in -lnetwork so that it will pass those checks.
Make fails, and not sure what the error message means.  Anyone able to help here?

media-libs/agar/agar-1.5.0.recipe
```diff
@@ -0,0 +1,59 @@
+SUMMARY="A modern open-source, cross-platform toolkit for graphical applications"
+DESCRIPTION="Cross-platform GUI toolkit.  Agar provides a base framework and \
+a collection of GUI widgets from which GUI applications can be built.  Agar \
+can be styled and themed using a CSS-style engine."
+HOMEPAGE="http://libagar.org/index.html.en"
+COPYRIGHT="2002-2016 Hypertrition, Inc."
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="http://stable.hypertriton.com/agar/agar-1.5.0.tar.gz"
```
fbrosson: You might wish to change SOURCE_URI to
```
SOURCE_URI="http://stable.hypertriton.com/agar/agar-$portVersion.tar.gz"
```

media-libs/agar/agar-1.5.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+CHECKSUM_SHA256="82342ded342c578141984befe9318f3d376176e5f427ae3278f8985f26663c00"
+
+ARCHITECTURES="!x86_gcc2 !x86 !x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 !x86"
+
+PROVIDES="
+	cmd:agar
```
fbrosson: Since agar needs libsndfile which does not build with gcc2, we can add this, between SECONDARY_ARCHITECTURES and PROVIDES.
(Borrowed from games-board/stockfish/stockfish-8.recipe)
```
	commandBinDir=$binDir
else
	commandSuffix=
	commandBinDir=$prefix/bin
fi
```
Then you can replace cmd:agar by
```
	cmd:agar$commandSuffix
```

korli:
libjpeg detection can't be worked without configure patching it seems. I don't see any check for libfontconfig. About network, this seems also a problem because the addrinfo check doesn't use any LDFLAGS one could define, so this needs also patching.

media-libs/agar/agar-1.5.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+	devel:libjpeg$secondaryArchSuffix
+	devel:libpng16$secondaryArchSuffix
+	devel:libsdl$secondaryArchSuffix
+	devel:libsndfile$secondaryArchSuffix
+	"
+
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtool$secondaryArchSuffix
```
fbrosson: Should be
```
```

media-libs/agar/agar-1.5.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+	cmd:make
+	cmd:perl
+	"
+	
+BUILD()
+{
+	./configure --prefix=$prefix
```
fbrosson: Should be
```
	./configure --prefix=$prefix --bindir=$commandBinDir
```

media-libs/agar/agar-1.5.0.recipe
scottmc: But it doesn't use libtoolize, it just uses libtool.  This is using bsdbuild instead of the standard gnu autotools.  I left it out and was getting a libtool missing error, so I added it in and the error went away.

scottmc:
Ok, merging the WIP recipe, and will open an issues for this one while I work out a possible patch.

media-libs/agar/agar-1.5.0.recipe
fbrosson: OK.
(I though we were supposed to always use cmd:libtoolize$secondaryArchSuffix instead of cmd:libtool, but I guess there are exceptions.)

Begasus:
Been doing some digging in here, got most dependencies found and package created, need to create a patchset and do some fine-tuning.
For detection I added the search paths for named libraries, for install I changed the path to $dataDir. So far only on 32bit checked.

PR #1094, opened 2017-01-21
botan-2.0.1:new recipe

dacianf:
I get this error when i try to build it http://pastebin.com/EDgJUhUC

### DIFF
--- a/dev-libs/botan/botan-2.0.1.recipe
+++ b/dev-libs/botan/botan-2.0.1.recipe
@@ -1,23 +1,35 @@
 SUMMARY="A cryptography library written in C++"
 DESCRIPTION="This is the best option for crypthography in C++"
 HOMEPAGE="https://botan.randombit.net/"
-COPYRIGHT="2000-2017 botan"
+COPYRIGHT="2000-2017 The Botan Authors"
 LICENSE="BSD (2-clause)"
 REVISION="1"
 SOURCE_URI="https://github.com/randombit/botan/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="f4c02bc4247eb78ac004ce10374992dd488f0d1ba21e712874cd77481332f731"
-SOURCE_DIR="botan-$portVersion"
 
-ARCHITECTURES="!x86_gcc2 !x86 !x86_64"
-SECONDARY_ARCHITECTURES="!x86"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandSuffix=$secondaryArchSuffix
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
 
 PROVIDES="
 	botan$secondaryArchSuffix = $portVersion
-	lib:libbotan_2$secondaryArchSuffix = 0.0.1 compat >= 0
-	cmd:botan$secondaryArchSuffix = $portVersion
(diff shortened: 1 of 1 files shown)


dev-libs/botan/botan-2.0.1.recipe
```diff
@@ -0,0 +1,71 @@
+SUMMARY="A cryptography library written in C++"
+DESCRIPTION="This is the best option for crypthography in C++"
+HOMEPAGE="https://botan.randombit.net/"
+COPYRIGHT="2000-2017 botan"
```
pulkomandy: Why would we reassing the copyright like this? If they say "1999-2013,2014,2015,2016,2017 The Botan Authors", then we should just use that. Reassigning to a single author is not correct.

dev-libs/botan/botan-2.0.1.recipe
```diff
@@ -0,0 +1,71 @@
...
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
+	cmd:make
+	cmd:python
+	"
+PATCH()
+{
+	#	sed -i -e "/^all\>/ s/-pthread //" src/build-data/cc/gcc.txt
+	sed -i -e "/^all\>/ s/-pthread -fstack-protector/-fno-stack-protector/" \
```
pulkomandy: Yes, the stack protection is missing and this leads to build failures (undefined references). We should patch our gcc to tell it how to support this feature.

dev-libs/botan/botan-2.0.1.recipe
```diff
@@ -0,0 +1,71 @@
...
+		src/build-data/cc/gcc.txt
+}
+
+BUILD()
+{
+	runConfigure --omit-buildspec ./configure.py \
+		--with-bzip2 --with-lzma --with-zlib \
+		--with-openssl --with-sqlite3 \
+		--destdir=/
```
pulkomandy: Usually running libtoolize to update libtool files will help with getting shared libraries built. Old libtool versions had a bug that prevented them to detect that shared libs can be built on Haiku.

dev-libs/botan/botan-2.0.1.recipe
fbrosson: Well, according to the authors.txt file, Jack Lloyd is the only author. The next 6 people have just made "extensive and ongoing contributions", so my understanding is that when they contribute they give their copyright to "The Botan Authors"... which is currently Jack Lloyd.

But I guess your suggestion is OK.
BTW, do we squeeze "1999-2013,2014,2015,2016,2017" to "1999-2017"?

dev-libs/botan/botan-2.0.1.recipe
fbrosson: OK, thanks!

dev-libs/botan/botan-2.0.1.recipe
pulkomandy: Well, at least in France you cannot give your authorship away. You can give the copyright, but as soon as you change a line of code, you are part of the "authors".

And yes, we should squeeze the date range, of course.

dev-libs/botan/botan-2.0.1.recipe
fbrosson: Thanks @pulkomandy for the suggestion.
I tried with a "libtoolize --force --copy --install" a few days ago, but it did not help.
I'll push it here, with the fixes I had requested in my [last review](#pullrequestreview-17802804), and maybe we can merge this PR even if the recipe does not yet build the shared lib.

fbrosson:
@dacianf, I assumed you were busy, so I pushed the fixes and merged the recipe.

BTW, now that this PR has been merged, the branch it is based on is no longer needed.

PR #1097, opened 2017-01-21
Lua: Haiku POSIX enough

miqlas:
Required for luarocks

### DIFF
--- a/dev-lang/lua/patches/lua-5.3.3.patchset
+++ b/dev-lang/lua/patches/lua-5.3.3.patchset
@@ -91,6 +91,7 @@ index fd447cc..367f1e7 100644
 +#define LUA_USE_ISATTY
 +#define LUA_USE_POPEN
 +#define LUA_USE_DLOPEN
++#define LUA_USE_POSIX
 +#endif
  
  /*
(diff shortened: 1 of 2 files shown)


dev-lang/lua/patches/lua-5.3.3.patchset
```diff
@@ -1,8 +1,13 @@
-From b1db2db24aca8b1f8a59949f78cc2cbe0d01bfb8 Mon Sep 17 00:00:00 2001
-From: begasus <begasus@gmail.com>
-Date: Sat, 24 Dec 2016 15:45:36 +0100
-Subject: import patches from 5.2.1
+From b7073ba00e3fcc6b6035fe791f19bf59bbb67137 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zolt=C3=A1n=20Mizsei?= <zmizsei@extrowerk.com>
+Date: Sat, 21 Jan 2017 09:11:37 +0100
+Subject: [PATCH] import patches from 5.2.1
```
pulkomandy: I would have preferred a separate commit in the patchset, rather than modifying the existing commit.
This helps tracking down what changed in the patch, and when.

scottmc: Several years ago I had an issue with trying to use lua's default posix option, so had to create one for Haiku.  I believe this is no longer the case, but just wanted to point it out, in case it still matters.
http://ports.haiku-files.org/attachment/ticket/lua-5.1.3-haiku.diff

korli:
I didn't see your PR and enabled the flag in the recipe.

PR #1098, opened 2017-01-21
Updated atk to 2.23.4, deleted old broken recipes

scottmc:
Configure isn't yet detecting the static libs can be made, so may still need to figure that out and then add a devel lib.  But it's building now and passing all of the included tests, which is better than the previous recipes.


dev-libs/atk/atk-2.23.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+COPYRIGHT="2000-2016 fixme"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://ftp.gnome.org/pub/GNOME/sources/atk/2.23/atk-2.23.4.tar.xz"
+CHECKSUM_SHA256="6888f4e902fc4727b1687ff512d5d087050f4c0d11035566484988cf698c2c1a"
+
+ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 ?x86"
+
+PATCHES="atk-$portVersion.patch"
```
fbrosson: Oops, PATCHES is now supposed to go after CHECKSUM_SHA256 ;-)

dev-libs/atk/atk-2.23.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+
+PROVIDES="
+	atk$secondaryArchSuffix=$portVersion
+	lib:libatk_1.0=$portVersion
```
fbrosson: Should be
```
	lib:libatk_1.0 = 0.22313.1 compat >= 0
```
BTW, the previous line would look nicer with spaces around the "=" ;-)

dev-libs/atk/atk-2.23.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	runConfigure ./configure
```
fbrosson: I guess you want
```
	runConfigure ./configure --enable-static
```
Upstream disabled static libs by default:
```
./configure --help | grep static
--enable-static[=PKGS]  build static libraries [default=no]
```

dev-libs/atk/atk-2.23.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+	make
+}
+
+INSTALL()
+{
+	make install
+	
+	rm $libDir/libatk-1.0.la
+	
+	fixPkgconfig
```
fbrosson: I think this is only needed if you have a _devel package.
BTW, I guess it would be nice to create it.

dev-libs/atk/atk-2.23.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+
+TEST()
+{
+	make check
+	cd tests
+	teststateset
+	testrelation
+	testrole
+	testvalue
+	testdocument
```
fbrosson: I think it would be better to prepend a "./" in front of each test\*, just in case we drop "." from PATH. BTW, I don't know why we have "." in PATH by default...

dev-libs/atk/atk-2.23.4.recipe
pulkomandy: Why do we want it? Shared libraries are what should be used.

fbrosson: I don't want static libs. (I even remove them from my packages after building and before installing.)
I just gave @scottmc the "trick" to have the static lib built, since I understood he want's it.
Of course, we still get the shared libs, even with "--enable-static".

scottmc:
This one should be ok now.

PR #1105, opened 2017-01-22
bash: bump to 4.4.011.

fbrosson:


waddlesplash:
I second @korli on that.

fbrosson:
Thanks @korli & @waddlesplash.
Yep, I knew it would be OK because I have successfully built it on each arch and even installed it on x86_64.
But since I did not install it on x86_gcc2 and x86 (because of the dependency on lib:libreadline >= 7 which would triggered the uninstallation of a lot of packages), I chose to create a PR.

BTW, "hp --test bash-4.4.011" has the same hick-ps as  "hp --test bash-4.4.005".

Issue #1101, opened 2017-01-22
links: build is broken


pulkomandy:
It depends on libpng instead of libpng16.

fbrosson:
Fixed in #1106.

PR #1106, opened 2017-01-22
links: use libpng16 instead of libpng.

fbrosson:
* Drop unneeded support for secondary architectures.
* Depend on libssl and libcrypto instead of openssl.
* Use **`$effectiveTargetArchitecture`** (instead of **`$targetArchitecture`**) when defining STDCPPLIBS (as **`-lstdc++`** or *``lbstdc++.r5`**) in order to have a working code for both primary and secondary architectures,
  should that code snippet be copied to other recipes.
* Add TEST() with **`make check`**.

### DIFF
--- a/www-client/links/links-2.8.recipe
+++ b/www-client/links/links-2.8.recipe
@@ -1,54 +1,55 @@
 SUMMARY="A graphics and text mode web browser"
 DESCRIPTION="Links is a multi-platform web browser you can run in Terminal."
 HOMEPAGE="http://links.twibright.com/"
-COPYRIGHT="1999 - 2011 Mikulas Patocka
-	2000 - 2011 Petr Kulhavy, Karel Kulhavy, Martin Pergel"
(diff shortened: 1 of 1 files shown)


fbrosson:
This PR fixes #1101. Tested on x86_gcc2. Both links and links_x86 build and run fine.
I have a few questions:
1) Should I drop support for secondary architectures (and the CONFLICTS I added)?
2) Should I replace the dependency on openssl by a dependency on libssl and/or libcrypto?
3) Should I bump revision?

korli:
1. why was support for secondary architectures present in the first place?
2. it depends, does links also build with openssl 1.0.2?
3. yes, this changes the resulting package.

fbrosson:
OK, thanks. I've updated the recipe.

1) I I don't know. I've just removed the support for 2ndary arch.
2) Yes, it builds fine with openssl 1.0.2. Dependencies changed to libssl and libcrypto.
3) done.

PR #1111, opened 2017-01-24
m4: bump to 1.4.18.

fbrosson:
* Switch HOMEPAGE & SOURCE_URI to https.
* ~~Add missing cmd:{cmp,diff,help2man} to BUILD_PREREQUIRES as TEST() needs them. Sort BUILD_PREREQUIRES while at it.~~
* ~~BTW, unlike 1.4.17, 1.4.18 fails to build (on Haiku) without help2man.~~
* Do not pass **`--disable-rpath`** to configure, otherwise the build fails if **`cmd:help2man`** is not in BUILD_PREREQUIRES.
* Define TEST_REQUIRES with **`cmd:{cmp,diff,help2man}`**. It might be used by HaikuPorter in the future.
* Remove duplicate **`cmd:awk`** in BUILD_PREREQUIRES (and sort it).

### DIFF
--- a/sys-devel/m4/m4-1.4.18.recipe
+++ b/sys-devel/m4/m4-1.4.18.recipe
@@ -11,12 +11,12 @@ integer arithmetic, manipulating text in various ways, recursion, etc. M4 can \
 be used either as a front-end to a compiler or as a macro processor in its own \
 right.
 One of the biggest users of M4 is the GNU Autoconf project."
-HOMEPAGE="http://www.gnu.org/software/m4/"
-COPYRIGHT="2000, 2005-2011 Free Software Foundation, Inc."
+HOMEPAGE="https://www.gnu.org/software/m4/"
+COPYRIGHT="2000, 2005-2016 Free Software Foundation, Inc."
 LICENSE="GNU GPL v3"
-REVISION="3"
-SOURCE_URI="http://ftp.gnu.org/gnu/m4/m4-$portVersion.tar.gz"
-CHECKSUM_SHA256="3ce725133ee552b8b4baca7837fb772940b25e81b2a9dc92537aeaf733538c9e"
+REVISION="1"
+SOURCE_URI="https://ftp.gnu.org/gnu/m4/m4-$portVersion.tar.gz"
+CHECKSUM_SHA256="ab2633921a5cd38e48797bf5521ad259bdc4b979078034a3b790d7fec5493fab"
 ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
 
 PROVIDES="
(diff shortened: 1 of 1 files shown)


fbrosson:
There is probably some way to build 1.4.18 without help2man.
I haven't investigated much but can try to, in case we don't want m4 to _build-depend_ on help2man.
BTW, I guess it would not be acceptable to have m4 _build-depend_ on cmd:{aclocal,autoconf,automake}, since automake itself depends on m4.
If the recipe is OK as is, i.e. with a build-dependency on help2man, please let me lnow ;-)
Thanks!

korli:
Weird, looking at the commits since 1.4.17, I don't see why help2man would be now necessary. Maybe a timestamp problem?
Otherwise, the usual problem of adding test requirements in BUILD_PREREQUIRES. Another section would be useful.

fbrosson:
You are probably right, and I guess that not passing $jobArgs (to make) will allow to build without help2man.
(I'll check asap.)

And yes, having a TEST_REQUIRES (or anything better) will allow to better test lots of packages which we don't already test, due to unwanted build dependencies.
BTW, I'd say that such a TEST_REQUIRES should only list what is needed but not **_already_** in BUILD_PREREQUIRES. (But IMO it should not error it overlaps BUILD_PREREQUIRES and/or BUILD_REQUIRES.)

pulkomandy:
You can do this:

```
TEST_REQUIRES = "..."

BUILD_REQUIRES = "...
    $TEST_REQUIRES
"
```

At least then it is clear which parts are only needed for the tests. We can later integrate the feature into HaikuPorter and start using it immediately, and remove the $TEST_REQUIRES line as we go.

Note that BUILD_PREREQUIRES makes no sense for the tests, as these can't be run in a cross-compiling environment.

fbrosson:
Hmm, contrary to what I was suspecting, not passing $jobArgs did not help.
After some investigation it turns out that there are 2 easy ways to fix the build:
* Add cmd:help2man to BUILD_PREREQUIRES. This is what I had suggested previously, but I guess we don't to add a build dependency on cmd:help2man.
* Avoid passing --disable-rpath to configure. I guess this would be OK. (Confirmation welcome ;-)

Regarding the TEST_REQUIRES, I'd rather add it **_after_** BUILD_PREREQUIRES and before INSTALL. It would be a bit weird to put it **_before_** BUILD_REQUIRES, don't you think? Shall we just define TEST_REQUIRES (and not include it in BUILD_REQUIRES), in order to keep BUILD_{,PRE}REQUIRES unchanged?
Once HaikuPorter is taught what to do with TEST_REQUIRES it will be possible to "hp --test m4" without even having to change this recipe.
What do you think?

scottmc:
+1 for adding a TEST_REQUIRES for case like this.  I've seen a small number
of case where this would have been handy to have.

-Scott

PR #1116, opened 2017-01-27
Luarocks bump version, recipe cleanup

miqlas:
Cannot find the lua.h yet during the configure process, any help really appreciated.

### DIFF
--- a/dev-lua/luarocks/luarocks-2.4.2.recipe
+++ b/dev-lua/luarocks/luarocks-2.4.2.recipe
@@ -7,7 +7,6 @@ LICENSE="MIT"
 REVISION="2"
 SOURCE_URI="http://luarocks.org/releases/luarocks-$portVersion.tar.gz"
 CHECKSUM_SHA256="0e1ec34583e1b265e0fbafb64c8bd348705ad403fe85967fd05d3a659f74d2e5"
-SOURCE_DIR=$portVersionedName
 
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 
@@ -44,7 +43,7 @@ BUILD()
 		--rocks-tree=$dataDir/luarocks \
 		--versioned-rocks-dir \
 		--force-config
-		
+
 	make build $jobArgs
 }
 INSTALL()
--- a/dev-lua/luarocks/luarocks-2.4.2.recipe
+++ b/dev-lua/luarocks/luarocks-2.4.2.recipe
@@ -1,0 +1,53 @@
+SUMMARY="A deployment and management system for Lua modules"
+DESCRIPTION="LuaRocks allows you to install modules as self-contained \
+packages called rocks, which also contain version dependency information."
+HOMEPAGE="http://luarocks.org/"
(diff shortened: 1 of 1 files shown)


dev-lua/luarocks/luarocks-2.4.2.recipe
```diff
@@ -0,0 +1,53 @@
+SUMMARY="A deployment and management system for Lua modules"
+DESCRIPTION="LuaRocks allows you to install modules as self-contained \
+packages called rocks, which also contain version dependency information."
+HOMEPAGE="http://luarocks.org/"
+COPYRIGHT="2007-2016 Kepler Project"
+LICENSE="MIT"
+REVISION="2"
+SOURCE_URI="http://luarocks.org/releases/luarocks-$portVersion.tar.gz"
+CHECKSUM_SHA256="0e1ec34583e1b265e0fbafb64c8bd348705ad403fe85967fd05d3a659f74d2e5"
+SOURCE_DIR=$portVersionedName
```
fbrosson: You can remove the "SOURCE_DIR=$portVersionedName" because that's the default value when haikuporter builds a recipe for a primary architecture.

BTW, I would recommend to avoid using $portVersionedName in a recipe because:
* $portVersionedName is the same as $portName + $portVersion
* and $portName ends with $secondaryArchSuffix which, as we know, is empty for primary arch and not empty for secondary arch.

This means that the places where we could technically use $portVersionedName (or $portName) in recipes **without** support for secondary architecturess.. would almost **never** be the same as the places where we might use $portVersionedName (or $portName) in recipes that **do have** support for secondary architecturess.

In short, and to keep consistency between all recipes,  $portVersionedName (and $portName) should never be used in SOURCE_DIR (and probably anywhere else too.)

BTW, I would also remove cmd:wget from BUILD_PREREQUIRES.
If the build process needs to download something, it should be downloaded **before** BUILD, with a SOURCE_URI_x.

miqlas:
It still cannot find lua.h

pulkomandy:
Can't it use pkg-config to locate lua?
Have you checked what it says in configure.log or similar, are there additional details?

dev-lua/luarocks/luarocks-2.4.2.recipe
```diff
@@ -0,0 +1,52 @@
...
+	cmd:md5sum
+	cmd:sed
+	cmd:wget
+	"
+
+BUILD()
+{
+	#not an autotools package
+	./configure --prefix=$prefix  \
+		--with-lua-include="$includeDir/lua53" \
```
Begasus: This should do the trick, as discussed on IRC :)
LUA_VERSION=$(lua -v | sed 's/Lua //' | head -c3 | tr -d '.')
./configure --prefix=$prefix --with-lua-include="$(finddir B_SYSTEM_DEVELOP_DIRECTORY)/headers/lua$LUA_VERSION"

Begasus:
the package still installs a "share" folder which fails when running haikuporter in strict mode ...

PR #1117, opened 2017-01-27
ATLAS : WIP recipe

miqlas:
Somehow broken in HaikuPorter, works outside.


sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,133 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	binutils$secondaryArchSuffix
+	cmd:awk
+	cmd:gfortran$secondaryArchSuffix
+	cmd:fgrep
```
fbrosson: You should probably use **`cmd:grep`** instead of **`cmd:fgrep`** because **fgrep is deprecated**, as we can see in this paragraph of the manual page:
```
       In addition, the variant programs egrep, fgrep and rgrep are
       the  same  as  grep -E,  grep -F, and grep -r, respectively.
       These variants are deprecated, but are provided for backward
       compatibility.
```
Of course, you should also replace every call to **`fgrep`** by **`grep -F`**.
BTW, this is the contents of **`/bin/fgrep`**:
```
#!/bin/sh
exec grep -F "$@"
```
Even if it's almost sure grep will continue to provide the fgrep variant, it should be avoided like any other deprecated stuff.

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,133 @@
+SUMMARY="Automatically Tuned Linear Algebra Software"
+DESCRIPTION="A software library for linear algebra. It provides a mature open \
+source implementation of BLAS APIs for C and Fortran77."
+HOMEPAGE="http://math-atlas.sourceforge.net/"
+COPYRIGHT="2011-2016 The ATLAS Project"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://downloads.sf.net/math-atlas/atlas3.10.3.tar.bz2"
```
fbrosson: You might wish to use $portVersion in SOURCE_URI:
```
SOURCE_URI="https://downloads.sf.net/math-atlas/atlas$portVersion.tar.bz2"
```

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+	binutils$secondaryArchSuffix
+	cmd:awk
+	cmd:bzip2
+	cmd:gfortran$secondaryArchSuffix
+	cmd:fgrep
+	cmd:find
+	cmd:g++$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	cmd:pwd
```
fbrosson: You don't need **`cmd:pwd`** in BUILD_PREREQUIRES since it is provided by coreutils.

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+	make build
+
+	echo 'Building shared libraries'
+	cd lib
+	if [ 1 -lt $NCPU ]; then
+		cp "$srcdir/makefile.shared.mt" makefile
+	else
+		cp "$srcdir/makefile.shared.st" makefile
+	fi
+	make -f makefile
```
fbrosson: You might wish to try replacing
```
	cd lib
	if [ 1 -lt $NCPU ]; then
		cp "$srcdir/makefile.shared.mt" makefile
	else
		cp "$srcdir/makefile.shared.st" makefile
	fi
	make -f makefile
```
by
```
	if [ "$NCPU" -gt 1 ]; then
		make -C lib -f ../makefile.shared.mt
	else
		make -C lib -f ../makefile.shared.st
	fi
```

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,145 @@
...
+HOMEPAGE="http://math-atlas.sourceforge.net/"
+COPYRIGHT="2011-2016 The ATLAS Project"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://downloads.sf.net/math-atlas/atlas$portVersion.tar.bz2"
+CHECKSUM_SHA256="2688eb733a6c5f78a18ef32144039adcd62fabce66f2eb51dd59dde806a6d2b7"
+SOURCE_DIR="ATLAS"
+PATCHES="atlas-$portVersion.patchset"
+ADDITIONAL_FILES="makefile.shared.mt
+	makefile.shared.st"
```
fbrosson: Instead of adding copies of these files in the HaikuPorts repo, you could use this:
```
archlinuxCommit="286cdb8eafb3756cdb3d63f0f6c6084e55a31169"
archlinuxRepoBaseUri="https://aur.archlinux.org/cgit/aur.git/plain/"
SOURCE_FILENAME_2="makefile.shared.mt"
SOURCE_FILENAME_3="makefile.shared.st"
SOURCE_URI_2="$archlinuxRepoBaseUri/$SOURCE_FILENAME_2?h=atlas-lapack&id=$archlinuxCommit"
SOURCE_URI_3="$archlinuxRepoBaseUri/$SOURCE_FILENAME_3?h=atlas-lapack&id=$archlinuxCommit"
CHECKSUM_SHA256_2="7627446a0d598b55bc8f2795f8ff2028f134928b9ca597b76bf8dd5622dbff69"
CHECKSUM_SHA256_3="18cfd59a104451c03c3b91020786260cf2b2581642a9c8f8598ba1d26b9cb3bf"
```
* makefile.shared.mt would be available as $sourceDir2/$SOURCE_FILENAME_2
* makefile.shared.st would be available as $sourceDir3/$SOURCE_FILENAME_3

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,145 @@
...
+
+	echo 'Building ATLAS'
+	make build $jobArgs
+
+	echo 'Building shared libraries'
+	cd lib
+	if [ 1 -lt $NCPU ]; then
+		cp "$portDir/additional-files/makefile.shared.mt" makefile
+	else
+		cp "$portDir/additional-files/makefile.shared.st" makefile
```
fbrosson: If you drop ADDITIONAL_FILES (with the suggestion above), you just need to use this here:
```
	if [ 1 -lt $NCPU ]; then
		cp "$sourceDir2/$SOURCE_FILENAME_2" makefile # shared.mt
	else
		cp "$sourceDir3/$SOURCE_FILENAME_3" makefile # shared.st
	fi
```

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,145 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	binutils$secondaryArchSuffix
+	cmd:awk
+	cmd:bzip2
+	cmd:gfortran$secondaryArchSuffix
+	cmd:fgrep
```
fbrosson: (This is a copy of one of the remarks I made on my previous review which is hiddent probably because the line to which it was attached was moved. It would not have happened if I had aborted my previous review (instead of just clicking "refresh") and rewritten it when I noticed there had been a new update to the PR. I'll be careful next time ;-)

You should probably use **`cmd:grep`** instead of **`cmd:fgrep`** because **fgrep is deprecated**, as we can see in this paragraph of the manual page:
```
       In addition, the variant programs egrep, fgrep and rgrep are
       the  same  as  grep -E,  grep -F, and grep -r, respectively.
       These variants are deprecated, but are provided for backward
       compatibility.
```
Of course, you should also replace every call to **`fgrep`** by **`grep -F`**.
BTW, this is the contents of **`/bin/fgrep`**:
```
#!/bin/sh
exec grep -F "$@"
```
Even if it's almost sure grep will continue to provide the fgrep variant, it should be avoided like any other deprecated stuff.

miqlas: The build tool explicitly calls fgrep. From C. With different kind of magic.
So yeah... It must be fgrep. I won't change it, for sure.

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,145 @@
...
+	lib:libgomp$secondaryArchSuffix
+	"
+CONFLICTS="
+	blas$secondaryArchSuffix
+	cblas$secondaryArchSuffix
+	lapack$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	atlas${secondaryArchSuffix}_devel = $portVersion compat >=3
```
fbrosson: For **`atlas${secondaryArchSuffix}_devel`** there should be no **`compat`** at all.

sci-libs/atlas/atlas-3.10.3.recipe
```diff
@@ -0,0 +1,145 @@
...
+
+BUILD()
+{
+	# Needs to be deleted!
+	rm -rf haiku_build || true
+	mkdir -p haiku_build
+	cd haiku_build
+
+	CORE=`sysinfo -cpu | grep -o "\w*MHz\w*" | grep -o '[0-9]*'` # CPU freq
+	NCPU=`sysinfo -cpu | fgrep "MHz" | awk '{print $1}'` # CPU count
```
fbrosson: > The build tool explicitly calls fgrep. From C. With different kind of magic.
So yeah... It must be fgrep. I won't change it, for sure.

OK.
BTW, I though fgrep was only being called from the recipe's BUILD()

miqlas:
@fbrosson : thanks for all the ideas, i'll inmplement them as fast as possible.

To everybody: i'm trying to get my patches upstreamed. ATLAS currently technically capable to get the required informations about the system to create an optimized build, so a generic x86_SSE2 build in the HaikuDepot wouldn't be too nice and actually impossible right now, because defining the arch, microarch, and so on drives the compiler crazy and i cannot debug, why. I suppose it runs some synthetic tests, but "none survives" (yeah, ATLAS told me). Maybe they can fix it in the next release, who knows.

So if ATLAS can create optimized build, then i would propose the following: we could disable this recipe for every platform, and everybody, who needs it, can compile it to get a machine optimized lib.
It would make the recipe lot cleaner and easier to maintain, but of course, everything, what using ATLAS would be machine specific.

This is my preferred way to close this case, but i'm interested what others thinks. Any thoughts?

@humdingerb : I know, this is again something what nobody would use. Never. Ever. Actually that's why i inveted time into this.

sci-libs/atlas/atlas-3.10.3.recipe
miqlas: WONTFIX

miqlas:
Disabled on every arch, to tidy the PRs up i merge this, as nobody had any input here since loong.

PR #1118, opened 2017-01-27
radare2: bump version to 1.1.0

return:
Tested on both x86 and x86_64, but not x86_gcc2. This patch fixes issues with the login_pty symbol when running radare2.

### DIFF
--- a/dev-util/radare2/radare2-1.1.0.recipe
+++ b/dev-util/radare2/radare2-1.1.0.recipe
@@ -8,16 +8,18 @@ COPYRIGHT="2009-2015 nibble, pancake"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
 REVISION="1"
-SOURCE_URI="http://radare.org/get/radare2-$portVersion.tar.xz"
-CHECKSUM_SHA256="024adba5255f12e58c2c1a5e2263fada75aad6e71b082461dea4a2b94b29df32"
+SOURCE_URI="http://cloud.radare.org/get/1.1.0/radare2-$portVersion.tar.gz"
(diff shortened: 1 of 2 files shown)


dev-util/radare2/radare2-1.1.0.recipe
```diff
@@ -8,16 +8,18 @@ COPYRIGHT="2009-2015 nibble, pancake"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
 REVISION="1"
-SOURCE_URI="http://radare.org/get/radare2-$portVersion.tar.xz"
-CHECKSUM_SHA256="024adba5255f12e58c2c1a5e2263fada75aad6e71b082461dea4a2b94b29df32"
+SOURCE_URI="http://cloud.radare.org/get/1.1.0/radare2-$portVersion.tar.gz"
```
pulkomandy: Please use $portVersion rather than 1.1.0 here.

dev-util/radare2/radare2-1.1.0.recipe
```diff
@@ -93,11 +98,6 @@ INSTALL()
 {
 	make install
 	cp libr/*/libr_*.so $libDir
-
-	mkdir -p `dirname $docDir`
-	mv $prefix/share/doc $docDir
-	rm -rf $prefix/share
```
pulkomandy: Are the docs properly installed in $docDir now (ie is that fixed upstream?)

return: I don't think this is fixed upstream, Line 98 causes the build to fail in 1.1.0  when moving the docs to $docDir.

return: But moving the directory from $prefix/data/doc to $docDir puts the documentation under documentation/packages/radare2/radare2/  

Is this correct documentation directory to place the html/markdown files?

PR #1119, opened 2017-01-28
Updated libusb recipe to latest and suggested cosmetic fixes

Vidrep:
I went step by step through each suggested change and tested the build each time for breakage.
The recipe builds on x86_gcc2.

### DIFF
--- a/dev-libs/libusb/libusb-1.0.21.recipe
+++ b/dev-libs/libusb/libusb-1.0.21.recipe
@@ -11,20 +11,21 @@ HOMEPAGE="http://libusb.info/"
 COPYRIGHT="2001 Johannes Erdfelt
 	2007-2009 Daniel Drake
 	2010-2012 Peter Stuge
-	2008-2013 Nathan Hjelm
+	2008-2016 Nathan Hjelm
 	2009-2013 Pete Batard
 	2009-2013 Ludovic Rousseau
 	2010-2012 Michael Plante
 	2011-2013 Hans de Goede
 	2012-2013 Martin Pieuchot
-	2012-2013 Toby Gray"
+	2012-2013 Toby Gray
+	2013-2015 Chris Dickens"
 LICENSE="GNU LGPL v2.1"
-REVISION="2"
-SOURCE_URI="https://github.com/libusb/libusb/archive/v1.0.20.tar.gz"
+REVISION="1"
+SOURCE_URI="https://github.com/libusb/libusb/archive/v$portVersion.tar.gz"
 SOURCE_FILENAME="libusb-$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


fbrosson:
Thanks @Vidrep for the update.
Everything looks OK in this recipe.
You just forgot to remove the recipe for the previous release, and you used master instead of a separate branch, but we can workaround all this ;-)
```
cd haikuports/dev-libs/libusb
```
Now, to make sure you will be able to update your commit with a "git commit --amend" (in order to avoid having to make complicated rebase actions), can you check that your commit is the last one you have in your local master?
To check this, you can compare the results of the following command
```
git log | head -5 | sed "2s/<.*>//"
```
with
```
commit 5a2505b1a10da2ea710343551a8dc008469262a2
Author: Vidrep 
Date:   Fri Jan 27 18:49:32 2017 -0700

    Updated libusb recipe to latest and suggested cosmetic fixes
```
If your commit hash does not match but all the other lines match, than this means your commit is till the last one, so that's still OK.

Now, if everything is OK, you can:
```
git rm libusb-1.0.20.recipe
git commit --amend --no-edit libusb-1.0.20.recipe
```
then you should push your commit to your github remote repo with the -f (for force) option.
If you don't know how to do that, send me the result of this command:
```
git remote -v | grep vidrep
```
If you get
```
origin	https://github.com/vidrep/haikuports.git (fetch)
origin	https://github.com/vidrep/haikuports.git (push)
```
then the git command to run will be:
```
git push -f
```
otherwise you'll need to tell me what "git remove -v | grep vidrep" gave you.

dev-libs/libusb/libusb-1.0.21.recipe
```diff
@@ -0,0 +1,85 @@
...
+	2012-2013 Martin Pieuchot
+	2012-2013 Toby Gray
+	2013-2015 Chris Dickens"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://github.com/libusb/libusb/archive/v$portVersion.tar.gz"
+SOURCE_FILENAME="libusb-$portVersion.tar.gz"
+CHECKSUM_SHA256="1a5b08c05bc5e38c81c2d59c29954d5916646f4ff46f51381b3f624384e4ac01"
+ARCHITECTURES="x86 x86_64 x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
```
fbrosson: Hmm, we can add x86_gcc2 here ;-)
```
```

@Vidrep, after you change libusb-1.0.21.recipe, you can:
```
git commit --amend --no-edit libusb-1.0.21.recipe
```
**REMINDER:** You should only use "--amend" if the commit on the top of your git history is the one you are using to update libusb.

Vidrep:
Should be good now. Please check and comment. Thanks.

PR #1122, opened 2017-01-30
keystone: create recipe

return:
Initial keystone recipe, tested on x86 but not on x86_64 but should work on both platforms.

### DIFF
--- a/dev-util/keystone/keystone-0.9.1.recipe
+++ b/dev-util/keystone/keystone-0.9.1.recipe
@@ -1,0 +1,72 @@
+SUMMARY="A Lightweight multi-platform, multi-architecture assembly framework"
+DESCRIPTION="* Multi-architecture: Arm, (AArch64/Armv8), Mips, PowerPC, Sparc, \
+SystemZ, XCore & X86 (include X86_64) (details).
+* Clean/simple/lightweight/intuitive architecture-neutral API.
+* Implemented in C/C++ language, with bindings for Powershell Perl, Python, \
+Ruby, C#, NodeJS, Java, GO, C++, OCaml, Lua, Rust, Delphi, Free Pascall & Vala available.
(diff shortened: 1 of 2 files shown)


dev-util/keystone/keystone-0.9.1.recipe
```diff
@@ -0,0 +1,73 @@
+SUMMARY="A Lightweight multi-platform, multi-architecture assembly framework"
+DESCRIPTION="* Multi-architecture: Arm, (AArch64/Armv8), Mips, PowerPC, Sparc, \
+SystemZ, XCore & X86 (include X86_64) (details). \
+* Clean/simple/lightweight/intuitive architecture-neutral API. \
```
korli: you probably don't need a \ when the next line starts with a *.

dev-util/keystone/keystone-0.9.1.recipe
```diff
@@ -0,0 +1,73 @@
...
+SOURCE_URI="https://github.com/keystone-engine/keystone/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="e9d706cd0c19c49a6524b77db8158449b9c434b415fbf94a073968b68cf8a9f0"
+PATCHES="keystone-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	keystone$secondaryArchSuffix = $portVersion
+	lib:libkeystone$secondaryArchSuffix = 0.9
```
korli: Arch Linux seems to only provide libkeystone.so.0, are you sure about 0.9?

korli:
There seems to be python bindings in bindings/python, they could end in a subpackage if you wish. Nothing that should prevent merging this PR anyway :)

PR #1127, opened 2017-01-31
lbreakout2: fix build dependency on libpng16.

fbrosson:


scottmc:
Can you check the others listed in the issue and is any of those have the same issue add them to this same pull request?

fbrosson:
* links-2.8 already build-depends on libpng16.
* libwebp already depends on libpng16.
* libusb does not depend on (any version of) libpng.

PR #1134, opened 2017-02-02
Freetype: add patch to activate subpixel rendering

humdingerb:
PATCHES is still commented out in the recipe. Added for easy testing.

### DIFF
--- a/media-libs/freetype/freetype-2.7.1.recipe
+++ b/media-libs/freetype/freetype-2.7.1.recipe
@@ -8,6 +8,7 @@ LICENSE="FreeType"
 REVISION="1"
 SOURCE_URI="https://download.savannah.gnu.org/releases/freetype/freetype-$portVersion.tar.bz2"
 CHECKSUM_SHA256="3a3bb2c4e15ffb433f2032f50a5b5a92558206822e22bfe8cbe339af4aa82f88"
+#PATCHES="freetype-subpixel.patchset" # enable patented subpixel rendering
 
 ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
 SECONDARY_ARCHITECTURES="x86_gcc2 x86"
--- a/media-libs/freetype/patches/freetype-subpixel.patchset
+++ b/media-libs/freetype/patches/freetype-subpixel.patchset
@@ -1,0 +1,22 @@
+From fd9b7dd32ca353efef65a47e508dd48149cfb278 Mon Sep 17 00:00:00 2001
+From: Humdinger <humdingerb@gmail.com>
+Date: Thu, 2 Feb 2017 08:50:24 +0100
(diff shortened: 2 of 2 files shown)


media-libs/freetype/freetype-2.7.1.recipe
```diff
@@ -8,6 +8,7 @@ LICENSE="FreeType"
 REVISION="1"
 SOURCE_URI="https://download.savannah.gnu.org/releases/freetype/freetype-$portVersion.tar.bz2"
 CHECKSUM_SHA256="3a3bb2c4e15ffb433f2032f50a5b5a92558206822e22bfe8cbe339af4aa82f88"
+#PATCHES="freetype-2.7.1-subpixel.patchset" # enable patented subpixel rendering
```
waddlesplash: You can just name it "freetype-subpixel", it should still work on future versions, too.

pulkomandy:
Can't we make a freetype_subpixel package that conflicts with the normal freetype one? This way we can even put it in the depot (or an alternate repository).

Also you apparently forgot to git add the patch after renaming (pro-tip: use "git mv").

humdingerb:
fixed the patch filename. next time with "git mv".. :)
I'm not sure how to create a "conflicting" package.

waddlesplash:
As I noted on the mailing list:
> 1. Microsoft's cleartype parents are actually generic and apply to all subpixel anti aliasing and not just fonts, and we use AGG as our system graphics engine, which does SPAA.
> 2. We ship an unmodified FFmpeg, which has *A LOT* of *definitely* patented code in it (mp3, mp4, ...), and MPEG LA has sent C&Ds numerous times before, as opposed to Microsoft who to my knowledge has never actually tried to enforce the ClearType patents. (Ubuntu, for instance, has to pay MP3 patent licenses at least -- I don't think they pay ClearType ones, despite having it enabled.)
> 3. The parents either expired last year, or will expire in 2018-2020 (as they issued in 2001-2003.)

So is there any real point to disabling FreeType subpixel rendering, when we already most likely infringe on a number of patents, including the ClearType ones already?

mmuman:
Why not disable it depending on the locale settings while we are at it? 😱

humdingerb:
@mmuman , in your case, it's not even "patches welcome", but "commits welcome" :)

In general, it wouldn't hurt to provide non-patented version of all those "contentious" packages. If anyone wants to work on that. In this case it's so exeedingly easy, with only one line of the config...

korli:
The package manager needs nonetheless a way to know which package to install between two providing the exact same lib:libfreetype provide. freetype should also be compiled against harfbuzz in the future, which is also not straightforward.

pulkomandy:
I'm merging this so we have the patch at hand. We can decide how to use it later.

PR #1141, opened 2017-02-05
IlmBase: remove recipes for old releases.

fbrosson:


fbrosson:
Should I update this PR to also remove ilmbase-2.1.0 and openexr-2.1.0?
(We already have ilmbase-2.2.0 and openexr-2.2.0.)

PR #1142, opened 2017-02-06
Fish: radical reduced build-time deps, thanks zanchey!

miqlas:
Now with less "unknown version" and with more "2.5.0" for free! Just today! Call us: 0900-miqlas!


miqlas:
Now with less "unknown version" and with more "2.5.0" for free! Just today! Call us: 0900-miqlas!

PR #1149, opened 2017-02-12
fix secondary arch _x86, switch to static build

tidux:
I disabled the --with-modules flag that split the commands and protocols out as .so files because LFTP was looking in the wrong directories for them.  The hpkg generated from this recipe is tested and working on hrev50951 gcc2h.


korli:
It seems Pulkomandy updated the recipe already. Sorry for that.

PR #1154, opened 2017-02-17
wxQt: enable positional printf argument support for FileZilla

miqlas:


scottmc:
Don't forget to update revision number when packages have been built using previous revision of the recipes.

korli:
I suppose these options are required by a package using wxwidgets. Please give this information in the commit message.

PR #1165, opened 2017-02-22
zeal: new recipe

return:
[Zeal](https://zealdocs.org/) is a documentation browser just like [Dash Docs](https://kapeli.com/dash) iOS and MacOS. Strangely, apps compiled with Qt do run but no windows are visible. Tested only on x86 but should work on x86_64.

### DIFF
--- a/app-doc/zeal/zeal-0.3.1.recipe
+++ b/app-doc/zeal/zeal-0.3.1.recipe
@@ -1,0 +1,59 @@
+SUMMARY="A simple offline documentation browser inspired by Dash"
+DESCRIPTION="Zeal is a cross-platform, open source, offline \
+documentation browser for software developers, inspired by Dash Docs \
+for macOS and iOS. Zeal has over 195+ docsets and are all compatible \
+with Dash docsets. "
+HOMEPAGE="https://zealdocs.org"
+COPYRIGHT="2013, 2017 Oleg Shparber"
+LICENSE="GNU GPL v3"
+REVISION="1"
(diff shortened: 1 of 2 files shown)


waddlesplash:
> Strangely, apps compiled with Qt do run but no windows are visible. 

??? Does this mean that this is useless?

app-doc/zeal/zeal-0.3.1.recipe
```diff
@@ -0,0 +1,59 @@
...
+with Dash docsets. "
+HOMEPAGE="https://zealdocs.org"
+COPYRIGHT="2013, 2017 Oleg Shparber"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/zealdocs/zeal/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="55f8511977818612e00ae87a4fddaa346210189531469690f2e3961bb4c2c318"
+PATCHES="zeal-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
```
korli: SECONDARY_ARCHITECTURES is missing.

return:
@waddlesplash No, it does indeed work but the issue is directed at Qt5 apps which I am having problems with. I am using Qt5-5.5.1-3 on HaikuDepot.

This is what I see went I try to run both Zeal and other Qt5 apps.

korli:
You need to check your installed fonts: in the Fonts preferences you'll probably see no Sans font is found.

return:
Thanks @korli!  I switched from Bitstream Charter to DejaVu Sans and all Qt5 apps work.

PR #1167, opened 2017-02-23
New recipe: extra-cmake-modules

miqlas:


kde-frameworks/extra-cmake-modules/extra_cmake_modules-5.31.0.recipe
```diff
@@ -0,0 +1,57 @@
...
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	extra_cmake_modules = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libQt5Core$secondaryArchSuffix
```
waddlesplash: Is this really a REQUIRES?

miqlas: Don't think so.

waddlesplash: Then you can remove it. :)

miqlas:
Already did it.

PR #1168, opened 2017-02-24
erlang: fix build

return:
This commit fixes a compilation issue with the ifaddrs.h library that Erlang needs. Tested on x86 only. x86_gcc2 needs to be tested.


return:
@korli I've amended the original commit message for more clarity on the changes made.

PR #1169, opened 2017-02-25
mpg123: enable SDL output

miqlas:
there is no vital recipe depending on mpg123


korli:
from Pulkomandy: 

> There is a reason for mpg123 not depending on SDL. It would bring the SDL dependency to all the things that use mpg123, and I think SDL itself uses it in some case (but I could remember wrong).
If adding it, we should make sure that the SDL-specific part is split into a separate package (mpg123_sdl? mpg123_tools?) and that the libraries can be used without the SDL dependency being pulled in."

mpg123_tools sounds good.

miqlas:
No interest anymore, closing for now.

PR #1172, opened 2017-02-25
X11 fixesproto: new recipe

miqlas:


waddlesplash:
This is quite a lot of recipes for protos (I mean the other PRs too). How do Linux distros do this? Can we combine them all into one recipe?

miqlas:
They have a separate recipe for every proto package.

waddlesplash:
:/

Must we do the same? Or is there some technical reason why this is the only way to go?

miqlas:
Have no idea, i’m just copying this: https://packages.gentoo.org/categories/x11-proto <https://packages.gentoo.org/categories/x11-proto>

waddlesplash:
Alright, well, seeing as they're all similar, perhaps it makes sense to at least put them all in one PR?

PR #1173, opened 2017-02-25
simutrans and related packages updated to new version 120.2

taosxx:
Any ideas how to use "secure" source files? Is there a way to get tarballs from sourceforge for a certain revision? As far as I can tell, snapshots must be requested before they are added for a limited time (probably not more than a day).

### DIFF
--- a/games-simulation/simutrans/simutrans-120.2.2.recipe
+++ b/games-simulation/simutrans/simutrans-120.2.2.recipe
@@ -1,0 +1,87 @@
+SUMMARY="A transport simulation game"
+DESCRIPTION="Simutrans is a freeware and open-source transportation \
+simulator. Your goal is to establish a successful transport company. \
+Transport passengers, mail and goods by rail, road, ship, and even air. \
+Interconnect districts, cities, public buildings, industries and \
+tourist attractions by building a transport network you always dreamed \
+of."
+HOMEPAGE="http://www.simutrans.com"
+COPYRIGHT="1997-2004 Hj. Malthaner
+	2005-2017 The Simutrans Team"
+LICENSE="Artistic"
+REVISION="1"
+SvnRevision="8162"
+GitRevision="fb14bfcef06f3c15e0cc5d8368a0bdfe411de65d"
+SOURCE_URI="https://github.com/aburch/simutrans/archive/$GitRevision.tar.gz"
+CHECKSUM_SHA256="7a9c6c2318c0098c84d58518248761d57f34bdbaabbaae5f66d16c3aed1a2227"
+SOURCE_FILENAME="simutrans-$portVersion.tar.gz"
+SOURCE_DIR="simutrans-$GitRevision"
+SOURCE_URI_2="https://downloads.sourceforge.net/project/simutrans/simutrans/120-2-2/simutrans-src-120-2-2.zip"
+CHECKSUM_SHA256_2="7f22c144377abf1a0ee49432f3f6f5eb9a620567960df31d2ea6399b17cf26fa"
+SOURCE_FILENAME_2="simutrans-sf-$portVersion.zip"
+SOURCE_DIR_2=""
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	simutrans$secondaryArchSuffix = $portVersion
+	app:simutrans$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	simutrans_pakset$secondaryArchSuffix >= 120.2
+	timgmsoundfont
+	lib:libbz2$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
+	lib:libsdl$secondaryArchSuffix
+	lib:libsdl_mixer$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libbz2$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
+	devel:libsdl$secondaryArchSuffix
+	devel:libsdl_mixer$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sdl_config$secondaryArchSuffix
+	cmd:strip$secondaryArchSuffix
+	"
+
+GLOBAL_WRITABLE_FILES="
+	non-packaged/$relativeAppsDir/simutrans/config directory keep-old
+	"
+
+BUILD()
+{
+	autoreconf configure.ac
+	export CFLAGS+=-DREVISION=$SvnRevision
+	runConfigure ./configure
+	make $jobArgs
+	strip sim
+}
+
+INSTALL()
+{
+	cp $sourceDir2/simutrans/text/*.tab simutrans/text/
+	cp sim simutrans/simutrans
+	mkdir -p $appsDir/simutrans
+	mkdir -p $prefix/non-packaged/$relativeAppsDir/simutrans/config
+	cp -r  simutrans $appsDir/
+	mv  $appsDir/simutrans/config \
+		$prefix/non-packaged/$relativeAppsDir/simutrans/
+	ln -s $prefix/non-packaged/$relativeAppsDir/simutrans/config \
+		$appsDir/simutrans/config
+
+	addAppDeskbarSymlink $appsDir/simutrans/simutrans "Simutrans"
+}
--- a/games-simulation/simutrans/simutrans-120.3~nightly.recipe
+++ b/games-simulation/simutrans/simutrans-120.3~nightly.recipe
@@ -10,13 +10,12 @@ bugs that will destroy your game, or not even run. They are for bug testing \
 only."
 HOMEPAGE="http://www.simutrans.com"
 COPYRIGHT="1997-2004 Hj. Malthaner
-	2005-2015 The Simutrans Team"
+	2005-2017 The Simutrans Team"
 LICENSE="Artistic"
-REVISION="3"
-SOURCE_URI="svn://tron.yamagi.org/simutrans/simutrans/trunk \
-	--username=anon --password="
+REVISION="1"
+SOURCE_URI="svn://servers.simutrans.org/simutrans/trunk"
 SOURCE_FILENAME="simutrans-$portVersion"
-SOURCE_DIR="trunk"
+SOURCE_DIR=""
 
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
@@ -27,10 +26,11 @@ GLOBAL_WRITABLE_FILES="
 
 PROVIDES="
 	simutrans$secondaryArchSuffix = $portVersion
+	app:simutrans$secondaryArchSuffix = $portVersion
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	simutrans_pakset$secondaryArchSuffix >= 120.0.1
+	simutrans_pakset$secondaryArchSuffix >= 120.2
 	timgmsoundfont
 	lib:libbz2$secondaryArchSuffix
 	lib:libpng$secondaryArchSuffix
@@ -50,8 +50,8 @@ BUILD_PREREQUIRES="
 	cmd:aclocal
 	cmd:autoreconf
 	cmd:curl
+	cmd:g++$secondaryArchSuffix
 	cmd:gcc$secondaryArchSuffix
-	cmd:git
 	cmd:ld$secondaryArchSuffix
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
@@ -71,7 +71,6 @@ BUILD()
 
 INSTALL()
 {
-	fixPkgconfig
 	cp sim simutrans/simutrans
 	mkdir -p $appsDir/simutrans
 	mkdir -p $prefix/non-packaged/$relativeAppsDir/simutrans/config
--- a/games-simulation/simutrans_pak128/simutrans_pak128-2.7.recipe
+++ b/games-simulation/simutrans_pak128/simutrans_pak128-2.7.recipe
@@ -1,0 +1,41 @@
+SUMMARY="Pakset for Simutrans"
+DESCRIPTION="A 'pakset' is the data set that Simutrans needs to run: \
+it contains all the data for the graphics of individual items (vehicles, \
+roads, railways, buildings, etc.) and all of the balancing information \
+(such as prices, speeds, capacities and so forth) necessary to make \
+the game complete.
+When Simutrans could only support 64px size graphics, pak128 already \
+started. First pak to feature a complex economy and have a very wide \
+variety of objects. It contains roughly 7 time more graphics data than \
+pak64 and thus requires by far the largest amount of RAM and \
+processing power of all Simutrans sets."
+HOMEPAGE="http://www.simutrans.com"
+COPYRIGHT="2004-2007 Tomas Kubes
+	2008-2017 The pak128-Team"
+LICENSE="Artistic"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/project/simutrans/pak128/pak128%20for%20ST%20120.2.2%20%282.7%2C%20minor%20changes%29/pak128.zip"
+CHECKSUM_SHA256="8fb29a07290a24495e7b2419de8659c28a2962f094b1c806f472c3b09d37cff4"
+SOURCE_FILENAME="simupak_128-$portVersion.zip"
+SOURCE_DIR="simutrans"
+ARCHITECTURES="any"
+
+PROVIDES="
+	simutrans_pak128 = $portVersion
+	simutrans_pakset = 120.2 compat >= 120.2
+	"
+
+GLOBAL_WRITABLE_FILES="
+	non-packaged/$relativeAppsDir/simutrans/pak128/config directory keep-old
+	"
+
+INSTALL()
+{
+	mkdir -p $appsDir/simutrans
+	mkdir -p $prefix/non-packaged/$relativeAppsDir/simutrans/pak128/config
+	cp -r pak128 $appsDir/simutrans/
+	mv $appsDir/simutrans/pak128/config \
+		$prefix/non-packaged/$relativeAppsDir/simutrans/pak128/
+	ln -s $prefix/non-packaged/$relativeAppsDir/simutrans/pak128/config \
+		$appsDir/simutrans/pak128/config
+}
--- a/games-simulation/simutrans_pak128/simutrans_pak128-2.7~nightly.recipe
+++ b/games-simulation/simutrans_pak128/simutrans_pak128-2.7~nightly.recipe
@@ -15,7 +15,7 @@ bugs that will destroy your game, or not even run. They are for bug testing \
 only."
 HOMEPAGE="http://www.simutrans.com"
 COPYRIGHT="2004-2007 Tomas Kubes
-	2008-2015 The pak128-Team"
+	2008-2017 The pak128-Team"
 LICENSE="Artistic"
 REVISION="1"
 SOURCE_URI="svn+https://svn.code.sf.net/p/simutrans/code/pak128/"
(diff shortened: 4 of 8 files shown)


waddlesplash:
Do we have to use Svn trunk revs?

taosxx:
It seems that simutrans doesn't offer the source files for stable versions of the official paksets (e.g. pak64 or pak128) as zip or tar files on sourceforge, but only the compiled packages (and in contrast to the simutrans base package there are not github mirrors that I know of). These probably could still be unpacked and put in a package for "any" architecture (see recipe for 120.0.1). However, as I noticed when I started working on the recipes in 2015, they seem to work for the 32bit version but not necessarily for the 64bit version (you can't run scenarios, for example). I've raised this issue on the simutrans forum at that time, but since simutrans was still aimed at 32bit systems it was not a priority. In addition, compiling the source code under haiku 64 bit solved the problem.

korli:
I see that Arch Linux has a x86_64 version of Simutrans which seems to use the compiled pakset. Shouldn't be an issue for us as well.

taosxx:
I'm using Arch Linux 64 bit (no multilib) as my everyday operating system and know that it uses the pre-compiled paksets - that's why scenarios are not working under Arch Linux (in contrast to Haiku 64bit with simutrans and pak64).

taosxx:
I've updated the pull request to include (in addition to the architecture specific recipes simutrans_pak64-120.2.0 and simutrans_pak128-2.6.0) two recipes (simutrans_pak64-120.2 and simutrans_pak128-2.6) that use the pre-compiled "official" zip files from sourceforge. With the latter, it is not possible to play scenarios on 64 bit architecture.

games-simulation/simutrans/simutrans-120.2.0.recipe
```diff
@@ -0,0 +1,82 @@
...
+	devel:libz$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
+	devel:libbz2$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
+	cmd:svn
+	cmd:curl
```
korli: Git svn and curl shouldn't be required for an offline tarball build.

taosxx: curl: It is included because the actual language files are downloaded via a script (get_lang_files.sh) during the build process.
svn: Without svn the build process used to stop when it tried to determine the svn revision (of course, with haikuporter it doesn't really get the correct number but then it includes a "0" - it's no problem als long as it can use "svnversion" in the first place). It might be better to use ("cmd:svnversion) in this case?

korli: All downloads should already be done before building. We can't be sure it doesn't download a new file, which would result in a different hpkg.

taosxx: Okay, then let's see if they now include at least the text files for english...

taosxx: So, there are absolutely no language files included if you don't use the get_lang_files.sh script. That means you can't run simutrans. It's - of course - possible to download the translations separately and include them as a separate SOURCE_URI, however, you still have the same problem. There's no revision control for the translation files and you get whatever is the latest version. You'll probably notice if there has been an update if the checksum doesn't match anymore.
The only workaround I see for this problem is going back to the sourceforge zip file for the source code. It's the only place where I see the language files included.

taosxx:
Update1: I've removed the architecture dependant pak64 and pak128 recipes (I got scenarios to work with the pre-compiled "official" zip files from sourceforge).

taosxx:
Update2: I've added a recipe for simutrans-120.2.1. This one gets rid of cmd:svn (it seems they changed a missing svn from error to warning), cmd:curl, and cmd:git and the get_lang_files.sh script but downloads the source files from sourceforge and therefore uses a different configure procedure. I'm not sure if this difference is to blame for the fact that the resulting simutrans binary can't run scenarios (same as the Arch Linux version for 64bit).

games-simulation/simutrans/simutrans-nightly.recipe
```diff
@@ -9,18 +9,17 @@ changes which make your savegames unreadable by any official version, or \
...
-	2005-2015 The Simutrans Team"
+	2005-2017 The Simutrans Team"
 LICENSE="Artistic"
-REVISION="2"
+REVISION="3"
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 PROVIDES="
-	simutrans$secondaryArchSuffix = $portVersion
+	simutrans$secondaryArchSuffix = $portVersion compat >= 120.2
```
korli: Why add compat here?

taosxx: I can leave it out but this is what I was thinking when I added it:
Not all pakset revisions are compatible with all revisions of simutrans. In this case, a nightly of simutrans (with the version "nightly") that was packaged with this recipe would be slightly more recent but still sufficiently comparable to version 120.2.x of simutrans and therefore compatible with the official paksets for 120.2. As soon as there is a compatibility breakage, the recipe would get bumped to a new revision with a new "compat" entry. The "-nightly" recipes are only intended to be built with haikuporter on one's own for those who want to live on the bleeding edge whereas the versioned recipes could be used to provide stable packages (unfortunately, haikuporter sees "nightly" as a newer revision and would build it if you don't specify the version).
At the moment, the paksets don't check the version of the simutrans package due to the hen/egg problem if you can't use HaikuDepot to install hpkgs (simutrans is only available in haikuports to be built with haikuporter): as soon as you move a pakset hpkg to the packages folder (or double-click on it to use HaikuDepot) you would get a complaint that there is nothing that provides a matching simutrans package so it wouldn't be activated, and if you would then try to move the simutrans package you would see the error message that tells you that there is no matching pakset available (because it wasn't activated before) - you would have to restart Haiku to solve this. This does not happen for packages in the repositories because one would be installed as dependency of the other.

pulkomandy: > (unfortunately, haikuporter sees "nightly" as a newer revision and would build it if you don't specify the version).

You can use 1.2.1~nightly for nightly builds before the 1.2.1 release. Using just "nightly" as a version should be avoided because there is no way to compare it to anything else.

I'm not sure the packsets need to depend on the game. You may want to install the packset for other reasons (extracting the data for display outside the game, using it with another game that would be compatible, etc). So, a strict dependency seems a bit too hard of a constraint. But we can express "softer" dependencies in our packages, and this may be a good place to use that.

taosxx: Okay, just for clarification, haikuporter would "count" like this:
1.2.1~nightly < 1.2.1 ?
1.2 > 1.2.0 ? (I had to add compat >= 120.2 to pakset 120.2.0 to get it past the check of "simutrans_pakset$secondaryArchSuffix >= 120.2".
BTW, now that the paksets (apart from the nightly ones) are back to "any" architecture, is it still okay to use "$secondaryArchSuffix" here? On 64bit you can still combine nightly simutrans package with release pakset or the other way round, but what happens on gcc2hybrid?

pulkomandy: The ~ is used for pre-releases (alpha, beta, etc) and is "before" the matching release. So 1.2.1~something < 1.2.1. In other cases, minor releases and the like, are "after" the matching release. So, 1.2.1 > 1.2.
It seems that 1.2 == 1.2.0.

See "version strings" in this document for some extra info:
https://dev.haiku-os.org/wiki/PackageManagement/BuildingPackages

taosxx: Thank you for the link. However, I still have some questions:
1.) "If no compatibility version is specified, the resolvable can only satisfy dependency constraints with an exactly matching version."
Does that mean that if I have a package with version 1.0.0 (and there is no compatibility declared in PROVIDES for "compat >= 1.0") it can't fullfill a dependency that is declared in REQUIRES as ">= 1.0"?
2) Okay, 1.2.3 > 1.2.3~nightly, but 1.2.3~nightly > 1.2.2 which means that haikuporter would still build the nightly instead of the stable 1.2.2 if you don't specify the version to build?

games-simulation/simutrans/simutrans-120.2.0.recipe
taosxx: I've uploaded simutrans-120.2.1~test.recipe. It downloads the source code twice, once as tarball from the git mirror (SOURCE_URI) and once as "official" zip (incl. the language files) from sourceforge (SOURCE_URI_2). Since I prefer the configure mechanism of the git mirror (you get at least a binary that can actually run scenarios), the sourceforge zip is only used for supplementing the language files so that svn, curl, git and get_lang_files.sh can be omitted.
If you prefer this solution I'll rename it to simutrans-120.2.1.recipe and will remove the other recipes.

korli:
Gentoo uses this zip file to complete the release tarball: http://simutrans-germany.com/translator/data/tab/language_pack-Base+texts.zip

taosxx:
That's the one the get_lang_files.sh downloads. As soon as there's an update to the language files the zip get's updated (remember, no version control) and you have the same problem as before. In simutrans-120.2.1~test, I extract the language files from the source zip from sourceforge.

games-simulation/simutrans/simutrans-nightly.recipe
taosxx: I've removed the compat entry and renamed the nightly recipes.

games-simulation/simutrans/simutrans-120.2.1.recipe
```diff
@@ -0,0 +1,76 @@
...
+SOURCE_DIR=""
+COPYRIGHT="1997-2004 Hj. Malthaner
+	2005-2017 The Simutrans Team"
+LICENSE="Artistic"
+REVISION="1"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+PROVIDES="
+	simutrans$secondaryArchSuffix = $portVersion
+	"
```
korli: a "Provides" is obviously missing for the binary.

taosxx: I thought we only do this for command line programs (e.g. diff, grep) that might get called from another program? I think I borrowed the PROVIDES section from openttd:

> PROVIDES="
>
>	openttd$secondaryArchSuffix = $portVersion
>"

pulkomandy: For everything that is installed to $binDir.
For things that are installed in $appsDir there should be an app:something provide as well, however this one currently isn't enforced by the policy checks (haikuporter --strict).

taosxx: Okay, I had a look at a few other recipes that use "app:something" but I'm not sure if I should add "app:simutrans" (the file name of the actual binary is lower case) or "app:Simutrans" (as in the addAppDeskbarSymlink entry). I've also seen for "something$secondaryArchSuffix" packages entries with ("app:something$secondaryArchSuffix") and without ("app:something") $secondaryArchSuffix.

taosxx:
Let me know if you have any preferences regarding the way the necessary language files should be included in the simutrans package. We have several examples:

1. simutrans-120.2.0.recipe:
- downloads a tarball from github without language files
- a script (get_lang_files.sh) downloads the latest language files (no specific version) and adds them to the work directory (needs curl or wget as dependencies)
- configuration is done by running 'autoreconf configure.ac' then 'runConfigure ./configure'
- resulting binary/package has support for network games and scenarios

2. simutrans-120.2.1.recipe:
- downloads the source zip from sourceforge, language files (specific version) are included (curl/wget not needed)
- 'chmod +x configure.sh' + 'runConfigure ./configure.sh'
- resulting binary/package does support network games but not scenarios

3. simutrans-120.2.1~test.recipe:
- downloads a tarball from github without language files (source-1) and the source zip from sourceforge including the language files (source-2)
- configuration is done in source-1 by running 'autoreconf configure.ac' then 'runConfigure ./configure'
- language files are copied from source-2 to source-1 in INSTALL() section
- resulting binary/package has support for network games and scenarios

taosxx:
@korli, @pulkomandy:
If none of the above options work for you, I could try a fourth one: having the necessary language files in a separate package (either using the latest "nightly version" from [http://simutrans-germany.com/translator](http://simutrans-germany.com/translator/data/tab/language_pack-Base+texts.zip) or extract them from the sourceforge zip.
Let me know what you think. I would like to update the pull request before the end of March since I probably won't have too much time in late March and April.

pulkomandy:
Why are the sourceforge zip and git configure scripts any different?
Anyway, as korli mentionned, 1) is not acceptable because all downloads must happen before the build starts and be checksummed. 2, 3 or 4 looks fine.

taosxx:
> Why are the sourceforge zip and git configure scripts any different?

I have not the faintest idea. Never found anything explaining this in build instructions, the simutrans forum, or by searching through the github mirror. There probably is an obvious explanation but I'm not experienced enough to see it (I compared both sources to find out why one would allow scenarios to run but not the other and couldn't find a reason for it, only noticed that the sequence of some lines for checking which libraries are available seems different - `diff` was not too happy that there seems to be different encoding).

> 2, 3 or 4 looks fine.

I'll probably go with 3 then since this allows to run scenarios.

taosxx:
I've updated the pull request over the weekend. Let me know if I should address some further issues.

games-simulation/simutrans/simutrans-120.2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+	autoreconf configure.ac
+	export CFLAGS+=-DREVISION=$SvnRevision
+	runConfigure ./configure
+	make $jobArgs
+	strip sim
+}
+
+INSTALL()
+{
+	fixPkgconfig
```
pulkomandy: Is this needed? It should be only if the package is poviding libraries, and I think this one is not doing that.

games-simulation/simutrans/simutrans-120.2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+	cp $sourceDir2/simutrans/text/*.tab simutrans/text/
+	cp sim simutrans/simutrans
+	mkdir -p $appsDir/simutrans
+	mkdir -p $prefix/non-packaged/$relativeAppsDir/simutrans/config
```
pulkomandy: Not sure what you are doing here, it is not possible to put things in non-packaged from a package so an rm would do the same?
Also, what about using $settingsDir instead? Is there a reason to not put things there?

games-simulation/simutrans_pak128/simutrans_pak128-2.6.recipe
```diff
@@ -0,0 +1,41 @@
...
+REVISION="1"
+ARCHITECTURES="any"
+
+PROVIDES="
+	simutrans_pak128 = $portVersion
+	simutrans_pakset = 120.2 compat >= 120.2
+	"
+
+GLOBAL_WRITABLE_FILES="
+	non-packaged/$relativeAppsDir/simutrans/pak128/config directory keep-old
```
pulkomandy: This path looks strange. Things in a package don't go in non-packaged.

games-simulation/simutrans_pak128/simutrans_pak128-2.6.recipe
```diff
@@ -0,0 +1,41 @@
...
+	"
+
+INSTALL()
+{
+	mkdir -p $appsDir/simutrans
+	mkdir -p $prefix/non-packaged/$relativeAppsDir/simutrans/pak128/config
+	cp -r pak128 $appsDir/simutrans/
+	mv $appsDir/simutrans/pak128/config \
+		$prefix/non-packaged/$relativeAppsDir/simutrans/pak128/
```
pulkomandy: Moving things to non-packaged essentially leaves them out of the package. Are you sure this works? Or does it just happen to work because you have simutrans datafiles copied manually to your non-packaged directory?

The package should put things either in $appsDir/simutrans somewhere, or use $dataDir/simutrans/.

taosxx: See above.

taosxx: I think github just ate my three replies...

games-simulation/simutrans/simutrans-120.2.1.recipe
taosxx: Yes, you're right! I had it removed before the first pull request. I must have overlooked it. After I had to run checkfs several times I had to reload the recipes from a backup.

games-simulation/simutrans/simutrans-120.2.1.recipe
taosxx: Actually yes, it is possible (I checked it several times for the previous recipes with a freshly installed system - took me some time to find a working solution). The config files must be writable - simutrans can't properly run if they're not. In addition, there must be an initial config file - the one from the package - otherwise simutrans won't start properly. The package includes a symlink in $appsDir/simutrans/config to the writable folder in non-packaged. When installing a simutrans_pak package for the first time a configuration file from the package is copied to non-packaged in the corresponding directory (I think I got the idea from the python package). When you run simutrans, settings like preferred language are saved to the file (if you can't write them simutrans shows an error message or might even crash). If you update the simutrans/simutrans_pak package the initial config file is not copied if there already is a config file (rule "keep-old" for the GLOBAL_WRITABLE_FILES).
The reason for not using $settingsDir was probably that - at the time of the very first simutrans recipe - more people on IRC were in favour of non-packaged or no one cared enough for one or the other (I didn't have any preferences of my own). Traditionally, even in Linux, simutrans generates a "Simutrans" folder rather than a ".simutrans" folder in the home directory that includes some config files, but also savegames or screenshots that a user might search for in order to share or upload them. The config files from the paks can be found in the apps directory. I thought it was a good idea to keep both the location of the paks config files and the Simutrans folder because people could just follow the "official" simutrans manual.
If you insist on a simutrans folder in the settings directory, is there a way to inform users during or after the installation that screenshots etc. can be found there instead of the Simutrans folder in the home directory as on the other platforms? I'll have to add a patchfile so that simutrans won't generate the home/Simutrans directory when saving games, etc - that will take some time to find in the simutrans source code.
If you don't mind the home/Simutrans folder I'll just change the symlink for the pak's config in the apps directory from non-packaged to the settings directory ($settingsDir/simutrans/config instead of $prefix/non-packaged/$relativeAppsDir/simutrans/config). Then the simutrans manual would still work for haiku - as long as I don't have to remove the symlink.

games-simulation/simutrans_pak128/simutrans_pak128-2.6.recipe
taosxx: Where would you put the GLOBAL_WRITABLE_FILES if not in a writable directory (I don't mind if it is $settingsDir instead)? Or do I misunderstand you?

taosxx: @pulkomandy 
I've updated the pull request to remove the fixPkgconfig entry.

However, it seems not trivial to go from `/non-packaged/$relativeAppsDir` to `$relativeSettingsDir`. While this still works for packages installed in the system hierarchy (config directory is copied to `/system/settings/simutrans/config` and `/apps/simutrans/config` links to it), I can't get it to work in the /home/config hierarchy because `/home/config/settings/simutrans/config` is not created at all, so the link from `/home/config/apps/simutrans/config` is broken. Using non-packaged for the config directory on the other hand works for both installation in system and in home.

scottmc:
What's the current status on this?

taosxx:
From my side:
1. I was having difficulties with packages that used "~nightly" as part of their version. For some reason, I could not get them to activate on my haiku installation. If I (only) changed the name to what I had before in the recipe I could activate them again. I was going to investigate this. And I completely wiped my haikuports directory as I got frequent bfs errors whenever I used haikuporter to try to build packages to make sure that the packages I used for testing didn't contain corrupt files that prevented them from getting activated.
2. I couldn't find a way around the problem when going from /non-packaged/$relativeAppsDir to $relativeSettingsDir for the settings directory: works fine if the package is installed in the /system/packages but not in the /home/config/packages hierarchy.
3. I'm affected by haiku bug #13769 [https://dev.haiku-os.org/ticket/13769](url), so anything git related does not work really well for me at the moment. Since I have to use haikuporter under haiku to build packages (or is there another way I'm not aware of?) and bfs is read-only under linux that's not a good situation.
4. I'm stuck with a max. 32 kBit/s download/upload mobile connection for at least another week, so downloading the simutrans and simupak source codes each time to make sure that the checksum hasn't changed in the mean time isn't going to happen due to timeouts.

scottmc:
Perhaps we mark the WIP recipe as broken on all architectures, and merge it.  Then open a new issue (or update an existing one?) with the status.  That way the code doesn't get lost and someone else can hack on it if they get the itch before you get back to it?

taosxx:
I don't mind you merging the WIP recipes (although it might not be necessary to mark them all as broken). Apart from the ~nightly version (that one doesn't work for me but no else complained about problems with "~" in package names so I'm not too sure), the other recipes from the pull request should work fine (they just still use the working non-packaged approach for configuration files - the same as the recipes that are already in haikuports - I didn't upload the not-completely-working switch-to-$relativeSettingsDir WIP recipe in spring).

scottmc:
This is now showing conflicts, Does that current simutrans recipe(s) take care of the issues on this pull request?  Can this one now be closed?  Do any new issues need to be opened to address any concerns raised above?

taosxx:
Updated to solve merge conflicts. No additional updates.

Begasus:
@taosxx is makeobj-60.0.recipe still being used? If not it could be removed (already superseded by 60.1 and 60.2)?

Begasus:
Related to https://github.com/haikuports/haikuports/issues/3779

taosxx:
makeobj-60.0 would be needed to compile a pakset for older simutrans versions. Since no one is going to do this (probably), let's just remove it. Apart from the nightlies all other pakset recipes use a pre-compiled zip file.

Begasus:
Created a PR at https://github.com/haikuports/haikuports/pull/3885 for this :)

PR #1197, opened 2017-02-28
Add "make bindcatalogs" to recipe for localized BePDF.

tsundeoku:
BePDF was missing a "make bindcatalogs" in its recipe, causing it to build with only the English locale.


diversys:
You need to bump revision

PR #1203, opened 2017-03-04
libsvg, new recipe

Begasus:


korli:
Do you need this library? It hasn't seen an update since 12 years, isn't packaged by Linux distributions. Maybe it's just obsolete.

Begasus:
Doing some checkups (it's required for libsvg-cairo), so no real hurry for it

Begasus:
libsvg-cairo is build also, both are needed(?) for svg support in tuxpaint, it builds with support with both libs (not sure how to test it though) ... no errors so far

scottmc:
Are you now able to draw vector images with tuxpaint?  That might be a good way to test.

korli:
Tuxpaint seems to use librsvg.

Begasus:
@scottmc not yet, but not sure how to do also, dismissed the search for svg, have been looking into sdl_pango which seem to be working also, but no idea there on how to check that one :)
@korli librsvg isn't an option at this moment (yet) :)

Begasus:
You can close this PR, libsvg-cairo is build also, haven't seen real changes in TP so far, if need be I still will leave both recipes on my own fork if someone is interested in testing them (as you said earlier, obsolete lib) :)

PR #1206, opened 2017-03-09
Fix vim 8 build on x86 gcc-5

freddietilley:
This patch fixes a lvalue required as left operand assignment error when trying to build vim with gcc-5.4.0


waddlesplash:
Looks OK I suppose.

pulkomandy:
Could use strButtons.SetByteAt(end, '\0'); instead, to avoid searching for the \n when its offset is already known?

freddietilley:
Ah, I was using the official class reference and not checking the String.h header file so I was not aware of the methods existence :-)

pulkomandy:
They were added when the [] operator was made read-only. I forgot to document them.

PR #1207, opened 2017-03-09
Fix vim 8 build on x86 gcc-5

freddietilley:
This patch fixes a lvalue required as left operand assignment error when trying to build vim with gcc-5.4.0. Now with the better SetByteAt method


korli:
Please bump the recipe revision (the commit changes the resulting package).

PR #1210, opened 2017-03-11
Fix vim 8 build on x86 gcc-5

freddietilley:
This patch fixes a lvalue required as left operand assignment error when trying to build vim with gcc-5.4.0.


miqlas:
Thanks!
Maybe you can try to update the vim recipe to the latest version, and/or fix the vim GUI: some Features doesn't working, @Pulkomandy can Tell which Features would be nice to have.

miqlas:
I just updated the vim recipe to the latest version with your patch: https://github.com/haikuports/haikuports/pull/1219
It makes this PR outdated, i think somebody should close this.

korli:
@freddietilley Thanks!

PR #1214, opened 2017-03-13
libsodium: bump version and remove secondary_arch

return:
libsodium is able to build on all architectures and patches are now up-streamed due to [this commit](https://github.com/jedisct1/libsodium/commit/9cae7b6b7c6f527df807157602ba48821281048b).


korli:
Any reason to remove secondary architectures? This is actually required for most libraries.

return:
@korli If a recipe 'architectures' field supports all architectures, then the $secondary_arch suffix is not need and can be removed. Or is that not the case with libsodium?

korli:
This is a wrong assumption: libraries can be needed for a secondary​ architecture.

pulkomandy:
"x86" in ARCHITECTURES: the library can be built with gcc5 for a gcc5 system.
"x86" in SECONDARY_ARCHITECTURES: the library can be built with gcc5, for a gcc2hybrid system

The build in both cases is mostly the same, but there are some changed paths and different package metadata, both needing special care. Hence the support for building as secondary arch package needs to be explicit.

return:
OK, my mistake for following those assumptions on removing the architecture suffixes. Is there a need to revert these changes  or open another PR to re-add the suffixes?

korli:
Please add a PR reintroducing the secondary suffixes.

return:
@korli OK, I have opened a PR addressing this at #1217

PR #1220, opened 2017-03-18
Qemu: bump version

miqlas:
Builds fine, but fails to package:

> Warning: POLICY ERROR: can't find used library "libssp.so.0"

It comes from the gcc package, but as this lib doesn't installed into the standard LD_PATH, i have no idea, what should i do now. Any help is really appreciated.

Edit: disabling stack protector solved the problem.


korli:
You should probably leave the recipes for mmlr repo alone.

waddlesplash:
Indeed; looks like this loses a bunch of code mmlr had.

korli:
@miqlas please don't remove qemu-2.1.2.recipe

PR #1221, opened 2017-03-18
ACPICA: new recipe

miqlas:


miqlas:
Yep. Should it be acpica_tools then?

PR #1223, opened 2017-03-19
erlang: bump version and enable ssl

return:
Tested only on x86. Erlang OTP 18 and higher with openssl enabled is needed to compile or run [Elixir](http://elixir-lang.org), a functional programming language built on top of the Erlang VM.

### DIFF
--- a/dev-lang/erlang/erlang-19.02.recipe
+++ b/dev-lang/erlang/erlang-19.02.recipe
@@ -1,0 +1,72 @@
+SUMMARY="The programming language Erlang/OTP"
+DESCRIPTION="Erlang is a programming language used to build massively \
+scalable soft real-time systems with requirements on high availability. \
+Some of its uses are in telecoms, banking, e-commerce, computer telephony \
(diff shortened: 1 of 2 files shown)


miqlas:
The official site have no 19.02, but 19.2, You are also defined 19.2 as source url, then why is it 19.02 in recipe name?

Please, add x86_64 to the arch list.

19.3 released just some days ago, can you maybe try to update your recipe to the latest, greatest erlang?

Thank you!

miqlas:
Just updated erlang to tle latest recipe, cleaned the recipe a bit and sent a PR, maybe you can look into it:

https://github.com/haikuports/haikuports/pull/1225

Tested only on x86_64, builds, packages, runs helloworld just fine.
It makes this PR outdated, so we can close this.

Btw the plenty "remove -lm" in patchset is unnecessary.

return:
@miqlas I've tested your recipe on x86 and it builds just fine. x86_gcc2 still needs to be tested but i'll close this PR as it is outdated.

PR #1225, opened 2017-03-19
Erlang: bump version, recipe cleanup

miqlas:
It makes https://github.com/haikuports/haikuports/pull/1223 outdated.


dev-lang/erlang/erlang-19.3.recipe
```diff
@@ -0,0 +1,118 @@
...
+	cmd:run_erl = $portVersion
+	cmd:to_erl = $portVersion
+	cmd:typer = $portVersion
+	"
+REQUIRES="
+	haiku
+	cmd:libtool
+	cmd:gcc
+	cmd:perl
+	cmd:wx_config
```
korli: Wx_config, really?

miqlas:
why not?

korli:
Well, Erlang didn't require wx_config until this change. You should definitely explain things, not hide every change for us to find out.

miqlas:
Wings3D requires it.

return:
@miqlas Erlang 19.3 doesn't seem compile on x86_gcc2. You should mark x86_gcc2 as broken until there is a patch that fixes this.

miqlas:
return, thanks, disabled.

korli:
REQUIRES should probably have some runtime dependency to wxqt, not related to wxqt_devel.

PR #1238, opened 2017-03-22
llvm: use official llvm 4.0.0 sources

return:
LLVM 4.0.0 released about a week ago and the official sources are now available. Built successfully and tested on x86_64.


korli:
1/ Could you mark x86_64 as tested? 
2/ Could you check that Mesa 17.x builds and works correctly with it?

return:
@korli 
1. Marked x86_64 as tested.
2. I was able to build Mesa and mesa_swpipe and it works fine.

I patched llvm-config to search for the libraries in /develop/lib rather than /lib as Mesa was using it to find the LLVM libraries.

korli:
Cool, thanks!

PR #1241, opened 2017-03-23
hgrep, fix recipe and pull a more recent version

Begasus:


korli:
LGTM

korli:
maybe remove the old recipes for hgrep?

Begasus:
I was thinking of that too, but there is still a zip in it and a hgrep file that can be used? (havent tested that one), but if it's ok I can remove them ...

waddlesplash:
Yes, please remove them.

Begasus:
Also the zip and related parts?

scottmc:
Looks like this is confusing two different hgreps.  The original one on
haikuports, was a script to search Haiku's headers, this new recipe is for
some different hgrep?

Also see lgrep, it was added at the same time as hgrep.  Both have a zip
file which contains a single script file.  I think that was done to fake
out haikuporter.  Not sure that anyone even uses those old hgrep and lgrep
scripts.

-scott

Begasus:
If not, this one could/should be deleted from the haikuports tree?

PR #1243, opened 2017-03-23
zoo, remove obselte bep files

Begasus:
If for some reason this should still be in here (maybe something in the patchfile could be handy for someone?) it could remain, otherwise, download link is broken, no need for the bep file imho ...


waddlesplash:
Patch looks pretty useless to me.

PR #1247, opened 2017-03-24
slang, fix recipe so it builds again

Begasus:


sys-libs/slang/slang-2.2.4.recipe
```diff
@@ -12,7 +12,7 @@ part of the S-Lang distribution."
 HOMEPAGE="http://www.jedsoft.org/slang/"
 COPYRIGHT="2004-2013 John E. Davis"
 LICENSE="GNU GPL v2"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="ftp://space.mit.edu/pub/davis/slang/v2.2/slang-2.2.4.tar.bz2"
```
pulkomandy: Why are we not getting it from http://www.jedsoft.org/releases/slang/old/slang-2.2.4.tar.gz ?
Also, why not update it to a more current version?

korli:
Let's say it's a hack, the version would need to be bumped.

PR #1256, opened 2017-03-26
Octave: new recipe

miqlas:


miqlas:
> #octave --version
> octave: no graphical display found
> octave: disabling GUI features
> GNU Octave, version 4.2.1
> Copyright (C) 2017 John W. Eaton and others.
> This is free software; see the source code for copying conditions.
> There is ABSOLUTELY NO WARRANTY; not even for MERCHANTABILITY or
> FITNESS FOR A PARTICULAR PURPOSE.
> 
> Octave was configured for "x86_64-unknown-haiku".
> 
> Additional information about Octave is available at http://www.octave.org.
> 
> Please contribute if you find this software useful.
> For more information, visit http://www.octave.org/get-involved.html
> 
> Read http://www.octave.org/bugs.html to learn how to submit bug reports.

miqlas:
Tested only on x86_64, other arches maybe requires more work, but i cannot test them. Patches welcome.
It builds and runs just fine.

sci-mathematics/octave/octave-4.2.1.recipe
```diff
@@ -0,0 +1,151 @@
...
+INSTALL()
+{
+	make install-strip
+
+	# Remove libtool .la files
+	# Move libs to $libDir
+	for f in liboctave liboctinterp; do
+		rm -rf $libDir/octave/$portVersion/$f.la
+		mv $libDir/octave/$portVersion/$f.so* $libDir
+	done
```
korli: Libs shouldn't be moved to $libDir when upstream decided otherwise.

miqlas:
How can i solve then the missing required libs error in haikuporter?

korli:
You could say it's a haikuporter bug. For most apps there is RPATH set on the binary, which haikuporter checks to find out an alternative loading path used by the runtime loader. Is it set for octave? How does octave find its libraries?
There is a similar problem for Qupzilla, or Opentoonz (this is one set LIBRARY_PATH with a script before executing its binary!!!)

miqlas:
@korli, i have no idea. As i told, it works on x86_64. Please, merge it, because i'm really overwhelmed right now, so i can't do much.  I think i should say farewell ASAP.

korli:
No problem. Thanks!

korli:
I added a TODO, it's really an issue if one wants to install different versions of Octave.
Thanks! And please don't say farewell yet :)

PR #1260, opened 2017-03-29
unixODBC, new recipe

Begasus:


dev-db/unixODBC/unixodbc-2.3.4.recipe
```diff
@@ -0,0 +1,95 @@
...
+        settings/odbc.ini keep-old
+        settings/odbcinst.ini keep-old
+	"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	unixodbc${secondaryArchSuffix} = $portVersion
+	cmd:odbc_config${secondaryArchSuffix}
```
korli: Question : is odbc_config meant to give libraries and headers path information? If yes, it belongs to the devel package.

Begasus: I tried to add it to the devel package but ran into problems with it then ... let me check

Begasus: Usage: odbc_config
                [--prefix]
                [--exec-prefix]
                [--include-prefix]
                [--lib-prefix]
                [--bin-prefix]
                [--version]
                [--libs]
                [--static-libs]
                [--libtool-libs]
                [--cflags]
                [--odbcversion]
                [--longodbcversion]
                [--odbcini]
                [--odbcinstini]
                [--header]
                [--ulen]

korli: Then add $binDir/odbc_config after $developDir. It should be ok.

Begasus: Warning: POLICY ERROR: "/packaging/unixodbc_devel/bin/odbc_config" needs library "libroot.so", but the package doesn't seem to declare that as a requirement

korli: You need to add haiku as a requirement of the devel package.

Begasus:
This should do it ... (build for x86_gcc2 just now)

PR #1262, opened 2017-03-29
gtk-doc, bump version

Begasus:
WIP ... the *.pc is still not in the correct place ...

### DIFF
--- a/dev-util/gtk_doc/gtk_doc-1.25.recipe
+++ b/dev-util/gtk_doc/gtk_doc-1.25.recipe
@@ -1,9 +1,8 @@
 SUMMARY="Generate documentation from C sources"
-DESCRIPTION="
-GTK-Doc is a project which was started to generate API documentation from \
-comments added to C code. It is typically used to document the public API of \
-GTK+ and GNOME libraries, but it can also be used to document application code.
-"
+DESCRIPTION="GTK-Doc is a project which was started to generate API \
+documentation from comments added to C code. It is typically used to \
+document the public API of GTK+ and GNOME libraries, but it can also \
+be used to document application code."
 HOMEPAGE="http://www.gtk.org/gtk-doc/"
 COPYRIGHT="2002-2010 Damon Chaplin and Owen Taylor"
 LICENSE="GNU GPL v2"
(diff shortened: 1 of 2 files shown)


korli:
I had a look at what arch Linux does, the pc file is in share/pkgconfig, which would translate to $dataDir/pkgconfig on Haiku. I'm not sure that pkg-config picks up this path though. To be tested.

Begasus:
you are right there, noticed it already yesterday (hence wip), can't seem to create a devel package for it so looking into mv it to the correct place ...

pulkomandy:
Moving it should be fine. I'm not in Haiku at the moment but it should be possible to ask pkg-config to print its search dir. However, having this in share makes it difficult to handle hybrid arch properly so it would be nice to avoid it.

korli:
Why is it needed to have secondary arches?

Begasus:
jikes seem to have messed some things up :/

dev-util/gtk_doc/gtk_doc-1.25.recipe
```diff
@@ -0,0 +1,102 @@
...
+	#TODO fix installation path so it installs 
+	#in the correct directory for each arch
+	# Remove stuff we don't need in the secondary architecture base package.
+	if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+	mkdir -p $developDir/lib/x86/pkgconfig/
+	mv $dataDir/pkgconfig/gtk-doc.pc $developDir/lib/x86/pkgconfig/gtk-doc.pc
+	else
+	mkdir -p $developDir/lib/pkgconfig/
+	mv $dataDir/pkgconfig/gtk-doc.pc $developDir/lib/pkgconfig/gtk-doc.pc
+	fi
```
Begasus: @korli this is the place where I moved the *.pc file to it's correct place, it's probably not the right way, but this is how far I came so far

Begasus:
@korli secondary arches are needed because they don't seem to pick up the *.pc file otherwise (tested with rasqal)

korli:
Please add a TODO where the pkgconfig files should be located. pkgconfig needs to be fixed to pick up files  in a architecture independent path aka $dataDir/pkgconfig.

Begasus:
in the recipe just above the section I already added a TODO ...

korli:
Well the TODO you inserted means something else to me. Whatever...

Begasus:
'proper' fix coming up, no need to check for the arch's anymore (just need to do some tests)

Begasus:
updated the recipe and added patch, hope this makes more sense :)

korli:
I don't know, it can always be improved later if needed.

PR #1266, opened 2017-04-01
rasqal, update recipe with new dependencies

Begasus:


korli:
Was it already broken on x86_gcc2?

Begasus:
@korli  yes, with the current one in the tree it doesn't build anymore also
http://sprunge.us/HhdY
checked also by @humdingerb 
still finetuning it ...

Begasus:
added gtkdoc-check for future, the system picks it up and all it's needed, but rasqal still doesn't build with it enabled (it's default disabled btw)
--with-html-dir=$developdocDir was already there so left it in, otherwise the html files will go into the subdir for gtk-doc/html ... not sure what is preferred so left it as it is :)

Begasus:
this one should be ok, the one for redland should also be set as "!x86_gcc2" as it depends on rasqal ...

PR #1267, opened 2017-04-01
Add dmd v2.067.1 recipe

raefaldhia:
This recipe isn't the lastest version of dmd, this is would be useful to get lastest version of dmd because any version above 2.067.1 would require bootstrap to build.

I'm not sure if the patches working correctly, because dmd require druntime to work correctly.

And now i'm working on porting druntime and phobos.
I always want to build app based on dlang in haiku lol, helps are welcome!


dev-lang/dmd/dmd-2.067.1.recipe
```diff
@@ -0,0 +1,71 @@
...
+PROVIDES="
+	dmd$secondaryArchSuffix = $portVersion
+	cmd:dmd$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	dmd_data  == $portVersion
+	"
+
+if [ -z "$secondaryArchSuffix" ]; then
```
korli: You can probably remove this check if you don't intend to have support for secondary architectures.

PR #1268, opened 2017-04-02
elixir: new recipe

return:
Elixir is a functional programming language built on top of the Erlang VM and is used for building scalable applications. 

Elixir fails to build from source due to a segfault in the Erlang VM. Until this is patched, the pre-compiled version was tested instead and so far works perfectly on x86_64.

### DIFF
--- a/dev-lang/elixir/elixir-1.4.2.recipe
+++ b/dev-lang/elixir/elixir-1.4.2.recipe
@@ -1,0 +1,45 @@
+SUMMARY="A dynamic, functional language for building scalable applications"
+DESCRIPTION="Elixir leverages the Erlang VM, known for running \
+low-latency, distributed and fault-tolerant systems, while also \
+being successfully used in web development and the embedded software \
+domain."
+HOMEPAGE="http://elixir-lang.org"
+COPYRIGHT="2012-2017 Plataformatec"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="https://github.com/elixir-lang/elixir/releases/download/v$portVersion/Precompiled.zip"
+SOURCE_DIR=""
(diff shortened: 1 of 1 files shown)


dev-lang/elixir/elixir-1.4.2.recipe
```diff
@@ -0,0 +1,45 @@
+SUMMARY="A dynamic, functional language for building scalable applications"
+DESCRIPTION="Elixir leverages the Erlang VM, known for running \
+low-latency, distributed and fault-tolerant systems, while also \
+being successfully used in web development and the embedded software \
+domain."
+HOMEPAGE="http://elixir-lang.org"
+COPYRIGHT="2012-2017 Plataformatec"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="https://github.com/elixir-lang/elixir/releases/download/v$portVersion/Precompiled.zip"
```
korli: The source file should be renamed with SOURCE_FILENAME.

dev-lang/elixir/elixir-1.4.2.recipe
```diff
@@ -0,0 +1,45 @@
...
+	cmd:iex = $portVersion
+	cmd:mix = $portVersion
+	"
+REQUIRES="
+	haiku
+	cmd:erl >= 18.0
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
```
korli: I think you don't need haiku_devel.

korli:
Such precompiled recipes are usually suffixes with _bin to clearly indicate that nothing is built by the recipe.

miqlas:
Wouldn't it be better to re-port this stuff?

Begasus:
@return any progress on this?

PR #1276, opened 2017-04-15
mesa: add pkg-config entry and missing header(s) for EGL

return:
The KHR/khrplatform.h header is required by libepoxy for running its testsuite. This commit also adds a
pkg-config entry for EGL, also used by libepoxy.


waddlesplash:
LGTM.

diversys:
Please bump revision.

return:
Revised recipe to include bumped revision.

PR #1277, opened 2017-04-16
Godot: new recipe

miqlas:
Builds fine, but freaks out at linking stage. Help required.

### DIFF
--- a/dev-games/godot/godot-2.1.3.recipe
+++ b/dev-games/godot/godot-2.1.3.recipe
@@ -23,6 +23,7 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
+	lib:libcrypto$secondaryArchSuffix
 #	lib:libfreetype$secondaryArchSuffix
 	lib:libGL$secondaryArchSuffix
 	lib:libglew$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


waddlesplash:
Paste of link errors?

miqlas:
@waddlesplash : Does not compute.

waddlesplash:
I mean, can you make a Pastebin or Gist or something of all the error messages you get while linking? You said it "freaks out" at the linking stage.

miqlas:
@waddlesplash : here you go: https://hastebin.com/puzanifozo

miqlas:
It seems hovewer it let to define not to use the shipped in-tree stuffs, it doesn't add the required LDFLAGS for the system libs.
But maybe i'm just wrong.

waddlesplash:
Yep, that looks like what's going on. Seems to need at least `-lpng -lglew -lssl -lvorbis`.

waddlesplash:
Works now?

miqlas:
Yeah :)
xsi

waddlesplash:
Oh, very nice.

miqlas:
Btw, i got a demo working: 
xs-

dev-games/godot/godot-2.1.3.recipe
```diff
@@ -0,0 +1,97 @@
...
+		#builtin_speex=no
+		#builtin_opus=no
+		#builtin_freetype=no
+		#builtin_squish=no
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/Godot
+	mv bin/godot.haiku.tools* $appsDir/Godot/Godot
```
diversys: If it's the only binary there is no need to put inside $appsDir/Godot/Godot

PR #1282, opened 2017-04-22
Opensound: fix for AMD64

miqlas:
Maybe something like this patch could be required on x86 too.

### DIFF
--- a/media-sound/opensound/patches/opensound-4.2_git.patchset
+++ b/media-sound/opensound/patches/opensound-4.2_git.patchset
@@ -1,4 +1,4 @@
-From 6227a674404258d70a8ac57aa355d668a763c215 Mon Sep 17 00:00:00 2001
+From 91023b01d5e1a1af5184d17ced555966dbce72f8 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Fran=C3=A7ois=20Revol?= <revol@free.fr>
 Date: Thu, 7 Aug 2014 22:20:13 +0200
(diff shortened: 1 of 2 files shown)


media-sound/opensound/patches/opensound-4.2_git.patchset
```diff
@@ -80,16 +80,12 @@ index 8083e50..b9ab441 100644
...
  static void
  add_kernel_flags (FILE * f)
  {
 -  fprintf (f, "CFLAGS=-O2 -D_KERNEL -D_KERNEL_MODE=1 -no-fpic\n");
-+  #if defined __x86_64__
-+    fprintf (f, "CFLAGS=-O2 -D_KERNEL -D_KERNEL_MODE=1\n");
-+  #else
-+    fprintf (f, "CFLAGS=-O2 -D_KERNEL -D_KERNEL_MODE=1 -no-fpic\n");
-+  #endif
++  fprintf (f, "CFLAGS=-O2 -D_KERNEL -D_KERNEL_MODE=1\n");
```
mmuman: Are you sure this doesn't break 32bit builds? I don't remember but usually -no-fpic is there to save a register being spilled by the compiler and avoid running out of them. Especially with gcc2 which is really picky...

miqlas:
Shouldn’t it be -fno-pic?
Not tested on 32 bit btw.

miqlas:
Can't test on other platforms, closing for now.

mmuman:
I'd keep it open as a reminder to test it.

PR #1295, opened 2017-04-23
lnlauncher: various fixes

m199:
Translations are working again, builds with gcc5 and on x86_64 are working now.


waddlesplash:
Is it still really v1.1.2? The version hasn't changed?

m199:
well, i didnt increased the version, on bitbucket its still "1.1" so maybe we should increase it to 1.1.3 for haikuporter?

PR #1299, opened 2017-04-24
DEVIL: updated recipe, bump version

miqlas:


media-libs/devil/devil-1.8.0.recipe
```diff
@@ -54,20 +67,33 @@ BUILD_PREREQUIRES="
...
 BUILD()
 {
-	runConfigure ./configure CFLAGS=-Dpng_set_gray_1_2_4_to_8=png_set_expand_gray_1_2_4_to_8
+	cd DevIL
+	mkdir -p haiku-build
+	cd haiku-build
+
+	cmake .. \
+		-DCMAKE_INSTALL_PREFIX:PATH=$prefix \
+		-DCMAKE_CXX_FLAGS="-fpermissive"
```
waddlesplash: Does this project use GNUInstallDirs in their CMake files? If so, you can use $cmakeDirArgs and probably ditch the directory moves below.

If they still have the configure script, maybe stick with that...

miqlas: No trace of GNUInstallDirs :(
The configure script doesn't works for me.

korli:
Could you eventually check that games-engines/love still works with this update?
Also when you can't test all previously enabled arch, then don't delete the old recipe.

miqlas:
@korli: the current love recipe is literally broken on every platform. So i just updated love to the latest version (oh my...) 
It builds and runs fine.

PR #1302, opened 2017-04-26
serd, updated the old bep file, bump version, wip

Begasus:


Begasus:
The installations fails for the devel package, guessing it has something to do with the "waf" script ... maybe someone else could look into this ...

korli:
libaubio uses waf. Might require patching wscript_build.

Begasus:
Already thought about that, will have a look :)

Begasus:
The information at libaubio didn't work out, but updated to a working version now, tests run ok ...

korli:
Thanks!

PR #1303, opened 2017-04-26
Armadillo: new recipe

miqlas:
Provides no tests.

### DIFF
--- a/sci-libs/armadillo/armadillo-7.800.2.recipe
+++ b/sci-libs/armadillo/armadillo-7.800.2.recipe
@@ -86,5 +86,7 @@ INSTALL()
 
 TEST()
 {
-	make test
+	#make test
+	# Armadillo provides no tests through cmake AFAIK
+	true
 }
--- a/sci-libs/armadillo/armadillo-7.800.2.recipe
+++ b/sci-libs/armadillo/armadillo-7.800.2.recipe
@@ -1,0 +1,90 @@
+SUMMARY="A C++ linear algebra library"
+DESCRIPTION="The library provides efficient classes for vectors, matrices and \
+cubes, as well as 200+ associated functions (eg. contiguous and non-contiguous \
+submatrix views). Various matrix decompositions are provided through \
+integration with LAPACK, or one of its high performance drop-in replacements."
+HOMEPAGE="http://arma.sourceforge.net/"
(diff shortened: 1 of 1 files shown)


miqlas:
Something wrong with the hdf5 recipe, as i can't install the generated armadillo package.

My hdf5 package defines himself as:
provides: lib:libhdf5 = 1.8.18 (compatible >= 10)
because it got: libhdf5.so.10.2.1

But the package server says:
nothing provides lib:libhdf5 = 1.8.18.

Maybe i made a mistake in the libhdf5 recipe.

pulkomandy:
There is no relation between the version of the lib and the version of the package. Do not mix them. (sometimes they happend to be identical, but that is not a requirement anywhere). Try this for PROVIDES:

    hdf5 = 1.8.18
    lib:libhdf5 = 10.2.1

and if there is a libhdf5.so.10 in the package:

    lib:libhdf5 = 10.2.1 compat >= 10

miqlas:
@PulkoMandy: thanks for the explanation, i'll fix the hdf5 recipe then.

sci-libs/armadillo/armadillo-7.800.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+BUILD()
+{
+	mkdir -p build && cd "$_"
+
+	cmake .. \
+		-DCMAKE_INSTALL_PREFIX=$prefix \
+		-DDETECT_HDF5=ON \
+		-DBUILD_SHARED_LIBS=ON \
+		-DINSTALL_INCLUDE_DIR=$includeDir \
+		-DINSTALL_DATA_DIR=$libDir
```
korli: Why not $dataDir?

sci-libs/armadillo/armadillo-7.800.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+}
+
+INSTALL()
+{
+	cd build
+	make install
+	strip $libDir/lib*
+
+	mkdir -p $libDir/cmake/
+	cp -r $libDir/Armadillo/CMake/. $libDir/cmake/.
```
korli: Well, is there a recipe which needs armadillo that you could test this against?

miqlas:
DATA_DIR is for .cmake files.
Tested already against mlpack, works ok.

PR #1306, opened 2017-04-27
quilt, bump version, update old bep file to working recipe

Begasus:
Haven't deleted the old bep file/patch (yet), maybe someone has a reason for the patch file to be imported?


waddlesplash:
Nope, looks like you can just remove the patch.

korli:
Please review why SECONDARY_ARCHITECTURES is needed when x86_gcc2 is enabled as primary architecture: this package provides commands and no libraries, which makes it useless to build on a secondary arch.

Begasus:
My bad on the SECONDARY_ARCHITECTURES, I added them to check the x86 build on gcc2h, will update in a bit ...

PR #1315, opened 2017-05-02
putty, bump version, remove old bep/patch file

Begasus:


net-misc/putty/patches/putty-0.60.patch
```diff
@@ -1,24 +0,0 @@
...
-+AC_CHECK_HEADERS([utmpx.h])
-+AC_CHECK_HEADERS([sys/select.h])
-+AC_CHECK_HEADERS([sys/types.h])
-+AC_CHECK_HEADERS([utmp.h]) 
- 
--AM_PATH_GTK([1.2.0], [all_targets="all-cli all-gtk"], [all_targets="all-cli"])
-+#AM_PATH_GTK([1.2.0], [all_targets="all-cli all-gtk"], [all_targets="all-cli"])
- AC_SUBST([all_targets])
- 
- AC_SEARCH_LIBS([socket], [xnet])
```
scottmc: Why remove the other check, wouldn't that break it on other platforms?  Goal should be to write patches such that they could be upstreamed, so that we don't have to repatch for every new release.

net-misc/putty/putty-0.69.recipe
```diff
@@ -0,0 +1,55 @@
+SUMMARY="A Free Telnet/SSH Client"
+DESCRIPTION="PPuTTY is a free implementation of SSH and Telnet for Windows \
```
diversys: PPuTTY?

net-misc/putty/putty-0.69.recipe
```diff
@@ -0,0 +1,55 @@
...
+and Unix platforms, along with an xterm terminal emulator.\
```
diversys: There should probably be no slash at the end of the line.

PR #1316, opened 2017-05-03
libsigsegv: bump version and update recipe format.

return:


dev-libs/libsigsegv/libsigsegv-2.11.recipe
```diff
@@ -0,0 +1,85 @@
...
+	"
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	aclocal -I m4
+	autoconf
+	autoheader
+	automake
+	runConfigure ./configure --prefix=$prefix --includedir=$includeDir \
```
waddlesplash: If you're using runConfigure, these two args should not be needed.

return: @waddlesplash Added mentioned changes to recipe, is this recipe ready to be merged?

return:
x86_gcc2 was tested by @Begasus and works perfectly.

PR #1321, opened 2017-05-06
Ruby: bump version

miqlas:


waddlesplash:
But does it work?

miqlas:
@waddlesplash : on x86_64 for sure, so i think it will work with other arch too.

PR #1325, opened 2017-05-06
spiff, revive old bep file to working recipe

Begasus:


sys-apps/spiff/spiff-1.recipe
```diff
@@ -1,58 +1,50 @@
...
+significant differences.  It also compares floating-point values with a \
+tolerance, so you can look in data files for discrepancies."
 HOMEPAGE="https://github.com/HaikuArchives/Spiff"
-SOURCE_URI="git+https://github.com/HaikuArchives/Spiff#50fa5e3cfe1431071b1de57828508f6903b052c2"
+COPYRIGHT="1988 Bellcore"
+LICENSE="SPIFF"
 REVISION="1"
-#STATUS_HAIKU="stable"
-ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+SOURCE_URI="https://github.com/HaikuArchives/Spiff/archive/master.zip"
```
miqlas: Yoou should use a git revision instead of "master" because it doesn't define clearly, which version is it, and you need to fix the checksum after every commit.

Begasus: I would, if I could find one ;)
A regular git checkout gives me a warning for a not secure download ...

humdingerb: The last commit is 89f030f4d88cc85092783c2344f69625be564e98, so this should work:

srcGitRev="89f030f4d88cc85092783c2344f69625be564e98"
SOURCE_URI="https://github.com/HaikuArchives/Spiff/archive/$srcGitRev.tar.gz"

Begasus: Ah, didn't know how to do it, I could pull the git source, but hp would tell me it's an unsecure download (no checksum) ... will look into it ;)

Begasus: updated with the comments of @humdingerb ... thanks!

PR #1326, opened 2017-05-07
Hello: new recipe

miqlas:
The plan is to have a nice example what can show the functionality of HaikuPorter for the newcomers.


miqlas:
The plan is to have a nice example what can show the functionality of HaikuPorter for the newcomers.

waddlesplash:
Already exists: https://github.com/haikuports/haikuporter/tree/master/generic -- though it looks like this example has some more than those do, maybe contribute it there.

Closing as invalid.

PR #1327, opened 2017-05-07
Boehm GC: add AMD64 support, recipe cleanup

miqlas:


miqlas:
The 2.6.x series is entirely different, needs a big recipe and maybe a patchset rework. But they stil lsupports this one.

About the test: yeah, it is 0/10. :(
Maybe somebody could look into it.

miqlas:
Doesn't work according the tests on x64, cannot test on other archs, so i suppose i just close it for now.

PR #1328, opened 2017-05-07
Ecl:  recipe cleanup

miqlas:


korli:
x86_64 isn't really enabled, is it?

miqlas:
It still not compiles on x64 :(
I think i shouldn't push my commits before it is fully working.

miqlas:
64 bit disabled, as it doesn't build right now.
Maybe you can merge it as cleanup or close it as invalid.

PR #1337, opened 2017-05-13
makedepend, bump version, remove old bep file

Begasus:


x11-misc/makedepend/makedepend-1.0.5.recipe
```diff
@@ -0,0 +1,44 @@
...
+HOMEPAGE="http://xorg.freedesktop.org/releases/individual/util/"
+COPYRIGHT="1993, 1994, 1998  The Open Group
+	1992 Network Computing Devices, Inc."
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="http://xorg.freedesktop.org/releases/individual/util/makedepend-$portVersion.tar.bz2"
+CHECKSUM_SHA256="f7a80575f3724ac3d9b19eaeab802892ece7e4b0061dd6425b4b789353e25425"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
korli: Why a secondary arch?

Begasus: I mearly added them as the requirement libs (util-macros and xproto) are also available as secondary arch) :)
If not maybe they could be changed too? (don't know where allong the line it probably would break on some x11 dependency)

waddlesplash: No, makedepend is a command, so it only needs to be built on the primary arch I believe.

Begasus: To build it as secondary arch I needed to enable it on gcc2h, if it's not needed I can remove the $secondaryArchSuffix

waddlesplash: Yes, remove all the secondaryArchSuffix.

PR #1348, opened 2017-05-17
ffmpeg: introduce ffmpeg_gpl for non-free stuff. Enabled encoders.

miqlas:


pulkomandy:
Is that needed? What is the problem with having GPL code in ffmpeg?

miqlas:
After talking a bit about it with korli i concluded it would be better to have a different recipe for the non-free stuff, so if somebody want to use the x26x stuff, can install this package.

Have you got better idea?

pulkomandy:
I'm confused now. Is the problem that some code is under the GPL, or that we are afraid of infringing on patents?

If it is GPL, then I would not worry: Haiku already includes GPL code eg. for NTFS filesystems handling. I would provide an ffmpeg package (with GPL stuff) and possibly an ffmpeg_nogpl with the GPL stuff disabled, in case someone would want that.
If it is patents, then ffmpeg_gpl does not sound like the correct name. But likewise, I would go with ffmpeg and ffmpeg_nopatents packages.

miqlas:
Let us wait a bit and collect opinions how thie situation could be handled.

korli:
Adrien, please see how we include the ffmpeg add-on in the build as a non GPL component: http://cgit.haiku-os.org/haiku/tree/src/add-ons/media/plugins/Jamfile#n20
But it links against an at the moment GPL build feature. 
Either we switch the ffmpeg add-on to SubIncludeGPL or the included package is LGPL (another GPL package would be provided). The naming isn't important in the end.

korli:
@miqlas it should be possible to build ffmpeg two times within the same recipe: build_gpl, build_lgpl and arrange packages correctly by license terms, aka ffmpeg_gpl_devel would only contain libpostproc, and depends on ffmpeg_devel.

miqlas:
@korli: have you got any recipe as example?
Should i build it in 2 different object folder?
As far i see, ffmpeg uses only the --enabled and --with libs, so if my recipe defines everything as build-req, the gpl-free version will use only the defined thing, it is good, but i'm not sure if we can mix and match files from 2 different build.

korli:
@miqlas yeah two build folders

miqlas:
I don't really know, how to accomplish it, so i just close it for now as it doesn't too much work to recreate my work.

korli:
Sorry that I couldn't help to get it further.

miqlas:
@korli: i just compared my build with the current one from HaikuDepot, and in my version not just the libpostproc requires libx264 and 5, but almost all the .so, so i think we cannot mix and match them. Somebody with deeper knowledge about library dependencies and law should do it if it is really needed.

PR #1349, opened 2017-05-17
muparser, new recipe/patch file

Begasus:
builds and installs ok on gcc2h (x86 arch), the .pc file doesn't contain the correct directions to the libs nor headers, I disabled $prefix for now, if it's set the pkg-config file is ok but I get a policy error for a top level include folder ... wip (but wanted to share already) :)

### DIFF
--- a/dev-cpp/muparser/muparser-2.2.5.recipe
+++ b/dev-cpp/muparser/muparser-2.2.5.recipe
@@ -1,0 +1,68 @@
+SUMMARY="Mathematical expressions parser library"
+DESCRIPTION="Many applications require the parsing of mathematical \
+expressions. The main objective of this library is to provide a fast \
+and easy way of doing this. muParser is an extensible high performance \
+math expression parser library written in C++. It works by transforming \
+a mathematical expression into bytecode and precalculating constant parts \
+of the expression."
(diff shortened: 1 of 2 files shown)


Begasus:
Updated the recipe, doing a manual install as "make install" insist on creating a ./self/include directory, need to check if it build for gcc2 ...

dev-cpp/muparser/muparser-2.2.5.recipe
```diff
@@ -0,0 +1,68 @@
...
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	cp `finddir B_SYSTEM_DATA_DIRECTORY`/libtool/config/config.guess build/autoconf/
+	cp `finddir B_SYSTEM_DATA_DIRECTORY`/libtool/config/config.sub build/autoconf/
```
diversys: Doesn't libtoolize do basically the same thing?

Begasus: have to re-check for this, not sure, normally I check things out in Terminal first ...

Issue #789, opened 2016-08-31
clang++ does not support TLS


Alexco500:
It seems to me that clang++ does not support TLS on haiku. Just check with a small test code:

```
~/Desktop/source/libobjc2-1.7/Buils> cat ./test.cc 

int forward(int receiver, int selector)
{
        __thread static int forwardingSlot;
        forwardingSlot = receiver+selector;
        return forwardingSlot;
}
```

Compile it with gcc (search x86):

```
~/Desktop/source/libobjc2-1.7/Buils> g++ ./test.cc 
./test.cc: In function 'int forward(int, int)':
./test.cc:4:2: warning: '__thread' before 'static'
  __thread static int forwardingSlot;
  ^
/boot/system/develop/lib/x86/start_dyn.o: In function `_start':
start_dyn.c:(.text+0x4c): undefined reference to `main'
collect2: error: ld returned 1 exit status
```

Compiles ok, but does not link, which is correct in this case.

With clang++ I get:

```
~/Desktop/source/libobjc2-1.7/Buils> clang++ ./test.cc 
./test.cc:4:2: error: thread-local storage is not supported for the current
      target
        __thread static int forwardingSlot;
        ^
1 error generated.
```

Since g++ seems to support TLS (and haiku obviously does it, too), clang++ should also support that.

return:
I recently ran into this issue when building LLVM with clang, needed by swift. This commit here: https://github.com/llvm-mirror/clang/commit/ef7bceadf4b43c953855013577afac4c2fcb1d62 disables Haiku thread-local storage for clang.

pulkomandy:
Maybe try enabling it and see what happens. I think we added TLS support on Haiku side since then?

return:
@pulkomandy I tested this on hrev51173 x86_64 with llvm-4.0.0 and it compiled the above test code and it works fine.

return:
Has anyone tested llvm-4.0.0-2 on x86 and tried compiling the test code? Until this is confirmed, I think this issue can be closed.

return:
The test code compiles and runs fine on x86_gcc4h, when fPIC is passed. Tested on hrev51193 with clang-4.0. Is this enough for this issue to be considered as solved?

korli:
Let's close this, support for PIE isn't specific to clang.

PR #1362, opened 2017-05-23
clang: Enable thread-local storage in clang

return:
This commit solves issue #789 by enabling TLS support in clang by default. This is known to work on x86_64 by passing the -fPIC flag, but I haven't tested it on x86 yet.

### DIFF
--- a/sys-devel/llvm/llvm-4.0.0.recipe
+++ b/sys-devel/llvm/llvm-4.0.0.recipe
@@ -31,7 +31,7 @@ other than the ones listed above.
 HOMEPAGE="http://www.llvm.org/"
 COPYRIGHT="2003-2014 University of Illinois at Urbana-Champaign"
 LICENSE="UIUC"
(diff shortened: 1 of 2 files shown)


korli:
You should probably disable PIE by default on x86_64 then.

return:
@korli  I've tried disabling PIE/PIC via -DLLVM_ENABLE_PIC=NO, but this fails to link a non-pie object. Is it related to this ticket: https://dev.haiku-os.org/ticket/13292 with gcc5?

korli:
Enabling PIE on x86_64 happens here : https://github.com/llvm-mirror/clang/blob/master/lib/Driver/ToolChains/Haiku.h#L27

return:
@korli PIE on x86_64 has been disabled.

korli:
@return incorrect, default PIE was only enabled on x86_64. it's now enabled for x86.

PR #1366, opened 2017-05-25
libzip, move zipconf.h to the right place

Begasus:


dev-libs/libzip/libzip-1.2.0.recipe
```diff
@@ -73,14 +73,14 @@ INSTALL()
...
+	# Move zipconf.h from lib/libzip/include/ to develop/headers/
+	mkdir -p $includeDir
+	cp $libDir/libzip/include/zipconf.h $includeDir
+	rm -R $libDir/libzip
 
 	# Adjust libincludedir to make it point to where zipconf.h was moved.
-	sed -i -e "s,^\(libincludedir=\)$libDir"'\(/libzip\)/include$,\1${prefix}/'"${relativeIncludeDir}\2," \
-		$developLibDir/pkgconfig/libzip.pc
+	#sed -i -e "s,^\(libincludedir=\)$includeDir"'\(/libzip\)/include$,\1${prefix}/'"${relativeIncludeDir}\2," \
+	#	$developLibDir/pkgconfig/libzip.pc
```
waddlesplash: Will this be needed in the future? If not, it should probably be removed rather than commented out.

Begasus: from the libzip.pc file, so it looks like those can be removed?

  prefix=/packages/libzip-1.2.0-2/.self
  exec_prefix=${prefix}
  libdir=/packages/libzip-1.2.0-2/.self/lib
  includedir=/packages/libzip-1.2.0-2/.self/develop/headers
  libincludedir=/packages/libzip-1.2.0-2/.self/develop/headers

Begasus:
Don't merge (yet), looks like the current version is ok with Haikuporter, I tried to build ebook-tools and the recipe works in haikuporter, if I try to build ebook-tools however from a Terminal it tells me it can't find the zipconf.h file (with this patched version from libzip it does however) ...

dev-libs/libzip/libzip-1.2.0.recipe
Begasus: This is moved to the patchset

Begasus:
This is working ok for ebook-tools ...

Begasus:
checked with odt2txt, that one is working fine also with these changes

PR #1367, opened 2017-05-27
ebook_tools, bump version, remove old bep file

Begasus:


Begasus:
Will check in a bit :)

Begasus:
Getting the zipconf.h problem with the current version of libzip ... (for x86)
Scanning dependencies of target epub
[ 12%] Building C object src/libepub/CMakeFiles/epub.dir/epub.o
In file included from /sources/ebook-tools-0.2.2/src/libepub/epublib.h:9:0,
from /sources/ebook-tools-0.2.2/src/libepub/epub.c:2:
/boot/system/develop/headers/x86/zip.h:59:21: fatal error: zipconf.h: No such file or directory

Begasus:
on x86_64 it suffers from the same problem, can't find zipconf.h
for x86 with the fixed libzip recipe it fails on finding xmlreader.h ... (it checks in libxml/* instead of what we have libxml2/*)

korli:
Did you try to add cmd:pkg_config to BUILD_PREREQUIRES?

Begasus:
still no luck, even with pkg-config added, seems we don't suffer from it alone, there is a patch online from others, but that doesn't seem to work (missing libzip cmake module?):
[https://github.com/pisilinux/main/blob/master/office/misc/ebook-tools/files/ebook-tools-0.2.1-libzip_pkgconfig.patch](url)

Begasus:
Builds ok now with the recent changes to libzip, also checked odt2txt as that one also uses libzip, added secondaryArchSuffix (builds ok for x86_gcc2, x86 and x86_64)
PR for libzip ...
https://github.com/haikuports/haikuports/pull/1366

PR #1369, opened 2017-05-27
luafilesystem, bump version, fix recipe

Begasus:


dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
```diff
@@ -0,0 +1,54 @@
...
+
+BUILD()
+{
+	make LUA_INC="$(finddir B_SYSTEM_DEVELOP_DIRECTORY)/headers/lua$LUAVERSION_WITHOUT_DOTS"
+}
+
+INSTALL()
+{
+	make LUA_LIBDIR=$libDir install
+	#make LUA_LIBDIR="$libDir/lua/$LUAVERSION_WITH_DOTS" install
```
Begasus: Not sure what path to go would be the best one here ...

korli: $libDir/lua/$LUAVERSION_WITH_DOTS is better (what arch linux uses btw).

Begasus: Right, there is already the subfolders for 5.* versions, just thought it would be safer to ask ;)

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
```diff
@@ -0,0 +1,54 @@
...
+COPYRIGHT="2004-2012 Kepler Project"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/keplerproject/luafilesystem/archive/v_1_6_3.tar.gz"
+CHECKSUM_SHA256="5525d2b8ec7774865629a6a29c2f94cb0f7e6787987bf54cd37e011bfb642068"
+SOURCE_DIR="luafilesystem-v_1_6_3"
+
+#check for Lua version, we need to remove the dot for the headers
+LUAVERSION_WITH_DOTS=$(lua -v | sed 's/Lua //' | head -c3)
+LUAVERSION_WITHOUT_DOTS=$(echo $LUAVERSION_WITH_DOTS | tr -d '.')
```
korli: LUAVERSION_WITH_DOTS should be moved in the INSTALL section, while LUAVERSION_WITHOUT_DOTS should be moved in the BUILD section. Reason is this calls a non-chroot shell which possibly doesn't have lua (or the right version)

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
```diff
@@ -0,0 +1,54 @@
...
+	haiku
+	lua
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	lua_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:lua
```
korli: you should define the lua version supported.

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
```diff
@@ -0,0 +1,54 @@
...
+	lib:lfs
+	"
+REQUIRES="
+	haiku
+	lua
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	lua_devel
```
korli: Should be devel:liblua with the version ie >= 5.3

korli:
On a side note, it seems possible to build luafilesystem for multiple lua versions, and package them accordingly. A problem is that at the moment the lua_devel, lua52_devel are declared as conflicting, which they might not. For instance, the pkgconfig file for lua51 is named lua.pc, should be lua51.pc to avoid the conflict.

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
Begasus: Been doing some checks yesterday (still wip) where I renamed LUAVERSION_WITH_DOTS to LUAVERSION  (it gives the dot default) and LUAVERSION_WITHOUT_DOT (as there is only one dot in there) ...

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
Begasus: wouldn't this break the other versions? for instance on x86_gcc2 there is still 5.2 as default against x86_64 where 5.3 is the default installed one ...

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
Begasus: iirc I already have lua_devel >= $LUAVERSION in there (not in Haiku atm) ...

Begasus:
There are more conflicts, for instance the headers for lua51 are installed in "5.1", after that they are installed in "52" and "53", there has been a chance after the 5.1 version on the install side from Lua, maybe this could be looked into also?

dev-lua/luafilesystem/luafilesystem-1.6.3.recipe
korli: if the lua version is available at Haikuports for x86_gcc2, this is just another package to build.

korli:
Why having headers in lua52 or lua5.2 is a conflict?

Begasus:
Default search goes out to the '.' version, if there is 52 it isn't found, this should be 5.2 (looking at configure scripts  to check for lua) ..

Begasus:
Updated the PR, added secondaryArchSuffix, builds for x86_gcc2, x86 and x86_64 now ...

PR #1390, opened 2017-06-05
odt2txt, no need to set the path for libzip anymore

Begasus:


app-text/odt2txt/odt2txt-0.5.recipe
```diff
@@ -8,9 +8,9 @@ LICENSE="GNU GPL v2"
 REVISION="1"
 SOURCE_URI="https://github.com/dstosberg/odt2txt/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="23a889109ca9087a719c638758f14cc3b867a5dcf30a6c90bf6a0985073556dd"
-PATCHES="odt2txt-$portVersion.patch"
+PATCHES="odt2txt-$portVersion.patchset"
 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="any"
```
korli: Why change to any?

Begasus: thought it would be ok as it's ok on all platforms?

korli: No, any means that a same package works on all architectures.

Begasus: ah, my bad, will change later, thanks for the update :)

PR #1398, opened 2017-06-07
redis, bump version, remove old bep/patch file

Begasus:
not checked (yet) on x86_64, but should be ok? :)


dev-db/redis/redis-3.2.9.recipe
```diff
@@ -0,0 +1,52 @@
+SUMMARY="Advanced key-value store"
+DESCRIPTION="
+Redis is an open source, advanced key-value store. It is often referred to as \
```
diversys: Start with the previous line

dev-db/redis/redis-3.2.9.recipe
```diff
@@ -0,0 +1,52 @@
...
+a data structure server since keys can contain strings, hashes, lists, sets \
+and sorted sets.
+"
```
diversys: Move to the previous line

Begasus: I'll move those over tomorrow (if I get to it), copied these lines over from the previous version ...

PR #1426, opened 2017-06-14
Qmplay2 bump version

khallebal:
For some reason Cmake requires glib2, harfbuzz and fontconfig for libass, and graphite2 for harfbuzz, but glib2 is only required on gcc2h not on pure X86.
youtube doesn't work, but i suspect a bug in our Qt5 port.
built with Qt5.5.1 on pure x86 and 5.6.1 on gcc2h.


waddlesplash:
@khallebal Why the close? Sorry I didn't get around to reviewing it yet.

khallebal:
Because, a) it seemed like it failed 1 test, b) i pushed the wrong recipe and because i'm new to github i don't know how to delete a pull request in order to start over , so any help would be nice.
I can reopen it though if there is a way to fix my mistakes.
Thanks.

PR #1430, opened 2017-06-16
Pe packaging fix

miqlas:
Dirty fix for missing libhekkel.
Now packages fine.

### DIFF
--- a/haiku-apps/pe/pe-2.4.5.recipe
+++ b/haiku-apps/pe/pe-2.4.5.recipe
@@ -15,7 +15,7 @@ SOURCE_URI="https://github.com/HaikuArchives/Pe/archive/7f4ca18.tar.gz"
 SOURCE_FILENAME="pe-$portVersion.tar.gz"
 SOURCE_DIR="Pe-7f4ca18b106c50cfdb3a4b699cb43ec60ea32499"
 CHECKSUM_SHA256="6c85968ae73ea6606a7665ec9f73844908327f43405121a7fe5851af4bcb187d"
-REVISION="5"
+REVISION="6"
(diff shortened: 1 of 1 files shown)


waddlesplash:
Does it actually provide libhekkel though (i.e. libhekkel is installed in `lib`) or is libhekkel just in `$appsDir/Pe/lib`?

miqlas:
It is in $appsDir/Pe/lib, so no.

waddlesplash:
So what's the full error then? Is `lpe` dependent on it? Otherwise, I think HaikuPorter is probably wrong here then.

miqlas:
Missing required lib: libhekkel, or something like that.

> 2017. jún. 16. dátummal, 13:32 időpontban waddlesplash <notifications@github.com> írta:
> 
> pe dependent on it? Otherwise, I think HaikuPorter is probably wrong here then.

pulkomandy:
Haikuporter scans all apps and binaries for required libs, and check that these libs are either in REQUIRES, or part of the package. But it does not check for a $appsDir/lib directory, which is quite an unusual case: the library is shared between different binaries which are part of the package (if there was just one, we would link the lib statically in it), but is useless to other packages.

Possible solutions:
- Haikuporter should look for libs next to the executable which requires them, or
- A special syntax for declaring such libraries could be added to recipes

waddlesplash:
> Haikuporter should look for libs next to the executable which requires them, or

This is the better solution IMO.

Begasus:
Is this fixed with the latest change to haikuporter?

korli:
Shouldn't be needed.

PR #1438, opened 2017-06-17
unarr: fix src_uri, bump rev

miqlas:


diversys:
No need to change revision

miqlas:
@diversys : i thought if i change a recipa, and don't bump version, then i have to bump revision, to let the buildbots know it is changed. Is it wrong?

diversys:
Lets's check with @waddlesplash

KapiX:
@miqlas You have to bump revision only if the RESULTING package has changed. If the SHA sum of the archive stays the same and none of the package metadata has changed, then there is no need for that.

PR #1441, opened 2017-06-17
ne recipe: update to 3.1.1

humdingerb:
It does build and seems to work in general, but I don't think the macros work.
It also uses ~/.ne as config folder, which I tried to change but failed...


diversys:
Maybe you can git mv old_recipe new_recipe?

humdingerb:
I could, but wasn't sure my INSTALL() was acceptable, seeing as the older version had a "haiku" target for the makefile in the patch. (I didn't see the sense in that.)

diversys:
Either that or also drop the old recipe :)

humdingerb:
Before I remove the old recipe and merge the PR, currently there's a "ne-x86" link to /bin/x86/ne in /bin. Is that OK, or is there a way to name that symlink just "ne"?

waddlesplash:
I think that's OK, for now anyway.

humdingerb:
All righty then. Thanks for the input, guys!

PR #1444, opened 2017-06-18
libmikmod, bump version

Begasus:


korli:
Why don't you drop 3.3.10?

Begasus:
Still can do it ;)

Begasus:
PS, libmikmod can be compiled against libSDL2 also, but don't know if it's possible to include that into one recipe or create another recipe for that or add a comment in the recipe, any ideas would be welcome :)

korli:
No need to depend on sdl2 IMO.

Begasus:
if you want support for sdl2_mixer it needs SDL2 ...

korli:
3.3.11.1 is out! Not sure why sdl2_mixer needs libmikmod too...

Begasus:
I'll look into it :) it doesn't 'need' libmikmod, but it can support mod files over libmikmod (sdl2_mixer) and fails to not finding libSDL if you don't compile it against libSDL2 :)

korli:
Well looking at https://www.archlinux.org/packages/extra/x86_64/sdl2_mixer/, no dependency on libmikmod, because they chose libmodplug.
https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/sdl2-mixer/sdl2-mixer-2.0.1.ebuild has one optionally (instead of libmodplug), and https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/libmikmod/libmikmod-3.3.11.1.ebuild disables sdl and sdl2.
I suppose we would need sdl2 for libmikmod because there is no other backend. The problem is that libmikmod has already sdl as dependency. I find not good to depend on sdl and sdl2 in an application, and foresee problems.

korli:
maybe simply switch sdl2_mixer to libmodplug, and it's OK.

Begasus:
Haven't checked with libmodplug (will do), maybe it could be an option to disable sdl all together for sdl in libmikmod? (not sure if it would bread other things?)

Begasus:
removing the dependency for sdl works ok, first check with sdl2_mixer (gcc2) succeeded :)
downside to removing sdl from the recipe is that mikmod player isn't using the sdl sound driver (so no output there atm ...)

Begasus:
imho this one could be merged, it's based on the previous work for it, I added libmodplug for sdl2_mixer so this solves the issue for libmikmod ...

PR #1449, opened 2017-06-19
boehm_gc: bump version

return:
This commit adds support for x86_64 for boehm_gc and passes 13 / 14 tests.
Here are the results from the [testsuite.](https://gist.github.com/return/21e4d5484afe9c995a47a09f9c2c0094) x86 and x86_gcc2 remain untested.


korli:
Maybe consider submitting a PR at https://github.com/ivmai/bdwgc

return:
I'll see if this works on x86/x86_gcc2, then I'll upstream these changes.

PR #1452, opened 2017-06-20
QMPlay2: bump version

khallebal:
Here i am again, i hope this time it'll work.

Cmake requires harfbuzz, fontconfig and fribidi for libass,it also requires graphite2 and glib2 for harfbuzz, the glib2 is only required when building on gcc2h but not on pure x86, is this a normal behaviour of cmake or is it just our port that's acting?
Youtube doesn't work, i get connection error msg if try i do a search, it used to with Qt4, so i suspect a bug in Qt5.


korli:
The patchset filename should be qmplay2-17.06.09.patchset

korli:
About build dependencies you added, this is mostly wrong because for instance the recipe for libass in master already includes the dependencies for fontconfig, fribidi and harfbuzz. Just make sure your local pkgs are in sync with master. I know it's mostly a problem of HaikuDepot on x86_gcc2, maybe leave this one out for now, given that the app doesn't really work anyway?

khallebal:
`Just make sure your local pkgs are in sync with master.```
I've installed all my hpkgs from the official repo, so i don't know why cmake required those extra deps. it doesn't even make sense?

diversys:
So it turns out you were doing `git mv` correctly as I've just run into the same issue.

@pulkomandy nicely explained what happened:
> git doesn't track moves/renames, it detects them when the files are similar enough
> (and sometimes wrongly detect two similar files as a copy when they are in fact unrelated)
> if you want to be really sure it gets it right, move, commit, make changes, commit again
> otherwise if there are too many changes it won't think it is the same file

korli:
Superseded by 9d6b017

PR #1455, opened 2017-06-21
Libelf : enable on x86 64

miqlas:


korli:
Why bumping the revision?

PR #1457, opened 2017-06-21
libdwarf: bump version

miqlas:
Still broken, help wanted:
Warning: POLICY ERROR: "/packaging/libdwarf/bin/dwarfdump" needs library "libdwarf.so.1", but the package doesn't seem to declare that as a requirement


korli:
Well when the problem already exists in the previous version, a bug report could be submitted and the PR merged.

PR #1459, opened 2017-06-21
OIIO: bump versoin

miqlas:
And some enabled stuff.


korli:
You should explain why new dependencies are added.

PR #1461, opened 2017-06-22
Nim: new recipe

miqlas:
For johnny_b


dev-lang/nim/nim-0.17.0.recipe
```diff
@@ -0,0 +1,103 @@
...
+	"
+
+PATCH()
+{
+	sed -i "s|@BINDIR@|$binDir|" install.sh
+	sed -i "s|@CONFIGDIR@|$settingsDir/nim|" install.sh
+	sed -i "s|@LIBDIR@|$libDir/nim|" install.sh
+	sed -i "s|@DOCDIR@|$docDir|" install.sh
+	sed -i "s|@DATADIR@|$dataDir/nim|" install.sh
+	sed -i "s|@NIMBLEDIR@|$dataDir/nim|" install.sh
```
korli: Warning for such "updates": the revision number is included in $libDir or $binDir, thus when bumping the revision number, the sources are already patches, install.sh won't be updated with the new revision.

korli: Why not simply going with the environment variables?

waddlesplash: +1 for environment variables

PR #1463, opened 2017-06-22
MPlayer: improved icon, it looks almost like the official one.

miqlas:
The earlier one looks a bit ehh in Deskbar


Giova84:
Hi, is nice, but why not to use the McClintock icon which is more in Haiku's style? :-)
Mplayer icon

miqlas:
‘cause unlike his other icons it looks terrible.

Giova84:
Obviously is a matter of personal taste :-)
So I will avoid to propose also the Haiku's style icon for QupZilla ;)
QupZilla Haiku's icon
To me, this icon, looks better than the newer one (I talk about Qupzilla)

humdingerb:
Though OT... :) I, too, like the old QupZilla icon better. But the QupZilla guys went with a new one, so I guess it's polite to follow. But we could work on a haikuified version, i.e. projected on sphere.

humdingerb:
@miqlas : Could you add the IOM file to the "additional-files"? Then it's easier to modify the icon. (Applies to your other added icons as well.)

miqlas:
One can convert the hrdef resource to iom, right?

humdingerb:
I think one can with IIRC "rc" and stuff, but why make it so difficult?
Sure, one can get the package and get the icon from the filetype, but that still loses the names of the shapes, styles, paths. Adding the IOM is really the easiest and most convenient.

miqlas:
They are svg Icons converted to hvif, shapes got default names. No Data loss i suppose,

PR #1473, opened 2017-06-26
gnupg, remove not installed binaries in the recipe

Begasus:
Some of the cmd: lines are not installed by default, no need to have them in the recipe (untill further notice) :)


miqlas:
Are you sure this package cannot provide cmd:gpg?
It is missing, clear, but why does it missing?

miqlas:
https://github.com/haikuports/haikuports/pull/1478 superseeds this.

PR #1476, opened 2017-06-26
Tinyxml: fix recipe

miqlas:


miqlas:
The PR from Bega looks nicer to me.

PR #1503, opened 2017-07-09
libmad, fix missing pkg-config file

Begasus:


media-libs/libmad/libmad-0.15.1b.recipe
```diff
@@ -46,11 +47,8 @@ BUILD()
 {
 	touch NEWS AUTHORS ChangeLog
 	sed -i 's/AM_CONFIG_HEADER/AC_CONFIG_HEADERS/' configure.ac
-	libtoolize --force --copy --install
-	aclocal-1.13
```
korli: you should also adjust the BUILD_PREREQUIRES when you want to use cmd:aclocal (with autoreconf).

Begasus: thanks missed that one, will fix :)

Issue #155, opened 2015-06-30
WIP recipe for pingus


soakbot:
(Original issue 155 created by scottmc on 2015-01-14T06:03:50.012666+00:00)

This is failing so far.  The $secondaryArhSuffix doesn't seem to be working for boost_signals.  Harding it gets past there but then it fails to find and use sdl_image, sdl_mixer, iconv and more.

soakbot:
(Original comment by korli on 2015-02-10T08:19:51.803516+00:00)

This goes in games-puzzle.

pulkomandy:
Were there attachments to this issue originally? Have they been lost?

scottmc:
Did you try here? http://ports.haiku-files.org/ticket/533

PR #1511, opened 2017-07-13
pingus, new recipe

Begasus:
still some work in progress ...

### DIFF
--- a/games-puzzle/pingus/pingus-0.7.6.recipe
+++ b/games-puzzle/pingus/pingus-0.7.6.recipe
@@ -1,0 +1,65 @@
+SUMMARY="Free Lemmings clone"
+DESCRIPTION="Pingus is a free Lemmings(tm) clone for GNU/Linux, Win32 and \
+other operating systems supported by SDL. It is covered under the GPL."
+HOMEPAGE="https://pingus.seul.org"
+COPYRIGHT="1998-2011 Ingo Ruhnke"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/Pingus/pingus/archive/v0.7.6.tar.gz"
+CHECKSUM_SHA256="c4cd89e1d350d2472f32de5f6266ac9f3658a3620eace3f79efac45db4323b65"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	pingus$secondaryArchSuffix = $portVersion
+	app:Pingus$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libboost_signals$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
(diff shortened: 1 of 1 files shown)


scottmc:
Nice.  Note that there is an open issue for this, but that previous work looks to be lost.
https://github.com/haikuports/haikuports/issues/155

Begasus:
It's there where I picked it up @scottmc  ;)

Begasus:
checked and working on gcc2h and x86_64, 2 issues atm
#1 settings are stored into ~/
#2 the binary pingus.bin doesn't work with Tracker, you need to start it from Terminal (the pingus script is working though)
one could create issues for both of them (it's using a private class for main, so not sure how to solve that) and couldn't find an option to set the userdir (needed for the files to go into $settingsDir) ...

games-puzzle/pingus/pingus-0.7.6.recipe
```diff
@@ -0,0 +1,65 @@
+SUMMARY="Free Lemmings clone"
+DESCRIPTION="Pingus is a free Lemmings(tm) clone for GNU/Linux, Win32 and \
+other operating systems supported by SDL. It is covered under the GPL."
+HOMEPAGE="https://pingus.seul.org"
+COPYRIGHT="1998-2011 Ingo Ruhnke"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/Pingus/downloads/blob/master/pingus-$portVersion.tar.bz2?raw=true"
```
korli: It would be better to use the project download based on a tag:
https://github.com/Pingus/pingus/archive/v0.7.6.tar.gz

korli:
For settings, maybe the following could help: 
https://github.com/Pingus/pingus/blob/master/src/util/system.cpp#L323
````
#elif defined(__HAIKU__)
	BPath settingsDir;
	if (find_directory(B_USER_SETTINGS_DIRECTORY, &path) == B_OK)
		return std::string(path.Path()) + "pingus-0.8";
````

and at the beginning:
````
#ifdef __HAIKU__
  #include <FindDirectory.h>
#endif
````

Begasus:
I'll look into it later, thanks for the pointers!

games-puzzle/pingus/pingus-0.7.6.recipe
Begasus: This one works ...

Begasus:
The last one system.cpp, is not the same anymore as in the above download ...

scottmc:
Also note this recently opened issue there: https://github.com/Pingus/pingus/issues/175

Begasus:
Updated the SOURCE_URI, for system.cpp it's just a bit over my head I think, been fiddling with it for a while but haven't succeeded yet ...

Begasus:
https://github.com/Pingus/pingus/commit/f1031ba8786bea8b4b700b0ddcebaabc9b1997b5
this is where they changed it for system.cpp ...

Begasus:
Could this be merged? Aside from the settingsdir things are working fine so far, maybe we could create an isseu for that one then?

PR #1523, opened 2017-07-21
libgcrypt, bump version

Begasus:
for x86_gcc2 I'm getting "`__GNUC_PATCHLEVEL__' undeclared", it is to a change in globals.c as far as I could trace it, 1.7.8 still builds ok for x86_gcc2 .... could use some attention :)


korli:
AFAIK you shouldn't change the compat 1.6 => 1.8, because, well, it's usually compatible.

Begasus:
Reverted back to compat >= 1.6, ps, how do you know we should use 1.6? ;)

korli:
@Begasus I don't know but basic is backward compatibility. It only makes sense to change the compat string when it's found to be incompatible.

PR #1530, opened 2017-07-24
Libmpdclient : bump version

miqlas:


miqlas:
Short question: should i bump the compat version in this recipe?

PR #1536, opened 2017-07-27
libedit: enable wide-character support

return:
This commit enables libedit to be built with support for wide characters. It is also required to support the integrated REPL in swift which uses the el_wgets symbol which was missing in libedit.


pulkomandy:
Does this use wchar_t? If so, have you tried that it works with gcc2 as well? Otherwise, the option should be enabled only for non-gcc2 builds.

return:
@pulkomandy It uses wchar_t as it says in the [changelog](https://github.com/cdesjardins/libedit/blob/18b682734c11a2bd0a9911690fca522c96079712/ChangeLog#L44) and I built and tested it with the examples on x86_gcc2.

pulkomandy:
Ok then, no problem if it works.
I wasn't sure how far we got with wchar_t support in gcc2. I think only wide char litterals are missing, so if this builds, we should be fine.
Thanks!

PR #1537, opened 2017-07-29
glib2, bump version

Begasus:


dev-libs/glib/patches/glib2-2.53.4.patchset
```diff
@@ -0,0 +1,246 @@
...
++esac
++
+ AC_MSG_CHECKING(thread related libraries)
+ AC_MSG_RESULT($G_THREAD_LIBS)
+ 
+-- 
+2.7.0
+
+
+From 2ae7f7f265f807896f61a19ad715d33a6ac0e6b2 Mon Sep 17 00:00:00 2001
```
Begasus: This one could use some looking into, I tried to re-implement the patch from previous version, but things have changed since then ...

Begasus:
set the credentials to the correct person for the patchset file :)

korli:
Is this PR ready?

Begasus:
I've been using it for quite some time here, so far no problems ... but maybe others could check? At least from my point of view it's ok ...

PR #1545, opened 2017-07-30
Ghostscript: enable shared lib

miqlas:
Introduce ghostscript_devel package


miqlas:
DO NOT MERGE YET, it is completly broken, as the built binaries are not able to run.

miqlas:
Totally broken.

PR #1549, opened 2017-08-02
open-sans, fix download/install

Begasus:


media-fonts/open_sans/open_sans-1.0.recipe
```diff
@@ -8,11 +8,30 @@ HOMEPAGE="http://www.google.com/fonts/specimen/Open+Sans"
...
+SOURCE_URI_9="https://github.com/google/fonts/raw/$srcGitRev/apache/opensans/OpenSans-SemiBold.ttf#noarchive"
+CHECKSUM_SHA256_9="b4c2050b25d3d296d5cf58589ca00816dc72df42262c2f629d5c6a984a161aa4"
+SOURCE_URI_10="https://github.com/google/fonts/raw/$srcGitRev/apache/opensans/OpenSans-SemiBoldItalic.ttf#noarchive"
+CHECKSUM_SHA256_10="a8f2af8e79f46686b1cfcfb3a1fd53e94e88308d7c6ee7f85c733f4796fcc3a0"
 LICENSE="Apache v2"
 COPYRIGHT="2010-2011 Google Corporation"
-REVISION="1"
+REVISION="2"
 
 ARCHITECTURES="?any"
```
korli: Can we enable the arch also?

Begasus: good catch, didn't notice it :) will try to see if I can check it later on x86_64 (don't have x86 install anymore), but as no compiler etc is dealing with it I think it should be ok (tested it with FontDemo on x86_gcc2) ...

PR #1558, opened 2017-08-04
psi_plus-0.16.475.recipe: enable build with QtWebkit

Giova84:


diversys:
Please bump the revision and edit commit message

Giova84:
Where I can change the revision? In the recipe name? So from 0.16.475 to 0.16.476?

korli:
Seems there is trailing space in this recipe (maybe not from your commit)

Giova84:
I am almost sure that I didn't touch somethin' else; in the case I am sorry.
However I cannot figure out where I have to bump the revision...

Giova84:
>Seems there is trailing space in this recipe (maybe not from your commit)

The check still failed and I think is due this trailing space; however, I can't recognize it. Would be fine if someone told me where this trailing space is present and I can fix it; otherwise feel free to fix it: I can learn by looking at the next fixed recipe.

However now I have to go.

diversys:
I meant to edit commit message title to be in line with other commits, see commit log. BTW, no need to mention that you bumped the revision unless it's the only change. There is also a typo there :)

Giova84:
Thanks for committing, @korli!
Hower by looking to the commit, I can say that I didn't introduced the trailing spaces :-) I just added `--enable-webkit`.
However the next time I will pay attention to the message title to be in line with other commits, as Diver pointed out.

PR #1563, opened 2017-08-06
PyGame: bump version

miqlas:


dev-python/pygame/python_pygame-1.9.3.recipe
```diff
@@ -0,0 +1,70 @@
...
+	"
+REQUIRES="
+	lib:libjpeg
+	lib:libpng16
+	lib:libsdl
+	lib:libsdl_gfx
+	lib:libsdl_image
+	lib:libsdl_mixer
+	lib:libsdl_ttf
+	lib:libsmpeg_0.4
```
korli: Doesn't it require python?

miqlas: Hmmm... could be :)

PR #1564, opened 2017-08-06
pv, new recipe

Begasus:


Begasus:
@miqlas ... thanks for the pointer ;)

Begasus:
didn't have the time to check on x86_64, but should be ok :)

PR #1577, opened 2017-08-12
openjpeg, bump version

Begasus:


media-libs/openjpeg/openjpeg-2.2.0.recipe
```diff
@@ -0,0 +1,89 @@
...
+	openjpeg${secondaryArchSuffix}_devel = $portVersion
+	devel:libopenjp2$secondaryArchSuffix = 7.0.2 compat >= 7
+	"
+REQUIRES_devel="
+	openjpeg$secondaryArchSuffix == $portVersion
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libpng$secondaryArchSuffix
```
korli: Should be devel:libpng16

Begasus: IIRC it wouldn't be needed for the devel package? I can change it, no prob (but I think I just left it as it was with the previous recipe) ...

korli: Yes, it's a leftover to fix. devel:libpng will be removed eventually.

Begasus: Ok, will check the build first with devel:libpng16 :)

korli:
I would leave openjpeg-2.1.2.recipe as is when it still works for x86_gcc2 though.

Begasus:
added v2.1.2 back into the PR

PR #1580, opened 2017-08-17
freeciv, bump version

Begasus:


korli:
Depends if the rest is also in the same did.

Begasus:
runConfigure ./configure * it's only Freeciv there ... "mv $binDir/freeciv-sdl $appsDir/Freeciv/Freeciv" ...

Begasus:
Better to have it as "$appsDir/Freeciv" imo, on that note ... do we need to keep pixmaps/application/appdata folders? (no valuable files in there for Haiku)

PR #1582, opened 2017-08-17
Haiku Radio, new recipe

Begasus:
First draft ...

### DIFF
--- a/haiku-apps/radio/radio-0.0.3_14.recipe
+++ b/haiku-apps/radio/radio-0.0.3_14.recipe
@@ -1,0 +1,45 @@
+SUMMARY="Find and listen to internet radio stations"
+DESCRIPTION="'Radio' helps you find and listen to internet radio streams."
+HOMEPAGE="https://github.com/fishpond-haiku/Haiku-Radio"
+LICENSE="MIT"
+COPYRIGHT="2017 Kai Niessen"
+REVISION="1"
+srcGitRev="26f3bbda29da165c849d246038af98e17270249c"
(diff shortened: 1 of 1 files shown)


humdingerb:
Can't comment inline with web+. My suggestions:

1. SUMMARY="Find and listen to internet radio stations"
2. DESCRIPTION="'Radio' helps you find and listen to internet radio streams."
3. Under PROVIDES: "radio = $portVersion" and "app:Radio = $portVersion"
3. Remove empty line 28
4. Put "makefile_engine" first in block
5. Add "addAppDeskbarSymlink"

Begasus:
Updated ...

humdingerb:
Looks good to me. Two questions:
1. When do we use "cmd:g++" instead of "cmd:gcc"? Is the distinction of interest?
2. Since in this case the author of the app is MIA, and the app hasn't been widely released before: How free are we to rename the app? "Radio" is so meh generic and also basically wrong... "StreamRadio", "NetRadio" or "NetStreamer" would be nicer.

Begasus:
1. I'm not the guy to ask here ;)
2. I'll hear it if there are any suggestions :)

Begasus:
#RadioStreamer# ? ;)

philxyz:
Name suggestion: Whisker

Begasus:
@humdingerb tried with both cmd:gcc and cmd:g++ and see no difference in building ...

humdingerb:
Thanks, I'll add the patch for the name change.

PR #1584, opened 2017-08-19
PHP: bump version

extrowerk:


dev-lang/php/patches/php-7.1.8.patchset
```diff
@@ -0,0 +1,481 @@
...
+ 
+     if test -z "$GD_WEBP_DIR"; then
+       AC_MSG_ERROR([webp/decode.h not found.])
+     fi
+ 
+-    for i in $PHP_WEBP_DIR /usr/local /usr; do
+-      test -f $i/include/webp/encode.h && GD_WEBP_DIR=$i && break
++    for i in $PHP_WEBP_DIR /usr/local /usr /system; do
++      test -f $i/develop/headers/webp/encode.h && GD_WEBP_DIR=$i && break
+     done
```
korli: You could also leave the loop as is, and add another check after the loop, this would be more acceptable upstream.
````
for i in $PHP_WEBP_DIR /usr/local /usr; do
   test -f $i/include/webp/encode.h && GD_WEBP_DIR=$i && break
done
test -f /system/develop/headers/webp/encode.h && GD_WEBP_DIR=/system
````

File: dev-lang/php/patches/php-7.1.8.patchset [hunk 2]
```diff
@@ -0,0 +1,481 @@
...
+ 
+     if test -z "$GD_WEBP_DIR"; then
+@@ -90,7 +90,7 @@ AC_DEFUN([PHP_GD_WEBP],[
+ 
+     PHP_CHECK_LIBRARY(webp,WebPGetInfo,
+     [
+-      PHP_ADD_INCLUDE($GD_WEBP_DIR/include)
++      PHP_ADD_INCLUDE($GD_WEBP_DIR/develop/headers)
```
korli: You could also leave PHP_ADD_INCLUDE($GD_WEBP_DIR/include) as is. No harm if it doesn't exist.

dev-lang/php/patches/php-7.1.8.patchset
```diff
@@ -0,0 +1,481 @@
...
+       PHP_ADD_LIBRARY(pthread)
+       PHP_ADD_LIBRARY_WITH_PATH(webp, $GD_WEBP_DIR/$PHP_LIBDIR, GD_SHARED_LIBADD)
```
korli: What is the value for PHP_LIBDIR? It should be develop/lib

dev-lang/php/php-7.1.8.recipe
```diff
@@ -0,0 +1,152 @@
...
+
+	runConfigure ./configure \
+		--with-libxml-dir=`finddir B_SYSTEM_DIRECTORY` \
+		--with-zlib-dir=`finddir B_SYSTEM_DIRECTORY` --with-zlib=shared \
+		--with-openssl \
+		--with-gd \
+		--with-pcre-dir=`finddir B_SYSTEM_DIRECTORY` \
+		--with-png-dir=`finddir B_SYSTEM_DIRECTORY` \
+		--with-jpeg-dir=`finddir B_SYSTEM_DIRECTORY` \
+		--with-webp-dir=`finddir B_SYSTEM_DIRECTORY` \
```
korli: in the m4 file we search explicitly /system, and here we define the webp directory, this is double work.

miqlas:
@korli: does it looks better to you? https://github.com/php/php-src/pull/2697/files

korli:
@miqlas like I already commented, each library check loop with /usr/ and /usr/local needless checks /system when it could easily so be done:
````
for i in $PHP_WEBP_DIR /usr/local /usr; do
   test -f $i/include/webp/encode.h && GD_WEBP_DIR=$i && break
done
test -f /system/develop/headers/webp/encode.h && GD_WEBP_DIR=/system
````

miqlas:
@korli : thanks for the tipp, i'll rework everything.
But i suppose i'll close this PR, because the PR what i did for php is in better shape.
I just made it to notice the other porters that i'm working on it.

PR #1587, opened 2017-08-19
Hermes: new recipe

miqlas:
build requires shouldn't needed, but they are.

### DIFF
--- a/games-arcade/hermes/hermes-1.02.recipe
+++ b/games-arcade/hermes/hermes-1.02.recipe
@@ -1,0 +1,60 @@
+SUMMARY="An oldschool Jump'n' Run game"
+DESCRIPTION="Hermes is an extraordinary Jump'n' Run game with plenty of \
+bad taste humour. If you feel offended by crude and dirty humour you may \
+run away now. The game also features pixel violence and promotes the \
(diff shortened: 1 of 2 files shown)


korli:
Versions for libraries should be hardcoded with the ones used to build the binaries. The ones present in haikuports might be incompatible.

miqlas:
@korli, i can do it, or i can provide a static binary. Wich one do you prefer?
Btw, have you got any idea why the devel req section required?

korli:
if you have built the binary yourself, you know which versions of hpkgs were used, then just declare them in the recipe, no need to rebuild a static binary.

haikuporter actually needs the libraries to make its policy check in the chroot:
https://github.com/haikuports/haikuporter/blob/master/HaikuPorter/Policy.py#L275
we could include the runtime dependencies when the build dependencies are empty.
haiku_devel isn't actually needed to build this recipe?

miqlas:
@korli : only the defined stuff is required, nothing else.
Ok, i'll define the correct versions then.
Thanks for the explanation and for the link.

korli:
You could do this: BUILD_REQUIRES=$REQUIRES

miqlas:
@korli: thanks, fixed.

PR #1589, opened 2017-08-20
zsh: Enable secure mem

miqlas:


korli:
Hey miqlas, you need to bump the revision :)

PR #1600, opened 2017-08-27
libraqm, new recipe, test still fails on most parts

Begasus:


Begasus:
Thanks for the merge, seems I overlooked some things when pushing :)

PR #1617, opened 2017-09-02
Audacious plugins fix

miqlas:
You tricky hacker....
https://www.youtube.com/watch?v=LmE3IhHBGjI


diversys:
--disable-mpris2 shouldn't be needed as it builds here just fine.
Audacious seems to not find its resources so the toolbar icons are missing and replaced with text. Could you take a look at that too? :)

diversys:
You can also enable ARCHITECTURES.

diversys:
Oh and --disable-dbus is needed for audacious recipe.

miqlas:
@diversys : it is... you know... actually... your recipe :D

PR #1628, opened 2017-09-09
luacheck, new recipe

Begasus:


Begasus:
Not sure if the secondaryArch is really needed, if not I can remove it :)

diversys:
If it's not a lib and not and add-on secondaryArch is not needed.

Begasus:
Updated :)

PR #1629, opened 2017-09-09
libdl, new recipe

Begasus:


waddlesplash:
Huh? Why do we need this? Can't we just patch anything that wants it?

Begasus:
It was one of the things I encountered when trying to build boswars ...

jessicah:
I agree with waddlesplash, this is not the solution. `dlopen` and friends are part of libroot. You need to modify boswars or whatever to not link to libdl on Haiku.

Begasus:
Ok, this can be closed then :)

PR #1638, opened 2017-09-15
mesa: switch to autoconf.

korli:
Mesa for Unix platforms is usually built with autoconf instead of scons. This switch should help us to follow build system changes for other platforms more easily, it provides a pkgconfig file for egl, library versioning, etc..
One hack is the use of libglapi for both libGL and libEGL, which is forbidden for the autoconf build system, but was allowed in scons. For the time being I added Haiku as an exception.
This commit is probably not so clean for upstream, I only tested libGL on x86_64. 
Please review.

### DIFF
--- a/sys-libs/mesa/mesa-17.1.7.recipe
+++ b/sys-libs/mesa/mesa-17.1.7.recipe
@@ -6,17 +6,18 @@ providing 3D rendering to Haiku applications."
 HOMEPAGE="http://www.mesa3d.org/"
 COPYRIGHT="1999-2017 Brian Paul"
 LICENSE="MIT"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://mesa.freedesktop.org/archive/mesa-${portVersion}.tar.xz"
 CHECKSUM_SHA256="69f472a874b1122404fa0bd13e2d6bf87eb3b9ad9c21d2f39872a96d83d9e5f5"
+PATCHES="mesa-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
 SECONDARY_ARCHITECTURES="?x86"
 
 PROVIDES="
 	mesa$secondaryArchSuffix = $portVersion
-	lib:libGL$secondaryArchSuffix
-	lib:libEGL$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix = 1.0.0 compat >= 1
+	lib:libEGL$secondaryArchSuffix = 1.0.0 compat >= 1
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


sys-libs/mesa/mesa-17.1.7.recipe
```diff
@@ -27,119 +28,106 @@ REQUIRES="
...
+	autoreconf -fi
+	runConfigure ./configure \
+		--disable-dri \
+		--disable-glx \
+		--disable-glx-tls \
+		--disable-gbm \
+		--disable-gles1 \
+		--disable-gles2 \
+		--disable-shared-glapi \
+		--with-gallium-drivers=swrast \
```
kallisti5: this doesn't look right.  swpipe ?

kallisti5: or maybe softpipe / llvmpipe

sys-libs/mesa/mesa-17.1.7.recipe
```diff
@@ -27,119 +28,106 @@ REQUIRES="
...
+	"
+REQUIRED_swpipe="
+	mesa$secondaryArchSuffix == $portVersion base
+	lib:libLLVM_4.0$secondaryArchSuffix
+	"
+
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	devel:libz$secondaryArchSuffix
+	llvm${secondaryArchSuffix}
```
kallisti5: This doesn't seem correct.  llvm is a static library used during linkage. I'm 90% sure we don't need llvm installed to leverage the llvmpipe renderer

sys-libs/mesa/mesa-17.1.7.recipe
```diff
@@ -27,119 +28,106 @@ REQUIRES="
...
 	"
 BUILD_PREREQUIRES="
-	llvm${secondaryArchSuffix}
```
kallisti5: see above about llvm being a static library linked in.

sys-libs/mesa/mesa-17.1.7.recipe
```diff
@@ -27,119 +28,106 @@ REQUIRES="
...
 	symlinkRelative -s $includeDir/EGL $includeDir/os/opengl/EGL
-
-	# KHR headers for EGL
-	mkdir -p $includeDir/KHR
-	cp ./include/KHR/* $includeDir/KHR
-
 	# Create KHR symlink in opengl kit
-	symlinkRelative -s $includeDir/EGL $includeDir/os/opengl/KHR
-
-	mkdir -p $developLibDir/pkgconfig
```
kallisti5: I don't remember doing these pkgconfig lines... are we sure fixPkgconfig properly corrects the pkg-config and the autotools install places it down properly?

sys-libs/mesa/mesa-17.1.7.recipe
```diff
@@ -149,24 +137,3 @@ EOF
 		"$addOnsDir/opengl/Software Pipe"
 }
 
-
```
kallisti5: Wait.. why are we dropping the renderer package? We want more of these going forward... not to remove them.

sys-libs/mesa/mesa-17.1.7.recipe
korli: yeah, I did't commit with swpipe. Will add it.

sys-libs/mesa/mesa-17.1.7.recipe
korli: llvm is now built with shared libs, there is a runtime dep. Here the line is just moved from BUILD_PREREQUIRES, because it's a library build dep, not a tool.

sys-libs/mesa/mesa-17.1.7.recipe
korli: see above.

sys-libs/mesa/mesa-17.1.7.recipe
korli: The hpkg is correct. There is already an egl pkgconfig file with autotools, we add the one for libGL.

sys-libs/mesa/mesa-17.1.7.recipe
korli: The definition is just moved before BUILD_REQUIRES, like for every other recipe.

korli:
Any comment on the patchset, in your opinion can this be upstreamed as is?

kallisti5:
So... not going to fix  --with-gallium-drivers=swpipe ?

korli:
This driver name doesn't actually exist in the configure script... It must be automatically enabled or something (in doubt check on Linux).

kallisti5:
weird. I tested this on my local machine and it worked well.
If you want to upstream this:

 1) make sure git-email is installed
 2) make sure git-email is configured in your .gitconfig
```
[sendemail]
	smtpencryption = tls
	smtpserver = ssl.mysmtp.com
	smtpuser = smtpuser
	smtpserverport = smtpport
```
  3) Checkout mesa
       git clone https://anongit.freedesktop.org/git/mesa/mesa.git
  4) Apply your changes / fix commit subject to "Haiku: Convert to autotools"
  5) Sign up for the mesa-dev mailing list. https://lists.freedesktop.org/mailman/listinfo/mesa-dev
  6) git send-email --to mesa-dev@lists.freedesktop.org origin

korli:
Hi Alex, thanks for the tips.
I've sent a patch against trunk, minus the change in src/mapi/glapi/glapi_dispatch.c, and the shared-glapi in configure.ac.
Since 17.2.x haiku-softpipe/GalliumContext.cpp doesn't build anymore, so I couldn't fully check the build. Your insight would definitely help :)

PR #1645, opened 2017-09-17
assimp, new recipe, wip

Begasus:
Needs to be checked on x86_64

### DIFF
--- a/media-libs/assimp/assimp-4.0.1.recipe
+++ b/media-libs/assimp/assimp-4.0.1.recipe
@@ -1,0 +1,73 @@
+SUMMARY="Importer library to import assets from 3D files"
+DESCRIPTION="Open Asset Import Library (short name: Assimp) is a portable \
+Open Source library to import various well-known 3D model formats in a \
+uniform manner. The most recent version also knows how to export 3d files \
+and is therefore suitable as a general-purpose 3D model converter."
+HOMEPAGE="https://github.com/assimp/assimp"
+COPYRIGHT="2006-2016, assimp team"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/assimp/assimp/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256='60080d8ab4daaab309f65b3cffd99f19eb1af8d05623fff469b9b652818e286e'
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	assimp$secondaryArchSuffix = $portVersion
+	cmd:assimp$secondaryArchSuffix
+	lib:libassimp$secondaryArchSuffix = 4.0.1 compat >= 4
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libglu$secondaryArchSuffix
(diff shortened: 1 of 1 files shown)


miqlas:
Closes #1645

media-libs/assimp/assimp-4.0.1.recipe
```diff
@@ -0,0 +1,75 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	mkdir -p build
+	cd build
+	cmake -DASSIMP_BIN_INSTALL_DIR=$binDir \
+		-DASSIMP_LIB_INSTALL_DIR=$libDir \
+		-DASSIMP_INCLUDE_INSTALL_DIR=$includeDir ..
```
korli: These INSTALL_DIR could theorically be replaced by $cmakeDirArgs, provided that https://github.com/assimp/assimp/blob/master/CMakeLists.txt#L186 is extended: if( UNIX or HAIKU )

media-libs/assimp/assimp-4.0.1.recipe
```diff
@@ -0,0 +1,75 @@
...
+REQUIRES_devel="
+	assimp$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libGL$secondaryArchSuffix
+	devel:libglu$secondaryArchSuffix
+	devel:libminizip$secondaryArchSuffix
+	devel:libIL$secondaryArchSuffix
```
korli: libIL is a dependency for assimp_qt_viewer. This shouldn't be needed.

Begasus: removed DevIL from the recipe ...

media-libs/assimp/assimp-4.0.1.recipe
Begasus: Have to look into this how to use it ;)

korli: That would be: cmake $cmakeDirArgs ..

provided that https://github.com/assimp/assimp/blob/master/CMakeLists.txt#L186 is extended: if( UNIX or HAIKU )

Begasus: I get the part of " cmake $cmakeDirArgs .. ", just not sure how to extend the correct way of extending ( UNIX or HAIKU ) ...

Begasus: PS, not too familiar with cmake myself ...

miqlas: @korli: arent't we UNIX for cmake already?

korli: Miqlas right, wasn't aware of it. Then it should just work.

Begasus: not using the INSTALL_DIR options fails on x86_gcc2 for secondaryArchSuffix ...

korli: Begasus, fine, the packages don't always correctly use the first provided by cmakeDirArgs. Ok to merge?

Begasus: Aside from removing DevIL from the recipe it builds ok here, miqlas tested the build on x86_64 :)

PR #1647, opened 2017-09-17
OIIO: bump version

extrowerk:


media-libs/openimageio/openimageio-1.7.17.recipe
```diff
@@ -0,0 +1,112 @@
...
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	export CXXFLAGS="-std=c++11 -DBOOST_NO_CXX11_CONSTEXPR"
+	make USE_OPENSSL=1 $jobArgs
```
korli: could we use "cmake . -DUSE_OPENSSL=ON $cmakeDirArgs" plus "make install" instead?

miqlas:
Broken in boost. 
"unique_ptr.hpp: error: constexpr constructor does not haave empty body."

I don't know what's up wit it.

extrowerk:
Builds and packages fine on 64 bit.

PR #1662, opened 2017-09-23
Trojita reworked receipe and corrected rdef

dasouth1:
Reworked the recipe for Trojita email client. Corrected rdef as it has an error that was preventing it from building.
[Trojita.zip](https://github.com/haikuports/haikuports/files/1326590/Trojita.zip)


miqlas:
Hi @dasouth1 , sorry, but your PR is empty.

PR #1663, opened 2017-09-23
pingus, add missing libz

Begasus:


korli:
@begasus I had pingus correctly built on x86_64, it failed for some reason on the buildbot...

PR #1700, opened 2017-10-07
zziplib, fix recipe

Begasus:
Not tested on x86_64 (yet)


Begasus:
checked and builds ok on all arch's

PR #1702, opened 2017-10-08
zopfli, minor adjustment

Begasus:


diversys:
No need to bump revision in this case

PR #1712, opened 2017-10-16
ffmpeg, bump version

Begasus:
not tested on x86_64 yet, probably would be good if some people tested it before merge ...


korli:
Looks like the versions in the PROVIDES block aren't updated.

Begasus:
will look into it :)

PR #1716, opened 2017-10-18
Fixing yaml_cpp checksum

robcsi:
I was building yaml_cpp for qutebrowser, and it wouldn't build because of the incorrect checksum in the recipe.


scottmc:
Strange that the checksum would need to be fixed.  @KapiX can you verify this one?  Also if this is the case, then the REVISION will need to be incremented as well.

KapiX:
I will later today.

korli:
No need to increment the revision, the resulting package is the same.

PR #1724, opened 2017-10-20
unarr: use upstream source

### LINKED ISSUE #1723: app-arch/unarr: obsolete version, please use upstream as source
The port currently uses a 2 year old version of unarr that has known bugs. This isn't unarr 1.0, it's a pre-1.0 snapshot (1.0~[number of git commit].[git shorthash], the tilde after the 1.0 indicates the version is lower than 1.0 in Debian version syntax).  This version has at least one major bug (no largefile support). For me this wasn't an issue as I simply patched that bug in the Debian package. For you it probably is as that patch isn't included in the source you're using.

So, please update your port ;) And don't use my openSuse buildservice repo as source, use the proper unarr 1.0.0 upstream release. It's available [here.](https://github.com/selmf/unarr/releases)

phoudoin:
Previous source was obsolete, thanks selmf, actual maintener, to notify us.
Enabled bz2, xz and zip support, too.
This should fix #1723.


app-arch/unarr/unarr-1.0.0.recipe
```diff
@@ -8,58 +8,84 @@ HOMEPAGE="https://github.com/sumatrapdfreader/sumatrapdf/tree/master/ext/unarr"
...
-SOURCE_DIR="libunarr"
+REVISION="3"
+SOURCE_URI="https://github.com/selmf/unarr/archive/v${portVersion}.tar.gz"
+CHECKSUM_SHA256="ef1610d636641751ca234da7998b3f6b53313c7e526492a6070dd572b99d4685"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
-	unarr$secondaryArchSuffix = $portVersion
```
korli: This provides is actually required.

app-arch/unarr/unarr-1.0.0.recipe
```diff
@@ -8,58 +8,84 @@ HOMEPAGE="https://github.com/sumatrapdfreader/sumatrapdf/tree/master/ext/unarr"
...
+	lib:libz$secondaryArchSuffix
 	"
 
 PROVIDES_devel="
 	unarr${secondaryArchSuffix}_devel = $portVersion
 	devel:libunarr$secondaryArchSuffix = $portVersion
 	"
+
 REQUIRES_devel="
-	unarr$secondaryArchSuffix == $portVersion base
```
korli: This is also required.

phoudoin: oops, didin't see these lines where lost.

btw, how to only produce a _devel package? Cause now the main package have not content at all, while the _devel one have the static library, headers and pkgconfig file... ?
Maybe I could build the shared library too, though.

korli: Usually only the main package contains everything, no devel package. Package name unarr, provides devel:linunarr.

phoudoin: Fix broken recipe, build the library both in static and shared versions and move everything into main package, no more devel package.

korli:
Question: now that you enabled the shared library, why not build a devel package (like for every Haikuports library) ?

phoudoin:
Would make sense indeed, dunno why I didn't though about it.

PR #1736, opened 2017-10-24
unarr: use upstream source

### LINKED ISSUE #1723: app-arch/unarr: obsolete version, please use upstream as source

phoudoin:
As notified by selmf (thanks!), actual maintener, we were using obsolete source.
Enabled bz2, xz and zip support too.
Add PkgConfig support.
And now both static and shared libraries are availables.

This should fix #1723.


phoudoin:
Kinda corrupt my previous branch, have to restart from a clean state.

PR #1758, opened 2017-10-30
Trojita: bump version

miqlas:


mail-client/trojita/trojita-0.7.0~git.recipe
```diff
@@ -32,7 +34,10 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku${secondaryArchSuffix}
+#	cmd:gpgconf$secondaryArchSuffix
 	lib:libcrypto$secondaryArchSuffix
+#	lib:libgpgmepp$secondaryArchSuffix
+#	lib:libgpg_error$secondaryArchSuffix
```
korli: Could you clarify in the commit message why it is needed to disable gpgme, etc?

miqlas: Because enabling GPGME means QGPGME also needs to be installed and afaik we got no QGPGME.

miqlas:
I think it is ok to merge now.

PR #1766, opened 2017-10-31
GLEW: bump version, recipe cleanup

miqlas:
@korli: does it looks good to you, or should i make a "glew2" recipe? A lib is missing from the new version.


waddlesplash:
Does software that uses GLEW work with a recompile (i.e. is it source compatible?) If not, then yes it should be glew2.

khallebal:
@miqlas  you're missing libglewmx from provides, both devel and non devel.

korli:
@khallebal it looks OK to me, version 2.0 doesn't provide libglewmx.

khallebal:
Oops, sorry did'nt know that about the new 2.0 version

PR #1784, opened 2017-11-11
Fish : bugfix

extrowerk:
Disabled 2.6.0 on every arch as broken.
Enabled 2.6b1 on every arch
Mimedb not exist anymore since 2015


korli:
Is 2.6b1 newer than 2.6.0?

extrowerk:
@korli : nope, but unlike 2.6 this doesn't crashing.

extrowerk:
Can't get this sht work.

PR #1785, opened 2017-11-12
symbola, fix checksum

Begasus:


korli:
We should switch to a mirror with a stable URL. for instance http://downloads.sourceforge.net/slackbuildsdirectlinks/symbola-font-ttf-10.03.zip

Begasus:
Will check later :)

Begasus:
Updated ... :)

PR #1793, opened 2017-11-17
openssl: allow co-installation of 1.0.2 and 1.1.0.

fbrosson:
* ~~Move `cmd:` items to a new `_bin` sub-package.~~
* ~~Base packages openssl11 and openssl now only ship the libs.~~
* ~~Make `openssl11_{bin,devel}` conflict with `openssl_{bin,devel}`.~~
* ~~Skip installation of `lib{ssl,crypto}.so` symlinks in openssl11.~~

The changes detailed below allow to co-install the base packages of 1.0.2 and 1.1.0 (as well as their debuginfo). The _devel (or _man) packages cannot be co-installed.

* Define a debuginfo package for openssl11.
* Fix typo in call to defineDebugInfoPackage for openssl which was producing an incorrect debuginfo on 2nd arch.
* Also add `cmd:openssl{,-1.1}` to their debuginfo packages.
* Move `$docDir/html/man3` to openssl11_devel.
* Add a **`-1.1`** suffix to each `cmd:{openssl,c_rehash}` for openssl11 to allow the co-installation of 1.0.2 & 1.1.
* Create new `_man` sub-packages for openssl and openssl11 and move all man pages (other than the man3 section) there to avoid overlaps when the 2 base packages are installed.
* Add symetric (openssl vs openssl11) `CONFLICTS` for their `_devel` and `_man` sup-packages. ~~Same thing for `REPLACES`.~~
* openssl{,11}_man REQUIRES and SUPPLEMENTS openssl{,11}.

* Introduce 4 custom variables (after portVersion):

portVersion | 1.0.2* | 1.1.0*
--------------- | ------- | --------
libVersion    | 1.0.0  | 1.1
cmdSuffix    |          | -1.1
mySuffix      |          | 11
altSuffix        | 11     |

These are used inside both recipes at the same places to make it easier to spot the real differences between 1.0.2 & 1.1.0.

### DIFF
--- a/dev-libs/openssl/openssl11-1.1.0f.recipe
+++ b/dev-libs/openssl/openssl11-1.1.0f.recipe
@@ -15,24 +15,29 @@ HOMEPAGE="https://www.openssl.org/"
 COPYRIGHT="1995-1998 Eric Young
 	1998-2017 The OpenSSL Project."
 LICENSE="OpenSSL"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://www.openssl.org/source/openssl-$portVersion.tar.gz"
 CHECKSUM_SHA256="12f746f3f2493b2f39da7ecf63d7ee19c6ac9ec6a4fcd8c229da8a522cb12765"
 SOURCE_DIR="openssl-$portVersion"
 PATCHES="openssl-$portVersion.patchset"
 
 ARCHITECTURES="?x86_gcc2 ?x86 ?x86_64 ?arm ?ppc"
-SECONDARY_ARCHITECTURES="?x86_gcc2 ?x86 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="?x86_gcc2 ?x86"
+
+libVersion=1.1
+cmdSuffix="-1.1"
+mySuffix=11
+altSuffix=
 
 PROVIDES="
-	openssl11$secondaryArchSuffix = $portVersion compat >= 1.1.0
-	lib:libcrypto$secondaryArchSuffix = 1.1 compat >= 1.1
-	lib:libssl$secondaryArchSuffix = 1.1 compat >= 1.1
+	openssl$mySuffix$secondaryArchSuffix = $portVersion compat >= 1.1.0
+	lib:libcrypto$secondaryArchSuffix = $libVersion compat >= $libVersion
+	lib:libssl$secondaryArchSuffix = $libVersion compat >= $libVersion
 	"
 if [ -z "$secondaryArchSuffix" ]; then
 	PROVIDES="$PROVIDES
-		cmd:c_rehash = $portVersion compat >= 1
-		cmd:openssl = $portVersion compat >= 1
+		cmd:c_rehash${cmdSuffix/-/_} = $portVersion compat >= 1
+		cmd:openssl${cmdSuffix/-/_} = $portVersion compat >= 1
 		"
 fi
 
@@ -42,13 +47,33 @@ REQUIRES="
 	ca_root_certificates
 	"
 
+if [ -z "$secondaryArchSuffix" ]; then
+	SUMMARY_man="Manual pages for openssl $portVersion"
+	ARCHITECTURES_man="any"
+	PROVIDES_man="
+		openssl${mySuffix}_man = $portVersion
+		"
+	REQUIRES_man="
+		openssl$mySuffix == $portVersion
+		"
+	SUPPLEMENTS_man="
+		openssl$mySuffix == $portVersion
+		"
+	CONFLICTS_man="
+		openssl${altSuffix}_man
+		"
+fi
+
 PROVIDES_devel="
-	openssl11${secondaryArchSuffix}_devel = $portVersion
-	devel:libcrypto${secondaryArchSuffix} = 1.1 compat >= 1.1
-	devel:libssl${secondaryArchSuffix} = 1.1 compat >= 1.1
+	openssl$mySuffix${secondaryArchSuffix}_devel = $portVersion
+	devel:libcrypto$secondaryArchSuffix = $libVersion compat >= $libVersion
+	devel:libssl$secondaryArchSuffix = $libVersion compat >= $libVersion
 	"
 REQUIRES_devel="
-	openssl11${secondaryArchSuffix} == $portVersion base
+	openssl$mySuffix$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_devel="
+	openssl$altSuffix${secondaryArchSuffix}_devel
 	"
 
 BUILD_REQUIRES="
@@ -56,19 +81,33 @@ BUILD_REQUIRES="
 	"
 BUILD_PREREQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	cmd:gcc${secondaryArchSuffix}
-	cmd:ld${secondaryArchSuffix}
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
 	cmd:make
 	cmd:perl >= 5
 	cmd:sed
-	cmd:awk
 	"
+
+if [ -n "$secondaryArchSuffix" ]; then
+	maybe_binDir_runtimes=
+else
+	maybe_binDir_runtimes="
+		$binDir/openssl$cmdSuffix \
+		"
+fi
+defineDebugInfoPackage openssl$mySuffix$secondaryArchSuffix \
+	$libDir/engines-1.1/capi.so \
+	$libDir/engines-1.1/padlock.so \
+	$libDir/libcrypto.so.$libVersion \
+	$libDir/libssl.so.$libVersion \
+	$maybe_binDir_runtimes
 
 BUILD()
 {
 	./config --prefix=$prefix --libdir=$relativeLibDir \
 		--openssldir=$dataRootDir/ssl \
-		zlib shared no-asm
+		zlib shared no-asm -g
 	make $jobArgs
 }
 
@@ -84,18 +123,33 @@ INSTALL()
 	prepareInstalledDevelLibs libcrypto libssl
 	fixPkgconfig
 
+	if [ -n "$secondaryArchSuffix" ]; then
+		maybe_docDir_html_man3=
+		maybe_manDir_man3=
+	else
+		maybe_docDir_html_man3=$docDir/html/man3
+		maybe_manDir_man3=$manDir/man3
+	fi
+
 	# devel package
 	packageEntries devel \
 		$developDir \
-		$docDir \
-		$manDir/man3
+		$maybe_docDir_html_man3 \
+		$maybe_manDir_man3
 
-	# Remove stuff we don't need in the secondary architecture base package,
-	# since we make it depend on the primary package.
+	# Remove stuff we don't need in the secondary architecture base package
 	if [ -n "$secondaryArchSuffix" ]; then
 		rm -rf $prefix/bin
 		rm -rf $dataRootDir/ssl
 		rm -rf $documentationDir
+	else
+		if [ -n "$cmdSuffix" ]; then
+			mv $binDir/openssl $binDir/openssl$cmdSuffix
+			mv $binDir/c_rehash $binDir/c_rehash$cmdSuffix
+		fi
+		# man package
+		packageEntries man \
+			$manDir
 	fi
 }
 
--- a/dev-libs/openssl/openssl-1.0.2m.recipe
+++ b/dev-libs/openssl/openssl-1.0.2m.recipe
@@ -15,7 +15,7 @@ HOMEPAGE="https://www.openssl.org/"
 COPYRIGHT="1995-1998 Eric Young
 	1998-2017 The OpenSSL Project."
 LICENSE="OpenSSL"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://www.openssl.org/source/openssl-$portVersion.tar.gz"
 CHECKSUM_SHA256="8c6ff15ec6b319b50788f42c7abc2890c08ba5a1cdcd3810eb9092deada37b0f"
 PATCHES="openssl-$portVersion.patchset"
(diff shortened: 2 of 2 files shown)


fbrosson:
Should we provide symlinks (`libssl.so.1.0.0 -> libssl.so.1.0.2`
and `libcrypto.so.1.0.0 -> libcrypto.so.1.0.2`) to simplify the transition?
(If we cannot provide such symlinks then we need to bump the revisions of all packages which need libssl and/or libcrypto.)
In case it is acceptable to provide such symlinks, should we do so in a separate "openssl" wrapper package? (If yes, I'll remove the REPLACES, of course.)

If anyone knows how to instruct "make install" to put the engines in `lib/openssl-1.0.2/engines/` instead of `lib/engines/` then I can fix the recipe.

I changed some compats a little. Are they OK?

waddlesplash:
@fbrosson is back! 🎊 🥂

Does this properly handle the devel packages too, or will they conflict (it doesn't look like you altered include, etc. paths?)

>Should we provide symlinks

I vote yes (I can't see any downside)?

fbrosson:
> @fbrosson is back! 🎊 🥂

Thanks!

Yep, I added a CONFLICTS for the devel package.
I did not even try to make them co-installable because I'm quite sure that multiple devel packages can co-exist in PACKAGES_PATH.
(So, if a package needs a specific version of openssl, haikuporter will activate the right devel package in the build environment.)

For the symlinks, I did not try anything yet. Actually, I wonder if HP would accept symlinks for `lib:` entries.
I also don't see any downsides, but I'm not yet sure there aren't any.

Shall we keep this PR open for a few days (or more) to figure out what is the best upgrade path for the openssl package and those which depend on it?

dev-libs/openssl/openssl102-1.0.2m.recipe
```diff
@@ -18,36 +18,53 @@ LICENSE="OpenSSL"
...
+SOURCE_DIR="openssl-$portVersion"
 PATCHES="openssl-$portVersion.patchset"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64 ?arm ?ppc"
 SECONDARY_ARCHITECTURES="x86_gcc2 x86"
 
 PROVIDES="
-	openssl$secondaryArchSuffix = $portVersion
-	lib:libcrypto$secondaryArchSuffix = 1.0.0 compat >= 1.0.0
-	lib:libssl$secondaryArchSuffix = 1.0.0 compat >= 1.0.0
```
korli: What's the point of changing the recipe name? The lib version?

fbrosson: Well, if we put 1.0.2 in openssl102 and 1.1 in openssl11 then we have a consistent naming convention and we can even add a wrapper package ("openssl") which would depend on the recommended openssl version, i.e. openssl102 for the moment, and then openssl11 once we have a working package for 1.1.
BTW, it might even be a good idea to add the symlinks (discussed above) in the openssl package.
I'm quite sure this will even simplify the transition of the packages which depend on libssl and/or libcrypto. Later, when all of these packages get bumped, I guess we could even drop the symlinks and only keep an enpty openssl package with just a dependency on our recommened/default openssl.

fbrosson: @korli, you are right. There's probably no need to change the package name. We can keep openssl for 1.0.2.
We could then add 1.1.0 in openssl11, keeping 1.0.2 in openssl.
Later we could rename openssl to openssl102 and openssl11 to openssl.

korli: openssl11 already exists AFAICT. Whether openssl 1.1 and openssl 1.0 can coexist in the runtime is a good question. Typically what happens when two libraries in one application each use one different version of openssl?

fbrosson: My mistake. I know openssl11 already exists. I should have written "enable" openssl11 instead of "add" openssl11.
BTW, I forgot to say that we would/will need to change openssl11 a little to have it provide `libssl.so.1.1.0` and `libcrypto.so.1.1.0` and also move `lib/engines/` to `lib/openssl-1.1/engines/`.

Hmm, I've just noticed that Debian uses `$libDir/openssl-1.0.2/engines/` for 1.0.2 and `$libDir/engines-1.1/` for 1.1.
So I think it would be nicer to use `$libDir/engines-1.0.2/` for 1.0.2 and `$libDir/engines-1.1/` for 1.1.

I'll do some tests and check whether or not it is possible to install openssl 1.0.2 (with the new layout and with the symlinks) without having to uninstall all the packages which depend on libssl and/or libcrypto.
I'll also need to check that building packages against 1.0.2 (with the new layout) only adds dependencies to `libssl.so.1.0.2` and `libcrypto.so.1.0.2` (and not to `libssl.so.1.0.0` and `libcrypto.so.1.0.0`, which will be symlinks).

fbrosson: > Typically what happens when two libraries in one application each use one different version of openssl?

I believe this would be OK because if two (or more) libs (required by an app) depend on the libs of different versions of openssl, then this means they use these openssl libs in different contexts.
All we need to do is make sure the base packages for 1.0.2 and 1.1.0 only ship the libs in openssl (and not the `cmd:openssl` runtime), and that they ship them with distinct sonames. That's what Debian does.

For anyone else catching this thread:
openssl11 would ship `lib{ssl,crypto}.so.1.1.0` and openssl102 (or openssl if we keep the package name unchanged) would ship `lib{ssl,crypto}.so.1.0.2`.
openssl11_bin would ship `cmd:openssl = 1.1.0` and openssl102_bin (or openssl_bin) would ship `cmd:openssl = 1.0.2`. So each user would be able to install one (and only one) cmd:openssl from the corresponding _bin sub-package.

fbrosson:
> Regarding this PR, it would be preferred to let the openssl 1.0.2 as is (please don't change a SONAME for no reason).

Excellent idea. Done.

> The openssl 1.1.0 recipe can rename its tools with a suffix, place its engines in another path, the devel package conflicts with the one of the 1.0 recipe.

I think it is better to keep the same names for the `cmd:` items and ship them in a `_bin` sub-package.

I still need to get rid of at least one of these two pairs of symlinks:
* the `lib{ssl,crypto}.so -> lib{ssl,crypto}.so.1.1` symlinks in openssl11
* the `lib{ssl,crypto}.so -> lib{ssl,crypto}.so.1.0.0` symlinks in openssl

Should I keep the symlinks in openssl and remove those in openssl11?

mmlr:
I just want to reference homebrew, which has apparently given up on a parallel migration to OpenSSL 1.1, instead waiting for everything to support the newer version and then switch in one large change/rebuild:

https://github.com/Homebrew/homebrew-core/issues/4543

Having a conflicts on the devel package may work, but I fear it's not that simple. If you look at the buildlog here:

https://kitchen.haiku-os.org/master/x86_gcc2/logviewer.html?buildruns/417/builds/26170.log

You can see that both openssl_x86_devel and openssl11_x86_devel were selected. I have not investigated what pulls in either package, but that both are there suggests that there are at least two direct or indirect dependencies causing them to be pulled in. With the declared conflicts, such builds would simple not be runable anymore.

The libpng devel packages for the plain (1.5) and 16 (1.6) versions can get by with providing separate include dirs, which may work here too. They still have some overlapping files though, critically both provide the libpng.pc.

Overlaps are very dangerous, because the package build time will influence what files will be visible. This will lead to very hard to track bugs that aren't easily reproduced. You also can't expect the newer version to always be the one having newer timestamps because point/maintenance releases for the respective older port may cause newer packages to be built for the older version. In the libpng case above, ports depending on libpng that use pkg-config will most likely find and use a more or less random version of libpng if both devel packages are pull in.

mmlr:
I've checked the built packages and you can see that curl_x86_devel pulls in devel:libssl_x86>=1.1 because it happend to be built against openssl11:

https://eu.hpkg.haiku-os.org/haikuports/master/repository/x86_gcc2/references_libssl.txt

I've made the same list for libcypto, but there it doesn't look like any devel:libcypto was pulled in from openssl11:

https://eu.hpkg.haiku-os.org/haikuports/master/repository/x86_gcc2/references_libcrypto.txt

Now possibly curl_x86_devel should not actually require the openssl lib, or rather only require it for the build?

fbrosson:
Thanks @mmlr for your feedback.

I've marked openssl11 as untested, as you have already noticed.

If the changes in this PR do not create new issues, then I think it could be worth merging it (after more testing and once everyone agrees) because it will make it easier (for those users who have `ALLOW_UNTESTED=yes`) to play with openssl 1.1 without having to drop openssl 1.0.2.

BTW, I think haikuporter should not activate conflicting packages.
I don't know, however, if making such a change would not break other things...

fbrosson:
> Now possibly curl_x86_devel should not actually require the openssl lib, or rather only require it for the build?

That's a good question. It would be interesting to see how many packages (among those which need `lib:libcurl`) would fail to build if we removed `devel:libssl` and/or `devel:libz` from curl's `REQUIRES_devel` (but keeping it in curl's `BUILD_REQUIRES`, of course.)
If only a few fail to build, then it _might_ be a good idea to add `devel:libssl` and/or `devel:libz` to their own `BUILD_REQUIRES`.
But maybe that would be just noise...

fbrosson:
> Regarding this PR, it would be preferred to let the openssl 1.0.2 as is (please don't change a SONAME for no reason). The openssl 1.1.0 recipe can rename its tools with a suffix, place its engines in another path, the devel package conflicts with the one of the 1.0 recipe.

@korli, I made the changes you suggested (plus a few others) and I have to admit that you were perfectly right about not changing the package for openssl 1.0.2.
The only tiny change I made to openssl was to move the man pages to a new _man subpackage (except for the man3 section, which goes to the devel package, of course) because it would not be possible to co-install 1.0.2 and 1.1.0 if we had kept the man pages in the base package, as we know.

Regarding openssl11, I renamed the 2 runtimes adding a suffix as you suggested.

I also added a debuginfo for 1.1.0 (on primary arch) and removed the one on 1.0.2 secondary arch because we still have https://github.com/haikuports/haikuporter/issues/85.

Now, the other changes I made are neutral to both versions of openssl. It's about the 4 custom variables.
I know we should avoid adding custom variables, but I think it would be worth having them on this particular package. But if you think the recipes would be better without them, that's OK. I can get rid of the custom variables and make the changes to obtain the same result.

I also added a SUPPLEMENT to the _man sub-packages. If this is not welcome (or appropriate) then I can drop that too.
BTW, I used "_man" but I can use any other name if you think _man is not the best choice.
I can rename _man to _doc (and put the html doc with the man pages) if you prefer.
Just let me know what changes you would like to have, in case anything is not ok.

I already tested everything and the only issue I found was the debuginfo on secondary arch, which was not providing the correct package name.

Thanks!

fbrosson:
> I also added a debuginfo for 1.1.0 (on primary arch) and removed the one on 1.0.2 secondary arch because we still have haikuports/haikuporter#85.

There was a typo in the call to defineDebugInfoPackage (from the previous commit):
```diff
-defineDebugInfoPackage openssl$secondaryArchSufix
+defineDebugInfoPackage openssl$secondaryArchSuffix
```
which I had not noticed, and which made me think that issue 85 was not fixed.
It turns out that it got fixed by @KapiX when he fixed issue 86.
So this is great news! Thanks KapiX!

> I already tested everything and the only issue I found was the debuginfo on secondary arch, which was not providing the correct package name.

I've updated my commit and now we should have correct debuginfo packages for openssl on secondary arch too.

fbrosson:
Now everything is OK. I'm 100% sure about that.
Shall I merge?

waddlesplash:
Sounds good to me.

dev-libs/openssl/openssl11-1.1.0f.recipe
```diff
@@ -84,18 +115,33 @@ INSTALL()
...
-	# since we make it depend on the primary package.
+	# Remove stuff we don't need in the secondary architecture base package
 	if [ -n "$secondaryArchSuffix" ]; then
 		rm -rf $prefix/bin
 		rm -rf $dataRootDir/ssl
 		rm -rf $documentationDir
+	else
+		if [ -n "$cmdSuffix" ]; then
+			mv $binDir/openssl $binDir/openssl-$libVersion
+			mv $binDir/c_rehash $binDir/c_rehash-$libVersion
```
korli: Should be: mv $binDir/c_rehash $binDir/c_rehash$cmdSuffix or something with -
Instead of another subpackage, please also rename the man pages and package them directly in the package containing the commands.

fbrosson: Renaming the man pages will be tricky, but I'll try.
Here's a small update to fix the first issue you pointed.

fbrosson: I fear that renaming the man pages will be a nightmare :-(
For 1.0.2 there are currently 100 files in documentation/man/man1.
Among these 100 files we have:
* openssl.1
* 42 files which start with openssl-
* 56 files which do not start with openssl

To keep the man pages in the base package, we could create a 1ssl section for the man pages of 1.0.2 and a 1ssl11 section for those from 1.1.x.
But having extra man sections might also be quite confusing.

Given that openssl 1.1.x has its html doc in the base package, would you agree with keeping the man pages in their sub-package ?
Most users who might be interested in reading openssl's man pages will probably know how to install openssl_man or openssl11_man.
Hmm, maybe we should add a "RECOMMENDS" keyword to haikuporter ;-)
Unless HaikuDepot already makes use of the SUPPLEMENTS (which I have already added.)

Hmm, do you think that "moving" (and renaming, for 1.1.x) to the base package *only* the 43 files (which start with openssl) would be a good idea? Oops, it will also be confusing...

pulkomandy: The general policy is that it is not possible to install _dev version of packages for two different versions at the same time (there can be exceptions). I would say the same could apply to the manpages.

We could even provide only the most up to date version of the docs, and do not package docs at all for older OpenSSL versions. I think their documentation do mentions the version when a feature was introduced, so this should not be a problem even for people still using the old version (they can even see what they are missing).

fbrosson: Yep, I (already) declared the CONFLICTS for both _devel and _man sub-packages, so we are safe here.
1.0.2 does not provide html docs. (I tried to have it produce these docs, but I suspect that it was not easily available on 1.0.2.)
I think the current commit is the best compromise we can have.

BTW, I'm just added `cmd:openssl{,-1.1}` to the debuginfo package.

fbrosson: > We could even provide only the most up to date version of the docs, and do not package docs at all for older OpenSSL versions.

If we choose to only ship the docs from 1.1.x, this means we cannot put it add it to the base package of 1.0.2 (unless we add the 1.1.x tarball to the 1.0.2 recipe, but that would not be a good idea.)
So we would need to ship the 1.1.x docs outside of 1.0.2.
Now, if we put the docs for 1.1.x in the base package for 1.1.x, then anyone willing to have the runtimes for 1.0.2 would be forced to install the 1.1.x base package.

So the best split is to have the 1.1.x doc in a _man sub-package for 1.1.x.
And since 1.1.x is still marked as untested, we also need to ship the docs for 1.0.2. The best place for that is in a _man sub-package for 1.0.2.

The current commit in this PR does exactly this ;-)

BTW, the tests work fine for 1.0.2. But they are not working for 1.1.x. When I launch the tests for 1.1.0 it does a few things than it hangs, doing nothing. There is probably a process waiting (forever) for standard input. I did not dig it.
BTW, we'll soon have mbedtls, which passes all tests :tada:

korli: I would rename the man files anyway, while updating the links between pages. Not a big deal for a sed expert :)
For instance openssl.1 => openssl_11.1, then update "L<openssl(1)>" => "L<openssl_11(1)>"

fbrosson: We could to that, but what would you do with the 56 files which do not even contain "openssl" in their names?
I think it will be much easier for the users to install the _man sub-package they want, and then use the same names as in all other OSes.
Those users who would be lost with the installion of the _man sub-package would not even use the man command. They would use Google ;-)

fbrosson: Shall we leave the man pages in their own sub-packages, without worrying about renaming any man page?
Reminder: WIth 1.1.x the html docs are *already* in the base package ;-)

PR #1812, opened 2017-11-29
Recipe for primegen 0.97

ohnx:
primegen is a small, fast library to generate prime numbers in order.

The recipe added primegen to haikuports at `haikuports/sci-libs/primegen/`.

I intentionally chose to create a git repository since there are a large number of patches required... I hope this is OK.

Please let me know if anything is wrong!

### DIFF
--- a/sci-libs/primegen/primegen-0.97.recipe
+++ b/sci-libs/primegen/primegen-0.97.recipe
@@ -1,0 +1,75 @@
+SUMMARY="Fast C/C++ prime number generator library"
+DESCRIPTION="primegen is a small, fast library to generate prime numbers \
+in order. It generates the 50847534 primes up to 1000000000 in just 8 \
+seconds on a Pentium II-350; it prints them in decimal in just 35 seconds. \
+primegen can generate primes up to 1000000000000000, although it is not \
+optimized for primes past 32 bits. It uses the Sieve of Atkin instead of the \
+traditional Sieve of Eratosthenes."
+HOMEPAGE="https://cr.yp.to/primegen.html"
+COPYRIGHT="1999, D. J. Bernstein"
+LICENSE="Public Domain"
+REVISION="1"
+SOURCE_URI="https://cr.yp.to/primegen/primegen-$portVersion.tar.gz"
+CHECKSUM_SHA256="54285baf8eed9e421ff2220a2112d38cfb20c1ebef6014ef3f0004c22c95f40d"
+PATCHES="primegen-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	primegen = $portVersion
+	cmd:primegaps = $portVersion
+	cmd:primes = $portVersion
+	cmd:primespeed = $portVersion
+	lib:libprimegen
+	"
+REQUIRES="
+	haiku
+	"
+
+PROVIDES_devel="
+	primegen_devel = $portVersion
+	devel:libprimegen
+	"
+REQUIRES_devel="
+	primegen == $portVersion base
+	"
+
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	cmd:nroff
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+BUILD()
+{
+	make
+}
+
+INSTALL()
+{
+	install -d $binDir $docDir $manDir/man1 $manDir/man3
+	install -d $includeDir/primegen $developLibDir $libDir
+	install -t $binDir primegaps primes primespeed
+	install -t $docDir BLURB README
+	install -t $manDir/man1 primegaps.1 primes.1 primespeed.1
+	install -t $manDir/man3 primegen.3
+	install -t $includeDir/primegen primegen.h uint32.h uint64.h
+	install -T primegen.a $developLibDir/libprimegen.a
+	install -T primegen.so $libDir/libprimegen.so
+	prepareInstalledDevelLib libprimegen
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+		$manDir/man3
+}
+
+TEST()
+{
+	[ "$(./primes 1 1000000 | md5sum | cut -d " " -f1)" = c13929ee9d2aea8f83aa076236079e94 ]
+}
--- a/sci-libs/primegen/patches/primegen-0.97.patchset
+++ b/sci-libs/primegen/patches/primegen-0.97.patchset
@@ -1,0 +1,34 @@
+From 7a947552d7374f9e99bdf128a9969ad042e1a6bf Mon Sep 17 00:00:00 2001
+From: Mason X <me@masonx.ca>
+Date: Wed, 29 Nov 2017 18:07:28 +0000
+Subject: [PATCH] Create shared object/library
+
+---
+ make-compile.sh | 2 +-
+ make-makelib.sh | 5 +++--
+ 2 files changed, 4 insertions(+), 3 deletions(-)
+
+diff --git a/make-compile.sh b/make-compile.sh
+index a1eb501..b0be539 100644
+--- a/make-compile.sh
++++ b/make-compile.sh
+@@ -1 +1 @@
+-echo exec "$CC" -c '${1+"$@"}'
++echo exec "$CC" -c -fPIC '${1+"$@"}'
+diff --git a/make-makelib.sh b/make-makelib.sh
+index d6b7c8c..ed01101 100644
+--- a/make-makelib.sh
++++ b/make-makelib.sh
+@@ -1,6 +1,7 @@
+ echo 'main="$1"; shift'
+-echo 'rm -f "$main"'
+-echo 'ar cr "$main" ${1+"$@"}'
++echo 'rm -f `echo "$main" | cut -f 1 -d '.'`.so "$main"'
++echo 'ar cr "$main" ${1+"$@"}'
++echo 'gcc -shared -o `echo "$main" | cut -f 1 -d '.'`.so ${1+"$@"}'
+ 
+ case "$1" in
+ sunos-5.*) ;;
+-- 
+2.14.2
+


KapiX:
A lot of patches is required but they add up to not so many changes. Please add them as patchset and use tar.gz from cr.yp.to as a source.

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+primegen can generate primes up to 1000000000000000, although it is not \
+optimized for primes past 32 bits. It uses the Sieve of Atkin instead of the \
+traditional Sieve of Eratosthenes."
+HOMEPAGE="https://cr.yp.to/primegen.html"
+COPYRIGHT="1999, D. J. Bernstein"
+LICENSE="Public Domain"
+REVISION="1"
+SOURCE_URI="https://github.com/ohnx/haiku-primegen/archive/$portVersion.tar.gz"
+SOURCE_DIR="haiku-primegen-$portVersion"
+CHECKSUM_SHA256="9eb6945fbf9f632eaed680b134025e84cb52703c0da1549e1dc14b3a47e6b497"
```
fbrosson: @ohnx, please swap SOURCE_DIR and CHECKSUM_SHA256 to meet the recommended order.
See [HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering)
Thanks!

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+
+ARCHITECTURES="x86 x86_gcc2 x86_64"
```
fbrosson: Not really important, but if you could change this to:
```
```
it would be great, because most of our recipes have them in that order.
Thanks again!

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+
+PROVIDES="
+    devel:primegen = $portVersion
+    cmd:primes = $portVersion
+    cmd:primegaps = $portVersion
+    "
```
fbrosson: Please replace the spaces by a tab (in each line where you have multiple spaces.)

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+
+REQUIRES="
+        haiku
+        "
```
fbrosson: Same request as above, regarding the indentation.
(Please also fix this in the rest of the recipe.)

sci-libs/primegen/primegen-0.0.97.recipe
fbrosson: I've noticed that the program builds fine without any changes and even seems to work fine too.
Have you tried this:
```
SOURCE_URI="https://cr.yp.to/primegen/primegen-$portVersion.tar.gz"
CHECKSUM_SHA256="54285baf8eed9e421ff2220a2112d38cfb20c1ebef6014ef3f0004c22c95f40d"
```
Of course, if you try this you should rename the recipe file to primegen-0.97.recipe

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+SOURCE_URI="https://github.com/ohnx/haiku-primegen/archive/$portVersion.tar.gz"
+SOURCE_DIR="haiku-primegen-$portVersion"
+CHECKSUM_SHA256="9eb6945fbf9f632eaed680b134025e84cb52703c0da1549e1dc14b3a47e6b497"
+
+ARCHITECTURES="x86 x86_gcc2 x86_64"
+
+PROVIDES="
+    devel:primegen = $portVersion
+    cmd:primes = $portVersion
+    cmd:primegaps = $portVersion
```
fbrosson: Should be:
```
PROVIDES="
	primegen = $portVersion
	cmd:primegaps = $portVersion
	cmd:primes = $portVersion
	cmd:primespeed = $portVersion
	"
```
BTW, we usually don't put a blank line between blocs PROVIDES and REQUIRES.

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+
+REQUIRES="
+        haiku
+        "
+
+BUILD_PREREQUIRES="
+        cmd:make
+        cmd:nroff
+        cmd:shar
+        cmd:gcc
```
fbrosson: It could be nice to also define a devel package (and remove unneeded cmd: lines in BUILD_REQUIRES):
```
PROVIDES_devel="
	primegen_devel = $portVersion
	devel:libprimegen
	"
REQUIRES_devel="
	primegen == $portVersion base
	"

BUILD_PREREQUIRES="
	cmd:gcc
	cmd:make
        "
```

sci-libs/primegen/primegen-0.0.97.recipe
```diff
@@ -0,0 +1,61 @@
...
+        "
+
+BUILD()
+{
+    make
+}
+
+INSTALL()
+{
+    make setup
```
fbrosson: This is what I used:
```
INSTALL()
{
	install -d $binDir $docDir $manDir/man1 $manDir/man3
	install -d $includeDir/primegen $developLibDir
	install -t $binDir primegaps primes primespeed
	install -t $docDir README
	install -t $manDir/man1 primegaps.1 primes.1 primespeed.1
	install -t $manDir/man3 primegen.3
	install -t $includeDir/primegen primegen.h uint32.h uint64.h
	install -T primegen.a $developLibDir/libprimegen.a

	# devel package
	packageEntries devel \
		$developDir \
		$manDir/man3
}
```

ohnx: Thanks for that!

The builtin `make setup` from the program compiles a number of extra programs that need the patching... I will try this instead, since those extra programs aren't necessarily needed.

ohnx:
@fbrosson I have made the changes you suggested and added the shared library and test function.

If it's all good now, please let me know and I'll squash all the commits :)

extrowerk:
YAY, @fbrosson is back!

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,81 @@
...
+
+PROVIDES="
+	primegen = $portVersion
+	cmd:primegaps = $portVersion
+	cmd:primes = $portVersion
+	cmd:primespeed = $portVersion
+	"
+REQUIRES="
+		haiku
+		"
```
fbrosson: One tab (per line, in lines 26 and 27) is enough for here ;-)

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,81 @@
...
+	install -d $binDir $docDir $manDir/man1 $manDir/man3
+	install -d $includeDir/primegen $developLibDir $libDir
+	install -t $binDir primegaps primes primespeed
+	install -t $docDir README
+	install -t $manDir/man1 primegaps.1 primes.1 primespeed.1
+	install -t $manDir/man3 primegen.3
+	install -t $includeDir/primegen primegen.h uint32.h uint64.h
+	install -T primegen.a $developLibDir/libprimegen.a
+	install -T primegen.so $developLibDir/libprimegen.so
+	install -T primegen.so $libDir/libprimegen.so
```
fbrosson: I think it would be better to replace lines 62 and 63 by:
```
	install -T primegen.so $libDir/libprimegen.so
	prepareInstalledDevelLib libprimegen
	fixPkgconfig
```
but I haven't checked yet.

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,81 @@
...
+TEST()
+{
+	[ "$(primes 1 1000000 | md5sum)" != "c13929ee9d2aea8f83aa076236079e94  -" ] && echo "Test failed!" && exit
+	[ "$(primes 1 10000000 | md5sum)" != "60e34d268bad671a5f299e1ecc988ff6  -" ] && echo "Test failed!" && exit
+	[ "$(primes 1 100000000 | md5sum)" != "4e2b0027288a27e9c99699364877c9db  -" ] && echo "Test failed!" && exit
+	[ "$(primes 1 1000000000 | md5sum)" != "92c178cc5bb85e06366551c0ae7e18f6  -" ] && echo "Test failed!" && exit
+	[ "$(primes 1 10000000000 | md5sum)" != "95ded65c9ddca18e1df966ba2a2b373a  -" ] && echo "Test failed!" && exit
+	[ "$(primes 1 100000000000 | md5sum)" != "9d8a8c2d7221497843ca289e6f463bfd  -" ] && echo "Test failed!" && exit
+	[ "$(primes 1 1000000000000 | md5sum)" != "c899bf5d0ccef05787ac969afd6606ff  -" ] && echo "Test failed!" && exit
+	echo "Test passed!"
```
fbrosson: Hmm, I think you can keep only the first test (1000000), or, maybe, the one with 10000000000, if it does not take too long, in order to check that it can handle ints bigger than 2^32.
BTW, I'd use `./primes` instead of just `primes`, just to make sure it will always be found.
You also don't need to have a `echo "Test passed!"` because haikuporter only cares about the return code.
So I 'd use something like:
```
	MD5="`./primes 1 1000000 | md5sum | cut -d" " -f1`"
	test "$MD5" = c13929ee9d2aea8f83aa076236079e94
```
But please feel free to keep your style if you wish. This is only cosmetics ;-)

BTW, if md5sum is not in coreutils, then you need to add `cmd:md5sum` in BUILD_PREREQUIRES.
Hmm, in case you haven't checked whether or not the recipe really needs `cmd:nroff` could you try to build it without nroff?
(It's always better to avoid unneeded dependencies.)

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,66 @@
...
+primegen can generate primes up to 1000000000000000, although it is not \
+optimized for primes past 32 bits. It uses the Sieve of Atkin instead of the \
+traditional Sieve of Eratosthenes."
+HOMEPAGE="https://cr.yp.to/primegen.html"
+COPYRIGHT="1999, D. J. Bernstein"
+LICENSE="Public Domain"
+REVISION="2"
+SOURCE_URI="https://cr.yp.to/primegen/primegen-$portVersion.tar.gz"
+CHECKSUM_SHA256="54285baf8eed9e421ff2220a2112d38cfb20c1ebef6014ef3f0004c22c95f40d"
+SOURCE_DIR="primegen-$portVersion"
```
fbrosson: You can (well, should, in fact) remove the "SOURCE_DIR" line when it has the default value. (That's the case here.)

sci-libs/primegen/primegen-0.97.recipe
fbrosson: Same remark for BUILD_PREREQUIRES and BUILD_REQUIRES.

sci-libs/primegen/primegen-0.97.recipe
ohnx: I tried it without nroff, the build does in fact need nroff.

ohnx: Does https://www.gnu.org/software/coreutils/manual/html_node/md5sum-invocation.html mean that md5sum is likely in coreutils?

sci-libs/primegen/primegen-0.97.recipe
ohnx: This doesn't affect the final output, so I made those changes.

sci-libs/primegen/primegen-0.97.recipe
fbrosson: Hmm, I guess you had already noticed that `nroff` is needed during the build, to create 4 files:
* primegaps.0
* primegen.0
* primes.0
* primespeed.0

But we don't really need them, so it might be worth patching the `Makefile` to replace
```
it: \
man prog
```
by
```
it: \
prog
```
The quick way to do that is to add a PATCH function before BUILD:
```
PATCH()
{
	# Make "it" depend on "prog" only, instead of "man" and "prog"
	sed -i -e "s/^man prog$/prog/" Makefile
}
```
But it might be better to use the patch file for that.

Let's see what the other contributors think about this.
Maybe, after all, it's better to keep the `cmd:nroff` and ignore the files it generates.

ohnx: The >32-bit test takes a while, so I decided to only keep the smallest one.

When and where is the test run? If it's after the `INSTALL` function, isn't using `primes` to test the correct installation of the `primes` program part of the testing? 😛

sci-libs/primegen/primegen-0.97.recipe
fbrosson: Hmm, I think there is a tiny difference. When we use `prepareInstalledDevelLib` we get a symbolic link (for develop/lib/libprimegen.so) in the primegen_devel package that points to the real file (lib/libprimegen.so) in the primegen package.
So I would recommend using `prepareInstalledDevelLib`.

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,81 @@
...
+PATCHES="primegen-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	primegen = $portVersion
+	cmd:primegaps = $portVersion
+	cmd:primes = $portVersion
+	cmd:primespeed = $portVersion
+	"
```
fbrosson: Since you have made the changes to build the shared lib, we need to declare it in `PROVIDES` with:
```
	lib:libprimegen
```

BTW, with the current recipe the 3 runtimes (primegaps, primes and primespeed) are being linked against the static lib.
It would be nice to make them use the shared lib. I'll let you dig that ;-)

ohnx: The way that the applications are compiled is rather complicated... Right now, I make minimal changes to the Makefile and most of the application is compiled in order to keep things simple... If the programs compiled using the shared library instead, it might mean a lot of changes...

sci-libs/primegen/primegen-0.97.recipe
fbrosson: > Does ... mean that md5sum is likely in coreutils?

Sorry, I should have given the trick:
```
package list -p /system/packages/coreutils-*.hpkg | grep ^bin/md5sum
```
which yelds
```
bin/md5sum
```
So yes, we (also) have it in the coreutils package in Haiku.
This means that we don't need to add a ```cmd:md5sum``` in BUILD_PREREQUIRES because coreutils is always "activated" by haikuporter during the builds.

sci-libs/primegen/primegen-0.97.recipe
fbrosson: Don't worry. In case the changes are not trivial then we can keep the static linking.

Regarging your questions about the tests:
> The >32-bit test takes a while, so I decided to only keep the smallest one.

Perfect.

> When and where is the test run? If it's after the INSTALL function, isn't using primes to test the correct installation of the primes program part of the testing? 😛

The TEST function is only called by haikuporter if you pass `--test` to it.
But you can only call `haikuporter --test primegen` _after_ a successful build.

BTW, it is recommended to use the ` --strict-policy` (or its short `-S` syntax) option during the build.
That's how I found that the `lib:libprimegen` was missing ;-)

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,81 @@
...
+HOMEPAGE="https://cr.yp.to/primegen.html"
+COPYRIGHT="1999, D. J. Bernstein"
+LICENSE="Public Domain"
+REVISION="2"
+SOURCE_URI="https://cr.yp.to/primegen/primegen-$portVersion.tar.gz"
+CHECKSUM_SHA256="54285baf8eed9e421ff2220a2112d38cfb20c1ebef6014ef3f0004c22c95f40d"
+SOURCE_DIR="primegen-$portVersion"
+PATCHES="primegen-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
```
fbrosson: I've ran the 5 first md5sums tests (i.e. from "primes 1 1000000" to "primes 1 10000000000") on both x86_64 and x86_gcc2 and everything is OK.
So you can drop the "?" in front of x86_gcc2 ;-)

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,78 @@
...
+	primegen_devel = $portVersion
+	devel:libprimegen
+	"
+REQUIRES_devel="
+	primegen == $portVersion base
+	"
+
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:gcc
```
fbrosson: I forgot to mention this earlier, but, when we merge new recipes, we like to sort the `cmd:` lines before merging.
So if you could swap the `cmd:make` and the `cmd:gcc` that would be great!

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,78 @@
...
+		$developDir \
+		$manDir/man3
+}
+
+TEST()
+{
+	# 32-bit test
+	[ "$(primes 1 1000000 | md5sum)" != "c13929ee9d2aea8f83aa076236079e94  -" ] && exit 1
+	# all tests passed
+	exit 0
```
fbrosson: I agree that we *currently* don't need to replace `primes` by `./primes`, but we can't be 100% sure it will always work without the `./`, so I'd really recomment replacing `primes` by `./primes`.
And while at it, I'd even do a few more cosmetic changes to have either:
```
[ "$(./primes 1 1000000 | md5sum)" != "c13929ee9d2aea8f83aa076236079e94  -" ]
```
or
```
[ "$(./primes 1 1000000 | md5sum | cut -d" " -f1 )" != c13929ee9d2aea8f83aa076236079e94 ]
```
The reason why I would prefer the second solution is that we don't need to care about what md5sum outputs right after the checksum. I know this is probably never going to change, but this way we don't hardcode an arbitrary strng (the two spaces and the dash) in our test.

The other advantage of removing the `&& exit 0` (and using `=` instead of `!=` is that we can use the exit code from the `[` command (which, as you know, is a kind of synonym for the `test` command) and, hence, have a TEST function with just one line in its body.

Personally, I'd use
```
test "$(./primes 1 1000000 | md5sum | cut -d" " -f1 )" != c13929ee9d2aea8f83aa076236079e94
```
but I'm perfectly OK with
```
[ "$(./primes 1 1000000 | md5sum | cut -d" " -f1 )" != c13929ee9d2aea8f83aa076236079e94 ]
```
too.
Well, I hope I'm not being too picky ;-)
So if you change TEST to:
```
TEST()
{
	[ "$(./primes 1 1000000 | md5sum)" != "c13929ee9d2aea8f83aa076236079e94  -" ]
]
```
I'll also be satisfied ;-)
Thanks again and sorry for being (a little bit) picky ;-)

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,78 @@
...
+{
+	make
+}
+
+INSTALL()
+{
+	install -d $binDir $docDir $manDir/man1 $manDir/man3
+	install -d $includeDir/primegen $developLibDir $libDir
+	install -t $binDir primegaps primes primespeed
+	install -t $docDir README
```
fbrosson: I saw that the README mentions another interesting file, BLURB.
I think it would be nice to also add it to the docDir.
Don't worry, I won't be picky this time ;-)))) I'll be happy with any of these:
```
	install -t $docDir README BLURB
```
or
```
	install -t $docDir BLURB README
```

BTW, if/when you squash your commits, the most used style for commit messages (in situations like this one) is:
```
primegen: add initial recipe.
```

Regarding the squash, it's nice if you do it yourself, but that's not required, since GitHub has a "Squash and merge" button which works fine.

fbrosson:
Hmm, I just noticed that your changes are in your master branch.
That's not recommended because it makes it harder for you to pull new commits, should you wish to update you copy of the repository before the merge.
But don't worry, it won't be a problem for us, and we can continue to use this PR.

(Once this PR is merged, you will need to update your local master branch and then "git push --force" your local master to your remote master (on GitHub) to re-sync your GitHub's master branch with the projects master. BTW, in case you need help for that, please do not hesitate to ask.)

For your next PRs, you should use [feature-branches](https://github.com/haikuports/haikuports/wiki/DevelopmentModel#what-are-feature-branches).

ohnx:
Aha... thanks for that tip!

I will definitely use those in the future.

ohnx:
@fbrosson ready for merge.

sci-libs/primegen/primegen-0.97.recipe
```diff
@@ -0,0 +1,78 @@
...
+		$developDir \
+		$manDir/man3
+}
+
+TEST()
+{
+	# 32-bit test
+	[ "$(./primes 1 1000000 | md5sum | cut -d " " -f1)" == "c13929ee9d2aea8f83aa076236079e94" ]
+	# all tests passed
+	exit 0
```
fbrosson: Oops, you forgot to remove the `exit 0`.
The exit code returned by `TEST()` is the exit code of the *last* command in that function, so, since we have removed the ` && exit 0` at the end of the `| ... ]` test, we should not run any command avec the test.

Oops, you also used `==` in the test, while you should use `=`. Both work... with ``bash``, but with ``sh`` the only accepted syntax for testing if 2 strings are identical is `=`. Using bash-epecific syntax is not recommended when something can be done (with no extra cost) using sh-syntax.

BTW, the c13929ee9d2aea8f83aa076236079e94 string does not need to be quoted, since it does not contain any special caracter. Well, keeping the quotes does not harm, but we can safely drop them.
(On the left side of the test, the quotes around the `$(...)` are fine and it would not be safe to remove them.)

So here what you can put in TEST to fix these tiny issues:
```
TEST()
{
	[ "$(./primes 1 1000000 | md5sum | cut -d " " -f 1)" = c13929ee9d2aea8f83aa076236079e94 ]
}
```
Regarding the `cut` command, the `-d` and `-f` options both accept a space before their values, but it's nice to either have no space at all for each of them, or a space each of them. Well, that's once again just style ;-)
BTW, you could also use `cut -d\  -f1` but in that case you really need two spaces after the backslash.

I hope you're not too angry with me ;-) because your first and second definitions of TEST were both correct, and then I suggested a style change...
So sorry again for my picky requests and thanks for your excellent work!

fbrosson: Cool, you squashed you commits and chose a nice commit message.

After you fix TEST, remember that you can use
```
git commit --amend primegen-0.97.recipe
```
so that you won't need to redo a squash.

ohnx: Yup! Amended :)

sci-libs/primegen/primegen-0.97.recipe
mmlr: For reference:
```
pkgman search cmd:md5sum
```

Would be more useful if you don't already have an idea where something may come from.

The packagefs also adds virtual attributes to all files that tell what package they are in. You can list them with listattr:

```
listattr -l /bin/md5sum
```

Finally, not adding cmd:md5sum because coreutils happens to be part of the default setup isn't very future proof. A lot of times it's hard to know what implicit tools are used by a certain buildsystem. But if you this obviously need a specific tool, I'd personally add it explicitly.

ohnx: Do the packages in coreutils change often?

ohnx: So I was looking through some other packages and I noticed that they explicitly require `cmd:make`, `cmd:gcc`, and whatnot... is that what you are suggesting I do?

fbrosson: I'm quite sure the list of available commands in coreutils never changes.
Regarding the suggested inclusion of `cmd:md5sum` in BUILD_PREREQUIRES, @mmlr was just suggesting that maybe we could decide to change.or remove the (non-written?) rule which allows us to not include such dependencies.
But I guess that removing this rule would cause so much trouble that this will never happen.
As a matter of fact, some of the commands available in coreutils are used in most recipes. So if we wanted to drop this rule we would need to include the `cmd:` lines in really a lot of recipes.
So, don't worry about the primegen's recipe. It's perfect!

pulkomandy: The fact that "it just works now" is not a reason to ignore this. We may decide to split coreutils in multiple smaller packages, or they may just remove commands from there and we would have to provide replacements in a different package, or we may replace the GPL licenced coreutils by some BSD version, or someone would want to provide a busybox-like all-in-one implementation, or...

And yes, as you said, when this happens it will be a lot of work. But this is not a reason to make it even worse when you know you use a command and don't add a requires for it. By raising awareness about this, we can do the work progressively as we notice such problems, instead of waiting for them to pile up and then having a lot of trouble when they actually start breaking things.

ohnx: Is there a `TEST_PREREQUIRES` that I could use for a command that is only needed for tests?

pulkomandy: Not at the moment, but we should probably add this to haikuporter.
I suggest the following pattern:

TEST_PREREQUIRES="..."
BUILD_PREREQUIRES += TEST_PREREQUIRES

This way, when HaikuPorter directly supports it, we can simply remove the += line.

PR #1823, opened 2017-12-02
Scooby: new recipe

nysnatuss:
Added [Scooby](https://github.com/HaikuArchives/Scooby), an open source full featured BeOS native e-mail client.

### DIFF
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -13,21 +13,21 @@ SOURCE_DIR="Scooby-$srcGitRev"
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 
 PROVIDES="
-	scooby = $portVersion
-	app:Scooby = $portVersion
+	scooby$secondaryArchSuffix = $portVersion
+	app:Scooby$secondaryArchSuffix = $portVersion
 	"
 REQUIRES="
-	haiku
-	lib:libsanta
+	haiku$secondaryArchSuffix
+	lib:libsanta$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
-	haiku_devel
-	devel:libsanta
+	haiku${secondaryArchSuffix}_devel
+	devel:libsanta$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
 	makefile_engine
-	cmd:gcc
+	cmd:gcc$secondaryArchSuffix
 	cmd:make
 	"
 
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -1,0 +1,42 @@
+SUMMARY="Full featured e-mail cilent"
+DESCRIPTION="Scooby is an open source full featured BeOS native e-mail client."
+HOMEPAGE="https://sourceforge.net/p/scooby/"
+COPYRIGHT="Atsushi Takamatsu"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="e53f65326329dd5a85fbdee640f4edefc2d043d3"
+SOURCE_URI="https://github.com/HaikuArchives/Scooby/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384c"
+SOURCE_FILENAME="$portVersionedName.tar.gz"
+SOURCE_DIR="Scooby-$srcGitRev"
+
+ARCHITECTURES="x86 x86_gcc2"
+
+PROVIDES="
+	scooby = $portVersion
+	app:Scooby = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libsanta
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	libsanta_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	makefile_engine
+	"
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -17,7 +17,7 @@ PROVIDES="
 	app:Scooby = $portVersion
 	"
 REQUIRES="
-	haiku$secondaryArchSuffix
+	haiku
 	lib:libsanta
 	"
 
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -10,24 +10,25 @@ CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384
 SOURCE_FILENAME="$portVersionedName.tar.gz"
 SOURCE_DIR="Scooby-$srcGitRev"
 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="x86_gcc2 !x86 !x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2"
 
 PROVIDES="
-	scooby$secondaryArchSuffix = $portVersion
-	app:Scooby$secondaryArchSuffix = $portVersion
+	scooby = $portVersion
+	app:Scooby = $portVersion
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libsanta$secondaryArchSuffix
+	lib:libsanta
 	"
 
 BUILD_REQUIRES="
-	haiku${secondaryArchSuffix}_devel
-	devel:libsanta$secondaryArchSuffix
+	haiku_devel
+	devel:libsanta
 	"
 BUILD_PREREQUIRES="
 	makefile_engine
-	cmd:gcc$secondaryArchSuffix
+	cmd:gcc
 	cmd:make
 	"
 
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -11,7 +11,6 @@ SOURCE_FILENAME="$portVersionedName.tar.gz"
 SOURCE_DIR="Scooby-$srcGitRev"
 
 ARCHITECTURES="x86_gcc2 !x86 !x86_64"
-SECONDARY_ARCHITECTURES="x86_gcc2"
 
 PROVIDES="
 	scooby = $portVersion
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -38,5 +38,8 @@ BUILD()
 
 INSTALL()
 {
-	make install
+	mkdir -p $appsDir
+	make install TARGET_DIR=$appsDir
+
+	addAppDeskbarSymlink $appsDir/Scooby
 }
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -10,7 +10,7 @@ CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384
 SOURCE_FILENAME="$portVersionedName.tar.gz"
 SOURCE_DIR="Scooby-$srcGitRev"
 
-ARCHITECTURES="x86_gcc2 !x86 !x86_64"
+ARCHITECTURES="?x86_gcc2 !x86 !x86_64"
 
 PROVIDES="
 	scooby = $portVersion
--- a/mail-client/scooby/scooby-0.9.9.recipe
+++ b/mail-client/scooby/scooby-0.9.9.recipe
@@ -1,4 +1,4 @@
-SUMMARY="Full featured e-mail cilent"
+SUMMARY="Full featured e-mail client"
 DESCRIPTION="Scooby is an open source full featured BeOS native e-mail client."
 HOMEPAGE="https://sourceforge.net/p/scooby/"
 COPYRIGHT="Atsushi Takamatsu"
@@ -10,7 +10,7 @@ CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384
 SOURCE_FILENAME="$portVersionedName.tar.gz"
 SOURCE_DIR="Scooby-$srcGitRev"
 
-ARCHITECTURES="x86 x86_gcc2"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
 
 PROVIDES="
 	scooby = $portVersion
@@ -23,12 +23,12 @@ REQUIRES="
 
 BUILD_REQUIRES="
 	haiku_devel
-	libsanta_devel
+	devel:libsanta
 	"
 BUILD_PREREQUIRES="
+	makefile_engine
 	cmd:gcc
 	cmd:make
-	makefile_engine
 	"
 
 BUILD()


mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+	app:Scooby = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libsanta
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	libsanta_devel
```
fbrosson: We prefer to write `devel:libsanta` instead of `libsanta_devel`.

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
+SUMMARY="Full featured e-mail cilent"
```
fbrosson: cilent →client

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+COPYRIGHT="Atsushi Takamatsu"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="e53f65326329dd5a85fbdee640f4edefc2d043d3"
+SOURCE_URI="https://github.com/HaikuArchives/Scooby/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384c"
+SOURCE_FILENAME="$portVersionedName.tar.gz"
+SOURCE_DIR="Scooby-$srcGitRev"
+
+ARCHITECTURES="x86 x86_gcc2"
```
fbrosson: I haven't tested yet, but it would be nice to see if it also builds on x86_64, in which case you could write:
```
```
Oh, yes, since most of our recipes list x86_gcc2 before x86, we try to keep the same order. Yep, that's cosmetics ;-)

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	libsanta_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	makefile_engine
```
fbrosson: We traditionally to put makefile_engine above all the cmd: lines.
Yep, that's cosmetics once more ;-)

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
```
fbrosson: Your recipe is missing an newline char (0x0A) on the very last line.
This does not break anything; but it is cool to have the newline when editing with vi.

nysnatuss:
@fbrosson I've applied the changes. Let me know if there's anything that need to be changed!

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="e53f65326329dd5a85fbdee640f4edefc2d043d3"
+SOURCE_URI="https://github.com/HaikuArchives/Scooby/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384c"
+SOURCE_FILENAME="$portVersionedName.tar.gz"
+SOURCE_DIR="Scooby-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
```
Begasus: if you want to build for x86 on x86_gcc2 maybe add $secondaryArchSuffix at the needed places?

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
```
Begasus: make install doesn't work for me (INSTALL_DIR is empty) ...

mail-client/scooby/scooby-0.9.9.recipe
fbrosson: @Begasus, we don't really need to add support for secondary architecture.

If it did not build on x86_gcc2 primary arch, or if it did provide at least one lib, or if the program would run much faster when built using a recent compiler, then, yes, it would be worth adding support for secondary architecture.

But maybe I'm missing something.

BTW, if you choose to keep support for secondary arch, then you need to add a SECONDARY_ARCHITECTURES="x86" as well as appropriate CONFLICTS (or small changes) to avoid overlaps between primary and secondary arch packages.

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,42 @@
...
+	"
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
```
Begasus: This should do the trick (and adds a menu item in the Deskbar) :)

mkdir -p $appsDir
make install TARGET_DIR=$appsDir
 
addAppDeskbarSymlink $appsDir/Scooby

mail-client/scooby/scooby-0.9.9.recipe
Begasus: Right there @fbrosson (ps nice to see you back!) :+1: 
I'm just wondering if x86 is still needed then?

Begasus:
Still crashes on exit here ... Otherwise builds and starts up ok on gcc2h :+1:

Begasus:
x86_64 build isn't working, fails with multiple errors ...

mail-client/scooby/scooby-0.9.9.recipe
fbrosson: Thanks @Begasus ;-)
I guess we can keep x86 (i.e. avoid putting "?x86"), at least for the moment.

If a package builds fine on x86_gcc2 primary arch and on x86_64 then it will certainly also build fine on x86.
And in that case we can use
```
```

For packages which don't build on x86_gcc2 but build fine for x86 secondary arch, we can probably avoid the "?" in front of the "x86" in ARCHITECTURES. We would have for such cases:
```
SECONDARY_ARCHITECTURES="x86"
```

I'd only use "?x86" (in ARCHITECTURES) in strange cases (if any) where a package would build on x86_64 but not on x86 (primary or secondary architectures). But I guess this should never happen.

fbrosson: Sorry @nysnatuss, I forgot to tell that you may remove all occurences of `$secondaryArchSuffix` (and `$[secondaryArchSuffix}`).

Well, in case the crashes only happen with x86_gcc2 primary arch build (and not with x86 secondary arch builds on x86_gcc2 hybrid), then it might be worth keeping the secondaryArchSuffix, addng SECONDARY_ARCHITECTURES="x86" and changing "x86_gcc2" to "!x86_gcc2" in ARCHITECTURES.

fbrosson:
Hmm, it looks like the sources are written in a C dialect which was deprecated after gcc2.
I wonder if inserting `CFLAGS="--std=c90"` or (`CFLAGS="--std=gnu90"`) to the right of the "make" commands in INSTALL and in BUILD would help.

Begasus:
command line option '-std=c90' is valid for C/ObjC but not for C++ (doesn't
seem to work in either case)

2017-12-02 12:16 GMT+01:00 fbrosson <notifications@github.com>:

mail-client/scooby/scooby-0.9.9.recipe
Begasus: didn't check the $secondaryArchSuffix (yet) to see if the crash is also there ...

nysnatuss: FYI, with $secondaryArchSuffix it builds ok with x86_gcc2.

Begasus:
Same issues for x86 (doesn't build)

mail-client/scooby/scooby-0.9.9.recipe
Begasus: if you still build for gcc2 yes, that deosn't effect the x86_gcc2 build (if you try to build for x86 (gcc5)) it fails ...
try "hp scooby_x86" ...

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,45 @@
...
+COPYRIGHT="Atsushi Takamatsu"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="e53f65326329dd5a85fbdee640f4edefc2d043d3"
+SOURCE_URI="https://github.com/HaikuArchives/Scooby/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384c"
+SOURCE_FILENAME="$portVersionedName.tar.gz"
+SOURCE_DIR="Scooby-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
fbrosson: @nysnatuss, since the package only builds on x86_gcc2, here's what you can do:
Replace
```
```
by
```
ARCHITECTURES="x86_gcc2 !x86 !x86_64"
SECONDARY_ARCHITECTURES="x86_gcc2"
```
and **keep** all the occurences of secondaryArchSuffix.

You should also replace
```
SOURCE_FILENAME="$portVersionedName.tar.gz"
```
by
```
SOURCE_FILENAME="scooby-$portVersion.tar.gz"
```

mail-client/scooby/scooby-0.9.9.recipe
Begasus: first add SECONDARY_ARCHITECTURES="x86" as fbrosson already mentioned after the ARCHITECTURES to be able to trigger the x86 build

fbrosson: We only add
```
SECONDARY_ARCHITECTURES="x86"
```
for packages which don't build on x86_gcc2 primary arch.
Here we are in the opposite situation. It only builds on x86_gcc2.
So what you can add is add:
```
SECONDARY_ARCHITECTURES="x86_gcc2"
```
It will allow users who build their own Haiku-x86 hybrid from source with support for x86_gcc2 secondary arch.
These users would then call
```
haikuporter scooby_x86_gcc2
```
after having added x86_gcc2 to the SECONDARY_ARCHITECTURES parameter in their haikuports.conf, of course.
See https://github.com/haikuports/haikuports/wiki/HaikuPorter-HaikuPortsConf for more details ;-)


Yes, this won't be very usefull to users of x86_64, unless we start shipping x86_64 hybrid (i.e. with support for x86_gcc2 as a secondary arch, like we did for x86 until a few months ago.)

humdingerb:
> ARCHITECTURES="x86_gcc2 !x86 !x86_64"
> SECONDARY_ARCHITECTURES="x86_gcc2"

I'm sorry, but why add the 2nd architecture if it only builds on gcc2?
x86 (32bit) isn't a supported arch anymore and x86_64 will most probably never get gcc2 as 2ndary arch. So, once it builds on 64bit too, just remove the "!" from x86_64 to get: ARCHITECTURES="x86_gcc2 x86_64"

fbrosson:
@humdingerb, I know that x86 is no longer supported.
The reason why I wrote we could add x86_gcc2 (or "?x86_gcc2" if you prefer) in SECONDARY_ARCHITECTURES was to allow those hypothetic users who would like to compile Haiku-x86 hybrid by themselves to build scooby for x86_gcc2 secondary arch.

That said, I do agree that adding support for a secondary architecture which is not available on any of the currently supported primary architectures would not be usefull to most users.
And I'm quite sure nobody will try to play with Haiku-x86 now that this arch is no longer supported.

So, after all, if you say that x86_64 will probably never get support for x86_gcc2 secondary arch, the best thing to do for this recipe is to remove all occurences of `$secondaryArchSuffix` and `${secondaryArchSuffix}`.

Begasus:
Think I already mentioned this in other PR's, as x86 isn't supported anymore it shouldn't be added if it fails on gcc2h :) (only if x86_gcc2 fails and you need to build x86 on a gcc2h it would be usefull) (just my 2cents :)

fbrosson:
Yep. That's the good strategy. To rephrase it a little:

* If a recipe builds fine on x86_gcc2 (primary arch) but not on x86  and x86_64, then we should no longer try to add support for secondary architectures on that recipe. We'd use:
```
ARCHITECTURES="x86_gcc2 !x86 !x86_64"
```

* If a recipe does not build on x86_gcc2 primary arch, but builds on x86 secondary arch, then it's OK to add support for x86 secondary arch. We'd use:
```
SECONDARY_ARCHITECTURES="x86"
```

humdingerb:
Exactly.

Begasus:
One more addition ;)
If a recipe builds ok on all arch's but needs a secondary arch (deppendencies) one should use:
SECONDARY_ARCHITECTURES="x86"
with the "$secondaryArchSuffix" :)

waddlesplash:
Just chiming in here to confirm @fbrosson's analysis with @Begasus' caveat (i.e. @fbrosson is correct for application recipes, but recipes with `lib:` provides may get a secondary arch recipe due to apps on that arch depending on it.)

waddlesplash:
Also: Some of the binary libraries used in Scooby have source code in the SilverWing repository. (Not all, but over half I think). So something could be done about that.

nysnatuss:
That's a lot to absorb, but I think I might get the hang of it. :)

waddlesplash:
No, if you have removed $secondaryArchSuffix then you should not have any `SECONDARY_ARCHITECTURES`. Sorry for inundating you with all this, we don't usually run into such issues :)

nysnatuss:
Thanks for the clarification, and also at least we all learn something today! 😁

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,45 @@
...
+SOURCE_DIR="Scooby-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 !x86 !x86_64"
+
+PROVIDES="
+	scooby = $portVersion
+	app:Scooby = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
```
Begasus: remove the $secondaryArchSuffix here also ;)

fbrosson: Yep, I had also noticed this one, but @Begasus was faster. Well, this is because I was preparing a longer comment ;-)

When haikuporter is building a recipe for a primary arch, $secondaryArchSuffix is an empty string.
When haikuporter is building a recipe for a secondary arch, e.g. x86, $secondaryArchSuffix is set to e.g. _x86.

But haikuporter will only accept to build a recipe for a secondary arch if that arch (e.g. x86) appears in SECONDARY_ARCHITECTURES.

This means that if SECONDARY_ARCHITECTURES is not defined in a given recipe then there is no need to have any $secondaryArchSuffix at all in that recipe.
So if you see SECONDARY_ARCHITECTURES in a recipe, then you should see secondaryArchSuffix here and there.
And if you don't see SECONDARY_ARCHITECTURES then you should not see any secondaryArchSuffix.

That said, having one or more occurrences of secondaryArchSuffix but no SECONDARY_ARCHITECTURES won't break anything. It is, however, not recommended because it would be just noise, for a recipe that has no support for secondary architectures.

mail-client/scooby/scooby-0.9.9.recipe
```diff
@@ -0,0 +1,45 @@
...
+COPYRIGHT="Atsushi Takamatsu"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="e53f65326329dd5a85fbdee640f4edefc2d043d3"
+SOURCE_URI="https://github.com/HaikuArchives/Scooby/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="29630d6ef1ff51cb463762d9bed23c8ebadac045e023c28b1ce1e8d72469384c"
+SOURCE_FILENAME="$portVersionedName.tar.gz"
+SOURCE_DIR="Scooby-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 !x86 !x86_64"
```
fbrosson: The app also crashes for me, on exit.
It looks like it's a SEGMENTATION VIOLATION happenning in libssl or libcrypto.
I guess it would be appropriate to mark x86_gcc2 as untested, i.e. with a "?":
```
ARCHITECTURES="?x86_gcc2 !x86 !x86_64"
```
Then we can merge ;-)
Thanks!

PR #1827, opened 2017-12-03
libpipeline, new recipe

Begasus:


Begasus:
New to this one, will look into it :)

Begasus:
Not tested on x86_64 (yet), but builds ok also for x86 ...

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,72 @@
...
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libpipeline$secondaryArchSuffix \
+	$libDir/libpipeline.so.1.5.0
```
fbrosson: I guess you wanted to replace 1.5.0 by $portVersion here too, since you use $portVersion for the `lib:libpipeline` and `devel:libpipeline` lines.

Begasus:
Updated @fbrosson hope this one is ok? :)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,72 @@
...
+
+	#remove libtool file
+	rm $libDir/*.la
+
+	prepareInstalledDevelLib libpipeline
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir
```
fbrosson: I've just tried building libpipeline and noticed that the man3 pages are in the main package.
It would be nice to move $manDir/man3 to the devel package.
You can do so by replacing
```
	# devel package
	packageEntries devel \
		$developDir
```
by
```
		rm -rf $documentationDir
		maybe_manDir_man3=
	else
		maybe_manDir_man3=$manDir/man3
	fi

	# devel package
	packageEntries devel \
		$developDir \
		$maybe_manDir_man3
```
Other benefit: it will get rid of the docs on secondary arch builds ;-)

Begasus: This doesn't work if you're building for $secondaryArchSuffix?

fbrosson: Oops, you are perfectly right. My mistake!
This should work:
```
		rm -rf $manDir
		maybe_manDir=
	else
		maybe_manDir=$manDir
	fi

	# devel package
	packageEntries devel \
		$developDir \
		$maybe_manDir

	rmdir $documentationDir
```

Actually, when I noticed that the only man section was man3, I though we could simplify the INSTALL without having to use the maybe_manDir... variable, but I did not test the "simplified" version...

BTW, I did not test what I wrote above.

Begasus: One question here, why remove the $mandir for $secondaryArchSuffix?

Begasus:
@fbrosson updated :)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
fbrosson: Sorry for this late reply. I guess you already found the answer. Here it is, for the other readers who might not yet know the answer.
We don't need the docs on secondary arch builds because the recipe builds fine on *all* primary architectures.

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+REVISION="1"
+SOURCE_URI="http://download.savannah.nongnu.org/releases/libpipeline/libpipeline-$portVersion.tar.gz"
+CHECKSUM_SHA256="0d72e12e4f2afff67fd7b9df0a24d7ba42b5a7c9211ac5b3dcccc5cd8b286f2b"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	libpipeline$secondaryArchSuffix = $portVersion
+	lib:libpipeline$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
+libVersion=$portVersion
+libVersionCompat="$libVersion compat >= 1.5"
+
 PROVIDES="
 	libpipeline$secondaryArchSuffix = $portVersion
-	lib:libpipeline$secondaryArchSuffix = $portVersion
+	lib:libpipeline$secondaryArchSuffix = $libVersionCompat
	"
```

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libpipeline${secondaryArchSuffix}_devel = $portVersion
+	devel:libpipeline$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES_devel="
 	libpipeline${secondaryArchSuffix}_devel = $portVersion
-	devel:libpipeline$secondaryArchSuffix = $portVersion
+	devel:libpipeline$secondaryArchSuffix = $libVersionCompat
	"
```

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libpipeline$secondaryArchSuffix \
+	$libDir/libpipeline.so.$portVersion
```
fbrosson:
```diff
 defineDebugInfoPackage libpipeline$secondaryArchSuffix \
-	$libDir/libpipeline.so.$portVersion
+	$libDir/libpipeline.so.$libVersion
```

dev-libs/libpipeline/libpipeline-1.5.0.recipe
Begasus: I've seen this passing by with a few others, what does the $libVersionCompat provide? :)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
Begasus: And why $libVersion here and not $libVersionCompat?

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+
+INSTALL()
+{
+	make install
+
+	#remove libtool file
+	rm $libDir/*.la
+
+	# prepare develop/lib
+	prepareInstalledDevelLibs libpipeline
```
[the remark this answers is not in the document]
Begasus: Right on that one, missed it :+1:

dev-libs/libpipeline/libpipeline-1.5.0.recipe
fbrosson: On recipes that provide one or more libraries which all have the same soname, adding these 2 custom variables (libVersion and libVersionCompat) can be handy because:
* we only define libVersion once, and we use it in the `lib:lib*` line(s) of PROVIDES and in the  `devel:lib*` line(s) of PROVIDES_devel.
* we define libVersionCompat right after libVersion, with the "compat" stuff, and we can use it in the defineDebugInfoPackage call.
* it simplifies future upgrades: no risk to update the version on a `lib:lib*` line and forget the `devel:lib*` (or the opposite).
* it helps new contributors to understand the difference between the version of the recipe and that of its libraries.

Of course, using these custom variables is just a possibility. I don't expect it to become a rule. It is much better to let each contributor decide if that's a good idea or not.
(Well, I think it is ;-)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
fbrosson: Because after the very first argument, defineDebugInfoPackage expects filenames. The "compat" stuff is just for the `PROVIDES{,_devel}`.

dev-libs/libpipeline/libpipeline-1.5.0.recipe
fbrosson: **Important:** I don't know if it is appropriate to have a compat for `devel:lib*` lines.
Well, even if it were only appropriate for `lib:lib*` lines it would still be nice, imo, to define libVersionCompat right after libVersion.

dev-libs/libpipeline/libpipeline-1.5.0.recipe
Begasus: Ah, you didn't mention "providing libVersion" yet ;) probably why the build failed earlier :)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
Begasus: libVersionCompat isn't used at the moment by any other recipe?

Begasus:
updated with libVersion and libVersionCompat, let me know if this would be ok ;)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
fbrosson: I recently added lib**V**ersion to openssl and later another contributor added lib**v**ersion to another recipe.
But for libVersionCompat, libpipeline would be the first recipe to have it ;-)
So maybe we should wait until some key contributors say if they agree with these new custom variables.

Begasus: Ok, added it, will wait for replies :)

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,86 @@
...
+	2003-2017 Colin Watson"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="http://download.savannah.nongnu.org/releases/libpipeline/libpipeline-$portVersion.tar.gz"
+CHECKSUM_SHA256="0d72e12e4f2afff67fd7b9df0a24d7ba42b5a7c9211ac5b3dcccc5cd8b286f2b"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion=1.5.0 #could be used with $portVersion also?
```
Begasus: I'm guessing using $portVersion would be obsolete here :)

fbrosson: Well, I'd still use
```sh
libVersion=$portVersion
```

fbrosson: Since it is likely that newer releases will also have libVersion == portVersion, I would recommend to define it this way:
```sh
libVersion=$portVersion
```
In case this becomes wrong in a future release, `haikuporter --strict-policy libpipeline` will detect it thanks to the defineDebugInfoPackage call. So it is safe, for now, to assume that libVersion == portVersion.
If haikuporter fails to build the debuginfo package, then we'll just need to switch to a numerical value (e.g. 1.5.0)  for libVersion.

dev-libs/libpipeline/libpipeline-1.5.0.recipe
```diff
@@ -0,0 +1,86 @@
...
+	libpipeline$secondaryArchSuffix = $portVersion
+	lib:libpipeline$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libpipeline${secondaryArchSuffix}_devel = $portVersion
+	devel:libpipeline$secondaryArchSuffix = $libVersionCompat
```
fbrosson: In a recent investigation I did for another PR, it seems we have, for PROVIDES_devel, around twice more recipes with just a version and no compat expression than recipes with both the version and the compat expression.
So it might be better to drop the compat here, in PROVIDES_devel:
```diff
-	lib:libpipeline$secondaryArchSuffix = $libVersionCompat
+	lib:libpipeline$secondaryArchSuffix = $libVersion
```

Begasus: Updated with the recent comments :)

Begasus:
Thanks! :+1:

PR #1830, opened 2017-12-03
PugiXML: new recipe

extrowerk:

PR #1831, opened 2017-12-03
Oiio pugi fix

extrowerk:
Depends on https://github.com/haikuports/haikuports/pull/1830


KapiX:
Can you add `defineDebugInfoPackage` while you're at it?

PR #1832, opened 2017-12-03
Add TEST() to various ports

alaviss:
Done as part of GCI 2017

This PR add `TEST()` to various ports, with fixes to a few ports I bumped into while running `TEST()`

# Status
## Tested
- [x] app-crypt/qca (PASS)
- [x] app-crypt/rhash (PASS)
- [x] app-text/podofo (PASS)
- [x] app-editors/retext (PASS)
- [x] app-i18n/mozc (FAIL, see below)
- [x] dev-lang/spidermonkey (FAIL, see below)
- [x] dev-lang/tcc (FAIL, see below)
- [x] dev-lang/yasm (PASS)
- [x] dev-libs/glib2 (FAIL, see below)
- [x] dev-libs/jsoncpp (PASS)
- [x] dev-libs/libtommath (PASS)
- [x] dev-libs/libxslt (PASS)
- [x] dev-libs/libyajl (PASS)
- [x] dev-libs/zziplib (PASS)
- [x] sys-devel/flex (PASS)

## Known issues
- app-editors/retext: tests operate on desktop, users should leave their device unattended during test
- app-i18n/mozc: most passes, following fails: `ipc_test`, `prediction_test`, `renderer_server_test`, `session_server_test`. [Log file](https://ghostbin.com/paste/xwsw8)
- dev-lang/spidermonkey: multiple tests failed, all `SIGSEGV`
- dev-lang/tcc: currently broken on Haiku due to [bug#13847](https://dev.haiku-os.org/ticket/13847)
- dev-libs/glib2:
  - tests won't build on `x86_gcc2`, `gcc2` lacks `-Wstrict-aliasing=2` 
  - most passes, some fails: [log](https://ghostbin.com/paste/47uu9)

### DIFF
--- a/app-crypt/qca/qca-2.1.3.recipe
+++ b/app-crypt/qca/qca-2.1.3.recipe
@@ -79,6 +79,7 @@ BUILD()
 		-DQCA_PRIVATE_INCLUDE_INSTALL_DIR=$includeDir \
 		-DQCA_DOC_INSTALL_DIR=$docDir \
 		-DQCA_MAN_INSTALL_DIR=$manDir \
+		-DBUILD_TESTS=1 \
 		-DPKGCONFIG_INSTALL_PREFIX=$developLibDir/pkgconfig
 
 	make $jobArgs
@@ -96,4 +97,9 @@ INSTALL()
 		$developDir \
 		$dataDir/Qt/mkspecs \
 		$libDir/cmake
+}
+
+TEST()
+{
+	QCA_PLUGIN_PATH="$sourceDir/lib/qca" make test
 }
--- a/app-crypt/rhash/rhash-1.3.5.recipe
+++ b/app-crypt/rhash/rhash-1.3.5.recipe
@@ -14,20 +14,20 @@ SOURCE_URI="https://downloads.sourceforge.net/rhash/rhash-$portVersion-src.tar.g
 CHECKSUM_SHA256="98e0688acae29e68c298ffbcdbb0f838864105f9b2bd8857980664435b1f1f2e"
 SOURCE_DIR="RHash-$portVersion"
 
-ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="?x86"
 
 PROVIDES="
 	rhash$secondaryArchSuffix = $portVersion
-	cmd:ed2k_link
-	cmd:gost_hash
-	cmd:has160_hash
-	cmd:magnet_link
-	cmd:rhash
-	cmd:sfv_hash
-	cmd:tiger_hash
-	cmd:tth_hash
-	cmd:whirlpool_hash
+	cmd:ed2k_link$secondaryArchSuffix
+	cmd:gost_hash$secondaryArchSuffix
+	cmd:has160_hash$secondaryArchSuffix
+	cmd:magnet_link$secondaryArchSuffix
+	cmd:rhash$secondaryArchSuffix
+	cmd:sfv_hash$secondaryArchSuffix
+	cmd:tiger_hash$secondaryArchSuffix
+	cmd:tth_hash$secondaryArchSuffix
+	cmd:whirlpool_hash$secondaryArchSuffix
 	lib:librhash$secondaryArchSuffix = 0
 	"
 REQUIRES="
@@ -44,7 +44,7 @@ REQUIRES_devel="
 	"
 
 BUILD_REQUIRES="
-	haiku_devel$secondaryArchSuffix
+	haiku${secondaryArchSuffix}_devel
 	"
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
--- a/app-crypt/rhash/rhash-1.3.5.recipe
+++ b/app-crypt/rhash/rhash-1.3.5.recipe
@@ -70,3 +70,10 @@ INSTALL()
 	packageEntries devel \
 		$developDir
 }
+
+TEST()
+{
+	cd tests
+	LIBRARY_PATH="$sourceDir/librhash${LIBRARY_PATH:+:$LIBRARY_PATH}" \
+		./test_rhash.sh --full ../rhash_shared
+}
--- a/app-text/podofo/podofo-0.9.5.recipe
+++ b/app-text/podofo/podofo-0.9.5.recipe
@@ -67,6 +67,7 @@ REQUIRES_devel="
 	"
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	devel:libcppunit$secondaryArchSuffix
 	devel:libfontconfig$secondaryArchSuffix
 	devel:libfreetype$secondaryArchSuffix
 	devel:libjpeg$secondaryArchSuffix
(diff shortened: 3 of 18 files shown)


dev-libs/zziplib/patches/zziplib-0.13.67.patchset
```diff
@@ -30,3 +30,73 @@ index 69070df..410487a 100644
...
+diff --git a/test/zziptests.py b/test/zziptests.py
+index 6e9df78..0f73300 100644
+--- a/test/zziptests.py
++++ b/test/zziptests.py
+@@ -199,7 +199,7 @@ class ZZipTest(unittest.TestCase):
+     filename = os.path.join(tmpdir,"README")
+     filetext = self.readme()
+     self.mkfile(filename, filetext)
+-    shell("{exe} -n README ../{zipfile} ??*.* README".format(**locals()), cwd=tmpdir)
++    shell("for f in ??*.*; do {exe} -un README ../{zipfile} $f README; done".format(**locals()), cwd=tmpdir)
```
pulkomandy: I think xargs could be used to make this less slow (it can pass the arguments by groups of arbitrary size to another process).

app-crypt/rhash/rhash-1.3.5.recipe
```diff
@@ -70,3 +70,10 @@ INSTALL()
 	packageEntries devel \
 		$developDir
 }
+
+TEST()
+{
+	cd tests
+	LIBRARY_PATH="$(pwd)/../librhash${LIBRARY_PATH:+:$LIBRARY_PATH}" \
```
fbrosson: You might wish to replace `$(pwd)/../` by `$sourceDir/`
That would be equivalent, since the working directory where `TEST()` starts is $sourceDir (which is the root directory of the tarball).

BTW, like @pulkomandy wrote, you did a really great work!

mmuman:
Hmm, which architecture are you building on? Because on gcc2h, spidermonkey fails to build both on x86_gcc2 and x86 as secondary arch (it misses passing libDir to configure and other things).

dev-lang/spidermonkey/spidermonkey-1.8.5.recipe
```diff
@@ -41,13 +42,14 @@ BUILD_PREREQUIRES="
 	cmd:xargs
 	cmd:pkg_config$secondaryArchSuffix
 	cmd:sed
+	cmd:zip
 	"
 
 BUILD()
 {
 	cd js/src
 	runConfigure --omit-dirs "libDir docDir dataRootDir" ./configure \
```
mmuman: having libDir used correctly is required for secondary arch support, currently INSTALL is trying to find something in lib/x86 which wasn't installed...

ohnx:
I also could not manage to make spidermonkey build on x86_64.

alaviss:
Updates:
- Squashed most tcc commits
- update rhash test following @fbrosson suggestions
- update zziplib test to use `xargs` as suggested by @pulkomandy 
- make spidermonkey works on x86_64, also fixed secondary arch build
- tested app-editors/retext

Todo:
- test mozc

scottmc:
For the ones that are failing the their TEST(), please open a new issue for each of those so that we can get them investigated further and resolved.  Nice progress here.

alaviss:
Updates:
- app-i18n/mozc tested

alaviss:
Updates:
- dev-lang/spidermonkey: finally fixed x86_64 build
- squashed some miscellaneous commits

fbrosson:
@alaviss, this is wonderful!! I haven't checked your commits yet, but I'm pretty sure everything is OK.

Questions to you and to all members:

Question 1:
Shall we merge this PR? If we choose to check everything before we merge, then there are chances that one or more recipes will be changed before we have the opportunity to check everything.
This is why I think we should merge. @alaviss has shown that he knows what he does. I'll review the changes without building, but I'm very confident that there are no errors. In case one of his commits were incorrect, we could always fix it later.

Question 2:
We have 2 merge methods:
* "Squash and merge": this is what we use for most PRs; it allows the member doing the merge to edit the commit summary. This method also automatically adds the reference to the PR being merged.
* "Rebase and merge": This method merges individually each commit in the PR after rebasing each one in turn. It does not add any reference to the PR. This is what I used to merge PR #1865 in 73de1ce3e17b811f866eeba7b1e5035e4ade966e.

I think the most appropriate merge method, for this PR, is the "Rebase and merge" because it will allow to have separate commits. @alaviss and all other members: what do you think would be most appropriate?

@alaviss, in case there is a consensus for the "Rebase and merge" method, your commit messages will appear exactly as you wrote them, and no reference to this PR will be added. Hence this question: if you would like each of your commits to include a refernce to this PR, you are the only one who can do that. So it's up to you. (For me that's perfectly ok to not have that reference.)

app-text/podofo/podofo-0.9.5.recipe
```diff
@@ -81,6 +81,7 @@ BUILD_PREREQUIRES="
 	cmd:cmake
 	cmd:make
 	cmd:sed
+	devel:libcppunit$secondaryArchSuffix
```
fbrosson: I might be missing something, but I'm quite sure `devel:libcppunit$secondaryArchSuffix` should be in BUILD_REQUIRES, not in BUILD_**PRE**REQUIRES.

app-crypt/qca/qca-2.1.3.recipe
```diff
@@ -97,3 +98,8 @@ INSTALL()
 		$dataDir/Qt/mkspecs \
 		$libDir/cmake
 }
+
+TEST()
+{
+	QCA_PLUGIN_PATH="$(pwd)/lib/qca" make test
```
fbrosson: What do you think of this variant?
```diff
-	QCA_PLUGIN_PATH="$(pwd)/lib/qca" make test
+	QCA_PLUGIN_PATH="$sourceDir/lib/qca" make test
```

fbrosson:
Hmm, if we choose the "Rebase and merge", it might be worth squashing some of the commits by category: "Add TEST()", "enable x86_64", etc..
If we choose "Squash and merge", of course, there is no need to squash anything before the merge.

dev-lang/spidermonkey/spidermonkey-1.8.5.recipe
```diff
@@ -34,19 +36,29 @@ BUILD_REQUIRES="
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize
```
fbrosson:
```diff
-	cmd:libtoolize
+	cmd:libtoolize$secondaryArchSuffix
```

alaviss: `libtoolize` should actually be architecture-independent (it's a bash script). But since it's packaged with libltdl...

alaviss:
Both merge methods eat away my GPG signature... So I'm fine with both methods

alaviss:
Updates:
- cosmetics changes by @fbrosson

korli:
FWIW It would have been nice to NOT squash commits for this PR.

fbrosson:
OK. Sorry @korli about that. As nobody, from appart @alaviss) dared to reply the [comment I wrote 2 days ago](#issuecomment-350495569) I though it was better to squash everything because most changes were just for the adding a TEST() function.
What we should have done was group all the 15 commits about the added TEST() together, and keep the various other changes in separate commits.
But given the really big work @alaviss did, and given that nobody had replied to my questions, I though that asking for a partial squash was not really needed.

Well, I'll avoid using "Squash and merge" in the future, on similar cases.

PR #1834, opened 2017-12-03
mbedtls: add initial recipe for 2.6.0

ohnx:
mbed TLS (formerly known as PolarSSL) makes it trivially easy for developers to include cryptographic and SSL/TLS capabilities in their (embedded) products, facilitating this functionality with a minimal coding footprint.

### DIFF
--- a/net-libs/mbedtls/mbedtls-2.6.0.recipe
+++ b/net-libs/mbedtls/mbedtls-2.6.0.recipe
@@ -1,0 +1,119 @@
+SUMMARY="Open source, portable, easy to use, readable and flexible SSL library"
+DESCRIPTION="mbed TLS (formerly known as PolarSSL) makes it trivially easy for \
+developers to include cryptographic and SSL/TLS capabilities in their (embedded) \
+products, facilitating this functionality with a minimal coding footprint."
+HOMEPAGE="https://tls.mbed.org/"
+COPYRIGHT="2006-2016, ARM Limited"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="https://tls.mbed.org/download/mbedtls-$portVersion-apache.tgz"
+CHECKSUM_SHA256="99bc9d4212d3d885eeb96273bcde8ecc649a481404b8d7ea7bb26397c9909687"
+
+ARCHITECTURES="x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	mbedtls = $portVersion
+	cmd:mbedtls_aescrypt2 = $portVersion
+	cmd:mbedtls_benchmark = $portVersion
+	cmd:mbedtls_cert_app = $portVersion
+	cmd:mbedtls_cert_req = $portVersion
+	cmd:mbedtls_cert_write = $portVersion
+	cmd:mbedtls_crl_app = $portVersion
+	cmd:mbedtls_crypt_and_hash = $portVersion
+	cmd:mbedtls_dh_client = $portVersion
+	cmd:mbedtls_dh_genprime = $portVersion
+	cmd:mbedtls_dh_server = $portVersion
+	cmd:mbedtls_dtls_client = $portVersion
+	cmd:mbedtls_dtls_server = $portVersion
+	cmd:mbedtls_ecdh_curve25519 = $portVersion
+	cmd:mbedtls_ecdsa = $portVersion
+	cmd:mbedtls_gen_entropy = $portVersion
+	cmd:mbedtls_gen_key = $portVersion
+	cmd:mbedtls_gen_random_ctr_drbg = $portVersion
+	cmd:mbedtls_gen_random_havege = $portVersion
+	cmd:mbedtls_generic_sum = $portVersion
+	cmd:mbedtls_hello = $portVersion
+	cmd:mbedtls_key_app = $portVersion
+	cmd:mbedtls_key_app_writer = $portVersion
+	cmd:mbedtls_mini_client = $portVersion
+	cmd:mbedtls_mpi_demo = $portVersion
+	cmd:mbedtls_pem2der = $portVersion
+	cmd:mbedtls_pk_decrypt = $portVersion
+	cmd:mbedtls_pk_encrypt = $portVersion
+	cmd:mbedtls_pk_sign = $portVersion
+	cmd:mbedtls_pk_verify = $portVersion
+	cmd:mbedtls_req_app = $portVersion
+	cmd:mbedtls_rsa_decrypt = $portVersion
+	cmd:mbedtls_rsa_encrypt = $portVersion
+	cmd:mbedtls_rsa_genkey = $portVersion
+	cmd:mbedtls_rsa_sign = $portVersion
+	cmd:mbedtls_rsa_sign_pss = $portVersion
+	cmd:mbedtls_rsa_verify = $portVersion
+	cmd:mbedtls_rsa_verify_pss = $portVersion
+	cmd:mbedtls_selftest = $portVersion
+	cmd:mbedtls_ssl_cert_test = $portVersion
+	cmd:mbedtls_ssl_client1 = $portVersion
+	cmd:mbedtls_ssl_client2 = $portVersion
+	cmd:mbedtls_ssl_fork_server = $portVersion
+	cmd:mbedtls_ssl_mail_client = $portVersion
+	cmd:mbedtls_ssl_server = $portVersion
+	cmd:mbedtls_ssl_server2 = $portVersion
+	cmd:mbedtls_strerror = $portVersion
+	cmd:mbedtls_udp_proxy = $portVersion
+	lib:libmbedcrypto = 0 compat >=0
+	lib:libmbedtls = 10 compat >= 10
+	lib:libmbedx509 = 0 compat >= 0
+	"
+REQUIRES="
+	haiku
+	"
+
+PROVIDES_devel="
+	mbedtls_devel = $portVersion
+	devel:libmbedcrypto = 0 compat >=0
+	devel:libmbedtls = 10 compat >= 10
+	devel:libmbedx509 = 0 compat >=0
+	"
+REQUIRES_devel="
+	mbedtls == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	cmd:perl
+	cmd:sed
+	"
+
+PATCH()
+{
+	sed -i \
+		-e "s|/include/mbedtls|/$relativeOldIncludeDir|;"\
+		-e "s|/include|/$relativeOldIncludeDir|;"\
+		-e "s|/bin/|/$relativeBinDir/|;"\
+		-e "s|/bin\$|/$relativeBinDir/|;"\
+		-e "s|/lib\$|/$relativeLibDir/|;"\
+		Makefile
+}
+
+BUILD()
+{
+	# Haiku is unix-like and network functions are in libnetwork
+	make CFLAGS="-Dunix" LDFLAGS="-lnetwork" SHARED=yes WARNING_CFLAGS=
+}
+
+INSTALL()
+{
+	make install SHARED=yes DESTDIR=$prefix
+	prepareInstalledDevelLibs libmbedcrypto libmbedtls libmbedx509
+	packageEntries devel $developDir
+}
+
+TEST()
+{
+	LIBRARY_PATH="$sourceDir/library:$LIBRARY_PATH" make check
+}
--- a/net-libs/mbedtls/mbedtls-2.6.0.recipe
+++ b/net-libs/mbedtls/mbedtls-2.6.0.recipe
@@ -14,7 +14,7 @@ SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	mbedtls$secondaryArchSuffix = $portVersion
-	lib:libmbedcrypto$secondaryArchSuffix = 0 compat >=0
+	lib:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
 	lib:libmbedtls$secondaryArchSuffix = 10 compat >= 10
 	lib:libmbedx509$secondaryArchSuffix = 0 compat >= 0
 	"
@@ -75,9 +75,9 @@ REQUIRES="
 
 PROVIDES_devel="
 	mbedtls${secondaryArchSuffix}_devel = $portVersion
-	devel:libmbedcrypto$secondaryArchSuffix = 0 compat >=0
+	devel:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
 	devel:libmbedtls$secondaryArchSuffix = 10 compat >= 10
-	devel:libmbedx509$secondaryArchSuffix = 0 compat >=0
+	devel:libmbedx509$secondaryArchSuffix = 0 compat >= 0
 	"
 REQUIRES_devel="
 	mbedtls$secondaryArchSuffix == $portVersion base
@@ -96,18 +96,31 @@ BUILD_PREREQUIRES="
 PATCH()
 {
 	sed -i \
-		-e "s|/include/mbedtls|/$relativeOldIncludeDir|;"\
-		-e "s|/include|/$relativeOldIncludeDir|;"\
-		-e "s|/bin/|/$relativeBinDir/|;"\
-		-e "s|/bin\$|/$relativeBinDir/|;"\
-		-e "s|/lib\$|/$relativeLibDir/|;"\
+		-e "s|/include/mbedtls|/$relativeOldIncludeDir|;" \
+		-e "s|/include|/$relativeOldIncludeDir|;" \
+		-e "s|/bin/|/$relativeBinDir/|;" \
+		-e "s|/bin\$|/$relativeBinDir/|;" \
+		-e "s|/lib\$|/$relativeLibDir/|;" \
 		Makefile
 }
+
+if [ -n "$secondaryArchSuffix" ]; then
+	maybe_binDir_mbedtls_runtimes=
+else
+	maybe_binDir_mbedtls_runtimes="`echo "$PROVIDES" | sed -n \
+		-e "s|cmd:\([^ ]*\).*|$binDir/\1|p" \
+		`"
+fi
+defineDebugInfoPackage mbedtls$secondaryArchSuffix \
+	$libDir/libmbedcrypto.so.0 \
+	$libDir/libmbedtls.so.10 \
+	$libDir/libmbedx509.so.0 \
+	$maybe_binDir_mbedtls_runtimes
 
 BUILD()
 {
 	# Haiku is unix-like and network functions are in libnetwork
-	make CFLAGS="-Dunix" LDFLAGS="-lnetwork" SHARED=yes WARNING_CFLAGS=
+	make CFLAGS="-Dunix" LDFLAGS="-lnetwork" SHARED=yes WARNING_CFLAGS=""
 }
 
 INSTALL()
--- a/net-libs/mbedtls/mbedtls-2.6.0.recipe
+++ b/net-libs/mbedtls/mbedtls-2.6.0.recipe
@@ -93,17 +93,6 @@ BUILD_PREREQUIRES="
 	cmd:sed
 	"
 
-PATCH()
-{
-	sed -i \
-		-e "s|/include/mbedtls|/$relativeOldIncludeDir|;" \
-		-e "s|/include|/$relativeOldIncludeDir|;" \
-		-e "s|/bin/|/$relativeBinDir/|;" \
-		-e "s|/bin\$|/$relativeBinDir/|;" \
-		-e "s|/lib\$|/$relativeLibDir/|;" \
-		Makefile
-}
-
 if [ -n "$secondaryArchSuffix" ]; then
 	maybe_binDir_mbedtls_runtimes=
 else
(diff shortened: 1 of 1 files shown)


ohnx:
@fbrosson changes made, as requested :)

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
...
+	cmd:mbedtls_ssl_client2 = $portVersion
+	cmd:mbedtls_ssl_fork_server = $portVersion
+	cmd:mbedtls_ssl_mail_client = $portVersion
+	cmd:mbedtls_ssl_server = $portVersion
+	cmd:mbedtls_ssl_server2 = $portVersion
+	cmd:mbedtls_strerror = $portVersion
+	cmd:mbedtls_udp_proxy = $portVersion
+	lib:libmbedcrypto
+	lib:libmbedtls
+	lib:libmbedx509
```
fbrosson: Please replace the 3 lines that start with`lib:lib` by:
```
	lib:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
	lib:libmbedtls$secondaryArchSuffix = 10 compat >= 10
	lib:libmbedx509$secondaryArchSuffix = 0 compat >= 0
```

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
...
+	"
+REQUIRES="
+	haiku
+	"
+
+PROVIDES_devel="
+	mbedtls_devel = $portVersion
+	devel:libmbedcrypto
+	devel:libmbedtls
+	devel:libmbedx509
```
fbrosson: Please replace the 3 lines that start with`devel:lib` by:
```
	devel:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
	devel:libmbedtls$secondaryArchSuffix = 10 compat >= 10
	devel:libmbedx509$secondaryArchSuffix = 0 compat >= 0
```

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
...
+
+PATCH()
+{
+	sed -i \
+	-e "s|/include/mbedtls|/$relativeOldIncludeDir|;"\
+	-e "s|/include|/$relativeOldIncludeDir|;"\
+	-e "s|/bin/|/$relativeBinDir/|;"\
+	-e "s|/bin\$|/$relativeBinDir/|;"\
+	-e "s|/lib\$|$relativeLibDir/|;"\
+	Makefile
```
fbrosson: Please add one indentation level (i.e; one extra tab) for the 5 lines that start with "-e" and for the line with "Makefile".

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
...
+
+BUILD()
+{
+	# Haiku is unix-like and network functions are in libnetwork
+	make CFLAGS="-Dunix" LDFLAGS="-lnetwork" SHARED=yes WARNING_CFLAGS=
+}
+
+INSTALL()
+{
+	make install SHARED=yes DESTDIR=$prefix/
```
fbrosson: It worked for my without the / after $prefix.
Could you try removing the / ?

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
+SUMMARY="An open source, portable, easy to use, readable and flexible SSL library"
+DESCRIPTION="mbed TLS (formerly known as PolarSSL) makes it trivially easy for \
+developers to include cryptographic and SSL/TLS capabilities in their (embedded) \
+products, facilitating this functionality with a minimal coding footprint."
+HOMEPAGE="https://tls.mbed.org"
```
fbrosson: If you could add a / after the `.org` that would be cool ;-)
(Some browsers hide them, but real URIs have a trailing slash when there is no path after.)

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
+SUMMARY="An open source, portable, easy to use, readable and flexible SSL library"
```
fbrosson: The SUMMARY line is a bit long (>70 chars). What do you think of:
```
SUMMARY="A lightweight crypto and SSL/TLS library with command line tools"
```
Hmm, I'm not really satisfied with this either, so if anyone has a better suggestion ;-)

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
...
+INSTALL()
+{
+	make install SHARED=yes DESTDIR=$prefix/
+	prepareInstalledDevelLibs libmbedcrypto libmbedtls libmbedx509
+	packageEntries devel $developDir
+}
+
+TEST()
+{
+	make check
```
fbrosson: We need another trick in TEST:
```
	LIBRARY_PATH="$sourceDir/library:$LIBRARY_PATH" make check
```
(The test would have worked without this trick if we had not enabled the shared libs.)

BTW, on x86_64 all the tests passed :-)
```
PASSED (60 suites, 6109 tests run)
```
I'm going to switch to x86_gcc2 to see if they also pass there.

net-libs/mbedtls/mbedtls-2.6.0.recipe
ohnx: I tried, it tried writing to `/packages/mbedtls-2.6.0-1/.selflib` instead of `/packages/mbedtls-2.6.0-1/.self/lib`

ohnx: Perhaps we are running different haikuporter versions? I am on `969cac0bb71b...`

net-libs/mbedtls/mbedtls-2.6.0.recipe
ohnx: I should really spin up a VM to do this... Thanks for testing!

ohnx: Also, thanks for this catch; I only tested when using static libraries and didn't check when using shared libraries :p

net-libs/mbedtls/mbedtls-2.6.0.recipe
ohnx: If you remove `An `, it becomes 69 characters :)

net-libs/mbedtls/mbedtls-2.6.0.recipe
fbrosson: I'm running a quite recent hrev on both x86_64 and x86_gcc2, but your issue about the $prefix/ does not comme from the version of Haiku.
The default value does not have a trailing slash:
```
grep ^DESTDIR Makefile 
DESTDIR=/usr/local
```
If you trie to clean the work dir with
```
haikuporter -c mbedtls
```
and then try to build it again, do you still get the error?
I have an "hp" alias for haikuporter which is set to:
```
haikuporter -G --get-dependencies --no-source-packages
```
And when I call `hp -S mbedtls` I don't have any ossues.

ohnx: I have tried deleting my work directory, still nothing... I will try to delete everything and run `haikuporter -c mbedtls` and try again.

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,118 @@
...
+	"
+
+PATCH()
+{
+	sed -i \
+	-e "s|/include/mbedtls|/$relativeOldIncludeDir|;"\
+	-e "s|/include|/$relativeOldIncludeDir|;"\
+	-e "s|/bin/|/$relativeBinDir/|;"\
+	-e "s|/bin\$|/$relativeBinDir/|;"\
+	-e "s|/lib\$|$relativeLibDir/|;"\
```
ohnx: Aha! there is a typo here...

ohnx: @fbrosson

ohnx:
Revised and amended commit.

ohnx:
All tests pass on x86_gcc2.

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	cmd:mbedtls_ssl_client2 = $portVersion
+	cmd:mbedtls_ssl_fork_server = $portVersion
+	cmd:mbedtls_ssl_mail_client = $portVersion
+	cmd:mbedtls_ssl_server = $portVersion
+	cmd:mbedtls_ssl_server2 = $portVersion
+	cmd:mbedtls_strerror = $portVersion
+	cmd:mbedtls_udp_proxy = $portVersion
+	lib:libmbedcrypto = 0 compat >=0
+	lib:libmbedtls = 10 compat >= 10
+	lib:libmbedx509 = 0 compat >= 0
```
fbrosson: You forgot to add the $secondaryArchSuffix.
On some rare cases, we would do this:
```
PROVIDES="
	mbedtls$secondaryArchSuffix = $portVersion
	cmd:mbedtls_aescrypt2$secondaryArchSuffix = $portVersion
...
	cmd:mbedtls_udp_proxy$secondaryArchSuffix = $portVersion
	lib:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
	lib:libmbedtls$secondaryArchSuffix = 10 compat >= 10
	lib:libmbedx509$secondaryArchSuffix = 0 compat >= 0
	"
```

But since x86_gcc2 seems to work fine we can get rid of the cmd: on the x86 secondary arch build.
Have a look at ../nettle/.
Well, here's how we handle this:
```
PROVIDES="
	mbedtls$secondaryArchSuffix = $portVersion
	lib:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
	lib:libmbedtls$secondaryArchSuffix = 10 compat >= 10
	lib:libmbedx509$secondaryArchSuffix = 0 compat >= 0
	"
	PROVIDES="$PROVIDES
		cmd:mbedtls_aescrypt2 = $portVersion
		cmd:mbedtls_benchmark = $portVersion
...
		cmd:mbedtls_strerror = $portVersion
		cmd:mbedtls_udp_proxy = $portVersion
		"
fi
```
Of couse, this needs a small change in INSTALL

Hmm, since we are going to `rm -rf $binDir` for secondary arch builds, we could, instead, `rm -rf $prefix/bin` and hence skip 2 of the sed commands (the ones which replace /bin by /bin/x86).
But IMHO it's better to keep all the sed commands, even if we know we're going to remove the $binDir tree on secondary arch builds.

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	"
+REQUIRES="
+	haiku
```
fbrosson: We need to add a $secondaryArchSuffix after haiku in REQUIRES to enable support for secondary arch:
```
REQUIRES="
	haiku$secondaryArchSuffix
	"
```

ohnx:
@fbrosson Is everything good for a merge? ;)

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	"
+
+PROVIDES_devel="
+	mbedtls_devel = $portVersion
+	devel:libmbedcrypto = 0 compat >=0
+	devel:libmbedtls = 10 compat >= 10
+	devel:libmbedx509 = 0 compat >=0
```
fbrosson: PROVIDES_devel should also needs several $secondaryArchSuffix :
```
PROVIDES_devel="
	mbedtls${secondaryArchSuffix}_devel = $portVersion
	devel:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
	devel:libmbedtls$secondaryArchSuffix = 10 compat >= 10
	devel:libmbedx509$secondaryArchSuffix = 0 compat >= 0
	"
```

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	"
+REQUIRES_devel="
+	mbedtls == $portVersion base
```
fbrosson: REQUIRES_devel should be:
```
REQUIRES_devel="
	mbedtls$secondaryArchSuffix == $portVersion base
	"
```

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
```
fbrosson: BUILD_REQUIRES should be:
```
BUILD_REQUIRES="
	"
```

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
```
fbrosson: `cmd:gcc` also needs the suffix (but not the other cmd: lines)
```
	cmd:gcc$secondaryArchSuffix
```

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,119 @@
...
+	# Haiku is unix-like and network functions are in libnetwork
+	make CFLAGS="-Dunix" LDFLAGS="-lnetwork" SHARED=yes WARNING_CFLAGS=
+}
+
+INSTALL()
+{
+	make install SHARED=yes DESTDIR=$prefix
+	prepareInstalledDevelLibs libmbedcrypto libmbedtls libmbedx509
+	packageEntries devel $developDir
+}
```
fbrosson: This is what we should add (after the call to `packageEntries devel $developDir`) to get rid of the $binDir tree on secondary arch builds:
```
		rm -rf $binDir
	fi
```

fbrosson:
Thanks for your previous updates ;-)
After you make the changes in my last review the recipe will be OK for a merge.

Hmm, should we create a debuginfo package (by calling defineDebugInfoPackage) ?
That's not difficult, but I don't know what's the general recommendation.

ohnx:
What is the purpose of a debuginfo package?

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,126 @@
...
+		-e "s|/bin/|/$relativeBinDir/|;"\
+		-e "s|/bin\$|/$relativeBinDir/|;"\
+		-e "s|/lib\$|/$relativeLibDir/|;"\
+		Makefile
+}
+
+BUILD()
+{
+	# Haiku is unix-like and network functions are in libnetwork
+	make CFLAGS="-Dunix" LDFLAGS="-lnetwork" SHARED=yes WARNING_CFLAGS=
```
extrowerk: I would add "" at the end of this line

fbrosson: Yep, why not ;-)

fbrosson:
While I was searching for information about the debuginfo packages I got the answser to my previous question about whether or not we should add a such a debuginfo package.
[HaikuPorter-Guidelines#debuginfo-package](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#debuginfo-package)

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,126 @@
...
+REVISION="1"
+SOURCE_URI="https://tls.mbed.org/download/mbedtls-$portVersion-apache.tgz"
+CHECKSUM_SHA256="99bc9d4212d3d885eeb96273bcde8ecc649a481404b8d7ea7bb26397c9909687"
+
+ARCHITECTURES="x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	mbedtls$secondaryArchSuffix = $portVersion
+	lib:libmbedcrypto$secondaryArchSuffix = 0 compat >=0
```
fbrosson: Please add a white space between ">=" and "0".

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,126 @@
...
+fi
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	mbedtls${secondaryArchSuffix}_devel = $portVersion
+	devel:libmbedcrypto$secondaryArchSuffix = 0 compat >=0
+	devel:libmbedtls$secondaryArchSuffix = 10 compat >= 10
+	devel:libmbedx509$secondaryArchSuffix = 0 compat >=0
```
fbrosson: Please add a white space between ">=" and "0" (in line 78 and 80).

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,126 @@
...
+	"
+
+PATCH()
+{
+	sed -i \
+		-e "s|/include/mbedtls|/$relativeOldIncludeDir|;"\
+		-e "s|/include|/$relativeOldIncludeDir|;"\
+		-e "s|/bin/|/$relativeBinDir/|;"\
+		-e "s|/bin\$|/$relativeBinDir/|;"\
+		-e "s|/lib\$|/$relativeLibDir/|;"\
```
fbrosson: Please add a white space before the last `\` of each line.
Yes, I know this is not necessary because we have tabs at the beginning of each (next) line, but it would look nicer with the space ;-) THanls!

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,126 @@
...
+		Makefile
+}
+
+BUILD()
```
fbrosson: Here's what we can insert, before BUILD, to have the debuginfo packages:
```
	maybe_binDir_mbedtls_runtimes=
else
	maybe_binDir_mbedtls_runtimes="`echo "$PROVIDES" | sed -n \
		-e "s|cmd:\([^ ]*\).*|$binDir/\1|p" \
		`"
fi
defineDebugInfoPackage mbedtls$secondaryArchSuffix \
	$libDir/libmbedcrypto.so.0 \
	$libDir/libmbedtls.so.10 \
	$libDir/libmbedx509.so.0 \
	$maybe_binDir_mbedtls_runtimes

```
(Make sure you keep an empty line before BUILD.)

I'm giving the trick here because it is not easy to figure out how to handle the debuginfo package for all cases (i.e. primary vs secondary arch builds), especially with 47 runtimes in PROVIDES.
We usually list them explicitly, but it is much better to compute (from `$PROVIDES`) the arguments we need to pass to defineDebugInfoPackage.

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,139 @@
...
+	prepareInstalledDevelLibs libmbedcrypto libmbedtls libmbedx509
+	packageEntries devel $developDir
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $binDir
+	fi
+}
+
+TEST()
+{
+	LIBRARY_PATH="$sourceDir/library:$LIBRARY_PATH" make check
```
fbrosson: Thanks for you update!

If you have ed installed, could you please run
```
(echo 96,106 m 119; echo w; echo q) | ed mbedtls-2.6.0.recipe
```
It will fix a small error, due to a wrong indication I gave you in my previous review.
The result is that the PATCH fonction will be moved after the call to defineDebugInfoPackage.
It's my fault, sorry.

While at it, would you mind doing an extra cosmetic change? Could you please replace, in `TEST()`
```
	LIBRARY_PATH="$sourceDir/library:$LIBRARY_PATH" make check
```
by
```
	LIBRARY_PATH="$sourceDir/library${LIBRARY_PATH:+:$LIBRARY_PATH}" make check
```
Yep, this is really cosmetics. But it looks nicer ;-)
It does not make any difference when LIBRARY_PATH is not an empty string.
But if LIBRARY_PATH is empty, then it makes a tiny difference, since, during the call to "make check" LIBRARY_PATH won't end with a colon. That's cosmetics, but it will be cool ;-)
Thanks!

pulkomandy: I find that this makes the line less readable, so I'd rather avoid it.

fbrosson: I don't really care.
It was just to make sure no purist would be able to say "this is not 100% correct" ;-)

net-libs/mbedtls/mbedtls-2.6.0.recipe
fbrosson: All my apologies. I wrote "before BUILD" but I should have written "before PATCH".
The fix is easy to do with ed:
```
(echo 96,106 m 119; echo w; echo q) | ed mbedtls-2.6.0.recipe
```
I assumes the PATH function starts at line 96 and that the BUILD function is at line 120.
I will move the 11 lines from 96 to 106 (i.e. the bloc containing the PATCH function and the blank line after it) a few lines below, just before BUILD.

net-libs/mbedtls/mbedtls-2.6.0.recipe
```diff
@@ -0,0 +1,139 @@
...
+	"
+
+PATCH()
+{
+	sed -i \
+		-e "s|/include/mbedtls|/$relativeOldIncludeDir|;" \
+		-e "s|/include|/$relativeOldIncludeDir|;" \
+		-e "s|/bin/|/$relativeBinDir/|;" \
+		-e "s|/bin\$|/$relativeBinDir/|;" \
+		-e "s|/lib\$|/$relativeLibDir/|;" \
```
fbrosson: Oops, it would be better to drop the last / in the last two lines of the sed command:
```diff
-		-e "s|/bin\$|/$relativeBinDir/|;" \
+		-e "s|/bin\$|/$relativeBinDir|;" \
-		-e "s|/lib\$|/$relativeLibDir/|;" \
+		-e "s|/lib\$|/$relativeLibDir|;" \
```
(old lines in red, new ones in green, as usual)

ohnx:
Thanks :)

fbrosson:

PR #1857, opened 2017-12-06
[WIP] python-setuptools: bump version.

netpositive36:


scottmc:
Any idea why the checksum changed?  If there was no change to the revision of the source code then why does it have a different checksum than before?

extrowerk:
@scottmc: version bump.
The Revision should be 1.

korli:
setuptools newer versions might not be compatible. I would suggest to rebuild everything depending on it locally before pushing.

netpositive36:
@korli you mean all hp packages depending on the setuptools, right?

korli:
@netpositive36 yes

scottmc:
Task here is to rebuild all haikuports packages that use this and verify nothing gets broken

Begasus:
Last version (at this moment) is 40.6.3

waddlesplash:
Obsolete.

PR #1863, opened 2017-12-07
Alembic: new recipe

extrowerk:
Sorry guys, Haiku is Windows now.


media-libs/alembic/patches/alembic-1.7.4.patchset
```diff
@@ -0,0 +1,62 @@
...
++    /boot/system
+ )
+ 
+ FIND_PATH(ILMBASE_INCLUDE_DIR
+@@ -95,6 +96,9 @@ FIND_PATH(ILMBASE_INCLUDE_DIR
+   PATH_SUFFIXES
+     include
+     include/OpenEXR
++    develop/headers/OpenEXR
++    develop/headers/x86/OpenEXR
```
fbrosson: Hmm, that's not recommended. There are better ways to pass the includes dir.
That said, if this does not hurt anyone else, then I can sleep with it ;-)

PR #1865, opened 2017-12-08
Add sci-mathematics/msieve

alaviss:
Done as part of GCI 2017

Changes to msieve:
- Convert msieve's `printf` format specifiers to their architecture-independent counterpart
- Fixed a small pointer bug: `gzFile*` used instead of `gzFile`
- Add Haiku support:
  - Add `HAIKU` switch to Makefile: disable `-ldl` on set
  - Disable `ffpoly.c` inline assembly (they won't build, and @waddlesplash said they're not *that* optimized anyway)
  - Implemented `set_idle_priority()` for Haiku: set `B_IDLE_PRIORITY` for the calling team

### DIFF
--- a/sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
+++ b/sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
@@ -1,0 +1,77 @@
+SUMMARY="Elliptic Curve Method for Integer Factorization"
+DESCRIPTION="C implementation of the Elliptic Curve Method for factorizing integers"
+HOMEPAGE="http://ecm.gforge.inria.fr/"
+COPYRIGHT="2001-2016 GMP-ECM Authors"
+LICENSE="
+	GNU GPL v3
+	GNU LGPL v3
+	"
+REVISION="1"
+SOURCE_URI="https://gforge.inria.fr/frs/download.php/file/36224/ecm-$portVersion.tar.gz"
+CHECKSUM_SHA256="0cf7b3eee8462cc6f98b418b47630e1eb6b3f4f8c3fc1fb005b08e2a1811ba43"
+SOURCE_DIR="ecm-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion=1.0.0
+libVersionCompat="$libVersion compat >= 1"
+
+PROVIDES="
+	gmp_ecm$secondaryArchSuffix = $portVersion
+	cmd:ecm$secondaryArchSuffix = $portVersion
+	lib:libecm$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	gmp_ecm${secondaryArchSuffix}_devel = $portVersion
+	devel:libecm$secondaryArchSuffix = $libVersion
+	"
+REQUIRES_devel="
+	haiku${secondaryArchSuffix}_devel
+	gmp_ecm$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libgmp$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:cmp
+	cmd:gcc$secondaryArchSuffix
+	cmd:m4
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure ./configure --enable-shared --enable-openmp
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	install -t "$docDir" -m 644 AUTHORS NEWS README*
+
+	rm "$libDir"/*.la
+
+	prepareInstalledDevelLib libecm
+
+	packageEntries devel \
+		"$developDir" \
+		"$docDir/NEWS"
+}
+
+TEST()
+{
+	make check
+}
--- a/sci-mathematics/msieve/msieve-1.53.recipe
+++ b/sci-mathematics/msieve/msieve-1.53.recipe
@@ -1,0 +1,72 @@
+SUMMARY="Library for Factoring Large Integers"
+DESCRIPTION="C library implementing a suite of algorithms to factor large integers"
+HOMEPAGE="https://sourceforge.net/projects/msieve/"
+COPYRIGHT="Public Domain"
+LICENSE="Public Domain"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/msieve/msieve/Msieve%20v$portVersion/msieve${portVersion/./}_src.tar.gz"
+CHECKSUM_SHA256="c5fcbaaff266a43aa8bca55239d5b087d3e3f138d1a95d75b776c04ce4d93bb4"
+PATCHES="msieve-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	msieve$secondaryArchSuffix = $portVersion
+	cmd:msieve$secondaryArchSuffix
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libecm$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	msieve${secondaryArchSuffix}_devel = $portVersion
+	devel:libmsieve$secondaryArchSuffix
+	"
+REQUIRES_devel="
+	haiku${secondaryArchSuffix}_devel
+	msieve$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libecm$secondaryArchSuffix
+	devel:libgmp$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	"
+
+PATCH()
+{
+	# Allow owns CFLAGS, could be useful when doing debug builds
+	sed -e 's/^CFLAGS =/CFLAGS +=/' -i Makefile
+	# Remove -march=native, while this flag provides speedup,
+	# it's not suitable for binary distribution
+	sed -e 's/-march=native//' -i Makefile
+}
+
+BUILD()
+{
+	make $jobArgs all ECM=1 HAIKU=1
+}
+
+INSTALL()
+{
+	install -d -m 755 "$binDir" "$docDir" "$includeDir" "$developLibDir"
+
+	install -t "$binDir" -m 755 msieve
+	install -t "$docDir" -m 644 Changes Readme*
+	install -t "$includeDir" -m 644 include/*
+	install -t "$developLibDir" -m 644 libmsieve.a
+
+	packageEntries devel \
+		"$developDir" \
+		"$docDir/Changes"
+}
--- a/sci-mathematics/msieve/patches/msieve-1.53.patchset
+++ b/sci-mathematics/msieve/patches/msieve-1.53.patchset
@@ -1,0 +1,636 @@
+From f5992c6e1069e230098aed13d160dcf8f9b3b718 Mon Sep 17 00:00:00 2001
+From: Leorize <alaviss@users.noreply.github.com>
+Date: Thu, 7 Dec 2017 20:49:15 +0700
+Subject: use correct types
+
+
+diff --git a/common/driver.c b/common/driver.c
+index ffecfa9..d9344a5 100644
+--- a/common/driver.c
++++ b/common/driver.c
+@@ -174,7 +174,7 @@ void msieve_run(msieve_obj *obj) {
+ 
+ 	status = evaluate_expression(obj->input, &n);
+ 	if (status < 0 || mp_is_zero(&n)) {
+-		printf("error %d converting '%s'\n", status, obj->input);
++		printf("error %"PRId32" converting '%s'\n", status, obj->input);
+ 		obj->flags |= MSIEVE_FLAG_FACTORIZATION_DONE;
+ 		return;
+ 	}
+diff --git a/common/lanczos/lanczos.c b/common/lanczos/lanczos.c
+index b26185a..4852a4a 100644
+--- a/common/lanczos/lanczos.c
++++ b/common/lanczos/lanczos.c
+@@ -1220,9 +1220,9 @@ static uint64 * block_lanczos_core(msieve_obj *obj,
+ 				uint32 eta = elapsed * (max_n - dim_solved) /
+ 						(dim_solved - first_dim_solved);
+ 
+-				fprintf(stderr, "linear algebra completed %u "
+-					"of %u dimensions (%1.1f%%, ETA "
+-					"%dh%2dm)    \r",
++				fprintf(stderr, "linear algebra completed %" PRIu32 " "
++					"of %" PRIu32 " dimensions (%1.1f%%, ETA "
++					"%" PRId32 "h%2" PRId32 "m)    \r",
+ 					dim_solved, max_n, 100.0 * dim_solved / 
+ 					max_n, eta / 3600, (eta % 3600) / 60);
+ 
+diff --git a/common/smallfact/gmp_ecm.c b/common/smallfact/gmp_ecm.c
+index fd1da3b..c3df19d 100644
+--- a/common/smallfact/gmp_ecm.c
++++ b/common/smallfact/gmp_ecm.c
+@@ -306,14 +306,14 @@ uint32 ecm_pp1_pm1(msieve_obj *obj, mp_t *n, mp_t *reduced_n,
+ 
+ 			if (log_rate && j == next_log) {
+ 				next_log += log_rate;
+-				fprintf(stderr, "%u of %u curves\r",
++				fprintf(stderr, "%" PRIu32 " of %" PRIu32 " curves\r",
+ 						j, curr_work->num_ecm_trials);
+ 				fflush(stderr);
+ 			}
+ 		}
+ 
+ 		if (log_rate)
+-			fprintf(stderr, "\ncompleted %u ECM curves\n", j);
++			fprintf(stderr, "\ncompleted %" PRIu32 " ECM curves\n", j);
+ 	}
+ 
+ clean_up:
+diff --git a/demo.c b/demo.c
+index a882732..a49dbb2 100644
+--- a/demo.c
++++ b/demo.c
+@@ -256,7 +256,7 @@ void factor_integer(char *buf, uint32 flags,
+ 			else
+ 				factor_type = "prp";
+ 
+-			printf("%s%d: %s\n", factor_type, 
++			printf("%s%" PRId32 ": %s\n", factor_type, 
+ 					(int32)strlen(factor->number), 
+ 					factor->number);
+ 			factor = factor->next;
+diff --git a/gnfs/fb.c b/gnfs/fb.c
+index afaee24..15b33c2 100644
+--- a/gnfs/fb.c
++++ b/gnfs/fb.c
+@@ -100,7 +100,7 @@ void create_factor_base(msieve_obj *obj,
+ 		if (report_progress && (num_found_r > next_report ||
+ 					num_found_a > next_report)) {
+ 			fprintf(stderr, "factor base: found "
+-					"%u rational and %u "
++					"%" PRIu32 " rational and %" PRIu32 " "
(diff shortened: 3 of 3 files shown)


sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	gmp_ecm${secondaryArchSuffix}_devel = $portVersion
+	devel:libecm$secondaryArchSuffix = $portVersion
+	"
+REQUIRES_devel="
+	haiku$secondaryArchSuffix
+	lib:libecm$secondaryArchSuffix = $portVersion
```
fbrosson: Two little mistakes:
```diff
 REQUIRES_devel="
-	haiku$secondaryArchSuffix
+	haiku${secondaryArchSuffix}_devel
-	lib:libecm$secondaryArchSuffix = $portVersion
+	gmp_ecm$secondaryArchSuffix == $portVersion
	"
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+BUILD()
+{
+	runConfigure ./configure --enable-shared --enable-openmp
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	install -Dm644 AUTHORS "$docDir/AUTHORS"
```
fbrosson: I'm not sure this would be an improvement, but you might wish to change this to:
```
	install -d -m 755 "$docDir"
	install -t "$docDir" -m 644 AUTHORS
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+
+	rm "$libDir"/*.la
+
+	prepareInstalledDevelLibs libecm
```
fbrosson: It might be nice to use prepareInstalledDevelLib (instead of prepareInstalledDevelLib**s**) since we only have one lib in this recipe.

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+CHECKSUM_SHA256="0cf7b3eee8462cc6f98b418b47630e1eb6b3f4f8c3fc1fb005b08e2a1811ba43"
+SOURCE_DIR="ecm-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	gmp_ecm$secondaryArchSuffix = $portVersion
+	cmd:ecm$secondaryArchSuffix = $portVersion
+	lib:libecm$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
+libVersion=1.0.0
+libVersionCompat="$libVersion compat >= 1"
+
 PROVIDES="
 	gmp_ecm$secondaryArchSuffix = $portVersion
 	cmd:ecm$secondaryArchSuffix = $portVersion
-	lib:libecm$secondaryArchSuffix = $portVersion
+	lib:libecm$secondaryArchSuffix = $libVersionCompat
	"
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	gmp_ecm${secondaryArchSuffix}_devel = $portVersion
+	devel:libecm$secondaryArchSuffix = $portVersion
```
fbrosson: I think it should be
```
	devel:libecm$secondaryArchSuffix = $libVersionCompat
```
or maybe
```
	devel:libecm$secondaryArchSuffix = $libVersion
```
I don't know which is correct. @korli (or any other expert), could you help us here? Thanks!

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,63 @@
...
+{
+	make $jobArgs all ECM=1 HAIKU=1
+}
+
+INSTALL()
+{
+	install -Dm755 msieve "$binDir"/msieve
+	mkdir -p "$includeDir"
+	cp include/* "$includeDir"
+	install -Dm755 libmsieve.a "$developLibDir/libmsieve.a"
```
fbrosson: What do you think of the following variant?
```sh
INSTALL()
{
	install -d -m 755 "$binDir" "$docDir" "$includeDir" "$developLibDir"
	install -t "$binDir" -m 755 msieve
	install -t "$docDir" -m 644 Changes Readme*
	install -t "$includeDir" -m 644 include/* 
	install -t "$developLibDir" -m 644 libmsieve.a
	packageEntries devel "$developDir"
}
```
Yep, for `libmsieve.a` I guess you meant 644, not 755.

BTW, in PROVIDES_DEVEL, maybe we should remove the "` = $portVersion`" in:
```
	devel:libmsieve$secondaryArchSuffix = $portVersion
```

Hmm, if you feel like trying to build libmsieve as a shared lib, that could be nice.
I can try to help you with that, but I'm sure this would be a piece of cake for some mentors.

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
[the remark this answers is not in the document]
alaviss: @fbrosson suggests doing so

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
alaviss: I would prefer `$libVersion`. While minor bumps may not break ABI compatibility, they may introduce new features, and we would like to use them

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
alaviss: I saw some ports with this style as well, so maybe I will change them for consistency

sci-mathematics/msieve/msieve-1.53.recipe
alaviss: It could be done, but virtually no distribution does this, and not having a shared lib might means that the ABI/API can break at anytime, so we shouldn't do this.

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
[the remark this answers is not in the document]
fbrosson: Well, I had suggested this variant:
```sh
	install -d -m 755 "$docDir"
	install -t "$docDir" -m 644 AUTHORS
```
but I just noticed that gmp_ecm also has a few more files that we could include in $docDir: NEWS, README, README.gpu and README.lib. So maybe we could use this;
```sh
	install -d "$docDir"
	install -t "$docDir" AUTHORS NEWS README*
```
(Yes, we can even drop the `-m 755` and `-m 644` since the files already have the right permissions.)

Hmm, maybe we should move NEWS to the devel package, with:
```sh
	packageEntries devel \
		"$developDir" \
		"$docDir/NEWS"
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
fbrosson: Sorry for this late reply. I was doing some stats about how many existing recipes have compat for their `devel:lib*`lines in PROVIDES_devel and how many don't have, but I could not figure out if there is a best practice about this.
Anyway, since we have many with, and many without, please feel free to do as you think is best.

Hot (good) news! If my find+grep+sed based scripts are not too wrong, it seems we have twice more recipes *without* compat versions (in PROVIDES_devel) than we have *with* compat versions (in PROVIDES_devel).

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
fbrosson: This is an exception.
Although the general rule is to declare the dependencies on libs with `lib:` lines, the dependency of a package's devel sub-package on its own base package should always be specified using the name of its own base package.

In other words, if foo_devel needs bar_devel (i.e. another package providing e.g. libbar) then we add a `devel:libbar$secondaryArchSuffix` in foo's REQUIRES_devel (and a `lib:libbar$secondaryArchSuffix` in foo's REQUIRES).
**But** in foo's REQUIRES_devel we *must not* put `lib:foo$secondaryArchSuffix` *but* `foo$secondaryArchSuffix`, even if the devel package does not need any `cmd:` item from its base package.
This is to make sure the devel package will depend on its base package.

Hmm, speaking about the "base" keyword, I forgot to add it in me previous comment. (I'll fix this in a second.)

fbrosson: Adding this again to make sure nobody will merge this recipe before this gets fixed.
```diff
 REQUIRES_devel="
-	haiku$secondaryArchSuffix
+	haiku${secondaryArchSuffix}_devel
-	lib:libecm$secondaryArchSuffix = $portVersion
+	gmp_ecm$secondaryArchSuffix == $portVersion base
 	"
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libgmp$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:cmp
+	cmd:make
+	cmd:m4
+	cmd:gcc$secondaryArchSuffix
```
fbrosson: Please sort alphabetically the lines with gcc, m4 and make (in this order, of course ;-) Thanks!

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
fbrosson: This just a reminder, for an improved INSTALL to include a few more files in $docDir.
```sh
	install -d "$docDir"
	install -t "$docDir" AUTHORS NEWS README*
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+INSTALL()
+{
+	make install
+	install -Dm644 AUTHORS "$docDir/AUTHORS"
+
+	rm "$libDir"/*.la
+
+	prepareInstalledDevelLibs libecm
+
+	packageEntries devel "$developDir"
```
fbrosson: Small reminder: we should probably put NEWS in the devel package.
```sh
	packageEntries devel \
		"$developDir" \
		"$docDir/NEWS"
```

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,70 @@
...
+}
+
+TEST()
+{
+	make $jobArgs check
```
fbrosson: If you really meant "make $jobArgs check" then that's OK for me.
I'm writing this comment just in case you meant "make check".

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,63 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libecm$secondaryArchSuffix
+	devel:libgmp$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:gcc$secondaryArchSuffix
```
fbrosson: We like to sort alphabetically the lines in BUILD_PREREQUIRES. Thanks!

alaviss:
Updates:
- removed some unnecessary patches from msieve
- implemented @fbrosson suggestions

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
```diff
@@ -0,0 +1,77 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:cmp
+	cmd:gcc$secondaryArchSuffix
+	cmd:m4
+	cmd:make
+	"
+
+BUILD()
```
fbrosson: To add a debuginfo package we just need to insert a call to [definedebuginfopackage](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#definedebuginfopackage) after BUILD_PREREQUIRES (and before BUILD:
```diff
 	cmd:make
 	"
 
+defineDebugInfoPackage gmp_ecm$secondaryArchSuffix \
+	$libDir/libecm.so.$libVersion
+
 BUILD()
 {
```

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,66 @@
+SUMMARY="Library for Factoring Large Integers"
+DESCRIPTION="C library implementing a suite of algorithms to factor large integers"
+HOMEPAGE="https://sourceforge.net/projects/msieve/"
+COPYRIGHT="Public Domain"
+LICENSE="Public Domain"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/msieve/msieve/Msieve%20v$portVersion/msieve${portVersion/./}_src.tar.gz"
+CHECKSUM_SHA256="c5fcbaaff266a43aa8bca55239d5b087d3e3f138d1a95d75b776c04ce4d93bb4"
+PATCHES="msieve-1.53.patchset"
```
fbrosson:
```diff
-PATCHES="msieve-1.53.patchset"
+PATCHES="msieve-$portVersion.patchset"
```

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,66 @@
...
+	devel:libgmp$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	"
+
+PATCH()
```
fbrosson:
```diff
 	cmd:sed
 	"
 
+defineDebugInfoPackage msieve$secondaryArchSuffix \
+	$binDir/msieve
+
 PATCH()
 {
```

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,66 @@
...
+INSTALL()
+{
+	install -d -m 755 "$binDir" "$docDir" "$includeDir" "$developLibDir"
+
+	install -t "$binDir" -m 755 msieve
+	install -t "$docDir" -m 644 Changes Readme*
+	install -t "$includeDir" -m 644 include/*
+	install -t "$developLibDir" -m 644 libmsieve.a
+
+	packageEntries devel "$developDir"
```
fbrosson: Hmm, the files Changes would be better in the devel package, after all. What do you think?
```diff
-	packageEntries devel "$developDir"
+	packageEntries devel \
+		"$developDir" \
+		"$docDir"/Changes
```

alaviss: `msieve` is more or less a demo for `libmsieve` anyway (the source file is literally `demo.c`). So I thing pretty much every file in `$docDir` should end up in the devel package. What's your opinion on this?

sci-mathematics/msieve/msieve-1.53.recipe
alaviss: Does `defineDebugInfoPackage` magically add `-g` to `CFLAGS`? If not then I don't think we have any debuginfo to extract. Moreover, it should be done for the lib, if that's even possible with static objects...

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
alaviss: Same as for `msieve`

sci-mathematics/msieve/msieve-1.53.recipe
fbrosson: Although I do agree that the information in the Readme files will be useful mostly to developers, having these files in the base package is slightly better because it shows that we provide credits to the authors.
Indeed, moving the Readme files to the devel package might be seen as an attempt to hide the credits.
If I were one of the authors I would wish to have the Readme files in the base package.
Moreover, these files are not very big ;-)
(For the Changes file, that's another story, since this can be seen as meta-data.)

sci-mathematics/msieve/msieve-1.53.recipe
fbrosson: I'm quite sure it does not add "-g".
But I've seen many packages which create the debuginfo package and which do not compile with "-g".
BTW, I'm still wondering if having debuginfo packages makes any sense if we don't compile with "-g".
Anyway, even if it does not make sense, at least we make the Haiku Debugger happy, by providing a debuginfo package.

fbrosson: OK. If adding a debuginfo package does not make sense without the "-g", then let's forget about it.
No regrets ;-)

alaviss: Then I don't really agree with this. Those others packages should be reported and have "false" debuginfo removed. While Haiku Debugger might be happy, it's users won't. No one will expect a so-called "debuginfo" package to install, then figured out that they were merely placeholder.

I could add `-g` myself, but even so the debuginfo won't make sense, as `msieve` uses extensive inlining with `-O3`.
Compling `msieve` with less optimization is counter-intuitive, as the main point of science applications is to extract as much juice as possible from your hardware to solve a problem.

BTW, I think building package with debuginfo enabled should be a haikuporter feature, so that we could have optimized binaries for daily usage and upon request, debuginfo for development purposes

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,66 @@
...
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	msieve${secondaryArchSuffix}_devel = $portVersion
+	devel:libmsieve$secondaryArchSuffix
+	"
+REQUIRES_devel="
+	haiku${secondaryArchSuffix}_devel
+	"
```
fbrosson: Oops, REQUIRES_devel is missing a dependency on the base package.
```diff
 REQUIRES_devel="
	haiku${secondaryArchSuffix}_devel
+	msive$secondaryArchSuffix == $portVersion base
 	"
```

alaviss: Why? It doesn't even use anything from the base, or is this a haikuports practice?

fbrosson: Yes, you are right, but we allways add this dependency to make sure that:
1) if we uninstall the base package, then its devel package will also be uninstalled.
2) if we upgrade the base package to a new release, then it will force the devel package to also be upgraded.

sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
fbrosson: OK, no problem, let's forget about the debuginfo package for gmp_ecm too.

BTW, I'm sure Haiku needs more skilled developers like you!

alaviss:
Changes:
- Remove `-march=native` from opt flags
- Some suggestions from @fbrosson implemented

sci-mathematics/msieve/msieve-1.53.recipe
```diff
@@ -0,0 +1,72 @@
...
+	cmd:sed
+	"
+
+PATCH()
+{
+	# Allow owns CFLAGS, could be useful when doing debug builds
+	sed -e 's/^CFLAGS =/CFLAGS +=/' -i Makefile
+	# Remove -march=native, while this flag provides speedup,
+	# it's not suitable for binary distribution
+	sed -e 's/-march=native//' -i Makefile
```
fbrosson: Good catch, about the `-march=native`.

Hmm, since the 2 sed commands change the same file, it might be a good idea to change PATCH to:
```sh
PATCH()
{
	# - Allow our own CFLAGS; could be useful when doing debug builds.
	# - Remove -march=native; while this flag provides speedup,
	#   it's not suitable for binary distribution.
	sed -i \
		-e 's/^CFLAGS =/CFLAGS +=/' \
		-e 's/-march=native//' \
		Makefile
}
```
I don't know if this would be an improvement, but this variant looks a little bit more eye-friendly.
What do you think ?
Everything else is OK, so we could even merge as is if you prefer.

alaviss: I personally like having them separated for no reason other than I hate tabs :P. So let's just merge them as is

fbrosson: OK, that's fine too.

fbrosson:

PR #1878, opened 2017-12-10
media-libs/libfdk_aac: add port

alaviss:
Done as part of GCI 2017

### DIFF
--- a/media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
+++ b/media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
@@ -1,0 +1,72 @@
+SUMMARY="Fraunhofer AAC codec library"
+DESCRIPTION='The Fraunhofer FDK AAC Codec Library for Android ("FDK AAC Codec") \
+	is software that implements the MPEG Advanced Audio Coding ("AAC") \
+	encoding and decoding scheme for digital audio.'
+HOMEPAGE="https://sourceforge.net/projects/opencore-amr"
+COPYRIGHT="1995-2012 Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V."
+LICENSE="Fraunhofer FDK AAC Codec"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/opencore-amr/fdk-aac/fdk-aac-$portVersion.tar.gz"
+CHECKSUM_SHA256="2164592a67b467e5b20fdcdaf5bd4c50685199067391c6fcad4fa5521c9b4dd7"
+SOURCE_DIR="fdk-aac-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="1.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	libfdk_aac$secondaryArchSuffix = $portVersion
+	lib:libfdk_aac$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libfdk_aac${secondaryArchSuffix}_devel = $portVersion
+	devel:libfdk_aac$secondaryArchSuffix = $libVersion
+	"
+REQUIRES_devel="
+	haiku${secondaryArchSuffix}_devel
+	libfdk_aac$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:awk
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage libfdk_aac$secondaryArchSuffix \
+	$libDir/libfdk-aac.so.$libVersion
+
+BUILD()
+{
+	runConfigure ./configure --enable-shared
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	install -t "$docDir" -m 644 ChangeLog \
+		documentation/*.pdf
+
+	rm -f "$libDir"/*.la
+
+	fixPkgconfig
+	prepareInstalledDevelLib libfdk-aac
+	packageEntries devel \
+		"$developDir" \
+		"$docDir"/ChangeLog \
+		"$docDir"/*.pdf
+}
--- a/media-libs/libfdk_aac/licenses/Fraunhofer FDK AAC Codec
+++ b/media-libs/libfdk_aac/licenses/Fraunhofer FDK AAC Codec
@@ -1,0 +1,80 @@
+Software License for The Fraunhofer FDK AAC Codec Library for Android
+
+© Copyright  1995 - 2012 Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V.
+  All rights reserved.
+
+1.    INTRODUCTION
+The Fraunhofer FDK AAC Codec Library for Android ("FDK AAC Codec") is software that implements
+the MPEG Advanced Audio Coding ("AAC") encoding and decoding scheme for digital audio.
+This FDK AAC Codec software is intended to be used on a wide variety of Android devices.
+
+AAC's HE-AAC and HE-AAC v2 versions are regarded as today's most efficient general perceptual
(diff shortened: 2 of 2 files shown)


media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
```diff
@@ -0,0 +1,70 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	install -t "$docDir" -m 644 ChangeLog
+
+	rm -rf "$libDir"/*.la
```
fbrosson: No need for recursion here ;-)
```diff
-	rm -rf "$libDir"/*.la
+	rm -f "$libDir"/*.la
```

BTW, everything else looks ok to me too. Great job, @alaviss, once more ;-)
I'll try to build this recipe on x86_gcc2 for the x86 secondary arch.
So we might even be able to mark x86 as tested in SECONDARY_ARCHITECTURES.

fbrosson:
Please do not merge yet. I'm writting an additional review. Thanks!

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
```diff
@@ -0,0 +1,70 @@
...
+HOMEPAGE="https://sourceforge.net/projects/opencore-amr"
+COPYRIGHT="1995-2012 Fraunhofer-Gesellschaft zur Förderung der angewandten Forschung e.V."
+LICENSE="Fraunhofer FDK AAC Codec"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/opencore-amr/fdk-aac/fdk-aac-$portVersion.tar.gz"
+CHECKSUM_SHA256="2164592a67b467e5b20fdcdaf5bd4c50685199067391c6fcad4fa5521c9b4dd7"
+SOURCE_DIR="fdk-aac-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="?x86"
```
fbrosson: Shall we mark it as tested on x86 secondary arch? (It builds fine for me on that arch.)
```diff
-SECONDARY_ARCHITECTURES="?x86"
+SECONDARY_ARCHITECTURES="x86"
```

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
```diff
@@ -0,0 +1,70 @@
...
+	runConfigure ./configure --enable-shared
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	install -t "$docDir" -m 644 ChangeLog
```
fbrosson: I think we should include, in the main package,  the file called NOTICE. Not including it could be seen as an attempt to cheat with the license. Including it, right from the beginning, will help us keep this package in our repo in case upstream finds out we have made a package for it.
```diff
-	install -t "$docDir" -m 644 ChangeLog
+	install -t "$docDir" -m 644 ChangeLog NOTICE \
+		documentation/aacDecoder.pdf \
+		documentation/aacEncoder.pdf
```
Regarding the 2 pdf files, I don't know if this is desirable or not. This is just a suggestion.

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
```diff
@@ -0,0 +1,70 @@
...
+
+	rm -rf "$libDir"/*.la
+
+	fixPkgconfig
+	prepareInstalledDevelLib libfdk-aac
+	packageEntries devel \
+		"$developDir" \
+		"$docDir/ChangeLog"
```
fbrosson: In case we ship the 2 pdf files in the devel package, we need this tiny change:
```diff
 	packageEntries devel \
-		"$docDir/ChangeLog"
+		"$docDir"/ChangeLog \
+		"$docDir"/*.pdf
```


Shall we add this after INSTALL?
```

TEST()
{
# "check" target exists but does nothing
	make check
}
```

alaviss: No-op check is no better than not having it, so I would disagree with this. If one day someone actually add check to the library, then we can proceed adding `TEST()` for the port

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
alaviss: The license file which came with the port is the NOTICE file. Given that `haikuporter` package this file, I don't see the need to install another copy.

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
alaviss: I've already built it myself, but forgot to commit that change :P

alaviss:
Updates:
- Package the PDFs
- x86 secondary arch marked as tested (tks @fbrosson)

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
fbrosson: Good catch!

Hmm, if I had noticed this was the case then I would have suggested an extra command at the end of INSTALL:
```
	rmdir "$docDir" "$documentationDir/packages"
```
Without this line, the base package has an empty directory. Well this is not an issue. It's just that if someone chooses to install only the base package and goes to /system/documentation/packages/ there will be an empty directory (libfdk_aac_x86 on x86_gcc2, libfdk_aac on x86_64).

To check if a package has empty directories, you can use the "package list" or  "package list -p".
For libfdk_aac_x86 this is what we get:
```
package list -p /system/packages/libfdk_aac_x86-0.1.5-1-x86_gcc2.hpkg  | grep documentation
documentation
documentation/packages
documentation/packages/libfdk_aac_x86
```
That said, there is no need to change the recipe just to remove that empty directory. It can be done next time there is a "real" change, or in case we need to trigger a rebuild for some reason.

media-libs/libfdk_aac/libfdk_aac-0.1.5.recipe
fbrosson: Well, there isn't really a consensus about this.
Some would argue that adding such a TEST() function with a comment stating that it does nothing makes it easier to find those other recipes which might do not have a TEST() function but have real tests.
Maybe we could/should add a TRUST_TEST variable with the following possible values :
* `yes` to indicate that a test suite exists and, in case the tests fail, this should trgger something, e.g. don't push the binary package to a public repo.
* `no` to indicate that a test suite exists but its result should be ignored. Some recipes work fine even if their tests fail.
* `fake` to indicate that the test suite does nothing.

pulkomandy: If the upstream decides to have no tests, it is their problem, not ours.

So, from our side we can consider we did everything needed. If the software turns out not to work and lands in the package repo, then it's upstream problem and they will have to deal with fixing the bugs in a hurry and the bad reputation.

So, empty test function (when there is nothing better provided by upstream) is enough, and there is no need to track the difference on our side.

fbrosson: What do you call an empty test function? No TEST() function at all, or a TEST() function which calls something like "make check" which does nothing?

pulkomandy: one that calls a do-nothing operation (empty "make check", or "true", or "#no tests"). This way we clearly mark that we investigated the problem, and found nothing to test.

fbrosson: OK, so I understand that you agree with me that having a TEST() function, even if is useless, is better than having no TEST() function at all.

This is where a `TRUST_TEST=fake` parameter makes sense. We would know that a test suite exists but does nothing. With such a parameter we would make everyone happy: those contributors who don't like empty TEST() functions, and those who like to know that a recipe has already been investigated about its tests.
We don't even need to change haikuporter. Adding a `TRUST_TEST=fake` (or anything more appropriate) cannot harm, as it will be ignored by haikuporter.

As usual, this is just a suggestion. If this is not an improvement, then let's forget about this.
We currently have some recipes with empty/fake tests, and some (most?) of these have a comment about that. But if we wanted to query the list of recipes in that case, there would be no easy method. With my suggestion it would be easy.
Needless to say, this TRUST_TEST would not need to be mandatory. It would perfectly make sense to use `TRUST_TEST=fake` only on those recipes with empty/fake tests, and leave all the other recipes unchanged, i.e. without any TRUST_TEST.

alaviss: This discussion is more about a META issue with haikuports than this PR. I suggest we move this conversation into an issue

PR #1887, opened 2017-12-12
mlpack: new recipe

extrowerk:


extrowerk:
Depends on Armadillo: https://github.com/haikuports/haikuports/pull/1888

Issue #720, opened 2016-07-16
Add a recipe for AliveJournal


richienyhus:
BeOS code: https://github.com/grahams/alivejournal

fbrosson:
Done in #1890.

PR #1890, opened 2017-12-13
AliveJournal Recipe

bach5000:
@raefaldhia The recipe itself should probably work, but AliveJournal does not.
There's a syntax error ("Syntax error before '*'") in src/LJApplication/LJAccount/LJViewHistory/LJViewHistory.h at lines 37 and 38.
The link to AliveJournal repo: https://github.com/grahams/alivejournal


extrowerk:
So the problematic part according the error message:
https://github.com/grahams/alivejournal/blob/master/src/LJApplication/LJAccount/LJViewHistory/LJViewHistory.h#L37

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,45 @@
+SUMMARY="Client for LiveJournal online journalling community"
+DESCRIPTION="AliveJournal is a revamped version of Simon Huet's original client."
+HOMEPAGE="http://www.livejournal.com/~alivejournal/"
+COPYRIGHT="2000 Simon Huet-Sean Graham"
+LICENSE="GNU GPL v2"
+REVISION="2"
```
fbrosson:
```diff
-REVISION="2"
+REVISION="1"
```

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,45 @@
...
+
+BUILD()
+{
+	make OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	cp objects/AliveJournal $appsDir
```
fbrosson:
```diff
-	mkdir -p $appsDir
+	install -d $appsDir $docDir
-	cp objects/AliveJournal $appsDir
+	install -t $appsDir objects/AliveJournal
```

raefaldhia:
Hi, @bach5000 #include <PopUpMenu.h> to fix it.

Look like you'll face other problem in https://github.com/grahams/alivejournal/blob/master/src/LJApplication/LJAccount/LJViewHistory/LJViewHistory.h#L40

I'll try to find a workaround for it.

bach5000:
Thanks @raefaldhia! Do I just add the patch file to this pull request?

raefaldhia:
Yes, you should add the patch file to this PR.
but look like there's a bunch of similar problem within the code, make sure you're patched 'em all.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="Client for LiveJournal online journalling community"
+DESCRIPTION="AliveJournal is a revamped version of Simon Huet's original client."
+HOMEPAGE="http://www.livejournal.com/~alivejournal/"
```
fbrosson: Although this program is a fork of another program, you should use the fork's homepage:
```diff
-HOMEPAGE="http://www.livejournal.com/~alivejournal/"
+HOMEPAGE="https://github.com/grahams/alivejournal"
```
BTW, the fork's README already mentions the original's homepage.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,46 @@
...
+COPYRIGHT="2000 Simon Huet-Sean Graham"
```
fbrosson: You should split this to make it clean that there are 2 authors. Otherwise some readers might think Simon Huet-Sean Graham is a single person. I'm not sure about the years for Simon Huet, so I just copied your 2000."
```diff
-COPYRIGHT="2000 Simon Huet-Sean Graham"
+COPYRIGHT="
+	2000 Simon Huet
+	2010 Sean Graham
+	"
```

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,46 @@
...
+	make OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/AliveJournal
+	addAppDeskbarSymlink $appsDir/AliveJournal
+	install -t $docDir COPYRIGHT README
+	sed -i -e '/^===$/ d;' $docDir/README
```
fbrosson: There's no real need to have a sed command to remove just a "===" line.
I would either drop this sed command or change it a little to get rid of the changelog that is embedded in the README file.

BTW, there is no COPYRIGHT file, so you should remove it from the previous line.

fbrosson:
If/when you create a PR on https://github.com/grahams/alivejournal to fix the Makefile and the sources you have patched, you might also wish to convert the .rsrc info a .rdef and have upstream replace the rsrc by the rdef (with the appropriate changes in the Makefile to handle this.)
You will have to investigate because I don't remember how this is done. (Help/pointers welcome)

All I remember right now is that if you add an rdef, say alivejournal.rdef, on our side, i.e. to haikuports/haiku-apps/alivejournal/additional-files/ then you would add this near the end of INSTALL():
```
addResourcesToBinaries $portDir/additional-files/alivejournal.rdef $appsDir/AliveJournal
```
Some of our .rdef files are rdef.in files with patterns. (Look for rdef.in in "git log -p" to find recipes which do so.)

richienyhus:
Will fix ticket #720

haiku-apps/alivejournal/alivejournal-1.61.recipe
bach5000: Whoops! I'll fix these right away.

bach5000:
Ok, @fbrosson. I did some research ...
and found out that to convert you use this command : `rc -d -o outputfile.rdef inputfile.rsrc`, according to https://www.haiku-os.org/documents/dev/compile_them_resources/.

Do I make a PR on upstream with the .rdef file or do I add the file here?

scottmc:
Once you get this all working, please do a pull request on upstream with the updated code and see if the author merges and if/when they do let us know and we can update the recipe to reflect at that time.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,50 @@
+SUMMARY="Client for LiveJournal online journalling community"
+DESCRIPTION="AliveJournal is a revamped version of Simon Huet's original client."
+HOMEPAGE="https://github.com/grahams/alivejournal"
+COPYRIGHT="
+	2000 Simon Huet
+	2010 Sean Graham
```
fbrosson: I'm so sorry, the style I used for the COPYRIGHT variable in my previous review was wrong with regards to the tabs and newlines. So you will have to do a second small edit here.
```diff
-COPYRIGHT="
-	2000 Simon Huet
+COPYRIGHT="2000 Simon Huet
-	2010 Sean Graham
-	"
+	2010 Sean Graham"
```

(I had in mind the style for PROVIDES, REQUIRES, etc..., but the style for COPYRIGHT is different.)

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,50 @@
...
+	make OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/AliveJournal
+	addAppDeskbarSymlink $appsDir/AliveJournal
+	install -t $docDir README
+	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
```
fbrosson: I think you meant:
```
	sed -i -n -e '/http:\/\/www\.bebits\.com\/app\/1920/q;p' $docDir/README
```
but I might be wrong. I'll check asap.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,50 @@
...
+	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef $appsDir/AliveJournal
```
fbrosson: When it's possible to keep the lines <= 80 chars (without having to do uggly things) we prefer to do so.
```diff
-	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef $appsDir/AliveJournal
+	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef \
+		$appsDir/AliveJournal
```

Hmm, did you create the rdef file? If yes, you should "git add" (and "git commit") it.
(You need to "cd haiku-apps/alivejournal && mkdir additional-files" then put the alivejournal.rdef file there.)

Well, if know how to change the upstream Makefile to have it use the rdef file, you could open a PR on the upstream site with all these changes, so that you won't even need to add the rdef file here.
That said, both methods are ok. And I guess we have more recipes with rdef files in our repo than recipes of programs providing their own rdef file.

bach5000:
@scottmc Okay. I'll do that right away.

haiku-apps/alivejournal/alivejournal-1.61.recipe
bach5000: Oh, it's fine. I make mistakes all the time :)

bach5000:
Oh BTW the upstream author merged my commits.
https://github.com/grahams/alivejournal/pull/1
Should I change the recipes and get rid of the patch file?

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -46,5 +46,6 @@ INSTALL()
 	addAppDeskbarSymlink $appsDir/AliveJournal
 	install -t $docDir README
 	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
+	rc -d -o ../../../additional-files/alivejournal.rdef work-$portVersion/sources/alivejournal-$gitSrcRevision/Resource.rsrc
 	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef $appsDir/AliveJournal
```
fbrosson: There are 3 scenarios:

1) if you only have the rsrc file and don't care about the rdef file:
```
	xres -o $appsDir/AliveJournal Resource.rsrc
```
(I'm not 100% sure if this works this way here.)

2) if you have the rdef and want to convert it to a rsrc:
```
	rc -o Resource.rsrc $portDir/additional-files/alivejournal.rdef
	xres -o $appsDir/AliveJournal Resource.rsrc
```

3) if you have the rdef in additional-files and want to let addResourcesToBinaries take care of everything:
```
	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef \
		$appsDir/AliveJournal
```

Here I think you are trying to generate the rdef from the rsrc at build time.
We prefer to generate the rdef and add it to out repo in haiku-apps/alivejournal/additional-files/.

The best would be to ask upstream to host the rdef and make the changes to use it instead of the rsrc.

But as @scottmc wrote, you can (should?) first create the rdef and write the recipe using the third scenario described above. Then, when everything is OK, you can open a PR on the upstream repo with all the changes and a pointer to your recipe in the description, showing that it builds fine using the rdef you generated from the upstream rsrc.

bach5000: Ok.

fbrosson:
> Oh BTW the upstream author merged my commits.
grahams/alivejournal#1

Cool!

> Should I change the recipes and get rid of the patch file?

Sure, if the latest snapshot works without your patch.
This is great news. This is also a very good practice: upstreaming your patches is always better than having to keep them on our side.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,48 @@
...
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/AliveJournal
+	addAppDeskbarSymlink $appsDir/AliveJournal
+	install -t $docDir README
+	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
+	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef \
```
fbrosson: Now that the rdef is available upstream, you can make this change:
```diff
-	addResourcesToBinaries $portDir/additional-files/alivejournal.rdef \
-		$appsDir/AliveJournal
+	addResourcesToBinaries Resource.rdef $appsDir/AliveJournal
```

Oops, I guess you meant "middle = 61" (instead of not "middle = 51") in https://github.com/grahams/alivejournal/commit/fdb75158274cae7b3a96d25c88baf49ba7b8bb32#diff-86a5e8efa2009f917362ed20cb06995eR5

If you create a PR to fix this tiny mistake, maybe you could also see if the Makefile can be changed to automatically process the rdef file to produce and insert the resource in the runtime right after the compilation.
If I understand correctly, makefile_engine has the logic that looks for the RSRCS parameter in the Makefile.
I don't know if it also has the logic to look for a RDEF(S) parameter to process an rdef file.

In case makefile_engine does not know how to process rdef files, then you can either create a tiny PR to fix your typo for the "middle = 61" parameter, or use sed to fix it during the build. But it would be better to create a PR, explaining you made a little typo. I'm sure upstream will kindly merge your second PR.

So if you have the typo fixed, you can then use the simplified addResourcesToBinaries (shown on the top of this comment.)
If not, you will need to fix the middle number before you call addResourcesToBinaries in INSTALL.

fbrosson: You removed the PATCH line in the recipe, but you forgot to remove the patch itself.
To do so, you can:
```
cd haiku-apps/alivejournal/patches/
git rm alivejournal-1.61.patchset
git commit alivejournal-1.61.patchset
```

scottmc:
In the Makefile you should change it to be like this:

#	Specify the resource definition files to use. Full or relative paths can be
#	used.
RDEFS = src/HaikuTwitter.rdef

#	Specify the resource files to use. Full or relative paths can be used.
#	Both RDEFS and RSRCS can be utilized in the same Makefile.
RSRCS = 


And also on the LIBS = 
change "stdc++.r4" to "$(STDCPPLIBS)"

I believe that's the trick to get it to compile on gcc2 and gcc5.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,47 @@
...
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/AliveJournal
+	addAppDeskbarSymlink $appsDir/AliveJournal
+	install -t $docDir README
+	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
+	addResourcesToBinaries Resource.rdef $appsDir/AliveJournal
```
fbrosson: Thanks @scottmc for the info about the RDEFS parameter in the Makefile.

@bach5000, this is good news. Now, to test a modified Makefile (i.e. with a "RDEFS = Resource.rdef"), you can create a patchset with the changes Scott explained, and see if everything builds find, and also check if the app has its icon.
(Don't forget to remove the call to addResourcesToBinaries in the recipe, otherwise you won't be sure that the resource is being inserted in the runtime by the very Makefile.)
Once you know that the modified Makefile works as expected, you can create a PR to change the "middle = 51" to "middle = 61" and the "RSRCS = Resource.rsrc" to "RDEFS = Resource.rdef".
The final step will be to drop the temporary patch you just created to test all this.

bach5000:
Uh oh.
There's another one of those `syntax error before '*'`: 
https://github.com/grahams/alivejournal/blob/master/src/LJApplication/LJAccount/LJViewHistory/LJViewHistory.h#L36
Is this just me? Could someone else try to build the recipe?

raefaldhia:
@bach5000 you're forgot to change this line in your recipe https://github.com/haikuports/haikuports/pull/1890/files#diff-649589151b664b944ab20e424eae4ac6R8

The latest commit for alivejournal repository is e3d2c76f0645076f4e05ea942427e0f6e919948b

bach5000:
Oops. Thanks again, @raefaldhia!

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -7,7 +7,7 @@ LICENSE="GNU GPL v2"
 REVISION="1"
 gitSrcRevision="e3d2c76f0645076f4e05ea942427e0f6e919948b"
 SOURCE_URI="https://github.com/grahams/alivejournal/archive/$gitSrcRevision.tar.gz"
-CHECKSUM_SHA256="198950a11a0cb38cd9093d08619991ad6bbd2d376e7b1999b09a8a556d9add68"
+CHECKSUM_SHA256="8215d4ec005fd345c15ece8223f022cc6d434d940dc3c699a00c87c4e8b6ac49"
```
fbrosson: I've tried the updated recipe but I still get the same error as with the previous patch:
```
sources/alivejournal-e3d2c76f0645076f4e05ea942427e0f6e919948b/src/LJApplication/LJAccount/LJViewHistory/LJViewHistory.h:37: syntax error before `*'
make: *** [/boot/system/develop/etc/makefile-engine:291: objects/LJAccount.o] Error 1
```
Well, at least this means the patch was correctly upstreamed :-)
I still have 4 sources that compile without any error, but the compilation fails with the fifth file:
```
cd haiku-apps/alivejournal/work-1.61/sources/alivejournal-e3d2c76*
find . -name "*.o" | wc -l ; find . -name "*.cpp" | wc -l
4
23
```
As expected, I also get the same error if I update to the latest gitSrcRevision:
```diff
-gitSrcRevision="e3d2c76f0645076f4e05ea942427e0f6e919948b"
+gitSrcRevision="cbd31ce1bdb61a58053d25eaca8e275efcf464ec"
 SOURCE_URI="https://github.com/grahams/alivejournal/archive/$gitSrcRevision.tar.gz"
-CHECKSUM_SHA256="8215d4ec005fd345c15ece8223f022cc6d434d940dc3c699a00c87c4e8b6ac49"
+CHECKSUM_SHA256="e4f8a66660ae3e89d6d9a8993f35f8912c28a3fee72896288f61f611d74998ad"
```

raefaldhia:
Yeah, same here i compiled the source with GCC5(x86_64), is the recipe compile in gcc2?

fbrosson:
I tried on x86_gcc2 and got the same error.
(I did not even try it on x86_64 because the chances were higher with gcc2 (than with gcc5), given that the sources were written a long time ago for BeOS.)

raefaldhia:
Yeah @bach5000 need to patch more code, recently i patched it and got it working but log in won't work lol but it give response on wrong password.

fbrosson:
This is great news, @raefaldhia !
Did you keep your patches? Are they big?
I'm sure @bach5000 would like to see your patches (or get hints from your memory if you lost the patchset.)
Hmm, after all, would you mind submitting them to the upstream author (wthough a PR) or sending the patchset to @bach5000 so that he can create the PR himself?
Thanks!

raefaldhia:
I am! here it is https://transfer.sh/R4cHC/Archive.zip

fbrosson:
Thanks @raefaldhia !

@bach5000, I recommend using
```
curl -LO https://transfer.sh/R4cHC/Archive.zip
```
instead of following the link with a browser.
It would (probably) work with a browser, unless you disable JavaScript.

bach5000:
Wow. This is a huge patchset. I'll make a PR on the upstream author. Thanks @raefaldhia, for taking your time patching this whole thing!

bach5000:
Also, thanks to everyone that helped me on this task :) Although this might not be finished, I couldn't gotten this far without your help.

bach5000:
Got the patch merged :)

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,47 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	cmd:sed
+	"
+
```
fbrosson: On x86_64 (and x86) STDCPPLIBS is set to stdc++ but on x86_gcc2 it is set to stdc++.r4 so we should fix the Makefile. This can be done with a PATCH function (before BUILD).
```
PATCH()
{
	sed -i -e '/^LIBS=/ s/stdc++/$(STDCPPLIBS)/' Makefile
}
```
I don't know, however, if it would be OK to upstream the resulting change.

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,47 @@
...
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/AliveJournal
+	addAppDeskbarSymlink $appsDir/AliveJournal
+	install -t $docDir README
+	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
+	addResourcesToBinaries Resource.rdef $appsDir/AliveJournal
```
fbrosson: Now that the rdef is being processed by the Makefile, we can drop the
```
	addResourcesToBinaries Resource.rdef $appsDir/AliveJournal
```

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,47 @@
...
+	make OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/AliveJournal
+	addAppDeskbarSymlink $appsDir/AliveJournal
+	install -t $docDir README
+	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
```
fbrosson: What do you think of this change:
```diff
-	sed -n '/http://www.bebits.com/app/1920/q;p' $docDir/README
+	sed -i -n -e '/^------------$/q;p' $docDir/README
```

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,47 @@
+SUMMARY="Client for LiveJournal online journalling community"
+DESCRIPTION="AliveJournal is a revamped version of Simon Huet's original client."
+HOMEPAGE="https://github.com/grahams/alivejournal"
+COPYRIGHT="2000 Simon Huet
+	2010 Sean Graham"
+LICENSE="GNU GPL v2"
+REVISION="1"
+gitSrcRevision="e3d2c76f0645076f4e05ea942427e0f6e919948b"
```
fbrosson:
```diff
-gitSrcRevision="e3d2c76f0645076f4e05ea942427e0f6e919948b"
+gitSrcRevision="c25a118cde4d97f2d0d514732b548e592cac9c29"
```

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,47 @@
...
+SOURCE_URI="https://github.com/grahams/alivejournal/archive/$gitSrcRevision.tar.gz"
+CHECKSUM_SHA256="8215d4ec005fd345c15ece8223f022cc6d434d940dc3c699a00c87c4e8b6ac49"
```
fbrosson:
```diff
-CHECKSUM_SHA256="8215d4ec005fd345c15ece8223f022cc6d434d940dc3c699a00c87c4e8b6ac49"
+CHECKSUM_SHA256="675345bbadccda762fd87c1f8d133955af41b5e36584b2d9d497246f361e563c"
```

haiku-apps/alivejournal/alivejournal-1.61.recipe
bach5000: Which parts of the README file would that command change/get rid of? There are so many of those dashes in the README file :(

haiku-apps/alivejournal/alivejournal-1.61.recipe
bach5000: To be honest, I feel guilty bothering the upstream author. I think it's best to use this command for now.

fbrosson: I also think it is better to use the PATCH function.
The first reason is that I don't know who sets it, makefile_engine or haikuporter. Well I think it is makefile_engine.
The second reason is that libstdc++.r4 will disapear some day, while libstdc++ will remain.
So, since the Makefile on the upstream repo already works fine with x86_64, as soon as x86_gcc2 is phased out, upstream will be OK on all supported architectures.

haiku-apps/alivejournal/alivejournal-1.61.recipe
fbrosson: Well, as you probably figured it out, this sed command scans each line to see if the pattern matches. As soon as it finds a match, it just **q**uits. And before it quits, each line which does not match gets **p**rinted. And since we used the `-n`, sed just _eats_ the contents of the file, performing the actions in the commands provided. The result is that this sed command keeps everything from the first line to the line just before the very first match. Well, I'm sure you understood all this soon after you asked the question ;-)

haiku-apps/alivejournal/alivejournal-1.61.recipe
```diff
@@ -0,0 +1,51 @@
+SUMMARY="Client for LiveJournal online journalling community"
```
fbrosson: Oops, I had not paid enough attention to the SUMMARY.
I think this would be better.
```diff
-SUMMARY="Client for LiveJournal online journalling community"
+SUMMARY="A client for the LiveJournal online journalling community"
```
The README has this sentence:
> AliveJournal is a client for the LiveJournal online journalling community.

BTW, did you know that haikuporter refuses to process recipes if the SUMMARY starts with the name of the package?
So if you ever start a draft recipe for a program, say foobar, and define SUMMARY="foobar bla bla bla" it will not work :D

bach5000: Wait, really? I did not know that.

haiku-apps/alivejournal/alivejournal-1.61.recipe
bach5000: Oh, ok.

bach5000:
Yay 😄

fbrosson:

Issue #539, opened 2016-04-02
Add recipe for PecoBeat


richienyhus:
https://github.com/HaikuArchives/PecoBeat

fbrosson:
Done in #1898.

PR #1898, opened 2017-12-15
PecoBeat: new recipe

nysnatuss:
Added PecoBeat, a MIDI drum computer for Haiku. Completes #539.

I also patched a printf error (not including stdio.h) and some include path errors.

### DIFF
--- a/haiku-apps/pecobeat/pecobeat-1.0.recipe
+++ b/haiku-apps/pecobeat/pecobeat-1.0.recipe
@@ -1,0 +1,46 @@
+SUMMARY="MIDI drum computer"
+DESCRIPTION="PecoBeat is a MIDI drum computer for Haiku."
+HOMEPAGE="https://github.com/HaikuArchives/PecoBeat"
+COPYRIGHT="2001-2008 Werner Freytag
+	2009 Haiku"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="2f2200f6cc9dc9f7bf736e3f921563bce4753c55"
+SOURCE_URI="https://github.com/HaikuArchives/PecoBeat/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="25687e5637638b834d6ac7f2b69edce70523dc5750da466ce174cf6d4a7f56f0"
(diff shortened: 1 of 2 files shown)


haiku-apps/pecobeat/pecobeat-1.0.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="MIDI drum computer"
+DESCRIPTION="PecoBeat is a MIDI drum computer for Haiku."
+HOMEPAGE="https://github.com/HaikuArchives/PecoBeat"
+COPYRIGHT="2001-2008 Werner Freytag
+	2009 Haiku"
```
fbrosson: Maybe we should add 2014.
```diff
-	2009 Haiku"
+	2009, 2014 Haiku"
```

haiku-apps/pecobeat/pecobeat-1.0.recipe
```diff
@@ -0,0 +1,46 @@
...
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="2f2200f6cc9dc9f7bf736e3f921563bce4753c55"
+SOURCE_URI="https://github.com/HaikuArchives/PecoBeat/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="25687e5637638b834d6ac7f2b69edce70523dc5750da466ce174cf6d4a7f56f0"
+SOURCE_FILENAME="PecoBeat-$srcGitRev.tar.gz"
+SOURCE_DIR="PecoBeat-$srcGitRev"
+PATCHES="pecobeat-1.0.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
fbrosson: pecobeat builds fine on both x86_gcc2 and x86_64.
I could install it on x86_gcc2 and run the app.
But on x86_64 it cannot be installed because it needs internalmidi... which does not build on that arch.
So I guess it would be better to mark x86_64 (and x86 too) as untested.
```diff
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
```

PR #1900, opened 2017-12-16
app-crypt/codecrypt: add port

alaviss:
Done as part of GCI 2017

Changes to crypto++:
  - Make the build system use more reliable way to detect architecture (this fixes secondary arch, ~~may also fixes x86_gcc2 but untested~~)
  - Package a pkg-config file. This file is needed for codecrypt to build
  - Install the versioned version of crypto++ instead of doing the rename hack

### DIFF
--- a/app-crypt/codecrypt/codecrypt-1.8.recipe
+++ b/app-crypt/codecrypt/codecrypt-1.8.recipe
@@ -1,0 +1,54 @@
+SUMMARY="Post-quantum cryptography tool"
+DESCRIPTION="GnuPG-like unix program for encryption and signing that uses only \
+quantum-computer-resistant algorithms"
+HOMEPAGE="http://e-x-a.org/codecrypt"
+COPYRIGHT="2013-2016 Mirek Kratochvil"
+LICENSE="
+	GNU GPL v3
+	GNU LGPL v3
+	"
+REVISION="1"
+SOURCE_URI="http://e-x-a.org/codecrypt/files/codecrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="3be7c2a548aa336dbac706c3cab2071664d02b11652c16c02a8b4fb5cac6bc88"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	codecrypt$secondaryArchSuffix = $portVersion
+	cmd:ccr$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libcryptopp$secondaryArchSuffix
+	lib:libfftw3$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libcryptopp$secondaryArchSuffix
+	devel:libfftw3$secondaryArchSuffix
+	devel:libgmp$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:cmp
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	CPPFLAGS="-D_BSD_SOURCE" LIBS="-lbsd" runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	install -d -m 755 "$docDir"
+	install -t "$docDir" -m 644 README
+}
--- a/dev-libs/crypto++/crypto++-5.6.5.recipe
+++ b/dev-libs/crypto++/crypto++-5.6.5.recipe
@@ -14,20 +14,25 @@ The library contains the following algorithms:
 * Elliptic curve cryptography: ECDSA, ECNR, ECIES, ECDH, ECMQV
 * Insecure or obsolescent algorithms: MD2, MD4, MD5
 * And many more other algorithms."
-HOMEPAGE="https://cryptopp.com"
+HOMEPAGE="https://cryptopp.com/"
 COPYRIGHT="2015-2017 Wei Dai"
 LICENSE="Boost v1.0"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://www.cryptopp.com/cryptopp565.zip"
 CHECKSUM_SHA256="a75ef486fe3128008bbb201efee3dcdcffbe791120952910883b26337ec32c34"
 SOURCE_DIR=""
+# Upstream patch, might be obsolete in a later version
+PATCHES="crypto++-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion=$portVersion
+libVersionCompat="$libVersion compat >= ${libVersion%.*}"
+
 PROVIDES="
 	crypto++$secondaryArchSuffix = $portVersion
-	lib:libcryptopp$secondaryArchSuffix = $portVersion compat >= $portVersion
+	lib:libcryptopp$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
@@ -35,7 +40,7 @@ REQUIRES="
 
 PROVIDES_devel="
 	crypto++${secondaryArchSuffix}_devel = $portVersion
-	devel:libcryptopp$secondaryArchSuffix = $portVersion compat >= $portVersion
+	devel:libcryptopp$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES_devel="
 	crypto++$secondaryArchSuffix == $portVersion base
@@ -48,23 +53,38 @@ BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld
 	cmd:make
-	cmd:pkg_config$secondaryArchSuffix
 	cmd:nasm
+	cmd:pkg_config$secondaryArchSuffix
 	"
+
+defineDebugInfoPackage crypto++$secondaryArchSuffix \
+	$libDir/libcryptopp.so.$libVersion
 
 BUILD()
 {
-	make shared dynamic static $jobArgs
+	make HAS_SOLIB_VERSION=1 dynamic static $jobArgs
 }
 
 INSTALL()
 {
-	make install PREFIX=$prefix INCLUDEDIR=$includeDir
+	make \
+		HAS_SOLIB_VERSION=1 LDCONF=true \
+		BINDIR=$binDir LIBDIR=$libDir \
+		DATADIR=$dataDir INCLUDEDIR=$includeDir \
+		install
 
-	mv $libDir/libcryptopp.so $libDir/libcryptopp.so.$portVersion
-	ln -s $libDir/libcryptopp.so.$portVersion $libDir/libcryptopp.so
-
+	# Yes, they created the link. But they forgot to install it...
+	ln -s libcryptopp.so.$libVersion \
+		$libDir/libcryptopp.so.${libVersionCompat##* }
 	prepareInstalledDevelLib libcryptopp
+
+	install -d -m 755 $developLibDir/pkgconfig
+	cat > $developLibDir/pkgconfig/libcrypto++.pc << EOF
+Name: libcrypto++
+Description: Class library of cryptographic schemes
+Version: $portVersion
+Libs: -lcryptopp
+EOF
 
 	packageEntries devel \
 		$relativeDevelopDir
--- a/dev-libs/crypto++/patches/crypto++-5.6.5.patchset
+++ b/dev-libs/crypto++/patches/crypto++-5.6.5.patchset
@@ -1,0 +1,25 @@
+From ec1e84e8ecefb81f7a233f047c75446eee9dfb54 Mon Sep 17 00:00:00 2001
+From: Leorize <alaviss@users.noreply.github.com>
+Date: Sat, 16 Dec 2017 15:06:42 +0700
+Subject: Makefile: fix architecture detection
+
+
+diff --git a/GNUmakefile b/GNUmakefile
+index 2a0ff4c..ddd5422 100644
+--- a/GNUmakefile
++++ b/GNUmakefile
+@@ -15,9 +15,9 @@ LN ?= ln -sf
+ LDCONF ?= /sbin/ldconfig -n
+ UNAME := $(shell uname)
+ 
+-IS_X86 := $(shell uname -m | $(EGREP) -v "x86_64" | $(EGREP) -i -c "i.86|x86|i86")
++IS_X86 := $(shell $(CXX) $(CXXFLAGS) -dumpmachine 2>/dev/null | $(EGREP) -v "64" | $(EGREP) -i -c "i.86|x86|i86")
+ IS_X32 ?= 0
+-IS_X64 := $(shell uname -m | $(EGREP) -i -c "(_64|d64)")
++IS_X64 := $(shell $(CXX) $(CXXFLAGS) -dumpmachine 2>/dev/null | $(EGREP) -i -c "(_64|d64)")
+ IS_PPC := $(shell uname -m | $(EGREP) -i -c "ppc|power")
+ IS_ARM32 := $(shell uname -m | $(EGREP) -i -c "arm")
+ IS_ARM64 := $(shell uname -m | $(EGREP) -i -c "aarch64")
+-- 
+2.15.0
+


dev-libs/crypto++/additional-files/libcrypto++.pc
```diff
@@ -0,0 +1,10 @@
+# Written by Alexander Rødseth <rodseth@gmail.com>
```
extrowerk: You can do it in your recipe too, it doesn't need an extra file.

extrowerk: Example: https://github.com/haikuports/haikuports/blob/409b5225c23081d9ca44e35f45871791bee5c4cd/app-arch/snappy/snappy-1.1.7.recipe

dev-libs/crypto++/patches/crypto++-5.6.5.patchset
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/GNUmakefile b/GNUmakefile
+index 2a0ff4c..ddd5422 100644
+--- a/GNUmakefile
++++ b/GNUmakefile
+@@ -15,9 +15,9 @@ LN ?= ln -sf
+ LDCONF ?= /sbin/ldconfig -n
+ UNAME := $(shell uname)
+ 
+-IS_X86 := $(shell uname -m | $(EGREP) -v "x86_64" | $(EGREP) -i -c "i.86|x86|i86")
++IS_X86 := $(shell $(CXX) $(CXXFLAGS) -dumpmachine 2>/dev/null | $(EGREP) -v "64" | $(EGREP) -i -c "i.86|x86|i86")
```
extrowerk: Haiku on x86 defines "BePC" instead of x86 afaik. Check the uname -a output.

alaviss: It's more reliable this way and have a plus: it's cross-platform (as long as the said platfrom have gcc-compatible compiler). This diff was actually extracted from [upstream](https://github.com/weidai11/cryptopp/commit/a5bf9626817e96f8c8611b40fb4699202d98638b)

alaviss:
Updates:
  - dev-libs/crypto++: Generate pkg-config file in recipe instead of bundling an external one

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -17,25 +17,29 @@ The library contains the following algorithms:
...
 SOURCE_URI="https://www.cryptopp.com/cryptopp565.zip"
 CHECKSUM_SHA256="a75ef486fe3128008bbb201efee3dcdcffbe791120952910883b26337ec32c34"
 SOURCE_DIR=""
+# Upstream patch, might be obsolete in a later version
+PATCHES="crypto++-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersionCompat=${portVersion%.*}
```
fbrosson:
```diff
+libVersion=$portVersion
-libVersionCompat=${portVersion%.*}
+libVersionCompat=${libVersion%.*}
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -17,25 +17,29 @@ The library contains the following algorithms:
 HOMEPAGE="https://cryptopp.com"
```
fbrosson:
```diff
-HOMEPAGE="https://cryptopp.com"
+HOMEPAGE="https://cryptopp.com/"
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -17,25 +17,29 @@ The library contains the following algorithms:
...
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersionCompat=${portVersion%.*}
+
 PROVIDES="
 	crypto++$secondaryArchSuffix = $portVersion
-	lib:libcryptopp$secondaryArchSuffix = $portVersion compat >= $portVersion
+	lib:libcryptopp$secondaryArchSuffix = $portVersion compat >= $libVersionCompat
```
fbrosson:
```diff
-	lib:libcryptopp$secondaryArchSuffix = $portVersion compat >= $libVersionCompat
+	lib:libcryptopp$secondaryArchSuffix = $libVersion compat >= $libVersionCompat
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -17,25 +17,29 @@ The library contains the following algorithms:
...
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
 
 PROVIDES_devel="
 	crypto++${secondaryArchSuffix}_devel = $portVersion
-	devel:libcryptopp$secondaryArchSuffix = $portVersion compat >= $portVersion
+	devel:libcryptopp$secondaryArchSuffix = $portVersion compat >= $libVersionCompat
```
fbrosson:
```diff
-	devel:libcryptopp$secondaryArchSuffix = $portVersion compat >= $libVersionCompat
+	devel:libcryptopp$secondaryArchSuffix = $libVersion compat >= $libVersionCompat
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -48,24 +52,38 @@ BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld
 	cmd:make
-	cmd:pkg_config$secondaryArchSuffix
 	cmd:nasm
+	cmd:pkg_config$secondaryArchSuffix
 	"
 
+defineDebugInfoPackage crypto++$secondaryArchSuffix \
+	$libDir/libcryptopp.so.$portVersion
```
fbrosson:
```diff
-	$libDir/libcryptopp.so.$portVersion
+	$libDir/libcryptopp.so.$libVersion
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -48,24 +52,38 @@ BUILD_PREREQUIRES="
...
-	mv $libDir/libcryptopp.so $libDir/libcryptopp.so.$portVersion
-	ln -s $libDir/libcryptopp.so.$portVersion $libDir/libcryptopp.so
+	make \
+		HAS_SOLIB_VERSION=1 LDCONF=true \
+		BINDIR=$binDir LIBDIR=$libDir \
+		DATADIR=$dataDir INCLUDEDIR=$includeDir \
+		install
 
+	# Yes, they created the link. But they forgot to install it...
+	ln -s libcryptopp.so.$portVersion $libDir/libcryptopp.so.$libVersionCompat
```
fbrosson:
```diff
-	ln -s libcryptopp.so.$portVersion $libDir/libcryptopp.so.$libVersionCompat
+	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.$libVersionCompat
```

fbrosson:
@alaviss, you rock!
I've just built your updated crypto++ on x86_gcc2 using the x86 secondary arch. The tests are all OK!
```
All tests passed!

Test ended at Sat Dec 16 11:32:06 2017
Seed used was: 1513423920
```
(Now I'll try to build the previous revision, without your patches, and run the tests again.)

alaviss:
`libVersion` should be added into [HaikuPorter Guidelines](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines)

Also the `HOMEPAGE` convention

alaviss:
Implemented all @fbrosson suggestions

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -48,24 +53,38 @@ BUILD_PREREQUIRES="
...
+		BINDIR=$binDir LIBDIR=$libDir \
+		DATADIR=$dataDir INCLUDEDIR=$includeDir \
+		install
 
+	# Yes, they created the link. But they forgot to install it...
+	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.$libVersionCompat
 	prepareInstalledDevelLib libcryptopp
 
+	install -d -m 755 $developLibDir/pkgconfig
+	cat > $developLibDir/pkgconfig/libcrypto++.pc << EOF
```
fbrosson: When I try to build codecrypt (after having installed crypto++ with your changes), I get:
```
checking for pkg-config... /bin/pkg-config
checking pkg-config is at least version 0.25... yes
checking for CRYPTOPP... no
configure: error: Package requirements (libcrypto++) were not met:

No package 'libcrypto++' found
```
The libcrypto++.pc we get has this:
```
Name: libcrypto++-5.6.5
Description: Class library of cryptographic schemes
Version: 5.6.5
Libs: -lcryptopp
```
But I think it should look like this:
```
prefix=/packages/crypto++-5.6.5-3/.self
exec_prefix=${prefix}
libdir=${prefix}/develop/lib
includedir=${prefix}/develop/headers
Name: libcrypto++-5.6.5
Description: Class library of cryptographic schemes
Version: 5.6.5
Libs: -L${libdir} -lcryptopp
```
I liked your idea of having a template in additional-files. You could put:
```
prefix=@PREFIX@
exec_prefix=${prefix}
libdir=${prefix}/develop/lib
includedir=${prefix}/develop/headers
Name: libcrypto++-@VERSION@
Description: Class library of cryptographic schemes
Version: @VERSION@
Libs: -L${libdir} -lcryptopp
```
in libcrypto++.pc.in and then use
```
	sed -i \
		-e "s|@PREFIX@|$prefix|;" \
		-e "s|@VERSION@|$portVersion|;" \
		$portDir/additional-files/libcrypto++.pc.in \
		> $developLibDir/pkgconfig/libcrypto++.pc
```

alaviss: Regarding the external file, @extrowerk [suggested](https://github.com/haikuports/haikuports/pull/1900#discussion_r157337733) against it.

Are you sure you're using my updated crypto++? It's working well here on both x86_gcc2 and x86_64.

And no, the only reason .pc looks bloated like so is for autotools interop. Since our library is in the compiler search path, it's redundant to put lib path in the .pc file

fbrosson: You are right. I think that, as you suspected, I had the old crypto++ when I tried to build codecrypt the first time.
(So sorry for the false alert.)

Hmm, all the other pc files I have on my system don't have the version appended to the Name: field. So maybe you can make this tiny change:
```diff
-Name: libcrypto++-$portVersion
+Name: libcrypto++
```

app-crypt/codecrypt/codecrypt-1.8.recipe
```diff
@@ -0,0 +1,52 @@
...
+BUILD()
+{
+	CPPFLAGS="-D_BSD_SOURCE" LIBS="-lbsd" runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
```
fbrosson: I think it would be nice to ship the README. What do you think?
```diff
 	make install
+	install -d "$docDir"
+	install -t "$docDir" README
```

I'll try building everything on x86_gcc2 for the x86 secondary arch.

BTW, for recipes which are not available on x86_gcc2 primary arch and which, consequently, need to be built for the x86 secondary arch, we sometimes choose to put the runtime in $prefix/bin instead of $binDir.
This is done by defining commandBinDir and commandSuffix, as you can see in e.g. app-office/calcurse/calcurse-4.3.0.recipe.

I don't know if it would be worth making the same changes for this recipe.
The only advantage would be that, on x86_gcc2, instead of typing ccr-x86 we would be able to type just ccr.

Let's leave codecrypt as it is for now, and, if other mentors/developers ask for the commandBinDir-commandSuffix "trick" before we merge, then, if you feel like doing it, that's ok for me.

dev-libs/crypto++/patches/crypto++-5.6.5.patchset
extrowerk: True

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -14,28 +14,33 @@ The library contains the following algorithms:
...
 CHECKSUM_SHA256="a75ef486fe3128008bbb201efee3dcdcffbe791120952910883b26337ec32c34"
 SOURCE_DIR=""
+# Upstream patch, might be obsolete in a later version
+PATCHES="crypto++-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion=$portVersion
+libVersionCompat=${libVersion%.*}
```
fbrosson: Hmm, libVersionCompat is already in 4 recipes:
* dev-libs/libpipeline/libpipeline-1.5.0.recipe
* net-libs/libblkmaker/libblkmaker-0.5.3.recipe
* sci-mathematics/gmp_ecm/gmp_ecm-7.0.4.recipe
* sys-libs/hwloc/hwloc-1.11.8.recipe

and it is defined differently.
I though about changing these 4 recipes to adjust to what we have in this PR right now, but the consequence would be that we would have much longer lines.
So I think it would be better to only change crypto++ this way:
1) first we rename to libVersionShort what we were calling libVersionCompat.
2) then we define libVersionCompat as `$libVersion compat >= $libVersionShort`

These changes would allow to have a consistent definition of libVersionCompat across the 4 (and soon 5) recipes which use it. I cannot see any better alternative...
What do you think of "libVersionShort" ? I first though naming this libVersionLink, then libVersionSymLink, but it looks like libVersionShort is the less confusing. What do you think?
Thanks!

```diff
 libVersion=$portVersion
-libVersionCompat=${libVersion%.*}
+libVersionShort=${libVersion%.*}
+libVersionCompat="$libVersion compat >= $libVersionShort"
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -14,28 +14,33 @@ The library contains the following algorithms:
...
+
 PROVIDES="
 	crypto++$secondaryArchSuffix = $portVersion
-	lib:libcryptopp$secondaryArchSuffix = $portVersion compat >= $portVersion
+	lib:libcryptopp$secondaryArchSuffix = $libVersion compat >= $libVersionCompat
```
fbrosson:
```diff
-	lib:libcryptopp$secondaryArchSuffix = $libVersion compat >= $libVersionCompat
+	lib:libcryptopp$secondaryArchSuffix = $libVersionCompat
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -14,28 +14,33 @@ The library contains the following algorithms:
...
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
 
 PROVIDES_devel="
 	crypto++${secondaryArchSuffix}_devel = $portVersion
-	devel:libcryptopp$secondaryArchSuffix = $portVersion compat >= $portVersion
+	devel:libcryptopp$secondaryArchSuffix = $libVersion compat >= $libVersionCompat
```
fbrosson:
```diff
-	devel:libcryptopp$secondaryArchSuffix = $libVersion compat >= $libVersionCompat
+	devel:libcryptopp$secondaryArchSuffix = $libVersionCompat
```

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -48,24 +53,38 @@ BUILD_PREREQUIRES="
...
-	mv $libDir/libcryptopp.so $libDir/libcryptopp.so.$portVersion
-	ln -s $libDir/libcryptopp.so.$portVersion $libDir/libcryptopp.so
+	make \
+		HAS_SOLIB_VERSION=1 LDCONF=true \
+		BINDIR=$binDir LIBDIR=$libDir \
+		DATADIR=$dataDir INCLUDEDIR=$includeDir \
+		install
 
+	# Yes, they created the link. But they forgot to install it...
+	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.$libVersionCompat
```
fbrosson:
```diff
-	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.$libVersionCompat
+	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.$libVersionShort
```

dev-libs/crypto++/crypto++-5.6.5.recipe
alaviss: The name doesn't actually make any sense. It's not the "short" version number, nor the version we symlinked, but rather the ABI version that's supported by that version of the library.

I suggest we drop `libVersionShort` and do this

```diff
 libVersion=$portVersion
-libVersionCompat=${libVersion%.*}
+libVersionCompat="$libVersion compat >= ${libVersion%.*}"
```

And for the symlink

```diff
-	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.$libVersionCompat
+	ln -s libcryptopp.so.$libVersion $libDir/libcryptopp.so.${libVersionCompat##* }
```

For the other ports, I suggest using `${libVersion%.*}` or `${libVersion%%.*}` depending on the authors versioning scheme. This will ensure that maintainers would only need to change one line when `SONAME` bump happens instead of two, reducing the margin of error. And IMO, the less constants we depend on the better, especially when it comes to versioning.

What's your opinion on this?

fbrosson: Excellent idea! You rock, @alaviss !
To summarize, we would use:

* for libs which have a `libfoobar.so.x -> libfoobar.x.y.z`:
```
```

* for libs which have a `libfoobar.so.x.y -> libfoobar.x.y.z`:
```
libVersionCompat="$libVersion compat >= ${libVersion%.*}"
```

and we would set libVersion to either portVersion, if this is appropriate, or to a the appropriate value if it is different from the portVersion.

And since there are very few recipes which need to compute `${libVersionCompat##* }`, this is not a problem.

So OK for me!

alaviss:
Updates:
- app-crypt/codecrypt: Package README.md
- dev-libs/crypto++: Use `libVersionCompat` scheme as discussed with @fbrosson

dev-libs/crypto++/crypto++-5.6.5.recipe
```diff
@@ -48,24 +53,39 @@ BUILD_PREREQUIRES="
...
+	ln -s libcryptopp.so.$libVersion \
+		$libDir/libcryptopp.so.${libVersionCompat##* }
 	prepareInstalledDevelLib libcryptopp
 
+	install -d -m 755 $developLibDir/pkgconfig
+	cat > $developLibDir/pkgconfig/libcrypto++.pc << EOF
+Name: libcrypto++
+Description: Class library of cryptographic schemes
+Version: $portVersion
+Libs: -lcryptopp
```
fbrosson: Thanks @alaviss for your recent updates. I have a last suggestion...

I understood that the following suggestion is not needed, but I'd like to suggest it for 2 reasons (explained below):
```diff
 	cat > $developLibDir/pkgconfig/libcrypto++.pc << EOF
+prefix=$prefix
+exec_prefix=\${prefix}
+libdir=\${prefix}/$relativeDevelopLibDir
+includedir=\${prefix}/$relativeIncludeDir
 Name: libcrypto++
 Description: Class library of cryptographic schemes
 Version: $portVersion
-Libs: -lcryptopp
+Libs: -L\${libdir} -lcryptopp
 EOF
```
* This would make the pc file (for this lib) consistent with the pc files that are generated for all the other packages.
* Paranoids like me would feel safer ;-)

The only downside would be that if/when pkg-config ever changes in a way that it stops adding all these lines, then we would have to revert this suggestion to keep the consistency.
What do you think?
Isn't it nice? (I've even "converted" it to avoid the additional-files directory ;-)

fbrosson: Oops, I forgot to tell that everything is OK, so you can add a ` (#1900)` to your commit summaries like we did for signify.
I hope you will accept my last suggestion, but, of course, I can merge this PR without it if you prefer and/or if you convince me that it is totally unneeded.

alaviss: IMO, consistency is unnecessary for a file that most humans will never read. They appear "consistent" as they are usually generated with autotools using templates like [this][0]

Those additional lines, which are freeform variable, have no special meaning in a pkg-config file. They are often used to [reduce clutter][1]
> Before describing the more useful fields, it will be helpful to demonstrate variable definitions. The most common usage is to define the installation paths so that they don't clutter the metadata fields. Since the variables are expanded recursively, this is very helpful when used in conjunction with autoconf derived paths.

Furthermore, since our library was put in Haiku's compiler search path (`/boot/system/develop/lib`), the `-L` directive (which tells the compiler where to find said library) becomes redundant in this case, which make those variables become more or less clutter themselves
```
$ gcc -print-search-dirs
**omitted**
libraries: **lots of stuff omitted**:/boot/system/develop/lib/
```

As for you being paranoid, be sure that all ports I made and/or modified are tested on both x86_gcc2, x86 (secondary) and x86_64 so they should always behave correctly

[0]: https://tecnocode.co.uk/2014/12/09/a-checklist-for-writing-pkg-config-files/
[1]: https://people.freedesktop.org/~dbn/pkg-config-guide.html#writing

fbrosson: OK, thanks for the pointers and explanation :)

PR #1901, opened 2017-12-16
Add worldclock recipe

localanu:
Part of GCI 2017


haiku-apps/wordclock/wordclock-1.0.recipe
```diff
@@ -0,0 +1,44 @@
+SUMMARY="An imprecise clock replicant"
```
humdingerb: Remove "replicant". Let's save that for the DESCRIPTION.

haiku-apps/wordclock/wordclock-1.0.recipe
```diff
@@ -0,0 +1,44 @@
...
+DESCRIPTION="WordClock is imprecise clock replicant and this show clock with word"
```
humdingerb: That's not very nicely phrased. How about:
"WordClock is an imprecise clock that displays the time in words instead of digits. It can be used as a Replicant as well."

haiku-apps/wordclock/wordclock-1.0.recipe
```diff
@@ -0,0 +1,44 @@
...
+srcGitRev="cbb5da9a498ec99a71a68f039042601207eafdd1"
+SOURCE_URI="https://github.com/Janus2/WordClock/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="fe29a3f17467927f88052d9c260b894d8424cf175b633aa2ad22deea2b25abba"
+SOURCE_FILENAME="WordClock-$srcGitRev.tar.gz"
+SOURCE_DIR="WordClock-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	WordClock = $portVersion
```
humdingerb: This should be the same low-casing as the recipe name: "wordclock"

haiku-apps/wordclock/wordclock-1.0.recipe
```diff
@@ -0,0 +1,44 @@
...
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:gcc
```
humdingerb: As this is C++, using "cmd:g++" is preferred.

haiku-apps/wordclock/wordclock-1.0.recipe
```diff
@@ -0,0 +1,44 @@
...
+	cmd:make
+	"
+
+BUILD()
+{
+	make $jobArgs
+}
```
humdingerb: WordClock is localized. Currently it was only translated to German. Nevertheless, after the "make $jobArgs", add a line "make bindcatalogs". This will add the localizations to the resources of the executable.

haiku-apps/wordclock/wordclock-1.0.recipe
fbrosson: Thanks for this PR!
There are only a couple of things to change.

1) Please change DESCRIPTON to:
```
DESCRIPTION="WordClock is an imprecise clock replicant which displays the time \
using words instead of numbers."
```

2) Remove the blank line at the end of the recipe.

3) Fix the rdef file by inserting a PATCH function before BUILD:
```
PATCH()
{
	sed -i \
		-e "/major/ s/=.*/= 1,/;" \
		-e "/middle/ s/=.*/= 0,/;" \
		-e "/minor/ s/=.*/= 0,/;" \
		sources/WordClock.rdef
}
```

haiku-apps/wordclock/wordclock-1.0.recipe
fbrosson: Good catch. Thanks @humdingerb!

fbrosson: @localanu, if you are unsure about what should be converted to lower-case, well, it's only for the first line of PROVIDES, not for the app: line.
So this is what we should have for the PROVIDES:
```
PROVIDES="
	wordclock = $portVersion
	app:WordClock = $portVersion
	"
```

For the DESCRIPTION, @humdingerb's suggestion will look like this:
```
DESCRIPTION="WordClock is an imprecise clock that displays the time in words \
instead of digits. It can be used as a Replicant as well."
```
Hmm, @humdingerb, wouldn't it be better to write "replicant" all lower-case?
(Unless, of course, the Haiku docs mention "Replicant").

@localanu, in BUILD, don't forget to add
```
	make bindcatalogs
```
after the
```
	make $jobArgs
```
as @humdingerb asked.

Also, please don't forget to remove the empty blank line at the end of the recipe.

Thanks!

localanu:
Updated

haiku-apps/wordclock/wordclock-0.1.1.recipe
```diff
@@ -0,0 +1,45 @@
+SUMMARY="An imprecise clock"
+DESCRIPTION="WordClock is an imprecise clock that displays the time in words instead of digits"
```
humdingerb: There's a full-stop missing. Also, didn't we wanted to add that it can be used as a Replicant?
When adding that, we can also point to another feature. So:
```
DESCRIPTION="WordClock is an imprecise clock that displays the time in words \
instead of digits. You can change the used font and embed the clock as a \
Replicant on the Desktop."
```

haiku-apps/wordclock/wordclock-0.1.1.recipe
```diff
@@ -0,0 +1,47 @@
...
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	make install TARGET_DIR=$appsDir
+
+	addAppDeskbarSymlink $appsDir/WordClock
+}
+
```
fbrosson: No need for a blank line at the end of the recipe.
Could you please remote it? Thanks :-)

humdingerb: oops. too late... :)

fbrosson: Never mind ;-)

humdingerb: Just for my education... wasn't it recommended at some time to always have a blank line at the end of files? For people using vim IIRC...

fbrosson: Some editors, like Pe (if I'm not mistaken) as well as those which run on a popular OS which uses CRLF-style end-of-lines (instead of LineFeed on Unix-like OSes) do not bother to add anything at the end of the very last line of a text file.
So when you open with vi a file that was saved with those "text editors", vi complains/warns about that by saying: "[noeol]" on Haiku or "[Incomplete last line]" on Linux. Then it adds the LF for you. This means you can "repair" the file just by saving it with vi.
So, for those who don't use vi, adding an empty line forces those "text editors" to add a CRLF+CR (or 2 LF), but I'm not really sure about what these editors add on such cases.

On GitHub, when a file is missing the LF at the last lines, we see a special zero character/glyph which makes it clear that the file is missing a LF there. So it is safe to ask for the deletion of the blank line if this spcial zero char/glyph was not shown :-)

localanu:
Okay,next task?

fbrosson:
Sorry @localanu, I was not logged on the GCI site! 🙃

haiku-apps/wordclock/wordclock-0.1.1.recipe
humdingerb: Thanks for the explanation.
> saying: "[noeol]" on Haiku

That's almost christmassy... :)

localanu:
no problem

fbrosson:

PR #1902, opened 2017-12-16
Vim: bump version

extrowerk:
Somebody should review this.


pulkomandy:
I assume it is just the last part of the patchset where you commented out input method code that needs a review? I have not much knowledge about vim code in that area, but I think we should check if this has any bad consequences on our Japanese input methods (Canna, Anthy, ...) or if it is something else.

Korli also told me that there is no libruby for gcc2, so we should either re-add one (from an old version of ruby because new ones won't work), or disable the ruby support in the gcc2 compiled version of vim. Or maybe just add secondary arch support to the recipe.

extrowerk:
Lets retry with current one, closing this as old.

PR #1904, opened 2017-12-17
Lapack: fix pkgconfig files

extrowerk:
lapack

Missing prefix declaration.


extrowerk:
@korli :true :(
So what's up with fixpkgconfig?

korli:
prefix is obviously missing from the pc file. Maybe just add it in the template.

extrowerk:
Fixed

PR #1905, opened 2017-12-17
GSL: bump version

extrowerk:


sci-libs/gsl/gsl-2.4.recipe
```diff
@@ -0,0 +1,120 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize
```
fbrosson:
```diff
 	cmd:automake
+	cmd:find
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
-	cmd:libtoolize
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
-	cmd:find
```

extrowerk:
It was a big version bump, but i'm not sure if it is compatible with 1.x or not. It would be maybe better to rename the old recipe to gsl1.
Have you got any info about the compatibility, @korli ?

waddlesplash:
It appears "xaos" and "krita" are the only things that require GSL, so if they still build under this version, then we can merge it.

extrowerk:
@waddlesplash : Xaos compiles fine

PR #1909, opened 2017-12-17
Recipe for WordGrinder (with lua issues not yet fixed)

bach5000:
This recipe (probably, but unlikely) should work, although there's some error with the source.
* Line 101: https://github.com/davidgiven/wordgrinder/blob/master/Makefile#L101
* Error: `ninja: unknown target 'all', did you mean 'dev'?`

Some help would greatly be appreciated.

### DIFF
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -1,0 +1,46 @@
+SUMMARY="A simple, Unicode-aware word processor that runs on the console"
+DESCRIPTION="WordGrinder is designed to get out of your way and let you write; \
+it does very little, but what it does it does it well. It supports basic paragraph \
+styles, basic character styles, basic screen markup, a menu interface that means you \
+don't have to remember complex key sequences, HTML import and export, and some other \
+useful features."
+HOMEPAGE="http://cowlark.com/wordgrinder/"
+COPYRIGHT="2007-2017 David Given"
+LICENSE="MIT"
+REVISION="1"
+gitSrcRevision="3b07ccffc3043ef03a589e7ff3ab22240e5dcdda"
+SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$gitSrcRevision.tar.gz"
+CHECKSUM_SHA256="713543928e7b5dd4677e7edcf5351f75ce63bf62db316cbf761868fd86cc79c5"
+SOURCE_DIR="wordgrinder-$gitSrcRevision"
+
+ARCHITECTURES="x86_gcc2 x86_64"
+
+PROVIDES="
+	wordgrinder = $portVersion
+	app:WordGrinder = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	cmd:ninja
+	"
+
+BUILD()
+{
+	make OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/WordGrinder
+	addAppDeskbarSymlink $appsDir/WordGrinder
+}
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -21,15 +21,16 @@ PROVIDES="
 REQUIRES="
 	haiku
 	lib:libncurses
-	lib:liblua5
+	lib:liblua
 	"
 
 BUILD_REQUIRES="
 	haiku_devel
 	devel:libncurses
-	devel:liblua5
+	devel:liblua
 	"
 BUILD_PREREQUIRES="
+	cmd:g++
 	cmd:make
 	cmd:ninja
 	"
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -10,7 +10,7 @@ LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
-SOURCE_DIR="wordgrinder-$gitSrcRevision"
+SOURCE_DIR="wordgrinder-$portVersion"
 
 ARCHITECTURES="x86_gcc2 x86_64"
 
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -11,7 +11,7 @@ REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
 
-ARCHITECTURES="x86_gcc2 x86_64"
+ARCHITECTURES="!x86_gcc2 !x86_64"
 
 PROVIDES="
 	wordgrinder = $portVersion
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -10,7 +10,6 @@ LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
-SOURCE_DIR="wordgrinder-$portVersion"
 
 ARCHITECTURES="x86_gcc2 x86_64"
 
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -16,7 +16,7 @@ ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
 	wordgrinder = $portVersion
-	app:WordGrinder = $portVersion
+	cmd:WordGrinder = $portVersion
 	"
 REQUIRES="
 	haiku
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -9,7 +9,7 @@ COPYRIGHT="2007-2017 David Given"
 LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256="713543928e7b5dd4677e7edcf5351f75ce63bf62db316cbf761868fd86cc79c5"
+CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
 SOURCE_DIR="wordgrinder-$gitSrcRevision"
 
 ARCHITECTURES="x86_gcc2 x86_64"
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -8,8 +8,7 @@ HOMEPAGE="http://cowlark.com/wordgrinder/"
 COPYRIGHT="2007-2017 David Given"
 LICENSE="MIT"
 REVISION="1"
-gitSrcRevision="3b07ccffc3043ef03a589e7ff3ab22240e5dcdda"
-SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$gitSrcRevision.tar.gz"
+SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="713543928e7b5dd4677e7edcf5351f75ce63bf62db316cbf761868fd86cc79c5"
 SOURCE_DIR="wordgrinder-$gitSrcRevision"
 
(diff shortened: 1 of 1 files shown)


app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,46 @@
...
+it does very little, but what it does it does it well. It supports basic paragraph \
+styles, basic character styles, basic screen markup, a menu interface that means you \
+don't have to remember complex key sequences, HTML import and export, and some other \
+useful features."
+HOMEPAGE="http://cowlark.com/wordgrinder/"
+COPYRIGHT="2007-2017 David Given"
+LICENSE="MIT"
+REVISION="1"
+gitSrcRevision="3b07ccffc3043ef03a589e7ff3ab22240e5dcdda"
+SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$gitSrcRevision.tar.gz"
```
fbrosson:
```diff
-gitSrcRevision="3b07ccffc3043ef03a589e7ff3ab22240e5dcdda"
-SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$gitSrcRevision.tar.gz"
+SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
```
You can (well, should) remove the SOURCE_DIR if you swich to a stable release.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,46 @@
...
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	cmd:ninja
+	"
+
+BUILD()
+{
+	make OBJ_DIR=objects
```
fbrosson: This might help, but I don't know if it will be enough.
```diff
-	make OBJ_DIR=objects
+	make PREFIX=$prefix BINDIR=$binDir MANDIR=$manDir
```
I think you will also need to add one of the versions of Lua: lib:liblua5 to REQUIRES and devel:liblua5 to BUILD_REQUIRES. (I'm not sure which version of Lua should be used.)

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,46 @@
...
+}
+
+INSTALL()
+{
+	install -d $appsDir $docDir
+	install -t $appsDir objects/WordGrinder
+	addAppDeskbarSymlink $appsDir/WordGrinder
```
fbrosson: I don't know if this change will work, but this is what I would try:
```diff
-	install -d $appsDir $docDir
-	install -t $appsDir objects/WordGrinder
-	addAppDeskbarSymlink $appsDir/WordGrinder
+	make install PREFIX=$prefix BINDIR=$binDir MANDIR=$manDir
```

bach5000:
Oh wait. I need to update my checksum. Please wait.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+styles, basic character styles, basic screen markup, a menu interface that means you \
+don't have to remember complex key sequences, HTML import and export, and some other \
+useful features."
+HOMEPAGE="http://cowlark.com/wordgrinder/"
+COPYRIGHT="2007-2017 David Given"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
+SOURCE_DIR="wordgrinder-$gitSrcRevision"
```
fbrosson:
```diff
-SOURCE_DIR="wordgrinder-$gitSrcRevision"
```
(We only define SOURCE_DIR if its default value is not correct.)
I can't run Haiku at the moment, so I won't be able to test.

Regarding ncurses, maybe you could try ncurses6 (instead of ncurses).
Regarding Lua, you might wish to try compiling against Lua 5.3.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+
+ARCHITECTURES="x86_gcc2 x86_64"
+
+PROVIDES="
+	wordgrinder = $portVersion
+	app:WordGrinder = $portVersion
```
fbrosson:
```diff
-	app:WordGrinder = $portVersion
+	cmd:wordgrinder = $portVersion
```
(probably)

app-text/wordgrinder/wordgrinder-0.7.1.recipe
bach5000: Whoops. I meant to change that :0
For lua, would I use liblua5?

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,46 @@
...
+styles, basic character styles, basic screen markup, a menu interface that means you \
+don't have to remember complex key sequences, HTML import and export, and some other \
+useful features."
+HOMEPAGE="http://cowlark.com/wordgrinder/"
+COPYRIGHT="2007-2017 David Given"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
+SOURCE_DIR="wordgrinder-$portVersion"
```
fbrosson: haikuporter already sets a default value for SOURCE_DIR, and that default value is `wordgrinder-$portVersion`.
So this means you don't need to define SOURCE_DIR.
(Actually, we prefer to not have a SOURCE_DIR in the recipe when its default value is correct for the tarball.)

bach5000: Whoops. I got that fixed.
Also, for some reason, there are errors saying that devel:liblua5 and devel:libncurses6  doesn't exist. Does this mean I have to install these manually?

fbrosson: Hmm, the packages are ncurses6 abd ncurses6_devel. I see that they provide lib:libncurses and devel:libncurses, so you just need to drop the 6.
BTW, to find what package provides devel:libncurses, you can use:
```
pkgman search devel:libncurses
```
It should tell that ncurses6_devel provides it. Then you can
```
pkgman install ncurses6_devel
```

fbrosson: I've just checked for Lua:
devel:liblua is provided by lua_devel, so you can
```
pkgman install lua_devel
```
You will get Lua 5.3.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+
+ARCHITECTURES="x86_gcc2 x86_64"
+
+PROVIDES="
+	wordgrinder = $portVersion
+	cmd:WordGrinder = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libncurses6
```
fbrosson:
```diff
 REQUIRES="
 	haiku
-	lib:libncurses6
+	lib:libncurses
 	lib:liblua
+	lib:libz
 	"
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+	lib:liblua
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libncurses6
```
fbrosson:
```diff
 BUILD_REQUIRES="
 	haiku_devel
-	devel:libncurses6
+	devel:libncurses
 	devel:liblua
+	devel:libz
 	"
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+	devel:liblua
+	"
+BUILD_PREREQUIRES="
+	cmd:g++
+	cmd:make
+	cmd:ninja
+	"
```
fbrosson:
```diff
 	cmd:ninja
+	cmd:pkg_config
 	"
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+
+BUILD()
+{
+	make PREFIX=$prefix BINDIR=$binDir DOCDIR=$docDir MANDIR=$manDir
```
fbrosson:
```diff
 BUILD()
 {
+	mkdir -p build
-	make PREFIX=$prefix BINDIR=$binDir DOCDIR=$docDir MANDIR=$manDir
+	make OBJDIR=$sourceDir/build PREFIX=$prefix BINDIR=$binDir \
+		DOCDIR=$docDir MANDIR=$manDir
 }
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+}
+
+INSTALL()
+{
+	make install PREFIX=$prefix BINDIR=$binDir DOCDIR=$docDir MANDIR=$manDir
```
fbrosson:
```diff
 INSTALL()
 {
-	make install PREFIX=$prefix BINDIR=$binDir DOCDIR=$docDir MANDIR=$manDir
+	make install OBJDIR=$sourceDir/build PREFIX=$prefix BINDIR=$binDir \
+		DOCDIR=$docDir MANDIR=$manDir
 }
```

bach5000:
@fbrosson Now I have the same error message as you do.
I didn't know Lua could make this much trouble :0

fbrosson:
The "devel:liblua" (in BUILD_REQUIRES) can be provided by several versions of Lua.
In our [repo](https://github.com/haikuports/haikuports/tree/master/dev-lang/lua) we currently have 5.1.5, 5.2.4, 5.3.3 and 5.3.4.
haikuporter tries to use the most recent among those which are installed or available in the haikuports/packages/ directory.

wordgrinder includes a copy of Lua 5.1.5, so if you comment out the lib:lua and devel:lua, the Makefile will not find the any Lua in the chroot and will, consequently, try to use its own copy.
(I guess this is what you tried.)

I'm a bit short of ideas at the moment. Let's hope a Lua expert will help us :)

bach5000:
In fact, yes, I did try that, and it still failed :(
Uh, for the part that says `os.getenv("HOME") or os.getenv("USERPROFILE")`: 
https://github.com/davidgiven/wordgrinder/blob/master/src/lua/main.lua#L17
The comment above says "HOME" should be the user's home directory. Could we change this?

fbrosson:
When a program stores its prefs in the home directory with a file, e.g. `$HOME/.wordgrinder.lua` we try to patch it (or configure it or pass variables) so that it uses /boot/home/config/settings/.

`finddir -l | grep PREFERENCES` shows that `finddir B_USER_PREFERENCES_DIRECTORY` is what we need. It yelds `/boot/home/config/preferences` and this is where we should have wordgrinder store its prefs.

You can look for recipes that call `finddir B_USER_PREFERENCES_DIRECTORY` to see how this is generally done.
Hmm, for cases where you need to patch the source code, it would be even better to use the C function that is equivalent to the finddir command.

I'll see if I can find a patch (among the recipes we already have) that could serve as an example.

bach5000:
Okay.

fbrosson:
I found the call that you can try to use for patching the source code:
```
find_directory(B_USER_SETTINGS_DIRECTORY, ...
```
https://github.com/haikuports/haikuports/search?&q=find_directory(B_USER_SETTINGS_DIRECTORY

Using find+grep+sed+sort (on my local git repo) I can tell that these patches are in these directories:
```
app-editors/vim/patches
app-emulation/basiliskii/patches
dev-util/global/patches
dev-vcs/git/patches
games-action/btanks/patches
games-arcade/f1spirit/patches
games-arcade/lbreakout2/patches
games-arcade/plee_the_bear/patches
games-arcade/rocksndiamonds/patches
games-arcade/slimevolley/patches
games-arcade/super_transball/patches
games-kids/dragonmemory/patches
games-puzzle/flobopuyo/patches
games-puzzle/lmarbles/patches
games-puzzle/lpairs/patches
games-puzzle/ltris/patches
games-puzzle/mirrormagic/patches
games-puzzle/pingus/patches
games-puzzle/pipepanic/patches
games-strategy/lgeneral/patches
games-strategy/openxcom/patches
games-strategy/vcmi/patches
media-libs/flickcurl/patches
media-sound/ocp/patches
net-misc/wget/patches
sys-apps/less/patches
sys-apps/most/patches
sys-libs/slang/patches
```

bach5000:
Thanks @fbrosson!
Just to check my understanding:
this is a list of patches that patches its source so that their apps could store their prefs to `/boot/home/config/preferences`, right? Or am I completely wrong?
If I'm correct, I should take a look at these recipes and learn how they patch their sources, right?
(Sorry for all these dumb questions; my brain's not functioning)

fbrosson:
You are right. This is a list of directories where I have found one or more patches which contain the string "find_directory(B_USER_SETTINGS_DIRECTORY".
Every recipe which uses such a patch does so for the same reason; teach the program how to find the /boot/home/config/settings/ directory and use it to store its prefs. BTW, these programs use a "." in the beginning of the file name they usually store in $HOME. In Haiku, we don't keep the dot because we don't want to hide the file. It is in a directory where all files have are there for similar reasons, storing prefs.
We could hardconding the path, and there are probably a few recipes which do so, but the best method is the patch.
I'm sure that, among all those examples, there will be some (or maybe many) which will be so close to what you need for wordgrinder that you could almost do a copy/paste.
But yes, you will need to put some glue around the call, for instance, to concatenate the path with the name of the config file.
BTW, you don't need to be sorry for anything! You are helping Haiku with many tasks and this is great!

bach5000:
@fbrosson,
where would we put the `find_directory` function? I'm kind of lost now :(

Begasus:
For lua you probably need to patch the "Makefile" as it's using the default (internal 5.1) version ...

fbrosson:
Thanks @Begasus for the hint. I read something about that, but it does not seem so easy.
So I'll probably merge this recipe soon.

Thanks @bach5000 for the update!

scottmc:
We have added a task to use the find_directory(B_USER_SETTINGS_DIRECTORY to fix one of several applications that have similar settings related open issues.  Search tasks for "settings"

fbrosson:
This is cool!
BTW, I've just added another task:

> [for review] Fix the (currently broken) recipe we have for WordGrinder. Skills required: Lua, Makefile, ninja, patching

fbrosson:

PR #1912, opened 2017-12-18
Add Prison 

localanu:

PR #1914, opened 2017-12-19
extra-cmake-modules: bump to 5.41.0, switch to tar.xz.

fbrosson:


fbrosson:
arch | test results
------ | -----
x86_gcc2 | 100% tests passed, 0 tests failed out of 46
x86_64 |  100% tests passed, 0 tests failed out of 50

Bumping  extra-cmake-modules to 5.41.0 will allow #1912 to add the latest release of prison, 5.41.0, instead of 5.39.0.
prison 5.41.0 does not build if extra-cmake-modules is not >= 5.41.0.

fbrosson:
Thanks @diversys! 👍

PR #1916, opened 2017-12-19
app-arch/brotli: add port

alaviss:
Done as part of GCI 2017

Aside from the fact that the recipe looks ugly, it's working for all architecture, including secondary

### DIFF
--- a/app-arch/brotli/brotli-1.0.2.recipe
+++ b/app-arch/brotli/brotli-1.0.2.recipe
@@ -1,0 +1,149 @@
+SUMMARY="A generic-purpose lossless compression algorithm"
+DESCRIPTION="A generic-purpose lossless compression algorithm that \
+compresses data using a combination of a modern variant of the LZ77 algorithm, \
+Huffman coding and 2nd order context modeling, with a compression ratio \
+comparable to the best currently available general-purpose compression methods. \
+It is similar in speed with deflate but offers more dense compression."
+HOMEPAGE="https://github.com/google/brotli"
+COPYRIGHT="2009, 2010, 2013-2017 Brotli Authors"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/google/brotli/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="c2cf2a16646b44771a4109bb21218c8e2d952babb827796eb8a800c1f94b7422"
+
+ARCHITECTURES="x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+# For now the library will follow $portVersion,
+# but upstream said that will change in the future.
+libVersion="1.0.2"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	brotli$secondaryArchSuffix = $portVersion
+	cmd:brotli$secondaryArchSuffix = $portVersion
+	lib:libbrotlicommon$secondaryArchSuffix = $libVersionCompat
+	lib:libbrotlidec$secondaryArchSuffix = $libVersionCompat
+	lib:libbrotlienc$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	brotli${secondaryArchSuffix}_devel = $portVersion
+	devel:libbrotlicommon$secondaryArchSuffix = $libVersionCompat
+	devel:libbrotlicommon_static$secondaryArchSuffix = $libVersion
+	devel:libbrotlidec$secondaryArchSuffix = $libVersionCompat
+	devel:libbrotlidec_static$secondaryArchSuffix = $libVersion
+	devel:libbrotlienc$secondaryArchSuffix = $libVersionCompat
+	devel:libbrotlienc_static$secondaryArchSuffix = $libVersion
+	"
+REQUIRES_devel="
+	brotli$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+PYTHON_PACKAGES=(python python3)
+PYTHON_VERSIONS=(2.7 3.6)
+# We don't have python2 for secondaryArch,
+# and gcc2 doesn't support the flags passed by python3
+if [ -n "$secondaryArchSuffix" ]; then
+	unset PYTHON_PACKAGES[0]
+	unset PYTHON_VERSIONS[0]
+elif [ "$effectiveTargetArchitecture" = "x86_gcc2" ]; then
+	unset PYTHON_PACKAGES[1]
+	unset PYTHON_VERSIONS[1]
+fi
+for i in "${!PYTHON_PACKAGES[@]}"; do
+pythonPackage=${PYTHON_PACKAGES[i]}
+pythonVersion=${PYTHON_VERSIONS[$i]}
+eval "PROVIDES_${pythonPackage}=\"\
+	${portName}_$pythonPackage = $portVersion\
+	\"; \
+REQUIRES_$pythonPackage=\"\
+	haiku\n\
+	cmd:python$pythonVersion\
+	\""
+BUILD_REQUIRES="$BUILD_REQUIRES
+	setuptools_$pythonPackage"
+BUILD_PREREQUIRES="$BUILD_PREREQUIRES
+	cmd:python$pythonVersion"
+done
+
+if [ -z "$secondaryArchSuffix" ]; then
+	REQUIRES_python+="
+		lib:libpython${PYTHON_VERSIONS[0]}
+		"
+	BUILD_PREREQUIRES+="
+		devel:libpython${PYTHON_VERSIONS[0]}
+		"
+fi
+if [ "$effectiveTargetArchitecture" != "x86_gcc2" ]; then
+	REQUIRES_python3+="
+		lib:libpython${PYTHON_VERSIONS[1]}m$secondaryArchSuffix
+		"
+	BUILD_PREREQUIRES+="
+		devel:libpython${PYTHON_VERSIONS[1]}m$secondaryArchSuffix
+		"
+fi
+
+BUILD()
+{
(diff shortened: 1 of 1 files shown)


app-arch/brotli/brotli-1.0.2.recipe
```diff
@@ -0,0 +1,143 @@
+SUMMARY="Generic-purpose lossless compression algorithm"
+DESCRIPTION="A generic-purpose lossless compression algorithm that \
+compresses data using a combination of a modern variant of the LZ77 algorithm, \
+Huffman coding and 2nd order context modeling, with a compression ratio \
+comparable to the best currently available general-purpose compression methods. \
+It is similar in speed with deflate but offers more dense compression"
```
fbrosson: I did not test this recipe but I'm quite sure there are no errors.
It looks like the only way you can improve it is by adding a full stop at the end of DESCRIPTION :)
Hmm, for SUMMARY maybe you could change "Generic-purpose" to "A generic-purpose".
BTW, for SUMMARY that's OK with no full stop.

app-arch/brotli/brotli-1.0.2.recipe
```diff
@@ -0,0 +1,143 @@
...
+	haiku\n\
+	cmd:python$pythonVersion\
+	\""
+BUILD_REQUIRES="$BUILD_REQUIRES
+	setuptools_$pythonPackage"
+BUILD_PREREQUIRES="$BUILD_PREREQUIRES
+	cmd:python$pythonVersion"
+done
+
+if [[ "$effectiveTargetArchitecture" == "$targetArchitecture" ]]; then
```
fbrosson: It would be better to use sh-only syntax (instead of ksh/bash), at least for where sh is enough.
```diff
-if [[ "$effectiveTargetArchitecture" == "$targetArchitecture" ]]; then
+if [ "$effectiveTargetArchitecture" = "$targetArchitecture" ]; then
```
Remark: the test above could also be written this way:
```
```
This is because secondaryArchSuffix is only defined when we are building for a secondary arch.
That said, your test is more readable. It is also more flexible. Image someone becomes crazy and decides to build a custom hybrid Haiku with a secondary arch identical to the primary arch. Your test continues to produce correct packages, even with weird configurations ;-)

Same suggestion 8 lines below:
```diff
-if [[ "$effectiveTargetArchitecture" != "x86_gcc2" ]]; then
+if [ "$effectiveTargetArchitecture" != "x86_gcc2" ]; then
```

I'm quite sure some parts, later, are not compatible with sh, but having the changes above will reduce the "risk" of other editors switching to ksh/bash syntax (after reading this recipe) in other recipes where sh is sufficient.

alaviss:
Implemented @fbrosson suggestions

app-arch/brotli/brotli-1.0.2.recipe
```diff
@@ -0,0 +1,143 @@
...
+		$python setup.py build install \
+			--root=/ --prefix=$prefix
+
+		packageEntries $pythonPackage \
+			$prefix/lib/python*
+	done
+
+	fixPkgconfig
+	prepareInstalledDevelLibs libbrotli{common,dec,enc}{,-static}
+	packageEntries devel "$developDir"
```
fbrosson: Builds fine on x86_gcc2 primary arch. (tests fail at some point, but on x86_gcc2 this happens.)
(I haven't tested yet on x86 secondary arch, nor on x86_64.)
Would be nice to include
```
	install -d "$docDir" "$manDir"/man1 "$manDir"/man3
	install -t "$manDir"/man1 docs/brotli.1
	install -t "$manDir"/man3 docs/*.3
	install -t "$docDir" README docs/*.pdf
```
and
```diff
-	packageEntries devel "$developDir"
+	packageEntries devel \
+		"$docDir"/*.pdf \
+		"$manDir"/man3 \
+		"$developDir"
```
(Yep, it seems there is no "install-doc" Makefile target.)
Maybe it would also be nice to use the $maybe_manDir_man3 "trick" to skip the installation of the man pages on secondary arch builds.
BTW, I added quotes because I know you like them and I agree that it is better with quotes than without. This is safe in situations where the path to the haikuports tree contains special chars (e.g. spaces, tabs, etc...) and the user chooses to call "haikuporter --nochroot" if this gets fixed some day.)

alaviss: Tests fails on x86_gcc2? I would appreciate a log. Hopefully those style changes doesn't broke anything (it shouldn't).

LOL, I totally forgot about the docs :P

app-arch/brotli/brotli-1.0.2.recipe
```diff
@@ -0,0 +1,143 @@
...
+INSTALL()
+{
+	make install
+
+	for i in "${!PYTHON_PACKAGES[@]}"; do
+		pythonPackage=${PYTHON_PACKAGES[i]}
+		pythonVersion=${PYTHON_VERSIONS[$i]}
+
+		python=python$pythonVersion
+		installLocation=$prefix/lib/$python/vendor-packages/
```
fbrosson: On x86_gcc2 my brotli_x86_python3 package is installing stuff in `/system/lib/python3.6/vendor-packages/` instead of `/system/lib/x86/python3.6/vendor-packages/`, so think installLocaion should be changed to:
```diff
-		installLocation=$prefix/lib/$python/vendor-packages/
+		installLocation=$libDir/$python/vendor-packages/
```
later:
```diff
 		packageEntries $pythonPackage \
-			$prefix/lib/python*
+			$libDir/python*
```

alaviss: As weird as it seems, on x86 snd arch, packages install to [`/system/lib/python3.6/vendor-packages/`](https://github.com/haikuports/haikuports/blob/master/dev-lang/python/python3-3.6.3.recipe#L128)

app-arch/brotli/brotli-1.0.2.recipe
alaviss: Looks like only the man pages have value here. The PDF is just an comparison, and README pretty much only have instructions to build the lib

alaviss:
Packaged those man pages

app-arch/brotli/brotli-1.0.2.recipe
fbrosson: I thought about this possibility but then found the x86 directory, so I assumed it was there for a reason.
Yep, this seems weird, but I believe you.

app-arch/brotli/brotli-1.0.2.recipe
```diff
@@ -0,0 +1,149 @@
+SUMMARY="A generic-purpose lossless compression algorithm"
+DESCRIPTION="A generic-purpose lossless compression algorithm that \
+compresses data using a combination of a modern variant of the LZ77 algorithm, \
+Huffman coding and 2nd order context modeling, with a compression ratio \
+comparable to the best currently available general-purpose compression methods. \
+It is similar in speed with deflate but offers more dense compression."
+HOMEPAGE="https://github.com/google/brotli"
+COPYRIGHT="2009, 2010, 2013-2016 Brotli Authors"
```
fbrosson: s/2016/2017/

Also, it would be nice to rename the tarball by adding this line after CHECKSUM_SHA256:
```
SOURCE_FILENAME="brotli-$portVersion.tar.gz"
```

app-arch/brotli/brotli-1.0.2.recipe
```diff
@@ -0,0 +1,149 @@
...
+	unset PYTHON_VERSIONS[1]
+fi
+for i in "${!PYTHON_PACKAGES[@]}"; do
+pythonPackage=${PYTHON_PACKAGES[i]}
+pythonVersion=${PYTHON_VERSIONS[$i]}
+eval "PROVIDES_${pythonPackage}=\"\
+	${portName}_$pythonPackage = $portVersion\
+	\"; \
+REQUIRES_$pythonPackage=\"\
+	haiku\n\
```
fbrosson: brotli_python{,3} should also depend on 
```
		brotli$secondaryArchSuffix == $portVersion base
```
Hmm, since this is exactly REQUIRES_devel, we can fix the dependencies of brotli_python{,3} with the following change:
```diff
 REQUIRES_$pythonPackage=\"\
-	haiku\n\
+	$REQUIRES_devel \
 	cmd:python$pythonVersion\
 	\""
```

Well, I don't know if it is desirable to _use_ `$REQUIRES_devel`, so please feel free to use whatever alternative you wish.

app-arch/brotli/brotli-1.0.2.recipe
alaviss: It's [2016](https://github.com/google/brotli/blob/master/LICENSE#L1) :P

I don't see any compelling reason to rename the tarball, after all they're contained within their respective port folder

app-arch/brotli/brotli-1.0.2.recipe
alaviss: It wouldn't make much sense to use `$REQUIRES_devel`, looking at the context, it would looks like those are some kind of devel package (they somewhat are lol).

Do we really need to make them depends on the base package? They're pretty much self-contained themselves

app-arch/brotli/brotli-1.0.2.recipe
fbrosson: As we can see in https://github.com/google/brotli/releases Brotli 1.0.2 was released on November 29, 2017. So it cannot be 2016 :)

Regarding the tarball, there are a few arguments in favor for the rename:
* Gentoo renames the tarball;
* We already do so in many similar cases;
* A tarball without the name of the program is not very meaningful.

That said, since the recipe is still correct without the rename, it is perfectly OK to keep the original filename if you wish to. That's one of the priviledges you have when you write a great recipe!

app-arch/brotli/brotli-1.0.2.recipe
fbrosson: I might be wrong, but I'm quite sure that if you install brotli_python without brotli then you cannot use it.
If this is correct then the package must declare this dependency.
Now, in the suggestion I made, I used `$REQUIRES_devel` because it was easier to write than to copy the exact same string that was used to define REQUIRES_devel.
Of course, you are free to use any construct you wish, as long as the package has this dependency.

alaviss: You can. The package have an version of brotli within itself (`_brotli.so`)

alaviss:
Bumped copyright year

app-arch/brotli/brotli-1.0.2.recipe
fbrosson: Hmm, you are perfectly right.
So this means the dependencies are correct.
I think everything is OK then. I guess we can merge.
> Bumped copyright year

Cool! I'll merge soon.

PR #1920, opened 2017-12-19
vala, bump version

Begasus:
added defineDebugInfoPackage and devel package ...


dev-lang/vala/vala-0.36.8.recipe
```diff
@@ -0,0 +1,133 @@
...
+	cmd:valac$secondaryArchSuffix
+	cmd:vapicheck$secondaryArchSuffix
+	cmd:vapigen$secondaryArchSuffix
+	cmd:vala_gen_introspect$secondaryArchSuffix
+	cmd:vala_${portVersion%.*}$secondaryArchSuffix
+	cmd:vala_gen_introspect_${portVersion%.*}$secondaryArchSuffix
+	cmd:valac_${portVersion%.*}$secondaryArchSuffix
+	cmd:vapicheck_${portVersion%.*}$secondaryArchSuffix
+	cmd:vapigen_${portVersion%.*}$secondaryArchSuffix
+	lib:libvala_${portVersion%.*}$secondaryArchSuffix = 0.0.0 compat >= 0
```
Begasus: @fbrosson didn't test the $libVersion here, but it probably would mean using something like $cmdVersion too (would seen strange to use $libVersion in the cmd:* section) ?

dev-lang/vala/vala-0.36.8.recipe
```diff
@@ -0,0 +1,133 @@
...
+PROVIDES_common="
+	vala${secondaryArchSuffix}_common = $portVersion
+	"
+REQUIRES_common="
+	haiku$secondaryArchSuffix
+	"
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	#lib:libvala_${portVersion%.*}$secondaryArchSuffix = 0.0.0 compat >= 0
+	"
```
Begasus: I'm not able to add the lib into the devel package, maybe missing the *.a lib is the reason here?

dev-lang/vala/vala-0.36.8.recipe
```diff
@@ -0,0 +1,133 @@
...
+	# for some reason it can't install the library
+	prepareInstalledDevelLibs #libvala-${portVersion%.*}
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $manDir
+		maybe_manDir=
+	else
+		maybe_manDir=$manDir
+	fi
```
Begasus: As it is now, there are no manpages installed, can't use the x86_gcc2 arch as that isn't supported, any hints here?

dev-lang/vala/vala-0.36.8.recipe
fbrosson: When `cmd:` items have a version, it can only be $portVersion. So a $cmdVersion is not necessary on any package.

Begasus: Ok, $libVersion wouldn't do any good here also, as it doesn't provide the .so version, so best leave it as is then? :)

pulkomandy: I would not be so sure about that. It could make sense to version a command if there are API (well command line interface) changes to it and other packages rely on a specific version. But admitedly this is probably very unusual.

dev-lang/vala/vala-0.36.8.recipe
korli: Shouldn't it be devel: here?

Begasus:
Bumped to 0.36.11, this one is still ok (not tested on x86_64 and didn't run the TEST() yet) ...

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -0,0 +1,133 @@
...
+		rm -rf $manDir
+		maybe_manDir=
+	else
+		maybe_manDir=$manDir
+	fi
+
+	#devel and common packages
+	packageEntries devel \
+		$developDir \
+		$maybe_manDir
```
fbrosson:
```diff
-	if [ -n "$secondaryArchSuffix" ]; then
-		rm -rf $manDir
-		maybe_manDir=
-	else
-		maybe_manDir=$manDir
-	fi
-
 	packageEntries devel \
-		$developDir \
+		$developDir
-		$maybe_manDir
```

And at the end if `INSTALL()`:

```diff
 	rmdir $dataDir/aclocal
-
-	rm -r $documentationDir
 }
```

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -0,0 +1,125 @@
...
+
+	rm -f $libDir/*.la
+
+	# for some reason it can't install the library
+	prepareInstalledDevelLibs #libvala-${portVersion%.*}
+	fixPkgconfig
+
+	#devel and common packages
+	packageEntries devel \
+		$developDir
```
fbrosson:
```diff
 	packageEntries devel \
-		$developDir
+		$developDir \
 		$manDir
```

Begasus: Bugger ...

Begasus:
PS,  if things are ok I could remove the older recipe?

fbrosson:
Shall we drop the recipe for the previous version, `vala-0.36.4.recipe` ?
No recipe needs `lib:libvala*`.

(But first we need to check that those recipes which need any of the `cmd:` items provided by vala also build fine with vala-0.36.11.)

Begasus:
Only reference I could spot (on github here) is for libsoup which is using --enable-vala=no as an option ...

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -28,26 +28,35 @@ PROVIDES="
...
+PROVIDES_common="
+	vala${secondaryArchSuffix}_common = $portVersion
+	"
 REQUIRES_common="
 	haiku$secondaryArchSuffix
 	"
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	devel:libvala_${portVersion%.*}$secondaryArchSuffix = 0.0.0 compat >= 0
+	"
```
Begasus: Not sure here, there is no lib in the develop package, only the *.pc files and headers, would it be safe to remove this line?

fbrosson: I don't know.
Does haikuporter complain if you remove that line and run `haikuporter -S vala` ?
If not, then I'd say that you can remove the line.

Begasus:
pkg-config-x86 libvala-0.36 --cflags is working ok (also --libs)
there is also vapigen pkgconfig file, not getting any output from then when checking with pkg-config ...

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -75,18 +98,24 @@ INSTALL()
...
+
 	packageEntries common \
 		$dataDir/aclocal/vapigen.m4 \
 		$dataDir/aclocal/vala.m4 \
 		$dataDir/vala/Makefile.vapigen
 
 	# Avoid having an empty data/aclocal/ in vala$secondaryArchSuffix
 	rmdir $dataDir/aclocal
+
+	rmdir $documentationDir
```
Begasus: Having it in or not doesn't seem to make any difference?

Begasus:
All tests passed on x86_gcc2 ...

dev-lang/vala/vala-0.36.11.recipe
fbrosson: Exact. The resulting package is the same.
That's because when haikuporter sees that the directory is empty (after `INSTALL()` has done its job) it does the `rmdir` for us.

That said, having the `rmdir $documentationDir` (near the end of `INSTALL()`) allows to trigger an error in case upstream decides to add something in a new release.

BTW, each choice has its small advantages and its small downsides, so I don't have any preference :)

Begasus: Looks ok to so far then :)

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -64,6 +73,20 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+defineDebugInfoPackage libvala$secondaryArchSuffix \
```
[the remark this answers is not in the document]
fbrosson: BTW, installing the debuginfo package is a good method to detect errors like this one.

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -75,18 +98,24 @@ INSTALL()
...
-	# remove libtool library files
-	rm $libDir/lib*.la
+	rm -f $libDir/*.la
 
 	fixPkgconfig
 
+	#devel and common packages
+	packageEntries devel \
+		$developDir \
+		$manDir
```
fbrosson: Well, after all, I'm not sure we should move the man to the devel package.
Here all man pages are in man1. And all `cmd:*` items are in the base package.
So I'd say that the man pages should also be in the base package.

That said, if we are 100% sure that all `cmd:` items may be put in the devel package, then it would be OK to move both `$manDir` and `$binDir`. But that might be risky, so the easiest solution is to keep the `$manDir` in the base package.

Begasus: I already thought it was strange, not having the manpages installed together with the cmd's ... :)

fbrosson:
Same results for me, on both x86 second arch and x86_64:
```
make  check-TESTS
make[4]: Entering directory '/sources/vala-0.36.11/tests'
TEST: Building...                                                      OK
...
All 295 tests passed
TEST: Building...                                                      OK
  /nullability/bug611223:                                              OK
All 1 tests passed
```

Does someone think we should abandon the _common subpackage and merge its contents into the base package?
(The _common subpackage was created in this commit: fb66e513a2ec009d2804f70fe8aa1400397ee8c5)

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -75,18 +98,23 @@ INSTALL()
...
+
 	packageEntries common \
 		$dataDir/aclocal/vapigen.m4 \
 		$dataDir/aclocal/vala.m4 \
 		$dataDir/vala/Makefile.vapigen
 
 	# Avoid having an empty data/aclocal/ in vala$secondaryArchSuffix
 	rmdir $dataDir/aclocal
+
+	#rmdir $documentationDir
```
fbrosson: Now that we have `$manDir` in the base package, it would not be possible to `rmdir $documentationDir`.
So I'd drop that line completely. (No need to keep it, even commented out.)

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -64,6 +73,20 @@ BUILD_PREREQUIRES="
...
+	$binDir/vapigen \
+	$binDir/vala-${portVersion%.*} \
+	$binDir/valac-${portVersion%.*} \
+	$binDir/vapicheck-${portVersion%.*} \
+	$binDir/vapigen-${portVersion%.*}
+	# looks like it doesn't take more then one "-" here
+	#$binDir/vala-gen-introspect \
+	#$binDir/vala-gen-introspect-${portVersion%.*} \
+
 BUILD()
```
fbrosson: I'd change this in `TEST()`:
```diff
 TEST()
 {
- 	sed -i -e 's/-lm/-lroot/g' $sourceDir/tests/testrunner.sh
 	make check
 }
```
and insert a `PATCH()` before `BUILD()`:
```
PATCH()
{
 	sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
}
```
(I dropped the `g` and the `$sourceDir/`)

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -28,26 +28,35 @@ PROVIDES="
...
 	"
+PROVIDES_common="
+	vala${secondaryArchSuffix}_common = $portVersion
+	"
 REQUIRES_common="
 	haiku$secondaryArchSuffix
 	"
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	#devel:libvala_${portVersion%.*}$secondaryArchSuffix = 0.0.0 compat >= 0
```
fbrosson:
```diff
-	#devel:libvala_${portVersion%.*}$secondaryArchSuffix = 0.0.0 compat >= 0
```
Shall we remove this line?
It's not needed at all because:
1) we are dropping the libtool file, and
2) we are not building the static lib.

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -64,6 +73,20 @@ BUILD_PREREQUIRES="
...
+	$binDir/valac \
+	$binDir/vapicheck \
+	$binDir/vapigen \
+	$binDir/vala-${portVersion%.*} \
+	$binDir/valac-${portVersion%.*} \
+	$binDir/vapicheck-${portVersion%.*} \
+	$binDir/vapigen-${portVersion%.*}
+	# looks like it doesn't take more then one "-" here
+	#$binDir/vala-gen-introspect \
+	#$binDir/vala-gen-introspect-${portVersion%.*} \
```
fbrosson:
```diff
-	# looks like it doesn't take more then one "-" here
-	#$binDir/vala-gen-introspect \
-	#$binDir/vala-gen-introspect-${portVersion%.*} \
```
`vala-gen-introspect` is a symlink to `vala-gen-introspect-0.36`, which is itself just a shell script.
(So this is why it can't be included in the debuginfo package.)

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -28,26 +28,35 @@ PROVIDES="
...
 REQUIRES_common="
 	haiku$secondaryArchSuffix
 	"
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	#devel:libvala_${portVersion%.*}$secondaryArchSuffix = 0.0.0 compat >= 0
+	"
+REQUIRES_devel="
+	vala$secondaryArchSuffix == $portVersion base
+	"
```
fbrosson: Because of this:
```bash
grep Requires /system/develop/lib/pkgconfig/libvala-0.36.pc 
Requires: glib-2.0 gobject-2.0
```
I think it might be appropriate to mention glib and gobject in `REQUIRES_devel`:
```diff
 REQUIRES_devel="
 	vala$secondaryArchSuffix == $portVersion base
+	devel:libglib_2.0
+	devel:libgobject_2.0
 	"
```
(But I'l not sure about this, as there could be downsides.)
Let's see what other contributors think about this.

Begasus: Been looking into updating cairo, this can also use glib2 (which in it turn produces libcairo-gobject, which could be used to fix gobject-introspection tests) :)

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -64,6 +73,20 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+defineDebugInfoPackage vala$secondaryArchSuffix \
+	$libDir/libvala-${portVersion%.*}.so.0.0.0 \
```
fbrosson: Shall we use a `libVersion` (and `libVersionCompat`)?
```diff
-	$libDir/libvala-${portVersion%.*}.so.0.0.0
+	$libDir/libvala-${portVersion%.*}.so.$libVersion
```
(Of course, if you define `libVersion` you will want to replace all other occurences of `0.0.0 compat >= 0` by `$libVersionCompat` and `0.0.0` by `$libVersion`.)

Begasus: Also leftover from last version, haven't looked into libVersion here yet :)

dev-lang/vala/vala-0.36.11.recipe
Begasus: PS, haven't seen any lib so far using devel:* from other libraries aside from it's own?

dev-lang/vala/vala-0.36.11.recipe
```diff
@@ -28,26 +31,28 @@ PROVIDES="
...
 REQUIRES="
 	haiku$secondaryArchSuffix
 	glib2${secondaryArchSuffix} >= 2.32.0
-	vala${secondaryArchSuffix}_common == $portVersion
+	lib:libdbus_1$secondaryArchSuffix
 	lib:libffi$secondaryArchSuffix
 	lib:libiconv$secondaryArchSuffix
 	lib:libintl$secondaryArchSuffix
 	"
-REQUIRES_common="
```
fbrosson: I'd recommend adding this (right after the `REQUIRES`):
```
REPLACES="
	vala${secondaryArchSuffix}_common
	"
```
This would be handy for users upgrading from `vala-0.36.4-1` (or `vala_x86-0.36.4-1`) to `vala-0.36.11-1` (or `vala_x86-0.36.11-1`) as it would instruct pkgman to delete `vala_common-0.36.4-1` (or `vala_x86_common-0.36.4-1`).

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -4,17 +4,20 @@ programming language features to GNOME developers without imposing any \
...
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
 	vala$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 libVersion="0.0.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"

 PROVIDES="
-	vala$secondaryArchSuffix = $portVersion
+	vala$secondaryArchSuffix = $portVersionCompat
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -28,26 +31,32 @@ PROVIDES="
...
 	glib2${secondaryArchSuffix} >= 2.32.0
-	vala${secondaryArchSuffix}_common == $portVersion
+	lib:libdbus_1$secondaryArchSuffix
 	lib:libffi$secondaryArchSuffix
 	lib:libiconv$secondaryArchSuffix
 	lib:libintl$secondaryArchSuffix
 	"
-REQUIRES_common="
-	haiku$secondaryArchSuffix
+
```
fbrosson:
```diff
 	lib:libintl$secondaryArchSuffix
 	"
-
 REPLACES="
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -28,26 +31,32 @@ PROVIDES="
...
+REPLACES="
+	vala${secondaryArchSuffix}_common
+	"
+
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
```
fbrosson:
```diff
-	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	vala${secondaryArchSuffix}_devel = $portVersionCompat
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -64,6 +73,22 @@ BUILD_PREREQUIRES="
...
+	$binDir/vapicheck \
+	$binDir/vapigen \
+	$binDir/vala-${portVersion%.*} \
+	$binDir/valac-${portVersion%.*} \
+	$binDir/vapicheck-${portVersion%.*} \
+	$binDir/vapigen-${portVersion%.*}
+
+PATCH()
+{
+		sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
```
fbrosson:
```diff
-		sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
+	sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -75,22 +100,16 @@ INSTALL()
...
-	packageEntries common \
-		$dataDir/aclocal/vapigen.m4 \
-		$dataDir/aclocal/vala.m4 \
-		$dataDir/vala/Makefile.vapigen
-
-	# Avoid having an empty data/aclocal/ in vala$secondaryArchSuffix
-	rmdir $dataDir/aclocal
+	#devel and common packages
+	packageEntries devel \
+		$developDir
```
fbrosson: I've just built vala-0.36.12 on x86_gcc2 (using x86 second arch) and it builds fine.
But I think some files should be moved to the devel package:
```diff
-	#devel and common packages
+	# devel package
 	packageEntries devel \
-		$developDir
+		$developDir \
+		$dataDir/aclocal \
+		$dataDir/vala/Makefile.vapigen
```
It might also be appropriate to move some runtimes, like valac, vapicheck and vapigen, but I'd need to look closer.

Begasus:
Updated, removed patch for lm/lroot, tests still run ok ...

Begasus:
All 309 tests passed
  TEST: Building...                                                      OK
    /nullability/bug611223:                                              OK
  All 1 tests passed

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -28,26 +32,33 @@ PROVIDES="
 	cmd:valac_${portVersion%.*}$secondaryArchSuffix
 	cmd:vapicheck_${portVersion%.*}$secondaryArchSuffix
 	cmd:vapigen_${portVersion%.*}$secondaryArchSuffix
-	lib:libvala_${portVersion%.*}$secondaryArchSuffix
-	"
-PROVIDES_common="
-	vala${secondaryArchSuffix}_common = $portVersion
+	lib:libvala_${portVersion%.*}$secondaryArchSuffix = $libVersionCompat
```
fbrosson:
```diff
 	cmd:vapigen$secondaryArchSuffix
 	cmd:vala_gen_introspect$secondaryArchSuffix
-	cmd:vala_${portVersion%.*}$secondaryArchSuffix
+	cmd:vala_$portVers$secondaryArchSuffix
 	cmd:vala_gen_introspect_${portVersion%.*}$secondaryArchSuffix
-	cmd:valac_${portVersion%.*}$secondaryArchSuffix
-	cmd:vapicheck_${portVersion%.*}$secondaryArchSuffix
-	cmd:vapigen_${portVersion%.*}$secondaryArchSuffix
-	lib:libvala_${portVersion%.*}$secondaryArchSuffix = $libVersionCompat
+	cmd:valac_$portVers$secondaryArchSuffix
+	cmd:vapicheck_$portVers$secondaryArchSuffix
+	cmd:vapigen_$portVers$secondaryArchSuffix
+	lib:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
 	"
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -28,26 +32,33 @@ PROVIDES="
...
 	"
-REQUIRES_common="
-	haiku$secondaryArchSuffix
+REPLACES="
+	vala${secondaryArchSuffix}_common
+	"
+
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersionCompat
+	"
```
fbrosson:
```diff
 PROVIDES_devel="
 	vala${secondaryArchSuffix}_devel = $portVersionCompat
+	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
 	"
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -64,6 +75,17 @@ BUILD_PREREQUIRES="
...
+defineDebugInfoPackage vala$secondaryArchSuffix \
+	$libDir/libvala-${portVersion%.*}.so.$libVersion \
+	$binDir/vala \
+	$binDir/valac \
+	$binDir/vapicheck \
+	$binDir/vapigen \
+	$binDir/vala-${portVersion%.*} \
+	$binDir/valac-${portVersion%.*} \
+	$binDir/vapicheck-${portVersion%.*} \
+	$binDir/vapigen-${portVersion%.*}
```
fbrosson: Suggestion:
```diff
 defineDebugInfoPackage vala$secondaryArchSuffix \
-	$libDir/libvala-${portVersion%.*}.so.$libVersion \
 	$binDir/vala \
 	$binDir/valac \
 	$binDir/vapicheck \
 	$binDir/vapigen \
-	$binDir/vala-${portVersion%.*} \
-	$binDir/valac-${portVersion%.*} \
-	$binDir/vapicheck-${portVersion%.*} \
-	$binDir/vapigen-${portVersion%.*}
+	$binDir/vala-$portVers \
+	$binDir/valac-$portVers \
+	$binDir/vapicheck-$portVers \
+	$binDir/vapigen-$portVers \
+	$libDir/libvala-$portVers.so.$libVersion
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -75,22 +97,18 @@ INSTALL()
...
 	fixPkgconfig
 
-	packageEntries common \
-		$dataDir/aclocal/vapigen.m4 \
-		$dataDir/aclocal/vala.m4 \
+	# devel package
+	packageEntries devel \
+		$developDir \
+		$dataDir/aclocal \
 		$dataDir/vala/Makefile.vapigen
```
fbrosson:
```diff
+	prepareInstalledDevelLib libvala-$portVers
 	fixPkgconfig
 
 	# devel package
 	packageEntries devel \
 		$developDir \
 		$dataDir/aclocal \
-		$dataDir/vala/Makefile.vapigen
+		$dataDir/vala \
+		$dataDir/vala-$portVers
 }
```
I haven't checked it if we need to add the call to `prepareInstalledDevelLib`.
I also don't know if it's OK to put `$dataDir/vala-0.36` in the devel package.
BTW, I guess it would be nice to move to the devel package some commands (or maybe all) that are in the base package:
- [ ] vala
- [x] valac
- [x] vapicheck
- [x] vapigen

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -4,20 +4,24 @@ programming language features to GNOME developers without imposing any \
...
 SOURCE_DIR="vala-$portVersion"
 PATCHES="vala-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
```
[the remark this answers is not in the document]
Begasus: You keep comming up with variables ... iirc I didn't call for portVersionCompat in the first place ...

dev-lang/vala/vala-0.36.12.recipe
Begasus: Iirc this was removed by comment?

dev-lang/vala/vala-0.36.12.recipe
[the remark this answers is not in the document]
fbrosson: I'm suggesting `portVers` because there are a lot of occurences of `"${portVersion%.*}"`.
I think the suggestion makes the recipe more eye-friendly, but this is just my opinion.
As indicated, this is just a suggestion. So please feel free to do whatever you wish.

dev-lang/vala/vala-0.36.12.recipe
fbrosson: I don't remember why it was removed, if that was the case.
So please ignore my suggestion if you wish, because I'm not sure about this.

Begasus: This is getting too confusing ... it was you that told me to remove the line (there are no lib's in the devel package ... )

dev-lang/vala/vala-0.36.12.recipe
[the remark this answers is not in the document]
Begasus: I don't see a benefit ... you provide the same ...

dev-lang/vala/vala-0.36.12.recipe
fbrosson: Oops, in that case I owe you all my apologies.
I suspect that I had not noticed that the recipe was not calling `prepareInstalledDevelLib libvala-$portVers` (before the call to `	fixPkgconfig`).
And without this call we cannot have the `devel:` in `PROVIDES_devel`.
But with call, the `devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat` in `PROVIDES_devel` is OK and `haikuporter -S` is happy.

Right now I'm trying to see if we can move `$binDir` to the devel package. I think all we need in the base package is the shared lib.

Begasus: Updated, not sure yet how to move the cmd's to the devel package ...

fbrosson: Cool, thanks for the update!
Yep, I haven't found yet how to use `defineDebugInfoPackage` on its files are not in the base package. So, unless someone has the answer, we can keep the `cmd:` entries in the base package.

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -0,0 +1,114 @@
...
+
+defineDebugInfoPackage vala$secondaryArchSuffix \
+	$binDir/vala \
+	$binDir/valac \
+	$binDir/vapicheck \
+	$binDir/vapigen \
+	$binDir/vala-$portVers \
+	$binDir/valac-$portVers \
+	$binDir/vapicheck-$portVers \
+	$binDir/vapigen-$portVers
```
fbrosson: Oops, you ate the lib ;) That's my fault, sorry...
```diff
-	$binDir/vapigen-$portVers
-	$binDir/vapigen-$portVers \
+	$libDir/libvala-$portVers.so.$libVersion
```

dev-lang/vala/vala-0.36.12.recipe
```diff
@@ -0,0 +1,114 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/*.la
+
+	fixPkgconfig
```
fbrosson: If you insert this:
```diff
+	prepareInstalledDevelLib libvala-$portVers
 	fixPkgconfig
```
then you can add a `devel:libvala_$portVers` in `PROVIDES_devel`:
```diff
 PROVIDES_devel="
 	vala${secondaryArchSuffix}_devel = $portVersionCompat
+	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
 	"
```

I think this is what we should do, but I might be wrong.

Begasus: Checking ...

dev-lang/vala/vala-0.36.12.recipe
fbrosson: BTW, what do you think of this variant?
```
 defineDebugInfoPackage vala$secondaryArchSuffix \
 	$binDir/vala \
	$binDir/vala-$portVers \
 	$binDir/valac \
	$binDir/valac-$portVers \
 	$binDir/vapicheck \
	$binDir/vapicheck-$portVers \
 	$binDir/vapigen \
	$binDir/vapigen-$portVers \
	$libDir/libvala-$portVers.so.$libVersion
```
(I've just move some lines so that similar lines together.)

dev-lang/vala/vala-0.36.12.recipe
Begasus: ok, this seems to work +1 but I also have vapigen.pc and vapigen-0.36.pc for the pkg-config files in the package, if I look at the content it looks ok, but "pkg-config-x86 vapigen --cflags" (--libs) doesn't produce any output?

fbrosson: Did you install both the base package and the devel subpackage?
If you did not install the devel subpackage then pkg-config will not find some files it needs.
BTW, Debian uses a very granular packaging: For vala, they have several binary pacakges, and one of them only has the shared lib. So I guess it would be safe to only have the shared lib in the base package.
Well, for now we need to keep the `cmd:` in the base package.

Hmm, it looks like my idea of moving some data files to the devel package is not appripriatte. (Sorry!)
After all, if a recipe has `cmd:vapigen` (and `cmd:pkg_config`) in its `BUILD_PREREQUIRES` then pkg-config and vapigen should be usable with no other dependencies.

It looks like we really need to check that the split is OK before the merge.
I did not do it, (but I did mention that I was not sure about the move :-)
So, sorry again for the bad suggestions, if it turns out that the moves are a bad idea.

fbrosson: OK.
So the question here is:
_Is it OK if `pkg-config-x86 vapigen --libs` does not produce any output?_
Good question.
(I don't know the answer.)

Regading the data files, it looks like the vapi files are required by valac.
I wonder if, after all, we should not just drop the devel package and put everything (except the debuginfo) in the base package.

fbrosson: Are any vala experts reading this thread?
What would you recommend, in particular, about the split?

fbrosson: It turns out that we never had a devel subpackage for vala. The only subpackage we've had is the vala_common.
So, shall we just drop these blocs?
```
PROVIDES_devel="
	vala${secondaryArchSuffix}_devel = $portVersionCompat
	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
	"
REQUIRES_devel="
	vala$secondaryArchSuffix == $portVersion base
	"
```
and
```
	prepareInstalledDevelLib libvala-$portVers
	fixPkgconfig

	# devel package
	packageEntries devel \
		$developDir \
		$dataDir/aclocal \
		$dataDir/vala \
		$dataDir/vala-$portVers
```

Begasus: Even after removing the devel package output is still the same ...

fbrosson: This is probably a bit exessive, but at least we know that we won't have any issues caused by an incorrect split.
BTW, we don't even need to have any `REPLACES` here, since any past vala_devel would be automatically removed when the user will upgrade vala.

Begasus: Rebuilding from scratch to check ...

fbrosson: Well, in that case I think we can assume that it is normal to have no output for `pkg-config-x86 vapigen --libs`.
What this means, if I'm not mistaken, is that GUI apps built with vala don't need libvala.
I guess the shared lib is only used by the runtimes provided by vala itself.

Begasus: Ok, build is still ok (withouth devel package) but out for "pkg-config-x86 vapigen --cflags" still doesn't produce any output ...

fbrosson: Well, I think this is normal.
I'd say that when a GUI app is built with vala, the only thing that is needed at build time are the headers.

BTW, I just realized that we need to keep the call to `fixPkgconfig`.
(I'm checking this right now.)
Without this, the `*.pc` files would be in `/boot/system/lib/pkgconfig/`, but we need them to be in `/boot/system/develop/lib/pkgconfig/` otherwise pkgconfig won't find them.

fbrosson: It looks like all we need in `INSTALL()` is:
```
INSTALL()
{
	make install

	rm -f $libDir/*.la

	fixPkgconfig
}
```

dev-lang/vala/vala-0.36.12.recipe
alaviss: @fbrosson This should work:

```bash
develPrefix=$(getPackagePrefix vala${secondaryArchSuffix}_devel)
defineDebugInfoPackage vala${secondaryArchSuffix}_devel \
	"$develPrefix/$relativeBinDir/<binaries here>"

Comment (fbrosson) [reply dev-lang/vala/vala-0.36.12.recipe]: Cool, thanks!
I'll try it.

BTW, what I have been trying to do, some time ago, was to use the `--directory` option of `defineDebugInfoPackage`, but I did not figure out how this is supposed to be used.

**Comment by fbrosson** [comment]
Cool, thanks @Begasus for your update. The recipe looks OK and build fine on x86_64 and x86 secondary arch.

Regarding the patchset, there are a few offsets mismatches:
```
patching file tests/Makefile.am
Hunk #1 succeeded at 316 (offset 31 lines).
patching file tests/Makefile.am
Hunk #1 succeeded at 316 (offset 31 lines).
patching file tests/Makefile.am
Hunk #1 succeeded at 126 (offset 10 lines).
```
This is just a warning, and the patchset works fine.

I have already regenerated the patchset and there are only a few changes:
```diff
diff vala-0.36.12.patchset vala-0.36.12.patchset.sync
8c8
< index 85780c1..91b3819 100644
---
> index 41bd338..52635c0 100644
11c11
< @@ -285,21 +285,6 @@ TESTS = \
---
> @@ -316,21 +316,6 @@ TESTS = \
44c44
< index 91b3819..9260d7b 100644
---
> index 52635c0..119d5a6 100644
47c47
< @@ -285,12 +285,6 @@ TESTS = \
---
> @@ -316,12 +316,6 @@ TESTS = \
71c71
< index 9260d7b..b1a530f 100644
---
> index 119d5a6..d4cfd3e 100644
74c74
< @@ -116,7 +116,6 @@ TESTS = \
---
> @@ -126,7 +126,6 @@ TESTS = \
```

Would like to regenerate the patchset to get rid of the warnings?
Again, that's not required. It's just recommended.

Hmm, if you agree I can push a tiny commit (with these small changes) to your PR, just before the merge.

Begasus:
It's ok by me if you push the commit into this one @fbrosson

Begasus:
Thanks! :+1:

PR #1922, opened 2017-12-19
Modified recipe for mini_xml-2.6

Lan72:
The address of the current download and the source folder has been fixed.


diversys:
This has dreamchess changes in which it shouldn't.

scottmc:
seems this recipe has since been updated, so this pull request is no longer needed.

PR #1932, opened 2017-12-21
htmltidy, bump version, fix build

Begasus:
Build for gcc2 is broken due to unbalanced #endif's and parse errors ...

### DIFF
--- a/app-text/htmltidy/tidy57-5.7.0~git.recipe
+++ b/app-text/htmltidy/tidy57-5.7.0~git.recipe
@@ -1,0 +1,90 @@
+SUMMARY="The granddaddy of HTML tools, with support for modern standards"
+DESCRIPTION="Tidy tidies HTML and XML. It can tidy your documents by itself, \
+and developers can easily integrate its features into even more powerful tools."
+HOMEPAGE="http://www.html-tidy.org/"
+COPYRIGHT="1998-2015 World Wide Web Consortium"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="28be79db50d2ecf6730068167d8ed23987fa8d59"
+SOURCE_URI="https://github.com/htacg/tidy-html5/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="5fddac299a385203e06a2158df3acd68aa83ffd10d5bdf73b9e7f2913dee1a88"
+SOURCE_FILENAME="tidy-html5-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="tidy-html5-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion=5.7.0
+libVersionCompat="$libVersion compat >= 5"
+
+PROVIDES="
+	tidy57$secondaryArchSuffix = $portVersion
+	lib:libtidy$secondaryArchSuffix = $libVersionCompat
+	"
+if [ -z "$secondaryArchSuffix" ]; then
+PROVIDES="$PROVIDES
+	cmd:tidy = $portVersion
+	"
+fi
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+CONFLICTS="
+	tidy$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	tidy57${secondaryArchSuffix}_devel = $portVersion
+	devel:libtidy$secondaryArchSuffix = $libVersion
+	devel:libtidys$secondaryArchSuffix
+	"
+REQUIRES_devel="
+	tidy57$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_devel="
+	tidy${secondaryArchSuffix}_devel
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:xsltproc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage tidy57$secondaryArchSuffix \
+	$libDir/libtidy.so.$libVersion
+
+BUILD()
+{
+	cd build/cmake
+
+	cmake ../.. -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX:PATH=$prefix \
+		-DLIB_INSTALL_DIR=$libDir \
+		-DINCLUDE_INSTALL_DIR=$includeDir \
+		-DMAN_INSTALL_DIR=$manDir/man1 \
+		-DBIN_INSTALL_DIR=$binDir
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make -C build/cmake install
+
+	prepareInstalledDevelLibs libtidy libtidys
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $prefix/bin
+		rm -rf $manDir
+	fi
+
+	# devel package
+	packageEntries devel \
+		$developDir
+}


Begasus:
Not tested on x86_64 yet

app-text/htmltidy/patches/tidy-5.6.0.patchset
```diff
@@ -0,0 +1,60 @@
...
++
+ /* === Convenience defines for Cygwin platforms === */
+ 
+ #if defined(__CYGWIN__)
+@@ -428,7 +437,7 @@ extern "C" {
+ #if PRESERVE_FILE_TIMES
+ 
+ #  ifndef HAS_FUTIME
+-#    if defined(CYGWIN_OS) || defined(BE_OS) || defined(OS2_OS) || defined(HPUX_OS) || defined(SOLARIS_OS) || defined(LINUX_OS) || defined(BSD_BASED_OS) || defined(MAC_OS) || defined(__MSL__) || defined(IRIX_OS) || defined(AIX_OS) || defined(__BORLANDC__) || defined(__GLIBC__)
++#    if defined(CYGWIN_OS) || defined(BE_OS) || defined(OS2_OS) || defined(HPUX_OS) || defined(SOLARIS_OS) || defined(LINUX_OS) || defined(BSD_BASED_OS) || defined(MAC_OS) || defined(__MSL__) || defined(IRIX_OS) || defined(AIX_OS) || defined(__BORLANDC__) || defined(__GLIBC__) || defined(__HAIKU__)
```
fbrosson: I'd put `__HAIKU__` right after the name of the last OS in that list, `AIX_OS`, and before the remaining names, which are not related to particular OSes: `__BORLANDC__` and `__GLIBC__`.
```diff
-+#    if defined(CYGWIN_OS) || defined(BE_OS) || defined(OS2_OS) || defined(HPUX_OS) || defined(SOLARIS_OS) || defined(LINUX_OS) || defined(BSD_BASED_OS) || defined(MAC_OS) || defined(__MSL__) || defined(IRIX_OS) || defined(AIX_OS) || defined(__BORLANDC__) || defined(__GLIBC__) || defined(__HAIKU__)
++#    if defined(CYGWIN_OS) || defined(BE_OS) || defined(OS2_OS) || defined(HPUX_OS) || defined(SOLARIS_OS) || defined(LINUX_OS) || defined(BSD_BASED_OS) || defined(MAC_OS) || defined(__MSL__) || defined(IRIX_OS) || defined(AIX_OS) || defined(__HAIKU__) || defined(__BORLANDC__) || defined(__GLIBC__)
```

app-text/htmltidy/tidy-5.6.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+	tidy$secondaryArchSuffix = $portVersion
+	cmd:tidy$commandSuffix = $portVersion
+	lib:libtidy$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+PROVIDES_devel="
+	devel:libtidy$secondaryArchSuffix = $libVersionCompat
+	devel:libtidys$secondaryArchSuffix
```
fbrosson: I think it should be:
```diff
-	devel:libtidys$secondaryArchSuffix
+	devel:libtidys$secondaryArchSuffix = $libVersion
```

app-text/htmltidy/tidy-5.6.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+	"
+REQUIRES_devel="
+	tidy$secondaryArchSuffix == $portVersion base
+	"
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:cmake
```
fbrosson:
```diff
 BUILD_PREREQUIRES="
+	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
-	cmd:cmake
 	cmd:make
 	"
```

app-text/htmltidy/tidy-5.6.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+
+	cmake ../.. -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX:PATH=$prefix \
+		-DLIB_INSTALL_DIR=$libDir \
+		-DINCLUDE_INSTALL_DIR=$includeDir \
+		-DBIN_INSTALL_DIR=$commandBinDir
+
+	make $jobArgs
+}
+INSTALL()
```
fbrosson:
```diff
 }
+
 INSTALL()
 {
```

app-text/htmltidy/tidy-5.6.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+	make install
+
+	prepareInstalledDevelLibs libtidy libtidys
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir
+}
+TEST()
```
fbrosson:
```diff
 }
+
 TEST()
 {
```

app-text/htmltidy/tidy-5.6.0.recipe
Begasus: it's a *.a lib, no version there?

app-text/htmltidy/tidy-5.6.0.recipe
Begasus: will fix :)

app-text/htmltidy/tidy-5.6.0.recipe
fbrosson: OK then. If there is no libtidy**s**.so (or one with no .so.x.y.z) then it's ok to have no version here.

Begasus: correct :)

app-text/htmltidy/tidy-5.6.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+CHECKSUM_SHA256="08a63bba3d9e7618d1570b4ecd6a7daa83c8e18a41c82455b6308bc11fe34958"
+SOURCE_DIR="tidy-html5-$portVersion"
+PATCHES="tidy-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
```
fbrosson: 3 cases:

1) the libs build on x86_gcc2 primary arch, but the `cmd:tidy` does not.
In this case you should not have the `cmd:tidy` in PROVIDES when building for x86_gcc2 primary arch.

2) both the libs and the cmd build (and work) fine on x86_gcc2 primary arch.
In this case there is no need for commandSuffix and commandBinDir. You can simply drop $binDir when building for a secondary arch.

3) the recipe does not build on x86_gcc2.
In this case you would have to mark this arch as broken.
(But I guess it builds)

Begasus: 3, forgot to mark it as broken, created an issue for this on htmltidy github ... will fix :)

app-text/htmltidy/tidy-5.6.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage libtidy$secondaryArchSuffix \
+		$libDir/libtidy.so.$libVersion
```
fbrosson: If you have a libtidys.so (and I guess you have) then you can add it here:
```diff
 defineDebugInfoPackage libtidy$secondaryArchSuffix \
-		$libDir/libtidy.so.$libVersion
+	$libDir/libtidy.so.$libVersion \
+	$libDir/libtidys.so
```

Begasus: no libtidys.so, only a libtidys.a

fbrosson: OK. So you only need to remove a tab :-)

app-text/htmltidy/tidy-5.6.0.recipe
fbrosson: Hmm, if there is no libtidy**s**.so then you are correct: lib:libtidy**s** should _not_ be in PROVIDES.

app-text/htmltidy/tidy-5.6.0.recipe
Begasus: jikes :)

Begasus:
@fbrosson if you got some time, could you check out https://github.com/htacg/tidy-html5/issues/663? 
Have been looking up on things, but can't seem to nail it :)

fbrosson:
I tried to build on x86_gcc2 and it fails to build, but it looks like the problem is not related to any supposedly unbalanced \#ifdefs.
It's just that some declarations are mixed in the code. For example, in tidylib.c at lines 1439 and 1440 we have:
```
    doc->pConfigChangeCallback = NULL;
    int bomEnc;
```
(I wouldn't bother trying to patch this, since Haiku will drop gcc2.)

Begasus:
https://github.com/htacg/tidy-html5/issues/663#issuecomment-353747446

Build log included there, parse error can be fixed, that's not the big dealbreaker here I think ...

fbrosson:
I think that with gcc2 you can't use inline lists to initialize something, e.g. ` int foobar[] = { 1, 2, 3 }`.
You would need to use
```
int foobar[3];
foobar[0]=1;
foobar[1]=2;
foobar[2]=3;
```
I think I saw this kind of initialization in tidylib.c.

fbrosson:
Oops, you can. I should have checked before writing!
Will delete the message to remove unneeded noise.

fbrosson:
Does it build if you change
```diff
-languageDefinition *languages[];
+languageDefinition *languages[256];
```

pulkomandy:
Why bother with gcc2 here? Can't you just build with gcc5?

Begasus:
That worked @fbrosson 
> grabbing tidy_git-5.7.0-1-x86_gcc2.hpkg and moving it to /boot/home/haikuports/packages/tidy_git-5.7.0-1-x86_gcc2.hpkg

Begasus:
@pulkomandy as it looks like, the patch wouldn't be that hard, so why not try to include gcc2 version?

Begasus:
@fbrosson they already patched the language headers for the unbalanced #endif's, that's why I used a commit to check with their latest changes +1

fbrosson:
I did not know that upstream had kindly made some changes to make it easier for Haiku/gcc2 to build it,
So I think that not trying to make a small patch for gcc2 might dissapoint the upstream author(s).
@Begasus is willing to try that, so why not consider this? He just needs to add some tests to make sure the program will not crash if it is told to load more than a hardcoded number of languages, 256 in the example above.

pulkomandy:
gcc2 hacks are unlikely to be accepted upstream, and as such they just
add noise to our patches, unless there is really a need. When something
builds with gcc5, it's enough and adding gcc2 support is just a waste of
time (even if you don't waste much time :)). Unless there is a library
and we need it for something BeOS compatible, I would try to avoid
spending any effort on gcc2 support.

Begasus:
Looks like Haiku will be added as a supported OS there :)
https://github.com/htacg/tidy-html5/issues/663#issuecomment-353792623

Begasus:
Patches upstreamed, if they backport it to 5.6.0 it will hit it soon, otherwise they will be in for 5.7.0 :+1: 
Thanks all!

fbrosson:
I hope this change (on the upstream project) will not cause any trouble for them.
The way the memory is allocated for that structure has to be special:
I guess that when the program wants to load N language difinitions, it first computes something like:
```
bytes_to_allocate = sizeof(tidyLanguagesType) + N*sizeof(languageDefinition *)
```
and then uses malloc.
But sizeof(tidyLanguagesType) with the patch will be `256*sizeof(languageDefinition *)` bigger than the `sizeof(tidyLanguagesType)` without the patch.
While this cannot crash tidy itself, I'm not sure this would be an issue to other programs linked against libtidy.so.
Well, I guess upstream will take care of all this.

@Begasus, maybe you should tell upstream that it's perfectly OK if they don't merge that patch. It would be sad if they get into trouble trying to support gcc2 just for Haiku.

Begasus:
Forwarded the comment to the issue there @fbrosson ... waiting for their reply ...

Begasus:
Reply from the developer there :)

https://github.com/htacg/tidy-html5/issues/663#issuecomment-353885736

fbrosson:
Yep, I noticed it because GitHub sends notifications to any user who is mentioned in a comment.
But thanks anyway :)

Begasus:
Updated now to the latest commit on their side, it's a pre 5.7.0, builds without patching now

app-text/htmltidy/tidy_git-5.7.0.recipe
```diff
@@ -0,0 +1,91 @@
+SUMMARY="The granddaddy of HTML tools, with support for modern standards"
+DESCRIPTION="Tidy tidies HTML and XML. It can tidy your documents by itself, \
+and developers can easily integrate its features into even more powerful tools."
+HOMEPAGE="http://www.html-tidy.org/"
+COPYRIGHT="1998-2015 World Wide Web Consortium"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="28be79db50d2ecf6730068167d8ed23987fa8d59"
+SOURCE_URI="https://github.com/htacg/tidy-html5/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="5fddac299a385203e06a2158df3acd68aa83ffd10d5bdf73b9e7f2913dee1a88"
```
fbrosson: I think it would be better to rename the recipe to `tidy-5.7.0~git.recipe` (or `tidy57-5.7.0~git.recipe`).

Because with the current name (tidy_git-5.7.0.recipe) your PROVIDES are not 100% correct.
Besides, it is missing a CONFLICTS with the package for tidy-5.1.25.

If you wish to have both the old and the new, then you need to ajust the PROVIDES for the new package and add a CONFLICTS.

Anyway, it would be nice to insert a:
```
SOURCE_FILENAME="tidy-html5-$portVersion~$srcGitRev.tar.gz"
```
between CHECKSUM_SHA256 and SOURCE_DIR.

Begasus: I will revert this one to the one from 5.6.0 with the patches, this is actually a post 5.6.0/pre 5.7.0 :)

Begasus:
Back to 5.6.0 with the patches ...

app-text/htmltidy/tidy_git-5.7.0.recipe
fbrosson: Why don't you just drop the old recipe and add the new one as `tidy-5.7.0~git.recipe` ?
Then, when 5.7.0 is out, you will just need to rename the recipe to `tidy-5.7.0.recipe` and switch to the stable release tarball. Moreover, this method has the advantage that you don't need to worry about adding any CONFLICTS, and your PROVIDES are already OK.
Tidy pre-5.7.0 seems to be working fine, isn't it?
Nobody will complain if you remove the old recipe :)

Begasus: These changes will probably to into 5.6.1 as the developer mentioned, so maybe wait to update to the latest one after this one, so then we can leave out the patch, if nobody has a problem with it, I can still use the 5.7.0-pre version, but then when 5.6.1 commes out it wouldn't make any sence?

fbrosson: Are we allowed to cheat here? I mean, put a `5.7.0~git` in `5.6.1~git` ?
Yeah, I know this is really bad.
Context:
`media-libs/libextractor/libextractor-1.3.recipe` is the **_only_** recipe which depends on `lib:libtidy` and it's not marked tested on any arch:
```
ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
SECONDARY_ARCHITECTURES="!x86_gcc2 ?x86"
```
(I don't want to push towards any direction. I'm just trying to see if we can avoid a patch, especially knowing that it will be removed when 5.6.1 becomes available.)

Another "solution" would be to leave this PR open for a few days or maybe one week, if nobody needs the new version of tidy right now.

Begasus: A bit confused here too ;)
We 'could' use a 5.6.1~git recipe that pull the 5.7.0~git sources
it would also mean changing the $libVersion though, as the ones created by the last one will be 5.7.0
Or I could leave this one in atm and add the one for 5.7.0~git to see where it goes

Choices ... ;)

fbrosson: I think I know what's the best thing to do:

* we keep the old recipe and
* you add a `tidy57-5.7~git.recipe` with modifed PROVIDES, REQUIRES, and a CONFLICTS.

This will be better than adding a patch for 5.6.0, and  much better than pretending we have a `5.6.1~git` that would actually be a `5.7.0~git`. (Yes, my previous idea about cheating was really bad.)

Adding the CONFLICTS and changing the PROVIDES is not hard.
Just rename your recipe to `tidy57-5.7.0~git.recipe` and I'll show you what changes you need to make.

Begasus: tidy57-5.7~git.recipe ... can't seem to build this with haikuporter? 'not found in the repository'

app-text/htmltidy/tidy57-5.7~git.recipe
```diff
@@ -0,0 +1,92 @@
...
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix
+fi
+
+libVersion=5.7.0
+libVersionCompat="$libVersion compat >= 5"
+
+PROVIDES="
+	tidy$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
-	tidy$secondaryArchSuffix = $portVersion
+	tidy57$secondaryArchSuffix = $portVersion
```

app-text/htmltidy/tidy_git-5.7.0.recipe
Begasus: Changed to the new recipe, also changed the provide for the $libVersion, looking forward on the next step :)

app-text/htmltidy/tidy57-5.7~git.recipe
```diff
@@ -0,0 +1,92 @@
...
+	cmd:tidy$commandSuffix = $portVersion
+	lib:libtidy$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	devel:libtidy$secondaryArchSuffix = $libVersionCompat
```
fbrosson:
```diff
+CONFLICTS="
+	tidy$secondaryArchSuffix
+
 PROVIDES_devel="
-	devel:libtidy$secondaryArchSuffix = $libVersionCompat
+	tidy57${secondaryArchSuffix}_devel = $libVersionCompat
 	devel:libtidys$secondaryArchSuffix
 	"
```

app-text/htmltidy/tidy57-5.7~git.recipe
```diff
@@ -0,0 +1,92 @@
...
+	devel:libtidys$secondaryArchSuffix
+	"
+REQUIRES_devel="
+	tidy$secondaryArchSuffix == $portVersion base
```
fbrosson:
```diff
 REQUIRES_devel="
-	tidy$secondaryArchSuffix == $portVersion base
+	tidy57$secondaryArchSuffix == $portVersion base
 	"
+CONFLICTS_devel="
+	tidy${secondaryArchSuffix}_devel
+

File: app-text/htmltidy/tidy_git-5.7.0.recipe (hunk 8 above)
Comment (fbrosson) [reply app-text/htmltidy/tidy_git-5.7.0.recipe]: You forgot the ".0" in `tidy57-5.7.0~git.recipe` when you renamed your recipe.

Comment (fbrosson) [reply app-text/htmltidy/tidy_git-5.7.0.recipe]: Oops, sorry @Begasus, I wrote `tidy57-5.7~git.recipe` instead of `tidy57-5.7.0~git.recipe`.
My mistake!

**Comment by Begasus** [comment]
Still can install both tidy-5.1.25 and tidy-5.7.0 next to each other without conflicts?

File: app-text/htmltidy/tidy57-5.7.0~git.recipe [hunk 12]
```diff
@@ -0,0 +1,96 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:xsltproc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libtidy$secondaryArchSuffix \
+		$libDir/libtidy.so.$libVersion
```
fbrosson: One tab is enough here :)

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,96 @@
...
+		rm -rf $manDir
+		maybe_manDir=
+	else
+		maybe_manDir=$manDir
+	fi
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+		$maybe_manDir
```
fbrosson: I guess there is a `$manDir/man1/tidy.1` so you should use `$maybe_manDir_man3` here instead of just `$maybe_manDir`, otherwise the man for tidy(1) will be in the devel package.
Of course, if this fix is required, what you will need to set, a few lines above, is not `maybe_manDir=$manDir` but `maybe_manDir_man3=$manDir/man3`.

fbrosson:
> Still can install both tidy-5.1.25 and tidy-5.7.0 next to each other without conflicts?

Hmm, I just noticed that, although you did add a `CONFLICTS_devel`, you forgot to add the `CONFLICTS` for the base package (after `REQUIRES`):
```
CONFLICTS="
	tidy$secondaryArchSuffix
	"
```

Begasus:
Think this should be ok :)

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+	make install
+
+	prepareInstalledDevelLibs libtidy libtidys
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $manDir
+		maybe_manDir=
+	else
+		maybe_manDir=$manDir
```
fbrosson:
```diff
 	if [ -n "$secondaryArchSuffix" ]; then
 		rm -rf $manDir
-		maybe_manDir=
+		maybe_manDir_man3=
 	else
-		maybe_manDir=$manDir
+		maybe_manDir_man3=$manDir/man3
 	fi
```

Begasus: Question (not so familiar with maybe_mandir), but shouldn't this be 
> maybe_manDir_man1=$manDir/man1
I don't see a *.1 man in $manDir/man3 ?

Begasus: man for tidy1 still ends up in the develop package ...

fbrosson: There is no need for a maybe_manDir_man**1** because what we store in the base package is:

**{** what `make install` installs **}** minus **{** anything we send to any sub-packages **}**

fbrosson: > man for tidy1 still ends up in the develop package ...

This is because you are setting `maybe_manDir` to `$manDir` and then sending it to the devel package.
What you need to send to the devel package is `$manDir/man3`.
But we can't use `$manDir/man3` because it might be already deleted. We need to use `$maybe_manDir_man3`.

fbrosson: There are 2 cases:
* when you build for a primary arch, `$secondaryArchSuffix` is an empty string and `maybe_manDir_man3` is `$manDir/man3`.
* when you build for a secondary arch e.g. x86, `$secondaryArchSuffix` is `_x86` and `maybe_manDir_man3` is an empty string.

You just need to image what happens in INSTALL for each of these cases.

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
+SUMMARY="The granddaddy of HTML tools, with support for modern standards"
+DESCRIPTION="Tidy tidies HTML and XML. It can tidy your documents by itself, \
+and developers can easily integrate its features into even more powerful tools."
+HOMEPAGE="http://www.html-tidy.org/"
+COPYRIGHT="1998-2015 World Wide Web Consortium"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="28be79db50d2ecf6730068167d8ed23987fa8d59"
+SOURCE_URI="https://github.com/htacg/tidy-html5/archive/$srcGitRev.tar.gz"
+SOURCE_FILENAME="tidy-html5-$portVersion"
```
fbrosson:
```diff
-SOURCE_FILENAME="tidy-html5-$portVersion"
 CHECKSUM_SHA256="5fddac299a385203e06a2158df3acd68aa83ffd10d5bdf73b9e7f2913dee1a88"
+SOURCE_FILENAME="tidy-html5-$portVersion"
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+BUILD()
+{
+	cd build/cmake
+
+	cmake ../.. -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX:PATH=$prefix \
+		-DLIB_INSTALL_DIR=$libDir \
+		-DINCLUDE_INSTALL_DIR=$includeDir \
+		-DMAN_INSTALL_DIR=$manDir/man1 \
+		-DBIN_INSTALL_DIR=$commandBinDir
```
fbrosson:
```diff
-		-DBIN_INSTALL_DIR=$commandBinDir
+		-DBIN_INSTALL_DIR=$binDir
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix
+fi
```
fbrosson:
```diff
-commandSuffix=$secondaryArchSuffix
-commandBinDir=$binDir
-if [ "$targetArchitecture" = x86_gcc2 ]; then
-	commandSuffix=
-	commandBinDir=$prefix
-fi
-
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+INSTALL()
+{
+	cd build/cmake
+	make install
+
+	prepareInstalledDevelLibs libtidy libtidys
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $manDir
```
fbrosson:
```diff
+		rm -rf $binDir
 		rm -rf $manDir
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
Begasus: Not working with "-DMAN_INSTALL_DIR=$manDir/man1 \"
 -- Installing: /packages/tidy57-5.7.0~git-1/.self/documentation/man/man1/tidy.1
 -- Installing: /packages/tidy57-5.7.0~git-1/.self/lib/pkgconfig/tidy.pc
  /packages/tidy57_devel-5.7.0~git-1/.self
  mv: cannot stat '/packages/tidy57-5.7.0~git-1/.self/documentation/man/man3': No such file or directory

If I change that to "-DMAN_INSTALL_DIR=$manDir/man3 \"
It works, but tidy.1 will end up in $manDir/man3 (in the develop package for x86_gcc2) and isn't in the packages for x86 (that part I understand) :) 
So, still wondering if it's correct that tidy1 should go into man3?

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+	commandSuffix=
+	commandBinDir=$prefix
+fi
+
+libVersion=5.7.0
+libVersionCompat="$libVersion compat >= 5"
+
+PROVIDES="
+	tidy57$secondaryArchSuffix = $portVersion
+	cmd:tidy$commandSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES="
 	tidy57$secondaryArchSuffix = $portVersion
-	cmd:tidy$commandSuffix = $portVersion
 	lib:libtidy$secondaryArchSuffix = $libVersionCompat
 	"
+if [ -n "$secondaryArchSuffix" ]; then
+PROVIDES="$PROVIDES
+	cmd:tidy$commandSuffix = $portVersion
+	"
+fi
 REQUIRES="
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+BUILD()
+{
+	cd build/cmake
+
+	cmake ../.. -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+		-DCMAKE_INSTALL_PREFIX:PATH=$prefix \
+		-DLIB_INSTALL_DIR=$libDir \
+		-DINCLUDE_INSTALL_DIR=$includeDir \
+		-DMAN_INSTALL_DIR=$manDir/man3 \
+		-DBIN_INSTALL_DIR=$commandBinDir
```
fbrosson: Don't forget this one. (GitHub is hiding my previous comment on this change.)
```diff
-		-DBIN_INSTALL_DIR=$commandBinDir
+		-DBIN_INSTALL_DIR=$binDir
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:xsltproc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libtidy$secondaryArchSuffix \
```
fbrosson:
```diff
-defineDebugInfoPackage libtidy$secondaryArchSuffix \
+defineDebugInfoPackage tidy57$secondaryArchSuffix \
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+	prepareInstalledDevelLibs libtidy libtidys
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $prefix/bin
+		rm -rf $manDir
+		maybe_manDir_man1=
+	else
+		maybe_manDir_man1=$manDir/man1
```
fbrosson: Sorry, I thought there would be a man3 section, but there isn't. So we can shrink this a bit :)
```diff
	if [ -n "$secondaryArchSuffix" ]; then
 		rm -rf $prefix/bin
 		rm -rf $manDir
-		maybe_manDir_man1=
-	else
-		maybe_manDir_man1=$manDir/man1
 	fi
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+		-DMAN_INSTALL_DIR=$manDir/man1 \
+		-DBIN_INSTALL_DIR=$binDir
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build/cmake
+	make install
```
fbrosson: Suggested neutral cosmetic change: use `make -C`:
```diff
-	cd build/cmake
-	make install
+	make -C build/cmake install
```

app-text/htmltidy/tidy57-5.7.0~git.recipe
fbrosson: That's funny, I noticed that SOURCE_FILENAME was not at the right place but I did not notice it was missing the `.tar.gz` suffix! 😲

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,90 @@
...
+DESCRIPTION="Tidy tidies HTML and XML. It can tidy your documents by itself, \
+and developers can easily integrate its features into even more powerful tools."
+HOMEPAGE="http://www.html-tidy.org/"
+COPYRIGHT="1998-2015 World Wide Web Consortium"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="28be79db50d2ecf6730068167d8ed23987fa8d59"
+SOURCE_URI="https://github.com/htacg/tidy-html5/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="5fddac299a385203e06a2158df3acd68aa83ffd10d5bdf73b9e7f2913dee1a88"
+SOURCE_FILENAME="tidy-html5-$portVersion"
```
fbrosson:
```diff
-SOURCE_FILENAME="tidy-html5-$portVersion"
+SOURCE_FILENAME="tidy-html5-$portVersion.tar.gz"
```

app-text/htmltidy/tidy57-5.7~git.recipe
fbrosson: Sorry, I wrote `$portVersion` instead of `$libVersion` !!!

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,90 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+CONFLICTS="
+	tidy$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	tidy57${secondaryArchSuffix}_devel = $portVersion
+	devel:libtidy$secondaryArchSuffix = $portVersion
```
fbrosson: Oops, I almost missed this one :)
```diff
-	devel:libtidy$secondaryArchSuffix = $portVersion
+	devel:libtidy$secondaryArchSuffix = $libVersion
```
Hmm, it was my fault! I made that mistake in an [old comment](https://github.com/haikuports/haikuports/pull/1932#discussion_r158699653). Sorry!

app-text/htmltidy/tidy57-5.7.0~git.recipe
```diff
@@ -0,0 +1,90 @@
...
+DESCRIPTION="Tidy tidies HTML and XML. It can tidy your documents by itself, \
+and developers can easily integrate its features into even more powerful tools."
+HOMEPAGE="http://www.html-tidy.org/"
+COPYRIGHT="1998-2015 World Wide Web Consortium"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="28be79db50d2ecf6730068167d8ed23987fa8d59"
+SOURCE_URI="https://github.com/htacg/tidy-html5/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="5fddac299a385203e06a2158df3acd68aa83ffd10d5bdf73b9e7f2913dee1a88"
+SOURCE_FILENAME="tidy-html5-$portVersion.tar.gz"
```
fbrosson: Oops, we both forgot to insert a `-$srcGitRev`.
```diff
-SOURCE_FILENAME="tidy-html5-$portVersion.tar.gz"
+SOURCE_FILENAME="tidy-html5-$portVersion-$srcGitRev.tar.gz"
```

PR #1933, opened 2017-12-21
Album, fix build

Begasus:


haiku-apps/album/album-0.9.4.recipe
```diff
@@ -5,28 +5,32 @@ may be viewed and tagged. \
...
 LICENSE="MIT"
-REVISION="4"
-SOURCE_URI="https://github.com/HaikuArchives/Album/archive/550632886979080247db552f38990690f77ee054.tar.gz"
-CHECKSUM_SHA256="ee1811b47bbb1178c88b66ef103bf8aba0ba551cfed70ce8a318842c4729c89e"
-SOURCE_DIR="Album-550632886979080247db552f38990690f77ee054"
+REVISION="5"
+COMMIT=3d9e8e25a1d6bb2684e6fe0342336d5a57ad7906
+SOURCE_URI="https://github.com/HaikuArchives/Album/archive/$COMMIT.tar.gz"
+CHECKSUM_SHA256="ad96483a2be75c203cec1ca4207a4d5568c88fa58ccb4b0a7583ddc625b50d48"
+SOURCE_DIR="Album-$COMMIT"
```
fbrosson: Some time ago I counted how many `srcGitRev`s and `COMMIT`s variables we had and it tourned out that we have many more `srcGitRev`s than `COMMIT`s.
And since all official variables are upper-case, it might be worth reserving upper-case variables for official variables only, and mixed case variables for everything else.
That said, we already have many `COMMIT`s, so do not hesitate to stick to it here too ;-)

Begasus: If needed I can still change it to srcGitRev :)

fbrosson: Well, it's not needed. It's just to try and use the most frequently used name.

BTW, my comment about mixed-case variables was uncomplete. I should have written that:
* the upper-case variables defined in a recipe serve as input to haikuporter.
* all haikuporter-supplied variables are lower-case-only or mixed-case (e.g. prefix, jobArgs, secondaryArchSuffix, binDir, etc...)
* haikuporter does not use any input from mixed-case (or lower-case-only) variables defined in a recipe.

Begasus:
Fix from @waddlesplash  (at haikuarchives) confirmed working on x86_64

PR #1951, opened 2017-12-24
hdf5, fix TEST

Begasus:


sci-libs/hdf5/hdf5-1.10.1.recipe
```diff
@@ -6,7 +6,7 @@ HOMEPAGE="http://www.hdfgroup.org/HDF5/"
 COPYRIGHT="2011-2016 The HDF Group,
 	The Board of Trustees of the University of Illinois."
 LICENSE="BSD (3-clause)"
-REVISION="1"
+REVISION="2"
```
fbrosson: We don't need to bump `REVISION`, unless the builders are doing things based on the result of `TEST()`.

Begasus: Ah, will revert :)

sci-libs/hdf5/hdf5-1.10.1.recipe
```diff
@@ -151,5 +151,5 @@ INSTALL()
 
 TEST()
 {
-	make tests
+	make check
```
fbrosson: At lines 15 and 16 of the Makefile we have:
```
all lib progs check test _test install uninstall dep depend: _config
	$(MAKE) $@
```
And "check" does not appear anywhere else.

However, at lines 34 and 35 of the Makefile we have:
```
.PHONY: all lib progs test install uninstall dep depend clean mostlyclean     \
	distclean maintainer-clean _config
```

Hence this question: Are you sure it's "make check" that you want, and not "make test" ?

Begasus: "make test" only builds the test cases (and they are already build in the process), it doesn't "actually" do something, "make check" will "actually" run a test :) (checked before)

fbrosson: OK. Let's merge, then :)

PR #1957, opened 2017-12-24
Enable Find Dictionaries in Wordgrinder

localanu:
- Enabled find dictionaries
- Change dir in spilchocker.lua
- Use in app dictionaries
To Do
- Find about aspell

### DIFF
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -94,7 +94,7 @@ INSTALL()
 	rmdir $docDir/wordgrinder
 if [ -n "$secondaryArchSuffix" ]; then
 	install -d $dataDir/wordgrinder/dictionaries
-	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
 	packageEntries dictionaries \
 		$dataDir
 fi
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -70,5 +70,8 @@ INSTALL()
 		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
+	mkdir $datadir/words
+	cp $sourceDir/extras/american-canadian.dictionary $dataDir/words
+	cp $sourceDir/extras/british.dictionary $dataDir/words
 	rmdir $docDir/wordgrinder
 }
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -9,7 +9,7 @@ COPYRIGHT="2007-2017 David Given"
 LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
-PATCHSET=wordgrinder-$portVerison.patchset
+PATCHES=wordgrinder-$portVersion.patchset
 CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -60,10 +60,6 @@ PROVIDES_dictionaries="
 	wordgrinder_dictionaries = $portVersion
 	"
 REQUIRES_dictionaries="
-	wordgrinder == $portVersion base
-	"
-SUPPLEMENTS_dictionaries="
-	wordgrinder
 	"
 fi
 
@@ -95,11 +91,10 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
-	install -d $dataDir/wordgrinder/dictionaries
-	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
 if [ -z "$secondaryArchSuffix" ]; then
 	packageEntries dictionaries \
 		$dataDir
 fi
-
+	install -d $dataDir/wordgrinder/dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
 }
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -39,6 +39,7 @@ CONFLICTS="
 	wordgrinder_dictionaries
 	"
 fi
+
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 #	devel:libfreetype$secondaryArchSuffix
@@ -62,7 +63,6 @@ PROVIDES_dictionaries="
 REQUIRES_dictionaries="
 	"
 fi
-
 
 defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -91,7 +91,7 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
-	
+
 	install -d $dataDir/wordgrinder/dictionaries
 	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
 if [ -z "$secondaryArchSuffix" ]; then
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -53,7 +53,7 @@ defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
 PATCH()
 {
-	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/$portName-$portVersion/words|;" \
+	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder-$portVersion/words/|*.dictionaries;" \
 		src/lua/addons/spillchocker.lua
 }
 
@@ -70,7 +70,7 @@ INSTALL()
 		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
-	mkdir $datadir/words
-	cp extras/*.dictionary $dataDir/words
+#	mkdir $datadir/words
+#	cp extras/*.dictionary $dataDir/words
 	rmdir $docDir/wordgrinder
 }
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -63,6 +63,7 @@ SUPPLEMENTS_dictionaries="
 	"
 fi
 
+
 defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
 
@@ -79,7 +80,7 @@ PATCH()
 BUILD()
 {
 	make PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
-		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
+		MANDIR=$manDir HOME="`finddir B_USER_SETTINGS_DIRECTORY`" \
 		LUA_PACKAGE=lua
 }
 
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -9,8 +9,8 @@ COPYRIGHT="2007-2017 David Given"
 LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
-PATCHES=wordgrinder-$portVersion.patchset
 CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
+PATCHES=wordgrinder-$portVersion.patchset
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
@@ -49,11 +49,26 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+if [ -n "$secondaryArchSuffix" ]; then
+SUMMARY_dictionaries="Built-in dictionaries for WordGrinder"
+ARCHITECTURES_dictionaries="any"
+PROVIDES_dictionaries="
+	wordgrinder_dictionaries = $portVersion
+	"
+REQUIRES_dictionaries="
+	wordgrinder == $portVersion base
+	"
+SUPPLEMENTS_dictionaries="
+	wordgrinder
+	"
+fi
+
 defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
+
 PATCH()
 {
-	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder-$portVersion/words/|*.dictionaries;" \
+	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder/dictionaries/|;" \
 		src/lua/addons/spillchocker.lua
 }
 
@@ -70,7 +85,12 @@ INSTALL()
 		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
-#	mkdir $datadir/words
-#	cp extras/*.dictionary $dataDir/words
 	rmdir $docDir/wordgrinder
+if [ -n "$secondaryArchSuffix" ]; then
+	install -d $dataDir/wordgrinder/dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionaries
+	packageEntries dictionaries \
+		$dataDir
+fi
+
 }
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -1,4 +1,3 @@
-export USER_SETTINGS=`/bin/finddir B_USER_SETTINGS_DIRECTORY`
 SUMMARY="A simple, Unicode-aware word processor that runs on the console"
 DESCRIPTION="WordGrinder is designed to get out of your way and let you write; \
 it does very little, but what it does it does it well. It supports basic \
@@ -71,24 +70,23 @@ PATCH()
 {
 	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder/dictionaries/|;" \
 		src/lua/addons/spillchocker.lua
-	sed -i -e "s|HOME ..||;" \
+	sed -i -e "s|/.wordgrinder.settings|/wordgrinder/settings|;" \
 		src/lua/settings.lua
-	sed -i -e "s|/.wordgrinder.settings|$USER_SETTINGS/wordgrinder/settings|;" \
-		src/lua/settings.lua
+	sed -i -e "s|/.wordgrinder.lua|/wordgrinder/wordgrinder.lua|;" \
+		src/lua/main.lua
 }
 
 BUILD()
 {
 	make PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
-		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
-		USER_SETTINGS_DIRECTORY=$USER_SETTINGS_DIRECTORY
+		MANDIR=$manDir HOME="`finddir B_USER_SETTINGS_DIRECTORY`" \
 		LUA_PACKAGE=lua
 }
 
 INSTALL()
 {
 	make install PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
-		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
+		MANDIR=$manDir HOME="`finddir B_USER_SETTINGS_DIRECTORY`" \
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
--- a/app-text/wordgrinder/wordgrinder-0.7.1.recipe
+++ b/app-text/wordgrinder/wordgrinder-0.7.1.recipe
@@ -79,14 +79,14 @@ PATCH()
 BUILD()
 {
 	make PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
-		MANDIR=$manDir HOME="`finddir B_USER_SETTINGS_DIRECTORY`" \
+		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
 		LUA_PACKAGE=lua
 }
 
 INSTALL()
 {
 	make install PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
-		MANDIR=$manDir HOME="`finddir B_USER_SETTINGS_DIRECTORY`" \
+		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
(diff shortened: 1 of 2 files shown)


Begasus:
wordgrinder_dictionary

Begasus:
Added the dictionaries to the recipe ...
wordgrinder_dictionary2

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -50,6 +51,11 @@ BUILD_PREREQUIRES="
 
 defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
+PATCH()
+{
+	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/words|;" \
+		src/lua/addons/spillchocker.lua
```
Begasus: wordgrinder will actually search for a dictionary called "words" here (check the provided screenshots in the PR)?

Begasus:
Once you change the settings to use one of the dictionaries, wordgrinder creates a file ~/.wordgrinder.settings, maybe this could be added to the task also? :)

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -70,5 +70,8 @@ INSTALL()
 		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
+	mkdir $datadir/words
+	cp $sourceDir/extras/american-canadian.dictionary $dataDir/words
+	cp $sourceDir/extras/british.dictionary $dataDir/words
```
Begasus: you can combine these 2 with:

> cp extras/*.dictionary $dataDir/words

app-text/wordgrinder/wordgrinder-0.7.1.recipe
Begasus: I used the next line, @fbrosson is the sed expert here though ...

> sed -i -e "s|/etc/dictionaries-common/words|$dataDir/words|*.dictionaries;"

Begasus: jikes, my bad:

>> sed -i -e "s|/etc/dictionaries-common/words|$dataDir/words/*.dictionaries|;"

fbrosson: I might be wrong (as I did not test anything yet), but I WordGrinder might accept a directory path instead of the full path of a single directory. If this is true, the I think this could work:
```
PATCH()
{
	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder/dictionaries/|;" \
		src/lua/addons/spillchocker.lua
}
```
It will tell WordGrinder to look for one (or all) dictionary in `/system/data/wordgrinder/dictionaries/`. So adding this in INSTALL;
```
	install -d $dataDir/wordgrinder/dictionaries
	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionaries
```
should populate that directory.

It would be nice, however, to split these files into a separate sub-package (created in the same recipe) that we could call wordgrinder_dictionaries or wordgrinder_words (or anything else).

Begasus: @fbrosson my guess here is that wordgrinder searches for the "dictionaries" instead of a "directory" in the sed commit?

also:

> install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary ;)

wordgrinder3

Begasus: forget the last thing I posted, still had the wordgrinder settings file installed :)

fbrosson: OK, I see. So it seems it's only on Windows that it accepts a directory.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -50,6 +51,11 @@ BUILD_PREREQUIRES="
 
 defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
+PATCH()
+{
+	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder-$portVersion/words/|*.dictionaries;" \
```
fbrosson:
```
	sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder-$portVersion/words/|*.dictionaries;" \
```
is incorrect for 2 reasons:
* using $portVersion here is not a good idea because $dataDir already has $portVersionedName in it.
* The `*.dictionaries` is not a sed command.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
localanu: I'll try it?

localanu: I'll try it?

Begasus: This should do it @fbrosson @localanu 

  sed -i -e "s|/etc/dictionaries-common/words|$dataDir/wordgrinder/dictionaries/|;" \
  install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -50,6 +51,11 @@ BUILD_PREREQUIRES="
 
 defineDebugInfoPackage wordgrinder$secondaryArchSuffix \
 	$commandBinDir/wordgrinder
```
Begasus: Add empty line here

app-text/wordgrinder/wordgrinder-0.7.1.recipe
fbrosson: Oops, I had not noticed I had written \*.dictionar**ies** instead of \*.dictionar**y**.
So thanks @Begasus!

Begasus: @fbrosson any example that we could use for a seperate packages for the directories?

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -9,6 +9,7 @@ COPYRIGHT="2007-2017 David Given"
 LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/davidgiven/wordgrinder/archive/$portVersion.tar.gz"
+PATCHES=wordgrinder-$portVersion.patchset
```
fbrosson: cosmetic change, to comply with the [HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering):
```diff
-PATCHES=wordgrinder-$portVersion.patchset
 CHECKSUM_SHA256="54085af7963e1f67342bc0b1b20d1ccc75494f2e23d401b601cf3089acea747c"
+PATCHES="wordgrinder-$portVersion.patchset"
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
Begasus: Got it :+1:

fbrosson: file is a good example.
Adding this after REQUIRES and before PATCH:
```
SUMMARY_dictionaries="Built-in dictionaries for WordGrinder"
ARCHITECTURES_dictionaries="any"
PROVIDES_dictionaries="
	wordgrinder_dictionaries = $portVersion
	"
REQUIRES_dictionaries="
	wordgrinder == $portVersion base
	"
SUPPLEMENTS_dictionaries="
	wordgrinder
	"
fi
```
And then inserting this in INSTALL:
```
	install -d $dataDir/wordgrinder/dictionaries
	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionaries
	packageEntries dictionaries \
		$dataDir
fi
```
shoud make the dictionaries available as a sub-package.
Remark: arch-indepedent packages cannot be built on secondary arch builds.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -65,4 +86,11 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
+if [ -n "$secondaryArchSuffix" ]; then
+	install -d $dataDir/wordgrinder/dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionaries
```
fbrosson:
```diff
-	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -48,13 +50,38 @@ BUILD_PREREQUIRES="
...
+		src/lua/settings.lua
+	sed -i -e "s|/.wordgrinder.settings|$USER_SETTINGS/wordgrinder/settings|;" \
+		src/lua/settings.lua
+}
+
 BUILD()
 {
 	make PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
 		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
+		USER_SETTINGS_DIRECTORY=$USER_SETTINGS_DIRECTORY
```
fbrosson: You don't need this line:
```
export USER_SETTINGS=`/bin/finddir B_USER_SETTINGS_DIRECTORY`
```
You just need to pass `$settingsDir/wordgrinder`
```diff
-		USER_SETTINGS_DIRECTORY=$USER_SETTINGS_DIRECTORY
+		USER_SETTINGS_DIRECTORY=$settingsDir/wordgrinder \
```

You will also need to insert this bloc after `SUPPLEMENTS_dictionaries` and before `PATCH()`:
```
USER_SETTINGS_FILES="
	settings/wordgrinder directory
```

localanu: hm,I checked and I not found the USER_SETTINGS_DIRECTORY ,I check user settings stored in HOME,can I change the HOME to B_USER_SETTINGS_DIRECTIORY?

localanu: I'm just try,for checks makefile have user settings var

localanu:
Why I change to B_USER_SETTINGS_DIRECTORY?,I see var HOME is using for store configurations

fbrosson:
On Linux many programs store their preferences in `$HOME`, but on Haiku this is not recommended. The Haiku philosophy recommends that all the settings be saved in `~/config/settings/` or in a subdirectory of that directory.
So here, for WordGrinder, if you find how to have the prefs file saved to /boot/home/config/settings/ or /boot/system/settings/ then it would be nice.
Otherwise, I guess it would not be that bad to save them in $HOME.

localanu:
@fbrosson so,I can't change home to B_USER_SETTINGS_DIRECONARY?

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -10,6 +10,7 @@ LICENSE="MIT"
 REVISION="1"
```
fbrosson: We need a revision bump here:
```diff
-REVISION="1"
+REVISION="2"
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -48,9 +49,33 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+if [ -n "$secondaryArchSuffix" ]; then
```
fbrosson: Oops, my mistake.
```diff
-if [ -n "$secondaryArchSuffix" ]; then
+if [ -z "$secondaryArchSuffix" ]; then
```
If we don't fix this error then we get no dictionaries sub-package on x86_64, and, on x86_gcc2, `hp wordgrinder_x86` produces an incorrect `wordgrinder_x86_dictionaries-0.7.1-2-x86_gcc2.hpkg` package instead of `wordgrinder_x86_dictionaries-0.7.1-2-any.hpkg` (the difference is mainly on the name: `-any.hpkg` vs `-x86_gcc2.hpkg`)

Yes, haikuporter does not support building arch-independent packages when it is building for a secondary arch.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -65,4 +90,11 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
+if [ -n "$secondaryArchSuffix" ]; then
```
fbrosson: Oops, same mistake, mine too :)
```diff
-if [ -n "$secondaryArchSuffix" ]; then
+if [ -z "$secondaryArchSuffix" ]; then
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
Begasus: This still doesn't create a dictionaries packages for me?

fbrosson: If you are running x86_64 then, with the fixes above, you should obtain a `wordgrinder_x86_dictionaries-0.7.1-2-any.hpkg`.
On x86_gcc2 you won't get that sub-package because wordgrinder does not build on x86_gcc2 primary arch. So in that arch you will get a `wordgrinder_x86-0.7.1-2-x86_gcc2.hpkg`.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
fbrosson: > Add empty line here

Oops, this was an old message which @Begasus had added when an empty line was missing. But the message was not attached to the correct line, so GitHub continued showing it.
But it was OK before this last change. One empty line is enough, after the "fi" and before the "defineDebugInfoPackage". (No need for 2 blank lines.)

app-text/wordgrinder/wordgrinder-0.7.1.recipe
Begasus: Still no dictionaries package ...

fbrosson: @Begasus, I guess you are running x86_gcc2. It is perfectly normal to not have the dictionaries sub-package when you run "hp wordginder_x86".
That said, you can build it on x86_64 and copy the wordginder_dictionaries hpkg from it to your x86_gcc2 and then install it there (after installing the wordginder_x86 hpkg)

BTW,it could be nice to have a repo for the arch-dependent packages ("-x86_gcc2" and "-x86_64"), and a repo for the arch-independent sub-packages (and packages), ("-any").

Reminder: Arch-independent packages, built on any primary arch, are supposed to be identical.

This way, for packages that do not build on x86_gcc2 primary arch we would be able to install their arch-independent packages from the "-any" repo.

When I say "It would be **nice**", I know this it would **not** be simple.

Begasus: @fbrosson wasn't the plan to make the dictionaries available? Not every one can switch to x86_64 to build the "any" package and then copy them over, IMHO it would be nice if they were able to build on x86_gcc2 also so one wouldn't have to switch?

Begasus: That said, if this is how it should be I'm ok with it

fbrosson: Hmm, what we can do is keep the dictionaries in the main package only when we are building for a secondary arch, and, for x86_64, put them in the external package.
This should not be too hard to do. I'll add a review with the changes, then we can decide if that's a good idea or not.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -65,4 +91,11 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
+if [ -z "$secondaryArchSuffix" ]; then
```
fbrosson: This will allow to have the dictionaries in the main package when we're building for a secondary arch.
```diff
-if [ -z "$secondaryArchSuffix" ]; then
 	install -d $dataDir/wordgrinder/dictionaries
 	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
+if [ -z "$secondaryArchSuffix" ]; then
 	packageEntries dictionaries \
 		$dataDir
 fi
```
Cool, isn't it?

fbrosson: So, what do we do?
Do we keep the package without the dictionaries, as I had though initially, or do we include it for wordgrinder_x86?

BTW, I realized that having  wordgrinder_dictionaries depend on wordgrinder (instead of the opposite) makes the wordgrinder_dictionaries hpkg installable only on the plafform on which it was built.

So if we wanted to be able to use wordgrinder_dictionaries on any arch, we would have to say that it does not depend on wordgrinder. Well, this can be easily done by removing the definition of REQUIRES_dictionaries.

@localanu, in case there is a consensus for including the dictionaries in wordgrinder_x86, there will be a small fix I had not forseen:
Inserting the following bloc after the definition of REQUIRES:
```diff
 	lib:libz$secondaryArchSuffix
 	"
+if [ -n "$secondaryArchSuffix" ]; then
+CONFLICTS="
+	wordgrinder_dictionaries
+	"
+fi
```
This will prevent the installation of wordgrinder_dictionaries on x86_gcc2 since on that arch the wordgrinder_x86 package will already include the dictionaries.

But please do not apply these changes before we reach a consensus on this question, as we might need to revert it.
Thanks!

Begasus: Just a thought here, then I'll silence ;)
Couldn't targetArchitecture come in handy in this case (to still be able to have a seperate package "any" on x86_gcc2), didn't check it (yet) ...

Begasus: Ok, just checked, still produces a x86 package instead of any
So I think we can be ok with this :+1: 
Good work!

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -65,4 +95,11 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
+	install -d $dataDir/wordgrinder/dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
+if [ -z "$secondaryArchSuffix" ]; then
+	packageEntries dictionaries \
+		$dataDir
+fi
+
```
fbrosson: Just a cosmetic "change" :)
There's a blank line near the end of INSTALL (see the small red "-"). Could you move it 6 line up? (at the small green "+")
```diff
 	rmdir $docDir/wordgrinder
+
 	install -d $dataDir/wordgrinder/dictionaries
 	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
 if [ -z "$secondaryArchSuffix" ]; then
 	packageEntries dictionaries \
 		$dataDir
 fi
-
 }
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -48,13 +53,38 @@ BUILD_PREREQUIRES="
...
 	"
 
+if [ -z "$secondaryArchSuffix" ]; then
+SUMMARY_dictionaries="Built-in dictionaries for WordGrinder"
+ARCHITECTURES_dictionaries="any"
+PROVIDES_dictionaries="
+	wordgrinder_dictionaries = $portVersion
+	"
+REQUIRES_dictionaries="
+	wordgrinder == $portVersion base
```
fbrosson: After all, there's no need to make wordgrinder_dictionaries depend on wordgrinder. We can also completely remove the `SUPPLEMENTS_dictionaries`.
```diff
 REQUIRES_dictionaries="
-	wordgrinder == $portVersion base
 	"
-SUPPLEMENTS_dictionaries="
-	wordgrinder
-	"
```

Also, could you please add a blank line before `BUILD_REQUIRES` ?
Thanls!

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -65,4 +91,10 @@ INSTALL()
 		LUA_PACKAGE=lua
 	mv $docDir/wordgrinder/README.wg $docDir
 	rmdir $docDir/wordgrinder
+if [ -z "$secondaryArchSuffix" ]; then
+	packageEntries dictionaries \
+		$dataDir
+fi
+	install -d $dataDir/wordgrinder/dictionaries
+	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
```
fbrosson: Oops, the last two lines (which install the dictionaries) are not in the right place.
This is what we should have:
```
INSTALL()
{
	make install PREFIX=$prefix BINDIR=$commandBinDir DOCDIR=$docDir \
		MANDIR=$manDir HOME="`finddir B_USER_DIRECTORY`" \
		LUA_PACKAGE=lua
	mv $docDir/wordgrinder/README.wg $docDir
	rmdir $docDir/wordgrinder

	install -d $dataDir/wordgrinder/dictionaries
	install -t $dataDir/wordgrinder/dictionaries extras/*.dictionary
	packageEntries dictionaries \
		$dataDir
fi
}
```

app-text/wordgrinder/wordgrinder-0.7.1.recipe
fbrosson: OK, I understand why there was a misunderstanding. The comment above was not very clear, sorry.
I should have written:
* green plus sign = insert a blank line here.
* red minus sign = delete this blank line.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -33,7 +34,11 @@ REQUIRES="
 	lib:libncurses$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
-
+if [ -n "$secondaryArchSuffix" ]; then
+CONFLICTS="
+	wordgrinder_dictionaries
+	"
+fi
 BUILD_REQUIRES="
```
fbrosson: We like to have a blank line before  BUILD_REQUIRES.

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -48,13 +53,34 @@ BUILD_PREREQUIRES="
...
+SUMMARY_dictionaries="Built-in dictionaries for WordGrinder"
+ARCHITECTURES_dictionaries="any"
+PROVIDES_dictionaries="
+	wordgrinder_dictionaries = $portVersion
+	"
+REQUIRES_dictionaries="
+	"
+fi
+
+
```
fbrosson: We like to have only one line before defineDebugInfoPackage.
No need for 2 lines.

localanu:
@fbrosson maybe,lua error cause by patch(),I try change dir to config/settings/wordgrinder

localanu:
@fbrosson can merged?,or had another problem?

app-text/wordgrinder/wordgrinder-0.7.1.recipe
localanu: I had do is

fbrosson:
Well, before we merge we need to make sure we don't get any LUA-related errors.
Hopefully there's no urgency at all.

BTW, we did not investigate on using the system dictionaries. This would be better IMO.
That said, if anyone else wants to merge this, I don't object. But the contributor who merges it will need to make sure there are no errors.

localanu:
@fbrosson this error?
`<command-line>:0:0: warning: "_XOPEN_SOURCE" redefined
`
package had error like this before I patch this package

fbrosson:
I might be wrong but I don't remember having seen these errors with the current package (i.e. without the changes in this PR.)
I remember having had Lua errors when trying the changes in this PR.
I have approved the task for the GCI so you don't need to worry about that, even if the PR has not been merged yet. BTW, we will probably merge it once we are sure everything is OK (and if we don't decide to use external dictionaries.)

Regarding the error you have mentioned, `<command-line>:0:0: warning: "_XOPEN_SOURCE" redefined`, are you sure you got it with the current package?

I'm asking this because when you try to install a package which is already installed with the exact same version and revision, the install fails but this might get unnoticed if you don't pay attention to what pkgman outputs.
So you might have though that the package which produced the error was the one which is currently in the repository, while it could be a newer version with some changes.

Anyway, we might also give up adding the embedded dictionaries, and, instead, add a dependency on hunspell (or aspell) with their external dictionaries. This would be a more flexible solution. But it needs much more work...

localanu:
@fbrosson  I'm sure, `<command-line>:0:0: warning: "_XOPEN_SOURCE" redefined`warrings have happened berfore I applying patch. I'm include haikuports log from my gist
[before](https://gist.github.com/localanu/181a8a911fc7ddc7734d37dacf2e5a4b)
[after](https://gist.github.com/localanu/b5798a79d8a2458c33ede5a1c8c6a416)

fbrosson:
Sorry @localanu, I had completely forgotten about this PR.
(I'll try to have a look at it tomorrow.)

localanu:
@fbrosson not problem

app-text/wordgrinder/wordgrinder-0.7.1.recipe
```diff
@@ -33,6 +34,11 @@ REQUIRES="
 	lib:libncurses$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
+if [ -n "$secondaryArchSuffix" ]; then
+CONFLICTS="
+	wordgrinder_dictionaries
+	"
+fi
```
alaviss: Isn't it just better to drop the dictionaries in this case?

Begasus:
The recipe looks ok to me, and even works with the new version 0.7.2 :)

Begasus:
Tried changing the path to use the new myspell dictionaries, seems like they are not recognized? (even when changing the path in Terminal to $dataDir/hunspell)

Begasus:
Maybe @davidgiven could shed a light on this if it's possible to use the dic/aff files from myspell?

davidgiven:
WordGrinder expects the dictionaries to be simple flat word lists, in the classic /usr/share/dict/words format. I don't know what format haiku's dictionaries are in but if they're anything more exotic like a compressed trie than that simply won't work.

Begasus:
Thanks for the fast answer @davidgiven ! The 2 dictionaries included in the source seem to work, if someone finishes this of one day we can close this :)

Begasus:
Merging as no activity for a long time, let's see what happens on next update to this

Begasus:
@localanu not sure if you are still have an interest here, just give a yell if you have, otherwise we'll close the PR :)

PR #1958, opened 2017-12-25
Toner: create recipe

owenca:
I will update COMMIT and CHECKSUM_SHA256 once [the other PR](https://github.com/HaikuArchives/Toner/pull/1) is merged.

### DIFF
--- a/haiku-apps/toner/toner-1.0.0.recipe
+++ b/haiku-apps/toner/toner-1.0.0.recipe
@@ -1,0 +1,42 @@
+SUMMARY="A tone generator"
+DESCRIPTION="Toner, a tone generator for aligning levels in your playback \
+system."
+HOMEPAGE="https://github.com/HaikuArchives/Toner"
+COPYRIGHT="2001 Ben Loftis"
+LICENSE="MIT"
+REVISION="1"
+COMMIT="eb78098c88b0d1eacb97a4e0085eff6bfd0a7f0b"
+SOURCE_URI="https://github.com/HaikuArchives/Toner/archive/$COMMIT.tar.gz"
+CHECKSUM_SHA256="9a3edf2de7436a33ce7706ad1347cbd3393e906c735ca097b42ded584e3e8a5f"
+SOURCE_DIR="Toner-$COMMIT"
+
+ARCHITECTURES="x86_gcc2 !x86 !x86_64"
+
+PROVIDES="
+	toner = $portVersion
+	app:Toner = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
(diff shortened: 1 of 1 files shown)


Begasus:
Looks ok at first sight :+1:

haiku-apps/toner/toner-1.0.0.recipe
```diff
@@ -0,0 +1,42 @@
+SUMMARY="A tone generator"
+DESCRIPTION="Toner, a tone generator for aligning levels in your playback \
+system."
+HOMEPAGE="https://github.com/HaikuArchives/Toner/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/HaikuArchives/Toner/"
+HOMEPAGE="https://github.com/HaikuArchives/Toner"
```
(trailing slash no needed here)

Everything else looks good to me too. (But I did not test anything yet.)
Oops:
```diff
-COPYRIGHT="2001 by Ben Loftis"
+COPYRIGHT="2001 Ben Loftis"
```

Begasus: @fbrosson for refference, when do they need to use the traling slash, and when not?

Begasus:
Confirmed running on x86_gcc2 (just can't test sound as I don't have any sound in my vmware install) :)

haiku-apps/toner/toner-1.0.0.recipe
fbrosson: For URIs with no path, ie. http://example.org/ the trailing slash _is_ required.
Some (most?) browsers hide the trailing slash, but this does not mean a URI without it is valid. As a matter of fact it is _not_ valid,

Now, for all other URIs, the rule is simple. When you type e.g.https://github.com/HaikuArchives/Toner in your browser's address bar, it will add a / if it is needed, or it will leave that unchanged if it is not needed. (That's the case for all project repos in GitHub.)
You can see this with curl by displaying the header of the reply to a request:
```
curl -I https://github.com/HaikuArchives/Toner
```
It does not send you a redirect.

On sites for which you should include the `/`, if you don't then you get a redirect:
```
curl -I https://www.w3.org/support
```
HTTP/1.1 301 Moved Permanently
Location: https://www.w3.org/support/
```

owenca: I will make the change when updating COMMIT to the merged commit.

fbrosson:
I can also confirm that Toner also works for me (on x86_gcc2).
But the sound is not working with Toner.
I'm using QEMU+KVM with some carefully chosen environment variables and qemu command line switches. With these settings I do have sound with VLC-0.8.6 and mpg321, but not with mpg123.
So either my conf is not good enough for all possible apps, or Toner needs fixes.
Maybe we should mark it as not tested.

owenca:
I have pushed a squashed commit and tested it.

PR #1959, opened 2017-12-25
OpenCV: bump version

extrowerk:


media-libs/opencv/opencv-3.4.0.recipe
```diff
@@ -12,8 +12,9 @@ LICENSE="BSD (3-clause)"
 REVISION="1"
 SOURCE_URI="https://github.com/opencv/opencv/archive/$portVersion.tar.gz"
 SOURCE_FILENAME="opencv-$portVersion.tar.gz"
-CHECKSUM_SHA256="5dca3bb0d661af311e25a72b04a7e4c22c47c1aa86eb73e70063cd378a2aa6ee"
+CHECKSUM_SHA256="678cc3d2d1b3464b512b084a8cca1fad7de207c7abdf2caa1fed636c13e916da"
 SOURCE_DIR="opencv-$portVersion"
```
fbrosson: Not needed, as it's the default value.
```diff
-SOURCE_DIR="opencv-$portVersion"
```

BTW, given that there are lots of "3.4.0", you might wish to make these neutral changes:
```diff
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
-	opencv$secondaryArchSuffix = $portVersion compat >= 3
+	opencv$secondaryArchSuffix = $portVersionCompat
```
and replace each `3.4.0 compat >= 3` by `$libVersionCompat`.

media-libs/opencv/opencv-3.4.0.recipe
```diff
@@ -84,23 +85,23 @@ REQUIRES="
 
 PROVIDES_devel="
 	opencv${secondaryArchSuffix}_devel = $portVersion compat >= 3
```
fbrosson:
```diff
 PROVIDES_devel="
-	opencv${secondaryArchSuffix}_devel = $portVersion compat >= 3
+	opencv${secondaryArchSuffix}_devel = $portVersionCompat
```

media-libs/opencv/opencv-3.4.0.recipe
```diff
@@ -249,6 +246,8 @@ INSTALL()
 
 TEST()
 {
+	# Fails as it needs the main packages, but they won't gets attached to the chroot
```
fbrosson: I guess the tests need the shared libs.
Well, you can workaround this by setting LIBRARY_PATH to an appropriate value.
Assuming the libs are available in src/lib/ after the build, you can do this:
```
TEST()
{
	LIBRARY_PATH="$sourceDir/src/lib${LIBRARY_PATH:+:$LIBRARY_PATH}" \
	make -C build_haiku test
}
```
(Don't forget to replace src/lib by the relative path to the real directory containing the libs.)

media-libs/opencv/opencv-3.4.0.recipe
```diff
@@ -249,6 +252,6 @@ INSTALL()
 
 TEST()
 {
-	cd build_haiku
-	make test
+	LIBRARY_PATH="$sourceDir/build_haiku/lib${LIBRARY_PATH:+:$LIBRARY_PATH}" \
+	make -C build_haiku test $jobArgs
```
fbrosson: Any opinions on using `$jobArgs` in `TEST()` ?

(Wouldn't it be "safer" to not have it here? Or am I being too paranoid ? Or is this OK only in some cases?)

extrowerk: There is way too much tests, more Than 10.000, that's why.

fbrosson: OK, I see. That's a very good reason then.

media-libs/opencv/opencv-3.4.0.recipe
```diff
@@ -241,6 +241,8 @@ INSTALL()
 	# fixCmake
 	sed -i 's,\/include,/'${relativeIncludeDir}',g' \
 		$libDir/cmake/OpenCV/OpenCVConfig.cmake
+	sed -i 's,\/..\/..\/,\/..\/..\/..\/,g' \
```
fbrosson: If you are trying to replace `/../../` by `/../../../` then this is not the correct command.
What would do this is:
```
sed -i 's,/\.\./\.\./,/../../,g' \
```
Remember: in sed, a _dot_ means _any character_. If you really want a dot you need to escape it (on the left side of the substitution. On the righ side, a dot is a dot.

extrowerk: My solution worked great, i see no point to changing it, as i fear i break it again.

fbrosson: The way the current sed command is written, what it does is that it looks for `/ab/cd/` with a, b, c and d being **any** characters, and then it replaces this match by `/../../../`.
So if you are lucky it will do what you are expecting it to do.
But if you are unlucky and there happens to be a match where any (or some, or all) of a, b, c or d are **not** a dot, then you will transform these non dot characters into dots.

BTW, since the sed substitution is using a comma as separator instead of the traditional slash, the slashes in the regexp don't need to be escaped. (What needs to be escaped here are the dots.)

With this release it looks like there are no such matches, but for future releases there could very well be a match. Imagine a `/../ET/` appears in a future release of the file you are patching. You would get `/../../../` and the `ET` would be gone.

You can safely change your sed command to:
```
sed -i -e "s|/\.\./\.\./|&../|g'
```
and it will do exactly what you want: add a `../` everytime you see a `/../../`.
If you try this I'm quite sure you will get the same result as with your current sed command.
The difference is that with this command your are much safer.

But if you wish to keep the sed command the way it is, that's OK for me. (I"m not using opencv on Haiku :)

extrowerk: Your example starts with " and ends with '
Which one should i use then?

Thanks for the explanation, you have right.

fbrosson: Oops, good catch.
Both work (if fixed to make the quotes match.)
(If we had to use e.g. $dataDir (or any environment variable) then the only correct answer would be with the double quotes, since simple quotes prevent evaluation of environment variables.)

PR #1962, opened 2017-12-25
Nomacs : new recipe

extrowerk:
Technically one can disable opencv with:    -DENABLE_OPENCV=OFF


extrowerk:
Currently it depends on opencv

PR #1964, opened 2017-12-26
BeAccessible: create recipe

owenca:
beaccessible

### DIFF
--- a/haiku-apps/beaccessible/beaccessible-1.0.0.recipe
+++ b/haiku-apps/beaccessible/beaccessible-1.0.0.recipe
@@ -1,0 +1,45 @@
+SUMMARY="An Access-like front-end to SQLite"
+DESCRIPTION="BeAccessible, an Access-like front-end to the SQLite database."
+HOMEPAGE="https://github.com/HaikuArchives/BeAccessible"
+COPYRIGHT="2004 randar"
+LICENSE="MIT"
+REVISION="1"
+COMMIT="724b55478f1fcd303f085518679b7d82d7ec1a3d"
+SOURCE_URI="https://github.com/HaikuArchives/BeAccessible/archive/$COMMIT.tar.gz"
+CHECKSUM_SHA256="5fc029666763f29eb1cb494fea1d333cf31b2df74be29849115d5d07f8cd4ad7"
+SOURCE_DIR="BeAccessible-$COMMIT"
+
+ARCHITECTURES="x86_gcc2 !x86 !x86_64"
+
+PROVIDES="
+	beaccessible = $portVersion
+	app:BeAccessible = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libexpat
+	lib:libsqlite3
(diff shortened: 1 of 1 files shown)


owenca:
I will update the recipe to point to the commit in this [PR](https://github.com/HaikuArchives/BeAccessible/pull/2).

haiku-apps/beaccessible/beaccessible-1.0.0.recipe
```diff
@@ -0,0 +1,46 @@
...
+
+BUILD()
+{
+	cd src
+	make $jobArgs OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	mkdir $appsDir
```
fbrosson: I think it is 
It's safer to use "mkdir -p".
Well, it is even better to use install:
```diff
-	mkdir $appsDir
+	install -d $appsDir
-	cp src/objects/BeAccessible $appsDir
+	install -t $appsDir src/objects/BeAccessible
```
I know the result is the same. But by using install we have more flexibility. Image we decide to do something (e.g. add a signature to each runtime we install) then we could modify `cmd:install` to do that, without having to change those recipes which already use install.

haiku-apps/beaccessible/beaccessible-1.0.0.recipe
```diff
@@ -0,0 +1,46 @@
...
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	"
+
+BUILD()
+{
+	cd src
+	make $jobArgs OBJ_DIR=objects
```
fbrosson:
```diff
-	cd src
-	make $jobArgs OBJ_DIR=objects
+	make -C src OBJ_DIR=objects $jobArgs
```

owenca: I had just learned "make -C" and wondered why the recipes I saw didn't use it...

haiku-apps/beaccessible/beaccessible-1.0.0.recipe
owenca: How should an add-on binary like Archiver be installed? I suppose it would be different than a regular executable like BeAccessible.

fbrosson: There is an $addOnsDir. Some examples of use (found right now):
* https://github.com/haikuports/haikuports/tree/master/haiku-apps/konfetti
* https://github.com/haikuports/haikuports/tree/master/app-emulation/virtualbox-guest-additions

owenca: Thanks!

owenca:
Just updated the PR with a new commit. The recipe now uses the [latest merged source files](https://github.com/HaikuArchives/BeAccessible/commit/724b55478f1fcd303f085518679b7d82d7ec1a3d).

PR #1969, opened 2017-12-27
Archiver: create recipe

owenca:
archiver
add-on


scottmc:
Perhaps you can review the work done by a GCI students from 2016 that is still sitting in a pending pull request for archiver?  See if there is anything worth keeping from it and updating it where requested?  Then submit a new pull requests with your fixes, and once that is accepted on HaikuArchives, then update this recipe to point to the newer commit.

owenca:
Just reviewed it. It looks like the student was trying to get Archiver to compile but didn't succeed. I don't see anything from that PR is needed now as Archiver already compiles and runs on Haiku after [this PR](https://github.com/HaikuArchives/Archiver/pull/4).

PR #1985, opened 2017-12-28
libICE, new recipe

Begasus:


x11-libs/libice/libice-1.0.9.recipe
```diff
@@ -0,0 +1,75 @@
...
+	libtoolize --force --copy --install
+	autoreconf -i
+	runConfigure ./configure
+}
+
+INSTALL()
+{
+	make install
+
+	rm -rf $libDir/*.la
```
fbrosson: No need for recursion here:
```diff
-	rm -rf $libDir/*.la
+	rm -f $libDir/*.la
```

Everything else looks OK.
Wait, you can switch HOMEPAGE to https and sort `BUILD_PREREQUIRES`.

And maybe move `devel:util_macros$secondaryArchSuffix` from `BUILD_PREREQUIRES` to `BUILD_REQUIRES`.
Unless it is there for some special reason.

(I did not test anything)

Begasus: Will do, this was based on an earlier X11 lib, so maybe there could be some moves needed too ..

Begasus: Just remembered, you need util_macros to be able to run autoreconf, so I think it should stay in there?

fbrosson: I don't know. Maybe you are right.
If it works this way then I'm not against keeping  `devel:util_macros$secondaryArchSuffix` in `BUILD_PREREQUIRES` (unless someone else sheds some more light telling it should be moved.)

x11-libs/libice/libice-1.0.9.recipe
```diff
@@ -0,0 +1,74 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libice$secondaryArchSuffix \
+	$libDir/libICE.so.$libVersion
+
+BUILD()
+{
+	autoreconf -vfi
+	runConfigure ./configure
```
alaviss: Did you forgot to run anything?

x11-libs/libice/libice-1.0.9.recipe
```diff
@@ -0,0 +1,74 @@
...
+SOURCE_URI="https://www.x.org/releases/individual/lib/libICE-$portVersion.tar.gz"
+CHECKSUM_SHA256="7812a824a66dd654c830d21982749b3b563d9c2dfe0b88b203cefc14a891edc0"
+SOURCE_FILENAME="libICE-$portVersion.tar.gz"
+SOURCE_DIR="libICE-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion=6.3.0
+libVersionCompat="$libVersion compat >= 6"
```
alaviss: I suggest using this construct as this save any future maintainer from editing an another line, which is easily forgotten (I did once and haikuporter doesn't catch it...)

```diff
-libVersionCompat="$libVersion compat >= 6"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```

x11-libs/libice/libice-1.0.9.recipe
Begasus: As? libtoolize is called by autoreconf if you mean that?

x11-libs/libice/libice-1.0.9.recipe
Begasus: Yeah, only seen it recently, so I need to get a hang of that one :)

x11-libs/libice/libice-1.0.9.recipe
alaviss: make?

Begasus: Jikes! (as I told, it was a base from one of the other recipes) didn't notice that one! :+1:

x11-libs/libice/libice-1.0.9.recipe
```diff
@@ -0,0 +1,74 @@
...
+	devel:xproto$secondaryArchSuffix
+	devel:xtrans$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	devel:util_macros$secondaryArchSuffix
+	cmd:aclocal
+	cmd:autoconf
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:gcc$secondaryArchSuffix
```
fbrosson:
```diff
-	cmd:make
 	cmd:gcc$secondaryArchSuffix
+	cmd:make
```

x11-libs/libice/libice-1.0.9.recipe
fbrosson: That's easy to remember:
* \# is for doing pattern maching starting from the left (hint: left = beginning of the line = where you put \# for comments)
* % is for doing pattern maching starting from the right (hint: the percentage character is on the right of the figures)
* use \#\# or %% to make the size of the pattern as big as possible.

x11-libs/libice/libice-1.0.9.recipe
extrowerk: Maybe you should check the original recipe and fix that too ;)

Begasus: Not sure (didn't check) who wrote the initial recipes :) but will check :)

x11-libs/libice/libice-1.0.9.recipe
```diff
@@ -0,0 +1,75 @@
+SUMMARY="Inter-Client Exchange Library"
+DESCRIPTION="The ICE library utilizes callbacks to process incoming \
+messages. Using callbacks allows ProtocolSetup messages and authentication \
+to happen behind the scenes. An additional benefit is that messages never \
+need to be buffered up by the library when the client blocks waiting for \
+a particular message."
+HOMEPAGE="http://xorg.freedesktop.org/releases/individual/lib/"
```
fbrosson:
```diff
-HOMEPAGE="http://xorg.freedesktop.org/releases/individual/lib/"
+HOMEPAGE="https://www.x.org/releases/individual/lib/"
```

x11-libs/libice/libice-1.0.9.recipe
```diff
@@ -0,0 +1,74 @@
...
+a particular message."
+HOMEPAGE="https://www.x.org/releases/individual/lib/"
+COPYRIGHT="1993, 1998 The Open Group"
+LICENSE="MIT (no promotion)"
+REVISION="1"
+SOURCE_URI="https://www.x.org/releases/individual/lib/libICE-$portVersion.tar.bz2"
+CHECKSUM_SHA256="8f7032f2c1c64352b5423f6b48a8ebdc339cc63064af34d66a6c9aa79759e202"
+SOURCE_DIR="libICE-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
```
It builds fine on both x86_gcc2 (using x86 second arch) and x86_64.
Shall we merge after marking x86_64 as tested?

Begasus: Ok, will adjust :+1:

PR #1988, opened 2017-12-28
libmcrypt: new recipe

nysnatuss:
Create recipe for libmcrypt, a replacements for the old Unix crypt.

### DIFF
--- a/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
+++ b/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
@@ -1,0 +1,61 @@
+SUMMARY="Replacements for the old Unix crypt"
+DESCRIPTION="libmcrypt are intended to be replacements for the \
+old Unix crypt, except that they are under the GPL and \
+support an ever-wider range of algorithms and modes."
+HOMEPAGE="http://mcrypt.sourceforge.net/"
+COPYRIGHT="1998-2002 Nikos Mavroyanopoulos"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/project/mcrypt/Libmcrypt/$portVersion/libmcrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="e4eb6c074bbab168ac47b947c195ff8cef9d51a211cdd18ca9c9ef34d27a373e"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	libmcrypt = $portVersion
+	lib:libmcrypt = $libVersionCompat
+	"
+REQUIRES="
+	haiku
+	"
+PROVIDES_devel="
+	devel:libmcrypt = $portVersion
+	"
+REQUIRES_devel="
+	libmcrypt == $portVersion base
+	"
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf -vfi
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/*.la
+
+	prepareInstalledDevelLib libmcrypt
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
+}
--- a/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
+++ b/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
@@ -12,9 +12,12 @@ CHECKSUM_SHA256="e4eb6c074bbab168ac47b947c195ff8cef9d51a211cdd18ca9c9ef34d27a373
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion="4.4.8"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libmcrypt$secondaryArchSuffix = $portVersion
-	lib:libmcrypt$secondaryArchSuffix = $portVersion
+	lib:libmcrypt$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
--- a/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
+++ b/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
@@ -66,6 +66,8 @@ INSTALL()
 	fi
 
 	packageEntries devel \
+		$binDir \
+		$dataDir \
 		$developDir \
 		$maybe_manDir
 }
--- a/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
+++ b/dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
@@ -13,7 +13,7 @@ ARCHITECTURES="x86_gcc2 x86 x86_64"
 
 PROVIDES="
 	libmcrypt = $portVersion
-	lib:libmcrypt = $libVersionCompat
+	lib:libmcrypt = $portVersion
 	"
 REQUIRES="
 	haiku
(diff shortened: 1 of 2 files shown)


dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
```diff
@@ -0,0 +1,61 @@
+SUMMARY="Replacements for the old Unix crypt"
+DESCRIPTION="libmcrypt are intended to be replacements for the \
+old Unix crypt, except that they are under the GPL and \
+support an ever-wider range of algorithms and modes."
+HOMEPAGE="http://mcrypt.sourceforge.net/"
+COPYRIGHT="1998-2002 Nikos Mavroyanopoulos"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/project/mcrypt/Libmcrypt/$portVersion/libmcrypt-$portVersion.tar.gz"
```
fbrosson:
```diff
-SOURCE_URI="https://downloads.sourceforge.net/project/mcrypt/Libmcrypt/$portVersion/libmcrypt-$portVersion.tar.gz"
+SOURCE_URI="https://downloads.sourceforge.net/mcrypt/libmcrypt-$portVersion.tar.gz"
```

You should also add a
```
SECONDARY_ARCHITECTURES="x86"
```
as well as the magic `$secondaryArchSuffix` where appropriate.
This variable is:
* empty for primary arch builds, 
* equal to _x86 when you build mcrypt_x86 on x86_gcc2.

We also like to have some blank lines at some places, as described in:
[HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering)

dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
```diff
@@ -0,0 +1,65 @@
...
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/mcrypt/libmcrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="e4eb6c074bbab168ac47b947c195ff8cef9d51a211cdd18ca9c9ef34d27a373e"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	libmcrypt$secondaryArchSuffix = $portVersion
+	lib:libmcrypt$secondaryArchSuffix = $portVersion
```
fbrosson: We use:
* `$portVersion` for `libmcrypt` and `libmcrypt_devel`
* `$libVersionCompat` for `lib:libmcrypt`
* `$libVersion` for `devel:libmcrypt`

And even if $libVersion is the same as $portVersion, we define it with `libVersion="$portVersion"`.

```diff
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libmcrypt$ secondaryArchSuffix= $portVersion
-	lib:libmcrypt$secondaryArchSuffix = $portVersion
+	lib:libmcrypt$secondaryArchSuffix = $libVersionCompat
 	"
```
We only know if libVersion is portVersion after a compilation (or if we read the Makefiles).

When you call haikuporter, you can use the `--strict-policy` option (same as `-S`).

dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
```diff
@@ -0,0 +1,68 @@
...
+{
+	make install
+
+	rm -f $libDir/*.la
+
+	prepareInstalledDevelLib libmcrypt
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
```
fbrosson: When you think your recipe is OK, you have to examine the contents of each hpkg (with `package list libmcrypt-2.5.8-1-x86_gcc2.hpkg` and `package list libmcrypt_devel-2.5.8-1-x86_gcc2.hpkg`).
With libmcrypt, we can see that :
* it has only one man page and that it is in section 3 (man3), so we can put $manDir in the devel subpackage. But since it builds on all architectures, we can drop it on secondary arch builds. Hence the `maybe_manDir` "trick".
* it has a `libmcrypt-config` in bin, so we need to move it to the devel subpackage. And because it's the only file in `$binDir`, whe can pass the whole `$binDir` tree to `packageEntries devel`.
* it has a `libmcrypt.m4` in `data/aclocal/` and that file is only needed in when you are compiling another program and wish to use autoconf to handle all the shared lib detection work. So you need to move that too to the devel subpackage.

So this is what we need:
```diff
 	prepareInstalledDevelLib libmcrypt
 	fixPkgconfig
 
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $manDir
+		maybe_manDir=
+	else
+		maybe_manDir=$manDir
+	fi
+
 	packageEntries devel \
		$binDir \
		$dataDir \
-		$developDir
+		$developDir \
		$maybe_manDir
 }
```

fbrosson: Oops, I forgot to insert the "+" in front of binDir and dataDir. (These are missing.)

dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
```diff
@@ -0,0 +1,78 @@
...
+	lib:libmcrypt$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libmcrypt${secondaryArchSuffix}_devel = $portVersion
+	devel:libmcrypt$secondaryArchSuffix = $libVersion
+	"
```
fbrosson: Oops, I forgot to spot a missing line in `PROVIDES_devel`. When you run haikuporter with the `--strict-policy` option (or its `-S` alias), it helps you find unadvertised shared library dependencies, as well as unadvertised `cmd:` items in `PROVIDES` and `PROVIDES_devel`. Here we have a `libmcrypt_config` in `bin/` so we need to add a line for it in the `PROVIDES_devel`:
```diff
 PROVIDES_devel="
 	libmcrypt${secondaryArchSuffix}_devel = $portVersion
 	devel:libmcrypt$secondaryArchSuffix = $libVersion
+	cmd:libmcrypt_config$secondaryArchSuffix = $portVersion
 	"
```

dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
```diff
@@ -0,0 +1,78 @@
+SUMMARY="Replacements for the old Unix crypt"
+DESCRIPTION="libmcrypt are intended to be replacements for the \
+old Unix crypt, except that they are under the GPL and \
+support an ever-wider range of algorithms and modes."
```
fbrosson: Hmm, "libmcrypt are intended" does not sound very well. Shall we use the following description, instead?
```
DESCRIPTION="libmcrypt is the companion library used by Mcrypt, which is \
itself an enhanced replacement for the old crypt command.
libmcrypt allows developers to use a wide range of encryption functions \
without making drastic changes to their code. It contains the actual \
encryption functions and provides a standardized mechanism for accessing \
them."
```
It's based on the description found in http://mcrypt.sourceforge.net/

dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
```diff
@@ -0,0 +1,78 @@
...
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
```
Begasus: Maybe add a "defineDebugInfoPackage" here?

fbrosson: Why not? That's a good idea. Thanks @Begasus !

@nysnatuss, this can be done by inserting these lines before BUILD (with a blank line before and another one after):
```
defineDebugInfoPackage libmcrypt$secondaryArchSuffix \
	$libDir/libmcrypt.so.$libVersion
```

dev-libs/libmcrypt/libmcrypt-2.5.8.recipe
nysnatuss: @fbrosson Do we also need to add it at `PROVIDES`?

fbrosson: No, it's only needed in `PROVIDES_devel` because the `libmcrypt-config` command will only be in the devel subpackage. (There are no files in bin/ shipped in the base package, for this recipe.)

PR #1843, opened 2017-12-03
man-db, new recipe

Begasus:
Not working yet, still getting errors in the build section ...

PR #1990, opened 2017-12-29
[WIP] app-text/mandoc: add recipe

alaviss:
**Not** done as part of GCI 2017

Changes to mandoc:
- Allow changing `/etc`
- Make configure preserve `PATH`

Status:
- `man` works
- `mandoc.db` can't be created as it has to be in the same directories as the man pages, which are read-only

Runs on:
- [x] x86_gcc2 (using x86 secondary arch)
- [ ] x86_64 (untested)


app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,92 @@
...
+SOURCE_URI="http://mandoc.bsd.lv/snapshots/mandoc-$portVersion.tar.gz"
+CHECKSUM_SHA256="0b0c8f67958c1569ead4b690680c337984b879dfd2ad4648d96924332fd99528"
+PATCHES="mandoc-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+CONFLICTS="
+	cmd:man
+"
```
fbrosson: @alaviss, you rock! It's great to have you with us, both as a GCI participant and as a mentor, which you also do very well!

And, of course, it is also great to have contributions outside the GCI contest!

== new topic ==

I haven't tested mandoc yet and I don't know what exactly you mean by "Allow to change /etc".
Haiku has a symlink for etc:
```
/etc -> /boot/system/settings/etc
```
so you could even leave that unchanged. But if you wish, you can use `finddir B_COMMON_ETC_DIRECTORY` (`/boot/system/settings/etc`) or `finddir B_SYSTEM_SETTINGS_DIRECTORY` (`/boot/system/settings`), or even `find_directory()`.
This is entirely up to you.

== new topic ==

We generally put the CONFLICTS bloc right after the REQUIRES bloc. BTW, the indentation of the closing double quotes needs a tab:
```diff
 CONFLICTS="
 	cmd:man
-"
+	"
```

alaviss: Sorry, I wasn't really clear back then. What I did was moving the `man.conf` to `B_SYSTEM_SETTINGS_DIRECTORY`

fbrosson: Hmm, Haiku already has a `man.conf` (and a `man.conf.default`) in there (/boot/system/settings/) so it might only be possible to put that file in this directory if it is compatible with the man package.
BTW, would it make sense to co-install man and mandoc ? Or is the latter a replacement for the former?

== new topic ==

BTW, for the implementation you might try using find_directory.

Regarding the DB cache you mentioned earlier, I guess B_SYSTEM_CACHE_DIRECTORY is where the files should be generated.

alaviss: The latter is a replacement for the former, given that our man port is ages old and upstream have abandoned it

I will definitely use `find_directory()`, however, the code path is complex and need further analysis before I could make changes to mandoc

fbrosson: We can easily add a GCI task for that so that you can easily have +1 in your counter of completed tasks. (If you add a 👍 to this comment I'll understand.)

fbrosson: FWIW, the task is already created (but is not yet published.)
GCI mentors, if you wish to co-mentor (or only have a look at) that task, look for "[PUBLISH ME PLEASE]" in TASKS (and uncheck "MY TASKS")

fbrosson:
@alaviss, in case you did not notice, @Begasus has #1843 for man-db, which, if I understand correctly, is needed by mandoc.
(I'll see if man-db can be merged.)

alaviss:
They're different `man` implementations. `man-db` is heavy and ugly, which is why I declare to port `mandoc`

fbrosson:
OK, thanks. I got it ;)

alaviss:
Updates:
- I don't remember what I changed lol

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,95 @@
...
+	cmd:soelim = $portVersion
+	cmd:whatis = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+CONFLICTS="
+	cmd:man
+"
```
fbrosson:
```diff
 CONFLICTS="
 	cmd:man
-"
+	"
```

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,95 @@
...
+BUILD_PREREQUIRES="
+	cmd:cc$secondaryArchSuffix
+	cmd:make
+	"
+
+TEST_PREREQUIRES="
+	cmd:diff
+	cmd:perl
+	"
+BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
```
fbrosson: Shall we shorten it a little?
```diff
-TEST_PREREQUIRES="
+TEST_REQUIRES="
 	cmd:diff
 	cmd:perl
 	"
-BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
+BUILD_PREREQUIRES+="$TEST_REQUIRES"
```

fbrosson:
FWIW, the `man` package, which we already have in HaikuPorts, has 2 `cmd:*` items which are not provided by `mandoc`: `cmd:man2dvi` and `cmd:man2html`. Hopefully, none of them are required by any other package. Moreover, `cmd:man2html` is even provided by another package: `dev-util/hxtools`.

This means we will be able to make the transition from `man` to `mandoc` (for the default distribution) if/when a consensus is found :)

alaviss:
Updates:
- Hopefully more infomative errors

alaviss:
Updates:
- Implemented the ability to dynamically create necessary folders as needed

app-text/mandoc/mandoc-1.14.3.recipe
alaviss: Any tips regarding commiting whitespace fixes? git doesn't register this one for me :(

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,94 @@
+SUMMARY="A suite of tools compiling mdoc from the OpenBSD project"
+DESCRIPTION="A suite of tools compiling mdoc, the roff macro language of choice \
+for BSD manual pages, and man, the predominant historical language for UNIX \
+manuals. It is small, ISO C, ISC-licensed, and quite fast. The main component \
+of the toolset is the mandoc utility program, based on the libmandoc validating \
+compiler, to format output for UTF-8 and ASCII UNIX terminals, HTML 5, \
+PostScript, and PDF."
+HOMEPAGE="http://mandoc.bsd.lv/"
+COPYRIGHT="1989-2017 mandoc Authors"
```
korli: Please copy the authors list in the license text here.

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,94 @@
...
+# Note: update LICENSE on every new version
+LICENSE="mandoc-$portVersion"
```
korli: You could simply list the licenses (ISC, BSD 2 and 3 clause).  ISC seems to be missing at https://git.haiku-os.org/haiku/tree/data/system/data/licenses though

app-text/mandoc/mandoc-1.14.3.recipe
alaviss: The list is growing and changing after every release, that's why I kept them outside

alaviss:
As for licensing, would it be OK if I do this:

- List all license, add ISC with only permission notice
- Export all copyright from "License.txt" included with the source code then save in a file called "Copyright" which is installed in `$docDir`

app-text/mandoc/mandoc-1.14.3.recipe
fbrosson: I'd use this:
```
LICENSE="mandoc-$portVersion
	BSD (2-clause)
	BSD (3-clause)
	ISC"
```

BTW, if a recipe for another release of mandoc is added it will probably have a very similar `LICENSE` file, so maybe you can rename `mandoc-$portVersion` to just `mandoc`. If anything is added to the `LICENSE` file on any future release of mandoc then you would only need to update `app-text/mandoc/licenses/mandoc`.

Regarding the `ISC` license, are there any plans to add it to the pool of licenses shipped in `/system/data/licenses/` by the haiku package?

app-text/mandoc/licenses/mandoc-1.14.3
```diff
@@ -0,0 +1,55 @@
+
```
fbrosson: Oops, there's a carriage-return (0x0d) in the first line. (This reminds me popular OS which I don't use.)
It would be nice to drop that CR. I'm OK with keeping an empty line, as long as it does not contain that 0x0d character :)

alaviss:
Updates:
- Follows @korli suggestions regarding licenses and copyright.
- Disable test since it would only works if UTF-8 support is enabled, which is blocked by [ticket#13948][0].

[0]: https://dev.haiku-os.org/ticket/13948

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,113 @@
...
+	cat > configure.local << EOF
+		CFLAGS='-O2 -pipe -D_BSD_SOURCE'
+		DBDIR='$(finddir B_SYSTEM_VAR_DIRECTORY)/mandoc'
+		LN='ln -sf'
+		MANDIR='$manDir'
+		MANPATH_BASE='$(IFS=:; echo "${manPathLint[*]}")'
+		MANPATH_DEFAULT='$(IFS=:; echo "${manPathDef[*]}")'
+		PREFIX='$prefix'
+		SBINDIR="\${PREFIX}/bin"
+		SYSCONFDIR='$portPackageLinksDir/.settings'
```
fbrosson: @alaviss, I noticed that you made this change:
```diff
-		SYSCONFDIR='$settingsDir'
+		SYSCONFDIR='$portPackageLinksDir/.settings'
```
I don't know yet if this is correct. I need to check.

alaviss: I grab it straight from HaikuPorter source: https://github.com/haikuports/haikuporter/blob/2b7ba23e93e799bb9ceeab552266cc1e35b26c5f/HaikuPorter/Port.py#L1173

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,113 @@
...
+	cmd:mandocd = $portVersion
+	cmd:soelim = $portVersion
+	cmd:whatis = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+CONFLICTS="
+	cmd:man
```
fbrosson: I though pkgman would not correctly handle the `cmd:man` in `CONFLICTS`, but it did :)
(`pkgman` correctly asked me the permission to uninstall `man` when I tried to install `mandoc`, and also correctly asked me the permission to uninstall `mandoc` when I tried to install `man`.)

But I still think it would be better to make the following change:
```diff
CONFLICTS="
-	cmd:man
+	man
 	"
```
This is beacause the `CONFLICTS` we have in **_all_** other recipes **_all_** use package names. So using a `cmd:` item here would mean we would have to be very careful about this in future modifications of pkgman.

alaviss: From [PackageManagement/BuildingPackages][0]

>     entity		::= entity_name [ "=" version_ref ] [ ( "compat" | "compatible" ) ">=" version_ref ]
>     entity_name	::= [ entity_type ":" ] entity_name_char+
>     entity_type	::= "lib" | "cmd" | "app" | "add_on"

>     required_entity		::= entity_name [ version_operator version_ref [ "base" ] ]
>     version_operator	::= "<" | "<=" | "==" | "!=" | ">=" | ">"

> - `conflicts`: A list of entities that this package conflicts with (i.e. only one of both can be installed at any time). The list elements must have the `required_entity` format. 

This means `cmd:man` as `CONFLICTS` will always be supported. And as in `REQUIRES`, we should be specific with what we `CONFLICTS`.

[0]: https://dev.haiku-os.org/wiki/Obsolete/MovedToTree/PackageManagement/BuildingPackages

fbrosson: Well, you are right in theory. But I guess the `CONFLICTS` variable was meant to be used with package names only. If it had been meant for `cmd:` items, then we could just use the `cmd:` in `PROVIDES` and not bother to define any `CONFLICTS`.
Moreover, there is a small advantage to using package names: there is no need to do a search.

Since we have no other recipe with `cmd:` items in a `CONFLICTS`, I'd like to get the green lights from one or two key contributors. If nobody objects having a `cmd:man` in mandoc's `CONFLICTS` then I'll be OK with that :)
(But I guess it will be better to switch to the variant.)

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,113 @@
...
+EOF
+	./configure
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
```
fbrosson: Some suggested polishing:
```diff
 INSTALL()
 {
 	make install
+
+	# man.conf is not in /etc/ but in $settingsDir
+	sed -i -e "s|/etc\(/man\.conf\)$|`finddir
+			B_SYSTEM_SETTINGS_DIRECTORY`\1|;" \
+		$manDir/man1/apropos.1 \
+		$manDir/man1/man.1 \
+		$manDir/man5/man.conf.5 \
+		$manDir/man8/makewhatis.8
 }
```
(Another sed command is comming soon.)

fbrosson: Since the recipe adds a default `MANPATH` in `configure.local`: `MANPATH_DEFAULT='/boot/home/config/non-packaged/documentation/man:/boot/home/config/documentation/man:/boot/system/non-packaged/documentation/man:/boot/system/documentation/man'`
you might wish to use the following sed:
```
	sed -i \
		-e "/^manpath \/usr\/share\/man/ i manpath `finddir \
			B_USER_NONPACKAGED_DOCUMENTATION_DIRECTORY`/man" \
		-e "/^manpath/ s|/usr/share/man|`finddir \
			B_USER_DOCUMENTATION_DIRECTORY`/man|;" \
		-e "/^manpath/ s|/usr/X11R6/man|`finddir \
			B_SYSTEM_NONPACKAGED_DOCUMENTATION_DIRECTORY`/man|;" \
		-e "/^manpath/ s|/usr/local/man|`finddir \
			B_SYSTEM_DOCUMENTATION_DIRECTORY`/man|;" \
		$manDir/man5/man.conf.5
```
to have these lines:
```
manpath /usr/share/man
manpath /usr/X11R6/man
manpath /usr/local/man
```
replaced by these ones:
```
manpath /boot/home/config/non-packaged/documentation/man
manpath /boot/home/config/documentation/man
manpath /boot/system/non-packaged/documentation/man
manpath /boot/system/documentation/man
```

BTW, the constants associated to the above paths (without the "/man" suffix) are respectively, as we know:
```
B_USER_NONPACKAGED_DOCUMENTATION_DIRECTORY
B_USER_DOCUMENTATION_DIRECTORY
B_SYSTEM_NONPACKAGED_DOCUMENTATION_DIRECTORY
B_SYSTEM_DOCUMENTATION_DIRECTORY
```

BTW, the recipe is hardcoding build-time defaults. That's OK for me, but if you would like to try some changes to have a dynamic default config by calling `find_directory()` when there is no `/system/settings/man.conf`, then this would be even better :)
Well, I guess this would be quite some extra work.

alaviss: @fbrosson There's no `/system/settings/man.conf` by default :laughing:

I would like to use `find_directory()`, but the amount of changes that has to be made is huge, and I don't think it's maintainable. Still, I'd have to figure it out somehow, since this would be a problem once we have proper multi-user support.

Ideas:
- Add a script that would automatically populate `man.conf` with `finddir` output (this would fail once multi-user support exists)
- Find a way to inject `find_directory()` to `mandoc`

fbrosson: Yep, I know. mandoc is smart enough to work without any config file :)
(But I did check that the config file was being read (and taken into account) when MANPATH is not defined. My conclusion is that everything is working as expected.)

BTW, there is a small issue: the man package that is currently pre-installed in the official Haiku images does come with a `/system/settings/man.conf`, and the sad news is that the config keywords are all upper-case (`MANPATH`), while mandoc requires it to be lower-case.
But I think it is OK to keep mandoc unchanged. The manual page is very clear about the keywords being lower case. Anyone willing to add custom manpaths can do so with the correct syntax.

Regarding the script to populate a default `man.conf`, I don't think it would be a good idea. The default built-in values are OK. And there is no urgency in teaching mandoc to use dynamic defaults. And, as you pointed, such a change would not be easy to maintain.
So I'd say that the recipe will be OK after you add the sed command to replace `/etc/man.conf` by `/boot/system/settings/man.conf` in the man pages (using the find_dir command in the recipe, of course.)
Regarding the complicated sed command to patch `$manDir/man5/man.conf.5`, it's just a nice to have and I'm OK with keeping `manpath /usr/share/man` + `manpath /usr/X11R6/man` + `manpath /usr/local/man` in that man page.

Regarding the _injection of `find_directory()` in mandoc_, I guess what you mean is finding a smart patch to do that in a way that would be easy to maintain. I'd say that this would be welcome but you really don't need to worry about that.

pulkomandy:
On Mon, Jan 29, 2018 at 09:02:20PM -0800, alaviss wrote:
> I would like to use `find_directory()`, but the amount of changes that has to be made is huge, and I don't think it's maintainable. Still, I'd have to figure it out somehow, since this would be a problem once we have proper multi-user support.

If the changes are done cleanly, they can be sent to upstream. Then,
they will take care of maintenance for us.

app-text/mandoc/mandoc-1.14.3.recipe
alaviss: Good news, the manpath code is simple enough that it would be easy to add dynamic folder searching.

Bad news, the code is based on the assumption that it won't fail, but `find_directory()` can. I would need some advice on what to do in case `find_directory()` fails.

pulkomandy: If it fails, you can still return an hardcoded path.

fbrosson: @pulkomandy, I'm a bit surprised by your answer :)
Wouldn't it be better to just skip `find_directory`'s result if it fails?
If `find_directory` ever fails, it will certainly be for a good reason. So returning a hardcoded path in that case would probably not be very useful.
And in case a directory did exist but `find_directory` would fail to return that directory's path, then it would be much easier to detect this error if mandoc were not told to fallback to a hardcoded path.

pulkomandy: > the code is based on the assumption that it won't fail

We could add error handling support, but I would say that it is not up to us. If the upstream decided to not handle errors at this level, the simplest thing to do is to always return some path, then let them handle errors in appropriate places.

If I was writing my own software, I would handle find_directory failure differently. But here we are doing a port, and I think patching in extra error handling where the original author did not think it was useful is not necessary. Especially as find_directory currently cannot fail, even if the API allows it.

fbrosson: mandoc could be patched to either:
* skip any directory that `find_directory()` fails to return,
* or provide a hardcoded fallback.

Now, since we know that `find_directory()` cannot (currently) fail, I suggest that mandoc silently skips any failure, if that ever happens in the future. There is no need, imho, to embed any hardcoded path.
Testing `find_directory()`s return code is the minimum, and this doesn't cost much if mandoc chooses to skip a manpath in case `find_directory()` ever fails.

That said, if it is easier to provide a hardcoded path than ignoring a failed call, then I'm OK with embedding hardcoded paths.

alaviss:
Updates:
- Make mandoc populate man directories dynamically using `find_directory()`
- Use functions from `libbsd` instead

fbrosson:
Cool! Thanks @alaviss for your recent enhancements!
(I've had a look at your patch and I think it is nicely done :)

Already tested on x86_gcc2 (using x86 secondary arch) and everything is OK.

I can merge your PR but I think it would be nice to have [this tiny change](#discussion_r164482894) before the merge. (Of course, you can drop the comment above de sed command.) Thanks!

alaviss:
Updates:
- Change `/etc/man.conf` to `B_SYSTEM_SETTINGS_DIR/man.conf` for man pages

app-text/mandoc/mandoc-1.14.3.recipe
```diff
@@ -0,0 +1,110 @@
...
+		DBDIR='$(finddir B_SYSTEM_VAR_DIRECTORY)/mandoc'
+		HAVE_ERR=1
+		HAVE_PROGNAME=1
+		HAVE_STRSEP=1
+		LDFLAGS="-lbsd"
+		LN='ln -sf'
+		MANDIR='$manDir'
+		PREFIX='$prefix'
+		SBINDIR="\${PREFIX}/bin"
+		SYSCONFDIR='$portPackageLinksDir/.settings'
```
fbrosson: This works fine, but I think this "variant" might be prefered:
```diff
-		SYSCONFDIR='$portPackageLinksDir/.settings'
+		SYSCONFDIR='$sysconfDir'
```
(Both possibilities produce the exact same path.)

If you prefer, you can use `SYSCONFDIR='$settingsDir'`. BTW, this was what I used for my tests.
(`$settingsDir` is in the wiki, whereas `$sysconfDir` is not, but I guess it is still acceptable to use `$sysconfDir`.)

app-text/mandoc/mandoc-1.14.3.recipe
fbrosson: @pulkomandy & @korli, are you OK with having `cmd:man` in `CONFLICTS` ?
It is working fine and, as I wrote previously, pkgman correctly handles this.

I'm asking the question because this would be the first recipe to use a `cmd:` item in a `CONFLICTS`.
I'm OK with this, but I'd like to have your thoughts before we merge. Thanks!

app-text/mandoc/mandoc-1.14.3.recipe
alaviss: More docs ahead (#2135)

alaviss:
Updates:
- Uses `sysconfDir` as suggested by @fbrosson

fbrosson:
LGTM.

@korli and/or @pulkomandy, could you tell me your thoughts about what we have in [`CONFLICTS`](#discussion_r164928571) ? Thanks!

pulkomandy:
On Tue, Jan 30, 2018 at 05:08:46PM -0800, fbrosson wrote:
> @pulkomandy & @korli, are you OK with having `cmd:man` in `CONFLICTS` ?
> It is working fine and, as I wrote previously, pkgman correctly handles this.

If pkgman is happy with it, I see no problem.

fbrosson:
Cool, thanks @pulkomandy!

@alaviss, if I'm not mistaken, you would like this PR to be merged with the "Rebase and merge" button (instead of the "Squash and merge" button) to avoid the "on-the-fly" squashing of your 2 commits.

So, if this is the case, would you like to update the commit title to include a ` (#1990)` for each of your commits (or the first one only) before I merge this PR?
(FWIW, this is because the "Rebase and merge" method does not allow GitHub to append a reference to this PR in the commit title.)

Issue #1983, opened 2017-12-28
lpairs-1.0.4 failed to build on x86_64


nysnatuss:
[games-puzzle/lpairs-1.0.4](https://github.com/haikuports/haikuports/blob/master/games-puzzle/lpairs/lpairs-1.0.4.recipe) failed to build on x86_64.

Error: ([Full log](https://pastebin.com/CGHLbFu9))
```
source='main.c' object='main.o' libtool=no \
DEPDIR=.deps depmode=none /bin/sh ../depcomp \
gcc -DHAVE_CONFIG_H -DSRC_DIR=\"/packages/lpairs-1.0.4-1/.self/data/lpairs/\" -DSOUND -DLOCALEDIR=\"/packages/lpairs-1.0.4-1/.self/data/locale\" -I. -I. -I..     -g -O2 -I/packages/libsdl-1.2.15-13/.self/develop/headers/SDL -D_GNU_SOURCE=1 -DSDL_1_1_5 -c main.c
/bin/sh: ../depcomp: No such file or directory
```
It seems like it tried to search depcomp in `/bin/sh/../depcomp` which is of course, non-existent in Haiku.

Begasus:
Try adding "autoreconf -vfi" after libtoolize and befor runConfigure?

Oyami-Srk:
Fixed according to @Begasus comment.

Issue #1982, opened 2017-12-28
bdhcalc-1.1 failed to build on x86_64


nysnatuss:
[haiku-apps/bdhcalc-1.1](https://github.com/haikuports/haikuports/blob/master/haiku-apps/bdhcalc/bdhcalc-1.1.recipe) failed to build on x86_64.

Error: ([Full log](https://pastebin.com/DgyFtWSV))
```
In file included from src/bdh_main.cpp:6:0:
src/bdh_view.h:38:21: error: extra qualification 'bdh_view::' on member 'pwr2' [-fpermissive]
  unsigned long long bdh_view::pwr2(unsigned long long power);
                     ^
src/bdh_view.h:39:21: error: extra qualification 'bdh_view::' on member 'pwr10' [-fpermissive]
  unsigned long long bdh_view::pwr10(unsigned long long power);
                     ^
In file included from /boot/system/develop/headers/os/interface/Alert.h:11:0,
                 from /boot/system/develop/headers/os/InterfaceKit.h:6,
                 from src/bdh_main.cpp:1:
/boot/system/develop/headers/os/interface/Window.h: In member function 'virtual void bdh_window::MessageReceived(BMessage*)':
/boot/system/develop/headers/os/interface/Window.h:291:18: error: 'typedef class BLooper BWindow::inherited' is private
  typedef BLooper inherited;
                  ^
src/bdh_main.cpp:44:2: error: within this context
  inherited::MessageReceived(msg);
  ^
```
I think we can just remove the `bdh_view::` and make `typedef class BLooper BWindow::inherited` public.

Oyami-Srk:
Fixed.

nysnatuss:
Fixed in #1993.

Issue #1980, opened 2017-12-28
ppviewer-1.0.0 failed to build on x86_64


nysnatuss:
[haiku-apps/ppviewer-1.0.0](https://github.com/haikuports/haikuports/blob/master/haiku-apps/ppviewer/ppviewer-1.0.0.recipe) failed to build on x86_64.

Error: [Full log](https://pastebin.com/58RaF4qX)
```
MainWindow.cpp: In constructor 'MainWindow::MainWindow(BRect, const char*, BFile*)':
MainWindow.cpp:69:98: error: taking address of temporary [-fpermissive]
  contentView=new BTextView(rect, "contentView", text_rect,be_fixed_font, &((rgb_color){0,0,0,255}),B_FOLLOW_ALL_SIDES, B_WILL_DRAW);
                                                                                                  ^
MainWindow.cpp: In constructor 'MainWindow::MainWindow(BRect, const char*, const char*, off_t)':
MainWindow.cpp:129:98: error: taking address of temporary [-fpermissive]
  contentView=new BTextView(rect, "contentView", text_rect,be_fixed_font, &((rgb_color){0,0,0,255}),B_FOLLOW_ALL_SIDES, B_WILL_DRAW);
                                                                                                  ^
```

Oyami-Srk:
Fixed.

nysnatuss:
Fixed in #1993.

Issue #1979, opened 2017-12-28
helios-1.7.2 failed to build on x86_64


nysnatuss:
[haiku-apps/helios-1.7.2](https://github.com/haikuports/haikuports/blob/master/haiku-apps/helios/helios-1.7.2.recipe) failed to build on x86_64.

The main error is: ([Full log](https://pastebin.com/ABxZYz8M))
```
error: call of overloaded 'RemoveItem(long int)' is ambiguous

...

note: candidate: bool BMenu::RemoveItem(BMenuItem*)
    bool    RemoveItem(BMenuItem* item);
            ^
note: candidate: BMenuItem* BMenu::RemoveItem(int32)
    BMenuItem*   RemoveItem(int32 index);
                 ^
note: candidate: bool BMenu::RemoveItem(BMenu*)
    bool    RemoveItem(BMenu* menu);
            ^
```
It seems like `RemoveItem()` does not accept long int and throw an error instead. ([Ref](https://stackoverflow.com/questions/4672152/call-of-overloaded-function-is-ambiguous))

Oyami-Srk:
Which version of Haiku are you using? I tested in Haiku 51602 and the atomic_add requires int32(I think in x64, atomic add should requires int64.

Oyami-Srk:
Fixed

scottmc:
Please upstream the patch to HaikuArchives, once that is accepted, then this issue can be closed.

nysnatuss:
Fixed in #1993.

PR #1993, opened 2017-12-29
Fix 4 building failed in x86_64 issues.

Oyami-Srk:
Issues: #1979 #1980 #1982 #1983


games-puzzle/lpairs/lpairs-1.0.4.recipe
```diff
@@ -27,14 +27,19 @@ BUILD_REQUIRES="
...
+	cmd:autoreconf
 	"
+if [ "$buildArchitecture"x == "x86_64"x ] ; then
+PATCHES="lpairs-1.0.4.patchset"
+fi
 
 BUILD()
 {
 	libtoolize --force --copy --install
+	autoreconf -vfi
```
alaviss: Please don't run `autoreconf` after modifying `configure` as this command will generate a brand new `configure`, overriding our changes. It's advised to modify `configure.ac` instead

haiku-apps/bdhcalc/bdhcalc-1.1.recipe
```diff
@@ -27,6 +27,10 @@ BUILD_PREREQUIRES="
 	cmd:g++
 	"
 
+if [ "$buildArchitecture"x == "x86_64"x ] ; then
+PATCHES="bdhcalc-1.1.patchset"
```
alaviss: Only x86_gcc2 use gcc2. And I believe we don't need an trailing `x`

```diff
-if [ "$buildArchitecture"x == "x86_64"x ] ; then
+if [ "$buildArchitecture" != "x86_gcc2" ] ; then
```

haiku-apps/helios/helios-1.7.2.recipe
```diff
@@ -24,8 +24,11 @@ BUILD_REQUIRES="
 BUILD_PREREQUIRES="
 	makefile_engine
 	cmd:make
-	cmd:gcc
+	cmd:g++
 	"
+if [ "$buildArchitecture"x == "x86_64"x ] ; then
```
alaviss: Same as above

```diff
-if [ "$buildArchitecture"x == "x86_64"x ] ; then
+if [ "$buildArchitecture" != "x86_gcc2" ] ; then
```

haiku-apps/ppviewer/ppviewer-1.0.0.recipe
```diff
@@ -25,6 +25,10 @@ BUILD_PREREQUIRES="
 	cmd:g++
 	"
 
+if [ "$buildArchitecture"x == "x86_64"x ] ; then
```
alaviss: Same as above

```diff
-if [ "$buildArchitecture"x == "x86_64"x ] ; then
+if [ "$buildArchitecture" != "x86_gcc2" ] ; then
```

games-puzzle/lpairs/lpairs-1.0.4.recipe
Oyami-Srk: Thanks for the review! autoreconf solved missing "../depcomp". I disabled dependency tracking instead of using autoreconf to fix it. But I don't know whether it's good or not.
Program build well and run properly for me.

haiku-apps/bdhcalc/bdhcalc-1.1.recipe
fbrosson: I'd strongly recommend using `$effectiveTargetArchitecture` which is the same as `$targetArchitecture` for primary architectures. On recipes which do not support secondary architectures, this makes no difference. But on the other recipes, using `$buildArchitecture` (to decide e.g. if a patch is needed or not) is generally not appropriate.
Therefore, to keep consistency across all recipes, it is better to use the variable which works on all cases, and this is `$effectiveTargetArchitecture`.

On secondary arch builds, e.g. x86 second arch on x86_gcc2, we have:
```
effectiveTargetArchitecture='x86'
buildArchitecture='x86_gcc2'
targetArchitecture='x86_gcc2'
```
BTW, this is the output of `grep rchitecture= wrapper-script` inside the `work-x86-$portVersion` of any package being built on x86_gcc2 for the x86 secondary arch.

nysnatuss:
Do `REVISION=1` need to be updated in this PR?

Oyami-Srk:
Thanks for your reminder, @fbrosson . I think I should separating problems of GCC and x64 itself by using `$effectiveTargetArchitecture` and `$buildArchitecture`. May Haiku release a 32-bits version with GCC5 instead of outdated GCC2?

haiku-apps/bdhcalc/bdhcalc-1.1.recipe
```diff
@@ -27,6 +27,10 @@ BUILD_PREREQUIRES="
 	cmd:g++
 	"
 
+if [ "$effectiveTargetArchitecture" == "x86_64" ] ; then
```
fbrosson: * If it's a gcc2 vs gcc5 test, then you should use:
```
if [ "$effectiveTargetArchitecture" != x86_gcc2 ] ; then
```
(This is because an (unsupported) Haiku/x86 exists and is 32 bits with gcc5.)

* If it's a 32 bit vs 64 bis issue, then I would recommend using `=` instead of `==`
```
if [ "$effectiveTargetArchitecture" = x86_64 ] ; then
```
(Using `==` is a bashism. `=` works on `sh` and `bash`, while `==` only works on `bash`)

* BTW, you can keep the quotes if you wish, but, as you certainly know, they are not required in string constants (as long as they don't have unprotected white spaces.)

haiku-apps/helios/helios-1.7.2.recipe
```diff
@@ -3,11 +3,11 @@ DESCRIPTION="Cdrecord GUI frontend."
 HOMEPAGE="https://github.com/HaikuArchives/Helios"
 COPYRIGHT="2000-2004 Maurice Michalski"
 LICENSE="MIT"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="git+https://github.com/HaikuArchives/Helios.git#ffbbca7"
```
alaviss: You should just move this to a later commit instead. Looks like most x86_64 issues has been fixed upstream. See https://github.com/HaikuArchives/Helios/commit/b870e477789928027dd4f3a444219cb820944678

haiku-apps/ppviewer/ppviewer-1.0.0.recipe
```diff
@@ -25,6 +25,10 @@ BUILD_PREREQUIRES="
 	cmd:g++
 	"
 
+if [ "$effectiveTargetArchitecture" == "x86_64" ] ; then
+PATCHES="ppviewer-1.0.0.patchset"
```
fbrosson: We like to use `$portVersion` when that's possible :)
```diff
-PATCHES="ppviewer-1.0.0.patchset"
+PATCHES="ppviewer-$portVersion.patchset"
```
(you can do the same change on ther other recipes you are already editing.)

haiku-apps/bdhcalc/bdhcalc-1.1.recipe
```diff
@@ -27,6 +27,10 @@ BUILD_PREREQUIRES="
...
+PATCHES="bdhcalc-1.1.patchset"
```
alaviss: I believe this patch could be upstreamed, and I encourage you to do so

haiku-apps/ppviewer/ppviewer-1.0.0.recipe
alaviss: This patch also looks upstreamable, and it should also work for gcc2 (untested)

haiku-apps/helios/helios-1.7.2.recipe
```diff
@@ -3,11 +3,11 @@ DESCRIPTION="Cdrecord GUI frontend."
 HOMEPAGE="https://github.com/HaikuArchives/Helios"
 COPYRIGHT="2000-2004 Maurice Michalski"
 LICENSE="MIT"
-REVISION="1"
-SOURCE_URI="git+https://github.com/HaikuArchives/Helios.git#ffbbca7"
-CHECKSUM_SHA256="ce82ede4fd218d3f0c7e30cafc061d2a377e473dbc88512a61e286f21ae51925"
+REVISION="3"
```
alaviss: Your PR hasn't been merged yet, so you don't have to bump revision. Revision bumping is only needed for packages that builds in the main repo

```diff
-REVISION="3"
+REVISION="2"
```

games-puzzle/lpairs/patches/lpairs-1.0.4.patchset
```diff
@@ -69,5 +63,28 @@ index 301878a..a786e39 100644
...
+diff --git a/lpairs/sdl.c b/lpairs/sdl.c
+index 5937097..e4779c4 100644
+--- a/lpairs/sdl.c
++++ b/lpairs/sdl.c
+@@ -480,7 +480,7 @@ void hardware_cap()
+ /*
+     update rectangle (0,0,0,0)->fullscreen
+ */
+-inline void refresh_screen(int x, int y, int w, int h)
++void refresh_screen(int x, int y, int w, int h)
```
alaviss: May I ask what error happened during build that requires you to remove `inline`?

Please also note that `inline` is also specified in `sdl.h`

Oyami-Srk: Because the `inline void refresh_screen(int x, int y, int w, int h)` has no definition in header, compiler cannot find the reference to this function. So the `undefined reference to refresh_screen` will happened. The inline function in a cpp file will be only available to that cpp file(it's inline function, so compiler will not export a alone function reference to object). Header  declares but no definition. Another solution is move the body to header(actually this is more reasonable, but I want to keep the minor fix).

haiku-apps/helios/helios-1.7.2.recipe
Oyami-Srk: Understand. Thanks.

Oyami-Srk:
Completed. Patches of BDH-Calc and PPViewer have been merged to upstream.

games-puzzle/lpairs/lpairs-1.0.4.recipe
```diff
@@ -27,15 +27,18 @@ BUILD_REQUIRES="
 	"
 BUILD_PREREQUIRES="
 	cmd:make
-	cmd:gcc
+	cmd:g++
 	cmd:ld
 	cmd:libtoolize
 	"
+if [ "$effectiveTargetArchitecture" == "x86_64" ] ; then
+PATCHES="lpairs-1.0.4.patchset"
```
alaviss: **WARNING:** All of this is just speculation, no actual testing has been done

It appears to me that your patchset won't cause trouble for gcc2, so you could safely remove this conditional

PR #2001, opened 2017-12-30
mcrypt: new recipe

nysnatuss:
Add recipe for [mcrypt](http://mcrypt.sourceforge.net/), an enhanced replacement for the old crypt command.

### DIFF
--- a/app-crypt/mcrypt/mcrypt-2.6.8.recipe
+++ b/app-crypt/mcrypt/mcrypt-2.6.8.recipe
@@ -1,3 +1,4 @@
+SUMMARY="Replacements for the old Unix crypt"
 DESCRIPTION="mcrypt is an enhanced replacement for the old crypt command. It \
 allows users to encrypt files or data streams without having to be \
 cryptographers. mcrypt uses libmcrypt, its companion library, which contains \
--- a/app-crypt/mcrypt/mcrypt-2.6.8.recipe
+++ b/app-crypt/mcrypt/mcrypt-2.6.8.recipe
@@ -1,0 +1,51 @@
+SUMMARY="Replacements for the old Unix crypt"
+DESCRIPTION="mcrypt is an enhanced replacement for the old crypt \
+command. mcrypt allows developers to use a wide range of encryption \
+functions without making drastic changes to their code. It contains the \
+actual encryption functions and provides a standardized mechanism for \
+accessing them."
+HOMEPAGE="http://mcrypt.sourceforge.net/"
+COPYRIGHT="1998-2007 Nikos Mavroyanopoulos"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/mcrypt/mcrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="5145aa844e54cca89ddab6fb7dd9e5952811d8d787c4f4bf27eb261e6c182098"
+SOURCE_DIR="mcrypt-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	mcrypt = $portVersion
+	cmd:mcrypt = $portVersion
+	cmd:mdecrypt = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libmcrypt
+	lib:libmhash
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libmcrypt
+	devel:libmhash
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	"
+
+BUILD()
+{
+	autoreconf -vfi
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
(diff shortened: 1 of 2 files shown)


app-crypt/mcrypt/mcrypt-2.6.8.recipe
```diff
@@ -0,0 +1,51 @@
...
+accessing them."
+HOMEPAGE="http://mcrypt.sourceforge.net/"
+COPYRIGHT="1998-2007 Nikos Mavroyanopoulos"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/mcrypt/mcrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="5145aa844e54cca89ddab6fb7dd9e5952811d8d787c4f4bf27eb261e6c182098"
+SOURCE_DIR="mcrypt-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
Begasus: Just being picky, x86 shouldn't be in here or commented as ?x86 as it builds on primary architectures x86_gcc2 and x86_64, other then that builds ok in x86_gcc2, nice work :+1:

app-crypt/mcrypt/mcrypt-2.6.8.recipe
```diff
@@ -0,0 +1,51 @@
+SUMMARY="Replacements for the old Unix crypt"
+DESCRIPTION="mcrypt is an enhanced replacement for the old crypt \
+command. mcrypt allows developers to use a wide range of encryption \
+functions without making drastic changes to their code. It contains the \
+actual encryption functions and provides a standardized mechanism for \
```
fbrosson: This variant is also based on the description found in http://mcrypt.sourceforge.net/.
```diff
-DESCRIPTION="mcrypt is an enhanced replacement for the old crypt \
-command. mcrypt allows developers to use a wide range of encryption \
-functions without making drastic changes to their code. It contains the \
-actual encryption functions and provides a standardized mechanism for \
+DESCRIPTION="mcrypt is an enhanced replacement for the old crypt command. It \
+allows users to encrypt files or data streams without having to be \
+cryptographers. mcrypt uses libmcrypt, its companion library, which contains \
+the actual encryption functions and provides a standardized mechanism for \
 accessing them."

File: app-crypt/mcrypt/mcrypt-2.6.8.recipe [hunk 3]
```diff
@@ -0,0 +1,51 @@
...
+accessing them."
+HOMEPAGE="http://mcrypt.sourceforge.net/"
+COPYRIGHT="1998-2007 Nikos Mavroyanopoulos"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/mcrypt/mcrypt-$portVersion.tar.gz"
+CHECKSUM_SHA256="5145aa844e54cca89ddab6fb7dd9e5952811d8d787c4f4bf27eb261e6c182098"
+SOURCE_DIR="mcrypt-$portVersion"
```
fbrosson:
```diff
-SOURCE_DIR="mcrypt-$portVersion"
```

app-crypt/mcrypt/mcrypt-2.6.8.recipe
fbrosson: Yes, you are right, because we did not test it on x86.
But since the support for x86 primary arch has been dropped, I think we can change the meaning of the `?` for x86 primary arch:
* for recipes **without* support for x86 secondary arch, marking x86 primary arch as tested should now be understood, imo, as _"this was **not tested** on x86 but everything shows that it **would work** without any glitches."_
* for recipes **with** support for x86 secondary arch, marking x86 as tested should be understood as _"it **was tested** on x86 secondary arch and would therefore **also work** on x86 primary arch."_
* we would only mark x86 as untested for recipes which we think they might work but cannot assume it would.
* we would not mention x86 at all on recipes which do not support x86 secondary arch and for which we have no idea about the outcome on that arch.

I think this change would be quite fair.
Of course, this could be adopted for new recipes only. I'm not suggesting that we change the status of existing recipes.

nysnatuss: @fbrosson So should I always mark "x86" as untested or left it out from primary arch in all future recipes?

fbrosson: If @Begasus agrees, I'd leave that as tested, otherwise you can either mark it as untested or remove it.
But I think it would be sad to not mention it.
And if we mention it, it would be sad to pretend we don't know if it works or not, because we all know it would perfectly work.

That said, some contributors could argue that x86 primary arch will disappear, so mentioning x86 in `ARCHITECTURES` is just noise, and they would be perfectly right.

So this is up to you. I'm not using x86 primary arch so any choice will be OK.

app-crypt/mcrypt/mcrypt-2.6.8.recipe
```diff
@@ -1,3 +1,4 @@
+SUMMARY="Replacements for the old Unix crypt"
```
fbrosson: What do you think of this variant?
```diff
-SUMMARY="Replacements for the old Unix crypt"
+SUMMARY="A replacement for the old Unix crypt command"
```

nysnatuss: Much better! 👍

app-crypt/mcrypt/mcrypt-2.6.8.recipe
Begasus: Fair enough for me :)

PR #2003, opened 2017-12-30
Partio: new recipe

extrowerk:


media-libs/partio/partio-1.1.0.recipe
```diff
@@ -0,0 +1,95 @@
...
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+PROVIDES="
+	partio$secondaryArchSuffix = $portVersion compat >= 1
+	cmd:partattr = $portVersion compat >= 1
+	cmd:partconv = $portVersion compat >= 1
+	cmd:partinfo = $portVersion compat >= 1
+	cmd:partview = $portVersion compat >= 1
```
alaviss: This part doesn't contain any library, so I believe we don't need `compat` here

media-libs/partio/partio-1.1.0.recipe
```diff
@@ -0,0 +1,95 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libGLU$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	partio${secondaryArchSuffix}_devel = $portVersion compat >= 1
+	devel:libpartio$secondaryArchSuffix = $portVersion compat >= 1
```
alaviss: Given the absent of `lib:` in `PROVIDES`, I believe this package generate an static library. As those libraries are added directly into a program, there should be no need for a `compat` version

media-libs/partio/partio-1.1.0.recipe
```diff
@@ -0,0 +1,95 @@
...
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:swig
+	"
+
+defineDebugInfoPackage partio$secondaryArchSuffix \
+	$binDir/partattr \
+	$binDir/partconv \
+	$binDir/partinfo \
+	$binDir/partview
```
alaviss: I see that you don't specify `$secondaryArchSuffix` for the `cmd:`s above. If your port installs directly into `$prefix/bin`, please change these into `$prefix/bin` or you will meet some packaging error with `SECONDARY_ARCHITECTURES`

media-libs/partio/partio-1.1.0.recipe
```diff
@@ -0,0 +1,107 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python
+	cmd:swig
+	"
+
+defineDebugInfoPackage partio$secondaryArchSuffix \
+	$prefix/bin/partattr$commandSuffix \
+	$prefix/bin/partconv$commandSuffix \
+	$prefix/bin/partinfo$commandSuffix \
+	$prefix/bin/partview$commandSuffix
```
alaviss:
```diff
-	$prefix/bin/partattr$commandSuffix \
-	$prefix/bin/partconv$commandSuffix \
-	$prefix/bin/partinfo$commandSuffix \
-	$prefix/bin/partview$commandSuffix
+	$commandBinDir/partattr$commandSuffix \
+	$commandBinDir/partconv$commandSuffix \
+	$commandBinDir/partinfo$commandSuffix \
+	$commandBinDir/partview$commandSuffix
```

media-libs/partio/partio-1.1.0.recipe
```diff
@@ -0,0 +1,107 @@
...
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	mkdir -p $includeDir
+	mv $prefix/include/* $includeDir/
+	rm -rf $prefix/include $prefix/test
```
alaviss: I believe this would be the correct solution https://github.com/wdas/partio/pull/68

PR #2006, opened 2017-12-30
GDAL : eliminate libtool files

extrowerk:


sci-libs/gdal/gdal-2.2.3.recipe
```diff
@@ -132,6 +132,9 @@ INSTALL()
 {
 	make install
 
+	# celanup
+	rm -rf $libDir/libgdal.la
```
fbrosson:
```diff
-	rm -rf $libDir/libgdal.la
+	rm -f $libDir/libgdal.la
```
(no need for recursion here)

PR #2021, opened 2017-12-31
gpart: new recipe, using the latest stable release

sfanxiang:
Most functionalities are working fine, including scanning and partition guessing.

Disk geometry reporting is not correct for heads/cylinders/sectors, which is likely due to the hard-coding of 1 for head/cylinders in the kernel:
http://cgit.haiku-os.org/haiku/tree/src/system/kernel/device_manager/devfs.cpp#n2184


sfanxiang:
The reason why the patchset is so large is that it contains some previous work (mostly a clang-format) already found on the upstream repo. It's included so that we have the proper fixes to known bugs. New patches are in the last commit of the patchset.

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,54 @@
...
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	"
+
+BUILD_PREREQUIRES="
+	haiku_devel
```
alaviss: Please move `haiku_devel` to `BUILD_REQUIRES`

sys-block/gpart/patches/gpart-0.3.patchset
```diff
@@ -0,0 +1,5672 @@
...
++s64_t skipsec = 0, maxsec = 0;
++FILE *logfile = 0;
+ 
+ void usage()
+ {
+-	FILE		*fp = stderr;
+-
+-	fprintf(fp,"Usage: %s [options] device\n",PACKAGE_NAME);
+-	fprintf(fp,"Options: [-b <backup MBR>][-C c,h,s][-c][-d][-E][-e][-f][-g][-h][-i]\n");
+-	fprintf(
```
alaviss: I believe this is the only patch that matters, and others were simply upstream patch. If they're not required, please remove them.

alaviss: There shouldn't be a need for `#else` here. These code paths share the same `return (1);` statement, so you would only need an `#endif` for this.

alaviss: Same as above

alaviss: We already have [these](http://xref.plausible.coop/source/xref/haiku/headers/compatibility/bsd/endian.h#33). Make sure the header include `<endian.h>` then add `-D_BSD_SOURCE` to `CFLAGS` to enable them

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,54 @@
+SUMMARY="A small tool which tries to guess what partitions are on a hard disk"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are \
+on a PC type, MBR-partitioned hard disk in case the primary partition table \
+was damaged."
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999-2001 Michail Brzitwa"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/baruch/gpart/archive/0.3.zip"
```
alaviss: Tarballs are generally recommended, also they're smaller. Please consider using `$portVersion` where applicable

```diff
-SOURCE_URI="https://github.com/baruch/gpart/archive/0.3.zip"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$portVersion.tar.gz"
```

sys-block/gpart/patches/gpart-0.3.patchset
alaviss: We have these in [`<endian.h>`](http://xref.plausible.coop/source/xref/haiku/headers/compatibility/bsd/endian.h#33). Define `_BSD_SOURCE` to use them

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,54 @@
...
+CHECKSUM_SHA256="6e2e1727af22615853d78958745230dd3a554b8e87c3212892ca7bd7117b1635"
+PATCHES="gpart-$portVersion.patchset"
+
+ARCHITECTURES="x86 x86_64"
```
fbrosson: x86 is not supported anymore, so you should mark it as untested (with a "?x86").
```diff
-ARCHITECTURES="x86 x86_64"
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
```
In case it does not build on x86_gcc2 primary arch, it is possible to add x86 as a secondary arch.

sys-block/gpart/patches/gpart-0.3.patchset.2
```diff
@@ -0,0 +1,170 @@
...
+ static int on_cyl_boundary(disk_desc *d, s64_t sec)
+ {
++#ifndef __HAIKU__
+ 	struct disk_geom *g = &d->d_dg;
+ 
+ 	if (g->d_h && g->d_s)
+ 		return ((sec % (g->d_h * g->d_s)) == 0);
+ 	return (1);
++#else
++	return (1);
```
alaviss: These code path share the same final `return (1);` statement. You could avoid `#else` code path by moving `#endif` to before `return (1);`. Still, this is only an suggestion, and your code is still vaild without this change

```diff
 
 	if (g->d_h && g->d_s)
 		return ((sec % (g->d_h * g->d_s)) == 0);
+#endif
	return (1);
-#else
-	return (1);
-#endif
 }
 
 static int on_head_boundary(disk_desc *d, s64_t sec)

File: sys-block/gpart/patches/gpart-0.3.patchset.2 [hunk 6]
```diff
@@ -0,0 +1,170 @@
...
+ static int on_head_boundary(disk_desc *d, s64_t sec)
+ {
++#ifndef __HAIKU__
+ 	struct disk_geom *g = &d->d_dg;
+ 
+ 	if (g->d_s)
+ 		return ((sec % g->d_s) == 0);
+ 	return (1);
++#else
++	return (1);
```
alaviss:
```diff
 
 	if (g->d_s)
 		return ((sec % g->d_s) == 0);
+#endif
 	return (1);
-#else
-	return (1);
-#endif
 }
 
 static void print_partition(disk_desc *d, dos_part_entry *p, int inset, s64_t offset)

File: sys-block/gpart/patches/gpart-0.3.patchset.2 (hunk 5 above)
Comment (sfanxiang) [reply sys-block/gpart/patches/gpart-0.3.patchset.2]: Thanks. I think repeating `return (1);` makes the code less error-prone if gpart were to actually support haiku.

File: sys-block/gpart/patches/gpart-0.3.patchset.2 [hunk 7]
```diff
@@ -105,51 +120,24 @@ index 3b44a75..71aef4a 100644
...
-+#		define htobe64(x) (x)
-+#		define le16toh(x) __builtin_bswap16(x)
-+#		define be16toh(x) (x)
-+#		define le32toh(x) __builtin_bswap32(x)
-+#		define be32toh(x) (x)
-+#		define le64toh(x) __builtin_bswap64(x)
-+#		define be64toh(x) (x)
-+#	endif
++#define _BSD_SOURCE
++#include <endian.h>
```
alaviss: Please don't do it like this. Instead add "-D_BSD_SOURCE" to your CFLAGS, and only include `<endian.h>` here. `<endian.h>` is present in both BSDs and Linux, so you may not need to put it in `#ifdef __HAIKU__`

alaviss: Actually, looking at btrfs code above, you may not even need to add `<endian.h>`, just add `-D_BSD_SOURCE` to CFLAGS and you're done

sfanxiang: `<endian.h>` is used elsewhere without being included, still leading to undefined references.
On linux, this is not an error probably because it's included by another header.

sfanxiang:
Ticket filed: https://dev.haiku-os.org/ticket/13918
I'm not sure if we should upstream this yet, with potentially some kernel fixes ahead.

sfanxiang:
BTW, the gpart task seems to have timed out for me :<

alaviss:
Try to see if you could reclaim it

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,58 @@
...
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are \
+on a PC type, MBR-partitioned hard disk in case the primary partition table \
+was damaged."
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999-2001 Michail Brzitwa"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="ec56d12ec9ffdb9877c12692ea6e51620b1ae44473d3d253b27fc31ed9ebb4dd"
+
```
fbrosson: You can remove the blank line before PATCHES.

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,58 @@
+SUMMARY="A small tool which tries to guess what partitions are on a hard disk"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are \
+on a PC type, MBR-partitioned hard disk in case the primary partition table \
+was damaged."
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999-2001 Michail Brzitwa"
```
fbrosson: There were commits from the author until 2017, so maybe it would be better to update this.
```diff
-COPYRIGHT="1999-2001 Michail Brzitwa"
+COPYRIGHT="1999-2017 Michail Brzitwa"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,58 @@
...
+REVISION="1"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="ec56d12ec9ffdb9877c12692ea6e51620b1ae44473d3d253b27fc31ed9ebb4dd"
+
+PATCHES="
+	gpart-$portVersion.patchset.1
+	gpart-$portVersion.patchset.2
+	"
+
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 x86_64"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,58 @@
...
+	cmd:ld
+	cmd:make
+	cmd:makeinfo
+	cmd:pkg_config
+	cmd:sed
+	"
+
+BUILD()
+{
+	chmod +x autogen.sh
```
fbrosson:
```diff
-	chmod +x autogen.sh
```
(not needed)

sfanxiang:
@fbrosson After rechecking this PR I found that I actually reopened d->d_dev in os_disk_geometry(), and I'm not sure if it's related to https://dev.haiku-os.org/ticket/13918. The code needs to be changed, anyway.

fbrosson:
Thanks for the update.
For the changes you mentioned in the patch, you don't need bother with it at the moment because there is some activity on the other PR and I don't know if it will get merged or not.

fbrosson:
(I'm marking this PR with the "wip-do-not-merge" label until the developers currently reviewing the other PR decide whether or not that other PR can be merged.)

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,56 @@
...
+	cmd:ld
+	cmd:make
+	cmd:makeinfo
+	cmd:pkg_config
+	cmd:sed
+	"
+
+BUILD()
+{
+	NOCONFIGURE=1 ./autogen.sh
```
alaviss: I do wish `NOCONFIGURE` was a thing [there](https://github.com/baruch/gpart/blob/23b651d1e9c6ae1454e5d3fe315d7d4e145661c0/autogen.sh#L7). I think your best bet right now is `autoreconf -vfi`

sys-block/gpart/patches/gpart-0.3.patchset.2
```diff
@@ -0,0 +1,139 @@
...
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++	device_geometry geom;
++	int fd;
++
++	if ((fd = open(d->d_dev, O_RDONLY)) < 0)
++		pr(FATAL, EM_OPENFAIL, d->d_dev, strerror(errno));
++
++	if (ioctl(fd, B_GET_GEOMETRY, &geom, sizeof(geom)) < 0) {
```
alaviss: POSIX [`ioctl()`](http://pubs.opengroup.org/onlinepubs/009695399/functions/ioctl.html#tag_03_282_04) error code is -1, and all other value is considered success.

```diff
-+	if (ioctl(fd, B_GET_GEOMETRY, &geom, sizeof(geom)) < 0) {
++	if (ioctl(fd, B_GET_GEOMETRY, &geom, sizeof(geom)) == -1) {

File: sys-block/gpart/patches/gpart-0.3.patchset.2 [hunk 14]
```diff
@@ -0,0 +1,139 @@
...
+ 	g.d_nsecs = o / u;
+ 	g.d_c = g.d_nsecs / g.d_h / g.d_s;
+ }
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++	device_geometry geom;
++	int fd;
++
++	if ((fd = open(d->d_dev, O_RDONLY)) < 0)
```
alaviss: Looks like an `fd` is already available at `d->d_fd`, so you should avoid having more side effects

sys-block/gpart/patches/gpart-0.3.patchset.2
```diff
@@ -0,0 +1,132 @@
...
+ 	g.d_c = g.d_nsecs / g.d_h / g.d_s;
+ }
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++	device_geometry geom;
++
++	if (ioctl(d->d_fd, B_GET_GEOMETRY, &geom, sizeof(geom)) == -1) {
++		pr(FATAL, EM_IOCTLFAILED, "B_GET_GEOMETRY", strerror(errno));
++	}
```
alaviss: [`pr()`][0] won't abort the execution flow. You should put the below code to an `else` branch instead

[0]: https://github.com/baruch/gpart/blob/deffb405040d6f70682657bcf930f92e2a9a7458/src/gpart.c#L106

sfanxiang: @alaviss For FATAL level it will: https://github.com/baruch/gpart/blob/deffb405040d6f70682657bcf930f92e2a9a7458/src/gpart.c#L124

The whole project is based on this logic.

fbrosson:
@alaviss 
> In the meantime, @fbrosson, can you try applying patchset 2 and see if things are better? If they are, we could do a hybrid between the two PRs :)

Yep, I've been thinking about doing that but have been postponing it since a long time...
I'll try it soon.

fbrosson:
Good news: as expected, `gpart-0.3.patchset.2` applies perfectly to gpart's `HEAD`.
I wrote the hybrid recipe, and it builds fine.
However, I get the same results as with the other PR.

I'll push an extra commit to this PR. The result will be that the two concurrent recipes will be quite similar. Most differences will be in the patchsets.

I can later "eat" my commit if needed, i.e. make it vanish, so that this PR will be exactly as it is right now.

fbrosson:
@alaviss: Here it is: our "hybrid" recipe :)
BTW, it was really easy to write the hybrid recipe because @sfanxiang had very nicely created 2 perfect patchsets.

@sfanxiang: You and @alaviss know the `ioctl` calls much better than me, so I'll let you discuss what to do from here.

Regarding the 2 commits I have added, as explained in my previous comment, I can "eat" them very easily (as well as you or any other contributor with write access to the repo), so if needed I can do so on request :)

sfanxiang:
> Good news: as expected, gpart-0.3.patchset.2 applies perfectly to gpart's HEAD.

Patchset 2 was originally written on HEAD. Only I thought a pointer to a random commit wouldn't be accepted. 😂

> However, I get the same results as with the other PR.

I'll look into this, but it used to work for me (at least on the installation disk).

sfanxiang:
@fbrosson Replying to your comment in #2059.

> So, as we can see, gpart is not reading the partition table.

The purpose of gpart is to __guess__ what partitions you have, in case you have a damaged partition table. On its homepage: "Gpart is a small tool which tries to __guess__ what partitions are on a PC type [...]". In fact, it reads the partition table if you turn on `-v`.

> Remark: the fact that I did not format any of the partitions does not matter.

Formation is the only clue of a partition's existence, without the partition table.

I would suggest you to format the partition, turn on `-v` (it mysteriously makes the scanning faster on my setup - I'm not sure why), and try again.

XF

sfanxiang:
@fbrosson 
> *** Fatal error: dev(/dev/disk/ata/0/slave/raw): read error near sector(524152): General system error.

How big, in bytes, is `/dev/disk/ata/0/slave/raw`? Because 524152 * 512 = 268365824, which is nearly 256 MiB. Could that be an EOF?

> BTW, if I pass -v and -d I get rid of the General system error:

`-d` skips the scanning, hence no error.

> Anyway, I guess we need to make a few extra changes to get rid of the General system error.
> I'll leave that to you :)
> (But don't worry about time. This can wait until you have enough free time.)

Sure. In the mean time, could you please test on a regular file instead of a block device? Just copy over the image to a regular file and run `gpart` on it.

Thanks.

XF

fbrosson:
I've created the virtual disk on my host machine with:
```
dd if=/dev/zero bs=1M count=256 of=/ram/256mb
```
(`/ram/` is itself a ramdisk, btw, but much bigger ;)
`/ram/256mb` is exactly 256 MiB:
```
stat -c %s /ram/256mb 
268435456
factor 268435456
268435456: 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2 2
```
And then I passed it to my Haiku VM. (I use QEMU+Xen+QXL+spicec)
```
grep 256mb haiku32.cfg 
disk = [ 'phy:/dev/mmcblk0p5,xvda,w', 'file:/ram/256mb,xvdb,w' ]
```
So it looks like gpart does not know how to probe the device's size.

I just did the test you suggested, with a real file instead of a device, and it works like a charm :)
```
~> gpart /ramwork/256mb 

Begin scan...
Possible partition(Linux ext2), size(128mb), offset(0mb)
Possible partition(Linux ext2), size(64mb), offset(128mb)
Possible partition(Linux ext2), size(31mb), offset(192mb)
Possible partition(Linux ext2), size(31mb), offset(224mb)
End scan.

Checking partitions...
Partition(Linux ext2 filesystem): primary 
Partition(Linux ext2 filesystem): primary 
Partition(Linux ext2 filesystem): primary 
Partition(Linux ext2 filesystem): primary 
Ok.

Guessed primary partition table:
Primary partition(1)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 128mb #s(262144) s(64-262207)
   chs:  (0/32/1)-(809/45/2)d (0/32/1)-(809/45/2)r

Primary partition(2)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 64mb #s(131072) s(262208-393279)
   chs:  (809/46/1)-(1023/161/2)d (809/46/1)-(1213/133/2)r

Primary partition(3)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 31mb #s(63488) s(393280-456767)
   chs:  (1023/161/2)-(1023/161/2)d (1213/134/1)-(1409/125/2)r

Primary partition(4)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 31mb #s(63488) s(458816-522303)
   chs:  (1023/161/2)-(1023/161/2)d (1416/16/1)-(1612/7/2)r

```

scottmc:
Nice work so far.  But can it detect and correctly ID Haiku (BeFS) partitions?  If so great.  If not, can add this functionality?  And better yet, upstream that fix so that it will detect Haiku partitions when on other OSes?

sfanxiang:
>  But can it detect and correctly ID Haiku (BeFS) partitions?

Yes. https://github.com/baruch/gpart/blob/master/README.md

Begasus:
Merged in: https://github.com/haikuports/haikuports/pull/9360

PR #2024, opened 2017-12-31
FFMPEG : bump version

extrowerk:
Changes:
- reorder
- debuginfo packages defined, but as the libs stripped it have not much use.


pulkomandy:
On Sun, Dec 31, 2017 at 08:20:08PM +0000, miqlas wrote:
> - debuginfo packages defined, but as the libs stripped it have not much use.

I think it's just a --enable-debug to pass to configure or something
similar? Should not be too hard.

-- 
Adrien.

korli:
Can you remove the patchset for 3.4.0 too?

extrowerk:
@korli: done.

PR #2026, opened 2018-01-01
dev-util/pkgconf: new recipe

alaviss:
Done as part of GCI 2017

| package | arch | status |
| --- | --- | --- |
| dev-libs/atf | x86_gcc2 | `Test cases: 319 total, 7 skipped, 0 expected failures, 0 broken, 0 failed` |
| dev-libs/atf | x86_64 | `Test cases: 319 total, 7 skipped, 0 expected failures, 0 broken, 0 failed` |
| dev-lua/lutok | x86_gcc2 | `Test cases: 101 total, 4 skipped, 0 expected failures, 0 broken, 0 failed` |
| dev-lua/lutok | x86_64 | `Test cases: 101 total, 4 skipped, 0 expected failures, 0 broken, 0 failed` |
| dev-util/pkgconf | x86_gcc2 | `Test cases: 109 total, 0 skipped, 0 expected failures, 0 broken, 0 failed` |
| dev-util/pkgconf | x86_64 | `Test cases: 109 total, 0 skipped, 0 expected failures, 0 broken, 0 failed` |
| dev-util/kyua | x86_gcc2 | `Test cases: 1215 total, 41 skipped, 4 expected failures,`[`3 broken, 3 failed`][1] |
| dev-util/kyua | x86_64 | `Test cases: 1215 total, 41 skipped, 4 expected failures,`[`3 broken, 3 failed`][2] |

Changes:
- dev-libs/atf:
  - don't include `<sys/mount.h>` on Haiku
  - disable tests that requires atf to be installed
- dev-util/kyua:
  - use `struct dirent` correctly. This prevents segfault on Haiku (thanks @pulkomandy for help)
  - use Haiku user configuration directory (this broke a lot of tests)
  - disable all tests that depends on `$HOME`
  - disable tests that use OS-specific behaviour
  - use `WIFCORED(x)` on Haiku
  - fix tests checking for `SIGKILL` on Haiku
  - implemented memory detection code for Haiku
  - and some misc fixes
- dev-util/pkgconf: swap `LIBRARY_PATH` with `BELIBRARIES` as this appears to be what our gcc [uses][0]

**NOTE:** Some changes to kyua are upstreamable but I won't do so as I don't want to sign their CLA. Feel free to upstream those changes if you wish

Known issues:
- dev-libs/atf: cyclic dependency on kyua due to the lack of `TEST_PREREQUIRES`
- dev-lua/lutok: cyclic dependency on kyua due to the lack of `TEST_PREREQUIRES`
- dev-util/kyua: cyclic dependency on `dev-libs/atf` and `dev-lua/lutok`

Meta issues:
Quoting self on GCI
> I'm having an situation here. ATF-based ports have their tests designed to be run after installation. This means that some of them won't execute correctly (or at all) when not installed. Currently I'm removing all of them from our install image, should I instead move them into an another package? And if we decide to do so, where should we put them? Currently they're in $prefix/tests.

[0]: https://github.com/haikuports/haikuports/blob/99b3c76c3e8178c62044b9f0e0594db0b1ff518c/sys-devel/gcc/patches/gcc-5.4.0_2016_06_04.patchset#L641
[1]: https://0bin.net/paste/vM-brFsQ-t2rWEzg#+2mWGpB8EmqSeobYlnAEjzq8mnJRTL7MO2ZBjWFvNry
[2]: https://0bin.net/paste/vOeuxcCguqWSSd3b#kaKOWBz3IJ9F16y42nuk8ZPCnnGVFRqS2XwngFkehl4

### DIFF
--- a/dev-libs/atf/atf-0.21.recipe
+++ b/dev-libs/atf/atf-0.21.recipe
@@ -1,0 +1,112 @@
+SUMMARY="Libraries to write tests in C, C++ and shell"
+DESCRIPTION="ATF, or Automated Testing Framework, is a collection of libraries \
+to write test programs in C, C++ and POSIX shell.
+
+The ATF libraries offer a simple API. The API is orthogonal through \
+the various bindings, allowing developers to quickly learn how to write \
+test programs in different languages.
+
+ATF-based test programs offer a consistent end-user command-line interface to \
+allow both humans and automation to run the tests.
+
+ATF-based test programs rely on an execution engine to be run and \
+this execution engine is not shipped with ATF. Kyua is the engine of choice."
+HOMEPAGE="https://github.com/jmmv/atf"
+COPYRIGHT="2007-2012 The NetBSD Foundation, Inc.
+	2011, 2012, 2014 Google Inc."
+LICENSE="BSD (2-clause)
+	BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/jmmv/atf/releases/download/atf-$portVersion/atf-$portVersion.tar.gz"
+CHECKSUM_SHA256="92bc64180135eea8fe84c91c9f894e678767764f6dbc8482021d4dde09857505"
+PATCHES="atf-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix="$secondaryArchSuffix"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=""
+fi
+
+libVersion="1.0.0"
+libCppVersion="2.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+libCppVersionCompat="$libCppVersion compat >= ${libCppVersion%%.*}"
+
+PROVIDES="
+	atf$secondaryArchSuffix = $portVersion
+	cmd:atf_sh$secondaryArchSuffix = $portVersion
+	lib:libatf_c$secondaryArchSuffix = $libVersionCompat
+	lib:libatf_c++$secondaryArchSuffix = $libCppVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	atf${secondaryArchSuffix}_devel = $portVersion
+	devel:libatf_c$secondaryArchSuffix = $libVersionCompat
+	devel:libatf_c++$secondaryArchSuffix = $libCppVersionCompat
+	"
+REQUIRES_devel="
+	atf$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:awk
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:grep
+	cmd:make
+	cmd:sed
+	"
+
+TEST_REQUIRES="
+	cmd:kyua$commandSuffix
+	"
+
+defineDebugInfoPackage atf$secondaryArchSuffix \
+	"$binDir/atf-sh" \
+	"$libDir/libatf-c.so.$libVersion" \
+	"$libDir/libatf-c++.so.$libCppVersion"
+
+BUILD()
+{
+	CPPFLAGS="-D_BSD_SOURCE" LIBS="-Wl,--as-needed -lbsd" \
+	runConfigure --omit-dirs "libExecDir" ./configure \
+		--libexecdir="$libExecDir/atf" \
+		--enable-developer=no
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm "$docDir"/{COPYING,README}
+
+	rm -rf "$prefix/tests"
+	rm -f "$libDir"/*.la
+
+	prepareInstalledDevelLibs libatf-c++ libatf-c
+	fixPkgconfig
+	# Turns out their .pc files don't have ${prefix}
+	sed -e "s|\${prefix}|$prefix|g" -i "$developLibDir/pkgconfig"/*.pc
+	sed \
+		-e "s|\${exec_prefix}/bin/atf-sh|$binDir/atf-sh|" \
+		-i "$developLibDir/pkgconfig/atf-sh.pc"
+	packageEntries devel \
+		"$developDir" \
+		"$dataDir/aclocal" \
+		"$manDir/man3"/atf-c*
+}
+
+TEST()
+{
+	CPATH="$PWD" kyua test
+}
--- a/dev-lua/lutok/lutok-0.4.recipe
+++ b/dev-lua/lutok/lutok-0.4.recipe
@@ -1,0 +1,103 @@
+SUMMARY="A lightweight C++ API library for Lua"
+DESCRIPTION="Lutok provides thin C++ wrappers around the Lua C API to ease the \
+interaction between C++ and Lua. These wrappers make intensive use of RAII to \
+prevent resource leakage, expose C++-friendly data types, report errors by means \
+of exceptions and ensure that the Lua stack is always left untouched in the face \
+of errors. The library also provides a small subset of miscellaneous utility \
+functions built on top of the wrappers.
+
+Lutok focuses on providing a clean and safe C++ interface; the drawback is that \
+it is not suitable for performance-critical environments. In order to implement \
+error-safe C++ wrappers on top of a Lua C binary library, Lutok adds several \
+layers of abstraction and error checking that go against the original spirit of \
+the Lua C API and thus degrade performance."
+HOMEPAGE="https://github.com/jmmv/lutok"
+COPYRIGHT="2011, 2012, 2014 Google Inc"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/jmmv/lutok/releases/download/lutok-$portVersion/lutok-$portVersion.tar.gz"
+CHECKSUM_SHA256="2cec51efa0c8d65ace8b21eaa08384b77abc5087b46e785f78de1c21fb754cd5"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix="$secondaryArchSuffix"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=""
+fi
+
+libVersion="3.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	lutok$secondaryArchSuffix = $portVersion
+	lib:liblutok$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:liblua$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	lutok${secondaryArchSuffix}_devel = $portVersion
+	devel:liblutok$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	lutok$secondaryArchSuffix == $portVersion base
+	devel:liblua$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:liblua$secondaryArchSuffix
+	devel:libatf_c++$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:atf_sh$secondaryArchSuffix
+	cmd:awk
+	cmd:diff
+	cmd:doxygen
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+TEST_REQUIRES="
+	cmd:kyua$commandSuffix
+	"
+
+defineDebugInfoPackage lutok$secondaryArchSuffix \
+	"$libDir/liblutok.so.$libVersion"
+
+BUILD()
+{
+	runConfigure --omit-dirs docDir ./configure \
+		--docdir="$developDocDir" \
+		--enable-developer=no
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	mv "$developDocDir/AUTHORS" "$docDir"
+	rm "$developDocDir"/{README,COPYING}
+
+	rm -f "$libDir"/*.la
+	rm -rf "$prefix"/tests
+
+	prepareInstalledDevelLib liblutok
+	fixPkgconfig
+	# Not all .pc file use ${prefix}
+	sed -e "s|\${prefix}|$prefix|g" -i "$developLibDir/pkgconfig/lutok.pc"
+	packageEntries devel "$developDir"
+}
+
+TEST()
+{
+	make check
+}
--- a/dev-util/kyua/kyua-0.13.recipe
+++ b/dev-util/kyua/kyua-0.13.recipe
@@ -1,0 +1,97 @@
+SUMMARY="A testing framework for infrastructure software"
+DESCRIPTION="Kyua is a testing framework for infrastructure software, \
+originally designed to equip BSD-based operating systems with a test suite. \
+This means that Kyua is lightweight and simple, and that Kyua integrates well \
+with various build systems and continuous integration frameworks.
+
+Kyua features an expressive test suite definition language, a safe runtime \
+engine for test suites and a powerful report generation engine.
+
+Kyua is for both developers and users, from the developer applying a simple \
+fix to a library to the system administrator deploying a new release on a \
+production machine.
+
+Kyua is able to execute test programs written with a plethora of testing \
+libraries and languages. The library of choice is ATF, for which Kyua was \
+originally designed, but simple, framework-less test programs and \
+TAP-compliant test programs can also be executed through Kyua."
+HOMEPAGE="https://github.com/jmmv/kyua"
+COPYRIGHT="2010-2015 The Kyua Authors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/jmmv/kyua/releases/download/kyua-$portVersion/kyua-$portVersion.tar.gz"
+CHECKSUM_SHA256="db6e5d341d5cf7e49e50aa361243e19087a00ba33742b0855d2685c0b8e721d6"
+PATCHES="kyua-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	kyua$secondaryArchSuffix = $portVersion
+	cmd:kyua$commandSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:liblua$secondaryArchSuffix
+	lib:liblutok$secondaryArchSuffix
+	lib:libsqlite3$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libatf_c++$secondaryArchSuffix
+	devel:liblutok$secondaryArchSuffix
+	devel:libsqlite3$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoreconf
+	cmd:awk
+	cmd:doxygen
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+TEST_REQUIRES="
+	cmd:find
+	cmd:sqlite3
+	"
+
+USER_SETTINGS_FILES="
+	settings/kyua directory
+	settings/kyua/kyua.conf template $dataDir/kyua/examples/kyua.conf"
+
+defineDebugInfoPackage kyua$secondaryArchSuffix \
+	"$commandBinDir/kyua"
+
+BUILD()
+{
+	autoreconf -vfi
+	LIBS="-Wl,--as-needed -lbe" \
+	runConfigure --omit-dirs "binDir settingsDir" ./configure \
+		--bindir=$commandBinDir --enable-developer=no
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -rf "$prefix/tests"
+	rm "$docDir"/{LICENSE,CONTRIBUTING.md}
+}
+
+TEST()
+{
+	make check
+}
--- a/dev-util/kyua/patches/kyua-0.13.patchset
+++ b/dev-util/kyua/patches/kyua-0.13.patchset
@@ -1,0 +1,671 @@
+From 235280aa551819a75c09da858a6451e9260488f7 Mon Sep 17 00:00:00 2001
+From: Leorize <alaviss@users.noreply.github.com>
+Date: Mon, 1 Jan 2018 22:35:07 +0700
+Subject: utils/process/child_test: fix type mismatch
+
+Align pipe_fail with syscall_pipe definition
+
+diff --git a/utils/process/child_test.cpp b/utils/process/child_test.cpp
+index 8996c31..2641716 100644
+--- a/utils/process/child_test.cpp
++++ b/utils/process/child_test.cpp
+@@ -256,7 +256,7 @@ open_fail(const char* path, const int flags, ...) throw()
+ ///
+ /// \return Always -1.
+ template< int Errno >
+-static pid_t
++static int
+ pipe_fail(int* UTILS_UNUSED_PARAM(fildes)) throw()
+ {
+     errno = Errno;
+-- 
+2.16.2
+
+
+From ea5b2176fc2fb7b89a14d39469fc33814506e84f Mon Sep 17 00:00:00 2001
+From: Leorize <alaviss@users.noreply.github.com>
+Date: Wed, 3 Jan 2018 20:56:21 +0700
+Subject: kyua: use Haiku user directory
+
+This broke a lot of tests that assume kyua will save it's data in
+$HOME/.kyua
+
+diff --git a/cli/common.cpp b/cli/common.cpp
+index 9204082..b937bdd 100644
+--- a/cli/common.cpp
++++ b/cli/common.cpp
+@@ -121,9 +121,9 @@ namespace {
+ static optional< fs::path >
+ get_historical_db(void)
+ {
+-    optional< fs::path > home = utils::get_home();
+-    if (home) {
+-        const fs::path old_db = home.get() / ".kyua/store.db";
++    optional< fs::path > data_dir = utils::get_home_data();
++    if (data_dir) {
++        const fs::path old_db = data_dir.get() / "store.db";
+         if (fs::exists(old_db)) {
+             if (old_db.is_absolute())
+                 return utils::make_optional(old_db);
+diff --git a/cli/config.cpp b/cli/config.cpp
+index 0049103..d35a533 100644
+--- a/cli/config.cpp
++++ b/cli/config.cpp
+@@ -101,9 +101,9 @@ load_config_file(const cmdline::parsed_cmdline& cmdline)
+     } else if (filename.str() != cli::config_option.default_value())
+         return engine::load_config(filename);
+ 
+-    const optional< fs::path > home = utils::get_home();
+-    if (home) {
+-        const fs::path path = home.get() / ".kyua" / config_basename;
++    const optional< fs::path > conf_home = utils::get_home_config();
++    if (conf_home) {
++        const fs::path path = conf_home.get() / config_basename;
+         try {
+             if (fs::exists(path))
+                 return engine::load_config(path);
+diff --git a/cli/main.cpp b/cli/main.cpp
+index 531c252..67820cd 100644
+--- a/cli/main.cpp
++++ b/cli/main.cpp
+@@ -239,10 +239,10 @@ fs::path
+ cli::detail::default_log_name(void)
+ {
+     // Update doc/troubleshooting.texi if you change this algorithm.
+-    const optional< std::string > home(utils::getenv("HOME"));
+-    if (home) {
+-        return logging::generate_log_name(fs::path(home.get()) / ".kyua" /
+-                                          "logs", cmdline::progname());
++    const optional< fs::path > data_home = utils::get_home_data();
++    if (data_home) {
++        return logging::generate_log_name(data_home.get() / "logs",
++                                          cmdline::progname());
+     } else {
+         const optional< std::string > tmpdir(utils::getenv("TMPDIR"));
+         if (tmpdir) {
+diff --git a/store/layout.cpp b/store/layout.cpp
+index 6bc67a8..b5f39d2 100644
+--- a/store/layout.cpp
++++ b/store/layout.cpp
+@@ -228,13 +228,13 @@ layout::new_db_for_migration(const fs::path& root,
+ fs::path
+ layout::query_store_dir(void)
+ {
+-    const optional< fs::path > home = utils::get_home();
+-    if (home) {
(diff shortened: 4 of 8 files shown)


dev-libs/atf/atf-0.21.recipe
```diff
@@ -0,0 +1,107 @@
...
+defineDebugInfoPackage atf$secondaryArchSuffix \
+	"$prefix/bin/atf-sh" \
+	"$libDir/libatf-c.so.$libVersion" \
+	"$libDir/libatf-c++.so.$libVersionCPlus"
+
+BUILD()
+{
+	CPPFLAGS="-D_BSD_SOURCE" LIBS="-lbsd" \
+		runConfigure --omit-dirs "binDir libExecDir" ./configure \
+			--bindir="$prefix/bin" --libexecdir="$libExecDir/atf"
```
fbrosson: We usually have a slightly different indentation:
```diff
 	CPPFLAGS="-D_BSD_SOURCE" LIBS="-lbsd" \
-		runConfigure --omit-dirs "binDir libExecDir" ./configure \
-			--bindir="$prefix/bin" --libexecdir="$libExecDir/atf"
+	runConfigure --omit-dirs "binDir libExecDir" ./configure \
+		--bindir="$prefix/bin" --libexecdir="$libExecDir/atf"
```

dev-lua/lutok/lutok-0.4.recipe
```diff
@@ -0,0 +1,97 @@
...
+	devel:liblua$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:atf_sh
+	cmd:awk
+	cmd:doxygen
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config
```
fbrosson:
```diff
-	cmd:pkg_config
+	cmd:pkg_config$secondaryArchSuffix
```

dev-lua/lutok/lutok-0.4.recipe
```diff
@@ -0,0 +1,97 @@
...
+	cmd:sed
+	"
+# Avoid cyclic dep
+if [ "$recipeAction" = "TEST" ]; then
+	BUILD_PREREQUIRES+="
+		cmd:kyua
+		"
+fi
```
fbrosson: I'd use this:
```
if [ "$recipeAction" = TEST ]; then
BUILD_PREREQUIRES+="
	cmd:kyua
	"
fi
```
because having this indentation will reduce the size of the diff when TEST_REQUIRES becomes available. If/when this happens, with the suggested indentation the diff will be:
```diff
-if [ "$recipeAction" = TEST ]; then
-BUILD_PREREQUIRES+="
+TEST_REQUIRES="
 	cmd:kyua
 	"
-fi
```

dev-libs/atf/atf-0.21.recipe
```diff
@@ -0,0 +1,107 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:grep
+	cmd:make
+	cmd:sed
+	"
+# Avoid cyclic dep
+if [ "$recipeAction" = "TEST" ]; then
+	BUILD_PREREQUIRES+="
+		cmd:kyua
+		"
```
fbrosson: I'd use this:
```
if [ "$recipeAction" = TEST ]; then
BUILD_PREREQUIRES+="
	cmd:kyua
	"
fi
```
because having this indentation will reduce the size of the diff when TEST_REQUIRES becomes available.
If/when this happens, with the suggested indentation the diff will be:
```diff
-if [ "$recipeAction" = TEST ]; then
-BUILD_PREREQUIRES+="
+TEST_REQUIRES="
 	cmd:kyua
 	"
-fi
```

alaviss: I could actually remove the entire block. It doesn't work at all

fbrosson:
Sorry @alaviss, I wanted to do a full review of this PR since a long time ago, but I kept postponning it this task...
(Well, I did review a few recipes (in this PR) once, but you have since made several enhancements, so I really need to review all of its recipes.)
I promisse I'll review it soon, probably in one or two days.

alaviss:
Updates:
- atf: cleaned up
- kyua: drop suffix on x86_gcc2

dev-util/kyua/kyua-0.13.recipe
```diff
@@ -0,0 +1,101 @@
...
+	"
+BUILD_REQUIRES+="$TEST_REQUIRES"
+BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
+
+USER_SETTINGS_FILES="
+	settings/kyua directory
+	settings/kyua/kyua.conf template $dataDir/kyua/examples/kyua.conf"
+
+defineDebugInfoPackage kyua$secondaryArchSuffix \
+	"$binDir/kyua"
```
fbrosson: `$binDir` ➡️ `$commandBinDir`

dev-lua/lutok/lutok-0.4.recipe
```diff
@@ -0,0 +1,102 @@
...
+
+TEST_REQUIRES="
+	devel:libatf_c++$secondaryArchSuffix
+	"
+TEST_PREREQUIRES="
+	cmd:atf_sh$secondaryArchSuffix
+	cmd:kyua$secondaryArchSuffix
+	"
+BUILD_REQUIRES+="$TEST_REQUIRES"
+BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
```
fbrosson: Workaround for our chicken/egg issue with `TEST_{PRE,}REQUIRES`:
```diff
-BUILD_REQUIRES+="$TEST_REQUIRES"
+#BUILD_REQUIRES+="$TEST_REQUIRES"
-BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
+#BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
```

dev-libs/atf/atf-0.21.recipe
```diff
@@ -0,0 +1,106 @@
...
+	cmd:awk
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:grep
+	cmd:make
+	cmd:sed
+	"
+
+TEST_PREREQUIRES="cmd:kyua"
+BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
```
fbrosson: Workaround for our chicken/egg issue with `TEST_PREREQUIRES`:
```diff
-BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
+#BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,105 @@
...
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="2.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	pkgconf$secondaryArchSuffix = $portVersion
+	cmd:pkgconf$secondaryArchSuffix = $portVersion
+	cmd:pkg_config$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 ARCHITECTURES="!x86_gcc2 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
 libVersion="2.0.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
 	pkgconf$secondaryArchSuffix = $portVersion
-	cmd:pkgconf$secondaryArchSuffix = $portVersion
-	cmd:pkg_config$secondaryArchSuffix = $portVersion
+	cmd:pkgconf$commandSuffix = $portVersion
 	lib:libpkgconf$secondaryArchSuffix = $libVersionCompat
 	"
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,105 @@
...
+	lib:libpkgconf$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+CONFLICTS="
+	cmd:pkg_config
```
fbrosson:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
-CONFLICTS="
-	cmd:pkg_config
-	"
+
+PROVIDES_pkgconfig="
+	pkgconf${secondaryArchSuffix}_pkgconfig = $portVersion
+	cmd:pkg_config$secondaryArchSuffix = $portVersion
+	"
+REQUIRES_pkgconfig="
+	pkgconf$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_pkgconfig="
+	cmd:pkg_config$secondaryArchSuffix
+	"
 ```

File: dev-util/pkgconf/pkgconf-1.3.90.recipe [hunk 10]
```diff
@@ -0,0 +1,105 @@
...
+	cmd:grep
+	cmd:make
+	cmd:sed
+	"
+
+TEST_REQUIRES="
+	devel:libatf_c$secondaryArchSuffix
+	"
+TEST_PREREQUIRES="
+	cmd:kyua$secondaryArchSuffix
```
fbrosson:
```diff
 TEST_PREREQUIRES="
-	cmd:kyua$secondaryArchSuffix
+	cmd:kyua$commandSuffix
 	"
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,105 @@
...
+	"
+BUILD_REQUIRES+="$TEST_REQUIRES"
+BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
+
+defineDebugInfoPackage pkgconf$secondaryArchSuffix \
+	"$binDir/pkgconf" \
```
fbrosson:
```diff
 defineDebugInfoPackage pkgconf$secondaryArchSuffix \
-	"$binDir/pkgconf" \
+	"$commandBinDir/pkgconf" \
 	"$libDir/libpkgconf.so.$libVersion"
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,105 @@
...
+		"$(finddir B_USER_DEVELOP_DIRECTORY)/$pkgConfSubDir"
+		"$(finddir B_SYSTEM_NONPACKAGED_DEVELOP_DIRECTORY)/$pkgConfSubDir"
+		"$(finddir B_SYSTEM_DEVELOP_DIRECTORY)/$pkgConfSubDir"
+		"$(finddir B_USER_NONPACKAGED_DATA_DIRECTORY)/pkgconfig"
+		"$(finddir B_USER_DATA_DIRECTORY)/pkgconfig"
+		"$(finddir B_SYSTEM_NONPACKAGED_DATA_DIRECTORY)/pkgconfig"
+		"$(finddir B_SYSTEM_DATA_DIRECTORY)/pkgconfig"
+		)
+
+	runConfigure ./configure \
```
fbrosson:
```diff
-	runConfigure ./configure \
+	runConfigure --omit-dirs binDir ./configure --bindir="$commandBinDir" \
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,105 @@
...
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir" "$developDocDir"
+	install -t "$docDir" -m 644 AUTHORS NEWS
+	install -t "$developDocDir" -m 644 doc/*.rst
+
+	ln -s pkgconf "$binDir/pkg-config"
```
fbrosson:
```diff
-	ln -s pkgconf "$binDir/pkg-config"
+	mkdir -p "$binDir"
+	ln -r -s "$commandBinDir"/pkgconf "$binDir"/pkg-config
+	packageEntries pkgconfig \
+		"$binDir"/pkg-config
+	if [ -n "$secondaryArchSuffix" ]; then
+		rmdir "$binDir"
+	fi
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
alaviss: For `pkgconf` I would want the suffix, as a small reminder that on x86_gcc2, it would only works with secondary arch libs.

fbrosson: If I'm not mistaken, pkgconfig should be usable on any recipe, including recipes that build on x86_gcc2 primary arch and that do not provide support for building on x86 secondary arch.

So, if this is true, then we really need to provide `cmd:pkgconf$commandSuffix` instead of `cmd:pkgconf$secondaryArchSuffix`.
Otherwise, on a recipe without support for secondary architectures, we would need to hardcode that, when building on x86_gcc2 primary arch, we would require pkgconf_x86.

So, by using `commandSuffix` for `cmd:pkgconf`, we keep things simple, since all cases are handled by:
* a `cmd:pkgconf` (or `cmd:pkg_config`) on recipes that do not define `SECONDARY_ARCHITECTURES`,
* a `cmd:pkgconf` (or `cmd:pkg_config$secondaryArchSuffix`) on recipes that define `SECONDARY_ARCHITECTURES` but do not define a `commandSuffix`,
* a `cmd:pkgconf$commandSuffix` (or `cmd:pkg_config$secondaryArchSuffix`) on recipes that define both `SECONDARY_ARCHITECTURES` and `commandSuffix`.

alaviss: It's basically the same reason why we need two version of `cmd:pkg_config` on x86 secondary arch.

The [`.pc` search path][0] is configured to look into `lib/$secondaryArchSuffix`. So using it for gcc2 wouldn't make any sense, as libs in `lib/x86` are incompatible with gcc2

[0]: https://github.com/haikuports/haikuports/pull/2026/files#diff-d10935f4784240a8d4b63c74253d0880R67

fbrosson: BTW, on x86_gcc2 **_non_** hybrid, it won't be possible to use `pkgconf` instead of `pkg-config`, so this means that, as long as x86_gcc2 lives:
1) we will need to keep the `cmd:pkg_config` in `BUILD_PREREQUIRES` for all recipes that build on x86_gcc2 primary arch.
2) it will **_only_** be OK to replace `cmd:pkg_config$secondaryArchSuffix` by `cmd:pkgconf` in those recipes which do **_not_** build on x86_gcc2 primary arch.

fbrosson: > It's basically the same reason why we need two version of cmd:pkg_config on x86 secondary arch.

Ok, I see. I guess you are correct.

alaviss:
Updates:
- dev-libs/atf, dev-lua/lutok:
  - use `commandSuffix` for dev-util/kyua
  - disabled `TEST()` to break dependency cycle
- dev-util/pkgconf:
  - move `cmd:pkg_config` to a separate package
  - cleaned up copyright years

All changes remains untested.

dev-util/pkgconf/pkgconf-1.3.90.recipe
fbrosson: OK, this means pkgconfig won't be usable for building anything on x86_gcc2 primary arch.

So you are perfectly right. The `cmd:pkgconf$secondaryArchSuffix` is correct.

fbrosson: Do you think it woud be possible, when building pkgconf on x86 second arch, to change the default search path to that of the primary arch (i.e. `/boot/system/develop/lib/pkgconfig/`), and create a wrapper script, `pkgconf-x86`, that would use the search path for the secondary arch, `/boot/system/develop/lib/x86/pkgconfig/` ?

This would allow, on x86_gcc2 hybrid, to uninstall pkgconfig and install pkgconf_pkgconfig (providing the symlink) and start testing the transition from pkgconfig to pkgconf even before x86_gcc2 is abandonned.

Well, after all, this seems a bit too complicated for a rather small result.
So let's ignore this comment.

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,121 @@
...
+PROVIDES_devel="
+	pkgconf${secondaryArchSuffix}_devel = $portVersion
+	devel:libpkgconf$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	pkgconf$secondaryArchSuffix == $portVersion base
+	"
+
+SUMMARY_pkgconfig="$SUMMARY (pkg-config compat symlink)"
+DESCRIPTION_pkgconfig="$DESCRIPTION (pkg-config compatibility symlink)"
```
fbrosson: I would either not define `DESCRIPTION_pkgconfig` at all or, if you prefer, keep it but with this change:
```diff
-DESCRIPTION_pkgconfig="$DESCRIPTION (pkg-config compatibility symlink)"
+DESCRIPTION_pkgconfig="$DESCRIPTION
+(pkg-config compatibility symlink)"
```

fbrosson: Well, after all, the `DESCRIPTION_pkgconfig` is OK as it is now. I'll delete my previous comment.

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,121 @@
...
+BUILD_REQUIRES+="$TEST_REQUIRES"
+BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
+
+defineDebugInfoPackage pkgconf$secondaryArchSuffix \
+	"$binDir/pkgconf" \
+	"$libDir/libpkgconf.so.$libVersion"
+
+BUILD()
+{
+	local pkgConfSubDir="lib/${secondaryArchSuffix/_}/pkgconfig"
```
fbrosson:
```diff
-	local pkgConfSubDir="lib/${secondaryArchSuffix/_}/pkgconfig"
+	local pkgConfSubDir="$lrelativeLibDir/pkgconfig"
```
This change is neutral on any secondary arch (e.g. x86), but not primary archs (e.g. x86_64).

dev-libs/atf/atf-0.21.recipe
```diff
@@ -0,0 +1,111 @@
...
+COPYRIGHT="2007-2012 The NetBSD Foundation, Inc.
+	2011, 2012, 2014 Google Inc."
+LICENSE="BSD (2-clause)
+	BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/jmmv/atf/releases/download/atf-$portVersion/atf-$portVersion.tar.gz"
+CHECKSUM_SHA256="92bc64180135eea8fe84c91c9f894e678767764f6dbc8482021d4dde09857505"
+PATCHES="atf-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
```
fbrosson: Shall we add `?x86` ?
```diff
-ARCHITECTURES="!x86_gcc2 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```

dev-libs/atf/atf-0.21.recipe
```diff
@@ -0,0 +1,111 @@
...
+	cmd:autoconf
+	cmd:awk
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:grep
+	cmd:make
+	cmd:sed
+	"
+
+TEST_PREREQUIRES="cmd:kyua$commandSuffix"
```
fbrosson: Shall we change the style?
```diff
-TEST_PREREQUIRES="cmd:kyua$commandSuffix"
+TEST_PREREQUIRES="
+       cmd:kyua$commandSuffix
+       "
+
 #BUILD_PREREQUIRES+="$TEST_PREREQUIRES"
```

dev-lua/lutok/lutok-0.4.recipe
```diff
@@ -0,0 +1,107 @@
...
+layers of abstraction and error checking that go against the original spirit of \
+the Lua C API and thus degrade performance."
+HOMEPAGE="https://github.com/jmmv/lutok"
+COPYRIGHT="2011, 2012, 2014 Google Inc"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/jmmv/lutok/releases/download/lutok-$portVersion/lutok-$portVersion.tar.gz"
+CHECKSUM_SHA256="2cec51efa0c8d65ace8b21eaa08384b77abc5087b46e785f78de1c21fb754cd5"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
```
fbrosson: Shall we add `?x86` ?
```diff
-ARCHITECTURES="!x86_gcc2 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```

dev-util/pkgconf/pkgconf-1.3.90.recipe
alaviss: It's not really an error, `//` :arrow_right: `/`. But yours looks way better, so I'm going to use it instead.

dev-libs/atf/atf-0.21.recipe
alaviss: Do we still support `x86` as a main architecture?

dev-util/pkgconf/pkgconf-1.3.90.recipe
```diff
@@ -0,0 +1,121 @@
...
+		"$(finddir B_SYSTEM_NONPACKAGED_DEVELOP_DIRECTORY)/$pkgConfSubDir"
+		"$(finddir B_SYSTEM_DEVELOP_DIRECTORY)/$pkgConfSubDir"
+		"$(finddir B_USER_NONPACKAGED_DATA_DIRECTORY)/pkgconfig"
+		"$(finddir B_USER_DATA_DIRECTORY)/pkgconfig"
+		"$(finddir B_SYSTEM_NONPACKAGED_DATA_DIRECTORY)/pkgconfig"
+		"$(finddir B_SYSTEM_DATA_DIRECTORY)/pkgconfig"
+		)
+
+	runConfigure ./configure \
+		--with-pkg-config-dir="$(IFS=:; echo \"${pkgConfPath[*]}\")"
```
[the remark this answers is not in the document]
alaviss: About the errors, it is because `/boot/system/develop/lib/pkgconfig` is passed to `--host` somehow. I'm planning to rework the patches to use `find_path` anyway, so this should not be a problem.

> BTW, I saw that pkgconf 1.4.2 was released very recently.

I noticed it when you suggested some changes weeks ago, but I'm currently lacking the time to work on this. I will see if I could manage to bring the VM up this weekend.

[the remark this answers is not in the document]
fbrosson: I'll test with just one dir:
```
    runConfigure ./configure \
        --with-pkg-config-dir="`finddir B_SYSTEM_DEVELOP_DIRECTORY/$relativeLibDir/pkgconfig"
```
I think this will work.

BTW, I've been thinking about a way to teach pkgconf how to work for both x86 secondary arch and x86_gcc2 primary arch:
Imagine you pass something that would roughly look like:
```
-DHAIKU_SYSTEM_DEVELOP_DIR="`finddir B_SYSTEM_DEVELOP_DIRECTORY`" \
-DHAIKU_RELATIVE_BIN_DIR="$relativeBinDir" \
-DHAIKU_RELATIVE_LIB_DIR="$relativeBinDir" \
```
to the compiler. pkgconf could then examine what `*argv` ends with. pkgconf_x86 would then be usable when building packages for x86_gcc2 primary arch, as well as for x86 secondary arch.
(Of course, with `find_directory()` we would not even need to pass a `HAIKU_SYSTEM_DEVELOP_DIR`.)

End of `*argv` | pkgconfig search path
-- | --
`/bin/pkgconf` | `/system/develop/lib/pkgconfig/`
`/bin/x86/pkgconf` | `/system/develop/lib/x86/pkgconfig/`
`/bin/pkgconf-x86` | `/system/develop/lib/x86/pkgconfig/`

Of course, we would include this after `SECONDARY_ARCHITECTURES="x86"`:
```bash
commandBinDir=$prefix/bin
	commandBinDir=$prefix/bin
fi
```
and this in `INSTALL()`:
```bash
		mkdir -p "$commandBinDir"
		ln -r -s "$binDir"/pkgconf "$commandBinDir"
	fi
```
and this after `PROVIDES`:
```bash
PROVIDES="$PROVIDES
	cmd:pkgconfig
	"
fi
```

This would be a lot of work... So don't worry if you don't have the time right now.
It can be done later. We could also ignore x86_gcc2, since we know it will be dropped in the future.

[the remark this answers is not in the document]
pulkomandy: Ideally no path should be hardcoded at build time, and pkgconf would use find_path to locate the directory. It could even examine the path to determine if a "setarch x86" had been used (and perhpaps we could ease this by providing some extra environment variable in the setarch environment if there is no easy way to detect it).

Anyways, I'm fine with pkgconf pointing to the default (that is x86_gcc2) directory if nothing is specified, and having setarch set the appropriate variables so it ends up using x86 directories when in a setarch environment. Haikuporter could likewise set these by default. I think that is how pkgconf is intended to be used and it is also more generic and allows us to use it for cross-compilation in the future.

dev-libs/atf/atf-0.21.recipe
fbrosson: Nope. x86 is only supported as a second arch for x86_gcc2 primary arch.

That said, as we know, if a package builds fine on x86 secondary arch then it will also build fine on x86 primary arch, unless the recipe uses hard coded paths or if it depends on recipes which do not have `x86` (or `?x86`) in `ARCHITECTURES`...

So, I think it would be nice to keep a `?x86` in `ARCHITECTURES` so that those users (like myself) who are still running Haiku x86 will be able to play with pkgconf.

BTW, I'm not against the switch to x86_64. Actually, I can run all "flavours" of Haiku at the same time using separate VMs of course. And they all use a common haikuports tree, shared via NFS.
My 3 Haiku systems are on (different partitions of) a single SDRAM which I can even use to boot real hardware.

So, if you agree with having a `?x86` in `ARCHITECTURES` that would be cool :)

alaviss:
Updates:
- pkgconf
  - updated to 1.4.2
  - now make use of `find_paths` for pkgconfig directories
- atf, lutok, kyua: now uses `TEST_REQUIRES`

fbrosson:
(I can't test it right now but I'll do so before tomorrow.)

dev-util/kyua/patches/kyua-0.13.patchset
```diff
@@ -0,0 +1,660 @@
...
+diff --git a/utils/process/status_test.cpp b/utils/process/status_test.cpp
+index 5a3e19e..603a90d 100644
+--- a/utils/process/status_test.cpp
++++ b/utils/process/status_test.cpp
+@@ -187,7 +187,7 @@ ATF_TEST_CASE_BODY(integration__coredump)
+     ATF_REQUIRE(!coredump.exited());
+     ATF_REQUIRE(coredump.signaled());
+     ATF_REQUIRE_EQ(SIGQUIT, coredump.termsig());
+-#if !defined(WCOREDUMP)
++#if !defined(WCOREDUMP) || !defined(__HAIKU__)
```
fbrosson: I might be wrong, but I think what you want here was a `&&`, not a `||`:
```diff
-+#if !defined(WCOREDUMP) || !defined(__HAIKU__)
++#if !defined(WCOREDUMP) && !defined(__HAIKU__)
```

BTW, it would be cool to include `?x86` in `ARCHITECTURES`. It does not cost much and it can be handy.
After all, that arch is not dead yet ;) Thanks :)

alaviss: It's probably an artifact of patching while sleepy :laughing:, it wasn't even necessary.

dev-util/kyua/kyua-0.13.recipe
```diff
@@ -0,0 +1,97 @@
...
+	cmd:doxygen
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+TEST_REQUIRES="
+	devel:libatf_c++$secondaryArchSuffix
```
fbrosson: Oops, you forgot to address [this comment](https://github.com/haikuports/haikuports/pull/2026/files#r178125185).
(I've just checked again, with the same conclusion.)

fbrosson:
Package | Version | Tests 
-- | -- | --
dev-libs/atf | 0.21 | 319/319 passed (0 failed) 
dev-lua/lutok | 0.4 | 101/101 passed (0 failed) 
dev-util/kyua | 0.13 | Test cases: 1215 total, 41 skipped, 4 expected failures, 3 broken, 3 failed | yes BUILD_REQUIRES devel:libatf_c++
dev-util/pkgconf | 1.4.2 | 112/112 passed (0 failed)

(Remark: I haven't played with pkgconf yet.)

alaviss:
I've mostly given up with kyua. Some test cases won't fail unless run with kyua, and the debugger doesn't stop at the point where the test case is executed :(

I'm working on my patch upstream and will update this port accordingly once the PR at pkgconf is merged.

fbrosson:
We don't need to worry about the (very small number of) tests that are failing for kyua's :)

dev-util/pkgconf/patches/pkgconf-1.4.2.patchset
```diff
@@ -0,0 +1,92 @@
...
+index 811e043..a832e83 100644
+--- a/libpkgconf/client.c
++++ b/libpkgconf/client.c
+@@ -78,7 +78,11 @@ pkgconf_client_init(pkgconf_client_t *client, pkgconf_error_handler_func_t error
+ 	pkgconf_path_build_from_environ("PKG_CONFIG_SYSTEM_INCLUDE_PATH", SYSTEM_INCLUDEDIR, &client->filter_includedirs, false);
+ 
+ 	/* GCC uses these environment variables to define system include paths, so we should check them. */
++#ifdef __HAIKU__
++	pkgconf_path_build_from_environ("BELIBRARIES", NULL, &client->filter_libdirs, false);
++#else
```
fbrosson:
```
#ifdef __HAIKU__
	pkgconf_path_build_from_environ("BELIBRARIES", NULL, &client->filter_libdirs, false);
#else
	pkgconf_path_build_from_environ("LIBRARY_PATH", NULL, &client->filter_libdirs, false);
#endif
```
I don't understand why you skip the `LIBRARY_PATH` when `__HAIKU__` is defined. `LIBRARY_PATH` **_is_** used in Haiku. (Actually, if I'm not mistaken, Haiku's `LIBRARY_PATH` is the equivalent of Linux's `LD_LIBRARY_PATH`.)
Well, I might be missing something, so maybe there is no error here.

pulkomandy: That's the problem, Haiku uses LIBRARY_PATH in the runtime_loader for a purpose similar to LD_LIBRARY_PATH in Linux. So if you change LIBRARY_PATH, some apps will stop running because they don't find their runtime libraries.

In Linux, LIBRARY_PATH is used to tell the dev tools where to look for libraries. This is indeed somewhat similar to our BELIBRARIES.

fbrosson: Thanks a lot, @pulkomandy, for this very important info.

alaviss:
Updates:
- kyua
  - Corrected a few folder locations (no effect on test results)
- pkgconf
  - Using upstreamed patch

Not tested on: x86_gcc2

Note: `find_paths` might also include x86_gcc2 lib paths into the returned pkgconfig paths, so this should be tested on x86_gcc2.

fbrosson:
Excellent! pkgconf seems to work fine on x86_64 and on x86 secondary arch.
(I'll test it on x86 primary arch soon. Thanks for making this possible, btw ;)

So, on x86_gcc2, right now we can only use it with packages which are installed for x86 secondary arch, but this is a great step, as it will allow to switch from pkg-config to pkgconf and even retire pkg-config once we have dropped support for x86_gcc2.

I have tested a few options (`--variable=prefix`, `--libs` and `--cflags`) of pkgconf on x86_64 (and pkgconf-x86 on x86_gcc2) with a few packages such as libpng16, openssl. The results are all OK.

So I'd say that the PR is ready for a merge and I guess the other reviewers will agree with this.

@alaviss, if you would like to merge you have my approval. Thanks again for your great work!

(I'll rerun the tests but I guess the results will be the same as the last time.)

@alaviss, I have a new challenge for you: teach pkgconf how to work for packages installed for a primary arch (e.g. x86_gcc2) when it was itself built for a secondary arch, e.g. x86. But please feel free to have some holidays before you try this ;)

alaviss:
Oops, I accidentally messed up with the git-remote. Can someone delete the pkgconf branch in haikuports/haikuports for me? Thanks.

alaviss:
Sorry for all the noise, I messed up with `git-remote`

PR #2028, opened 2018-01-01
libpciacces, bump version, move to the proper place

Begasus:


x11-libs/libpciaccess/libpciaccess-0.14.recipe
```diff
@@ -1,32 +1,42 @@
...
 REVISION="1"
-SOURCE_URI="https://xorg.freedesktop.org/releases/individual/lib/libpciaccess-$portVersion.tar.gz"
-CHECKSUM_SHA256="fe26ec788732b4ef60b550f2d3fa51c605d27f646e18ecec878f061807a3526e"
+SOURCE_URI="https://www.x.org/releases/individual/lib/libpciaccess-$portVersion.tar.bz2"
+CHECKSUM_SHA256="3df543e12afd41fea8eac817e48cbfde5aed8817b81670a4e9e493bb2f5bf2a4"
 PATCHES="libpciaccess-$portVersion.patchset"
 
-ARCHITECTURES="?x86_gcc2 ?x86 ?x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
```
fbrosson: @Begasus, if you feel like writing a patch for fixing an error at line 230 of `scanpci/scanpci.c`, you can mark x86_gcc2 as tested :)
```diff
-ARCHITECTURES="?x86_gcc2 x86 x86_64"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
Here's the diff for the patch:
```diff
     ret = pci_system_init();
     if (ret != 0)
-	err(1, "Couldn't initialize PCI system");
+	err(1, "%s", "Couldn't initialize PCI system");
```

I'm not sure, however, if the correct format to use is just `"%s"` or if we should use `"%s\n"`. I'll leave that to you ;-)

BTW, everything else looks OK in your recipe, and I'll test it soon on x86_64.

fbrosson: Shall we use a PATCH() function?
```
PATCH()
{

		sed -i -e "s/\(err(1, \)\(\"Couldn't initialize PCI system\");\)/\1\"%s\", \2/;" \
			scanpci/scanpci.c
	fi
}
```
Or would you prefer to mark this recipe as broken on x86_gcc2?

Begasus: Insterted it as a PATCH function, easier to remove is we would like to set x86_gcc2 as broken, it builds ok, but there is no need for it as other deppendencies that would call it won't work with gcc2*

x11-libs/libpciaccess/libpciaccess-0.14.recipe
```diff
@@ -41,13 +51,26 @@ BUILD_PREREQUIRES="
...
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+defineDebugInfoPackage libpciaccess$secondaryArchSuffix \
+	$libDir/libpciaccess.so.$libVersion
+
+PATCH()
+{
+
```
fbrosson: No need for a blank line here.

Oops, that's my fault. Sorry, I had not noticed I had inserted a blank line in my comment.

PR #2035, opened 2018-01-01
gnupg, bump version

Begasus:
Needs new version of libassuan: https://github.com/haikuports/haikuports/pull/2036


Begasus:
We still have 1.4.21 that still builds on gcc2, 2.2.* versions are not supported there, should we rename this recipe to gnu**2**-2.2.4.recipe?

Begasus:
Actually depot would take care if you want to install the gcc2 or gcc5 version, so no need to rename it

PR #2036, opened 2018-01-01
libassuan, bump version, cleanup

Begasus:


dev-libs/libassuan/libassuan-2.5.1.recipe
```diff
@@ -80,5 +84,6 @@ INSTALL()
 		$binDir/libassuan-config
 
 	packageEntries devel \
-		$developDir $binDir/libassuan-config
+		$developDir \
+		$binDir/libassuan-config
```
fbrosson:
```diff
 	packageEntries devel \
 		$developDir \
-		$binDir/libassuan-config
+		$binDir/libassuan-config \
+		$dataDir
```
(This will move `data/aclocal/libassuan.m4` to the devel package.)

Begasus: Any reason why not to include the *.info file? I see multiple files in the documentation dir?

dev-libs/libassuan/libassuan-2.5.1.recipe
```diff
@@ -1,26 +1,28 @@
...
+components. Both, server and client side functions are provided."
+COPYRIGHT="2001-2013 Free Software Foundation, Inc.
+	2001-2015 g10 Code GmbH"
 LICENSE="
 	GNU GPL v3
 	GNU LGPL v2.1
 	"
 REVISION="1"
-HOMEPAGE="http://www.gnupg.org/related_software/libassuan/index.en.html"
+HOMEPAGE="https://www.gnupg.org/related_software/libassuan/index.en.html"
```
fbrosson:
```diff
-HOMEPAGE="https://www.gnupg.org/related_software/libassuan/index.en.html"
+HOMEPAGE="https://www.gnupg.org/related_software/libassuan/"
```
This is better for 3 reasons:
* The webmaster already has `index.en.html` as default page and sends it without forcing a redirect.
* Should the webmaster add a translated page for another language spoken in the country where the client is, he would receive the content for the translated page (or be redirected to the alternate page).
* This allows the webmaster to switch from HTML to PHP (or any other language), should such a decision be taken.

So for non english-speaking users, it is better to not specify any language in `HOMEPAGE`.

PR #2046, opened 2018-01-02
Update icu-60.1.recipe

scottmc:


kenmays:
@scottmc - I've successfully tested ICU 61.1 on both Haiku x86/86_64 with GCC 7.3. I'll
leave it up to you...

PR #2049, opened 2018-01-03
p11-kit: new recipe [GCI 2017]

sfanxiang:
Current status:
- builds are successful
- all commands segfault

### DIFF
--- a/app-crypt/p11-kit/p11_kit-0.23.9.recipe
+++ b/app-crypt/p11-kit/p11_kit-0.23.9.recipe
@@ -11,15 +11,18 @@ CHECKSUM_SHA256="e1c1649c335107a8d33cf3762eb7f57b2d0681f0c7d8353627293a58d6b4db6
 SOURCE_DIR="p11-kit-$portVersion"
 PATCHES="p11_kit-$portVersion.patchset"
 
-ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="x86"
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+libVersion="0.23.9"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
 	p11_kit$secondaryArchSuffix = $portVersion
 	cmd:p11_kit$secondaryArchSuffix = $portVersion
 	cmd:trust$secondaryArchSuffix = $portVersion
-	lib:libp11_kit$secondaryArchSuffix = $portVersion
-	lib:p11_kit_proxy$secondaryArchSuffix = $portVersion
+	lib:libp11_kit$secondaryArchSuffix = $libVersionCompat
+	lib:p11_kit_proxy$secondaryArchSuffix = $libVersionCompat
 	"
 
 # TODO: decide if ca_root_certificates is actually necessary
@@ -32,8 +35,7 @@ REQUIRES="
 
 PROVIDES_devel="
 	p11_kit${secondaryArchSuffix}_devel = $portVersion
-	devel:libp11_kit$secondaryArchSuffix = $portVersion
-	devel:p11_kit_proxy$secondaryArchSuffix = $portVersion
+	devel:libp11_kit$secondaryArchSuffix = $libVersion
 	"
 REQUIRES_devel="
 	p11_kit$secondaryArchSuffix == $portVersion base
@@ -58,12 +60,21 @@ BUILD_PREREQUIRES="
 	"
 
 USER_SETTINGS_FILES="
-	settings/pkcs11 directory
+	settings/pkcs11/pkcs11.conf template data/pkcs11/pkcs11.conf.example
 	"
 
 GLOBAL_WRITABLE_FILES="
-	settings/pkcs11/pkcs11.conf.example keep-old
+	settings/pkcs11/pkcs11.conf keep-old
 	"
+
+defineDebugInfoPackage p11_kit$secondaryArchSuffix \
+	$binDir/p11-kit \
+	$binDir/trust \
+	$libDir/libp11-kit.so.$libVersion \
+	$libDir/p11-kit/p11-kit-remote \
+	$libDir/p11-kit/p11-kit-server \
+	$libDir/pkcs11/p11-kit-client.so \
+	$libDir/pkcs11/p11-kit-trust.so
 
 BUILD()
 {
@@ -80,12 +91,26 @@ INSTALL()
 {
 	make install
 
+	# Replace, in the config example, the real path of the settings dir by
+	# the user-friendly path. Then install it as the config file.
+	sed -i \
+		-e "s|$sysconfDir|`finddir B_SYSTEM_SETTINGS_DIRECTORY`|;" \
+		$settingsDir/pkcs11/pkcs11.conf.example
+	install -d $dataDir/pkcs11
+	install -t $dataDir/pkcs11 $settingsDir/pkcs11/pkcs11.conf.example
+	mv \
+		$settingsDir/pkcs11/pkcs11.conf.example \
+		$settingsDir/pkcs11/pkcs11.conf
+
 	# clean up
 	rm -f $libDir/*.la $libDir/pkcs11/*.la
 
-	prepareInstalledDevelLibs \
-		libp11-kit \
-		p11-kit-proxy
+	# fix incorrect install directory on secondary arch builds
+	if [ -n "$secondaryArchSuffix" ]; then
+		mv $prefix/lib/p11-kit $libDir/p11-kit
+	fi
+
+	prepareInstalledDevelLib libp11-kit
 	fixPkgconfig
 
 	packageEntries devel \
--- a/app-crypt/p11-kit/p11_kit-0.23.9.recipe
+++ b/app-crypt/p11-kit/p11_kit-0.23.9.recipe
@@ -11,16 +11,24 @@ CHECKSUM_SHA256="e1c1649c335107a8d33cf3762eb7f57b2d0681f0c7d8353627293a58d6b4db6
 SOURCE_DIR="p11-kit-$portVersion"
 PATCHES="p11_kit-$portVersion.patchset"
 
-ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandBinDir=$binDir
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]
+then
+	commandBinDir=$prefix/bin
+	commandSuffix=
+fi
 
 libVersion="0.3.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
 	p11_kit$secondaryArchSuffix = $portVersion
-	cmd:p11_kit$secondaryArchSuffix = $portVersion
-	cmd:trust$secondaryArchSuffix = $portVersion
+	cmd:p11_kit$commandSuffix = $portVersion
+	cmd:trust$commandSuffix = $portVersion
 	lib:libp11_kit$secondaryArchSuffix = $libVersionCompat
 	lib:p11_kit_proxy$secondaryArchSuffix = $libVersionCompat
 	"
@@ -67,8 +75,8 @@ GLOBAL_WRITABLE_FILES="
 	"
 
 defineDebugInfoPackage p11_kit$secondaryArchSuffix \
-	$binDir/p11-kit \
-	$binDir/trust \
+	$commandBinDir/p11-kit \
+	$commandBinDir/trust \
 	$libDir/libp11-kit.so.$libVersion \
 	$libDir/p11-kit/p11-kit-remote \
 	$libDir/p11-kit/p11-kit-server \
@@ -79,7 +87,7 @@ BUILD()
 {
 	certsInstallDir="$portPackageLinksDir/ca_root_certificates"
 	certsDir="$certsInstallDir/$relativeDataRootDir/ssl"
-	runConfigure ./configure \
+	runConfigure --omit-dirs binDir ./configure --bindir $commandBinDir \
 		LDFLAGS="-lnetwork" \
 		--with-trust-paths="${certsDir}/CARootCertificates.pem" \
 		--with-user-config="`finddir B_USER_SETTINGS_DIRECTORY`/pkcs11"
--- a/app-crypt/p11-kit/p11_kit-0.23.9.recipe
+++ b/app-crypt/p11-kit/p11_kit-0.23.9.recipe
@@ -85,9 +85,8 @@ INSTALL()
 	make install
 
 	# clean up
-	rm -f $libDir/*.la
-	rm -f $libDir/pkcs11/*.la
-	
+	rm -f $libDir/*.la $libDir/pkcs11/*.la
+
 	prepareInstalledDevelLibs \
 		libp11-kit \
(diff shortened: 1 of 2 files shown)


korli:
the lib: and devel: should come with compat strings, ie = 0.3.0 compat >= 0

what's actually lib:p11_kit_proxy?

sfanxiang:
It's a symlink (p11-kit-proxy.so -> libp11-kit.so.0.3.0) created by make install.

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,97 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libffi$secondaryArchSuffix
+	devel:libtasn1$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autom4te
+	cmd:automake
+	cmd:autoreconf
```
korli: When you don't use autoreconf/autogen.sh explicitly these should not be required.

sfanxiang: Actually none of these are required because configure is already available.
Should I remove these?

sfanxiang:
Some segfaults are fixed. Some more exist and they are straight inside the memory manager:

/home/builder/builds/haiku/src/system/libroot/posix/malloc/superblock.h:194:void BPrivate::superblock::putBlock(BPrivate::block*): getNumAvailable() < getNumBlocks()

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+USER_SETTINGS_FILES="
+	settings/pkcs11 directory
+	"
+
+GLOBAL_WRITABLE_FILES="
+	settings/pkcs11/pkcs11.conf.example keep-old
```
fbrosson:
```diff
 USER_SETTINGS_FILES="
 	settings/pkcs11 directory
+	settings/pkcs11/pkcs11.conf template data/pkcs11/pkcs11.conf.example
 	"
 
 GLOBAL_WRITABLE_FILES="
-	settings/pkcs11/pkcs11.conf.example keep-old
+	settings/pkcs11/pkcs11.conf keep-old
 	"
 
+defineDebugInfoPackage p11_kit$secondaryArchSuffix \
+	$binDir/p11-kit \
+	$binDir/trust \
+	$libDir/libp11-kit.so.$libVersion \
+	$libDir/p11-kit/p11-kit-remote \
+	$libDir/p11-kit/p11-kit-server \
+	$libDir/pkcs11/p11-kit-client.so \
+	$libDir/pkcs11/p11-kit-trust.so
+
```

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+
+INSTALL()
+{
+	make install
+
+	# clean up
+	rm -f $libDir/*.la $libDir/pkcs11/*.la
+
+	prepareInstalledDevelLibs \
+		libp11-kit \
```
fbrosson:
```diff
 INSTALL()
 {
 	make install
 
+	# Replace, in the config example, the real path of the settings dir by
+	# the user-friendly path. Then install it as the config file.
+	sed -i \
+		-e "s|$sysconfDir|`finddir B_SYSTEM_SETTINGS_DIRECTORY`|;" \
+		$settingsDir/pkcs11/pkcs11.conf.example
+	install -d $dataDir/pkcs11
+	install -t $dataDir/pkcs11 $settingsDir/pkcs11/pkcs11.conf.example
+	mv \
+		$settingsDir/pkcs11/pkcs11.conf.example \
+		$settingsDir/pkcs11/pkcs11.conf
+
 	# clean up
 	rm -f $libDir/*.la $libDir/pkcs11/*.la
 
+	# fix incorrect install directory on secondary arch builds
+	if [ -n "$secondaryArchSuffix" ]; then
+		mv $prefix/lib/p11-kit $libDir/p11-kit
+	fi
+
-	prepareInstalledDevelLibs \
-		libp11-kit \
-		p11-kit-proxy
+	prepareInstalledDevelLib libp11-kit
 	fixPkgconfig
 
 	packageEntries devel \
 		$developDir
 }
```
Notes: `p11-kit-proxy.so` is just a symlink to `libp11-kit.so.0.3.0`. We only need to provide that symlink in the base package. In the devel package it is (probably) not needed. So we don't need to pass it to `prepareInstalledDevelLib`.

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+	haiku$secondaryArchSuffix
+	ca_root_certificates
+	lib:libffi$secondaryArchSuffix
+	lib:libtasn1$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	p11_kit${secondaryArchSuffix}_devel = $portVersion
+	devel:libp11_kit$secondaryArchSuffix = $portVersion
+	devel:p11_kit_proxy$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES_devel="
 	p11_kit${secondaryArchSuffix}_devel = $portVersion
-	devel:libp11_kit$secondaryArchSuffix = $portVersion
+	devel:libp11_kit$secondaryArchSuffix = $libVersion
-	devel:p11_kit_proxy$secondaryArchSuffix = $portVersion
 	"
```

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	p11_kit$secondaryArchSuffix = $portVersion
+	cmd:p11_kit$secondaryArchSuffix = $portVersion
+	cmd:trust$secondaryArchSuffix = $portVersion
+	lib:libp11_kit$secondaryArchSuffix = $portVersion
+	lib:p11_kit_proxy$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
-ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
-SECONDARY_ARCHITECTURES="x86"
+SECONDARY_ARCHITECTURES="?x86"

+libVersion="0.3.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	p11_kit$secondaryArchSuffix = $portVersion
 	cmd:p11_kit$secondaryArchSuffix = $portVersion
 	cmd:trust$secondaryArchSuffix = $portVersion
-	lib:libp11_kit$secondaryArchSuffix = $portVersion
+	lib:libp11_kit$secondaryArchSuffix = $libVersionCompat
-	lib:p11_kit_proxy$secondaryArchSuffix = $portVersion
+	lib:p11_kit_proxy$secondaryArchSuffix = $libVersionCompat
	"
```

sfanxiang:
@fbrosson
We can't merge yet. I just fixed another segfault (haven't pushed) but there's still problem (`trust extract' cannot write to file), and keep in mind that the library is largely untested because we haven't ported any modules.

fbrosson:
OK, no problem. I did not know you were working on fixing the segfaults. This is great news!
So, this is perfectly OK to keep the PR open.

Regarding the "`'trust extract' cannot write to file`" issue, this is probobably due to the filesystem being read-only where it is trying to write.(Many directories are readonly.)
Some directories are, however, read-write. One (or more) of the following directories may be appropriate for what `trust extract` is likely to have to write:

- [ ] B_SYSTEM_TEMP_DIRECTORY
- [ ] B_USER_VAR_DIRECTORY
- [ ] B_SYSTEM_VAR_DIRECTORY
- [ ] B_SYSTEM_CACHE_DIRECTORY
- [ ] B_USER_CACHE_DIRECTORY
- [ ] B_USER_VAR_DIRECTORY

sfanxiang:
@fbrosson OK, but I was asking it to write to the current directory :)

app-crypt/p11-kit/p11_kit-0.23.9.recipe
sfanxiang: What is libVersion? If it means "the version of the library" then it's 0.23.9, right?

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,118 @@
...
+	$libDir/pkcs11/p11-kit-client.so \
+	$libDir/pkcs11/p11-kit-trust.so
+
+BUILD()
+{
+	certsInstallDir="$portPackageLinksDir/ca_root_certificates"
+	certsDir="$certsInstallDir/$relativeDataRootDir/ssl"
+	runConfigure ./configure \
+		LDFLAGS="-lnetwork" \
+		--with-trust-paths="${certsDir}/CARootCertificates.pem" \
```
sfanxiang: I'm still not sure with this one. On my linux, the trust path is at `/etc/ssl/certs/ca-certificates.crt`, which is generated when `update-ca-trust` is run.

So what's the equivalent for `update-ca-trust` on Haiku?

app-crypt/p11-kit/p11_kit-0.23.9.recipe
sfanxiang: Alright I was wrong.

app-crypt/p11-kit/p11_kit-0.23.9.recipe
fbrosson: On Haiku, [haikuports/haiku-data/ca_root_certificates](https://github.com/haikuports/haikuports/tree/master/haiku-data/ca_root_certificates) does this:
```
INSTALL()
{
	installDir="$dataRootDir/ssl/"
	mkdir -p "$installDir"
	cp ca-bundle-*.crt "$installDir/CARootCertificates.pem"
}
```
which lands in `/boot/system/data/ssl/CARootCertificates.pem`

(But this is readonly.)

sfanxiang: Thanks. So is this recipe OK then?

fbrosson:
Congratulations @sfanxiang ! You fixed the crashes, and on all architectures!
Both `p11-kit-x86 list-modules` (on x86_gcc2) and `p11-kit list-modules` (on x86_64) produce the correct output
```
p11-kit-trust: p11-kit-trust.so
    library-description: PKCS#11 Kit Trust Module
    library-manufacturer: PKCS#11 Kit
    library-version: 0.23
    token: CARootCertificates.pem
        manufacturer: PKCS#11 Kit
        model: p11-kit-trust
        serial-number: 1
        hardware-version: 0.23
        flags:
               write-protected
               token-initialized
```

BTW, before the merge, shall we add the commandBinDir "trick" in order to get rid of the `-x86` suffix on x86_gcc2?

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,117 @@
...
+COPYRIGHT="2017 Red Hat, Inc."
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/p11-glue/p11-kit/releases/download/$portVersion/p11-kit-$portVersion.tar.gz"
+CHECKSUM_SHA256="e1c1649c335107a8d33cf3762eb7f57b2d0681f0c7d8353627293a58d6b4db63"
+SOURCE_DIR="p11-kit-$portVersion"
+PATCHES="p11_kit-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
```
fbrosson:
```diff
-ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+SECONDARY_ARCHITECTURES="x86"
 
+commandBinDir=$binDir
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]
+then
+	commandBinDir=$prefix/bin
+	commandSuffix=
+fi
+
```

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,117 @@
...
+
+libVersion="0.3.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	p11_kit$secondaryArchSuffix = $portVersion
+	cmd:p11_kit$secondaryArchSuffix = $portVersion
+	cmd:trust$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES="
 	p11_kit$secondaryArchSuffix = $portVersion
-	cmd:p11_kit$secondaryArchSuffix = $portVersion
+	cmd:p11_kit$commandSuffix = $portVersion
-	cmd:trust$secondaryArchSuffix = $portVersion
+	cmd:trust$commandSuffix = $portVersion
 	lib:libp11_kit$secondaryArchSuffix = $libVersionCompat
 	lib:p11_kit_proxy$secondaryArchSuffix = $libVersionCompat
 	"
```

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,117 @@
...
+	settings/pkcs11/pkcs11.conf template data/pkcs11/pkcs11.conf.example
+	"
+
+GLOBAL_WRITABLE_FILES="
+	settings/pkcs11/pkcs11.conf keep-old
+	"
+
+defineDebugInfoPackage p11_kit$secondaryArchSuffix \
+	$binDir/p11-kit \
+	$binDir/trust \
```
fbrosson:
```diff
 defineDebugInfoPackage p11_kit$secondaryArchSuffix \
-	$binDir/p11-kit \
+	$commandBinDir/p11-kit \
-	$binDir/trust \
+	$commandBinDir/trust \
```

sfanxiang:
@fbrosson Like this?
``` sh
	commandBinDir=$binDir
else
	commandBinDir=$prefix/bin
fi
```

BTW, does `x86_gcc2` exist for historical reasons?

app-crypt/p11-kit/p11_kit-0.23.9.recipe
```diff
@@ -0,0 +1,117 @@
...
+	$libDir/p11-kit/p11-kit-remote \
+	$libDir/p11-kit/p11-kit-server \
+	$libDir/pkcs11/p11-kit-client.so \
+	$libDir/pkcs11/p11-kit-trust.so
+
+BUILD()
+{
+	certsInstallDir="$portPackageLinksDir/ca_root_certificates"
+	certsDir="$certsInstallDir/$relativeDataRootDir/ssl"
+	runConfigure ./configure \
```
fbrosson:
```diff
-	runConfigure ./configure \
+	runConfigure --omit-dirs binDir ./configure --bindir $commandBinDir \
```

fbrosson:
Oops, I saw you question only after I had finished my last review.

> BTW, does x86_gcc2 exist for historical reasons?

Yep. Haiku x86_gcc2 is compatible with BeOS, at both binary and source levels. Any (or most?) program compiled for BeOS can run on Haiku without any modification. So if you have an app but don't have its source code, you can still use it out of the box. This has a "small" cost: on x86_gcc2 we need to compile the system and the x86_gcc2 primary arch packages with a modified version of gcc 2.95.3.

The x86 secondary arch (available on x86_gcc2 hybrid), however, allows to build and use programs that would not build with gcc 2.95.3 (which we call gcc2.)

PR #2055, opened 2018-01-04
check: unsplit check_common.

fbrosson:
~Until now, check_86, which builds fine on x86_gcc2, had a runtime dependency on check_common, which is the arch-independent subpackage which can only be built when check is built on x86_gcc2 primary arch.
But since the latter no longuer builds for check 0.11.0, there was no way to obtain check_common on x86_gcc2 and meet check_x86's requirement when installing it.
The workaround for this issue is to simply stop creating the check_common subpackage when building on x86_gcc2. All contents of check_common are now part of check_x86.~

This is required because check 0.11.0 no longer builds on x86_gcc2 primary arch, so the arch-independent check_common could not be built on x86_gcc2, hence preventing the installation of check_x86.

### DIFF
--- a/dev-libs/check/check-0.11.0.recipe
+++ b/dev-libs/check/check-0.11.0.recipe
@@ -9,7 +9,7 @@ HOMEPAGE="https://libcheck.github.io/check/"
 COPYRIGHT="2001-2015 Arien Malec, Branden Archer, Chris Pickett, Fredrik \
 Hugosson, and Robert Lemmen."
 LICENSE="GNU LGPL v2.1"
-REVISION="1"
+REVISION="2"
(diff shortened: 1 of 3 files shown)


fbrosson:
I've "implemented" this because this is what we were suggeting in #2034, but I think it would be better, after all, to just "unsplit" check_common for all architectures. I had forgotten that the check_common was really small. It's only half the size of the base package, on x86_64. Moreover, it only contains documentation. When I had split check a long time ago, I though it was a good idea, but it was not.

So, shall I update this PR to just revert the split and get rid of this "check_common" for all architectures?
Reminder: I'll also make the same changes for the older versions of check, `0.9.13` and `0.9.14`.
BTW, we had kept `0.9.13` because `0.9.14` was not passing some tests.

PR #2059, opened 2018-01-04
Add gpart

localanu:


sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,50 @@
...
+#BUILD_PREREQUIRES
+
+#defineDebugInfoPackage
+
+BUILD()
+{
+./autogen.sh
+CFLAGS="-D_BSD_SOURCE" ./configure --prefix=$prefix
+make
+}
```
tts2k:
``` diff
 BUILD()
 {
-./autogen.sh
-CFLAGS="-D_BSD_SOURCE" ./configure --prefix=$prefix
-make
+	./autogen.sh
+	CFLAGS="-D_BSD_SOURCE" ./configure --prefix=$prefix
+	make
 }

File: sys-block/gpart/gpart-0.3.recipe [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+INSTALL()
+{
+make install
+}
+#TEST()
```
tts2k:
``` diff
 INSTALL()
{
-make install
+	make install
}
 #TEST()

File: sys-block/gpart/gpart-0.3.recipe [hunk 3]
```diff
@@ -0,0 +1,50 @@
...
+SUMMARY="A small tool which tries to guess what partitions are on a PC type"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are on a PC type, \
+MBR-partitioned hard disk in case the primary partition table was damaged."
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999, Michail Brzitwa"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$COMMIT_ID.tar.gz"
+#CHECKSUM_SHA256
+CHECKSUM_SHA256="c492b7dc677e4c6af197f9598ba815a4e14a03578ee8a963430529767bc6c947"
```
tts2k:
``` diff
-#CHECKSUM_SHA256
 CHECKSUM_SHA256="c492b7dc677e4c6af197f9598ba815a4e14a03578ee8a963430529767bc6c947"

File: sys-block/gpart/gpart-0.3.recipe [hunk 4]
```diff
@@ -0,0 +1,50 @@
...
+REQUIRES="
+	haiku
+	"
+#PROVIDES_devel
+#REQUIRES_devel
+
+BUILD_REQUIRES="
+	haiku_devel
+	autoconf
+	automake
```
tts2k:
``` diff
-	autoconf
-	automake
+	cmd:autoconf
+	cmd:automake

Comment (tts2k) [reply sys-block/gpart/gpart-0.3.recipe]: "cmd:" must be added

File: sys-block/gpart/gpart-0.3.recipe (hunk 3 above)
Comment (tts2k) [reply sys-block/gpart/gpart-0.3.recipe]: Remove unnescessary line

File: sys-block/gpart/gpart-0.3.recipe [hunk 5]
```diff
@@ -0,0 +1,49 @@
+COMMIT_ID="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
```
fbrosson:
```diff
-COMMIT_ID="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+SUMMARY="A small tool which tries to guess what partitions are on a PC type"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are on a PC type, \
+MBR-partitioned hard disk in case the primary partition table was damaged."
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999, Michail Brzitwa"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$COMMIT_ID.tar.gz"
```
fbrosson:
```diff
 REVISION="1"
+srcGitRev="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
-SOURCE_URI="https://github.com/baruch/gpart/archive/$COMMIT_ID.tar.gz"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$srcGitRev.tar.gz"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+CHECKSUM_SHA256="c492b7dc677e4c6af197f9598ba815a4e14a03578ee8a963430529767bc6c947"
+SOURCE_FILENAME="gpart-$portVersion.tar.gz"
+SOURCE_DIR="gpart-$COMMIT_ID"
```
fbrosson:
```diff
-SOURCE_FILENAME="gpart-$portVersion.tar.gz"
+SOURCE_FILENAME="gpart-$portVersion-$srcGitRev.tar.gz"
-SOURCE_DIR="gpart-$COMMIT_ID"
+SOURCE_DIR="gpart-$srcGitRev"
 PATCHES="gpart-$portVersion.patchset"
-#ADDITIONAL_FILES
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+PATCHES="gpart-$portVersion.patchset"
+#ADDITIONAL_FILES
+
+ARCHITECTURES="x86 x86_64"
+#SECONDARY_ARCHITECTURES
```
fbrosson:
```diff
-ARCHITECTURES="x86 x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 x86_64"
-#SECONDARY_ARCHITECTURES

 PROVIDES="
+	gpart = $portVersion
 	cmd:gpart
 	"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+PROVIDES="
+	cmd:gpart
+	"
+REQUIRES="
+	haiku
+	"
+#PROVIDES_devel
+#REQUIRES_devel
```
fbrosson:
```diff
-#PROVIDES_devel
-#REQUIRES_devel
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+BUILD_REQUIRES="
+	haiku_devel
+	cmd:autoconf
+	cmd:automake
+	cmd:make
+	cmd:gcc
+	"
+#BUILD_PREREQUIRES
```
fbrosson:
```diff
 BUILD_REQUIRES="
 	haiku_devel
+ 	"
+BUILD_PREREQUIRES
 	cmd:autoconf
 	cmd:automake
- 	cmd:make
 	cmd:gcc
+	cmd:make
 	"
-#BUILD_PREREQUIRES
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+#defineDebugInfoPackage
```
fbrosson:
```diff
-#defineDebugInfoPackage
+defineDebugInfoPackage gpart \
+	$binDir/gpart
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+BUILD()
+{
+	./autogen.sh
+	CFLAGS="-D_BSD_SOURCE" ./configure --prefix=$prefix
```
fbrosson:
```diff
 BUILD()
 {
 	./autogen.sh
-	CFLAGS="-D_BSD_SOURCE" ./configure --prefix=$prefix
+	CFLAGS="-D_BSD_SOURCE" runConfigure ./configure
- 	make
+ 	make $jobArgs
 }
+
 INSTALL()
 {
 	make install
 }
-#TEST()
```

sys-block/gpart/patches/gpart-0.3.patchset
```diff
@@ -0,0 +1,91 @@
...
++	ioctl(d->d_fd, B_GET_GEOMETRY, & dg);
++	g -> d_h = dg.head_count;
++	g -> d_c = dg.cylinder_count;
++	g -> d_s = dg.sectors_per_track * dg.cylinder_count;
++}
++#endif
++
+ #else
+-#error Only Linux and FreeBSD supported
++#error Only Haiku,Linux,and FreeBSD supported
```
fbrosson:
```diff
-+#error Only Haiku,Linux,and FreeBSD supported
++#error Only Linux, FreeBSD and Haiku supported
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
+COMMIT_ID="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
+SUMMARY="A small tool which tries to guess what partitions are on a PC type"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are on a PC type, \
+MBR-partitioned hard disk in case the primary partition table was damaged."
```
fbrosson:
```diff
-DESCRIPTION="Gpart is a small tool which tries to guess what partitions are on a PC type, \
-MBR-partitioned hard disk in case the primary partition table was damaged."
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are \
+on a PC type, MBR-partitioned hard disk in case the primary partition table \
+was damaged."
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999, Michail Brzitwa"
```
fbrosson:
```diff
-COPYRIGHT="1999, Michail Brzitwa"
+COPYRIGHT="1999-2017 Michail Brzitwa"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
+COMMIT_ID="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
+SUMMARY="A small tool which tries to guess what partitions are on a PC type"
```
fbrosson:
```diff
-SUMMARY="A small tool which tries to guess what partitions are on a PC type" 
+SUMMARY="A tool for recovering MBR partition tables in hard disks"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,46 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="c492b7dc677e4c6af197f9598ba815a4e14a03578ee8a963430529767bc6c947"
+SOURCE_FILENAME="gpart-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="gpart-$srcGitRev"
+PATCHES="gpart-$portVersion.patchset"
+
+
```
fbrosson: Please leave only one blank line between
```
PATCHES="gpart-$portVersion.patchset"
```
and
```
 ARCHITECTURES="x86_gcc2 ?x86 x86_64"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="A tool for recovering MBR partitions tables in hard disk"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are\
+on a PC type, MBR-partitioned hard disk in case the primary partition table was damaged."
```
fbrosson:
```diff
-DESCRIPTION="Gpart is a small tool which tries to guess what partitions are\
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are \
-on a PC type, MBR-partitioned hard disk in case the primary partition table was damaged."
+on a PC type, MBR-partitioned hard disk in case the primary partition table \
+was damaged."
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,46 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc
+	cmd:make
+	"
+
+defineDebugInfoPackage gpart \
+	$binDir\gpart
```
fbrosson:
```diff
 defineDebugInfoPackage gpart \
-	$binDir\gpart
+	$binDir/gpart
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,46 @@
...
+
+BUILD()
+{
+	./autogen.sh
+	CFLAGS="-D_BSD_SOURCE" runConfigure ./configure
+	make $jobArgs
+}
+INSTALL()
```
fbrosson: Please insert one blank line before
```
INSTALL()
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,47 @@
...
+
+PROVIDES="
+	gpart = $portVersion
+	cmd:gpart = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
```
fbrosson:
```diff
 BUILD_REQUIRES="
 	haiku_devel
+	"
+BUILD_PREREQUIRES="
 	cmd:autoconf
 	cmd:automake
 	cmd:gcc
 	cmd:make
 	"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,50 @@
...
+	cmd:gpart = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BULID_PREREQUIRES="
+	haiku_devel
+	"
+
```
fbrosson: We prefer not to have any blank line between `BUILD_REQUIRES` and `BUILD_PREREQUIRES`.

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,50 @@
+SUMMARY="A tool for recovering MBR partitions tables in hard disks"
```
fbrosson:
```diff
-SUMMARY="A tool for recovering MBR partitions tables in hard disks"
+SUMMARY="A tool for recovering MBR partition tables in hard disks"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+	cmd:gpart = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BULID_PREREQUIRES="
+	haiku_devel
+	"
+BUILD_REQUIRES="
```
fbrosson: You swaped `BUILD_REQUIRES` and `BUILD_PREREQUIRES`. If you look at [my previous comment](https://github.com/haikuports/haikuports/pull/2059/#discussion_r160011209), `BUILD_PREREQUIRES` comes _after_ `BUILD_REQUIRES`, not _before_.

Here's the new diff to correct this:
```diff
-BUILD_PREREQUIRES="
+BUILD_REQUIRES="
 	haiku_devel
 	"
-BUILD_REQUIRES="
+BUILD_PREREQUIRES="
 	cmd:autoconf
 	cmd:automake
 	cmd:gcc
 	cmd:make
 	"
```

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+PROVIDES="
+	gpart = $portVersion
+	cmd:gpart = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BULID_REQUIRES="
```
fbrosson:
```diff
-BULID_REQUIRES="
+BUILD_REQUIRES="
```

scottmc:
Looking for a review of the included patch on this one.

sys-block/gpart/patches/gpart-0.3.patchset
```diff
@@ -0,0 +1,91 @@
...
+diff --git a/src/gm_btrfs.c b/src/gm_btrfs.c
+index cdf0a36..155bb1d 100644
+--- a/src/gm_btrfs.c
++++ b/src/gm_btrfs.c
+@@ -12,7 +12,7 @@
+  * Created:   20.11.2015 <mwilck@arcor.de>
+  *
+  */
+-
++#define _BSD_SOURCE
```
alaviss: There should be no need for this. Your recipe already contains `-D_BSD_SOURCE` in `CFLAGS`

```diff
-+#define _BSD_SOURCE

File: sys-block/gpart/patches/gpart-0.3.patchset [hunk 27]
```diff
@@ -0,0 +1,91 @@
...
+diff --git a/src/gpart.h b/src/gpart.h
+index e6e213c..2e60a54 100644
+--- a/src/gpart.h
++++ b/src/gpart.h
+@@ -17,8 +17,11 @@
+ #ifndef _GPART_H
+ #define _GPART_H
+ 
++#define __BSD_SOURCE
++
```
alaviss:
```diff
-+#define _BSD_SOURCE
-+

File: sys-block/gpart/patches/gpart-0.3.patchset [hunk 28]
```diff
@@ -0,0 +1,91 @@
...
+diff --git a/src/l64seek.h b/src/l64seek.h
+index 5e8dc4b..cb7969a 100644
+--- a/src/l64seek.h
++++ b/src/l64seek.h
+@@ -27,7 +27,7 @@
+  * offsets.
+  */
+ 
+-typedef loff_t off64_t;
++typedef off_t off64_t;
```
alaviss: Please don't do this. For easier upstreaming, wrap this under a conditional instead

```diff
--typedef loff_t off64_t;
-+typedef off_t off64_t;
++#ifdef __HAIKU__
++typedef off_t off64_t;
++#else
++typedef loff_t off64_t;
++#endif

File: sys-block/gpart/patches/gpart-0.3.patchset [hunk 29]
```diff
@@ -0,0 +1,91 @@
...
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++	device_geometry dg;
++	ioctl(d->d_fd, B_GET_GEOMETRY, & dg);
++	g -> d_h = dg.head_count;
++	g -> d_c = dg.cylinder_count;
++	g -> d_s = dg.sectors_per_track * dg.cylinder_count;
++}
++#endif
```
alaviss: Why is there an `#endif` here? This would invaildate the `#else` branch below

sys-block/gpart/patches/gpart-0.3.patchset
```diff
@@ -0,0 +1,91 @@
...
+ 	g.d_nsecs = o / u;
+ 	g.d_c = g.d_nsecs / g.d_h / g.d_s;
+ }
++
++
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++	device_geometry dg;
++	ioctl(d->d_fd, B_GET_GEOMETRY, & dg);
```
alaviss: [`ioctl()`][0] can fail, please add code to handle that case

This could be a reference for you:
https://github.com/baruch/gpart/blob/23b651d1e9c6ae1454e5d3fe315d7d4e145661c0/src/disku.c#L54

[0]: http://pubs.opengroup.org/onlinepubs/009695399/functions/ioctl.html

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
+SUMMARY="A tool for recovering MBR partition tables in hard disks"
+DESCRIPTION="Gpart is a small tool which tries to guess what partitions are \
+on a PC type, MBR-partitioned hard disk in case the primary partition table \
+was damaged."
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999-2017 Michail Brzitwa"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
```
alaviss: Your recipe is version 0.3, but you're building from HEAD. Is there any reason not to use the [latest release tag][1]?

[1]: https://github.com/baruch/gpart/releases/tag/0.3

localanu: No,because contains necessary patch

alaviss: @fbrosson should this recipe version be `0.3~git` instead? Given that it's not using the [latest release tag] upstream

[0]: https://github.com/baruch/gpart/releases/tag/0.3

sys-block/gpart/patches/gpart-0.3.patchset
localanu: Need,error in building

sys-block/gpart/patches/gpart-0.3.patchset
localanu: Need,error in building

sys-block/gpart/patches/gpart-0.3.patchset
localanu: gm_bsddl.c:40:22: note: in expansion of macro 'le32'
  if ((dl->d_magic == le32(DISKMAGIC)) && (dl->d_magic2 == le32(DISKMAGIC))) {
                      ^
gpart.h:29:17: warning: implicit declaration of function 'htole16' [-Wimplicit-function-declaration]
 #define le16(x) htole16(x)

localanu: and another gpart.h error

alaviss: It shouldn't. If there's any error, you should edit the `Makefile` to make sure CFLAGS is used when compiling this file

sys-block/gpart/patches/gpart-0.3.patchset
localanu: Okay,remove it

sys-block/gpart/patches/gpart-0.3.patchset
localanu: wait,make file generated by autotools,not directly wirte

sys-block/gpart/gpart-0.3.recipe
```diff
@@ -0,0 +1,49 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:gcc
+	cmd:make
+	"
+
+defineDebugInfoPackage gpart \
+	$binDir/gpart
```
alaviss: On x86_gcc2, it appears to me that `gpart` was built without the `-g` flags. This means there won't be any (useful) debuginfos available in the binary, so this should be removed

```diff
-defineDebugInfoPackage gpart \
-	$binDir/gpart

File: sys-block/gpart/gpart-0.3.recipe [hunk 33]
```diff
@@ -0,0 +1,49 @@
...
+
+BUILD()
+{
+	./autogen.sh
```
alaviss: You should use `autoreconf -vfi` instead. The autogen script included would run configure once after it has done generating, and we should avoid running configure multiple times

sys-block/gpart/gpart-0.3.recipe
fbrosson: @alaviss, you had convinced me about the fact that an empty debuginfo package is worse than no debugpackage at all, but I wonder if it would not be, after all, better to keep these empty debuginfo packages and, as you suggested during that discussion, teach haikuporter how to trigger a debug build.
So, maybe we could keep the defineDebugInfoPackage in the hope that haikuporter gains this feature.
What do you think?

alaviss:
I've got an KDL running `gpart -v /dev/disk/scsi/1/0/0/raw` (this is my `/boot` drive) on x86_gcc2

```
PANIC: last transaction (7) still open!

Welcome to Kernel Debugging Land...
Thread 285 "daemon" running on CPU 0
stack trace for thread 285 "daemon"
    kernel stack: 0x814e5000 to 0x814e9000
      user stack: 0x7125c000 to 0x7129c000
frame               caller     <image>:function + offset
 0 814e8bc8 (+  32) 8014bad6   <kernel_x86> arch_debug_stack_trace + 0x12
 1 814e8be8 (+  16) 800a9013   <kernel_x86> stack_trace_trampoline(NULL) + 0x0b
 2 814e8bf8 (+  12) 8013db16   <kernel_x86> arch_debug_call_with_fault_handler + 0x1b
 3 814e8c04 (+  48) 800aaa9b   <kernel_x86> debug_call_with_fault_handler + 0x5f
 4 814e8c34 (+  64) 800a9227   <kernel_x86> kernel_debugger_loop(0x8018ff37 "PANIC: ", 0x801663e0 "last transaction (%ld) still open!
", 0x814e8ce0 "", int32: 0) + 0x20f
 5 814e8c74 (+  48) 800a95cb   <kernel_x86> kernel_debugger_internal(0x8018ff37 "PANIC: ", 0x801663e0 "last transaction (%ld) still open!
", 0x814e8ce0 "", int32: 0) + 0x77
 6 814e8ca4 (+  48) 800aae1a   <kernel_x86> panic + 0x3a
 7 814e8cd4 (+ 112) 80053bc8   <kernel_x86> cache_start_transaction + 0xc0
 8 814e8d44 (+  48) 81418939   <bfs> Journal<0x823add80>::Lock(Transaction*: 0x814e8e64, false) + 0x9d
 9 814e8d74 (+  48) 81418e5b   <bfs> Transaction<0x814e8e64>::Start(Volume*: 0x8221f900, int64: 2101582) + 0x3b
10 814e8da4 (+  96) 81412810   <bfs> Inode<0xddf60010>::WriteAt(Transaction&: 0x814e8e64, int64: 211214, 0x71296f2c "KERN: das_write(): request.Wait() returned: Invalid Argument
", 0x814e8f2c) + 0x190
11 814e8e04 (+ 112) 8142163d   <bfs> bfs_write(fs_volume*: 0x82388bc0, fs_vnode*: 0xddf2b668, 0xddf6a698, int64: 104930, 0x71296f2c, 0x814e8f2c) + 0xe1
12 814e8e74 (+  64) 800fbd24   <kernel_x86> file_write(file_descriptor*: 0xddf784c0, int64: 104930, 0x71296f2c, 0x814e8f2c) + 0x58
13 814e8eb4 (+  96) 800e6b74   <kernel_x86> common_user_io(int32: 3, int64: -1, 0x71296f2c, uint32: 0x3d (61), true) + 0x1a0
14 814e8f14 (+  48) 800e70b4   <kernel_x86> _user_write + 0x28
15 814e8f44 (+ 100) 801405ff   <kernel_x86> handle_syscall + 0xdc
user iframe at 0x814e8fa8 (end = 0x814e9000)
 eax 0x90          ebx 0x25ab19c      ecx 0x71296e3c  edx 0x620f2114
 esi 0x3d          edi 0x71299062     ebp 0x71296e78  esp 0x814e8fdc
 eip 0x620f2114 eflags 0x3202    user esp 0x71296e3c
 vector: 0x63, error code: 0x0
16 814e8fa8 (+   0) 620f2114   <commpage> commpage_syscall + 0x04
17 71296e78 (+ 112) 013e62a0   <_APP_> __tf12SyslogDaemon (nearest) + 0x364
18 71296ee8 (+8432) 013e6549   <_APP_> __tf12SyslogDaemon (nearest) + 0x60d
19 71298fd8 (+8288) 013e5bf6   <_APP_> SyslogDaemon<0x7245b008>::_Daemon(0x0) + 0xc6
20 7129b038 (+  48) 013e5c6f   <_APP_> SyslogDaemon<0x7245b008>::_DaemonThread(NULL) + 0x1f
21 7129b068 (+  48) 024fcf6b   <libroot.so> _get_next_team_info (nearest) + 0x5f
22 7129b098 (+   0) 620f2250   <commpage> commpage_thread_exit + 0x00

**Comment by pulkomandy** [comment]
On Sat, Jan 06, 2018 at 09:30:21AM +0000, fbrosson wrote:
> @alaviss, you had convinced me about the fact that an empty debuginfo package is worse than no debugpackage at all, but I wonder if it would not be, after all, better to keep these empty debuginfo packages and, as you suggested during that discussion, teach haikuporter how to trigger a debug build.
> So, maybe we could keep the defineDebugInfoPackage in the hope that haikuporter gains this feature.

This is not an haikuporter problem. Proper options should be passed to
the build system (./configure or other) to do a debug build. There is
no generic way to do this. So all recipes which need debug info should
be revisited to give the proper options to their build system to
generate debug informations.

Since we need to revisit the recipes to do this anyways, we can at the
same time add the appropriate defineDebugInfoPackage. There is no use
for keeping it, and actually it is harmful: we can't know if a recipe
was already handled to generate debug info, it looks like so, but in
fact it doesn't work. So I prefer to leave this out, then it is easy to
grep for recipes which don't generate debug infos yet.

**Comment by localanu** [comment]
Okay,now I have a lot of changes

File: sys-block/gpart/gpart-0.3~git.recipe [hunk 34]
```diff
@@ -0,0 +1,49 @@
...
+HOMEPAGE="https://github.com/baruch/gpart"
+COPYRIGHT="1999-2017 Michail Brzitwa"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="23b651d1e9c6ae1454e5d3fe315d7d4e145661c0"
+SOURCE_URI="https://github.com/baruch/gpart/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="c492b7dc677e4c6af197f9598ba815a4e14a03578ee8a963430529767bc6c947"
+SOURCE_FILENAME="gpart-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="gpart-$srcGitRev"
+#PATCHES="gpart-$portVersion.patchset"
```
fbrosson: If the patch is required then you should uncomment this line.
If it is not required then this line should be removed as well as the patch itself.

Please ignore this comment in case I am missing something.

localanu: Sorry,I forgot for uncomment patches

localanu:
Ugh, I forgot for uncommenting -_-

sys-block/gpart/gpart-0.3.recipe
mmlr: How would HaikuPorter do that? Making debug builds varies with each port, there isn't really a way to standardize or factor this out. The recipe needs to take care of this. If generally enabling debug builds is undesireable for some reason (performance for example), then there should just not be any debuginfo package. The recipe could also build the binaries twice and provide a debug build as a separate subpackage, but the user would then explicitly have to install that one. Making a debuginfo package from a separate debug build wouldn't work however, as the info wouldn't match the non-debug build.

fbrosson: I might be wrong, but I **_think_** that with a **_very dirty_** hack like asking haikuporter to define a `HAIKU_CFLAGS="-g"` and patching our gcc to examine HAIKU_CFLAGS it would be possible to have non fake debuginfo packages, and this would probably work for many recipes.
There might even be less ugly solutions.

But @alaviss and @pulkomandy have already convinced me that it is not a good idea to have fake debuginfo packages.

BTW, @mmlr, I'm just replying here because you asked how I would do that. But I don't even plan to try that :)

sys-block/gpart/patches/gpart-0.3~git.patchset
```diff
@@ -0,0 +1,74 @@
...
+ 
+ static void geometry_from_num_sectors(struct disk_geom *g, uint64_t nsects)
+@@ -95,8 +99,20 @@ static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
+ 	g.d_nsecs = o / u;
+ 	g.d_c = g.d_nsecs / g.d_h / g.d_s;
+ }
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++device_geometry dg;
```
alaviss:
```diff
-+device_geometry dg;
++	device_geometry dg;

File: sys-block/gpart/patches/gpart-0.3~git.patchset [hunk 36]
```diff
@@ -0,0 +1,74 @@
...
++	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg)){
```
alaviss: [`ioctl()`][0] returns -1 on error, and any value other than -1 indicate success

[0]: http://pubs.opengroup.org/onlinepubs/009695399/functions/ioctl.html#tag_03_282_04

sys-block/gpart/patches/gpart-0.3~git.patchset
```diff
@@ -0,0 +1,74 @@
...
++		g -> d_h = dg.head_count;
++		g -> d_c = dg.cylinder_count;
++		g -> d_s = dg.sectors_per_track * dg.cylinder_count;
++	}
++	else
++	pr(FATAL, EM_IOCTLFAILED, "B_GET_GEOMETRY", strerror(errno));
```
alaviss:
```diff
++	pr(FATAL, EM_IOCTLFAILED, "B_GET_GEOMETRY", strerror(errno));
++		pr(FATAL, EM_IOCTLFAILED, "B_GET_GEOMETRY", strerror(errno));

File: sys-block/gpart/gpart-0.3~git.recipe [hunk 38]
```diff
@@ -0,0 +1,49 @@
...
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
```
alaviss: Missing `cmd:autoreconf`

sys-block/gpart/gpart-0.3~git.recipe
```diff
@@ -0,0 +1,49 @@
...
+	cmd:gcc
+	cmd:make
+	"
+
+defineDebugInfoPackage gpart \
+	$binDir/gpart
```
alaviss: As mentioned earlier, this package is not built with `-g`, thus there won't be any debuginfo to extract

sys-block/gpart/patches/gpart-0.3~git.patchset
```diff
@@ -0,0 +1,74 @@
...
+ static void geometry_from_num_sectors(struct disk_geom *g, uint64_t nsects)
+@@ -95,8 +99,20 @@ static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
+ 	g.d_nsecs = o / u;
+ 	g.d_c = g.d_nsecs / g.d_h / g.d_s;
+ }
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++device_geometry dg;
++	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg) == 0 ){
```
alaviss: I think I should be more clear. [`ioctl()`][0] would return -1 on error, and any other value indicate success. This means that you should do this instead:

```diff
-+	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg) == 0 ){
++	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg) != -1){
```
[0]: http://pubs.opengroup.org/onlinepubs/009695399/functions/ioctl.html#tag_03_282_04

sys-block/gpart/gpart-0.3.recipe
alaviss: @fbrosson we don't have to patch gcc for that. Simply appending the flags to our CFLAGS is fine enough.

@mmlr There are two kinds of debug builds:
- The one where you get a binary with all debugging logic on (ie. `ASSERT()`): For this, it would have to be done by the port as different build system have different ways to enable debugging logic
- The one where you only have good backtrace and step-by-step debugging: For this all you need is to make the compiler generate debugging infomations, mainly by appending `-g` to CFLAGS.

Gentoo has an [excellent write up][0] on this topic, and I would recommend checking it out.

[0]: https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Backtraces

mmlr: Yes, I know about the different debug flags. Also our `defineDebugInfoPackage` basically is what `splitdebug` does in gentoo.

But messing with compiler flags is still just not the job of HaikuPorter, it is the job of the recipe. Only the individual recipe really knows / should know what the software in question needs. Maybe the software isn't even compiled with a compiler that understands certain flags, or it isn't compiled at all.

Having such setup done locally by the recipe is just a lot more obvious and can be tailored to the actual situation.

sys-block/gpart/patches/gpart-0.3~git.patchset
```diff
@@ -0,0 +1,78 @@
...
+ 
+ static void geometry_from_num_sectors(struct disk_geom *g, uint64_t nsects)
+@@ -95,8 +99,20 @@ static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
+ 	g.d_nsecs = o / u;
+ 	g.d_c = g.d_nsecs / g.d_h / g.d_s;
+ }
++#elif defined(__HAIKU__)
++static void os_disk_geometry(disk_desc *d, struct disk_geom *g)
++{
++ device_geometry dg;
```
alaviss:
```diff
-+ device_geometry dg;
++ 	device_geometry dg;

File: sys-block/gpart/patches/gpart-0.3~git.patchset [hunk 42]
```diff
@@ -0,0 +1,78 @@
...
++	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg) != 
+-1 ){
```
alaviss: I believe you've edited this patch by hand with `nano`. To prevent nano from wrapping lines, use `nano -w`

```diff
-+	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg) != 
--1 ){
++	if (ioctl(d->d_fd, B_GET_GEOMETRY, & dg) != -1 ){

File: sys-block/gpart/patches/gpart-0.3~git.patchset [hunk 43]
```diff
@@ -0,0 +1,78 @@
...
++		g -> d_h = dg.head_count;
++		g -> d_c = dg.cylinder_count;
++		g -> d_s = dg.sectors_per_track * dg.cylinder_count;
++	}
++	else
++	 pr(FATAL, EM_IOCTLFAILED, 
+"B_GET_GEOMETRY", 
+strerror(errno));
```
alaviss:
```diff
-+	 pr(FATAL, EM_IOCTLFAILED, 
-"B_GET_GEOMETRY", 
-strerror(errno));
++	pr(FATAL, EM_IOCTLFAILED, "B_GET_GEOMETRY", strerror(errno));

**Comment by alaviss** [comment]
Currently this recipe won't build. Please always double check before pushing changes

```
Applying patchset "/boot/home/haikuports/sys-block/gpart/patches/gpart-0.3~git.patchset" ...
error: corrupt patch at line 32
error: could not build fake ancestor
Command '['git', 'am', '--ignore-whitespace', '-3', '--keep-cr', u'/boot/home/haikuports/sys-block/gpart/patches/gpart-0.3~git.patchset']' returned non-zero exit status 128

waddlesplash:
@alaviss is that KDL reproducible?

localanu:
@alaviss I have fix building error in patch

sys-block/gpart/gpart-0.3~git.recipe
fbrosson:
```diff
-defineDebugInfoPackage gpart \
-	$binDir/gpart
-
```
Exact. We don't want a debuginfo package here.

If we comment out the call to `defineDebugInfoPackage gpart $binDir/gpart` and rebuild the recipe, then install the newly built package and run:
```bash
readelf /bin/gpart | grep debug
```
you will see no output. This means `-g` is not being passed to the compiler, as @alaviss has already noticed.

fbrosson:
@localanu, I've added a :-1: to my previous message because I was not using `gpart` correctly!!!

Contrary to what I wrote, `gpart` **_does_** need to see formated partitions, otherwise it won't show any useful output. Well, at least when not given the `-v` switch.

So, I've just rebuilt your `gpart` and tried it on a (virtual) disk with a few formated partitions.
Here are the real partitions I have on that disk:
```
~> fdisk -l /dev/disk/ata/0/slave/raw 
Disk /dev/disk/ata/0/slave/raw: 0 B, 0 bytes, 0 sectors
Units: sectors of 1 * 512 = 512 bytes
Sector size (logical/physical): 512 bytes / 512 bytes
I/O size (minimum/optimal): 512 bytes / 512 bytes
Disklabel type: dos
Disk identifier: 0x00000000

Device                     Boot  Start    End Sectors  Size Id Type
/dev/disk/ata/0/slave/raw1          64 262207  262144  128M 83 Linux
/dev/disk/ata/0/slave/raw2      262208 393279  131072   64M 83 Linux
/dev/disk/ata/0/slave/raw3      393280 456767   63488   31M 83 Linux
/dev/disk/ata/0/slave/raw4      458816 522303   63488   31M 83 Linux
```
And here are the results of a few invocations of `gpart`:
* Without any switches:
```
~> gpart /dev/disk/ata/0/slave/raw 

Begin scan...

*** Fatal error: dev(/dev/disk/ata/0/slave/raw): read error near sector(524287): General system error.
```
* With `-v` :
```
~> gpart -v /dev/disk/ata/0/slave/raw 

dev(/dev/disk/ata/0/slave/raw) mss(512) chs(1/1/524288)(LBA) #s(0) size(0mb)
Primary partition(1)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 128mb #s(262144) s(64-262207)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/65)-(0/0/262208)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 00 00 00 00 04 00

Primary partition(2)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 64mb #s(131072) s(262208-393279)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/262209)-(0/0/393280)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 04 00 00 00 02 00

Primary partition(3)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 31mb #s(63488) s(393280-456767)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/393281)-(0/0/456768)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 06 00 00 F8 00 00

Primary partition(4)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 31mb #s(63488) s(458816-522303)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/458817)-(0/0/522304)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 07 00 00 F8 00 00


Begin scan...

*** Fatal error: dev(/dev/disk/ata/0/slave/raw): read error near sector(524287): General system error.
```
* With `-v` and `-d`:
```
~> gpart -v -d /dev/disk/ata/0/slave/raw 

dev(/dev/disk/ata/0/slave/raw) mss(512) chs(1/1/524288)(LBA) #s(0) size(0mb)
Primary partition(1)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 128mb #s(262144) s(64-262207)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/65)-(0/0/262208)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 00 00 00 00 04 00

Primary partition(2)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 64mb #s(131072) s(262208-393279)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/262209)-(0/0/393280)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 04 00 00 00 02 00

Primary partition(3)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 31mb #s(63488) s(393280-456767)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/393281)-(0/0/456768)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 06 00 00 F8 00 00

Primary partition(4)
   type: 131(0x83)(Linux ext2 filesystem)
   size: 31mb #s(63488) s(458816-522303)
   chs:  (1023/255/63)-(1023/255/63)d (0/0/458817)-(0/0/522304)r
   hex:  00 FF FF FF 83 FF FF FF 40 00 07 00 00 F8 00 00

```
But don't worry about this yet. @alaviss will review your patchset as well as the one for the concurrent PR, and I don't know which one will be prefered.
Thanks again and sorry for not having investigated enough your other PR, about wordgrinder's dictionaries. (I haven't forgotten about it, it's just that I'm a bit lazy... :)

scottmc:
can we get some eyes on this one and get it wrapped up soon?

Begasus:
Merged in: https://github.com/haikuports/haikuports/pull/9360

PR #2062, opened 2018-01-04
ProjectConceptor: create recipe and patchset

owenca:


scottmc:
Have you tried sending a pull request on the upstream repo yet?  Try that and if/when it is accepted update the recipe to use the latest source code.

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,50 @@
...
+	"
+
+BUILD()
+{
+	make -C src OBJ_DIR=objects $jobArgs
+}
+
+INSTALL()
+{
+	install -d $appsDir/Plugins
```
fbrosson: The Plugins directory should not be created in `$appsDir` but in a subdirectory, e.g. `$appsDir/ProjectConceptor/Plugins/` or `$appsDir/ProjectConceptor-Plugins` or, even better, imho, in `$libDir/ProjectConceptor/Plugins/`.
Obviously, if you choose `$appsDir/ProjectConceptor/Plugins/` you will need to install the main runtime as `$appsDir/ProjectConceptor/ProjectConceptor` instead of `$appsDir/ProjectConceptor`.

pulkomandy: $appsDir/ProjectConceptor/add-ons/ is my preferred solution (with the executable in $appsDir/ProjectConceptor/ProjectConceptor).

$libDir and $addOndDir could be used, but I think they are relevant only for things that can be shared by multiple applications (eg. translators). In this case, the plugins are specific to the application so they should stay in the application directory.

(and in BeOS/Haiku terminology, we say "add-on", not "plugin").

owenca: I have added a subdirectory ProjectConceptor and used "add-ons" in place of "Plugins" underneath it. Please see the updated commit below.

owenca:
The PR for the upstream update can be found [here](https://github.com/Paradoxianer/ProjectConceptor/pull/47).

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
fbrosson: I agree with some of your remarks, but:
* The terminology used inside an app should probably not be changed, because if ProjectConceptor's documentation mentions a "Plugins" directory then, imo, we should keep that name.
* There are several packages which create subdirectories in `/system/lib/` which are not used by any other recipes. I can think of a few: media-video/vlc, net-misc/openssh, app-misc/mc,
* `/system/lib/git/` is used by only 2 packages: git and git_flow.

For those reasons I would recommend using either `$appsDir/ProjectConceptor/Plugins/` or an better alternative imo: `$libDir/ProjectConceptor/Plugins/`. After all, if someone wanted to write a plugin for ProjectConceptor using another license, putting it in there would be more appropriate than putting it in `$appsDir/ProjectConceptor/add-ons/` or even `$appsDir/ProjectConceptor/Plugins/`.

Another advantage of using `$libDir/ProjectConceptor/Plugins/` is that the app can be installed directly in `$libDir`, which is nice because `/boot/system/apps` is in `$PATH`, so you can easily launch it from a terminal.

owenca: Ok. I have changed it back to "Plugins". :)

pulkomandy: > There are several packages which create subdirectories in /system/lib/ which are not used by any other recipes. I can think of a few: media-video/vlc, net-misc/openssh, app-misc/mc,

If these libs are private to the application, then these should be fixed. If the libs are meant to be public, but just happens to not be used by anyone else, then it is ok. For example this is the case for NetSurf's libraries: they have an API and are documented separately from NetSurf, but not many projects use them (yet).

> Another advantage of using $libDir/ProjectConceptor/Plugins/ is that the app can be installed directly in $libDir, which is nice because /boot/system/apps is in $PATH, so you can easily launch it from a terminal.

First, plug-ins should be stored in $addOnsDir, not in $libDir. This is where load_add_on will look for them by default. There is not much technical difference between applications, libs and add-ons in Haiku (you could make a file that behaves well in all 3 cases). But we try to separate them cleanly.

The definitions are as follows:
- app: something you can run
- library: something you can link your application to. Usually an hard dependency (you cannot build the app without the lib)
- add-on: something optional that extends the app. App is not linked to it, instead it is loaded dynamically with load_add_on or dlopen. Can be added to an app after the fact.

Then there is the distinction of "public" vs "private" add-ons and libraries
- public: multiple apps can use the same library (libSDL, libbe, ...) or add-on (translators, media codecs, ...)
- private: only one application can use them

Here we are clearly in the private case. You may be right that these should go in a subdir of $addOnDir. This is the case for example for screensavers. So let's go for that (but in $addOnDir, not $libDir).

fbrosson: Isn't `$addOnsDirs` reserved for stuff closely related to the system?
I consider that screensavers, as well as some very special exceptions like Qt5, are OK in `$addOnsDirs`.
But I still think that stuff like media-video/vlc, net-misc/openssh, app-misc/mc or app-crypt/johntheripper are fine with their subdirectories in `/system/lib/` (and should not be moved to `/system/add-ons/`.)

So I'd still recommend using `$libDir/ProjectConceptor/Plugins/` (and not `$addOnsDirs/ProjectConceptor/Plugins/`) for this recipe.

pulkomandy: I have a description of add-ons vs libraries in my message above. It is not a matter of public/private, but of dependency direction. An app *depends* on a library.  An app *uses* an add-on, but does not depend on it and can work without it.

UNIX apps (vlc, openssh, ...) does not usually have this distinction, and we have been lazy about patching them. But for native apps we should do the right thing.

fbrosson: I do admit that ProjectConceptor's plugins are shared objects, but I'm not saying that they are libraries that other apps can use like they use "classical" libraries.

I claim that, in my opinion, putting these shared objects in `/system/lib/ProjectConceptor/Plugins/` is more appropriate than putting them in `/system/add-ons/ProjectConceptor/`.

If you are suggesting that we should move `/system/lib/vlc/`, `/system/lib/openssh/`, `/system/lib/mc/` and `/system/lib/john/` to `/system/add-ons/` then I disagree.

fbrosson: OK, thanks for the explanation.
So, if I understand correctly, it is acceptable to create a subdirectory in `/system/lib/` and put ProjectConceptor's plugins there.

Would it also be acceptable to create one extra "Plugins" subdirectory in there, to have the plugins in `/system/lib/ProjectConceptor/Plugins/` ?
Thanks!

owenca: The app is hardwired to the directory name "Plugins" and assumes that it's in the same directory where the app is. So "ProjectConceptor/Plugins" and "ProjectConceptor/ProjectConceptor" is the only way for now?

owenca:
projectconceptor

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,49 @@
+SUMMARY="A Haiku diagramming application like Microsoft Visio"
+DESCRIPTION="ProjectConceptor is a modular application for Haiku and BeOS \
+designed to view, edit and process information which can be represented by a \
+graph structure."
+APP="ProjectConceptor"
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="2017 Paradoxianer"
+LICENSE="MIT"
```
alaviss: Looks like there's [more][0] than just MIT

- [MIT (no promotion)][1]
- [Zlib][2]
- [BSD (3-clause)][3]

[0]: https://github.com/Paradoxianer/ProjectConceptor/search?utf8=%E2%9C%93&q=copyright&type=
[1]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/ColorTools.h#L4
[2]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/tinyxml/tinyxmlparser.cpp#L3
[3]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/ToolBar/CalendarToolItem/LICENSE

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,49 @@
+SUMMARY="A Haiku diagramming application like Microsoft Visio"
+DESCRIPTION="ProjectConceptor is a modular application for Haiku and BeOS \
+designed to view, edit and process information which can be represented by a \
+graph structure."
+APP="ProjectConceptor"
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="2017 Paradoxianer"
```
alaviss: There're [a few more][0] copyright holders

- [Lee Thomason][1]
- [Be Incorporated][2]
- (could be a holder) [Al.V. Sarikov][3]

[0]: https://github.com/Paradoxianer/ProjectConceptor/search?utf8=%E2%9C%93&q=copyright&type=
[1]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/tinyxml/tinyxmlparser.cpp#L3
[2]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/ColorTools.h#L4
[3]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/ToolBar/CalendarToolItem/MonthWindow.cpp#L2

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+}
+
+INSTALL()
+{
+	install -d $appsDir/$APP/Plugins
+	install src/app/$APP $appsDir/$APP
+	addAppDeskbarSymlink $appsDir/$APP/$APP
+
+	cd src/plugins
+	install $(find * -mindepth 2 -type f -executable) $appsDir/$APP/Plugins
```
alaviss: Is there any reason to use `*` here?

```diff
-	install $(find * -mindepth 2 -type f -executable) $appsDir/$APP/Plugins
+	install -t $appsDir/$APP/Plugins $(find . -mindepth 2 -type f -executable)

Comment (owenca) [reply haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe]: If you use the -t format, the line will exceed 80 characters? I prefer the shorter format if they are equivalent. As to why using ‘*’ not ‘.’, compare the outputs of “find” and you will see why.

Comment (fbrosson) [reply haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe]: I think we can keep what we have right now:
```
	install $(find * -mindepth 2 -type f -executable) $appsDir/$APP/Plugins
```
Or maybe use a mix:
```
	install -t $appsDir/$APP/Plugins $(find * -mindepth 2 -type f -executable)
```

File: haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe [hunk 5]
```diff
@@ -0,0 +1,49 @@
+SUMMARY="A Haiku diagramming application like Microsoft Visio"
+DESCRIPTION="ProjectConceptor is a modular application for Haiku and BeOS \
+designed to view, edit and process information which can be represented by a \
+graph structure."
+APP="ProjectConceptor"
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="2017 Paradoxianer, 2000-2002 Lee Thomason, 1991-2000 Be Incorporated"
```
alaviss: Sorry, but that's not how we handle [`COPYRIGHT`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#copyright). We generally have [one copyright per line]((https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#copyright)).

```diff
-COPYRIGHT="2017 Paradoxianer, 2000-2002 Lee Thomason, 1991-2000 Be Incorporated"
+COPYRIGHT="2017 Paradoxianer
+	2000-2002 Lee Thomason
+	1991-2000 Be Incorporated"

Comment (owenca) [reply haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe]: And start the first item on its own line, according to the given [examples](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#examples). :)

Comment (alaviss) [reply haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe]: Some [metrics][0] from @fbrosson for you ;)

We should consider fixing those [examples][1] as well.

[0]: https://github.com/haikuports/haikuports/pull/1896#pullrequestreview-83766512
[1]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#examples

File: haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe [hunk 6]
```diff
@@ -0,0 +1,60 @@
...
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="
+	2017 Paradoxianer
+	2000-2002 Lee Thomason
+	1991-2000 Be Incorporated
+	"
+LICENSE="
+	BSD (3-clause)
+	MIT
+	MIT (no promotion)
```
alaviss: Re-reading the [source file][0], I don't think this would qualify as "MIT (no promotion)". This is more or less a custom license, so sorry for my earlier suggestion

[0]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/ColorTools.h#L4

owenca: No problem. I can certainly appreciate that you took the pain in trying to get the legal side right.

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
owenca: If some 1,600+ cases have multiple Copyright holders on the same line and only 500 or so use the multiple-line style, shouldn't we fix the guideline document instead? After all, it's just a style and not set in stone. If the devs/admins feel strongly about one way or the other, it can be enforced by `haikuporter`, like what's done to `LICENSE`.

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
alaviss: We still have to ship that license. Please add it as the Open Tracker License

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+APP="ProjectConceptor"
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="
+	2017 Paradoxianer
+	2000-2002 Lee Thomason
+	1991-2000 Be Incorporated
+	"
+LICENSE="
+	BSD (3-clause)
+	MIT
```
alaviss: We have to ship a copy of [Open Tracker License][0] also

[0]: https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/ColorTools.h#L4

owenca: I had already seen Open Tracker in the ColorTools code, but it's not a valid license as far as `haikuporter` is concerned.

alaviss: You can add custom licenses by making a folder in the same place as your recipe called "licenses". In there, add a file called "Open Tracker License" with the content of the said [license](https://github.com/Paradoxianer/ProjectConceptor/blob/dfef4b25fbd3dc620e144fef66a0879c4442a8f1/src/app/Tools/ColorTools.h#L4), then it would become vaild for `haikuporter`.

owenca: Thanks it works.

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,60 @@
+SUMMARY="A Haiku diagramming application like Microsoft Visio"
+DESCRIPTION="
+	ProjectConceptor is a modular application for Haiku and BeOS designed to \
+	view, edit and process information which can be represented by a graph \
+	structure.
+	"
```
fbrosson: @owenca, could you change DESCRIPTION to:
```
DESCRIPTION="ProjectConceptor is a modular application for Haiku and BeOS \
designed to view, edit and process information which can be represented by a \graph structure."
```

Yes, this is our fault. The [example with the build recipe for gawk](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#a-simple-example-using-an-outdated-style-for-description) was using an outdated style for `DESCRIPTION`, so I added an [example with the build recipe for which](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#a-simple-example-using-the-recommended-style-for-description) which uses the recommended style.

(I'll check which are the mostly used styles for `COPYRIGHT` and `LICENSE`, so I might ask you to change these too.)
Thanks, and sorry for making these picky requests...

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
fbrosson:
```diff
-COPYRIGHT="
-	2017 Paradoxianer
+COPYRIGHT="2017 Paradoxianer
 	2000-2002 Lee Thomason
-	1991-2000 Be Incorporated
-	"
+	1991-2000 Be Incorporated"
-LICENSE="
-	BSD (3-clause)
+LICENSE="BSD (3-clause)
 	MIT
 	Open Tracker License
-	Zlib
-	"
+	Zlib"
```
The recommended style for "multi-line" `COPYRIGHT` and `LICENSE` is to have the first author or license in the same line as the name of the parameter, right after the `="`, and the closing quotes right after the last author or license.

I just did some metrics on `COPYRIGHT` and `LICENSES`. The following tables **only** take into account thoses recipes which have **2 or more lines** in these variables.

Number of recipes with multi-line `COPYRIGHT` | 553
--- | ---
Number of recipes using the _recommended_ style for `COPYRIGHT` | 493
Number of recipes using the _alternate_ style for `COPYRIGHT` | 60


Number of recipes with multi-line `LICENSE` | 232
--- | ---
Number of recipes using the _recommended_ style for `LICENSE` | 170
Number of recipes using the _alternate_ style for `LICENSE` | 62

Since this PR is for adding a new recipe, it would be nice to follow the trend :)
Thanks!

Hmm, if you are away from your Haiku system then I can push these small changes for you if you agree :)

(I've already updated the wiki with an [example of a recipe having a multi-line](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#a-simple-example-using-the-recommended-style-for-a-multi-line-copyright) `COPYRIGHT`.)

owenca: Thanks! Will fix it soon.

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+APP="ProjectConceptor"
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="
+	2017 Paradoxianer
```
scottmc: 2017 Matthias “Paradoxer” Lindner

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
owenca: Maybe we can use this recipe as a simple example to avoid having two simple examples with the recommended style?

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
```diff
@@ -0,0 +1,54 @@
+SUMMARY="A Haiku diagramming application like Microsoft Visio"
+DESCRIPTION="ProjectConceptor is a modular application for Haiku and BeOS \
+designed to view, edit and process information which can be represented by a \
+graph structure."
+APP="ProjectConceptor"
+HOMEPAGE="https://github.com/Paradoxianer/$APP"
+COPYRIGHT="2017 Matthias “Paradoxer” Lindner
```
fbrosson:
```diff
-COPYRIGHT="2017 Matthias “Paradoxer” Lindner
+COPYRIGHT="2017 Matthias \“Paradoxer\” Lindner
```
(I guess you want to see the double quotes :)

fbrosson: @owenca, I saw the recent changes in the upstream repo, and this is great!

The latest stable release is 0.0.7, so I was wondering if jumping from 0.0.7 to 1.0.0 was not too "big" a jump, since the recent changes were mainly to make the app build on Haiku. Wouldn't a smaller jump, like 0.0.7 to 0.1.0 better reflect the recent changes?

Hmm, I saw that the project has an rsrc:
https://github.com/Paradoxianer/ProjectConceptor/blob/master/src/app/ProjectConceptor.rsrc

Would you like to convert it to an rdef? That would be terrific :)

owenca: I believe I had tried to convert the rsrc file to rdef but `rc` couldn't decompile it. It also happened to some of the other apps I worked. Will double check though.

owenca: Oops! It turns out there are a few more rsrc files. The ProjectConceptor.rsrc can be decompiled. Will try the others later and make a new PR to the upstream repo. We can either merge this one now or wait for the rdef files got merged upstream.

fbrosson: Cool!
For the merge, we can do whatever you wish. If you prefer to have this PR merged before you convert the rsrc files, that's perfectly OK.
But since there is no urgency at all, it is also perfectly OK to keep this PR open, and wait for your future upstream PRs to get merged.
It's up to you.

owenca: Let’s wait then.

Paradoxianer: Actually i think moving from 0.0.7 to 0.1.0 makes much more sense,  since its not even beta yet.. (still 3 bugs left .. since 5 years or so ;) ) i just dont have to much time left ot work on it  :) But i am happy to upstream any usefull change :) And i can make a release  for 0.1.0 if this helps :-D

Paradoxianer: btw. thanks for the recepie... this is something i didnt achived also i tried some weeks .(1h or 2h i have left every week)

owenca:
The rsrc files have been converted to rdef files and corresponding makefiles updated in this [PR](https://github.com/Paradoxianer/ProjectConceptor/pull/48).

owenca:
The recipe has been updated and it now refers to the latest upstream commit.

haiku-apps/projectconceptor/projectconceptor-1.0.0.recipe
owenca: I will study the code in the summer (if not sooner) and try to fix the bugs. :)

Paradoxianer: I alread told scott.. that my higher priority was / would be the index_server from haiku... actually it only need a good settings implemention then we could have a good full text search and thinks like thumbnails generation, mp3 methadata to attribute conversion and so on... 
i dont say projectconceptor isnt worth it .. but it would help tenthousden times more to get the index_server running... I am out of haiku since 6 Mont now.. but i didnt worked on index_server for 1 Year (did mostly testing) but if  you decide to work on this... i could try to help ;-)...

owenca: I'd love to work on index_server with you! I'm not sure if it can be done as a GCI task, although there are only a few days left. Otherwise, I will see if I have time during spring break.

PR #2063, opened 2018-01-04
sdl-rtf, revive old bep file

Begasus:
Builds ok for gcc5, didn't test on x86_64 ...

### DIFF
--- a/media-libs/sdl_rtf/sdl_rtf-0.1.0.recipe
+++ b/media-libs/sdl_rtf/sdl_rtf-0.1.0.recipe
@@ -1,33 +1,79 @@
-DESCRIPTION="SDL-rtf"
-HOMEPAGE="http://www.libsdl.org/projects/SDL_rtf"
-SOURCE_URI="http://www.libsdl.org/projects/SDL_rtf/release/SDL_rtf-0.1.0.tar.gz"
-CHECKSUM_MD5="fe36733167b5c89f128414f32612121a"
+SUMMARY="Display Rich Text Format (RTF) documents"
+DESCRIPTION="This is a sample library which allows you to display \
+Rich Text Format (RTF) documents in your SDL applications. It comes \
+with an example program showrtf which displays an RTF file using an \
+SDL_ttf font engine."
+HOMEPAGE="https://www.libsdl.org/projects/SDL_rtf/"
+COPYRIGHT="2003-2009 Sam Lantinga"
+LICENSE="GNU LGPL v2"
 REVISION="1"
-STATUS_HAIKU="stable"
-DEPEND="media-libs/sdl-ttf >= 2.0.9"
+SOURCE_URI="https://www.libsdl.org/projects/SDL_rtf/release/SDL_rtf-$portVersion.tar.gz"
+CHECKSUM_SHA256="3dc0274b666e28010908ced24844ca7d279e07b66f673c990d530d4ea94b757e"
+SOURCE_DIR="SDL_rtf-$portVersion"
+
+ARCHITECTURES="x86_gcc2 ?x86_64"
+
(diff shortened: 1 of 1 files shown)


media-libs/sdl_rtf/sdl_rtf-0.1.0.recipe
```diff
@@ -1,33 +1,80 @@
...
-DEPEND="media-libs/sdl-ttf >= 2.0.9"
+SOURCE_URI="https://www.libsdl.org/projects/SDL_rtf/release/SDL_rtf-$portVersion.tar.gz"
+CHECKSUM_SHA256="3dc0274b666e28010908ced24844ca7d279e07b66f673c990d530d4ea94b757e"
+SOURCE_DIR="SDL_rtf-$portVersion"
+
+ARCHITECTURES="x86_gcc2 ?x86_64"
+
+PROVIDES="
+	sdl_rtf = $portVersion
+	#lib:libSDL_rtf
```
extrowerk: Hmmm?

media-libs/sdl_rtf/sdl_rtf-0.1.0.recipe
```diff
@@ -1,33 +1,80 @@
...
+REQUIRES="
+	haiku
+	lib:libSDL
+	lib:libsdl_ttf
+	"
+
+PROVIDES_devel="
+	sdl_rtf = $portVersion
+	devel:libSDL_rtf
+	#devel:libSDL_rtf_0.1 = 0.0.0
```
extrowerk: Hmmm?

media-libs/sdl_rtf/sdl_rtf-0.1.0.recipe
```diff
@@ -1,33 +1,80 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	cmd:pkg_config
+	"
+
```
extrowerk: Too much newline.

media-libs/sdl_rtf/sdl_rtf-0.1.0.recipe
Begasus: Yeah, already mentioned it yesterday at IRC, libSDL_rtf is not in the regular package, even if I include it here, same for the develop package, libSDL_rtf_1.so.0.0.0 is not included there, even if it's declared in the PROVIDES ... any thoughts on this?

pulkomandy: What is the content of the package? Is there only a .a file? .so only in the devel package?

Begasus: In the main package there is only the so libs for libSDL_rtf_1.0
In the devel package there is only the so and a lib for libSDL_rtf (and the header)

pulkomandy: So the prepareInstalledDevelLibs is incorrect. It should be:

    prepareInstalledDevelLibs libSDL_rtf libSDL_rtf_1

This will make sure the lib is installed in lib and develop/lib under the right names.

Begasus: Even with that included and the lib enabled in lib: and devel: the content of the packages is the same, no libSDL_rtf in lib package and no libSDL_rtf-0.1 in the devel package ...

scottmc:
Also, doesn't this also create the demo cmd file showrtf?  Or is that only built if make test is run?  I forget as it was years ago when i wrote the original bep file.

PR #2067, opened 2018-01-05
command_not_found: fix recipe

tts2k:


fbrosson:
@tts2k, it seems there is another bug: we need to also `chmod +x` the post-install script.
The following changes fix this issue. Could you plaase apply them? Thanks!
```diff
 INSTALL()
 {
-	mkdir -p $binDir
+	install -d $binDir $postInstallDir
 	install -m 0755 haiku_cnf.py $binDir/command_not_found
-
-	mkdir -p $postInstallDir
+	install -m 0755 install_cnf.py $postInstallDir
 }
```

fbrosson:
Hmm, I've investigated a little bit. There are a few issues:
* The `install_cnf.py` script uses `/boot/common/bin/python` instead of just `/bin/python`. (`/boot/system/bin/python` would also work.)
* It seems `/boot/system/settings/etc/profile.d/command-not-found.sh` will only be installed by `/system/boot/post-install/install_cnf.py` after a reboot.
* Even if I install manually the previous file and the `options.json` file (in `/boot/home/config/settings/command-not-found/`), the package still does not work.

Now I'm hitting another issue:
```
No such file or directory: 'optional-packages/OptionalPackages'
```

Begasus:
Builds ok for primary arch on x86_gcc2, issues for the options.json file are the same when trying to run "command_not_found updatedb" (even after reboot)

app-shells/command_not_found/command_not_found-0.0.1~git.recipe
```diff
@@ -39,9 +43,9 @@ BUILD()
...
 INSTALL()
 {
-	mkdir -p $binDir
+	install -d $binDir $postInstallDir
 	install -m 0755 haiku_cnf.py $binDir/command_not_found
 
-	mkdir -p $postInstallDir
-	mv install_cnf.py $postInstallDir
+	install -m 0755 install_cnf.py $postInstallDir
+	chmod +x install_cnf.py
```
fbrosson: No need for the `chmod +x install_cnf.py`, since the `install -m 0755 install_cnf.py $postInstallDir` already takes care of that for us.

BTW, the chmod is not being applied on the installed script but on the one in the source directory, so it would have no effect in the hpkg.

Hmm, in case there was a TEST() then the chmod could have been required, but that's not the case here so you can safely drop the chmod (as long as you keep the `-m 0755` in the previous line.)

Begasus:
When trying to run the postinstall script (after a reboot) manually I'm getting this:

  /boot/system/boot/post-install> python install_cnf.py 
    File "install_cnf.py", line 18
      if not os.path.exists("/boot/common/etc/profile.d/command-not-found.sh")
                                                                             ^
  SyntaxError: invalid syntax

fbrosson:
Yep, that's also what I noticed, as indicated in point 1 of [this message](#issuecomment-355729841).
(I changed it manually in the build directory to see how far I could get.)

tts2k:
We need a python expert for this

tts2k:
@fbrosson is there anyway to write a default user configuration file ?

tts2k:
`command_not_found updatedb` should be working now if you copy default user configuration manually.
Post-install script isn't automatically executed. When i check the log it said:
```
KERN: package_daemon [610256015:   355] Volume::CommitTransactionHandler::_RunPostInstallScript(): running post-installation script "boot/post-install/install_cnf.py" of package command_not_found-0.0.1~git-1-x86_64.hpkg failed: 127 (errno: No such file or directory)
```
Also, there are some trailing whitespaces at https://github.com/haikuports/haikuports/pull/2067/files#diff-e4097288dd72a6d5a07d029a3907bcabR46 but TravisCI still let me pass

app-shells/command_not_found/command_not_found-0.0.1~git.recipe
```diff
@@ -2,15 +2,19 @@ SUMMARY="A tool to find not-installed applications"
...
 LICENSE="GNU GPL v2"
 REVISION="1"
-SOURCE_URI="git+https://github.com/jrabbit/haiku-cnf.git#87cd5d0c212e377cffa5d3e2a00c143fe751c6fb"
-PATCHES="command_not_found-0.0.1.patchset"
-DISABLE_SOURCE_PACKAGE="yes"
+srcGitRev="db1be8b57dedff781dc68caa46f403d88c5eedb9"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="a1ce4e6c8a7307c0d5c1e65a90b3d2a7ba9d16d9506a70aab1005129333d6121"
+SOURCE_FILENAME="haiku-cnf-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="haiku-cnf-$srcGitRev"
```
korli: Please remove this blank line.

app-shells/command_not_found/command_not_found-0.0.1~git.recipe
```diff
@@ -2,15 +2,19 @@ SUMMARY="A tool to find not-installed applications"
...
 
-ARCHITECTURES="?x86 ?x86_gcc2"
+PATCHES="command_not_found-0.0.1~git.patchset"
+
+ARCHITECTURES="?x86 ?x86_gcc2 x86_64"
```
alaviss: This is a pure python script. As such, it would work on any arch that has python.

```diff
-ARCHITECTURES="?x86 ?x86_gcc2 x86_64"
+ARCHITECTURES="any"

File: app-shells/command_not_found/command_not_found-0.0.1~git.recipe [hunk 4]
```diff
@@ -39,9 +43,8 @@ BUILD()
 
 INSTALL()
 {
-	mkdir -p $binDir
+	install -d $binDir $postInstallDir
```
alaviss: Trailing spaces. I wonder why Travis CI is not covering this.

```diff
-	install -d $binDir $postInstallDir	
+	install -d $binDir $postInstallDir

**Comment by Begasus** [comment]
Confirmed and seems to be working on x86_gcc2 :+1: 
>   command_not_found updatedb
  DEBUG:__main__:Running for first time
  DEBUG:__main__:Updating database

**Comment by tts2k** [comment]
Fixed post install script, all thanks to @alaviss for his help

**Comment by fbrosson** [comment]
`command_not_found-0.0.1~git-1-any.hpkg` installs `install_cnf.py` in `/system/boot/post-install/`, but line 8 of that post-install script has a reference to a file in `/boot/system/non-packaged/`.
Shouldn't we patch `install_cnf.py` with:
```diff
 if [ -e /system/bin/python ]; then
     command_not_found_handle(){
-        /boot/system/bin/python /boot/system/non-packaged/bin/command_not_found.py "$1"
+        /boot/system/bin/python /boot/system/bin/command_not_found "$1"
     }
```

tts2k:
@fbrosson You're right. It's funny how i forgot to test the main function of the port

app-shells/command_not_found/patches/command_not_found-0.0.1~git.patchset
```diff
@@ -0,0 +1,84 @@
...
+-#!/bin/env python
++#!/bin/python
+ import os
++import commands
+ 
+ hacks = """# command-not-found tomfoolery
+ if [ -e /system/bin/python ]; then
+     command_not_found_handle(){
+-        /boot/system/bin/python /boot/system/non-packaged/bin/command_not_found.py "$1"
++        /boot/system/bin/python /boot/system/bin/command_not_found "$1"
```
fbrosson: `finddir B_SYSTEM_BIN_DIRECTORY` currently returns `/boot/system/bin`, but this could change some day, in _theory_.
Hence this question: should we replace `/boot/system/bin` by the result of the call to finddir?
Or is this a bad idea?
I really don't know.
If we keep this hardcoded path, then we will need to update the patchset if we ever change `B_SYSTEM_BIN_DIRECTORY`.
On the other side, I guess such a change is unlikely to happen in the near future.

mmlr: It is _always_ a good idea to remove hardcoded paths. With the rational "it's not going to change anyway" we could just as well get rid of find_directory in the OS.

tts2k: Like you said:
>I guess such a change is unlikely to happen in the near future.
So i don't think we need to change it _now_
Beside, see my comment on GCI

mmlr: The problem with that is, that you are touching it now, so it's a reasonably easy and low overhead change to do now. If the hardcoded path is kept now, noone will feel responsible and bother to change it later.  At the very least add a TODO so that it can be found more easily.

Also, hardcoding the python path like that is a bad idea, as that will break not only if the system path changes, but also when the user installs python in a different install location (i.e. into home). That is exactly what the `/package-links` directory with its automatic links was made for.

fbrosson: Thanks @mmlr for your remark. I have to admit that you are right.

@tts2k, shall we keep this PR open until we reach a consensus?

Maybe we could even install the handler directly in `$settingsDir/profile.d/` (making it visible in `/boot/system/settings/profile.d/`) instead of installing a post-install script.

tts2k: @mmlr is right and i'm not going against him
But this recipe is gonna be rewritten sooner or later, because using post install script is definately not a good option for long-term.
@fbrosson just keep this PR for a few days, i'll try to rewrite it if i have free time

fbrosson: Thanks @mmlr for suggesting the use of the package links. That's a very good idea.

@tts2k, these package links work this way:
When haikuporter sees a `cmd:python` is in REQUIRES, it creates a `cmd~python` symlink in a special directory which command_not_found will see. Then whenever a script (from that package) needs to access any file provided by the package which provides `cmd:python`, it just needs to look in `$portPackageLinksDir/cmd~python/`, which will be pointing to the `$prefix` of the package providing `cmd:python`.
This `portPackageLinksDir` variable is defined in a temporary file called `wrapper-script` which haikuporter creates in the `haikuports/app-shells/command_not_found/work-0.0.1~git` when it starts building that package.

And to really see these package links, you can call `haikuporter --enter-chroot command_not_found` (or `haikuporter -E command_not_found`) and then go to `/packages/command_not_found-0.0.1~git-1/ and you will see the `cmd~python` symlink to a directory. To exit the chroot you just need to type exit. Sometimes you need to type this twice.

This package links stuff is a bit complex to use, but this is one of the key features of the packaging.

tts2k:
Rewrote recipe. Post-install script isn't needed anymore

PR #2070, opened 2018-01-05
mesa: GLView: Don't unconditionally unlock GL context

i80and:
Addresses https://dev.haiku-os.org/ticket/13191

I'm not 100% confident this is the right solution; in particular, I don't know what guarantees are desired when dealing with multiple threads. Is `IsLocked()` even the right tool here?

### DIFF
--- a/sys-libs/mesa/patches/mesa-17.1.7.patchset
+++ b/sys-libs/mesa/patches/mesa-17.1.7.patchset
@@ -1,4 +1,4 @@
-From ec8d0a5d3dffdb41b48d2dd1e83c5a05bad0e95f Mon Sep 17 00:00:00 2001
+From 955e760862127dbc1970faabac4e8f08a85d574f Mon Sep 17 00:00:00 2001
 From: Jerome Duval <jerome.duval@gmail.com>
 Date: Mon, 11 Sep 2017 21:29:14 +0200
 Subject: Haiku patch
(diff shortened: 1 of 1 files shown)


i80and:
Received feedback that I should contribute directly to mesa, rather than going through haikuports

waddlesplash:
@pulkomandy is correct that we need to post this to Mesa trunk, but that usually takes some time to propagate to a new release, so we should patch here as well as submitting upstream.

waddlesplash:
@kallisti5, could you please check this.

kallisti5:
@i80and  I actually have commit access and can get it in (with you as the author) if you don't have the time.

I have some large patches to push adding Haiku to their new meson build system:
  https://github.com/kallisti5/mesa/commits/haiku-meson

The generated binaries don't work at the moment though... still investigating.

i80and:
I appreciate the offer, if you think this patch is good enough!

waddlesplash:
More discussion at https://dev.haiku-os.org/ticket/13191#comment:27.

kallisti5:
For the locking changes.. i'd say lets keep them in a haikuports patch for now until we are sure they work like we want... for the rest of the changes... I just need to find the time.

I'd be interested in finding a backup person with commit access who would want to maintain Mesa... email me if interested.  I'm pretty sure I could get you commit access.

waddlesplash:
Fixed instead by https://github.com/haikuports/haikuports/commit/d1d70a5c4be3e45f46a05270edd1fc853a428c62

PR #2076, opened 2018-01-07
app-arch/unshield: new recipe

alaviss:
Done as part of GCI 2017

### DIFF
--- a/app-arch/unshield/unshield-1.4.2.recipe
+++ b/app-arch/unshield/unshield-1.4.2.recipe
@@ -1,0 +1,94 @@
+SUMMARY="Tool and library to extract CAB files from InstallShield installers"
+DESCRIPTION="Unshield extracts CAB files from InstallShield installers, \
+used to install software on Microsoft Windows based machines."
+HOMEPAGE="https://github.com/twogood/unshield"
+COPYRIGHT="2017 David Eriksson"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/twogood/unshield/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="5dd4ea0c7e97ad8e3677ff3a254b116df08a5d041c2df8859aad5c4f88d1f774"
+SOURCE_FILENAME="unshield-$portVersion.tar.gz"
+PATCHES="unshield-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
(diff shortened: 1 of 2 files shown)


app-arch/unshield/unshield-1.4.2.recipe
```diff
@@ -0,0 +1,81 @@
...
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	install -t "$docDir" ChangeLog
+
+	prepareInstalledDevelLib libunshield
+	fixPkgconfig
+	packageEntries devel "$developDir"
```
Begasus: Why the qoutes around $developDir?

app-arch/unshield/unshield-1.4.2.recipe
```diff
@@ -0,0 +1,81 @@
...
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir"
+	install -t "$docDir" ChangeLog
+
+	prepareInstalledDevelLib libunshield
+	fixPkgconfig
```
Begasus: Not structural, but I would put an empty line after this one :)

Begasus:
man page should be moved to $docDir instead of $dataDir :)

app-arch/unshield/unshield-1.4.2.recipe
alaviss: Because why not? Quote is generally my go to for every unknown variable to avoid whitespace expansion. While currently running haikuporter without chroot is broken, it could be fixed in the future and this would be useful then

alaviss:
Updates:
- Now with working iconv support (https://github.com/twogood/unshield/pull/71)
- Moved man pages to `$manDir`

alaviss:
Updates:
- Added Expander rules (need [ticket#13934][0])

[0]: https://dev.haiku-os.org/ticket/13934

app-arch/unshield/unshield-1.4.2.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$docDir" "$manDir" "$dataDir/expander/rules"
+	install -t "$docDir" ChangeLog
+
+	mv "$dataDir/man/man1" "$manDir/man1"
+	rm -rf "$dataDir"
```
Begasus: Should be: rm -rf "$dataDir/man"

I asked yesterday also, why the qoutes here (and in other places mentioning $paths)?

Begasus:
Looks ok to me :+1:

PR #2081, opened 2018-01-07
shc, new recipe

Begasus:


dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,47 @@
+SUMMARY="A generic shell script compiler"
+DESCRIPTION="Shc takes a script, which is specified on the command line \
+and produces C source code. The generated source code is then compiled \
+and linked to produce a stripped binary executable. \
+The compiled binary will still be dependent on the shell specified in the \
+first line of the shell code (i.e shebang) (i.e. #!/bin/sh), thus shc does \
+not create completely independent binaries."
+HOMEPAGE="https://github.com/neurobin/shc/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/neurobin/shc/"
+HOMEPAGE="https://github.com/neurobin/shc"
```

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,47 @@
...
+	cmd:autoconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	"
+
+BUILD()
+{
+	libtoolize -fci
+	autogen.sh
```
fbrosson:
```diff
-	autogen.sh
+	./autogen.sh
```

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,47 @@
+SUMMARY="A generic shell script compiler"
+DESCRIPTION="Shc takes a script, which is specified on the command line \
+and produces C source code. The generated source code is then compiled \
+and linked to produce a stripped binary executable. \
```
fbrosson:
```diff
-DESCRIPTION="Shc takes a script, which is specified on the command line \
-and produces C source code. The generated source code is then compiled \
-and linked to produce a stripped binary executable. \
+DESCRIPTION="Shc takes a script, which is specified on the command line and \
+produces C source code. The generated source code is then compiled +and linked \
+to produce a stripped binary executable. \
```

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,47 @@
...
+The compiled binary will still be dependent on the shell specified in the \
+first line of the shell code (i.e shebang) (i.e. #!/bin/sh), thus shc does \
+not create completely independent binaries."
```
fbrosson:
```diff
-first line of the shell code (i.e shebang) (i.e. #!/bin/sh), thus shc does \
-not create completely independent binaries."
+first line of the shell code (i.e shebang) (i.e. #!/bin/sh), thus shc does not \
+create completely independent binaries."
```

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,47 @@
+SUMMARY="A generic shell script compiler"
+DESCRIPTION="Shc takes a script, which is specified on the command line and \
+produces C source code. The generated source code is then compiled \
+and linked to produce a stripped binary executable. \
```
fbrosson:
```diff
-produces C source code. The generated source code is then compiled \
+produces C source code. The generated source code is then compiled and linked \
-and linked to produce a stripped binary executable. \
+to produce a stripped binary executable. \
```
```diff
 CHECKSUM_SHA256="da6a2a3ff4c356a61e086c616561bf681489993cab00c426bad0cfd703a68063"
+SOURCE_FILENAME="shc-$portVersion.tar.gz"
```

Begasus:
Wanted to give this a try with tartos https://github.com/neurobin/tartos
Running shc from Terminal on it works ok, but when I want to try it with a recipe for tartos it doesn't work :/

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,48 @@
...
+The compiled binary will still be dependent on the shell specified in the \
```
fbrosson:
```diff
-produces C source code. The generated source code is then compiled \
+produces C source code. The generated source code is then compiled and linked \
-and linked to produce a stripped binary executable. \
+to produce a stripped binary executable. \
```

Begasus: I know we have the 80 char limit, but is it needed to fill out the lines to meet the limit?
PS, no problem to change it :)

fbrosson: I hope I'm not being too picky, but if you are inside a sentence, then there is no reason to keep a short line.
That said, if you start a new sentence then, for me, it's OK to have short lines that end with a full stop or with a full stop and a backslash. (This allows e.g. to insert new sentences with less "diff-noise".)

We don't want to see recipes with unnecessarily short line and without even a tiny reason for that, don't we?

Begasus: Fair enough :+1:

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,48 @@
+SUMMARY="A generic shell script compiler"
+DESCRIPTION="Shc takes a script, which is specified on the command line and \
+produces C source code. The generated source code is then compiled and linked \
+to produce a stripped binary executable. \
+The compiled binary will still be dependent on the shell specified in the \
+first line of the shell code (i.e shebang) (i.e. #!/bin/sh), thus shc does not \
+create completely independent binaries."
+HOMEPAGE="https://github.com/neurobin/shc"
+COPYRIGHT="2015-2017 Francisco Garcia
+	Md Jahidul Hamid"
```
fbrosson: According to https://github.com/neurobin/shc/commits?author=neurobin you can add "2015-2017":
```diff
 COPYRIGHT="2015-2017 Francisco Garcia
-	Md Jahidul Hamid"
+	2015-2017 Md Jahidul Hamid"
```
I don't know, however, if the years are correct for Francisco Garcia. (But I guess this can be fixed later, if someone finds out it is not correct.)

Begasus: According to the link on gentoo Francisco is the (founder?) of shc? 
http://gpo.zugaina.org/dev-util/shc

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,48 @@
...
+	cmd:aclocal
+	cmd:autoconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	"
+
+BUILD()
+{
+	libtoolize -fci
```
korli: libtoolize will be called by autogen.

Begasus: Will need to check, but iirc it was needed before calling autogen.sh ...

alaviss: Running `autoreconf -fci` could be an option.

Begasus: Turns out it also works without  libtoolize -fci, I'm just getting a warning from aclocal then:
aclocal: warning: couldn't open directory 'm4': No such file or directory

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,55 @@
...
+
+INSTALL()
+{
+	make install
+}
+
+TEST()
+{
+	#the produced binary doesn't work when running the test
+	#with haikuporter, running it from Terminal works well
```
fbrosson: The runtimes produced by shc need to be able to run the shell interpreter, so I guess that adding `cmd:sh` or `cmd:bash` to `BUILD_REQUIRES` will make the test succeed.
(If a test succeeds outside of the chroot, then it should also work in the chroot, as long as you add the required runtime dependencies.)

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,55 @@
...
+	cd test
+	$sourceDir/src/shc -v -f test.bash -o test
```
fbrosson:
```diff
 TEST()
 {
-	#the produced binary doesn't work when running the test
-	#with haikuporter, running it from Terminal works well
 	cd test
 	$sourceDir/src/shc -v -f test.bash -o test
+	echo | ./test Testing $portVersionedName on $targetArchitecture.
 }
```

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,55 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	"
```
fbrosson:
```diff
	cmd:make
 	"
+if [ "$recipeAction" = TEST ]; then
+BUILD_PREREQUIRES="$BUILD_PREREQUIRES
+	cmd:bash
+	"
+fi
```
Currently this is not really needed because bash is already our default shell, but it might be a good idea to have it here, in case we switch to a different shell.
That said, this is a lot of noise for something that is unlikely to happen. So please feel free to skip this suggestion :-)

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,55 @@
+SUMMARY="A generic shell script compiler"
+DESCRIPTION="Shc takes a script, which is specified on the command line and \
+produces C source code. The generated source code is then compiled and linked \
+to produce a stripped binary executable. \
+The compiled binary will still be dependent on the shell specified in the \
+first line of the shell code (i.e shebang) (i.e. #!/bin/sh), thus shc does not \
+create completely independent binaries."
+HOMEPAGE="https://github.com/neurobin/shc"
+COPYRIGHT="1997-2013 Francisco Garcia
+ 	2015-2017 Md Jahidul Hamid"
```
fbrosson:
```diff
COPYRIGHT="1997-2013 Francisco Garcia
- 	2015-2017 Md Jahidul Hamid"
+	2015-2017 Md Jahidul Hamid"
```
(There's a white space before the tab.)

dev-util/shc/shc-3.9.6.recipe
Begasus: Running the test isn't the issue here, that part is ok, the produces binary doesn't work :)

  ~/haikuports/dev-util/shc/work-3.9.6/sources/shc-3.9.6/test> ./test
  ./test:UaÐ¶1dóš‰
  ~/haikuports/dev-util/shc/work-3.9.6/sources/shc-3.9.6/test> ./test
  ./test: L.
  F
  ޘ_

Begasus: In haikuports with and without cmd: sh (and bash) ...

Begasus: In Terminal (outside of haikuporter) ... 
  ~/haikuports/dev-util/shc/work-3.9.6/sources/shc-3.9.6/test> shc -v -f test.bash -o test
  shc shll=bash
  shc [-i]=-c
  shc [-x]=exec '%s' "$@"
  shc [-l]=
  shc opts=
  shc: cc  test.bash.x.c -o test
  shc: strip test
  shc: chmod ug=rwx,o=rx test
  ~/haikuports/dev-util/shc/work-3.9.6/sources/shc-3.9.6/test> ./test
  $@ is 
  command line: ./test 
  hello world
  [15288] PAUSED... Hit return!

fbrosson: It worked fine for me, on x86_gcc2.
(I'll test on x86_64.)

fbrosson: `hp --test shc` also works fine for me on x86_64:
```
heap size:                      765
TOC size:                        41
package attributes size:        851
total size:                     845
Testing ...
shc shll=bash
shc [-i]=-c
shc [-x]=exec '%s' "$@"
shc [-l]=
shc opts=
shc: cc  test.bash.x.c -o test
shc: strip test
shc: chmod ug=rwx,o=rx test
$@ is Testing shc-3.9.6 on x86_64.
command line: ./test Testing shc-3.9.6 on x86_64.
hello world
[2928] PAUSED... Hit return!
/data/haikuports/dev-util/shc> echo $?
0
```

Begasus: As explained earlier :) running the test is ok :) (the produced binary is not)

dev-util/shc/shc-3.9.6.recipe
Begasus: Wouldn't hurt to include it also :)

dev-util/shc/shc-3.9.6.recipe
```diff
@@ -0,0 +1,59 @@
...
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:gcc
+	cmd:libtoolize
+	cmd:make
+	"
+if [ "$recipeAction" = TEST ]; then
```
fbrosson: Oops, sorry, I just removed my comment suggesting the inclusion of this trick... just before your last update.
That's because I just recalled that @alaviss had noticed that this trick had stopped working.
So, since bash is already activated, it might be worth dropping the trick.

That said, if you want you can add this:
```diff
 	cmd:make
 	"
+
+TEST_REQUIRES="
+	cmd:bash
+	"
+BUILD_PREREQUIRES="$BUILD_PREREQUIRES
+	$TEST_REQUIRES
+	"
+
 BUILD()
```

Or you might prefer something else: adding `cmd:bash` directly to `BUILD_PREREQUIRES`.

I don't know which of the 3 possibilities is the best.
To summarize, we can:
* Drop any reference to `cmd:bash`. _(I think I would choose this.)_ (This works fine for now, but the recipe will need to be modified if/when we change the default shell to either require `cmd:bash` or update `TEST()` to make it use the new default shell.)
* Put `cmd:bash` directly in `BUILD_PREREQUIRES`. This does not hurt but is unneeded at the moment.
* Put `cmd:bash` in `TEST_REQUIRES` and insert the temporary hack to append it to `BUILD_PREREQUIRES`. _(This is what's in the diff in the beginning of this message.)_ But this seems to be a bit excessive, after all, given that the two above alternatives work fine and are much shorter.

Begasus: Removed cmd:bash for now, if someone sees the need to have it in I'm open to suggestions :)

PR #2082, opened 2018-01-07
Voxophone [recipe] [GCI]

bach5000:
Finally back to recipes 😉 
* Tested on x86_gcc2, and works 🎊 
* Not tested on x86_64 yet... since I don't have it available.

### DIFF
--- a/haiku-apps/voxophone/patches/voxophone-1.0.patchset
+++ b/haiku-apps/voxophone/patches/voxophone-1.0.patchset
@@ -1,0 +1,18 @@
+From fbc0ce8e97430146e9962b6dc10ea91ad8d9e286 Mon Sep 17 00:00:00 2001
+From: Yourself <bach5000@gmail.com>
+Date: Mon, 8 Jan 2018 18:46:06 +0000
+Subject: support x86_64
+
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 9cc5a6f..74e8b26 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -4,4 +4,4 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.8)
+ FILE(GLOB MAIN_SRC "src/*.cpp")
+ 
+ ADD_EXECUTABLE(Voxophone ${MAIN_SRC})
+-TARGET_LINK_LIBRARIES(Voxophone be root tracker media network netapi)
++TARGET_LINK_LIBRARIES(Voxophone be root tracker media network bnetapi)
+-- 
+2.7.0
--- a/haiku-apps/voxophone/patches/voxophone-1.0.patchset
+++ b/haiku-apps/voxophone/patches/voxophone-1.0.patchset
@@ -1,5 +1,5 @@
 From fbc0ce8e97430146e9962b6dc10ea91ad8d9e286 Mon Sep 17 00:00:00 2001
-From: Yourself <bach5000@gmail.com>
+From: bach5000 <bach5000@users.noreply.github.com>
 Date: Mon, 8 Jan 2018 18:46:06 +0000
 Subject: support x86_64
 
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -29,6 +29,9 @@ BUILD_PREREQUIRES="
 	cmd:g++
 	cmd:make
 	"
+	
+defineDebugInfoPackage voxophone \
+	$appsDir/Voxophone
 
 BUILD()
 {
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -9,6 +9,7 @@ SOURCE_URI="https://github.com/aarroyoc/Voxophone/archive/$srcGitRev.tar.gz"
 CHECKSUM_SHA256="bc98a70248057aa482f038e3689876dc0632aca4aa65d2417d9f692c86b486b5"
 SOURCE_FILENAME="voxophone-$portVersion-$srcGitRev.tar.gz"
 SOURCE_DIR="Voxophone-$srcGitRev"
+PATCHES="voxophone-$portVersion.patchset
 
 ARCHITECTURES="x86_gcc2 x86_64"
 
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -9,7 +9,7 @@ SOURCE_URI="https://github.com/aarroyoc/Voxophone/archive/$srcGitRev.tar.gz"
 CHECKSUM_SHA256="bc98a70248057aa482f038e3689876dc0632aca4aa65d2417d9f692c86b486b5"
 SOURCE_FILENAME="voxophone-$portVersion-$srcGitRev.tar.gz"
 SOURCE_DIR="Voxophone-$srcGitRev"
-PATCHES="voxophone-$portVersion.patchset
+PATCHES="voxophone-$portVersion.patchset"
 
 ARCHITECTURES="x86_gcc2 x86_64"
 
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -13,7 +13,7 @@ SOURCE_DIR="Voxophone-$srcGitRev"
 ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
-  voxophone = $portVersion
+	voxophone = $portVersion
 	app:Voxophone = $portVersion
 	"
 REQUIRES="
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -24,7 +24,6 @@ BUILD_REQUIRES="
 	haiku_devel
 	"
 BUILD_PREREQUIRES="
-	makefile_engine
 	cmd:cmake
 	cmd:g++
 	cmd:make
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -29,7 +29,7 @@ BUILD_PREREQUIRES="
 	cmd:g++
 	cmd:make
 	"
-	
+
 defineDebugInfoPackage voxophone \
 	$appsDir/Voxophone
 
--- a/haiku-apps/voxophone/voxophone-1.0.recipe
+++ b/haiku-apps/voxophone/voxophone-1.0.recipe
@@ -1,0 +1,44 @@
+SUMMARY="An online radio player"
+DESCRIPTION="Voxophone is a online radio player application for Haiku."
+HOMEPAGE="https://github.com/aarroyoc/Voxophone"
+COPYRIGHT="2015 Adrian Arroyo Calle"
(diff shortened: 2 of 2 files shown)


haiku-apps/voxophone/voxophone-1.0.recipe
```diff
@@ -0,0 +1,44 @@
...
+BUILD()
+{
+	cmake .
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	cp Voxophone $appsDir
```
fbrosson: > Finally back to recipes 😉

Cool!

Suggestion:
```diff
-	mkdir -p $appsDir
+	install -d $appsDir
-	cp Voxophone $appsDir
+	install Voxophone $appsDir
```

BTW, I'm trying to build it on x86_64, but it needs libnetapi, which is provided by samba.
So I think you should add `lib:libnetapi` do `REQUIRES` and `devel:libnetapi` do `BUILD_REQUIRES`. (I'll try that.)

bach5000: Good idea, @fbrosson 😃 
I'll also add those requirements.

haiku-apps/voxophone/voxophone-1.0.recipe
```diff
@@ -0,0 +1,46 @@
...
+
+ARCHITECTURES="x86_gcc2 x86_64"
+
+PROVIDES="
+	voxophone = $portVersion
+	app:Voxophone = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libnetapi
```
alaviss: `lib:libnetapi` is a part of `haiku`, so you don't have to specify it

```diff
-	lib:libnetapi

File: haiku-apps/voxophone/voxophone-1.0.recipe [hunk 3]
```diff
@@ -0,0 +1,46 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libnetapi
```
alaviss:
```diff
-	devel:libnetapi

File: haiku-apps/voxophone/voxophone-1.0.recipe [hunk 4]
```diff
@@ -0,0 +1,46 @@
...
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
```
alaviss: This port is not using `makefile_engine` to build

```diff
-	makefile_engine

File: haiku-apps/voxophone/voxophone-1.0.recipe [hunk 5]
```diff
@@ -0,0 +1,46 @@
+SUMMARY="An online radio player"
+DESCRIPTION="Voxophone is a online radio player application for Haiku."
```
alaviss:
```diff
-DESCRIPTION="Voxophone is a online radio player application for Haiku."
+DESCRIPTION="Voxophone is an online radio player application for Haiku."

Comment (bach5000) [reply haiku-apps/voxophone/voxophone-1.0.recipe]: Whoops what's grammar again 😄

File: haiku-apps/voxophone/voxophone-1.0.recipe (hunk 2 above)
Comment (fbrosson) [reply haiku-apps/voxophone/voxophone-1.0.recipe]: @alaviss, you are right!

For some unknown reason, when I tried to build the recipe (before makefile_engine was removed), I got an error about lnetapi, so I typed:
```
pkgman search lib:libnetapi
Status  Name       Description                                 
---------------------------------------------------------------
s       samba      Library bits of the samba network filesystem
s       samba_x86  Library bits of the samba network filesystem
```
and thought it needed samba.
But now, without `makefile_engine` and without `lib:libnetapi` the recipe builds fine on x86_gcc2 :)

So sorry @bach5000 for my incorrect suggestion and thanks for your help!

(I'll try on x86_64 soon.)

Comment (fbrosson) [reply haiku-apps/voxophone/voxophone-1.0.recipe]: Hmm, on x86_64 it fails to build:
```
Scanning dependencies of target Voxophone
[ 25%] Building CXX object CMakeFiles/Voxophone.dir/src/app.cpp.o
[ 50%] Building CXX object CMakeFiles/Voxophone.dir/src/dashboard.cpp.o
[ 75%] Building CXX object CMakeFiles/Voxophone.dir/src/player_manager.cpp.o
[100%] Linking CXX executable Voxophone
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/5.4.0/../../../../x86_64-unknown-haiku/bin/ld: cannot find -lnetapi
CMakeFiles/Voxophone.dir/build.make:146: recipe for target 'Voxophone' failed
```
What's funny is that on x86_64 it says it `cannot find -lnetapi` but on x86_gcc2 it did not link against it:
```
objdump -x /system/apps/Voxophone |grep -w NEEDED
  NEEDED      libbe.so
  NEEDED      libroot.so
  NEEDED      libtracker.so
  NEEDED      libmedia.so
  NEEDED      libnetwork.so
  NEEDED      libbnetapi.so
  NEEDED      libstdc++.r4.so
```
Hopefully, making this change in CMakeLists.txt:
```diff
-TARGET_LINK_LIBRARIES(Voxophone be root tracker media network netapi)
+TARGET_LINK_LIBRARIES(Voxophone be root tracker media network bnetapi)
```
allows to build on x86_64 :)

BTW, I added a debuginfo package with:
```
defineDebugInfoPackage voxophone \
	$appsDir/Voxophone
```
and it is creating a real package, not a fake one :)

pulkomandy:
On Sun, Jan 07, 2018 at 04:43:03PM -0800, fbrosson wrote:
> @alaviss, you are right!
> 
> For some unknown reason, when I tried to build the recipe (before makefile_engine was removed), I got an error about lnetapi, so I typed:
> ```
> pkgman search lib:libnetapi
> Status  Name       Description                                 
> ---------------------------------------------------------------
> s       samba      Library bits of the samba network filesystem
> s       samba_x86  Library bits of the samba network filesystem
> ```
> and thought it needed samba.

There is a libnetapi in samba.
There is a libbnetapi in Haiku (notice the extra b), and unfortunately
for compatibility reasons we provide a symlink libnetapi -> libbnetapi,
which we should eventually remove before it creates more confusion :)

fbrosson:
Thanks @pulkomandy for the info about the symlink.
Actually, as I wrote in [my previous message](https://github.com/haikuports/haikuports/pull/2082#discussion_r160066682), I found the fix:
> Hopefully, making this change in CMakeLists.txt:
```diff
-TARGET_LINK_LIBRARIES(Voxophone be root tracker media network netapi)
+TARGET_LINK_LIBRARIES(Voxophone be root tracker media network bnetapi)
```
> allows to build on x86_64 :)

bach5000:
@fbrosson I would have to make a patchset for this, don't I?

fbrosson:
Not necessarily. If it's a simple change that can be done with a `PATCH()` function, I'd say that it's OK to avoid the patchset. You could, for instance, do something like:
```bash
PATCH()
{
	sed -i -e "/TARGET_LINK_LIBRARIES/ s/ \(netapi\)/ b\1/" \
		CMakeLists.txt
}
```
I wanted to test this... but then completely forgot about it.
(What we need to check is if this sed command only changes the line we need to change, and nothing else. But I'm pretty sure it will work.)

pulkomandy:
I would prefer a PATCHES to a PATCH() function.

PATCH() function is a little less readable (one has to parse the sed logic), and may accidentally do more or less than needeed in next version of the software without warnings. Moreover, they do not easily allow to trace the history of each patch. They are also less easy to upstream.

As a result, they should be reserved for when there are a lots of repetitive changes (typically, replacing /usr/bin in shell script headers). In these cases, the changes are quite obvious and having them repeated many times in the patchset becomes more annoying than having to understand the sed expression.

So if you only need to change the libraries in 1 or 2 places, a PATCHSET seems more appropriate.

fbrosson:
OK, thanks @pulkomandy for your message. (I was not really sure it was OK.)

@bach5000, let's go for the patchset, then :)
Thanks!

haiku-apps/voxophone/voxophone-1.0.recipe
korli: @fbrosson are you sure? Libnetapi isn't the network API on R5?

fbrosson: Sorry @korli, these messages are outdated. I should have stroken the text.
The explanation is [here](https://github.com/haikuports/haikuports/pull/2082#discussion_r160065744).

fbrosson:
@pulkomandy, would this variant be more readable?
```bash
PATCH()
{
	sed -i -e "s,\(TARGET_LINK_LIBRARIES(Voxophone be root tracker media network\) \(netapi\),\1 b\2," \
		CMakeLists.txt
}
```
The only risk with that one is that it would fail if the pattern changes after "netapi". But in that case it would be detected because on x86_64 the build would fail. And if the pattern changes, the patchset would also fail to apply.
The added value here is that it is, imo, more readable and easier to maintain than a patchset.

Reminder: This is just a suggestion. I did not even test it. Thanks!

bach5000:
I will try to make a patch (if necessary) and test it when I have Haiku available.
I should be able to do this after all the homework :/

bach5000:
Please wait...
I'm going to try to test these.

haiku-apps/voxophone/patches/voxophone-1.0.patchset
```diff
@@ -0,0 +1,18 @@
+From fbc0ce8e97430146e9962b6dc10ea91ad8d9e286 Mon Sep 17 00:00:00 2001
+From: Yourself <bach5000@gmail.com>
```
fbrosson: You shoud probably replace `Yourself` by `bach5000` and `gmail.com` by `users.noreply.github.com` (which, assembled, produce your github e-mail alias.)

BTW, you might also wish to configure your git user with:
```
git config --global user.name bach5000
git config --global user.email "your github e-mail alias here"
```
then squash your commits, and then
```
git commit --amend --author="bach5000 <your github e-mail alias here>"
```
I've even seen some antispam digits being used in front of the github e-mail alias.

haiku-apps/voxophone/voxophone-1.0.recipe
```diff
@@ -0,0 +1,44 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:g++
+	cmd:make
+	"
+
+BUILD()
```
fbrosson:
```diff
  	cmd:make
 	"
 
+defineDebugInfoPackage voxophone \
+	$appsDir/Voxophone
+
 BUILD()
```

bach5000: Oh, sorry!
I noticed this earlier but forgot to put it in :(

haiku-apps/voxophone/patches/voxophone-1.0.patchset
bach5000: Hmmm... I thought I already configured my git things.
I guess I haven't.

haiku-apps/voxophone/voxophone-1.0.recipe
```diff
@@ -0,0 +1,47 @@
...
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:g++
+	cmd:make
+	"
+
```
fbrosson:
```diff
-	
+
```
(there's a tab)

bach5000: Whoops!
(I think I told you I'm terrible with whitespaces)
If only Github had an option to show all the whitespaces :(

PR #2085, opened 2018-01-08
media-libs/libpgf: new recipe

alaviss:
Done as part of GCI

Changes to libpgf:
- Add Haiku support

### DIFF
--- a/media-libs/libpgf/libpgf-7.15.32.recipe
+++ b/media-libs/libpgf/libpgf-7.15.32.recipe
@@ -1,0 +1,73 @@
+SUMMARY="An implementation of the Progressive Graphics File (PGF)"
+DESCRIPTION="libpgf is a library implementing an interface for handling \
+PGF (Progressive Graphic File) format files."
+HOMEPAGE="http://www.libpgf.org/"
+COPYRIGHT="2006 xeraina GmbH, Switzerland"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/libpgf/libpgf/$portVersion-latest/libPGF-codec-and-console-src.zip"
+CHECKSUM_SHA256="f1dd054b09b0f3651209d67529490b4cbc156a985d8b019ff3bfe575260f4221"
+SOURCE_FILENAME="libpgf-$portVersion.zip"
+SOURCE_DIR="PGF/Codec"
+PATCHES="libpgf-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="7.0.15"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	libpgf$secondaryArchSuffix = $portVersion
+	lib:libpgf$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libpgf${secondaryArchSuffix}_devel = $portVersion
+	devel:libpgf$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	libpgf$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:automake
+	cmd:doxygen
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage libpgf$secondaryArchSuffix \
+	"$libDir/libpgf.so.$libVersion"
+
+BUILD()
+{
+	sh ./autogen.sh
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f "$libDir"/*.la
+
+	install -d -m 755 "$developDocDir"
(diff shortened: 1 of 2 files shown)


media-libs/libpgf/libpgf-7.15.32.recipe
```diff
@@ -0,0 +1,73 @@
...
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$developDocDir"
+	mv "$dataDir/doc/libpgf-$portVersion/html" "$developDocDir"
```
fbrosson: It looks like the trend is to have the html docs in /system/documentation/packages/libpgf_x86/html/` (on x86_gcc2 or the same without the _x86 on x87_64).
So it might be better to use a different mv:
```diff
-	mv "$dataDir/doc/libpgf-$portVersion/html" "$developDocDir"
+	mv "$dataDir/doc/libpgf-$portVersion" "$docDir"
```
and later pass`"$docDir"/html` to `packageEntries devel`.
(I need to check if this works as I am expecting.)

You have probably already noticed that adding a `false` in `INSTALL()` after the `make install` is a nice trick to be able to examine in `media-libs/libpgf/work-x86-7.15.32/packaging/` what gets installed before `INSTALL()` continues its job. (this path is on x86_gcc2 using x86 secondary arch, of course.)

Regarding the `SOURCE_DIR`, although it works fine with a / in it, I think some **older** versions of haikuporter did not correctly handle a `SOURCE_DIR` with more than one level, so I would change it to PGF (instead of PGF/Codec) and later do a "cd Codec", but this is because I am a paranoid :)

media-libs/libpgf/libpgf-7.15.32.recipe
```diff
@@ -0,0 +1,73 @@
...
+	"
+
+defineDebugInfoPackage libpgf$secondaryArchSuffix \
+	"$libDir/libpgf.so.$libVersion"
+
+BUILD()
+{
+	sh ./autogen.sh
+	runConfigure ./configure
+	make $jobArgs
```
fbrosson: I haven't tried this yet, but I think it might be a good idea. (This assumes you set `SOURCE_DIR="PGF"`)

```diff
 BUILD()
 {
+	for d in Codec Console; do
+		pushd $d
-	sh ./autogen.sh
+		sh ./autogen.sh
-	runConfigure ./configure
+		runConfigure ./configure
-	make $jobArgs
+		make $jobArgs
+		popd
+	done
 }
```
It might be nice to add a libpdf_bin subpackage. What do you think?
Hmm, maybe it won't build without a lot of changes. I'll try it.

media-libs/libpgf/libpgf-7.15.32.recipe
fbrosson: Hmm, after all, let's forget my previous suggestion about moving the html docs to another place.
Following a trend just _because it is a trend_ is not a good reason. Sorry for the noise.
I guess your recipe is perfect as it is now.

alaviss: From [BuildRecipes](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes)

>- $developDocDir: The absolute base directory for development-related documentation specific to the port being built. Documentation in html- or pdf-format should be put here, manpages should be put into $manDir.

Probably you should start enforcing this now

Regarding `SOURCE_DIR`, given that our builder always use the latest commit from https://github.com/haikuports/haikuporter, and most are [probably using](https://github.com/haikuports/haikuports/wiki#setting-up-haikuports) it, I don't see why this should be a problem.

fbrosson:
There are some files that libpgf is installing in man3 and which have names that are too generic, like `config.h.3`, `Decoder.h.3`, `Encoder.cpp.3`, `Encoder.h.3`, or `IOException.3`:
```
package list -p libpgf_x86_devel-7.15.32-1-x86_gcc2.hpkg | \
	grep documentation/man/man3/ | grep -v PGF
documentation/man/man3/BitStream.h.3
documentation/man/man3/CDecoder.3
documentation/man/man3/CDecoder_CMacroBlock.3
documentation/man/man3/CEncoder.3
documentation/man/man3/CEncoder_CMacroBlock.3
documentation/man/man3/CSubband.3
documentation/man/man3/CWaveletTransform.3
documentation/man/man3/Decoder.cpp.3
documentation/man/man3/Decoder.h.3
documentation/man/man3/Encoder.cpp.3
documentation/man/man3/Encoder.h.3
documentation/man/man3/IOException.3
documentation/man/man3/ROIBlockHeader.3
documentation/man/man3/ROIBlockHeader_RBH.3
documentation/man/man3/Subband.cpp.3
documentation/man/man3/Subband.h.3
documentation/man/man3/WaveletTransform.cpp.3
documentation/man/man3/WaveletTransform.h.3
documentation/man/man3/config.h.3
```
I don't know how this should be handled.
Should we just drop some of these files?
Or move (all or some of them) to a new man3pgf section?

alaviss:
I think we could just drop the man pages. The code examples looks terrible there, and their content is in those html pages anyway

alaviss:
Updates:
- Get rid of the man pages as their content is mostly a duplicate of the html files, not to mention the code examples look terrible there

alaviss:
Anyway, could you publish the mandoc task, @fbrosson?

media-libs/libpgf/libpgf-7.15.32.recipe
```diff
@@ -0,0 +1,71 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	install -d -m 755 "$developDocDir"
```
fbrosson:
```diff
 	make install
 
+	rm -f "$libDir"/libpgf.la
+
 	install -d -m 755 "$developDocDir"
```
Oops, you forgot to get rid of the libtool file.
(You can use `"$libDir"/*.la` if you prefer.)

alaviss:
Updates:
- Get rid of libtool files (thx @fbrosson)

fbrosson:
The mandoc task is available

PR #2087, opened 2018-01-08
Eventual [recipe] [GCI] 

bach5000:
**\~Work in Progress\~**
This recipe is currently broken 🙁 
`cp: cannot create regular file '/boot/home/config/non-packaged/lib/libActivity.so'`
I believe this is a permission error.
I ran the commands in `BUILD` with `sh` in the source folder, and it works okay (on gcc2).

Any help/suggestions would be welcome 😄


haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,60 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku. \
+Intended to be a full-featured organizer. "
+HOMEPAGE="https://sourceforge.net/projects/eventual/"
+COPYRIGHT="2016(?) Alex Hitech" #Help on this one
```
diversys: COPYRIGHT="2011 Alex Hitech"

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+	sh makelib.sh
+	cd src/Libraries
+	for LIBDIR in */ ; do
+		LIBDIR=${LIBDIR%?}
+		NAME=lib$LIBDIR.so
+		LIB=$LIBDIR/objects.x86-cc2-release/$NAME
+		cp $LIB ../EventEditor/$NAME
+		cp $LIB ../EventServer/$NAME
+		cp $LIB ../PreferencesPreflet/$NAME
+		cp $LIB /boot/home/config/non-packaged/lib/$NAME #Error here
```
diversys: package libs should be put in $appsDir/Eventual/lib

bach5000: Okay, @diversys.
I'll make the commit, but I can't test it :(

bach5000:
I think the recipe should get through the `BUILD` stage with @diversys' fixes (hopefully)

bach5000:
Oh. I figured out what's wrong.
One of the makefiles had the wrong path for makefile-engine.

bach5000:
The recipe builds, until the very last moment, when I get a trail of `POLICY ERROR`s.
All of them say `Warning: POLICY ERROR: can't find used library "lib[some-lib].so"`

fbrosson:
You just need to
```
pkgman search lib:[some-lib]
pkgman search devel:[some-lib]
```
and then use `pkgman install` to install the appropriate packages.
Of course, you need to add the lib:[some-lib] to your `REQUIRES` and devel:[some-lib] to your `BUILD_REQUIRES`.

bach5000:
@fbrosson Sorry :(
Forgot to specify that these libs are part of the program.
This program has its own libs. Did I forget to put it where Haikuporter could find it?

fbrosson:
OK, no problem.
Hmm, I should try to compile your recipe... (but you forgot to git add your patchset)

bach5000:
Ahhh!
Sorry ... I'm very forgetful on Mondays :(
I will do that right away.

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,64 @@
...
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/Eventual
+	#mkdir -p $appsDir/Eventual/lib
+	cp -r 'src/EventEditor/objects/Eventual Editor' $appsDir/Eventual
+	cp -r 'src/EventServer/objects/Eventual Server' $appsDir/Eventual
+	cp -r 'src/PreferencesPreflet/objects/Eventual Preferences' $appsDir/Eventual
+	#cp -r src/LibraryObjects/ $appsDir/Eventual/lib
```
fbrosson: OK. The problem comes from the `INSTALL()` funciton. It is not copying any lib. When the build fails you can:
```
cd /boot/home/haikuports/haiku-apps/eventual/work-1.0/packaging/
cd eventual
find . -type f -name "lib*.so*"
```
this should normally list all the libs, but it is not showing anything.


If you comment out the call to cp which copies the libraries, nobody else will install the shared libs :(

bach5000: Did I comment those lines? :(
I uncommented them, rebuilt the recipe, and now the libs show up when I type
`find . -type f -name "lib*.so*"`.

```
./apps/Eventual/lib/LibraryObjects/libActivity.so
./apps/Eventual/lib/LibraryObjects/lib[some-lib].so

...
```
But the build still gives me `POLICY ERROR`s (?)

bach5000: Oh wait... let me guess...
There shouldn't be `LibraryObjects` in the directory, right?

bach5000:
Yes! It works 🎊 😄 
Perhaps time for some cosmetic edits?

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,64 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku. \
+Intended to be a full-featured organizer. "
+HOMEPAGE="https://sourceforge.net/projects/eventual/"
+COPYRIGHT="2011 Alex Hitech"
```
alaviss: There're two [copyright holders][0]:

- [Alexey Burshtein][1]
- [Kevin H. Patterson][2]

[0]: https://github.com/HaikuArchives/Eventual/search?utf8=%E2%9C%93&q=copyright&type=
[1]: https://github.com/HaikuArchives/Eventual/blob/1433b19ca24d7217ae104e5a7bc075979cbce701/src/Libraries/Activity/SoundSetupView.cpp#L2
[2]: https://github.com/HaikuArchives/Eventual/blob/1433b19ca24d7217ae104e5a7bc075979cbce701/src/PreferencesPreflet/License.txt#L1

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,64 @@
...
+LICENSE="MIT"
```
alaviss: Not only MIT, but also [BSD (3-clause)](https://github.com/HaikuArchives/Eventual/blob/1433b19ca24d7217ae104e5a7bc075979cbce701/src/PreferencesPreflet/License.txt)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,64 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	cmd:sh
```
alaviss: This is included by default, so it's not necessary

```diff
-	cmd:sh

File: haiku-apps/eventual/eventual-1.0.recipe [hunk 7]
```diff
@@ -0,0 +1,64 @@
...
+HOMEPAGE="https://sourceforge.net/projects/eventual/"
+COPYRIGHT="2011 Alex Hitech"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="846290b16e76a8a4ffbd955057306a18ac96d5e3"
+SOURCE_URI="https://github.com/HaikuArchives/Eventual/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="e0b9844cebb6ed2cf42f93e592fe2e4ce0b801457b31cbac7ff7f5b5d493f76e"
+SOURCE_FILENAME="Eventual-$srcGitRev.tar.gz"
+SOURCE_DIR="Eventual-$srcGitRev"
+PATCHES="eventual-$portVersion.patchset"
```
alaviss: As the patch is upstreamable, please do so we could drop this line :)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,64 @@
...
+
+INSTALL()
+{
+	mkdir -p $appsDir/Eventual
+	mkdir -p $appsDir/Eventual/lib
+	cp -r 'src/EventEditor/objects/Eventual Editor' $appsDir/Eventual
+	cp -r 'src/EventServer/objects/Eventual Server' $appsDir/Eventual
+	cp -r 'src/PreferencesPreflet/objects/Eventual Preferences' $appsDir/Eventual
+	cp -r src/lib $appsDir/Eventual
+	addAppDeskbarSymlink $appsDir/Eventual
```
fbrosson: Here's a variant which has a small advantage compare to your `INSTALL()`: it hides the lib subdirectory (of the Eventual directory) in the Deskbar menu.
```
INSTALL()
{
	install -d $appsDir/Eventual/lib
	install -t $appsDir/Eventual/lib src/lib/*
	install -t $appsDir/Eventual \
		"src/EventEditor/objects/Eventual Editor" \
		"src/EventServer/objects/Eventual Server" \
		"src/PreferencesPreflet/objects/Eventual Preferences"

	install -d $dataDir/deskbar/menu/Applications/Eventual
	addAppDeskbarSymlink "$appsDir/Eventual/Eventual Editor" Eventual
	addAppDeskbarSymlink "$appsDir/Eventual/Eventual Server" Eventual
	addAppDeskbarSymlink "$appsDir/Eventual/Eventual Preferences" Eventual
}
```
It has a small downside: we are using addAppDeskbarSymlink in an unusual way (i.e. with a slash

I have an error when I launch any of the Eventual apps:
```
Did not succeed to read the preferences from the file! Error = -2147475440
```

bach5000: Hmm...
I'll make the changes and look into this error... after school :(

haiku-apps/eventual/eventual-1.0.recipe
bach5000: @alaviss Would I just type BSD? Or BSD (3-clause)?

fbrosson: If I'm not mistaken, it should be
```
LICENSE="MIT
	BSD (3-clause)"
```
or maybe
```
LICENSE="BSD (3-clause)
	MIT"
```
I don't know if we sort them alphabetically or if we list those licenses which are mostly used in the source. So I guess you can chose whatever order you want.

scottmc:
I would list MIT first, as it's the license they aimed to use for this project. The BSD was likely pulled in by some code they used from elsewhere, but still needs to be listed.

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -2,15 +2,15 @@ SUMMARY="A time management system"
 DESCRIPTION="Eventual is a personal time management system for Haiku. \
 Intended to be a full-featured organizer. "
 HOMEPAGE="https://sourceforge.net/projects/eventual/"
-COPYRIGHT="2011 Alex Hitech"
-LICENSE="MIT"
+COPYRIGHT="2011 Alexey Burshtein \
+2001, 2002 Kevin H. Patterson"
+LICENSE="MIT BSD"
```
fbrosson:
```diff
 COPYRIGHT="2011 Alexey Burshtein \
-2001, 2002 Kevin H. Patterson"
+	2001, 2002 Kevin H. Patterson"
-LICENSE="MIT BSD"
+LICENSE="MIT
+	BSD (3-clause)"
```
(I'm not sure if it's 2, 3, or 4-clause.)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,67 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku. \
+Intended to be a full-featured organizer. "
+HOMEPAGE="https://sourceforge.net/projects/eventual/"
+COPYRIGHT="2011 Alexey Burshtein \
+2001, 2002 Kevin H. Patterson"
+LICENSE="MIT BSD"
```
fbrosson:
```diff
 COPYRIGHT="2011 Alexey Burshtein \
-2001, 2002 Kevin H. Patterson"
+	2001, 2002 Kevin H. Patterson"
-LICENSE="MIT BSD"
+LICENSE="MIT
+	BSD (3-clause)"
```
(I'm not sure if it's 2, 3, or 4-clause.)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,67 @@
...
+REVISION="1"
+srcGitRev="846290b16e76a8a4ffbd955057306a18ac96d5e3"
+SOURCE_URI="https://github.com/HaikuArchives/Eventual/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="e0b9844cebb6ed2cf42f93e592fe2e4ce0b801457b31cbac7ff7f5b5d493f76e"
+SOURCE_FILENAME="Eventual-$srcGitRev.tar.gz"
```
fbrosson:
```diff
-SOURCE_FILENAME="Eventual-$srcGitRev.tar.gz"
+SOURCE_FILENAME="Eventual-$portVersion-$srcGitRev.tar.gz"
```

scottmc:
Upstream has been merged now.

bach5000:
@scottmc Okay 👍

haiku-apps/eventual/eventual-1.0.recipe
bach5000: Whoops.
Didn't see this until now :(

bach5000: @fbrosson It seems to be 3-clause.

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,68 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku. \
+Intended to be a full-featured organizer. "
+HOMEPAGE="https://sourceforge.net/projects/eventual/"
+COPYRIGHT="2011 Alexey Burshtein \
+	   2001-2002 Kevin H. Patterson"
+LICENSE="MIT
+	 BSD (3-clause)"
```
fbrosson: Oops, there are a three unwanted spaces in line 6 and one in line 8. (We like to have only one tab and no spaces in there.)
```diff
 COPYRIGHT="2011 Alexey Burshtein \
-	   2001-2002 Kevin H. Patterson"
+	2001-2002 Kevin H. Patterson"
 LICENSE="MIT
-	 BSD (3-clause)"
+	BSD (3-clause)"
```

bach5000:
@fbrosson Done!

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -47,9 +47,13 @@ BUILD()
...
-	cd EventEditor && make OBJ_DIR=objects && cd ../
-	cd EventServer && make OBJ_DIR=objects && cd ../
-	cd PreferencesPreflet && make OBJ_DIR=objects
+	make -C EventEditor OBJ_DIR=objects
+	make -C EventServer OBJ_DIR=objects
+	make -C PreferencesPreflet OBJ_DIR=objects
+	
+	#cd EventEditor && make OBJ_DIR=objects && cd ../
+	#cd EventServer && make OBJ_DIR=objects && cd ../
+	#cd PreferencesPreflet && make OBJ_DIR=objects
```
fbrosson: Don't forget to remove these 3 lines.

BTW, I've got a small variant:
```
	make OBJ_DIR=objects -C EventEditor
	make OBJ_DIR=objects -C EventServer
	make OBJ_DIR=objects -C PreferencesPreflet
```
But if you prefer to have the `-C` before the `OBJ_DIR=objects`, thats OK for me.

bach5000: Okay :)

bach5000:
Oops.
Forgot to change srcGitRev and hash.

bach5000:
Hmm...
That's strange. Just noticed that the programs work fine in their source folders.
But the symlinks seem to be giving errors?

fbrosson:
* Is it the app itself that creates its default configuration in an empty file? In which case it would be nice to either modify the program to prevent it from creating that empty file, or, alternatively, make it create a config file with a minimal (but non empty) content.

* Or is the default config being created by the recipe? In which case we should just skip that and let the app handle the situation.

* If neither the recipe nor the app itself create an empty config file, then the recommended behaviour would be, imo, to silently hide the error message and start with some default configuration in memory.

bach5000:
@fbrosson When I open any one of the 3 Eventual apps, the error message appears and after that a new empty pref file is made. So I guess this is the app that's doing this.

fbrosson:
OK. Thanks for the test :)
Would like to try and fix this on the upstream project, and then update the srcGitRev before we merge?

Hmm, I'd like to see if I can simplify the `BUILD()` a bit more. Shall we keep your PR open for a few  days before we merge?
Thanks!

bach5000:
@fbrosson Sure.
It'll probably take me a while to solve this error, anyway :)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,68 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku. \
+Intended to be a full-featured organizer. "
+HOMEPAGE="https://sourceforge.net/projects/eventual/"
+COPYRIGHT="2011 Alexey Burshtein \
```
korli: The backslash can be removed.

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,68 @@
...
+
+BUILD()
+{
+	sh makelib.sh
+	cd src/Libraries
+	mkdir -p ../lib
+	for LIBDIR in */ ; do
+		LIBDIR=${LIBDIR%?}
+		NAME=lib$LIBDIR.so
+		LIB=$LIBDIR/objects.x86-cc2-release/$NAME
```
korli: This looks wrong for x86_64. Maybe don't use makelib.sh and instead call "make OBJ_DIR=objects -C $LIBDIR" for each library.

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,68 @@
...
+	"
+
+BUILD()
+{
+	cd src/Libraries
+	mkdir -p ../lib
+	for LIBDIR in */ ; do
+		LIBDIR=${LIBDIR%?}
+		NAME=lib$LIBDIR.so
+		make OBJ_DIR=objects -C LIBDIR
```
fbrosson:
```diff
-		make OBJ_DIR=objects -C LIBDIR
+		make OBJ_DIR=objects -C $LIBDIR
```
(I guess you forgot the `$`)

scottmc:
This ready yet?

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,68 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku. \
+Intended to be a full-featured organizer. "
```
fbrosson:
```diff
 DESCRIPTION="Eventual is a personal time management system for Haiku. \
-Intended to be a full-featured organizer. "
+It is intended to be a full-featured organizer."
```
You can also join the two sentences together (in a single sentence) if you prefer.

fbrosson:
I forgot I wanted to try some polishing in `BUILD()`.
(I'm on it right now.)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,68 @@
...
+		LIBDIR=${LIBDIR%?}
+		NAME=lib$LIBDIR.so
+		make OBJ_DIR=objects -C $LIBDIR
+		LIB=$LIBDIR/objects/$NAME
+		cp $LIB ../EventEditor/$NAME
+		cp $LIB ../EventServer/$NAME
+		cp $LIB ../PreferencesPreflet/$NAME
+		cp $LIB ../lib/$NAME
+	done
+	cd ../
```
fbrosson:
```diff
 BUILD()
 {
-	cd src/Libraries
-	mkdir -p ../lib
-	for LIBDIR in */ ; do
-		LIBDIR=${LIBDIR%?}
-		NAME=lib$LIBDIR.so
-		make OBJ_DIR=objects -C $LIBDIR
-		LIB=$LIBDIR/objects/$NAME
-		cp $LIB ../EventEditor/$NAME
-		cp $LIB ../EventServer/$NAME
-		cp $LIB ../PreferencesPreflet/$NAME
-		cp $LIB ../lib/$NAME
-	done
-	cd ../
+	cd src
+	mkdir -p lib
+	for LIBDIR in Libraries/* ; do
+		make OBJ_DIR=objects -C $LIBDIR
+		NAME=lib${LIBDIR#Libraries/}.so
+		mv $LIBDIR/objects/$NAME lib
+		ln -s ../lib/$NAME EventEditor
+		ln -s ../lib/$NAME EventServer
+		ln -s ../lib/$NAME PreferencesPreflet
+	done
 	make OBJ_DIR=objects -C EventEditor
 	make OBJ_DIR=objects -C EventServer
 	make OBJ_DIR=objects -C PreferencesPreflet
 }
```
@bach5000, here's a variant which produces the exact same package as your current `BUILD()`.
There are 2 small advantages: it's 3 lines smaller, and it does not make 4 copies of each lib. It just moves it to src/lib/ and creates symlinks so that each app can be built.
What do you think?

(Sorry, BTW, for this late comment.)

bach5000: @fbrosson,
It would be better to not make 4 copies of each lib, but I only did it because the build instructions (https://github.com/HaikuArchives/Eventual/blob/master/INSTALL) said so. If there are no conflicts building then I suppose we should go ahead and use it 😃 

Also, late comments are fine. I've made plenty late comments multiple times before 😬

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,65 @@
+SUMMARY="A time management system"
+DESCRIPTION="Eventual is a personal time management system for Haiku, \
+intended to be a full-featured organizer. "
```
fbrosson: Oops, you forgot to remove the space near the end of line 3.

Hmm, the comma between `Eventual is a personal time management system for Haiku` and `intended to be a full-featured organizer.` does not sound very nice. What do you think of these possible variants?
```
DESCRIPTION="Eventual is a personal time management system for Haiku and is \
intended to be a full-featured organizer."
```
```
DESCRIPTION="Eventual is a personal time management system intended to be a \
full-featured organizer."
```

(BTW, thanks for the previous update in `BUILD()` :)

haiku-apps/eventual/eventual-1.0.recipe
```diff
@@ -0,0 +1,65 @@
...
+LICENSE="MIT
+	BSD (3-clause)"
+REVISION="1"
+srcGitRev="2fd70537847d3b9712e96ebbc4ff37d5e961b480"
+SOURCE_URI="https://github.com/HaikuArchives/Eventual/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="229cd7fa206da0606d32fc0db1f44a5eee20f0e8428e3dec141e4aab272ead68"
+SOURCE_FILENAME="Eventual-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="Eventual-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 ?x86_64"
```
fbrosson: Eventual builds and runs on x86_gcc2, but it fails to build on x86_64 with this error:
```
In file included from ActivityWindow.h:19:0,
                 from ActivityData.cpp:8:
../../Libraries/Category/Category.h:9:21: fatal error: ostream.h: No such file or directory
```
On x86_gcc2, `ostream.h` is in `/system/develop/headers/c++/2.95.3/`.
Does someone know if there is an easy patch that would make the app build on x86_gcc2?
Thanks!

fbrosson: This patch allows to skip the missing ostream.h and go a few steps forward.
```
diff --git a/src/Libraries/Category/Category.h b/src/Libraries/Category/Category.h
index deaaef1..02032bd 100644
--- a/src/Libraries/Category/Category.h
+++ b/src/Libraries/Category/Category.h
@@ -6,7 +6,6 @@
 #define _CATEGORY_H_
 
 #include <iostream>
-#include <ostream.h>
 #include <stdlib.h>
 
 #include <InterfaceKit.h>
diff --git a/src/Libraries/Category/CategoryItem.h b/src/Libraries/Category/CategoryItem.h
index 118f4b4..4ba9779 100644
--- a/src/Libraries/Category/CategoryItem.h
+++ b/src/Libraries/Category/CategoryItem.h
@@ -6,7 +6,6 @@
 #define _CATEGORY_ITEM_H_
 
 #include <iostream>
-#include <ostream.h>
 #include <stdlib.h>
 
 #include <InterfaceKit.h>
-- 
2.16.0
```
But then we hit other issues (still on x86_64):
```
CalendarModulePreferences.cpp:42:1: warning: deprecated conversion from string constant to 'char*' [-Wwrite-strings]
CalendarModulePreferences.cpp: In constructor 'CalendarModulePreferences::CalendarModulePreferences(const BString&, BMessage*)':
CalendarModulePreferences.cpp:538:43: warning: cast to pointer from integer of different size [-Wint-to-pointer-cast]
    ( this->weekends )->AddItem( ( void* ) tempUint32 );
```
So let's forget about x86_64. We can probably leave it as is (`?x86`).
(I'll delete this message and the previous one.)

fbrosson:
@bach5000, I will merge after you remove the white space near the end of line 3.
(There is no urgency, of course.)

haiku-apps/eventual/eventual-1.0.recipe
bach5000: Perhaps the first one? We shouldn't forget Haiku!!!

pulkomandy:
On Wed, Jan 24, 2018 at 02:37:53PM -0800, Bach Nguyen wrote:
> @fbrosson,
> It would be better to not make 4 copies of each lib, but I only did it because the build instructions (https://github.com/HaikuArchives/Eventual/blob/master/INSTALL) said so. If there are no conflicts building then I suppose we should go ahead and use it 😃 

symlinks would make the package smaller and behave the same.

fbrosson:
Actually, the symlinks we create during the build don't get installed. It's just a trick to keep the build directory as small as possible.
Indeed, the variant did not have any impact at all on the contents of the archive. Both the previous `BUILD()` and the the variant (which was adopted ;) produce the exact same files, with the exact same contents, and even the exact same set of symlinks.

BTW, what led me to suggest the variant was just a more or less futile motivation: the previous `BUILD()` was following the instructions provided by upstream, but these instructions were not optimal.

(I wrote myself, a long time ago, a hpkg2md5 script which creates a directory, calls `package extract` in there, and then computes the MD5 checksums of each file in the hpkg and puts all these checksums (with the paths of their files) in a file with an ".md5" extension. It also sorts that file, which makes it suitable for comparisons. I often use this script, in particular when I want to make sure that dropping a build dependency in a recipe does not have any impact, or, in the case of the recipe of this PR, to make sure the `BUILD()` variant was absolutely equivalent to the previous :)

PR #2089, opened 2018-01-09
Chessx: Initial recipe

tts2k:


extrowerk:
What the command-not-found recipe doing here?

tts2k:
@extrowerk What recipe ? i don't see anything. You must be drunk

extrowerk:
tts2k, it was not funny and i got no screenshot, but i'm sure, github presented me the comman_not_found recipe and patchset too in your commit.
Ofc, --force helps a lot in rewriting history, or maybe it was just a caching bug either on GitHub site or on my site.

tts2k:
@extrowerk I was just joking, don't take it too serious. Sorry if it insulted you.
It was an issue won my side and i fixed it by dropping all unnecessary commit

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,61 @@
+SUMMARY="A free and open source chess database application"
+DESCRIPTION="With ChessX you can operate your collection of chess games \
+in many ways: browse, edit, add, organize, analyze, etc."
+HOMEPAGE="https://chessx.sourceforge.net/"
+COPYRIGHT="2005 Michal Rudolf"
+LICENSE="GNU GPL v2"
+REVISION='1'
+SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx/$portVersion/chessx-$portVersion.tgz"
```
fbrosson: You can use SourceForge's redirector service:
```diff
-SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx/$portVersion/chessx-$portVersion.tgz"
+SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx-$portVersion.tgz"
```
This is handy because it automatically finds the right subdirectory for you. So if the owner moves the archive to another directory, SF will find it and serve it, after an automatic redirect.

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,61 @@
...
+	qmake
+	make $jobArgs
+}
+
+INSTALL()
+{
+	install -d $appsDir $developDocDir
+	install -t $appsDir -m755 release/chessx
+	install -t $developDocDir -m644 ChangeLog TODO
+	addAppDeskbarSymlink $appsDir/chessx chessx
```
fbrosson:
```diff
-	addAppDeskbarSymlink $appsDir/chessx chessx
+	addAppDeskbarSymlink $appsDir/chessx ChessX
```

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,61 @@
...
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libQt5Xml$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libgl$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:linguist$secondaryArchSuffix >= 5
+	cmd:make
+	cmd:qmake
```
fbrosson:
```diff
-	cmd:qmake
+	cmd:qmake$secondaryArchSuffix
```
Explanation: qmake is not available on x86_gcc2 primary arch, and the runtime provided by qt5_x86_devel is in /system/bin/x86/qmake and has a qmake-x86 symlink in /system/bin/ pointing to x86/qmake.
So this means qmake needs a `$secondaryArchSuffix` on x86_gcc2, otherwise the recipe will only build on x86_64.

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,64 @@
+SUMMARY="A free and open source chess database application"
+DESCRIPTION="With ChessX you can operate your collection of chess games \
+in many ways: browse, edit, add, organize, analyze, etc."
+HOMEPAGE="https://chessx.sourceforge.net/"
+COPYRIGHT="2005 Michal Rudolf"
+LICENSE="GNU GPL v2"
+REVISION='1'
+SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx/$portVersion.tgz"
```
fbrosson:
```diff
-SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx/$portVersion.tgz"
+SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx-$portVersion.tgz"
```
( `/` -> `-` )

(Everything else is OK)

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,64 @@
+SUMMARY="A free and open source chess database application"
+DESCRIPTION="With ChessX you can operate your collection of chess games \
+in many ways: browse, edit, add, organize, analyze, etc."
+HOMEPAGE="https://chessx.sourceforge.net/"
+COPYRIGHT="2005 Michal Rudolf"
```
fbrosson: Sorry, I had noticed this earlier but forgot to leave a comment about it.
The program is still being updated, so I guess you should replace `2005` by `2005-2017` (or `2007-2017` if you are not sure about 2005) in `COPYRIGHT`.
Thanks!

fbrosson: Now I know why I forgot to leave a comment about the `COPYRIGHT` text.
I wanted to see which authors we could add, from those listed in http://chessx.sourceforge.net/node/7 but there are no years associated to each author, so we would have to either dig this or leave an incomplete text.

Anyone has an idea of what we could put in `COPYRIGHT`? Thanks!

fbrosson: After all, shall we mention all the authors (and key contributors) together?
```
COPYRIGHT="2005-2017 Michal Rudolf, Jens Nissen, Marius Roets, Sean \
Estabrooks, Rico Zenklusen, William Hoggarth, Ejner Borgbjerg, Bruno Rizzuti"
```
I don't think anyone would disagree with this, but if that happens I'm open to any suggestion :)
Thanks!

pulkomandy: It is not something we can decide ourselves. Either each contributor wants to keep his own copyright on his own work, or he reassigns it to someone else (when/where that is possible, for example in France it is not legally possible to do it). We should just follow on what upstream says, either at the header of their LICENCE file, or in their sourcecode headers. It is up to them to provide us with accurate information.

fbrosson: OK. Thanks for your message.

I understand we should dig each header to find the information. That's OK for me, but I don't have the time to do that, and I wouldn't feel comfortable asking @tts2k to do that too.

Assuming nobody wants to search for the correct text, would we need to postpone the merge until we have that information?

Is there something we could put in `COPYRIGHT` for the merge before someone fixes it?

After all, what we have in `COPYRIGHT` is only informative. It's not supposed to be an authority.

What would you do here? Can't we make an exception and merge it with an incomplete (or incorrect) content?

pulkomandy: > It is up to them to provide us with accurate information.

So just ask upstream if they didn't bother to do so until now.

tts2k: Found it !
https://sourceforge.net/p/chessx/code/2415/tree//trunk/src/database/index.cpp#l3

alaviss: So as of version 1.4.6 we've these copyright holders:

- 2012-2016 Jens Nissen
- 2011 Nokia Corporation and/or it's subsidiary(-ies)
- 2005-2007, 2010 Michal Rudof
- 2010 P. Sereno
- 2014 Markus Kolb
- 1999-2004 Shane Hudson
- 1998-2005 Gilles Vollant
- 2005-2011 Sergey A. Tachenov
- 2005 Ejner Borgbjerg
- 2005-2007 Marius Roets
- 2005, 2006, 2007 Heinz R. Hopfgartner
- 2008 Aliaksandr Antonik

Extracted with this 15s craft

    find -type f -exec grep -i copyright '{}' + | grep -v .svn | grep -v i18n | awk -F: '!a[$1]++' | less

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,64 @@
...
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Multimedia$secondaryArchSuffix
+	lib:libQt5Network$secondaryArchSuffix
+	lib:libQt5PrintSupport$secondaryArchSuffix
+	lib:libQt5Svg$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	lib:libQt5Xml$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libgl$secondaryArchSuffix
+	"
```
alaviss: Alphabetical order, please

```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
+	lib:libgl$secondaryArchSuffix
 	lib:libQt5Gui$secondaryArchSuffix
 	lib:libQt5Multimedia$secondaryArchSuffix
 	lib:libQt5Network$secondaryArchSuffix
 	lib:libQt5PrintSupport$secondaryArchSuffix
 	lib:libQt5Svg$secondaryArchSuffix
 	lib:libQt5Widgets$secondaryArchSuffix
 	lib:libQt5Xml$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
-	lib:libgl$secondaryArchSuffix
 	"

File: games-board/chessx/chessx-1.4.6.recipe [hunk 7]
```diff
@@ -0,0 +1,64 @@
...
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Multimedia$secondaryArchSuffix
+	devel:libQt5Network$secondaryArchSuffix
+	devel:libQt5PrintSupport$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libQt5Xml$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libgl$secondaryArchSuffix
+	"
```
alaviss:
```diff
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	devel:libgl$secondaryArchSuffix
 	devel:libQt5Gui$secondaryArchSuffix
 	devel:libQt5Multimedia$secondaryArchSuffix
 	devel:libQt5Network$secondaryArchSuffix
 	devel:libQt5PrintSupport$secondaryArchSuffix
 	devel:libQt5Svg$secondaryArchSuffix
 	devel:libQt5Widgets$secondaryArchSuffix
 	devel:libQt5Xml$secondaryArchSuffix
 	devel:libz$secondaryArchSuffix
-	devel:libgl$secondaryArchSuffix
 	"

File: games-board/chessx/chessx-1.4.6.recipe [hunk 8]
```diff
@@ -0,0 +1,64 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	install -d $appsDir $developDocDir
+	install -t $appsDir -m755 release/chessx
+	install -t $developDocDir -m644 ChangeLog TODO
+	addAppDeskbarSymlink $appsDir/chessx ChessX
+}
```
alaviss:
```diff
 INSTALL()
 {
-	install -d $appsDir $developDocDir
+	install -d $appsDir $docDir
 	install -t $appsDir -m755 release/chessx
-	install -t $developDocDir -m644 ChangeLog TODO
+	install -t $docDir -m644 ChangeLog TODO
 	addAppDeskbarSymlink $appsDir/chessx ChessX
 }

File: games-board/chessx/chessx-1.4.6.recipe [hunk 9]
```diff
@@ -0,0 +1,64 @@
+SUMMARY="A free and open source chess database application"
+DESCRIPTION="With ChessX you can operate your collection of chess games \
+in many ways: browse, edit, add, organize, analyze, etc."
+HOMEPAGE="https://chessx.sourceforge.net/"
+COPYRIGHT="2005 Michal Rudolf"
+LICENSE="GNU GPL v2"
```
alaviss: Also [BSD (3-clause)](https://sourceforge.net/p/chessx/code/2415/tree//trunk/src/database/downloadmanager.cpp#l7)

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -3,9 +3,18 @@ DESCRIPTION="With ChessX you can operate your collection of chess games \
...
+	2005-2007 Heinz R. Hopfgartner
+	2005-2010 Michal Rudolf
+	2005-2011 Sergey A. Tachenov
 	2007 Rico Zenklusen
-	2007-2009 Michal Rudolf
+	2008 Aliaksandr Antonik
+	2010 P. Sereno
+	2011 Nokia Corporation and/or it's subsidiary(-ies)
+	2012-2016 Jens Nissen	
 	"
```
fbrosson:
```diff
-COPYRIGHT="
-	2005-2006 Michal Rudolf
+COPYRIGHT="2005-2006 Michal Rudolf
 	1998-2005 Gilles Vollan
```
```diff
 	2011 Nokia Corporation and/or it's subsidiary(-ies)
-	2012-2016 Jens Nissen	
- 	"
+	2012-2016 Jens Nissen"
 LICENSE="GNU GPL v2"
```
Can you please make the changes above, in order to use the recommended style for `COPYRIGHT`?
Source: https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#a-simple-example-using-the-recommended-style-for-a-multi-line-copyright
Thanks!

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,76 @@
...
+	2005 Ejner Borgbjerg
+	2005-2007 Marius Roets
+	2005-2007 Heinz R. Hopfgartner
+	2005-2010 Michal Rudolf
+	2005-2011 Sergey A. Tachenov
+	2007 Rico Zenklusen
+	2008 Aliaksandr Antonik
+	2010 P. Sereno
+	2011 Nokia Corporation and/or it's subsidiary(-ies)
+	2012-2016 Jens Nissen"
```
alaviss: After some more time crafting script, here's the end result. Now if only someone could do the same for license...

- 2005-2016 ChessX developers
- 2014 Digia Plc and/or its subsidiary(-ies).
- 2005-2006 Ejner Borgbjerg
- 1998-2005 Gilles Vollant
- 2007 Heinz R. Hopfgartner
- 1990-2000 Info-ZIP
- 2012-2016 Jens Nissen
- 2005 Kamil Przybyla
- 2005-2007 Marius Roets
- 2014 Markus Kolb
- 2005-2010 Michal Rudolf
- 2011 Nokia Corporation and/or its subsidiary(-ies).
- 2006-2007 Pascal Georges
- 2010 P. Sereno
- 2007 Rico Zenklusen
- 2006-2007 Sean Estabrooks
- 2005-2011 Sergey A. Tachenov
- 1999-2004 Shane Hudson
- 2003 Sune Fischer
- 2006 Tobias Koenig
- 2005-2006 William Hoggarth

fbrosson:
I've got a small suggestion to do, before the merge.

fbrosson:
Oops, I'm on the wrong PR :)

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,76 @@
...
+LICENSE="GNU GPL v2
+	BSD (3-clause)"
```
alaviss: As of licenses:

- LGPL2, Zlib in `src/quazip`
- LGPL2.1 with Digia Qt LGPL Exception version 1.1 (you would have to dig this up yourself, sorry) in `src/gui/textedit.cpp`

Hopefully that's all

fbrosson:
The recipe is working fine, and everything looks OK, including the metadata (`COPYRIGHT` and `LICENSE`), so I guess I can merge :)

If you also think everything is OK then please add a 👍 and I'll merge.
Thanks!

games-board/chessx/chessx-1.4.6.recipe
```diff
@@ -0,0 +1,88 @@
...
+	Digia Qt LGPL Exception v1.1"
+REVISION='1'
+SOURCE_URI="https://downloads.sourceforge.net/chessx/chessx-$portVersion.tgz"
+CHECKSUM_SHA256="ad951910a3bd53f5655f9e85101e7969b536078473e09a2a7dded6ef171a68ed"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	chessx = $portVersion
```
fbrosson:
```diff
 PROVIDES="
-	chessx = $portVersion
+	chessx$secondaryArchSuffix = $portVersion
-	app:chessx = $portVersion
+	app:chessx$secondaryArchSuffix = $portVersion
 	"
```
Does anyone know if it's correct to also add the `$secondaryArchSuffix` after `app:chessx`?

alaviss: @fbrosson Doesn't [look][0] like so to me

[0]: https://github.com/haikuports/haikuporter/blob/2b7ba23e93e799bb9ceeab552266cc1e35b26c5f/HaikuPorter/Port.py#L1205

fbrosson: I might be wrong but this does not seem to be related to the `app:*` elements but to the `$appsDir` path only.

Well, I just did some metrics (using find, grep, sort, cut and comm, which are great for that as we know :)

Total number of recipes with an `app:` and 2nd arch | 132
--- | ---
`app:` **_with_** `$secondaryArchSuffix` | 78
`app:` **_without_** `$secondaryArchSuffix` | 54

BTW, it's funny, because [qemacs](https://github.com/haikuports/haikuports/tree/master/app-editors/qemacs) has the `$secondaryArchSuffix` for its `app:`, while [xemacs](https://github.com/haikuports/haikuports/tree/master/app-editors/xemacs) does not.

There does not seem to be a real trend, but I think it would be safer to recommend using the `$secondaryArchSuffix` for `app:` elements.

(In my metrics above I did not count my "copy" of chessx.)

pulkomandy: The $secondaryArchSuffix makes sense only if the app is available both for the secondary and primary archs at the same time. This is an unlikely case. If the app is only available for one, then it makes sense to not use the $secondaryArchSuffix. This way, people can "pkgman install app:chessx" and not worry about which compiler was used.

fbrosson: OK, thanks for the explanation. Now I understand why there is no real trend.

I do agree that, on x86_gcc2, we would not be able to run `pkgman install app:chessx`, but with a simple `pkgman search app:chessx` we would get the answer.
For instance, this is what we get with notepadqq on x86_gcc2:
```
pkgman search app:notepadqq
Status  Name           Description                        
----------------------------------------------------------
        notepadqq_x86  Multiplatform Notepad++-like editor
```

IMO there is something worse than having to call to `pkgman install app:notepadqq_x86` instead of just `pkgman install app:notepadqq`:
It's having to remember that if your first attempt did not succeed, then this could be because the package you are trying to install... has the suffix on its `app:`.

Don't you agree that adding the suffix increases the consistency?

BTW, among the 78 recipes which already have the suffix for their `app:`, 65 are broken on x86_gcc2 primary arch, 4 are marked as untested (on that same arch), and 7 are marked as tested (also on that same arch.)
So, among those 78 recipes, the only which would be correctly using the suffix, according to your rationale, would be:
* these 4 recipes, which are marked as untested:
```
kde-base/kolourpaint/kolourpaint-17.11.80.recipe
kde-base/kstars/kstars-2.8.8.recipe
media-gfx/krita/krita-3.3.1.recipe
media-gfx/nomacs/nomacs-3.8.0.recipe
```
* and these 7 other recpes, which are marked as tested:
```
app-editors/qemacs/qemacs-0.4.1~pre20170225.recipe
games-puzzle/fish-fillets/fish_fillets-1.0.1.recipe
games-strategy/homeworld_sdl/homeworld_sdl-0.6.recipe
haiku-apps/cdplayer/cdplayer-1.0.recipe
haiku-apps/clockwerk/clockwerk-1.0~git.recipe
haiku-apps/paladin/paladin-1.4~git.recipe
haiku-games/puri/puri-0.3.9.1.recipe
haiku-games/superfreecell/superfreecell-0.1.0.recipe
mail-client/mailnews/mailnews-2.0.0.25.recipe
```
**All** the other **65** recipes (which have the suffix for their `app:` elements) are **broken** on x86_gcc2 primary arch (and yet have the suffix for their `app:` elements.)

Now, in my previous table I wrote that among those recipes which have support for 2nd arch, there are only 54 recipes without the suffix for the `app:` elements, which is less than the number of recipes that do not follow your rationale.

Since your argument is beaten by the metrics above, wouldn't it be easier to just decide that, from now on, all new recipes (with support for secondary arch) should have the suffix on their `app:` elements?

Please don't get me wrong. I'm not trying to fight anyone. You are here since much longer than me, and you also know Haiku and HaikuPorts much better than me.
I'm just trying to shed some light on a (small) inconsistency to improve things a little. Thanks!

pulkomandy: As I said, having an app exist on both arches is a *very* rare case. The only possible reason would be that the app uses add-ons, and you want to be able to use different add-ons that are not all available for gcc2 or for gcc5. I am not aware of any app where this is the case.

So I think the better thing to do is make sure all "app:" recipe are enabled for only one of either x8-_gcc2 (primary) or x86 (secondary), and that they never have a $secondaryArchSuffix. Or if one of them has a reason for having one, it should be written in the recipe in big red letters (ok, we can't do that) in a comment next to it.

fbrosson: OK.
I'm not convinced that droping the suffix for `app:chessx` is a good idea, but can live with that :)
(I'll update my previous review.)

P.S. I guess the reason why we have so many recipes with the suffix, although they are broken on x86_gcc2) is that there are more contributors who think the suffix should be used than contributors who think it shouldn't. But I do admit that this is not a valid argument :)

pulkomandy: Maybe we should explain when to use and when not to use it. I think people think the suffix is mandatory everywhere when using the secondary arch (if that was the case, we wouldn't require recipe writers to add it and the tools would do it on their own).

The $secondaryArchSuffix is used to make sure the entries for primary and secondary arch do not conflict wit each other. This makes sense only if all these conditions are true:
- The recipe builds for both primary and secondary arch
- The result of the build is different in each case and it makes sens to have both versions installed

In all other cases, the suffix should not be used.
For example, the ffmpeg recipe builds for both architectures (in different versions) and provide libraries with the secondaryArchSuffix. But, only the x86 package provides binaries (ffplay, ffmpeg, ffprobe, ...). So, these binaries are provided without $secondaryArchSuffix.

fbrosson: Guess what? You have now convinced me, and I have even found one more argument for you ;)

When we use the commandBinDir "trick", we set commandSuffix to the empty string, and we do so because the `cmd:` elements are in `$prefix/bin` and don't have the suffix (even when wr're on x86_gcc2, since this is the purpose of the commandBinDir trick.)

So we just need to say that `app:` elements follow the same rules as the `cmd:` elements.
Since we are providing `$appsDir/chessx` (and not `$appsDir/chessx_x86`), it is perfectly normal to declare `app:chessx` **without** any suffix.

So congratulations, and thanks for having convinced me!
That wasn't easy, sorry ;D

PR #2091, opened 2018-01-09
Create monsterz

vrindag:


Begasus:
Not good enough, you need to provide the needed requirements to be able to build it, I guess you should file for one of the task to build a recipe at GCI first ...

games-puzzle/monsterz
```diff
@@ -0,0 +1,45 @@
+SUMMARY="A little arcade puzzle game."
+DESCRIPTION="The goal of the game is to create rows of similar monsters, either horizontally or vertically.\
+The only allowed move is the swap of two adjacent monsters, on the condition that it creates a row of three \
+or more. When alignments are cleared, pieces fall from the top of the screen to fill the board again. Chain \
+reactions earn you even more points."
+HOMEPAGE="http://sam.zoy.org/monsterz/"
+COPYRIGHT="?"
+LICENSE="ISC"
+REVISION="1"
+SOURCE_URI="?"
```
fbrosson:
```
SOURCE_URI="?"
```
@vrindag, you can't submit this :)
You're supposed to test the recipe. 
Without a correct `SOURCE_URI` you can't even kick a build!

fbrosson: Sorry, I though you had clicked Submit on the GCI site.
That's OK, then :)

vrindag: I just claimed the task, then I had tons of homework to finish. Sorry!

fbrosson: No problem. Sorry about my inappropriate message.
May I delete this comment and the previous four too? If that's OK for you, you can add a thumbs up and I'll understand.
(This would allow to remove 5 messages in this discussion, and I'd also feel much better without my comment :)

Once again, I'm really sorry about my comment. BTW, you are perfectly right to do your homework before anything else.

vrindag: That's completely fine with me. I just finished my homework, so lots more time for me to build recipes!!!

vrindag: Sorry for the late reply.

fbrosson:
Oops, I can delete my own messages without leaving any traces of it, but when I delete someone else's messages, it leaves a trace of the action. I'll think twice next time.
Anyway, I'm happy you have finished your homework! Good packaging!

vrindag:
In https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines , it says that a matching parent directory must be created on the Gentoo portage. The game Monsterz is available there.(Gentoo users: run emerge monsterz, it’s already in your distribution). Does that mean that I have to add the SOURCE_URI ,the SOURCE_DIR and the SOURCE_FILENAME from there?

vrindag:
Like last time(openclonk), I found a list of libraries that debian uses. Is there any way to check what libraries Haiku has?  And, can I use the linux example to build the recipe?(it, again contains way too many libraries). Is pygame installed by default on Haiku?

https://tests.reproducible-builds.org/debian/rb-pkg/unstable/i386/monsterz.html

fbrosson:
What I generally do is start with a small set of `devel:lib*` dependencies in `BUILD_REQUIRES`. (I only put what I am sure is required, and only if I know the exact names of these libs.)
Then I kick a build and watch carefully the output of the configure script. If it says it cannot find a library, e.g. libcurl, then I run:
```
pkgman search devel:libcurl
pkgman install curl_devel
```
and then I add `devel:libcurl` to `BUILD_REQUIRES` and retry. I do that until the configure script reports that it has found everything it looks for.

Later, when the build succeeds, haikuporter will report any missing dependencies (if you call it with `-S`). So if it tells me that the runtime needs `libcurl.so` (which would not be surprising), I would add `lib:libcurl` to the `REQUIRES`.

This is just an example. I don't know if monsterz needs libcurl.

vrindag:
Thanks

vrindag:
Each time I try to compile the recipe, it fails, saying 'Port' object has no attribute 'recipeKeys'

fbrosson:
To answer [your previous question](https://github.com/haikuports/haikuports/pull/2091#issuecomment-356581303), yes, we need to use the same category and directory name as Gentoo.
For the `SOURCE_URI`, if we package the same version, then it is likely that we will have the same content.
`SOURCE_DIR` is determined by the archive.
`SOURCE_FILENAME` is only required if the name of the archive does not contain the version of the project or a git hash. But it is very welcome, in particular when the name is made of only a git hash and the ".tar.gz" extension.

vrindag:
So, for the SOURCE_URI I should use 
https://gentoobrowse.randomdan.homeip.net/packages/games-puzzle/monsterz , right?

fbrosson:
No, SOURCE_URI should be the link to the tarball on the download page.
(It is used by haikuporter with wget.)

games-puzzle/monsterz
```diff
@@ -4,7 +4,8 @@ The only allowed move is the swap of two adjacent monsters, on the condition tha
 or more. When alignments are cleared, pieces fall from the top of the screen to fill the board again. Chain \
 reactions earn you even more points."
 HOMEPAGE="http://sam.zoy.org/monsterz/"
-COPYRIGHT="<copyright notice>"
+COPYRIGHT="2007 Sam Hocevar, Brendan Rackley, MenTaLguY, Sun Microsystems\
```
fbrosson:
```diff
-COPYRIGHT="2007 Sam Hocevar, Brendan Rackley, MenTaLguY, Sun Microsystems\
+COPYRIGHT="2007 Sam Hocevar, Brendan Rackley, MenTaLguY, Sun Microsystems, \
```

vrindag:
This works on x86_gcc2

games-puzzle/monsterz_game/monsterz-0.7.1.recipe
```diff
@@ -0,0 +1,43 @@
+SUMMARY="A little arcade puzzle game"
+DESCRIPTION="The goal of the game is to create rows of similar monsters, either horizontally or vertically.\
+The only allowed move is the swap of two adjacent monsters, on the condition that it creates a row of three \
+or more. When alignments are cleared, pieces fall from the top of the screen to fill the board again. Chain \
+reactions earn you even more points."
+HOMEPAGE="http://sam.zoy.org/monsterz/"
+COPYRIGHT="2007 Sam Hocevar, Brendan Rackley, MenTaLguY, Sun Microsystems\
+Michael Speck, David White Mike Kershaw"
+LICENSE="WTFPL
```
fbrosson:
```diff
-LICENSE="WTFPL
+LICENSE="WTFPL v2
```
(This is because you are adding the license as "WTFPL v2", not "WTFPL".)

BTW, It's a very good idea to add the " v2".

Regarding the `SOURCE_URI`, I'm quite sure it is recommended to use http or, even better, https, instead of svn.

games-puzzle/monsterz_game/monsterz-0.7.1.recipe
```diff
@@ -0,0 +1,43 @@
+SUMMARY="A little arcade puzzle game"
+DESCRIPTION="The goal of the game is to create rows of similar monsters, either horizontally or vertically.\
+The only allowed move is the swap of two adjacent monsters, on the condition that it creates a row of three \
+or more. When alignments are cleared, pieces fall from the top of the screen to fill the board again. Chain \
+reactions earn you even more points."
```
fbrosson:
```diff
-DESCRIPTION="The goal of the game is to create rows of similar monsters, either horizontally or vertically.\
-The only allowed move is the swap of two adjacent monsters, on the condition that it creates a row of three \
-or more. When alignments are cleared, pieces fall from the top of the screen to fill the board again. Chain \
-reactions earn you even more points."
+DESCRIPTION="Monsterz is a little arcade puzzle game, similar to the famous \
+Bejeweled or Zookeepe. The goal of the game is to create rows of similar \
+monsters, either horizontally or vertically. The only allowed move is the swap \
+of two adjacent monsters, on the condition that it creates a row of three or \
+more. When alignments are cleared, pieces fall from the top of the screen to \
+fill the board again. Chain reactions earn you even more points."
```

fbrosson:
Hmm, the game uses inkscape to create png files from its svg, and we don't have inkscape on HaikuPorts.
Moreover, the Makefile will need some changes. This recipe is not going to be easy.

games-puzzle/monsterz/monsterz-0.7.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+	"
+	
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/Monsterz
```
fbrosson: OK, all the files are now in their correct directories. That's fine.

Now, the hardest part: fixing INSTALL.
Because it currently does not install anything. It just creates an empty directory and a symlink to an inexisting file.
But I'm quite sure you will also need to pass some arguments to make in BUILD, like `prefix=$prefix`.

BTW, if you wish to create a Monsterz directory in $appsDir, you can use this variant:
```diff
-	mkdir -p $appsDir/Monsterz
+	install -d $appsDir/Monsterz
```

vrindag:
In the Makefile, will I need to replace Inkscape with something else?

fbrosson:
I don't know. Maybe it can build without inkscape.

What you might try, if there are not too many SVG files, is do the conversion yourself on another OS and then add the PNG files to an "additional-files" subdirectory. There are quite a few recipes which also have such a subdirectory (in the haikuports tree) so you can look how we set ADDITIONAL_FILES and how we use it.

But before you start making any conversion, I'd recommend that you wait for other opinions about this, because I'm not sure it would be OK to merge png files in that "additional-files" subdirectory.

vrindag:
Ok, so should I fix INSTALL first?

pulkomandy:
I think it should be possible to use nanosvgtranslator and the "translate" command to convert SVG to PNG in Haiku. This could be included in the buildsystem, maybe?

mmlr:
Adding binary files to HaikuPorts should be avoided because it will bloat the repository. Instead they should be put in an archive or separate repository and then fetched as an additional source.

Begasus:
Closing this one as no activity, new one created at https://github.com/haikuports/haikuports/pull/3587

PR #2097, opened 2018-01-10
zziplib : enable for x86_64

extrowerk:


dev-libs/zziplib/zziplib-0.13.67.recipe
```diff
@@ -5,13 +5,13 @@ HOMEPAGE="http://zziplib.sourceforge.net/"
 COPYRIGHT="1999-2010 Guido Draheim"
 LICENSE="GNU LGPL v2.1
 	MPL v1.1"
-REVISION="2"
+REVISION="3"
```
fbrosson: No need to bump REVISION here, if it's just to enable x86_64.
(The build servers will see that there is no package yet for that arch and will kick the build.)

PR #2099, opened 2018-01-10
SageBrush: new recipe

Oyami-Srk:


haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+srcGitRev="d9b3945cb78c7070aa8b48d2e0bc6a2007fa3601"
+SOURCE_URI="https://github.com/HaikuArchives/SageBrush/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="d152e0d572a8eb0b8fc39a531cc1466827a9b96526bc5669dd181ef77ed32940"
+SOURCE_FILENAME="SageBrush-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="SageBrush-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 !x86_64"
+
+PROVIDES="
+	sagebrush$secondaryArchSuffix = $portVersion
```
extrowerk: Currently you have no secondary arch defined in your recipe, and this won't build with gcc5 so you can remove all the $secondaryArchSuffix safely.

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,54 @@
...
+	gcc -fPIC -shared -o libsagebrush.so -Isupport -Iinterface \
+		support/*.cpp interface/*.cpp
+}
+
+INSTALL()
+{
+	mkdir -p $libDir
+	cp -a libsagebrush.so $libDir
+	mkdir -p $includeDir
+	cp support/*.h $includeDir
```
extrowerk: I'mnot sure if it is a good idea to put all the header files to the top level in the $includeDir. Maybe you can create a sagebrush folder for them. Is anything depends on it?

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
extrowerk: Or you can introduce it like:
SECONDARY_ARCH="!x86"
in the line 14.

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
Oyami-Srk: I think it shouldn't be a dependence. Putting header files like this makes test projects build successfully without specifing an include path or changing the codes manually.

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
...
+	"
+REQUIRES_devel="
+	sagebrush$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
```
extrowerk: then you have to add secondaryarch to gcc too.

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
...
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="d9b3945cb78c7070aa8b48d2e0bc6a2007fa3601"
+SOURCE_URI="https://github.com/HaikuArchives/SageBrush/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="d152e0d572a8eb0b8fc39a531cc1466827a9b96526bc5669dd181ef77ed32940"
+SOURCE_FILENAME="SageBrush-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="SageBrush-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 !x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 !x86_64"
```
extrowerk: remove x86_64 please, there is no x86_64as secarch, just x86

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
...
+DESCRIPTION="SageBrush is an interface library for image editor. It provides \
+a kit to build an image editor easily without starting from scratch."
+HOMEPAGE="https://github.com/HaikuArchives/SageBrush"
+COPYRIGHT="2012 Vladislav Burundukov"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="d9b3945cb78c7070aa8b48d2e0bc6a2007fa3601"
+SOURCE_URI="https://github.com/HaikuArchives/SageBrush/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="d152e0d572a8eb0b8fc39a531cc1466827a9b96526bc5669dd181ef77ed32940"
+SOURCE_FILENAME="SageBrush-$portVersion-$srcGitRev.tar.gz"
```
extrowerk: you can remove the portversion here

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
...
+DESCRIPTION="SageBrush is an interface library for image editor. It provides \
+a kit to build an image editor easily without starting from scratch."
+HOMEPAGE="https://github.com/HaikuArchives/SageBrush"
+COPYRIGHT="2012 Vladislav Burundukov"
+LICENSE="MIT"
+REVISION="1"
+srcGitRev="d9b3945cb78c7070aa8b48d2e0bc6a2007fa3601"
+SOURCE_URI="https://github.com/HaikuArchives/SageBrush/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="d152e0d572a8eb0b8fc39a531cc1466827a9b96526bc5669dd181ef77ed32940"
+SOURCE_FILENAME="SageBrush-$srcGitRev.tar.gz"
```
fbrosson:
```diff
-SOURCE_FILENAME="SageBrush-$srcGitRev.tar.gz"
+SOURCE_FILENAME="SageBrush-$portVersion-srcGitRev.tar.gz"
```
This is highly recommended (although not strictly required.)

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
+SUMMARY="An interface library for image editor"
```
bach5000: SUMMARY="An interface library for image editors"

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
...
+DESCRIPTION="SageBrush is an interface library for image editor. It provides \
```
bach5000: DESCRIPTION="SageBrush is an interface library for image editors. It provides \

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
```diff
@@ -0,0 +1,55 @@
...
+a kit to build an image editor easily without starting from scratch."
```
bach5000: a kit to build image editors easily without starting from scratch."

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
alaviss: It's [somewhat required][0]

> ### Archive name
>
>Our policy specifies that all archives' names should be in <port name>-<port version> format. Again, most packages are named like that. For those that are not, use SOURCE_FILENAME attribute.

Also, as this project have no release, shouldn't the port version have `~git` suffix?

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#archive-name

haiku-libs/sagebrush/sagebrush-1.0.0.recipe
fbrosson: No, we should not remove the `-$portVersion` here.
Just adding a comment here so that we understand why the `-$portVersion` is back.
See [this comment](https://github.com/haikuports/haikuports/pull/2099#discussion_r160833040) from @alaviss.

BTW, as @alaviss also suggests in the same comment, we should also rename the recipe to: `sagebrush-1.0.0~git.recipe`.
Well, I'd say that if we don't have any release yet, maybe we could even replace the "1.0.0" by a single "0", thus renaming the recipe to `sagebrush-0~git.recipe`.
@alaviss (and others), what do you think?

fbrosson:
> @fbrosson Should we change the version to `1.0.0~git`?

@alaviss, to answer [that question](https://github.com/haikuports/haikuports/pull/2099#pullrequestreview-88074030) I'd say we could probably even change it to just `0~git`. (see [a comment I wrote recently](https://github.com/haikuports/haikuports/pull/2099#discussion_r160876832) :)

haiku-libs/sagebrush/sagebrush-0~git.recipe
```diff
@@ -56,11 +56,16 @@ INSTALL()
 
 TEST()
 {
-	gcc tools/Property\ Editor/*.cpp -Iinterface -Isupport -L. -lbe -lsagebrush
+	gcc "tools/Property Editor"/*.cpp -Iinterface -Isupport -L. -lbe -lsagebrush
+	LIBRARY_PATH=.:$LIBRARY_PATH ./PropertyWindow
```
fbrosson: I think we should avoid putting `.` in `LIBRARY_PATH`.
If there are libs in the root directory of the build, then this would be better imo:
```diff
-	LIBRARY_PATH=.:$LIBRARY_PATH ./PropertyWindow
+	LIBRARY_PATH="$sourceDir${LIBRARY_PATH:+:$LIBRARY_PATH}" ./PropertyWindow
```
You can even insert
```
	export LIBRARY_PATH="$sourceDir${LIBRARY_PATH:+:$LIBRARY_PATH}"
```
at the beginning of the `TEST()` function and never have to worry about it anymore. In particular, there is no more need to add `../..` to the `LIBRARY_PATH` later.

haiku-libs/sagebrush/sagebrush-0~git.recipe
```diff
@@ -0,0 +1,73 @@
...
+	mkdir -p $includeDir/sagebrush
+	cp support/*.h $includeDir/sagebrush
+	cp interface/*.h $includeDir/sagebrush
+	prepareInstalledDevelLib libsagebrush
+	packageEntries devel $developDir
+}
+
+TEST()
+{
+	export LIBRARY_PATH="$sourceDir:$LIBRARY_PATH"
```
fbrosson: This not 100% correct. It works, but if LIBRARY_PATH is not already defined, then having a `:` at the end has the same effect, if I'm not mistaken, as having a `:.` at the end.
So, even if this looks a bit more complicated, you should do this change:
```diff
-	export LIBRARY_PATH="$sourceDir:$LIBRARY_PATH"
+	export LIBRARY_PATH="$sourceDir${LIBRARY_PATH:+:$LIBRARY_PATH}"
```

Oyami-Srk: Thanks for the instruction!

haiku-libs/sagebrush/sagebrush-0~git.recipe
```diff
@@ -0,0 +1,73 @@
...
+		support/*.cpp interface/*.cpp
+}
+
+INSTALL()
+{
+	mkdir -p $libDir
+	cp -a libsagebrush.so $libDir
+	mkdir -p $includeDir/sagebrush
+	cp support/*.h $includeDir/sagebrush
+	cp interface/*.h $includeDir/sagebrush
```
fbrosson:
```diff
 INSTALL()
 {
-	mkdir -p $libDir
+	install -d $libDir $includeDir/sagebrush
-	cp -a libsagebrush.so $libDir
+	install -t $libDir libsagebrush.so
-	mkdir -p $includeDir/sagebrush
-	cp support/*.h $includeDir/sagebrush
-	cp interface/*.h $includeDir/sagebrush
+	install -t $includeDir/sagebrush support/*.h interface/*.h
 	prepareInstalledDevelLib libsagebrush
 	packageEntries devel $developDir
 }
```
(Cool, isn't it ?)

PR #2103, opened 2018-01-11
Cutter: new recipe

extrowerk:


dev-util/cutter/cutter-1.1.recipe
```diff
@@ -0,0 +1,117 @@
...
+winedbg, windbg)
+* Run on Linux, *BSD, Windows, OSX, Android, iOS, Solaris and Haiku
+* Perform forensics on filesystems and data carving
+* Be scripted in Python, Javascript, Go and more
+* Support collaborative analysis using the embedded webserver
+* Visualize data structures of several file types
+* Patch programs to uncover new features or fix vulnerabilities
+* Use powerful analysis capabilities to speed up reversing
+* Aid in software exploitation"
+HOMEPAGE="https://github.com/radareorg/cutter/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/radareorg/cutter/"
+HOMEPAGE="https://github.com/radareorg/cutter"
```
(The trailing slash is not needed here.)

I haven't tested the recipe, but everything else looks OK :)

fbrosson:
```diff
-HOMEPAGE="https://github.com/radareorg/cutter/"
+HOMEPAGE="https://github.com/radareorg/cutter"
```
(The trailing slash is not needed here.)

I haven't tested the recipe, but everything else looks OK :)

korli:
This one fails on x86_gcc2, see http://vmpkg.haiku-os.org/master/x86_gcc2/buildruns/718/builds/26934.log

CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
RADARE2_INCLUDE_DIRS
   used as include directory in directory /sources/cutter-1.1/src
   used as include directory in directory /sources/cutter-1.1/src
   used as include directory in directory /sources/cutter-1.1/src

PR #2105, opened 2018-01-12
MemoChip: Initial recipe

tts2k:
- License is `unknown` because i haven't found any license for this
- The package build fine with a placeholder license
- The app opened up but doesn't seems to be functioning

### DIFF
--- a/haiku-apps/memochip/memochip-1.3~git.recipe
+++ b/haiku-apps/memochip/memochip-1.3~git.recipe
@@ -10,7 +10,7 @@ CHECKSUM_SHA256="14d0bb56c469e369b553e2b61a78f8c7597198a9f218e242f8468b8140d4671
 SOURCE_FILENAME="MemoChip-$portVersion-$srcGitRev.tar.gz"
 SOURCE_DIR="MemoChip-$srcGitRev"
 
-ARCHITECTURES="?x86_gcc2 x86_64"
+ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
 	MemoChip = $portVersion
@@ -39,5 +39,5 @@ INSTALL()
 	install -d $appsDir
 	install -t $appsDir -m755 objects/MemoChip
 
-	addAppDeskbarSymlink $appsDir/MemoChip MemoChip
+	addAppDeskbarSymlink $appsDir/MemoChip
 }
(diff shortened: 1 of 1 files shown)


haiku-apps/memochip/memochip-1.3~git.recipe
```diff
@@ -0,0 +1,43 @@
...
+{
+	make $jobArgs OBJ_DIR=objects
+}
+
+INSTALL()
+{
+	install -d $appsDir
+	install -t $appsDir -m755 objects/MemoChip
+
+	addAppDeskbarSymlink $appsDir/MemoChip MemoChip
```
fbrosson:
```diff
-	addAppDeskbarSymlink $appsDir/MemoChip MemoChip
+	addAppDeskbarSymlink $appsDir/MemoChip
```
(The last paramter is not required, since we are using the same name.)

haiku-apps/memochip/memochip-1.3~git.recipe
```diff
@@ -0,0 +1,43 @@
...
+COPYRIGHT="2000 Hideki Naito(bo0wy)"
+LICENSE="unknown"
+REVISION="1"
+srcGitRev="f7d08de2d71f9d73ea89aeda627fd24871ce3512"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="14d0bb56c469e369b553e2b61a78f8c7597198a9f218e242f8468b8140d46710"
+SOURCE_FILENAME="MemoChip-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="MemoChip-$srcGitRev"
+
+ARCHITECTURES="?x86_gcc2 x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="?x86_gcc2 x86_64"
+ARCHITECTURES="x86_gcc2 x86_64"
```
(I've just tested it on x86_gcc2.)

fbrosson:
The recipe is OK and the app builds fine on all architectures. It is also working, but something is not OK: the 2 buttons in the main window have no effect.

But since the license is unknown, the builders will refuse to create a package for it.
So I guess we can merge this as is. We don't even need to mark it as untested.

Issue #2074, opened 2018-01-06
Document $cmakeDirArgs


alaviss:
Looks like we have an [convenience variable](https://github.com/haikuports/haikuporter/blob/2b7ba23e93e799bb9ceeab552266cc1e35b26c5f/HaikuPorter/Port.py#L1189) for cmake also. It would be nice to document this.

PR #2111, opened 2018-01-13
googletest, fix recipe for secondaryArchSuffix

Begasus:
Patch based on their own upstream patch ... https://github.com/google/googletest/pull/1160

### DIFF
--- a/dev-libs/gtest/gtest-1.9.0~git.recipe
+++ b/dev-libs/gtest/gtest-1.9.0~git.recipe
@@ -1,0 +1,97 @@
+SUMMARY="C++ test framework"
+DESCRIPTION="C++ testing utility
+Features:
+* An xUnit test framework
+* Test discovery
+* A rich set of assertions
+* User-defined assertions
+* Death tests
+* Fatal and non-fatal failures
+* Value-parameterized tests
+* Type-parameterized tests
+* Various options for running the tests
+* XML test report generation"
+HOMEPAGE="https://github.com/google/googletest"
+COPYRIGHT="2008-2018 Google Test developers"
+LICENSE="BSD (3-clause)"
+srcGitRev="ea31cb15f0c2ab9f5f5b18e82311eb522989d747"
+REVISION="1"
+SOURCE_URI="https://github.com/google/googletest/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="41a782bacfed269409e533599f2272910e6c7ccd0c3d03f9694c693916070180"
+SOURCE_FILENAME="googletest-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="googletest-$srcGitRev"
+PATCHES="gtest-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	gtest$secondaryArchSuffix = $portVersion
+	lib:libgmock$secondaryArchSuffix
+	lib:libgmock_main$secondaryArchSuffix
+	lib:libgtest$secondaryArchSuffix
+	lib:libgtest_main$secondaryArchSuffix
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	gtest${secondaryArchSuffix}_devel = $portVersion
+	devel:libgmock$secondaryArchSuffix
+	devel:libgmock_main$secondaryArchSuffix
+	devel:libgtest$secondaryArchSuffix
+	devel:libgtest_main$secondaryArchSuffix
+	"
+REQUIRES_devel="
+	gtest$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python2
+	"
+
+defineDebugInfoPackage gtest$secondaryArchSuffix \
+	$libDir/libgmock.so \
+	$libDir/libgmock_main.so \
+	$libDir/libgtest.so \
+	$libDir/libgtest_main.so
+
+BUILD()
+{
+	mkdir -p build && cd build
+
+	cmake .. $cmakeDirArgs \
+		-DCMAKE_BUILD_TYPE='Release' \
+		-DBUILD_SHARED_LIBS:BOOL=ON
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	prepareInstalledDevelLibs libgmock libgmock_main libgtest libgtest_main
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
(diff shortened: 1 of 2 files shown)


Begasus:
Need to test for x86_64

dev-libs/gtest/patches/gtest-1.8.0.patchset
```diff
@@ -0,0 +1,82 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -4,10 +4,22 @@ project( googletest-distribution )
+ 
+ enable_testing()
+ 
++include(CMakeDependentOption)
++if (CMAKE_VERSION VERSION_LESS 2.8.5)
++  set(CMAKE_INSTALL_BINDIR "bin" CACHE STRING "User executables (bin)")
++  set(CMAKE_INSTALL_LIBDIR "lib${LIB_SUFFIX}" CACHE STRING "Object code libraries (lib)")
```
fbrosson: This might work, but you are hardcoding, and this is not recommended.

BTW, I think @mmlr will confirm that `$relativeLibDir` should not be hardcoded in a patch but, instead, passed from the recipe.

Begasus: Using -DCMAKE_INSTALL_LIBDIR=$libDir" would be better to use in the recipe then?

fbrosson: Yep, I think so. BTW, `dev-qt/prison/` does so:
```
		-DCMAKE_INSTALL_PREFIX:PATH="$prefix" \
		-DCMAKE_INSTALL_LIBDIR:PATH="$libDir" \
		-DCMAKE_INSTALL_INCLUDEDIR:PATH="$includeDir"
```

Begasus: Yeah, they probably work because they are mentioned in the cmake file(s), without the GnuInstallDirs it won't work, need to investigate ... :)

Begasus: One question (well maybe) :) should I keep the patch and use -DCMAKE_INSTALL_* for the others in the recipe?

fbrosson: IDK. I'd have to look closer...

alaviss: @Begasus 
Don't forget about [`$cmakeDirArgs`][0]

[0]: https://github.com/haikuports/haikuports/issues/2074

Begasus:
Bumped to git revision (with latest changes), seems to work ok but still getting a problem with pkg-config afterwords, ps not related to the build ... 
https://github.com/google/googletest/issues/1438

dev-libs/gtest/gtest-1.9.0~git.recipe
```diff
@@ -0,0 +1,90 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python3
```
fbrosson: Python3 has support for secondary architectures, so I guess it would be safer to use it here.
```diff
-	cmd:python3
+	cmd:python3$secondaryArchSuffix
```

Begasus: @ your service :)

Begasus: Need to revert this, looks like python3_x86 doesn't provide the cmd:python3_x86 :)

fbrosson: Oops, my mistake! I'm so sorry!

Well, if, on x86_gcc2, haikuporter activates python3 instead of python3_x86, then it might be worth trying to "convince" it to activate python3_x86.
Maybe adding `devel:libpython3.6m$secondaryArchSuffix` to `BUILD_REQUIRES` will do the trick. Good luck :)

pulkomandy: Beware, python3_x86 seems to have problems with library paths and may crash.

Begasus: Should we use python2 here instead then? I don't see any errors on x86_gcc2 (x86) nor x8_64

dev-libs/gtest/patches/gtest-1.8.0.patchset
Begasus: @alaviss is this needed? we are way past version 2.8.5 already?
Taking a quick look I can't find any recipe forcing a version for cmake (could be that I missed) :)

dev-libs/gtest/gtest-1.9.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+* Test discovery
+* A rich set of assertions
+* User-defined assertions
+* Death tests
+* Fatal and non-fatal failures
+* Value-parameterized tests
+* Type-parameterized tests
+* Various options for running the tests
+* XML test report generation"
+HOMEPAGE="https://github.com/google/googletest/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/google/googletest/"
+HOMEPAGE="https://github.com/google/googletest"
```

dev-libs/gtest/gtest-1.9.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python2
```
fbrosson: @Begasus, in case you would like to revert to python3:
```diff
-	cmd:python2
+	cmd:python3
```
I think this would work, thanks to commit 298e4d49689975287246a3bf99592a169f8f29d8 by @pulkomandy.

Begasus: Ok, I reverted it since there wasn't any response on whether to use 2 or 3 ...

dev-libs/gtest/gtest-1.9.0~git.recipe
```diff
@@ -0,0 +1,97 @@
...
+* A rich set of assertions
+* User-defined assertions
+* Death tests
+* Fatal and non-fatal failures
+* Value-parameterized tests
+* Type-parameterized tests
+* Various options for running the tests
+* XML test report generation"
+HOMEPAGE="https://github.com/google/googletest"
+COPYRIGHT="2008-2016 Google Test developers"
```
fbrosson: s/201**6**/201**8**/

fbrosson:
FWIW, I have built **gtest** `1.9.0~git` on all possible architectures.
I have also built protobuf_x86 3.5.1.1 on x86_gcc2 using gtest_x86 `1.8.0` vs `1.9.0~git` and the resulting packages are identical.

(Of course, protobuf's static libs have different checksums at every build, so we can ignore this in the comparison. The metadata is also slightly diferent, but that's because the `.PackageInfo` file keeps track of which version of `lib:libgmock` is required by protobuf's base package.)

In other words, I think this PR is ready for a merge.
(Is everyone OK with that?)

PR #2112, opened 2018-01-13
softhsm: new recipe

sfanxiang:
(Done as part of GCI 2017.)

### DIFF
--- a/dev-libs/softhsm/softhsm-2.3.0.recipe
+++ b/dev-libs/softhsm/softhsm-2.3.0.recipe
@@ -1,0 +1,85 @@
+SUMMARY="An implementation of a cryptographic store"
+DESCRIPTION="SoftHSM is an implementation of a cryptographic store \
+accessible through a PKCS #11 interface. You can use it to explore PKCS #11 \
+without having a Hardware Security Module. It is being developed as a part \
+of the OpenDNSSEC project. SoftHSM uses Botan for its cryptographic \
+operations."
+HOMEPAGE="https://www.opendnssec.org/softhsm/"
+COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation and SURFnet bv"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://dist.opendnssec.org/source/softhsm-$portVersion.tar.gz"
+CHECKSUM_SHA256="5ed604c89a3a6ef9d7d1ee92c28a2c4b3cd1f86f302c808e2d12c8f39aa2c127"
+SOURCE_DIR="softhsm-$portVersion"
+PATCHES="softhsm-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandBinDir=$binDir
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]
+then
+	commandBinDir=$prefix/bin
+	commandSuffix=
+fi
+
+PROVIDES="
+	softhsm_2$secondaryArchSuffix = $portVersion
+	cmd:softhsm2_dump_file$commandSuffix = $portVersion
+	cmd:softhsm2_keyconv$commandSuffix = $portVersion
+	cmd:softhsm2_util$commandSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	botan$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	botan${secondaryArchSuffix}_devel
+	cmd:awk
+	cmd:diff
+	cmd:find
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:makeinfo
+	cmd:pkg_config$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	"
+
+USER_SETTINGS_FILES="
+	settings/softhsm2.conf template data/softhsm/softhsm2.conf.sample
+	"
+
+GLOBAL_WRITABLE_FILES="
+	settings/softhsm2.conf keep-old
+	var/lib/softhsm directory keep-old
+	"
+
+BUILD()
+{
+	runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/softhsm/*.a $libDir/softhsm/*.la
+
+	mkdir -p $dataDir/softhsm
+	mv $settingsDir/softhsm2.conf.sample $dataDir/softhsm
+}
+
+TEST()
+{
+	make check
+}
--- a/dev-libs/softhsm/softhsm-2.3.0.recipe
+++ b/dev-libs/softhsm/softhsm-2.3.0.recipe
@@ -1,16 +1,15 @@
 SUMMARY="An implementation of a cryptographic store"
-DESCRIPTION="SoftHSM is an implementation of a cryptographic store \
-accessible through a PKCS #11 interface. You can use it to explore PKCS #11 \
-without having a Hardware Security Module. It is being developed as a part \
-of the OpenDNSSEC project. SoftHSM uses Botan for its cryptographic \
-operations."
+DESCRIPTION="SoftHSM is an implementation of a cryptographic store accessible \
+through a PKCS #11 interface. You can use it to explore PKCS #11 without \
+having a Hardware Security Module. It is being developed as a part of the \
+OpenDNSSEC project. SoftHSM uses Botan for its cryptographic operations."
 HOMEPAGE="https://www.opendnssec.org/softhsm/"
-COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation and SURFnet bv"
+COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation
+	2010 SURFnet bv"
 LICENSE="BSD (2-clause)"
 REVISION="1"
 SOURCE_URI="https://dist.opendnssec.org/source/softhsm-$portVersion.tar.gz"
 CHECKSUM_SHA256="5ed604c89a3a6ef9d7d1ee92c28a2c4b3cd1f86f302c808e2d12c8f39aa2c127"
-SOURCE_DIR="softhsm-$portVersion"
 PATCHES="softhsm-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
@@ -24,23 +23,28 @@ then
 	commandSuffix=
 fi
 
+majorVersion="${portVersion%%.*}"
+
 PROVIDES="
-	softhsm_2$secondaryArchSuffix = $portVersion
-	cmd:softhsm2_dump_file$commandSuffix = $portVersion
-	cmd:softhsm2_keyconv$commandSuffix = $portVersion
-	cmd:softhsm2_util$commandSuffix = $portVersion
+	softhsm$majorVersion$secondaryArchSuffix = $portVersion
+	cmd:softhsm${majorVersion}_dump_file$commandSuffix = $portVersion
+	cmd:softhsm${majorVersion}_keyconv$commandSuffix = $portVersion
+	cmd:softhsm${majorVersion}_util$commandSuffix = $portVersion
+	lib:libsofthsm$majorVersion$secondaryArchSuffix
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	botan$secondaryArchSuffix
+	cmd:botan$commandSuffix
+#	lib:libbotan_2$secondaryArchSuffix
 	lib:libssl$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	devel:libbotan_2$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
-	botan${secondaryArchSuffix}_devel
 	cmd:awk
 	cmd:diff
 	cmd:find
@@ -50,17 +54,22 @@ BUILD_PREREQUIRES="
 	cmd:make
 	cmd:makeinfo
 	cmd:pkg_config$secondaryArchSuffix
-	devel:libssl$secondaryArchSuffix
 	"
 
 USER_SETTINGS_FILES="
-	settings/softhsm2.conf template data/softhsm/softhsm2.conf.sample
+	settings/softhsm${majorVersion}.conf template data/softhsm/softhsm${majorVersion}.conf.sample
 	"
 
 GLOBAL_WRITABLE_FILES="
-	settings/softhsm2.conf keep-old
+	settings/softhsm${majorVersion}.conf keep-old
 	var/lib/softhsm directory keep-old
 	"
+
+defineDebugInfoPackage softhsm${majorVersion}${secondaryArchSuffix} \
+	$commandBinDir/softhsm${majorVersion}-dump-file \
+	$commandBinDir/softhsm${majorVersion}-keyconv \
+	$commandBinDir/softhsm${majorVersion}-util \
+	$libDir/libsofthsm${majorVersion}.so
 
 BUILD()
 {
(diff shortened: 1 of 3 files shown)


dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+	cmd:awk
+	cmd:diff
+	cmd:find
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:makeinfo
+	cmd:pkg_config$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
```
fbrosson:
```
	devel:libssl$secondaryArchSuffix
```
should go to `BUILD_REQUIRES`.

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+	runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/softhsm/*.a $libDir/softhsm/*.la
```
fbrosson:
```diff 
-	rm -f $libDir/softhsm/*.a $libDir/softhsm/*.la
+	rm -f $libDir/softhsm/*.la
```
I also think it would be a good idea to drop the static libs, but the current concensus is to keep them.

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
+SUMMARY="An implementation of a cryptographic store"
+DESCRIPTION="SoftHSM is an implementation of a cryptographic store \
+accessible through a PKCS #11 interface. You can use it to explore PKCS #11 \
+without having a Hardware Security Module. It is being developed as a part \
+of the OpenDNSSEC project. SoftHSM uses Botan for its cryptographic \
+operations."
```
fbrosson:
```diff
-DESCRIPTION="SoftHSM is an implementation of a cryptographic store \
-accessible through a PKCS #11 interface. You can use it to explore PKCS #11 \
-without having a Hardware Security Module. It is being developed as a part \
-of the OpenDNSSEC project. SoftHSM uses Botan for its cryptographic \
-operations."
+DESCRIPTION="SoftHSM is an implementation of a cryptographic store accessible \
+through a PKCS #11 interface. You can use it to explore PKCS #11 without \
+having a Hardware Security Module. It is being developed as a part of the \
+OpenDNSSEC project. SoftHSM uses Botan for its cryptographic operations."
```
(rewrapping only)

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+
+PROVIDES="
+	softhsm_2$secondaryArchSuffix = $portVersion
+	cmd:softhsm2_dump_file$commandSuffix = $portVersion
+	cmd:softhsm2_keyconv$commandSuffix = $portVersion
+	cmd:softhsm2_util$commandSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	botan$secondaryArchSuffix
```
fbrosson:
```diff
-	botan$secondaryArchSuffix
+	lib:libbotan_2$secondaryArchSuffix
```

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+	lib:libssl$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	botan${secondaryArchSuffix}_devel
```
fbrosson: We prefer to use the `devel:libbotan_2` syntax rather than `devel:libbotan`. BTW, it needs to go to `BUILD_REQUIRES`.
```diff
-	botan${secondaryArchSuffix}_devel
+	devel:libbotan_2$secondaryArchSuffix
```

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+	devel:libssl$secondaryArchSuffix
+	"
+
+USER_SETTINGS_FILES="
+	settings/softhsm2.conf template data/softhsm/softhsm2.conf.sample
+	"
+
+GLOBAL_WRITABLE_FILES="
+	settings/softhsm2.conf keep-old
+	var/lib/softhsm directory keep-old
```
fbrosson:
```diff
 GLOBAL_WRITABLE_FILES="
 	settings/softhsm2.conf keep-old
	var/lib/softhsm directory keep-old
 	"
+
+defineDebugInfoPackage softhsm2$secondaryArchSuffix \
+	$commandBinDir/softhsm2-dump-file \
+	$commandBinDir/softhsm2-keyconv \
+	$commandBinDir/softhsm2-util \
+	$libDir/libsofthsm2.so
+
 BUILD()
 {
```
I don't know if it's OK to have a `var/lib/softhsm` directory. We might need to change this in some way to make it adhere to the Haiku/HaikuPorts rules.

dev-libs/softhsm/softhsm-2.3.0.recipe
alaviss: Same as above

dev-libs/softhsm/softhsm-2.3.0.recipe
alaviss: May I ask why you removed the static library archives (`*.a`)?

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/softhsm/*.a $libDir/softhsm/*.la
+
+	mkdir -p $dataDir/softhsm
+	mv $settingsDir/softhsm2.conf.sample $dataDir/softhsm
```
alaviss: Shouldn't `$docDir` be more suitable for this?

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
+SUMMARY="An implementation of a cryptographic store"
+DESCRIPTION="SoftHSM is an implementation of a cryptographic store \
+accessible through a PKCS #11 interface. You can use it to explore PKCS #11 \
+without having a Hardware Security Module. It is being developed as a part \
+of the OpenDNSSEC project. SoftHSM uses Botan for its cryptographic \
+operations."
+HOMEPAGE="https://www.opendnssec.org/softhsm/"
+COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation and SURFnet bv"
```
alaviss: They're two [separated][0] holders

```diff
-COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation and SURFnet bv"
+COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation
+	2010 SURFnet bv"
```

[0]: https://github.com/opendnssec/SoftHSMv2/blob/2.3.0/LICENSE#L1

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,85 @@
...
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://dist.opendnssec.org/source/softhsm-$portVersion.tar.gz"
+CHECKSUM_SHA256="5ed604c89a3a6ef9d7d1ee92c28a2c4b3cd1f86f302c808e2d12c8f39aa2c127"
+SOURCE_DIR="softhsm-$portVersion"
```
alaviss: No need as `SOURCE_DIR` defaults to `softhsm-$portVersion` anyway

```diff
-SOURCE_DIR="softhsm-$portVersion"

File: dev-libs/softhsm/softhsm-2.3.0.recipe (hunk 1 above)
Comment (alaviss) [inline dev-libs/softhsm/softhsm-2.3.0.recipe]: Shouldn't this be in `BUILD_REQUIRES`?

File: dev-libs/softhsm/softhsm-2.3.0.recipe (hunk 6 above)
Comment (alaviss) [reply dev-libs/softhsm/softhsm-2.3.0.recipe]: Is the said directory included in the package? If yes then it's necessary that we include it, otherwise it have to be removed

From [BuildingPackages](https://dev.haiku-os.org/wiki/Obsolete/MovedToTree/PackageManagement/BuildingPackages)

> `global-writable-files`: A list of global writable file infos. The list elements must have the following format:
>
>      global_writable_file_info	::= path [ "directory" ] [ "keep-old" | "manual" | "auto-merge" ]
>
> `path` is the relative path of the writable file or directory, starting with "settings/" or any other writable directory. If the "directory" keyword is given, the path refers to a directory. If no other keyword is given after the path respectively after the "directory" keyword, the file or directory is not included in the package. It will be created by the software or by the user.

To be honest the definition of this variable implicts that we need one for every `USER_SETTINGS_FILES`. I'm a bit confused with this myself, probably needs some knowledgeable ppl to explain.

File: dev-libs/softhsm/softhsm-2.3.0.recipe (hunk 9 above)
Comment (fbrosson) [reply dev-libs/softhsm/softhsm-2.3.0.recipe]: Good catch! Thanks @alaviss!

File: dev-libs/softhsm/softhsm-2.3.0.recipe (hunk 4 above)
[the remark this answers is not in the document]
Comment (fbrosson) [reply dev-libs/softhsm/softhsm-2.3.0.recipe]: It only needs the `lib:` and `devel:`. I made the change on my side and the recipe builds fine.

(I already worte a tiny review with these changes: https://github.com/haikuports/haikuports/pull/2112#pullrequestreview-88660189 :)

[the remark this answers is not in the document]
Comment (alaviss) [reply dev-libs/softhsm/softhsm-2.3.0.recipe]: You beat me to it lol. Anyway, there's no `lib:libbotan_2`

https://github.com/haikuports/haikuports/blob/fc0936b04009e713f9bf551851781c83f9f23871/dev-libs/botan/botan-2.0.1.recipe#L23

File: dev-libs/softhsm/softhsm-2.3.0.recipe (hunk 2 above)
Comment (fbrosson) [reply dev-libs/softhsm/softhsm-2.3.0.recipe]: I already left a comment about this: https://github.com/haikuports/haikuports/pull/2112#discussion_r161374439

File: dev-libs/softhsm/softhsm-2.3.0.recipe [hunk 10]
```diff
@@ -0,0 +1,85 @@
...
+then
+	commandBinDir=$prefix/bin
+	commandSuffix=
+fi
+
+PROVIDES="
+	softhsm_2$secondaryArchSuffix = $portVersion
+	cmd:softhsm2_dump_file$commandSuffix = $portVersion
+	cmd:softhsm2_keyconv$commandSuffix = $portVersion
+	cmd:softhsm2_util$commandSuffix = $portVersion
```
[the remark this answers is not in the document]
fbrosson: Yes, that's a good idea.
```diff
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
 commandBinDir=$binDir
 commandSuffix=$secondaryArchSuffix
 if [ "$targetArchitecture" = x86_gcc2 ]
 then
 	commandBinDir=$prefix/bin
 	commandSuffix=
 fi
+
+majorVersion="${portVersion%%.*}"
+
 PROVIDES="
-	softhsm2$secondaryArchSuffix = $portVersion
+	softhsm$majorVersion$secondaryArchSuffix = $portVersion
-	cmd:softhsm2_dump_file$commandSuffix = $portVersion
+	cmd:softhsm${majorVersion}_dump_file$commandSuffix = $portVersion
-	cmd:softhsm2_keyconv$commandSuffix = $portVersion
+	cmd:softhsm${majorVersion}_keyconv$commandSuffix = $portVersion
-	cmd:softhsm2_util$commandSuffix = $portVersion
+	cmd:softhsm${majorVersion}_util$commandSuffix = $portVersion
-	lib:libsofthsm2$secondaryArchSuffix
+	lib:libsofthsm$majorVersion$secondaryArchSuffix
	"
REQUIRES="
	haiku$secondaryArchSuffix
-	lib:libbotan_2$secondaryArchSuffix
+	lib:libbotan_$majorVersion$secondaryArchSuffix
	lib:libssl$secondaryArchSuffix
	"
```

dev-libs/softhsm/softhsm-2.3.0.recipe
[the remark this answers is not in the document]
fbrosson: You are perfectly right, @alaviss. Our current recipe for botan only produces the static lib.

BTW, I merged #1094 one year ago, but we did not manage to create the shared lib, so we added a commented-out `lib:libbotan_2`.

Hmm, I just realized that haikuporter does not even try to see if the dependencies we put in `REQUIRES` actually exist. (And this is perfectly normal.)
It's only at install time that we are blocked: pkgman tells me that nothing provides lib:libbotan_2, which is "needed by softhsm2-2.3.0-1".

dev-libs/softhsm/softhsm-2.3.0.recipe
fbrosson: (Sorry for this late reply.)
Well, I'd say that if that file was not supposed to be a template config, then it would be OK to put it in `$docDir`. But since it is a template, it should be, imo, in a subdirectory under `$dataDir`.

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,99 @@
...
+having a Hardware Security Module. It is being developed as a part of the \
+OpenDNSSEC project. SoftHSM uses Botan for its cryptographic operations."
+HOMEPAGE="https://www.opendnssec.org/softhsm/"
+COPYRIGHT="2010 .SE, The Internet Infrastructure Foundation
+	2010 SURFnet bv"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://dist.opendnssec.org/source/softhsm-$portVersion.tar.gz"
+CHECKSUM_SHA256="5ed604c89a3a6ef9d7d1ee92c28a2c4b3cd1f86f302c808e2d12c8f39aa2c127"
+PATCHES="softhsm-$portVersion.patchset"
```
fbrosson: With**out** changing the name of the directory, could you please rename the recipe to `softhsm2-2.3.0.recipe` ?
(This means we need to add the `SOURCE_DIR` back.)
```diff
 CHECKSUM_SHA256="5ed604c89a3a6ef9d7d1ee92c28a2c4b3cd1f86f302c808e2d12c8f39aa2c127"
+SOURCE_DIR="softhsm-$portVersion"
 PATCHES="softhsm-$portVersion.patchset"
```

dev-libs/softhsm/softhsm-2.3.0.recipe
```diff
@@ -0,0 +1,99 @@
...
+defineDebugInfoPackage softhsm${majorVersion}${secondaryArchSuffix} \
+	$commandBinDir/softhsm${majorVersion}-dump-file \
+	$commandBinDir/softhsm${majorVersion}-keyconv \
+	$commandBinDir/softhsm${majorVersion}-util \
+	$libDir/libsofthsm${majorVersion}.so
+
+BUILD()
+{
+	runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir
```
fbrosson:
```diff
-	runConfigure --omit-dirs binDir ./configure \
-		--bindir=$commandBinDir
+	runConfigure --omit-dirs binDir ./configure --bindir=$commandBinDir
```
(We're still under the 80 chars recommended limit.)

dev-libs/softhsm/softhsm-2.3.0.recipe
fbrosson: The messages above are outdated.
(We do need the `SOURCE_DIR` because we have chosen to rename the recipe.)

dev-libs/softhsm/softhsm2-2.3.0.recipe
```diff
@@ -0,0 +1,99 @@
...
+GLOBAL_WRITABLE_FILES="
+	settings/softhsm${majorVersion}.conf keep-old
+	var/lib/softhsm directory keep-old
+	"
+
+defineDebugInfoPackage softhsm${majorVersion}${secondaryArchSuffix} \
+	$commandBinDir/softhsm${majorVersion}-dump-file \
+	$commandBinDir/softhsm${majorVersion}-keyconv \
+	$commandBinDir/softhsm${majorVersion}-util \
+	$libDir/libsofthsm${majorVersion}.so
```
fbrosson:
```diff
-defineDebugInfoPackage softhsm${majorVersion}${secondaryArchSuffix} \
+defineDebugInfoPackage softhsm${majorVersion}$secondaryArchSuffix \
 	$commandBinDir/softhsm${majorVersion}-dump-file \
 	$commandBinDir/softhsm${majorVersion}-keyconv \
 	$commandBinDir/softhsm${majorVersion}-util \
-	$libDir/libsofthsm${majorVersion}.so
+	$libDir/softhsm/libsofthsm${majorVersion}.so
```
* The first change is neutral, but since we prefer to drop the curly braces when they are not necessary, it would be nice to drop them here, for `secondaryArchSuffix`. (Please do not drop them for `majorVersion`, unless you really wish to, but not before you have read the following paragraph.)
* The second change is required, because the lib is in a subdirectory.

BTW, regarding the removal of unneeded curly braces, I had suggested to remove them for `majorVersion` too, but you chose to keep them at some strategic places, and I did not comment about that because I found that, after all, they are more eye-friendly the way you wrote them, I mean, with the curly braces.

Once you have addressed the 2 changes in the above diff the recipe will be OK for me :) Thanks!

PR #2113, opened 2018-01-13
util-linux: bump to 2.31.1, fix policy error in 2nd arch.

fbrosson:
* Use the commandBinDir "trick" to fix the policy error.
* Mark x86_gcc2 as broken because a gcc2 patch would be required.
* ~Drop the call to `autogen.sh`.~ (Package contents are identical.) _[Keeping the call to `autogen.sh` because it might be needed if we ever choose to enable some programs]_
* ~x86 secondary arch builds are hanging on two configure tests which crash. Dropping a paragraph in Makefile.am eliminates these crashes and has no impact on the resulting runtimes.~ _[Thanks @alaviss for the hint on passing "have_scanf_alloc_modifier=no" to configure.]_
* On x86, pass "have_scanf_alloc_modifier=no" to configure to skip a conftest which crashes and makes configure hang.
* Switch to tar.xz.

### DIFF
--- a/sys-apps/util-linux/patches/util_linux-2.31.1.patchset
+++ b/sys-apps/util-linux/patches/util_linux-2.31.1.patchset
@@ -10,7 +10,7 @@ Subject: [PATCH] Haiku patches
  3 files changed, 42 insertions(+)
 
 diff --git a/include/c.h b/include/c.h
-index a5162b9..d9ffef7 100644
+index b0670dc..8808c32 100644
 --- a/include/c.h
 +++ b/include/c.h
 @@ -29,6 +29,10 @@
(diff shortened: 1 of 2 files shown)


sys-apps/util-linux/patches/util_linux-2.31-x86.patchset
```diff
@@ -0,0 +1,54 @@
...
+
+diff --git a/configure.ac b/configure.ac
+index 1899ec3..d03f518 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -601,39 +601,6 @@ int main()
+ 	return rc;
+ }])
+ 
+-AC_CACHE_VAL([scanf_cv_alloc_modifier],
```
alaviss: Since it's cached, maybe all you need is to add `scanf_cv_alloc_modifier=no` to `./configure` and it will happily skip the test.

Issue #1114, opened 2017-01-27
Write a recipe for Sigil


pulkomandy:
Sigil (https://sigil-ebook.com) is a WYSIWYG editor for epub e-books. It is not too different from a word processor. It is Qt based.

augiedoggie:
I think this can be closed.  Looks like @threedeyes added a recipe in 5e6e5f24b730f519a37822a7d540efaf5e792e45

PR #2117, opened 2018-01-14
Sigil: new recipe

extrowerk:


app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+	2012  Dave Heiland
+	2012  John Schember
+	2012 Grant Drake
+	2016 Kevin B. Hendricks"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/Sigil-Ebook/Sigil/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="12c1ace50598aef0a5e1397da3fdcd9c5ed9eba753c476d0ee5ab00a5d2d3764"
+SOURCE_DIR="Sigil-$portVersion"
+ADDITIONAL_FILES="sigil.rdef"
```
fbrosson:
```diff
-ADDITIONAL_FILES="sigil.rdef"
+ADDITIONAL_FILES="sigil.rdef.in"
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+	local APP_SIGNATURE="application/x-vnd.qt5-sigil"
+	local MAJOR="`echo "$portVersion" | cut -d. -f1`"
+	local MIDDLE="`echo "$portVersion" | cut -d. -f2`"
+	local MINOR="`echo "$portVersion" | cut -d. -f3`"
+	sed \
+		-e "s|@APP_SIGNATURE@|$APP_SIGNATURE|" \
+		-e "s|@MAJOR@|$MAJOR|" \
+		-e "s|@MIDDLE@|$MIDDLE|" \
+		-e "s|@MINOR@|$MINOR|" \
+		$portDir/additional-files/sigil.rdef > sigil.rdef
```
fbrosson:
```diff
-		$portDir/additional-files/sigil.rdef > sigil.rdef
+		$portDir/additional-files/sigil.rdef.in > sigil.rdef
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
+SUMMARY="A free, open source, multi-platform ebook editor"
+DESCRIPTION="Sigil is designed to edit books in ePub format (both ePub 2 and \
+ePub 3)."
+HOMEPAGE="http://sigil-ebook.com"
```
fbrosson:
```diff
-HOMEPAGE="http://sigil-ebook.com"
+HOMEPAGE="https://sigil-ebook.com/
```
(Please don't forget the trailing slash. Thanks :)

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
+SUMMARY="A free, open source, multi-platform ebook editor"
+DESCRIPTION="Sigil is designed to edit books in ePub format (both ePub 2 and \
+ePub 3)."
+HOMEPAGE="http://sigil-ebook.com/"
```
fbrosson: Oops, you forgot the switch from http to https :)

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5WebKit$secondaryArchSuffix
+	devel:libQt5WebKitWidgets$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:cmake
+	cmd:make
+	cmd:pkg_config
```
fbrosson:
```diff
-	cmd:pkg_config
+	cmd:pkg_config$secondaryArchSuffix
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python3
+	"
+
+BUILD()
+{
+	mkdir -p haiku_build && cd haiku_build
+
+	cmake .. \
+		-DCMAKE_INSTALL_PREFIX=$appsDir/Sigil \
```
fbrosson: I think you can put Sigil directly in `$appsDir` and leave it's libs in `$libDir/Sigil/` with:
```diff
 	cmake .. \
-		-DCMAKE_INSTALL_PREFIX=$appsDir/Sigil \
+		-DCMAKE_INSTALL_PREFIX=$prefix \
+		-DCMAKE_INSTALL_BINDIR=$appsDir \
+		-DCMAKE_INSTALL_LIBDIR=$libDir \
 		-DUSE_SYSTEM_LIBS=ON
```

Obviously, this would require some changes in `INSTALL()`, but you would have fewer commands than you have now.

extrowerk: I'd like to have everything in one folder.

fbrosson: I don't know if Sigil, when installed directly in `$appsDir/`, would be able to find its libs in `$libDir/sigil/`, but **_if_** that's the case **_then_** it would be better, imo, to avoid having a subdirectory in `$appsDir` with only the app and another subdirectory for the libs.

If I understand correctly, with your current recipe you get these files:
```
/system/apps/Sigil/Sigil
/system/apps/Sigil/lib/libsigilgumbo.so
```
With what I'm suggesting you would have these:
```
/system/apps/Sigil
/system/lib/sigil/libsigilgumbo.so
```
There are 3 advantages with this layout:
1) It creates only 1 subdirectory, instead of 2.
2) Sigil can be launched from a terminal with just "Sigil", since `/boot/system/apps/` is in `$PATH`.
3) In case Sigil builds on both x86_gcc2 primary arch and x86 secondary arch, you could, with a few small changes in the recipe, have this if you install both sigil and sigil_x86:
```
/system/apps/Sigil
/system/apps/Sigil-x86
/system/lib/sigil/libsigilgumbo.so
/system/lib/x86/sigil/libsigilgumbo.so
```
whereas with your current recipe you would have something like:
```
/system/apps/Sigil/Sigil
/system/apps/Sigil/Sigil-x86
/system/apps/Sigil/lib/libsigilgumbo.so
/system/apps/Sigil/lib/x86/libsigilgumbo.so
```
or
```
/system/apps/Sigil/Sigil
/system/apps/Sigil/lib/libsigilgumbo.so
/system/apps/Sigil-x86/Sigil
/system/apps/Sigil-x86/lib/libsigilgumbo.so
```
or
```
/system/apps/Sigil/Sigil
/system/apps/Sigil/lib/libsigilgumbo.so
/system/apps/Sigil/x86/Sigil
/system/apps/Sigil/x86/lib/libsigilgumbo.so
```
But all these layouts add unneeded directories instead of using those already available with the usual layout.

BTW, if I understand correctly, the directory with the shared library will only have one file, the lib itself. So once you know it is there, you don't need to care about it's location.

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+	2012 Grant Drake
+	2016 Kevin B. Hendricks"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/Sigil-Ebook/Sigil/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="12c1ace50598aef0a5e1397da3fdcd9c5ed9eba753c476d0ee5ab00a5d2d3764"
+SOURCE_DIR="Sigil-$portVersion"
+ADDITIONAL_FILES="sigil.rdef.in"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+	sigil = $portVersion
+	app:Sigil$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	cmd:python3
+	lib:libhunspell_1.6$secondaryArchSuffix
+	lib:libminizip$secondaryArchSuffix
+#	lib:libpcre$secondaryArchSuffix
+	lib:libpython${PYTHON_VERSION}m
```
fbrosson:
```diff
-	lib:libpython${PYTHON_VERSION}m
+	lib:libpython${PYTHON_VERSION}m$secondaryArchSuffix
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
...
+ADDITIONAL_FILES="sigil.rdef.in"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PYTHON_VERSION="3.6"
+
+PROVIDES="
+	sigil = $portVersion
+	app:Sigil$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES="
-	sigil = $portVersion
+	sigil$secondaryArchSuffix = $portVersion
-	app:Sigil$secondaryArchSuffix = $portVersion
+	app:Sigil = $portVersion
 	"
```

app-text/sigil/sigil-0.9.9.recipe
fbrosson: FWIW, I have built sigil, and I am able to run the app. But it displays some errors when try to do anything. Those errors mention python scripts. I haven't looked close enough.

@extrowerk, do you also get these errors?

In case a python expert would like to try this recipe, here are some things to know:

* If we only set the prefix and don't override any paths, the layout of a default installation is:

File | remark
---- | ----
/boot/system/bin/sigil | a shell script
/boot/system/lib/sigil/sigil | the runtime (around 10 MB)
/boot/system/lib/sigil/libsigilgumbo.so | the private lib (500 KB)

* The shell script is supposed to be the entry point. It sets some environment variables and launches `/boot/system/lib/sigil/sigil`which then loads `/boot/system/lib/sigil/libsigilgumbo.so`.

* If we want to create a symlink in the applications folder then we should do so with the wrapper script, not with the big runtime. So it might be worth trying to set some default values for the environment variables which this script sets in order to call the main runtime directly, instead of the kicking the wrapper script.

extrowerk: I also have Seen some errors. But then magic happend, and it started. Can you Post your errors, please?

fbrosson: The errors I get are all shown in a dialog box having "Sigil" as window name.
The first error I get has this message: `Embedded Python Error`, as well as 2 buttons: "Show Details..." and a "Close". When I click on "Show Details.." I can see this:
```

  File "/packages/sigil-0.9.9-1/.self/share/sigil/python3lib/xmlprocessor.py", line 6, in <module>
    from sigil_bs4.builder._lxml import LXMLTreeBuilderForXML

  File "/packages/sigil-0.9.9-1/.self/share/sigil/plugin_launchers/python/sigil_bs4/builder/_lxml.py", line 26, in <module>
    from lxml import etree

ModuleNotFoundError: No module named 'lxml'
```
When I click "Close" I get another message: `error in xmlprocessor repairXML:-2`. But the details of this one are identical to those of the previous error.

I did not investigate at all about this "LXML" module.

The next errors seem to be the same. After I have close a few of them, I don't get anymore errors. But soon after, the app crashes.
Well, if I'm not mistaken, yesterday I reached a point where I could see the Sigil's main window, with a menu and all the decorations.

pulkomandy: It looks like the python module should be packaged (if that's not the case already) and the sigil package should depend on it.

alaviss: @fbrosson Since the lib is private, and it's packaged with an executable, I think this should be in `$libExecDir` (yes, I'm aware that it's the same as `$libDir` for now). Also, should we use [`$cmakeDirArgs`][0] instead?

[0]: https://github.com/haikuports/haikuports/issues/2074

extrowerk:
closes #1114

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,100 @@
+SUMMARY="A free, open source, multi-platform ebook editor"
+DESCRIPTION="Sigil is designed to edit books in ePub format (both ePub 2 and \
+ePub 3)."
+HOMEPAGE="https://sigil-ebook.com/"
+COPYRIGHT="2009, 2010, 2011  Strahinja Markovic
+	2011, 2012  John Schember
+	2012  Dave Heiland
+	2012  John Schember
```
alaviss:
```diff
-COPYRIGHT="2009, 2010, 2011  Strahinja Markovic
-	2011, 2012  John Schember
-	2012  Dave Heiland
-	2012  John Schember
+COPYRIGHT="2009, 2010, 2011 Strahinja Markovic
+	2011, 2012 John Schember
+	2012 Dave Heiland
+	2012 John Schember
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
+SUMMARY="A free, open source, multi-platform ebook editor"
+DESCRIPTION="Sigil is designed to edit books in ePub format (both ePub 2 and \
+ePub 3)."
+HOMEPAGE="https://sigil-ebook.com/"
+COPYRIGHT="2009, 2010, 2011 Strahinja Markovic
+	2011, 2012 John Schember
+	2012 Dave Heiland
+	2012 John Schember
```
fbrosson:
```diff
-	2012 John Schember
+	2012 John Schember"
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,98 @@
+SUMMARY="A free, open source, multi-platform ebook editor"
+DESCRIPTION="Sigil is designed to edit books in ePub format (both ePub 2 and \
+ePub 3)."
+HOMEPAGE="https://sigil-ebook.com/"
+COPYRIGHT="2009, 2010, 2011 Strahinja Markovic
+	2011, 2012 John Schember
+	2012 Dave Heiland
+	2012 John Schember"
```
fbrosson: Oops, I just undestood what happened. The [diff](https://github.com/haikuports/haikuports/pull/2117#pullrequestreview-116764124) @alaviss wrote did not cover the full `COPYRIGHT`, and you ate the [next two lines](https://github.com/haikuports/haikuports/pull/2117/commits/dbb7404828f7a4e5bf0d7eae4d3dcf3490a66efb#diff-ed6ed231d7f69ac1e469f24e27f14a71R9) when you made the changes.
```
	2012 Grant Drake
	2016 Kevin B. Hendricks"
```
Then I noticed the missing quotes and thought you had eaten the closing quotes after `2012 John Schember`.

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,100 @@
...
+	2011, 2012 John Schember
+	2012 Dave Heiland
+	2012 John Schember
+	2012 Grant Drake
+	2016 Kevin B. Hendricks"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/Sigil-Ebook/Sigil/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="12c1ace50598aef0a5e1397da3fdcd9c5ed9eba753c476d0ee5ab00a5d2d3764"
+SOURCE_DIR="Sigil-$portVersion"
```
alaviss:
```suggestion
SOURCE_FILENAME="sigil-$portVersion.tar.gz"
SOURCE_DIR="Sigil-$portVersion"
```

app-text/sigil/sigil-0.9.9.recipe
```diff
@@ -0,0 +1,100 @@
...
+	devel:libQt5WebKitWidgets$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:cmake
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python3
+	"
```
alaviss: Can you please sort `BUILD_PREREQUIRES`?

Begasus:
Superseded by https://github.com/haikuports/haikuports/commit/5e6e5f24b730f519a37822a7d540efaf5e792e45

PR #2119, opened 2018-01-14
net-misc/youtube-dl: version bump.

netpositive36:


fbrosson:
If everything is working fine with the new release, could you please drop the old recipe (`youtube_dl-2017.11.15.recipe`)? Thanks!

PR #2120, opened 2018-01-14
PostgreSQL: bump, adjust, etc

extrowerk:
Enabled icu and xml, xslt cause i thought it could be nice.
My patch (the latest) maybe have to be fixed before merge. Any constructive idea will be greatly appreciated.


extrowerk:
Tests won't run as it cannot find the required so files. Somebody have an LD command for it, but i can't seems to find it right now.

dev-db/postgresql/postgresql-10.1.recipe
```diff
@@ -0,0 +1,147 @@
...
+	mkdir -p $developLibDir/postgresql/pgxs
+	mv $libDir/postgresql/pgxs $developLibDir/postgresql/
+
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
```
fbrosson: I haven't checked yet, but if the shared libs are built in e.g. `dev-db/postgresql/work-9.3.6/sources/postgresql-10.1/src/lib/` then this should work:
```diff
-	make check
+	LIBRARY_PATH="$sourceDir/src/lib${LIBRARY_PATH:+:$LIBRARY_PATH}" make check
```

scottmc:
anything here worth keeping/merging with the other pending pull request?

waddlesplash:
Actually, I didn't notice the USE_DSM_MMAP; we should probably use that instead of the stubbed sys/shm. I'll take a look at that.

Otherwise that one probably supersedes this, despite the fact this uses a newer postgres.

PR #2122, opened 2018-01-14
Conway [recipe]

bach5000:
Tested on _gcc2 and _64. 
* x86_gcc2 👍 
* x86_64     👎  (I guess I'll mark this arch as broken)

Also, does anyone know how to get rid of the `modification time [x] seconds in the future` warning, without having to use that crazy looking `find` command in BUILD?

### DIFF
--- a/haiku-apps/conway/conway-0.9.recipe
+++ b/haiku-apps/conway/conway-0.9.recipe
@@ -7,9 +7,9 @@ HOMEPAGE="https://github.com/HaikuArchives/Conway"
 COPYRIGHT="1998 Alex Chamberlain"
 LICENSE="GNU GPL v2"
 REVISION="1"
-srcGitRev="7c509bca14ce4e286e8b1a2b057f5b1520ae91af"
+srcGitRev="cc589fdda35b4f5b07c1bd3a4dc797638a45b50e"
 SOURCE_URI="https://github.com/HaikuArchives/Conway/archive/$srcGitRev.tar.gz"
-CHECKSUM_SHA256="98aa350c9f0fa454c04447ffd21c116981546a929068eaaf211cc0e4c3539466"
+CHECKSUM_SHA256="" # This will be added shortly
 SOURCE_FILENAME="Conway-$portVersion-$srcGitRev.tar.gz"
 SOURCE_DIR="Conway-$srcGitRev"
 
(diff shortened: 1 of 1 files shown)


fbrosson:
One possible reason for your timestamps being in the future is an incorrect timezone setting.
Run `date` and `TZ=UTC date` on your Haiku system. The first should give you your localtime. The second is for the GMT time. If you did not set your timezone, you will get the same numbers.

To set your timezone, type `Time` (or choose "Preferences" -> "Time") and go to the "Time zone" tab.

bach5000:
Thanks, @fbrosson!
It works fine now 😃

haiku-apps/conway/conway-0.9.recipe
```diff
@@ -0,0 +1,45 @@
+SUMMARY="A simple implementation of \"Conway's Game of Life\""
```
fbrosson: I think it would look nicer with the quotes before Game:
```diff
-SUMMARY="A simple implementation of \"Conway's Game of Life\""
+SUMMARY="A simple implementation of Conway's \"Game of Life\""
```
What do you think?

bach5000:
The recipe finally builds now, without any timestamp problems 😄

haiku-apps/conway/conway-0.9.recipe
```diff
@@ -0,0 +1,45 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:gcc
+	cmd:make
+	"
+
+BUILD()
```
fbrosson:
```diff
 	cmd:make
 	"
+
+defineDebugInfoPackage conway \
+ 	$appsDir/Conway
+
BUILD()
```
(Could you please add a debuginfo package? Thanks :)

haiku-apps/conway/conway-0.9.recipe
```diff
@@ -0,0 +1,48 @@
...
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:gcc
+	cmd:make
+	"
+
+defineDebugInfoPackage conway \
+ 	$appsDir/Conway
```
fbrosson: Oops, there's a space before the tab (in line 36).

fbrosson: Do you think we should rename the recipe to `conway-0.9~git.recipe` ?
(According to our rules, it would be better, but since the app is probably not going to change, I guess we can keep the current name.)

fbrosson: Sorry, I just noticed that the space came from the diff I wrote in my comment above.
(I got rid of it, and then merged the PR, so we're done with this recipe :)

Issue #661, opened 2016-06-19
Add a recipe for MP3Gain


richienyhus:
http://mp3gain.sourceforge.net

Old BeOS port: 
[MP3GainBeOS-1.3.1_src-incl.zip](https://github.com/haikuports/haikuports/files/322286/MP3GainBeOS-1.3.1_src-incl.zip)

Interesting to see who the original porters were for this one. 😉

mmuman:
I have no idea who that can be 😄

richienyhus:
Fixed in #2123

PR #2123, opened 2018-01-15
mp3gain: Add recipe

tts2k:
~~I need suggestion for `$portVersion`. Not sure if it's ok~~

### DIFF
--- a/media-sound/mp3gain/mp3gain-1.5.2.recipe
+++ b/media-sound/mp3gain/mp3gain-1.5.2.recipe
@@ -7,7 +7,7 @@ MP3Gain can adjust the volume in a completely lossless way by modifying the \
 mp3 file directly, without decoding and re-encoding. This works with all mp3 \
 players, i.e. no support for a special tag is required."
 HOMEPAGE="http://mp3gain.sourceforge.net/"
-COPYRIGHT="1995, 1996, 1997 Michael Hipp
+COPYRIGHT="1995-1997 Michael Hipp
 	1999 Mark Taylor
 	1999 Takehiro TOMINAGA
 	1999-2000 Albert L. Faber
--- a/media-sound/mp3gain/mp3gain-1.5.2.recipe
+++ b/media-sound/mp3gain/mp3gain-1.5.2.recipe
@@ -1,15 +1,22 @@
 SUMMARY="Lossless mp3 normalizer with statistical analysis"
 DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
 volume. MP3Gain does not just do peak normalization, as many normalizers do. \
-Instead, it does some statistical analysis (using the replaygain algorithm)to \
+Instead, it does some statistical analysis (using the replaygain algorithm) to \
 determine how loud the file actually sounds to the human ear.
-MP3Gain can adjust the volume in a completely lossless way by modifying the mp3 \
-file directly, without decoding and re-encoding. This works with all mp3 \
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
+mp3 file directly, without decoding and re-encoding. This works with all mp3 \
 players, i.e. no support for a special tag is required."
 HOMEPAGE="http://mp3gain.sourceforge.net/"
-COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
-	2002 John Zitterkopf"
-LICENSE="GNU LGPL v2.1"
+COPYRIGHT="1995, 1996, 1997 Michael Hipp
+	1999 Mark Taylor
+	1999 Takehiro TOMINAGA
+	1999-2000 Albert L. Faber
+	2000 Don Melton
+	2001-2009 David Robinson and Glen Sawyer
+	2002 John Zitterkopf
+	2004-2009 David Lasker, Altos Design, Inc."
+LICENSE="GNU GPL v2
+	GNU LGPL v2.1"
 REVISION="1"
 SOURCE_URI="https://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
 CHECKSUM_SHA256="dbadc7a41a8a3b87d3a21a1989701cfe72d2090c3ead02b0b396a19a8acf6fca"
--- a/media-sound/mp3gain/mp3gain-1.5.2.recipe
+++ b/media-sound/mp3gain/mp3gain-1.5.2.recipe
@@ -41,9 +41,6 @@ BUILD_PREREQUIRES="
 	cmd:make
 	"
 
-defineDebugInfoPackage mp3gain \
-	$binDir/mp3gain
-
 BUILD()
 {
 	make OSTYPE=beos
@@ -52,5 +49,5 @@ BUILD()
 INSTALL()
 {
 	install -d $binDir
-	install -t $binDir mp3gain
+	install -t $binDir --strip mp3gain
 }
--- a/media-sound/mp3gain/mp3gain-1.5.2.recipe
+++ b/media-sound/mp3gain/mp3gain-1.5.2.recipe
@@ -1,22 +1,22 @@
 SUMMARY="Lossless mp3 normalizer with statistical analysis"
 DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
 volume. MP3Gain does not just do peak normalization, as many normalizers do. \
-Instead, it does some statistical analysis (using the replaygain algorithm) \
-to determine how loud the file actually sounds to the human ear.
-MP3Gain can adjust the volume in a completely lossless way by modifying the \
-mp3 file directly, without decoding and re-encoding. This works with all \
-mp3 players, i.e. no support for a special tag is required."
-HOMEPAGE="http://mp3gain.sourceforge.net"
+Instead, it does some statistical analysis (using the replaygain algorithm)to \
+determine how loud the file actually sounds to the human ear.
+MP3Gain can adjust the volume in a completely lossless way by modifying the mp3 \
+file directly, without decoding and re-encoding. This works with all mp3 \
+players, i.e. no support for a special tag is required."
+HOMEPAGE="http://mp3gain.sourceforge.net/"
 COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
 	2002 John Zitterkopf"
 LICENSE="GNU LGPL v2.1"
 REVISION="1"
-SOURCE_URI="http://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
+SOURCE_URI="https://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
 CHECKSUM_SHA256="dbadc7a41a8a3b87d3a21a1989701cfe72d2090c3ead02b0b396a19a8acf6fca"
 SOURCE_FILENAME="mp3gain-${portVersion//./_}-src.zip"
 SOURCE_DIR=""
 
-ARCHITECTURES="?x86_gcc2 x86_64"
+ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
 	mp3gain = $portVersion
@@ -33,6 +33,9 @@ BUILD_PREREQUIRES="
 	cmd:gcc
 	cmd:make
 	"
+
+defineDebugInfoPackage mp3gain \
+	$binDir/mp3gain
 
 BUILD()
 {
--- a/media-sound/mp3gain/mp3gain-1.5.2.recipe
+++ b/media-sound/mp3gain/mp3gain-1.5.2.recipe
@@ -12,7 +12,7 @@ COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
 LICENSE="GNU LGPL v2.1"
 REVISION="1"
 SOURCE_URI="http://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
-CHECKSUM_SHA256="3378d32c8333c14f57622802f6a92b725f36ee45a6b181657b595b1b5d64260f"
+CHECKSUM_SHA256="dbadc7a41a8a3b87d3a21a1989701cfe72d2090c3ead02b0b396a19a8acf6fca"
 SOURCE_FILENAME="mp3gain-${portVersion//./_}-src.zip"
 SOURCE_DIR=""
 
--- a/media-sound/mp3gain/mp3gain-1.5.2.r2.recipe
+++ b/media-sound/mp3gain/mp3gain-1.5.2.r2.recipe
@@ -1,0 +1,46 @@
+SUMMARY="Lossless mp3 normalizer with statistical analysis"
+DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
+volume. MP3Gain does not just do peak normalization, as many normalizers do. \
+Instead, it does some statistical analysis (using the replaygain algorithm) \
+to determine how loud the file actually sounds to the human ear.
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
+mp3 file directly, without decoding and re-encoding. This works with all \
+mp3 players, i.e. no support for a special tag is required."
+HOMEPAGE="http://mp3gain.sourceforge.net"
+COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
+	2002 John Zitterkopf"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
+CHECKSUM_SHA256="3378d32c8333c14f57622802f6a92b725f36ee45a6b181657b595b1b5d64260f"
+SOURCE_FILENAME="mp3gain-${portVersion//./_}-src.zip"
+SOURCE_DIR=""
+
+ARCHITECTURES="?x86_gcc2 x86_64"
+
+PROVIDES="
+	mp3gain = $portVersion
+	cmd:mp3gain = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	"
+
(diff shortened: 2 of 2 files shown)


korli:
Just use 1.5.2

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,46 @@
...
+COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
+	2002 John Zitterkopf"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
+CHECKSUM_SHA256="dbadc7a41a8a3b87d3a21a1989701cfe72d2090c3ead02b0b396a19a8acf6fca"
+SOURCE_FILENAME="mp3gain-${portVersion//./_}-src.zip"
+SOURCE_DIR=""
+
+ARCHITECTURES="?x86_gcc2 x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="?x86_gcc2 x86_64"
+ARCHITECTURES="x86_gcc2 x86_64"
```

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="Lossless mp3 normalizer with statistical analysis"
+DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
+volume. MP3Gain does not just do peak normalization, as many normalizers do. \
+Instead, it does some statistical analysis (using the replaygain algorithm) \
+to determine how loud the file actually sounds to the human ear.
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
+mp3 file directly, without decoding and re-encoding. This works with all \
+mp3 players, i.e. no support for a special tag is required."
+HOMEPAGE="http://mp3gain.sourceforge.net"
```
fbrosson:
```diff
-HOMEPAGE="http://mp3gain.sourceforge.net"
+HOMEPAGE="http://mp3gain.sourceforge.net/"
```

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,46 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	"
+
```
fbrosson:
```diff
 	cmd:make
 	"
+
+defineDebugInfoPackage mp3gain \
+	$binDir/mp3gain
+
 BUILD()
```
(It's nice to have the debuginfo package.)

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,46 @@
...
+to determine how loud the file actually sounds to the human ear.
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
+mp3 file directly, without decoding and re-encoding. This works with all \
+mp3 players, i.e. no support for a special tag is required."
+HOMEPAGE="http://mp3gain.sourceforge.net"
+COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
+	2002 John Zitterkopf"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="http://downloads.sourceforge.net/mp3gain/mp3gain-${portVersion//./_}-src.zip"
```
fbrosson: http -> https
(SF also accepts to serve the downloads through https since 1 or 2 years.)

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="Lossless mp3 normalizer with statistical analysis"
+DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
+volume. MP3Gain does not just do peak normalization, as many normalizers do. \
+Instead, it does some statistical analysis (using the replaygain algorithm) \
+to determine how loud the file actually sounds to the human ear.
```
fbrosson:
```diff
-Instead, it does some statistical analysis (using the replaygain algorithm) \
+Instead, it does some statistical analysis (using the replaygain algorithm) to \
-to determine how loud the file actually sounds to the human ear.
+determine how loud the file actually sounds to the human ear.
```

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,46 @@
...
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
+mp3 file directly, without decoding and re-encoding. This works with all \
+mp3 players, i.e. no support for a special tag is required."
```
fbrosson:
```diff
-mp3 file directly, without decoding and re-encoding. This works with all \
+mp3 file directly, without decoding and re-encoding. This works with all mp3 \ 
-mp3 players, i.e. no support for a special tag is required."
+players, i.e. no support for a special tag is required."
```

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,49 @@
+SUMMARY="Lossless mp3 normalizer with statistical analysis"
+DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
+volume. MP3Gain does not just do peak normalization, as many normalizers do. \
+Instead, it does some statistical analysis (using the replaygain algorithm)to \
```
fbrosson:
```diff
-Instead, it does some statistical analysis (using the replaygain algorithm)to \
+Instead, it does some statistical analysis (using the replaygain algorithm) to \
```
(you ate a white space :)

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,49 @@
...
+determine how loud the file actually sounds to the human ear.
+MP3Gain can adjust the volume in a completely lossless way by modifying the mp3 \
+file directly, without decoding and re-encoding. This works with all mp3 \
```
fbrosson:
```diff
-MP3Gain can adjust the volume in a completely lossless way by modifying the mp3 \
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
-file directly, without decoding and re-encoding. This works with all mp3 \
+mp3 file directly, without decoding and re-encoding. This works with all mp3 \
```
Line 6 was OK. Now it has is 81 chars...
(Yep, we count the backslash.)
(Sorry for being a little bit picky.)

(Everything else is OK :)

media-sound/mp3gain/mp3gain-1.5.2.recipe
korli: Well, stripping the binary is enough, who wants a debug package for every command line tool?

fbrosson: I thought it was a general recommendation to create debuginfo packages whenever that was possible.
But I'm OK with not creating the debuginfo package.

That said, I think we should not decide, ourselves, which recipes should have a debuginfo package, and which should not. We can't guess what would be the needs of all the users. If someone would like to debug that program, then we should not strip it. So having the debuginfo package is the only way to satisfy all possible users.

richienyhus:
Will fix #661

pulkomandy:
On Mon, Jan 15, 2018 at 08:19:25AM -0800, fbrosson wrote:
> That said, I think we should not decide, ourselves, which recipes should have a debuginfo package, and which should not. We can't guess what would be the needs of all the users. If someone would like to debug that program, then we should not strip it. So having the debuginfo package is the only way to satisfy all possible users.

This is not the use case for debuginfo packages. If you are going to
debug a program, you will soon need to download the sources and rebuild
it anyway. So the debuginfo package is of little use.

However, when you are debugging a program, and find out there is
something unexpected happening inside a library, then it is useful to
have the debuginfo package to see what's inside the library, while you
debug the program.

Another use case for debuginfo package is if some software is crashing
on users and the developper cannot reproduce the problem. In that case,
letting the user install the debuginfo package will let him generate a
more useful debug report. But if the program doesn't crash, there is no
need for that.

fbrosson:
OK, thanks @pulkomandy.

Hmm, do the build servers create the source packages when `DISABLE_SOURCE_PACKAGE` is not set to `yes` ?
And, if yes, are they available for `pkgman install` ?

Hmm, maybe we could modify haikuporter to have it ignore the `defineDebugInfoPackage` directive, and only create the debuginfo package if the user explicitly asks for it.

This would make everyone's life much easier. If a user wants to debug some package, he would be able to build it without having to modify the recipe.

We could even have the buildbots run haikuporter with the option to not ignore the `defineDebugInfoPackage` directive when the recipe has at least one lib in its `PROVIDES`.
What do you think?

fbrosson:
Please ignore my previous message.

Why wouldn't we create debuginfo packages for all packages where we have the `defineDebugInfoPackage` directive? After all, it allows to save some disk space.

And even if the source package is not available, haikuporter has an option to create it without building, so having the debuginfo packages for all recipes is imo a very good thing.

pulkomandy:
On Mon, Jan 15, 2018 at 06:28:52PM +0000, fbrosson wrote:
> OK, thanks @pulkomandy.
> 
> Hmm, do the build servers create the source packages when `DISABLE_SOURCE_PACKAGE` is not set to `yes` ?
> And, if yes, are they available for `pkgman install` ?

Currently, source packages are not installable (this is a bug in the
package daemon, basically).

> 
> Hmm, maybe we could modify haikuporter to have it ignore the `defineDebugInfoPackage` directive, and only create the debuginfo package if the user explicitly asks for it.
> 
> This would make everyone's life much easier. If a user wants to debug some package, he would be able to build it without having to modify the recipe.

If you are debugging something, it's unlikely that you will use
haikuporter to build it. You will run the makefile manually. This is why
it is important to upstream changes.

> 
> We could even have the buildbots run haikuporter with the option to not ignore the `defineDebugInfoPackage` directive when the recipe has at least one lib in its `PROVIDES`.
> What do you think?

I think this is useless complexity. We should define debug info packages
only where they make sense: for libraries.

pulkomandy:
On Mon, Jan 15, 2018 at 06:35:32PM +0000, fbrosson wrote:
> Please ignore my previous message.
> 
> Why wouldn't we create debuginfo packages for all packages where we have the `defineDebugInfoPackage` directive? After all, it allows to save some disk space.

The problem is that it will automatically be added to the haikuports
repository. If there was a way to filter that after building, then yes,
that may make sense. But at the moment there isn't such a thing.

And I see no point in building a debuginfo package and then dropping it.
Stripping the binaries is enough in that case.

fbrosson:
Debian creates dbg packages for a lot of packages, including command line progs which do not provide any shared library at all. One example: acpitool.

I think we should provide debuginfo packages for all packages which already have the directive. Users would be able to install exactly what they want.
(BTW, striping the runtimes in a package and not providing its debuginfo is not, IMO, very user-friendly.)

fbrosson:
@korli, @pulkomandy, shall we keep the debuginfo for mp3gain? Or would you prefer to drop it?
@tts2k is probably waiting for us to reach some consensus.

You are here since much longer than me, so I'm not going to object anything if you don't want to have a debuginfo package for mp3gain.

mmlr:
FWIW I agree with @pulkomandy that debuginfo packages are mostly useful for libraries and should therefore only be provided for them. If an application is so buggy that it needs a debuginfo package because of all its crashes then it should IMO not be marked tested and automatically built at all. For all other cases, stripping the binaries is enough.

fbrosson:
OK then. Let's drop the debuginfo package :)
Thanks @mmlr for your comment.

@tts2k, may I update the PR for you? (I'll drop the debuginfo package and make these 2 [cosmetic changes](#pullrequestreview-88830685) in `DESCRIPTION`.)

media-sound/mp3gain/mp3gain-1.5.2.recipe
fbrosson: The paragraph about the debuginfo package in the wiki, [HaikuPorter-Guidelines#debuginfo-package](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#debuginfo-package), should probably be updated because it currently says:
> All packages should define a `debuginfo` package. This way when downloading a library user downloads only necessary data. This can be done with `defineDebugInfoPackage` macro.

The way it is written does not suggest that the macro should only be added to recipes which provide libraires. It clearly states that **all** packages should use that macro. The word "library" appears only to show that debuginfo packages are convenient.

alaviss:
@mmlr @pulkomandy 
Since `defineDebugInfoPackage` strips really well and preserve added resources (thanks @fbrosson for that), I believe we would need a simillar directive for stripping (or have haikuporter strips everything by default).

media-sound/mp3gain/mp3gain-1.5.2.recipe
alaviss: @korli It's nice when you could have somewhat meaningful backtrace when something gone wrong :). Besides, we've already spent time building binaries with debuginfo (`-O2 -g` is pretty common as default CFLAGS whenever the user don't have one specified, and also lengthen the time it takes to compile).

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,49 @@
...
+players, i.e. no support for a special tag is required."
+HOMEPAGE="http://mp3gain.sourceforge.net/"
+COPYRIGHT="2001-2009 David Robinson and Glen Sawyer
+	2002 John Zitterkopf"
```
alaviss: Missing holders:
- [2004-2009 David Lasker, Altos Design, Inc.][0]
- [1999 Mark Taylor][1]
- [2000 Don Melton][2]
- [1999][4]-[2000][3] Albert L. Faber
- [1995, 1996, 1997 Michael Hipp][5]
- [1999 Takehiro TOMINAGA][6]

[0]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mp3gain.c#l6
[1]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mpglibDBL/lame.h#l4
[2]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mpglibDBL/lame.h#l765
[3]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mpglibDBL/layer2.h#l2
[4]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mpglibDBL/machine.h#l4
[5]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mpglibDBL/layer3.c#l4
[6]: https://sourceforge.net/p/mp3gain/code/ci/cli_1_5_2_r2/tree/mp3gain/mpglibDBL/bitstream.h#l4

pulkomandy:
On Tue, Jan 16, 2018 at 12:24:20AM +0000, alaviss wrote:
> @mmlr @pulkomandy 
> Since `defineDebugInfoPackage` strips really well and preserve added resources (thanks @fbrosson for that), I believe we would need a simillar directive for stripping (or have haikuporter strips everything by default).

It seems you can use extractDebugInfo and throw away the output. This is
just a subset of defineDebugInfoPackage, which does not actually define
a package, so it is what we need here.

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,56 @@
+SUMMARY="Lossless mp3 normalizer with statistical analysis"
+DESCRIPTION="MP3Gain analyzes and adjusts mp3 files so that they have the same \
+volume. MP3Gain does not just do peak normalization, as many normalizers do. \
+Instead, it does some statistical analysis (using the replaygain algorithm) to \
+determine how loud the file actually sounds to the human ear.
+MP3Gain can adjust the volume in a completely lossless way by modifying the \
+mp3 file directly, without decoding and re-encoding. This works with all mp3 \
+players, i.e. no support for a special tag is required."
+HOMEPAGE="http://mp3gain.sourceforge.net/"
+COPYRIGHT="1995, 1996, 1997 Michael Hipp
```
fbrosson: Thanks for the update!
May I ask you one last tiny edit?
```diff
-COPYRIGHT="1995, 1996, 1997 Michael Hipp
+COPYRIGHT="1995-1997 Michael Hipp
```
(Here this is not very important because there are only 3 years in the list, but since we try do that for all our recipes, it would be more consistent with the other recipes.)
Thanks!

fbrosson: Shall we also drop the debugInfoPackage to please @korli, @pulkomandy and @mmlr ?
(This will allow to get green lights for the merge ;)

fbrosson: It seems we can:
```diff
-defineDebugInfoPackage mp3gain \
-	$binDir/mp3gain
+extractDebugInfoPackage mp3gain
```

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,56 @@
...
+
+BUILD()
+{
+	make OSTYPE=beos
+}
+
+INSTALL()
+{
+	install -d $binDir
+	install -t $binDir mp3gain
```
fbrosson:
```diff
-	install -t $binDir mp3gain
+	install -t $binDir --strip mp3gain
```

media-sound/mp3gain/mp3gain-1.5.2.recipe
```diff
@@ -0,0 +1,56 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	"
+
+defineDebugInfoPackage mp3gain \
+	$binDir/mp3gain
```
fbrosson:
```diff
 	cmd:make
 	"
-
-defineDebugInfoPackage mp3gain \
-	$binDir/mp3gain
 
 BUILD()
```

PR #2124, opened 2018-01-15
gsasl: new recipe

sfanxiang:
(Done as part of GCI 2017.)

### DIFF
--- a/net-misc/gsasl/gsasl-1.8.0.recipe
+++ b/net-misc/gsasl/gsasl-1.8.0.recipe
@@ -1,12 +1,12 @@
 SUMMARY="The GNU SASL tool"
 DESCRIPTION="GNU SASL is an implementation of the Simple Authentication and \
 Security Layer framework and a few common SASL mechanisms. SASL is used by \
-network servers (e.g., IMAP, SMTP) to request authentication from clients, \
-and in clients to authenticate against servers.
+network servers (e.g., IMAP, SMTP) to request authentication from clients, and \
+in clients to authenticate against servers.
 
 Supported mechanisms are ANONYMOUS, EXTERNAL, LOGIN, PLAIN, SECURID, NTLM, \
 DIGEST-MD5, CRAM-MD5, SCRAM-SHA-1, SCRAM-SHA-1-PLUS, GS2-KRB5, GSSAPI."
-HOMEPAGE="http://www.gnu.org/software/gsasl"
+HOMEPAGE="https://www.gnu.org/software/gsasl/"
 COPYRIGHT="2004-2012 Simon Josefsson"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
@@ -37,6 +37,9 @@ BUILD_PREREQUIRES="
 	cmd:libtoolize
 	cmd:make
 	"
+
+defineDebugInfoPackage gsasl \
+	$binDir/gsasl
 
 BUILD()
 {
--- a/net-misc/gsasl/gsasl-1.8.0.recipe
+++ b/net-misc/gsasl/gsasl-1.8.0.recipe
@@ -66,4 +66,6 @@ INSTALL()
 
 	rm -rf $libDir
 	rm -rf $includeDir
+	rm -rf $manDir/man3
+	rm -rf $infoDir
 }
--- a/net-misc/gsasl/gsasl-1.8.0.recipe
+++ b/net-misc/gsasl/gsasl-1.8.0.recipe
@@ -24,13 +24,23 @@ PROVIDES="
 REQUIRES="
 	haiku
 	lib:libgsasl
+	lib:libgnutls
+	lib:libiconv
+	lib:libintl
 	"
 
 BUILD_REQUIRES="
 	haiku_devel
+	devel:libgnutls
 	devel:libgsasl
+	devel:libiconv
+	devel:libintl
 	"
 BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
 #	cmd:dia
 	cmd:gcc
 	cmd:ld
--- a/net-misc/gsasl/gsasl-1.8.0.recipe
+++ b/net-misc/gsasl/gsasl-1.8.0.recipe
@@ -23,8 +23,8 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku
-	lib:libgsasl
 	lib:libgnutls
+	lib:libgsasl
 	lib:libiconv
 	lib:libintl
 	"
--- a/net-misc/gsasl/gsasl-1.8.0.recipe
+++ b/net-misc/gsasl/gsasl-1.8.0.recipe
@@ -1,0 +1,56 @@
+SUMMARY="The GNU SASL tool"
+DESCRIPTION="GNU SASL is an implementation of the Simple Authentication and \
+Security Layer framework and a few common SASL mechanisms. SASL is used by \
+network servers (e.g., IMAP, SMTP) to request authentication from clients, \
+and in clients to authenticate against servers.
+
+Supported mechanisms are ANONYMOUS, EXTERNAL, LOGIN, PLAIN, SECURID, NTLM, \
+DIGEST-MD5, CRAM-MD5, SCRAM-SHA-1, SCRAM-SHA-1-PLUS, GS2-KRB5, GSSAPI."
+HOMEPAGE="http://www.gnu.org/software/gsasl"
+COPYRIGHT="2004-2012 Simon Josefsson"
+LICENSE="GNU GPL v3
+	GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="https://ftp.gnu.org/gnu/gsasl/gsasl-$portVersion.tar.gz"
+CHECKSUM_SHA256="310262d1ded082d1ceefc52d6dad265c1decae8d84e12b5947d9b1dd193191e5"
+PATCHES="gsasl-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	gsasl = $portVersion
+	cmd:gsasl = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libgsasl
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libgsasl
+	"
+BUILD_PREREQUIRES="
+#	cmd:dia
+	cmd:gcc
+	cmd:ld
+	cmd:libtoolize
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure ./configure
+
+	# ignore lack of dia
+	make -i $jobArgs
+}
+
+INSTALL()
+{
+	# ignore lack of dia
+	make -i install
+
+	rm -rf $libDir
+	rm -rf $includeDir
+}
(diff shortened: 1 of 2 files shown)


net-misc/gsasl/gsasl-1.8.0.recipe
```diff
@@ -0,0 +1,56 @@
+SUMMARY="The GNU SASL tool"
+DESCRIPTION="GNU SASL is an implementation of the Simple Authentication and \
+Security Layer framework and a few common SASL mechanisms. SASL is used by \
+network servers (e.g., IMAP, SMTP) to request authentication from clients, \
+and in clients to authenticate against servers.
```
fbrosson:
```diff
-network servers (e.g., IMAP, SMTP) to request authentication from clients, \
+network servers (e.g., IMAP, SMTP) to request authentication from clients, and \
-and in clients to authenticate against servers.
+in clients to authenticate against servers.
```

net-misc/gsasl/gsasl-1.8.0.recipe
```diff
@@ -0,0 +1,56 @@
...
+
+Supported mechanisms are ANONYMOUS, EXTERNAL, LOGIN, PLAIN, SECURID, NTLM, \
+DIGEST-MD5, CRAM-MD5, SCRAM-SHA-1, SCRAM-SHA-1-PLUS, GS2-KRB5, GSSAPI."
+HOMEPAGE="http://www.gnu.org/software/gsasl"
```
fbrosson:
```diff
-HOMEPAGE="http://www.gnu.org/software/gsasl"
+HOMEPAGE="https://www.gnu.org/software/gsasl/"
```
( `http` -> `https` & `/gsasl` -> `/gsasl/` )

net-misc/gsasl/gsasl-1.8.0.recipe
```diff
@@ -0,0 +1,56 @@
...
+	"
+BUILD_PREREQUIRES="
+#	cmd:dia
+	cmd:gcc
+	cmd:ld
+	cmd:libtoolize
+	cmd:make
+	"
+
+BUILD()
```
fbrosson:
```diff
 	cmd:make
 	"
+
+defineDebugInfoPackage gsasl \
+	$binDir/gsasl
+
 BUILD()
```

net-misc/gsasl/gsasl-1.8.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	gsasl = $portVersion
+	cmd:gsasl = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libgsasl
+	"
```
fbrosson: I saw this in `config.log`:
```
configure:17001: WARNING: GnuTLS not found.  STARTTLS support disabled.
configure:17004: checking if STARTTLS support via GnuTLS should be built
configure:17006: result: no
```
I noticed that the it also checks the availability of libiconv and libintl, so maybe we could also add all this stuff:
```diff
 REQUIRES="
 	haiku
 	lib:libgsasl
+	lib:libgnutls
+	lib:libiconv
+	lib:libintl
 	"
```
(I don't know if this is a good idea, I'm trying it right now.)

net-misc/gsasl/gsasl-1.8.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libgsasl
+	"
```
fbrosson:
```diff
 BUILD_REQUIRES="
 	haiku_devel
+	devel:libgnutls
 	devel:libgsasl
+	devel:libiconv
+	devel:libintl
 	"
 BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoheader
+	cmd:automake
 #	cmd:dia
```
I'm trying to see if we could call `make` without the `-i` and avoid the error with by changing something, maybe this line in `configure.ac`:
```
AM_MISSING_PROG(DIA, dia, $missing_dir)
```
But I did not succeed yet.
`dia` seems to be needed to create a png file. So I guess this is only for the docs.
I tried to pass `--disable-gtk-doc --disable-gtk-doc-html` to configure but it does not help...

net-misc/gsasl/gsasl-1.8.0.recipe
fbrosson: Oops, sorry, I should have placed libgnutls above libgsasl, not bellow.
Could you please swap these 2 lines, so that all `lib:lib*` lines be sorted alphabetically? Thanks!

net-misc/gsasl/gsasl-1.8.0.recipe
```diff
@@ -0,0 +1,69 @@
...
+}
+
+INSTALL()
+{
+	# ignore lack of dia
+	make -i install
+
+	rm -rf $libDir
+	rm -rf $includeDir
+}
```
fbrosson:
```diff
 	rm -rf $libDir
 	rm -rf $includeDir
+	rm -rf $manDir/man3
+	rm -rf $infoDir
 }
```
I'm not sure about the infoDir. Shall we keep it? It allows to have `/system/documentation/info/gsasl.info` which should probably in be another package, libgsasl_devel, but it is not. So maybe we can keep it here until we add it to the recipe in `net-libs/libgsasl/`.
What do you think?
Any thoughts from other contributors?

BTW, I wonder if it would not be possible to drop the patch (or parts of it) by passing `CPPFLAGS="-D_BSD_SOURCE"` and `LIBS="-lbsd"` to `make`.
Would you like to investigate on this? (We can keep the PR open for a few more days if you wish, as there is no urgency.)

sfanxiang: About the patch, poll() is available without _BSD_SOURCE, but polling stdin strangely fails, and that forced me to change to the select() equivalent.

fbrosson: OK, thanks.
If you are sure there is no way to shrink the patch then I can merge the PR soon after you add the `rm $manDir/man3`.
But if you agree, we can keep the PR open for a few days to see if someone finds a way to either avoid the patch or shrink it.

mmlr: If poll() is broken and select() works, that hints at a bug in Haiku. Please file a bug report at https://dev.haiku-os.org/ regarding that, pointing at the reproducible test case.

fbrosson: @mmlr, shall I merge this PR?

fbrosson: @sfanxiang, I'll let you open the bug report :)

sfanxiang: @fbrosson
I just figured that polling stdin is successful in this test program:
```c
#include <poll.h>
#include <stdio.h>
#include <unistd.h>

void main()
{
	struct pollfd pfd;
	pfd.fd = STDIN_FILENO;
	pfd.events = POLLIN;
	pfd.revents = 0;

	poll(&pfd, 1, -1);
	printf("%d %d\n", pfd.revents & POLLIN, pfd.revents & POLLERR);
}
```

fbrosson: Well, it _does_ wait until we type something, and it _does_ not consume any character from what we type, but it _only_ quits _if/when_ we press return.
You can try. Run:
```
./stdin-poll-test ; date
```
Now type `date` but don't press return yet. First wait for e.g. 10 seconds and then press return.
You will notice that the program only exits after you press return.

I get the same results on all architectures. (x86_gcc2 primary, x86 secondary, and x86_64.)

So it looks like poll is _not_ working as expected.

sfanxiang: @fbrosson
Thanks for testing.

> Well, it does wait until we type something, and it does not consume any character from what we type, but it only quits if/when we press return.

This is the expected behavior. The terminal by default sends the input only when a delimiter is reached (return, EOF, ...). What I meant by "not working" is that, in gsasl (but not in the test program), polling stdin results in POLLERR.

fbrosson: OK, I see. My mistake :)

sfanxiang:
In gsasl, `poll()` is a wrapper function (part of gnulib) implemented using `select()`. It calls `recv()` to determine the fd status, but `recv()` errors in `ENOTSOCK` for stdin. Released 6 years ago shipping gnulib with it, gsasl 1.8.0 didn't properly handle this case ([link](https://git.savannah.gnu.org/cgit/gsasl.git/tree/gl/poll.c?h=gsasl-1-8-0&id=524db10028ee3c2ea6f7f89ecf0be087b0dc7c2e#n290)), although it has been fixed upstream.

I think the best solution is to patch `poll()` in gsasl. @fbrosson What do you think?

pulkomandy:
We do have a native poll() in Haiku, why is it not being used?

fbrosson:
Thanks @pulkomandy for the hint!

@sfanxiang, if I'm not mistaken, `-D_POSIX_C_SOURCE` needs to be added to the `CFLAGS` (or `CPPFLAGS`). (This is because Haiku's `poll.h` is in `/system/develop/headers/posix/`.)

Would you like to improve the patch you wrote for libdiscid? (This would be great!)
(Don't forget, however, that this can wait until you have enough free time.)

pulkomandy:
Actually, we don't protect posix stuff with _POSIX_C_SOURCE.

We do this for BSD extensions with _DBSD_SOURCE because this is required for strict POSIX conformity (POSIX headers are normally not allowed to provide more functions than what's in the spec)

You can check in the header file that there is not any _POSIX_C_SOURCE or other guards: https://git.haiku-os.org/haiku/tree/headers/posix/poll.h

So, we need to investigate why the buildsystem decided against using it. Probably a check of the config.log would reveal something.

fbrosson:
OK. Thanks for the info :)

Regarding the "include" vs "headers" directory name, would it be acceptable to add to haiku_devel a symlink from the former to the latter? Or would this be unthinkable?
(BTW, I guess this name came from BeOS.)

sfanxiang:
> So, we need to investigate why the buildsystem decided against using it.

Because this check (extracted from configure) is failing:
```c
#include <fcntl.h>
#include <poll.h>
         int main()
         {
           int result = 0;
           struct pollfd ufd;
           /* Try /dev/null for reading.  */
           ufd.fd = open ("/dev/null", O_RDONLY);
           /* If /dev/null does not exist, it's not MacOS X nor AIX. */
           if (ufd.fd >= 0)
             {
               ufd.events = POLLIN;
               ufd.revents = 0;
               if (!(poll (&ufd, 1, 0) == 1 && ufd.revents == POLLIN))
                 result |= 1;
             }
           /* Try /dev/null for writing.  */
           ufd.fd = open ("/dev/null", O_WRONLY);
           /* If /dev/null does not exist, it's not MacOS X nor AIX. */
           if (ufd.fd >= 0)
             {
               ufd.events = POLLOUT;
               ufd.revents = 0;
               if (!(poll (&ufd, 1, 0) == 1 && ufd.revents == POLLOUT))
                 result |= 2;
             }
           /* Trying /dev/tty may be too environment dependent.  */
           return result;
         }
```

The first `poll()`, for example, returns `POLLIN | POLLERR | POLLHUP` in `revents`. Perhaps something is wrong with Haiku's implementation too?

pulkomandy:
It's possible that our /dev/null behaves unusually on this. poll itself is just deffering the work to underlying file systems, so things can be wrong on that side, especially in special cases like polling /dev/null (why would one do that?)

sfanxiang:
I should have checked `POLLOUT` - polling `/dev/null` with `POLLOUT` results in `POLLOUT | POLLERR | POLLHUP`, but in POSIX we have: (emphasis mine)
> POLLHUP:
> ... This event and POLLOUT are __mutually-exclusive__; a stream can never be writable if a hangup has occurred.

pulkomandy:
Time to report the bug on Haiku bugtracker then.

sfanxiang:
Ticket: https://dev.haiku-os.org/ticket/13965

PR #2126, opened 2018-01-15
dev-python/cython: version bump.

netpositive36:


korli:
I don't understand why there are four recipes in the end...

netpositive36:
Fixed.

pulkomandy:
You are renaming the recipes (from python_cython to cython_python). This is fine, but you must add a REPLACE entry to make sure users get the update automatically.

netpositive36:
Thanks for the notification. It's actually REPLACES and the [documentation](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes) has a broken link about this entry.

pulkomandy:
Fixed the broken links.

Some people at the Haiku project like to break links, and also they think serving wiki pages in a nicely formated HTML is useless, so now you get a link to the sourcecode only. How nice of them.

-- 
Adrien.

netpositive36:
It links a markdown (plain text) page now.

dev-python/cython/cython_python-0.27.3.recipe
```diff
@@ -3,11 +3,13 @@ DESCRIPTION="Cython is an optimising static compiler for both the Python \
...
 Pyrex). It makes writing C extensions for Python as easy as Python itself."
 HOMEPAGE="http://cython.org/"
-COPYRIGHT="2007-2016 Stefan Behnel, Robert Bradshaw, et al."
+COPYRIGHT="2007-2017 Stefan Behnel, Robert Bradshaw, et al."
 LICENSE="Apache v2"
-REVISION="2"
-SOURCE_URI="https://pypi.python.org/packages/b7/67/7e2a817f9e9c773ee3995c1e15204f5d01c8da71882016cac10342ef031b/Cython-0.25.2.tar.gz"
-CHECKSUM_SHA256="f141d1f9c27a07b5a93f7dc5339472067e2d7140d1c5a9e20112a5665ca60306"
+REVISION="1"
+REPLACES="python_cython-0.25.2"
```
korli: This REPLACES is useless, right?

dev-python/cython/cython_python-0.27.3.recipe
```diff
@@ -22,7 +24,9 @@ REQUIRES="
 	haiku
 	cmd:python2
 	"
-
+REPLACES="
+	python_cython
```
korli: Please also fix the PROVIDES (which ATM is the same as the REPLACES)

dev-python/cython/cython_python3-0.27.3.recipe
```diff
@@ -23,7 +24,9 @@ REQUIRES="
 	haiku$secondaryArchSuffix
 	cmd:python3
 	"
-
+REPLACES="
+	python3_cython
```
korli: Please fix the PROVIDES

dev-python/cython/cython_python-0.27.3.recipe
pulkomandy: The recipe is renamed from python_cython to cython_python.

The REPLACES is needed to make sure pkgman does the update in that case.

netpositive36:
@korli : thanks for the comments, duplicate REPLACES removed.

alaviss:
Both packages seems to be built from the same source. I think this package should be updated to use constructs simillar to https://github.com/haikuports/haikuports/blob/259001925d643c9f37843ff5712a42b8c4b4edbe/dev-python/markdown/markdown-2.6.9.recipe#L62

Also, one could be interested in this https://github.com/cython/cython/releases for shorter `SOURCE_URI`

netpositive36:
@alaviss Thanks for the comments. I'd stick with the separated recipes, imo it's simpler, more readable and flexible solution.

PR #2127, opened 2018-01-16
PEGTL: Initial recipe

tts2k:


dev-libs/pegtl/pegtl-2.3.3.recipe
```diff
@@ -0,0 +1,42 @@
...
+DESCRIPTION="The Parsing Expression Grammar Template Library (PEGTL) is a \
+zero-dependency C++11 header-only parser combinator library for creating \
+parsers according to a Parsing Expression Grammar (PEG)."
+HOMEPAGE="https://github.com/taocpp/PEGTL"
+COPYRIGHT="2014-2018 Dr. Colin Hirsch and Daniel Frey"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="$HOMEPAGE/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="a734e4e7a913e1243bc066a50a8448c241b036e017a26c6d00d366c11ac7dd05"
+SOURCE_FILENAME="$portVersion.tar.gz"
```
fbrosson:
```diff
-SOURCE_FILENAME="$portVersion.tar.gz"
+SOURCE_FILENAME="PEGTL-$portVersion.tar.gz"
```
(I guess this is what you meant to do.)

dev-libs/pegtl/pegtl-2.3.3.recipe
```diff
@@ -0,0 +1,42 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:g++
+	cmd:make
+	"
+
+INSTALL()
+{
+	install -d $includeDir
```
fbrosson: Shall we follow Debian's layout? They put the header files directly in `includes/pegtl` instead of `includes/tao/pegtl` and they also ship the README and the examples. These changes do just this. What do you think?
```diff
 INSTALL()
 {
-	install -d $includeDir
+	install -d "$includeDir" "$developDocDir"/examples
-	cp -r include/tao $includeDir
+	cp -r include/tao/* "$includeDir"
+	install -t "$developDocDir" README.md
+	install -t "$developDocDir"/examples src/example/pegtl/*
 }
```

BTW, the tests are OK on x86_64 but are failing on x86_gcc2.

tts2k: Ya, i agree with you. Didn't notice that README also have instructions

tts2k: Also, are those quotation marks needed ?

PR #2128, opened 2018-01-16
libdiscid: fix recipe

sfanxiang:
(Done as part of GCI 2017.)

### DIFF
--- a/media-libs/libdiscid/libdiscid-0.6.2.recipe
+++ b/media-libs/libdiscid/libdiscid-0.6.2.recipe
@@ -5,7 +5,7 @@ generates an identifier which can be used to lookup the CD at MusicBrainz. \
 Additionally, it provides a submission URL for adding the disc ID to the \
 database and gathers ISRCs and the MCN (=UPC/EAN) from disc."
 HOMEPAGE="https://musicbrainz.org/doc/libdiscid"
-COPYRIGHT="2017 Robert Kaye et al"
+COPYRIGHT="2013-2017 Robert Kaye et al"
 LICENSE="GNU LGPL v2.1"
 REVISION="3"
 SOURCE_URI="http://ftp.musicbrainz.org/pub/musicbrainz/libdiscid/libdiscid-$portVersion.tar.gz"
@@ -15,9 +15,12 @@ PATCHES="libdiscid-$portVersion.patchset"
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libdiscid$secondaryArchSuffix = $portVersion
-	lib:libdiscid$secondaryArchSuffix = 0.6.2 compat >= 0
+	lib:libdiscid$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
@@ -26,7 +29,7 @@ REQUIRES="
 
 PROVIDES_devel="
 	libdiscid${secondaryArchSuffix}_devel = $portVersion
-	devel:libdiscid$secondaryArchSuffix = 0.6.2 compat >= 0
+	devel:libdiscid$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES_devel="
 	libdiscid$secondaryArchSuffix == $portVersion base
@@ -47,6 +50,9 @@ BUILD_PREREQUIRES="
 	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	"
+
+defineDebugInfoPackage libdiscid$secondaryArchSuffix \
+	$libDir/libdiscid.so.$libVersion
 
 BUILD()
 {
--- a/media-libs/libdiscid/libdiscid-0.6.2.recipe
+++ b/media-libs/libdiscid/libdiscid-0.6.2.recipe
@@ -37,9 +37,9 @@ BUILD_REQUIRES="
 	devel:libmusicbrainz5$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
-	cmd:aclocal$secondaryArchSuffix
-	cmd:automake$secondaryArchSuffix
-	cmd:autoreconf$secondaryArchSuffix
+	cmd:aclocal
+	cmd:automake
+	cmd:autoreconf
 	cmd:awk
 	cmd:cmp
 	cmd:gcc$secondaryArchSuffix
--- a/media-libs/libdiscid/libdiscid-0.6.2.recipe
+++ b/media-libs/libdiscid/libdiscid-0.6.2.recipe
@@ -4,14 +4,15 @@ disc IDs from audio CDs. It reads a CD's table of contents (TOC) and \
 generates an identifier which can be used to lookup the CD at MusicBrainz. \
 Additionally, it provides a submission URL for adding the disc ID to the \
 database and gathers ISRCs and the MCN (=UPC/EAN) from disc."
-HOMEPAGE="http://musicbrainz.org/doc/libdiscid"
-COPYRIGHT="2012 Andrew Hawkins"
+HOMEPAGE="https://musicbrainz.org/doc/libdiscid"
+COPYRIGHT="2017 Robert Kaye et al"
 LICENSE="GNU LGPL v2.1"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="http://ftp.musicbrainz.org/pub/musicbrainz/libdiscid/libdiscid-$portVersion.tar.gz"
 CHECKSUM_SHA256="f9e443ac4c0dd4819c2841fcc82169a46fb9a626352cdb9c7f65dd3624cd31b9"
+PATCHES="libdiscid-$portVersion.patchset"
 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
@@ -36,13 +37,20 @@ BUILD_REQUIRES="
 	devel:libmusicbrainz5$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
+	cmd:aclocal$secondaryArchSuffix
+	cmd:automake$secondaryArchSuffix
+	cmd:autoreconf$secondaryArchSuffix
 	cmd:awk
 	cmd:cmp
 	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	"
+
 BUILD()
 {
+	autoreconf
 	runConfigure ./configure
 	make $jobArgs
 }
@@ -64,4 +72,3 @@ TEST()
 {
 	make check
 }
-
--- a/media-libs/libdiscid/libdiscid-0.6.2.recipe
+++ b/media-libs/libdiscid/libdiscid-0.6.2.recipe
@@ -5,7 +5,7 @@ generates an identifier which can be used to lookup the CD at MusicBrainz. \
 Additionally, it provides a submission URL for adding the disc ID to the \
 database and gathers ISRCs and the MCN (=UPC/EAN) from disc."
 HOMEPAGE="https://musicbrainz.org/doc/libdiscid"
-COPYRIGHT="2013-2017 Robert Kaye et al"
+COPYRIGHT="2006-2017 MetaBrainz Foundation"
 LICENSE="GNU LGPL v2.1"
 REVISION="3"
 SOURCE_URI="http://ftp.musicbrainz.org/pub/musicbrainz/libdiscid/libdiscid-$portVersion.tar.gz"
@@ -64,6 +64,8 @@ BUILD()
 INSTALL()
 {
 	make install
+	install -d $docDir
+	install -t $docDir AUTHORS README
 
 	rm $libDir/libdiscid.la
 
--- a/media-libs/libdiscid/libdiscid-0.6.2.recipe
+++ b/media-libs/libdiscid/libdiscid-0.6.2.recipe
@@ -12,7 +12,7 @@ SOURCE_URI="http://ftp.musicbrainz.org/pub/musicbrainz/libdiscid/libdiscid-$port
 CHECKSUM_SHA256="f9e443ac4c0dd4819c2841fcc82169a46fb9a626352cdb9c7f65dd3624cd31b9"
 PATCHES="libdiscid-$portVersion.patchset"
 
-ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
 libVersion="$portVersion"
@@ -56,7 +56,7 @@ defineDebugInfoPackage libdiscid$secondaryArchSuffix \
 
 BUILD()
 {
-	autoreconf
+	autoreconf -fi
 	runConfigure ./configure
 	make $jobArgs
 }
--- a/media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
+++ b/media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
@@ -1,4 +1,4 @@
-From b15b936bd13b2d0fc1e280a55c8e8dd3e9839714 Mon Sep 17 00:00:00 2001
+From 7d3edd19c385a5f6869acd8bc285ff7abe1a51d3 Mon Sep 17 00:00:00 2001
 From: sfanxiang <sfanxiang@gmail.com>
 Date: Mon, 15 Jan 2018 02:01:29 +0000
 Subject: port to Haiku
@@ -46,10 +46,10 @@ index 0d21232..accef71 100644
  AM_CONDITIONAL([OS_GENERIC], [test x$os = xgeneric])
 diff --git a/src/disc_haiku.c b/src/disc_haiku.c
 new file mode 100644
-index 0000000..6c07574
+index 0000000..0680884
 --- /dev/null
 +++ b/src/disc_haiku.c
-@@ -0,0 +1,125 @@
+@@ -0,0 +1,126 @@
 +/* --------------------------------------------------------------------------
 +
 +   MusicBrainz -- The Internet music metadatabase
@@ -104,7 +104,7 @@ index 0000000..6c07574
 +	scsi_toc toc;
 +	int ret = ioctl(fd, B_SCSI_GET_TOC, &toc);
 +
-+	if ( ret < 0 )
++	if (ret == -1)
 +		return 0; /* error */
 +
 +	disk_toc->first_track_num = toc.toc_data[2];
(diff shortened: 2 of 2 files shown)


media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -36,13 +37,20 @@ BUILD_REQUIRES="
 	devel:libmusicbrainz5$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
+	cmd:aclocal$secondaryArchSuffix
+	cmd:automake$secondaryArchSuffix
+	cmd:autoreconf$secondaryArchSuffix
```
fbrosson:
```diff
 BUILD_PREREQUIRES="
-	cmd:aclocal$secondaryArchSuffix
+	cmd:aclocal
-	cmd:automake$secondaryArchSuffix
+	cmd:automake
-	cmd:autoreconf$secondaryArchSuffix
+	cmd:autoreconf
```
(Your new lines 46 and 47 are correct.)

media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -36,13 +37,20 @@ BUILD_REQUIRES="
...
+	cmd:autoreconf
 	cmd:awk
 	cmd:cmp
 	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	"
+
 BUILD()
```
fbrosson: Shall we add a debuginfo package?
```diff
 	cmd:make
 	"
+
+defineDebugInfoPackage libdiscid$secondaryArchSuffix \
+	$libDir/libdiscid.so.$libVersion
+
 BUILD()
```

media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -4,14 +4,15 @@ disc IDs from audio CDs. It reads a CD's table of contents (TOC) and \
...
 LICENSE="GNU LGPL v2.1"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="http://ftp.musicbrainz.org/pub/musicbrainz/libdiscid/libdiscid-$portVersion.tar.gz"
 CHECKSUM_SHA256="f9e443ac4c0dd4819c2841fcc82169a46fb9a626352cdb9c7f65dd3624cd31b9"
+PATCHES="libdiscid-$portVersion.patchset"
 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
```
fbrosson: Since you are fixing this recipe, could you please also make this small change?
```diff
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libdiscid$secondaryArchSuffix = $portVersion
-	lib:libdiscid$secondaryArchSuffix = 0.6.2 compat >= 0
+	lib:libdiscid$secondaryArchSuffix = $libVersionCompat
 	"
```
And a few lines below:
```diff
 PROVIDES_devel="
 	libdiscid${secondaryArchSuffix}_devel = $portVersion
-	devel:libdiscid$secondaryArchSuffix = 0.6.2 compat >= 0
+	devel:libdiscid$secondaryArchSuffix = $libVersion
 	"
```
(I can't attach any comment to the PROVIDES_devel bloc because it is not near a modified line.)

@alaviss, what would you put here, for the PROVIDES_devel? Would you use `$libVersion` or `$libVersionCompat`?

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,176 @@
...
++
++
++#include "discid/discid.h"
++#include "discid/discid_private.h"
++#include "unix.h"
++
++#define NUM_CANDIDATES 2
++
++static char *device_candidates[NUM_CANDIDATES] = {"/dev/disk/atapi/0/slave/raw",
++					         "/dev/disk/atapi/1/slave/raw"};
```
fbrosson: Would libdiskid be able to work on devices if they are not in this candidates list?
If not, would it be easily feasible to handle any device instead of just those in the candidate list?

media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -4,14 +4,15 @@ disc IDs from audio CDs. It reads a CD's table of contents (TOC) and \
 generates an identifier which can be used to lookup the CD at MusicBrainz. \
 Additionally, it provides a submission URL for adding the disc ID to the \
 database and gathers ISRCs and the MCN (=UPC/EAN) from disc."
-HOMEPAGE="http://musicbrainz.org/doc/libdiscid"
-COPYRIGHT="2012 Andrew Hawkins"
+HOMEPAGE="https://musicbrainz.org/doc/libdiscid"
+COPYRIGHT="2017 Robert Kaye et al"
```
fbrosson:
```diff
-COPYRIGHT="2017 Robert Kaye et al"
+COPYRIGHT="2013-2017 Robert Kaye et al"
```

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
sfanxiang: > Would libdiskid be able to work on devices if they are not in this candidates list?

Yes, but it cannot automatically detect them.

By specifying `device` here: https://github.com/metabrainz/libdiscid/blob/1a39085ae119c71ed6c113c3e5dda47c4aee06fa/src/disc.c#L160, programs can use other devices given by the user.

It is feasible, though, that we can use some Haiku native API (or simply read the directory of "/dev/disk") to completely solve the problem, but an easier solution would be to add most commonly used devices, namely (the list in the patch is incomplete):
- /dev/disk/atapi/0/master/raw
- /dev/disk/atapi/0/slave/raw
- /dev/disk/atapi/1/master/raw
- /dev/disk/atapi/1/slave/raw

What do you think?

fbrosson: Since the answer to my first question is yes I am perfectly OK with your current patch :)

Any program willing to access a device which is not in the candidate list can perfectly specify it.
And I guess most users would probably have their device in the candidate list.

So we can merge the PR.

sfanxiang: OK, thanks. Would you please wait till I update the candidates list (it's not complete)?

fbrosson: Hmm, if you would like to add the `/dev/disk/atapi/{0,1}/master/raw` I think it's a good idea. `/dev/disk/atapi/1/master/raw` can be quite common, after all.

media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -36,13 +40,23 @@ BUILD_REQUIRES="
...
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	"
+
+defineDebugInfoPackage libdiscid$secondaryArchSuffix \
+	$libDir/libdiscid.so.$libVersion
+
 BUILD()
 {
+	autoreconf
```
fbrosson:
```diff
-	autoreconf
+	autoreconf -fi
```
(This fixes the build on x86 second arch)

media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -4,19 +4,23 @@ disc IDs from audio CDs. It reads a CD's table of contents (TOC) and \
...
+COPYRIGHT="2013-2017 Robert Kaye et al"
 LICENSE="GNU LGPL v2.1"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="http://ftp.musicbrainz.org/pub/musicbrainz/libdiscid/libdiscid-$portVersion.tar.gz"
 CHECKSUM_SHA256="f9e443ac4c0dd4819c2841fcc82169a46fb9a626352cdb9c7f65dd3624cd31b9"
+PATCHES="libdiscid-$portVersion.patchset"
 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,180 @@
...
++AM_CONDITIONAL([OS_HAIKU], [test x$os = xhaiku])
+ AM_CONDITIONAL([OS_DARWIN], [test x$os = xdarwin])
+ AM_CONDITIONAL([OS_FREEBSD], [test x$os = xfreebsd])
+ AM_CONDITIONAL([OS_GENERIC], [test x$os = xgeneric])
+diff --git a/src/disc_haiku.c b/src/disc_haiku.c
+new file mode 100644
+index 0000000..6c07574
+--- /dev/null
++++ b/src/disc_haiku.c
+@@ -0,0 +1,125 @@
```
fbrosson:
```diff
-index 0000000..6c07574
+index 0000000..58f6ade
 --- /dev/null
 +++ b/src/disc_haiku.c
-@@ -0,0 +1,125 @@
+@@ -0,0 +1,126 @@
```
(Yep, I know. You don't need to make this change because `haikuporter -b` will generate this for you.)

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,180 @@
...
++	disk_toc->first_track_num = toc.toc_data[2];
++	disk_toc->last_track_num = toc.toc_data[3];
++
++	return 1;
++}
++
++
++int mb_disc_unix_read_toc_entry(int fd, int track_num, mb_disc_toc_track *track) {
++	scsi_toc toc;
++	int ret = ioctl(fd, B_SCSI_GET_TOC, &toc);
```
fbrosson:
```diff
-+	int ret = ioctl(fd, B_SCSI_GET_TOC, &toc);
++	int base, ret, address;
++
++	ret = ioctl(fd, B_SCSI_GET_TOC, &toc);
```

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,180 @@
...
++
++	if ( ret < 0 )
++		return 0; /* error */
++
++	/* leadout - track number 170 (0xaa) */
++	if (track_num == 0xaa)
++		track_num = toc.toc_data[3] + 1;
++
++	int base = ((track_num - toc.toc_data[2]) * 8) + 4;
```
fbrosson:
```diff
-+	int base = ((track_num - toc.toc_data[2]) * 8) + 4;
++	base = ((track_num - toc.toc_data[2]) * 8) + 4;
```

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,180 @@
...
++	"/dev/disk/atapi/0/slave/raw",
++	"/dev/disk/atapi/1/master/raw",
++	"/dev/disk/atapi/1/slave/raw"
++};
++
++int mb_disc_unix_read_toc_header(int fd, mb_disc_toc *disk_toc) {
++	scsi_toc toc;
++	int ret = ioctl(fd, B_SCSI_GET_TOC, &toc);
++
++	if ( ret < 0 )
```
alaviss: According to the POSIX [ioctl()][0], only `-1` is considered an error code, values other than that indicates success

```diff
-+	if ( ret < 0 )
++	if ( ret == -1 )
```

[0]: http://pubs.opengroup.org/onlinepubs/009695399/functions/ioctl.html#tag_03_282_04

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,180 @@
...
++
++	return 1;
++}
++
++
++int mb_disc_unix_read_toc_entry(int fd, int track_num, mb_disc_toc_track *track) {
++	scsi_toc toc;
++	int ret = ioctl(fd, B_SCSI_GET_TOC, &toc);
++
++	if ( ret < 0 )
```
alaviss:
```diff
-+	if ( ret < 0 )
++	if ( ret == -1 )
```

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
```diff
@@ -0,0 +1,180 @@
...
++#include "discid/discid_private.h"
++#include "unix.h"
++
++#define NUM_CANDIDATES 4
++
++static char *device_candidates[NUM_CANDIDATES] = {
++	"/dev/disk/atapi/0/master/raw",
++	"/dev/disk/atapi/0/slave/raw",
++	"/dev/disk/atapi/1/master/raw",
++	"/dev/disk/atapi/1/slave/raw"
```
alaviss: I don't really know if these are sufficient as candidates :confused:

media-libs/libdiscid/libdiscid-0.6.2.recipe
```diff
@@ -4,19 +4,23 @@ disc IDs from audio CDs. It reads a CD's table of contents (TOC) and \
 generates an identifier which can be used to lookup the CD at MusicBrainz. \
 Additionally, it provides a submission URL for adding the disc ID to the \
 database and gathers ISRCs and the MCN (=UPC/EAN) from disc."
-HOMEPAGE="http://musicbrainz.org/doc/libdiscid"
-COPYRIGHT="2012 Andrew Hawkins"
+HOMEPAGE="https://musicbrainz.org/doc/libdiscid"
+COPYRIGHT="2013-2017 Robert Kaye et al"
```
alaviss: As there're many more copyright holders, you could instead use "2006-2013 MetaBrainz Foundation" as the value for `COPYRIGHT` then install the `AUTHORS` file included in the distribution.

media-libs/libdiscid/libdiscid-0.6.2.recipe
alaviss: I would use `$libVersionCompat` as the library already provides ABI stability, so there's no doubt that the API is stable as well

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
sfanxiang: @alaviss Do you know any Haiku API that can list all cdroms? Potentially there can also be SATA/SCSI cdroms so the perfect solution would require OS support.

alaviss: Probably we would need @pulkomandy to help with that.

pulkomandy:
On Tue, Jan 16, 2018 at 08:18:46PM -0800, alaviss wrote:
> Probably we would need @pulkomandy to help with that.

My guess would be BDiskDevice{,Roster,Visitor}. I don't know how to use
it to specifically target CD drives, however. IIRC, cdrecord (or more
precisely sg-utils) will test for that using ATAPI/SCSI commands. It may
be possible to get the info from the device, still (how do we decide to
put a CD icon on devices rather than a hard disk one?)

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
fbrosson: @sfanxiang, x86_gcc2 primary arch needs the above diff.

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
fbrosson: @sfanxiang, x86_gcc2 primary arch needs the above diff.

media-libs/libdiscid/patches/libdiscid-0.6.2.patchset
sfanxiang: @fbrosson Thanks for your reminder. I know, but I was on my phone when I pushed this. I'll finish it ASAP.

fbrosson: No problem :)

media-libs/libdiscid/libdiscid-0.6.2.recipe
fbrosson: That's a good idea. Thanks @alaviss for the suggestion!
Hmm, did you really mean "2006-201**3** MetaBrainz Foundation" ?
Or should we use "2006-201**7** MetaBrainz Foundation" ?

@sfanxiang, if you agree, could you please make the following changes? Thanks!
```diff
-COPYRIGHT="2013-2017 Robert Kaye et al"
+COPYRIGHT="2006-2017 MetaBrainz Foundation"
```
(I'm not sure about the year.)

```diff
 INSTALL()
 {
 	make install
+	install -d $docDir
+	install -t $docDir AUTHORS README
 
 	rm $libDir/libdiscid.la
```

alaviss: @fbrosson Sorry, I forgot to link to [reference][0]. [Scanning][1] through the release there aren't any copyright that gone further than 2013. I suggest you fix this one.

[0]: https://github.com/metabrainz/libdiscid/blob/v0.6.2/versioninfo.rc.in#L23
[1]: https://github.com/metabrainz/libdiscid/search?p=1&q=copyright&type=&utf8=%E2%9C%93

fbrosson: No problem :)
But since the timestamp of the latest tarball, `libdiscid-0.6.2.tar.gz`, is:
```
Last-Modified: Sun, 29 Jan 2017 14:00:22 GMT
```
I'd say that we can keep 2017. What do you think?

fbrosson: Even the ChangeLog has a timestamp in 2017: `2017-01-29 12:43:54 UTC`.
How come would the recent changes not also be Copyright MetaBrainz Foundation ?

But maybe I'm missing something. Would you say that the changes after 2013 should have another copyright?
Or would you say that we should not update our copyright years if upstream forgets to do so in their files? (I don't like this idea, but that's a point of view.)

alaviss: I don't believe we've the rights to bump copyright year without authors permission. But IANAL, so we would need some legal advice from a lawyer.

fbrosson: Well, the information we provide in our recipes is not authoritative. It's only informative.
Now, there are 2 possible situations:
* if the "legally" correct year is 2013 and we write 2017, nobody will ever complain.
* if the "legally" correct year is 2017 and we write 2013, then the owners could ask: _"Why didn't you write 2017? You know the archive was released in 2017, so you should assume the authors forgot to update the metadata!"_ And we would have to answer: _"We did not dare to refresh the copyright years in our archive..."_

For these reasons I think we can keep 2017. Moreover, writing 2013 might make potential users think the library is not maintained.

sfanxiang: @fbrosson @alaviss
IANAL either, but I found this clause [here](https://www.copyright.gov/title17/92chap4.html):
> in the case of compilations or derivative works incorporating previously published material, the year date of first publication of the compilation or derivative work is sufficient.

And IMO, I can't find anything against writing 2017 as the year.

Of course, the actual situations can be much more complex.

PR #2129, opened 2018-01-16
net-p2p/hefur: new recipe

alaviss:
The last PR as part of GCI 2017

Changes to hefur:
- Make packaging easier, also fixes linking on Haiku (https://github.com/abique/hefur/pull/26)

Changes to mimosa:
- Make mimosa works on Haiku (https://github.com/abique/mimosa/pull/5)

Requires: [Ticket#13960](https://dev.haiku-os.org/ticket/13960)

Not tested on x86 secondary architecture due to problems compiling Haiku x86_gcc2 :disappointed:

### DIFF
--- a/net-p2p/hefur/hefur-0.6.recipe
+++ b/net-p2p/hefur/hefur-0.6.recipe
@@ -1,0 +1,105 @@
+SUMMARY="A standalone and lightweight BitTorrent tracker"
+DESCRIPTION="Hefur is a standalone C++ BitTorrent tracker.
+
+Features:
+ - support HTTP, HTTPS and UPD announce and scrape
+ - support both IPv4 and IPv6
+ - no database required (everything in memory)
+ - should be efficient
+ - perfect for small setup
+ - info_hash white list based on .torrent in a directory tree
+ - control interface
+ - control program: hefurctl"
+HOMEPAGE="https://github.com/abique/hefur"
+COPYRIGHT="2012 Alexandre Bique"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/abique/hefur/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="ce693eb5f0410e47cd1f65906b77c9b0179eb807943119b900a8fa8f630d0d7b"
+SOURCE_FILENAME="hefur-$portVersion.tar.gz"
+PATCHES="hefur-$portVersion.patchset" # Merged upstream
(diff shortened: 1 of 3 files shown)


korli:
Could you eventually add a haiku version requirement? See the example https://github.com/haikuports/haikuports/blob/master/dev-util/ninja/ninja-1.8.2.recipe#L29

alaviss:
Updates:
- As [Ticket#13960](https://dev.haiku-os.org/ticket/13960) is merged, a version requirement for `haiku` and `haiku_devel` has been added

net-p2p/hefur/hefur-0.6.recipe
```diff
@@ -0,0 +1,105 @@
...
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	hefur$secondaryArchSuffix = $portVersion
+	cmd:hefurctl$commandSuffix = $portVersion
+	cmd:hefurd$commandSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix >= r1~alpha4_pm_hrev51768-1
```
korli: Please drop the version requirement in the REQUIRES section.
Haikuporter will use whatever version is used to build the package instead.

alaviss: `memmem(3)` is required, which had been added to `libgnu` in `hrev51768`. Since libgnu is dynamically linked, we need the newer version resides in `hrev51768` for the port to work.

emrysn:
@alaviss Please check your efivar task for @scottmc's comment: "please create a new issue on haikuports reporting the current status of the evifar port."

net-p2p/hefur/hefur-0.6.recipe
korli: Sorry I must request once again. "Haikuporter will use whatever version is used to build the package instead."

fbrosson: @alaviss, I think the reason why @korli is asking for the removal of the reference to a specific hrev is that having such a reference causes other problems.
(Anyone failing to build hefur will just need to update Haiku.)

alaviss: I think what @korli means is `>= r1~alpha4_pm_hrev51768-1` is added by HaikuPorter at packaging anyway (this happens to libs, so I believe the same holds for everything else).

Also, I've already enforced it in `BUILD_REQUIRES`, so user wouldn't be able to build it unless they have a new enough version.

korli: Thanks!

Issue #2130, opened 2018-01-16
GNU libsigsegv 2.12 is released


bhaible:
GNU libsigsegv 2.12 is released. You find the download link at the homepage
https://www.gnu.org/software/libsigsegv/

New in 2.12:

* Added support for catching stack overflow on Hurd/i386.

* **Added support for catching stack overflow on Haiku.**

* Corrected distinction between stack overflow and other fault on AIX.

* Reliability improvements on Linux, FreeBSD, NetBSD.

You are welcome to update haikuports/dev-libs/libsigsegv/libsigsegv-2.11.recipe .

scottmc:
@bhaible thanks for the heads up!

scottmc:
@fbrosson Can you update the recipe for this to use the newest release?

fbrosson:
Done in #2139.

PR #2139, opened 2018-01-18
libsigsegv: bump to 2.12, add debuginfo.

fbrosson:


fbrosson:
Thils will close #2130.
@bhaible, did I put the right years in every line of [`COPYRIGHT`](https://github.com/fbrosson/haikuports/blob/libsigsegv/dev-libs/libsigsegv/libsigsegv-2.12.recipe#L16)? Thanks!

dev-libs/libsigsegv/libsigsegv-2.12.recipe
```diff
@@ -13,30 +13,32 @@ This library supports three sets of functions, all defined in <sigsegv.h>:
...
+	lib:libsigsegv$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
 
 PROVIDES_devel="
 	libsigsegv${secondaryArchSuffix}_devel = $portVersion
-	devel:libsigsegv$secondaryArchSuffix = $portVersion
+	devel:libsigsegv$secondaryArchSuffix = $libVersion
```
fbrosson: @alaviss, do you know if I should use `$libVersionCompat` here, instead of `$libVersion`? Thanks!
(I need to learn how to find the answer by myself...)

bhaible:
> @bhaible, did I put the right years in every line of [`COPYRIGHT`]

Yes, this is correct.

dev-libs/libsigsegv/libsigsegv-2.12.recipe
alaviss: So here's how it works. A shared library typically provides an ABI/API compatibility version, which means software linked with the old version is guaranteed to work with the new release of the same compatibility version (not always the case). This means that shared library could be swapped to a new version that fixes security/minor issues and program linked to it would still work. To achieve this, the API must also supports the same thing that the older versions (within compatibility version) does. This means it's safe to provides a compat version for the devel package as well.

TL;DR: You should use `$libVersionCompat`, as ports that use the version 2 of the API could do this

```bash
BUILD_REQUIRES="
	...
	devel:libsigsegv$secondaryArchSuffix >= 2
	...
	"
```

instead of

```bash
BUILD_REQUIRES="
	...
	devel:libsigsegv$secondaryArchSuffix >= 2.0.5
	...
	"
```

fbrosson: Thanks @alaviss.
Actually, I already knew which is the correct setting (between `$libVersionCompat` and `$libVersion`) for the `lib:libsigsegv` in `PROVIDES`, as well as for any lib in any package. The answer depends on whether or not the "make install" installs the symbolic link that "acts" as a "compat".
 ~What I don't know is wether or not the setting for `PROVIDES_devel` should always be the same as the setting for `PROVIDES`.~

OK, I think I got it now.

I have updated the recipe to use `$libVersionCompat` (instead of `$libVersion`) in `PROVIDES_devel`.
I guess it is OK now (but I'll wait for a confirmation before I merge :)

PR #2140, opened 2018-01-19
RoadFighter: override LDFLAGS

owenca:


games-arcade/roadfighter/patches/roadfighter-1.0.1269.patchset
```diff
@@ -61,3 +61,29 @@ index 5f5bfd8..284b1b9 100644
...
+--- a/build/linux/Makefile
++++ b/build/linux/Makefile
+@@ -38,10 +38,7 @@ OBJS = \
+ 
+ CC = g++
+ CFLAGS = -g3 -O3 `sdl-config --cflags` -I/usr/X11R6/include
+-LDFLAGS = `sdl-config --libs` -L/usr/X11R6/lib/ -lSDL_image -lSDL_mixer -lSDL_sound -lSDL_ttf
+-#ifdef __HAIKU__
+-LDFLAGS += -lbe
+-#endif
```
fbrosson: `build/linux/Makefile` is modified by each of the two patches in this patchset.
In the first patch, we have at lines 16 to 18:
```diff
+#ifdef __HAIKU__
+LDFLAGS += -lbe
+#endif
```
But in the second patch, we remove it with:
```diff
-#ifdef __HAIKU__
-LDFLAGS += -lbe
-#endif
```

@owenca, If you are comfortable with squashing commits, could you please combine these two patches?
I don't know if you need any hints, but here are some :)
You will need to find the hash of the "import" (run `git log` in the build directory), then type `EDITOR=vi git rebase --interactive $the_has_of_the_import` (replacing `$the_has_of_the_import` by the hash of the import) and type `j3xrsZZ`.

fbrosson: @owenca, I had a better idea and just updated your PR. (See the last commit: "RoadFighter: simplify patchset".)
If you agree I can merge the PR.

This PR does not have any impact on the resulting package, and I guess this is the reason why you did not bump `REVISION`.
But if anyone would prefer to have a `REVISION` bump, then I think we should do so.

owenca: The truth is that I forgot to bump the revision number. :(  Also, I didn’t know how to squash a patch. (Now I know, thanks to you.) Do you want me to bump the revision number before you merge the PR with your last commit?

fbrosson: Cool!
Yep, that would be nice to bump the revision. I can do it, but if you would like to practice, here are some hints:

The easiest way to copy (to your local branch) the commit I have pushed (to your remote branch) is:
1) make sure you are on your roadfighter branch.
2) run `git pull` (or `git pull origin roadfighter` if you forgot to use `--set-upstream` when you pushed your first commit.)
3) bump the revision in `roadfighter-1.0.1269.recipe`
4) `git commit -m "RoadFighter: bump revision" roadfighter-1.0.1269.recipe`

Then you just need to push. You don't even need to use the force option.

If, however, you would like to squash the 3 commits:
```
EDITOR=vi git rebase --interactive 1c85a5a8233fc3797076c634b0a9cff76fd11c3f
```
and type `j3xrsj3xrfZZ`.
Of course, if you have chosen to squash the 3 commits you will need to `git push --force`.

BTW, there is an easier method to squash the 3 most recent commits, but I haven't looked for it yet.

owenca: I had already pushed the new commit. :) I had to do a `git pull --rebase origin roadfighter` in order to avoid using `push -f`.

In hindsight, I could have squashed all 3 commits by using `git rebase -i HEAD~3`. (or is it `HEAD~2`?) I was confused when reading your hint that started with `EDITOR=vi` and thought that squashing patch commits would be somehow different...

fbrosson: Yep, I only saw that you had pushed the commit with the revision bump after I had clicked on "Comment". So you were faster than me :)

BTW, thanks for the `git rebase -i HEAD~3`! When I read about this a long time ago I was not comfortable with squashing, so I never used it, and had not realized how nice it is!

Regarding the `EDITOR=vi`, I like to specify it (when I suggest some changes in a text file) because it is much easier to describe the changes using vi commands. And yes, most users don't need to specify it because most programs which need to spawn an external editor have vi as default value :)

PR #2142, opened 2018-01-20
tartos, new recipe

Begasus:
Running the shc command from Terminal works ok, but using it here in the recipe doesn't seem to work to have a working binary ...

### DIFF
--- a/dev-util/tartos/tartos-1.0.0~git.recipe
+++ b/dev-util/tartos/tartos-1.0.0~git.recipe
@@ -1,0 +1,48 @@
+SUMMARY="TAR to shell script"
+DESCRIPTION="The tartos script is a standalone bash shell script which binds \
+different type of TAR archives to a single shell script (bash). This kind of \
+approach is generally adopted to make a single install scipt from an existing \
+tar archives with executable or installable software. It can be used to make \
+a self extracting archive/installer which does some additional jobs like \
+installing softwares, compared to the conventional tar archive files. The \
+tartos script itself is written in bash and thus depends on it and on the \
+package tar."
+HOMEPAGE="https://github.com/neurobin/tartos"
+COPYRIGHT="2015-2017 Md Jahidul Hamid"
+LICENSE="GNU GPL v3"
+REVISION="1"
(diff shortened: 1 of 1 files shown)


dev-util/tartos/tartos-1.0.0~git.recipe
```diff
@@ -0,0 +1,48 @@
...
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:shc
+	"
+
+BUILD()
+{
+	shc -f tartos.sh -o tartos
```
alaviss: Is there any good reason to compile this? You could just give the sh file executable permission then install it without the `.sh` suffix

dev-util/tartos/tartos-1.0.0~git.recipe
```diff
@@ -0,0 +1,48 @@
...
+DESCRIPTION="The tartos script is a standalone bash shell script which binds \
+different type of TAR archives to a single shell script (bash). This kind of \
+approach is generally adopted to make a single install scipt from an existing \
+tar archives with executable or installable software. It can be used to make \
+a self extracting archive/installer which does some additional jobs like \
+installing softwares, compared to the conventional tar archive files. The \
+tartos script itself is written in bash and thus depends on it and on the \
+package tar."
+HOMEPAGE="https://github.com/neurobin/tartos"
+COPYRIGHT="2015-2017 Md Jahidul Hamid"
```
alaviss: I can't find any copyright statement in the source archive. Does this mean the script is uncopyrighted?

Begasus: Not really a Copyright included, only thing popping up is 2015 and author in the man page ...

dev-util/tartos/tartos-1.0.0~git.recipe
Begasus: Maybe not ... but as it was from the same author I thought I'd give shc a try on this one to see if it could be used in a recipe ...

PR #2010, opened 2017-12-30
sys-block/sg3_utils: new recipe

alaviss:
Done as part of GCI 2017

Changes to sg3_utils:
- Added Haiku support

I would like to have some input for [my fork](https://github.com/alaviss/sg3_utils/tree/haiku) before upstreaming this patch.
**EDIT:** PR made: https://github.com/hreinecke/sg3_utils/pull/24

If you happens to have an SCSI-capable device (they could be found in `/dev/disk/scsi`), please help me test various utilities provided by this package.

Thanks

PR #2144, opened 2018-01-20
x11-libs/libXaw: new recipe.

fbrosson:


x11-libs/libxaw/libxaw-1.0.13.recipe
```diff
@@ -0,0 +1,117 @@
...
+libXaw6Version="6.0.1"
+libXaw7Version="7.0.0"
+
+libXaw6VersionCompat="$libXaw6Version compat >= ${libXaw6Version%%.*}"
+libXaw7VersionCompat="$libXaw7Version compat >= ${libXaw7Version%%.*}"
+
+PROVIDES="
+	libxaw$secondaryArchSuffix = $portVersion
+	lib:libxaw6$secondaryArchSuffix = $libXaw6VersionCompat
+	lib:libxaw7$secondaryArchSuffix = $libXaw7VersionCompat
```
alaviss: Don't worry, this fits with tabspace = 4, which is also [applied to][0] HaikuPorts

```diff
-	lib:libxaw6$secondaryArchSuffix = $libXaw6VersionCompat
-	lib:libxaw7$secondaryArchSuffix = $libXaw7VersionCompat
+	lib:libxaw${libVersionCompat##* }$secondaryArchSuffix = $libVersionCompat
+	lib:libxaw${libVersionCompat2##* }$secondaryArchSuffix = $libVersionCompat2
```

[0]: https://github.com/haikuports/haikuports/pull/2010#discussion_r159844382

x11-libs/libxaw/libxaw-1.0.13.recipe
```diff
@@ -0,0 +1,117 @@
...
+	lib:libxext$secondaryArchSuffix
+	lib:libXmu$secondaryArchSuffix
+	lib:libXpm$secondaryArchSuffix
+	lib:libXt$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libxaw${secondaryArchSuffix}_devel = $portVersion
+	devel:libXaw6$secondaryArchSuffix = $libXaw6VersionCompat
+	devel:libXaw7$secondaryArchSuffix = $libXaw7VersionCompat
```
alaviss: Oops, this part doesn't fit (1 char off). Hopefully it's OK here...

```diff
-	devel:libXaw6$secondaryArchSuffix = $libXaw6VersionCompat
-	devel:libXaw7$secondaryArchSuffix = $libXaw7VersionCompat
+	devel:libxaw${libVersionCompat##* }$secondaryArchSuffix = $libVersionCompat
+	devel:libxaw${libVersionCompat2##* }$secondaryArchSuffix = $libVersionCompat2

File: x11-libs/libxaw/libxaw-1.0.13.recipe [hunk 3]
```diff
@@ -0,0 +1,117 @@
...
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:makeinfo
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+defineDebugInfoPackage libxaw$secondaryArchSuffix \
+	$libDir/libXaw6.so.$libXaw6Version \
+	$libDir/libXaw7.so.$libXaw7Version
```
alaviss:
```diff
-	$libDir/libXaw6.so.$libXaw6Version \
-	$libDir/libXaw7.so.$libXaw7Version
+	$libDir/libXaw${libVersionCompat##* }.so.$libVersion \
+	$libDir/libXaw${libVersionCompat2##* }.so.$libVersion2

File: x11-libs/libxaw/libxaw-1.0.13.recipe [hunk 4]
```diff
@@ -0,0 +1,117 @@
...
+	rm -f $libDir/libXaw.so*
+
+	prepareInstalledDevelLibs \
+		libXaw6 \
+		libXaw7
+	fixPkgconfig
+
+	# add the appropriate symlinks
+	ln -s libXaw6.so.$libXaw6Version $libDir/libXaw6.so.6
+	ln -s libXaw7.so.$libXaw7Version $libDir/libXaw7.so.7
```
alaviss: Since the soname is

     0x000000000000000e (SONAME)             Library soname: [libXaw.so.7]

there's no need for these

```diff
-	ln -s libXaw6.so.$libXaw6Version $libDir/libXaw6.so.6
-	ln -s libXaw7.so.$libXaw7Version $libDir/libXaw7.so.7

File: x11-libs/libxaw/libxaw-1.0.13.recipe [hunk 5]
```diff
@@ -0,0 +1,117 @@
...
+SOURCE_DIR="libXaw-$portVersion"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libXaw6Version="6.0.1"
+libXaw7Version="7.0.0"
+
+libXaw6VersionCompat="$libXaw6Version compat >= ${libXaw6Version%%.*}"
+libXaw7VersionCompat="$libXaw7Version compat >= ${libXaw7Version%%.*}"
```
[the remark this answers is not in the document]
fbrosson: Hmm, that's a good idea. But I would prefer a more symetric variant:
```
libVersion6="6.0.1"
libVersion7="7.0.0"
```
(BTW, thanks @alaviss for your review! I'll answer your other comments asap.)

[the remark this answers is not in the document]
fbrosson: Thanks @alaviss for your review!

You gave me a good idea. We can have a shorter name for the variables.
But since `libXaw.so.8` will probably never be used, I think we can stick to a hardcoded "6" and  "7". The recipe will probably remain valid forever with these hardcoded major versions.
(I'll update the PR asap.)

x11-libs/libxaw/libxaw-1.0.13.recipe
```diff
@@ -0,0 +1,117 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/libXaw*.la
+
+	# remove unwanted symlinks
+	rm -f $libDir/libXaw.so*
```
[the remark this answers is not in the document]
fbrosson: I might be wrong, but given that what we have for `Libs:` in `$libDir/pkgconfig/xaw6.pc` and `$libDir/pkgconfig/xaw7.pc` is (respectively) `Libs: -L${libdir} -lXaw6` and `Libs: -L${libdir} -lXaw7`, I think we don't need to have a `libXaw.so` (nor any `libXaw.so.*`) symlinks.

pulkomandy:
On Sun, Jan 21, 2018 at 07:51:00AM -0800, fbrosson wrote:
> +	# remove unwanted symlinks
> +	rm -f $libDir/libXaw.so*
> 
> I might be wrong, but given that what we have for `Libs:` in `$libDir/pkgconfig/xaw6.pc` and `$libDir/pkgconfig/xaw7.pc` is (respectively) `Libs: -L${libdir} -lXaw6` and `Libs: -L${libdir} -lXaw7`, I think we don't need to have a `libXaw.so` (nor any `libXaw.so.*`) symlinks.

Just let prepareInstalledDevelLibs do the Right Thing (tm)?
You probably also want to use fixPkgConfig, because .pc files should
refer to $developLibDir, not $libDir.

x11-libs/libxaw/libxaw-1.0.13.recipe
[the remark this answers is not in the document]
alaviss: Library soname is what the dynamic linker uses to find a library upon program activation. You could try to run this simple test case.

    $ echo 'int main() { return 0; }' | gcc -lXaw7 -x c -
    $ ./a.out
    runtime_loader: Cannot open file libXaw.so.7: No such file or directory

fbrosson:
I tried to rely on prepareInstalledDevelLibs only, but since `make install` puts some symlinks which I think we should not keep, I chose to drop these symlinks.

I've just updated the recipe, but, of course, I won't merge it until we reach a nice consensus on everything, including the symlinks.

I can try and build something which needs libXaw7 to make sure we don't need to provide a `libXaw.so` symlink.
We have groff (which needs libXaw7 on Debian but not on Haiku) and which I could try to build against libXaw7.
There are also several other packages on Debian which also require libXaw7 and which I could try to build on Haiku to verify my claim about the unneeded/unwanted symlinks:
* xcolors
* xcal
* xless
* xbill
* xbomb
* xarclock
* xjump
* xgammon
* ddd
* xvile
* nitpic
* ngspice

(The order I used here is what I'll follow until I find something that builds.)

x11-libs/libxaw/libxaw-1.0.13.recipe
```diff
@@ -0,0 +1,117 @@
...
+
+	# fix path in the Xaw(3) man page
+	sed -i \
+		-e "s|$docDir|`finddir B_SYSTEM_DIRECTORY`/$relativeDocDir|;" \
+		$manDir/man3/Xaw.3
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+		$documentationDir
```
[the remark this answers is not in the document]
fbrosson: You are right. Good catch. Thanks!
(I've made the changes to have the doc in `$developDocDir`.)

alaviss:
They would build, but they won't [run][0]

[0]: https://github.com/haikuports/haikuports/pull/2144#discussion_r162817606

x11-libs/libxaw/libxaw-1.0.13.recipe
fbrosson: > BTW, I think we need a shorthand for `${libVersionCompat##* }`. Do you have any suggestions for the name?

Maybe we could choose `libMajorVersion`. What do you think of that?

But for this recipe I think it would be better to stick to hardcoded major versions, since `libXaw.so.8` is not part of the source tarball and is almost not used by any X client.
And if any update ever happens, it will probably be for libXaw7 only.

alaviss: `libMajorVersion` sounds good enough for me.

However, I think hardcoding also works here. HaikuPorter would catch when `lib:libxaw8` came up from nowhere. And hopefully, we would catch cases where someone forgot to bump the numbers in `INSTALL()`

x11-libs/libxaw/libxaw-1.0.13.recipe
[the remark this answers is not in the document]
fbrosson: You are right. But the following change will fix this issue :)
```diff
-	ln -s libXaw6.so.$libVersion6 $libDir/libXaw6.so.6
+	ln -s libXaw6.so.$libVersion6 $libDir/libXaw.so.6
-	ln -s libXaw7.so.$libVersion7 $libDir/libXaw7.so.7
+	ln -s libXaw7.so.$libVersion7 $libDir/libXaw.so.7
```

x11-libs/libxaw/libxaw-1.0.13.recipe
fbrosson: :)

For the second part of your comment, I'm not sure I understand correctly. Do you think we should use define and use `libMajorVersion` in this recipe?
I'd say that, for this recipe, it would not be a good idea to add this variable because we know that there libXaw8 will probably never be shipped in this tarball. And if this ever happens, we would just need to change one digit.
We don't really need more flexibility than what we already have with `libVersion6` and `libVersion7`. Don't you agree?

fbrosson:
I've dropped this:
```diff
- 	# remove unwanted symlinks
-	rm -f $libDir/libXaw.so*
```
and also made these changes:
```diff
-	# add the appropriate symlinks
+	# add missing symlinks
-	ln -s libXaw6.so.$libVersion6 $libDir/libXaw.so.6
+	ln -s libXaw6.so.$libVersion6 $libDir/libXaw6.so.6
-	ln -s libXaw7.so.$libVersion7 $libDir/libXaw.so.7
+	ln -s libXaw7.so.$libVersion7 $libDir/libXaw7.so.7
```
The recipe is now working fine and I think it is ready for a merge.
Thanks @alaviss and @pulkomandy for your feedback.
Shall I merge the PR now or would you prefer to wait until I create a recipe (e.g. xcolors) which I would use to make sure this one works perfectly?

x11-libs/libxaw/libxaw-1.0.13.recipe
alaviss: I've looked into Debian changelog. Looks like there [was][0] Xaw8, but upstream dropped it, so I don't think it's much of a concern any more. So I agree with `libVersion6` and `libVersion7`

[0]: https://tracker.debian.org/news/138754

x11-libs/libxaw/libxaw-1.0.13.recipe
```diff
@@ -0,0 +1,116 @@
...
+
+libVersion6="6.0.1"
+libVersion7="7.0.0"
+
+libVersionCompat6="$libVersion6 compat >= 6"
+libVersionCompat7="$libVersion7 compat >= 7"
+
+PROVIDES="
+	libxaw$secondaryArchSuffix = $portVersion
+	lib:libxaw$secondaryArchSuffix
```
alaviss: Do you think we need a compat version for this? If a port has been built with Xaw, HaikuPorter will complain that `lib:libxaw` is missing from `REQUIRES`. But then, a definition like this `lib:libxaw >= 6` won't work to get Xaw6

pulkomandy:
On Mon, Jan 22, 2018 at 01:10:53AM +0000, alaviss wrote:
> +PROVIDES="
> +	libxaw$secondaryArchSuffix = $portVersion
> +	lib:libxaw$secondaryArchSuffix
> 
> Do you think we need a compat version for this? If a port has been built with Xaw, HaikuPorter will complain that `lib:libxaw` is missing from `REQUIRES`. But then, a definition like this `lib:libxaw >= 6` won't work to get Xaw6

As usual there is no magic or guessing involved.
- Either the lib is named libxaw6.so*, and in this case, the PROVIDES
  should be lib:libxaw6
- Or, the lib is named libxaw.so.6*, and in this case, the PROVIDES
  should be lib:libxaw

The former form is "legacy", things were done this way before library
versionning was introduced. Sometimes it is still used when two versions
of a lib are very different (for example, SDL and SDL2) - more like a
rewrite than a new version.

If a project decides to use that scheme, you have to link with
appropriate link flags (-lSDL2 for SDL2, for example). And the
PROVIDES/REQUIRES should just match that.

In fact I don't know why we are still populating these by hand in
haikuporter. We should auto-generate them from "ls $libDir" and
"ls $developLibDir" because there isn't any intelligent work needed,
just copy-pasting versions.

fbrosson:
> The former form is "legacy", things were done this way before library versionning was introduced. Sometimes it is still used when two versions of a lib are very different (for example, SDL and SDL2) - more like a rewrite than a new version.

Yep, that's exactly what I did here :)
I **_first_** tried to limit the `lib:lib*` elements in `PROVIDES` to just `lib:libxaw6` and `lib:libxaw7` **_but_** haikuporter's strict policy mode was complaining about a missing declares for `libxaw.so` (which is just a symlink to `livXaw7.so` which is itself a symlink to `libXaw7.so.7.0.0`), so I had to add a `lib:libxaw`:
```diff
 PROVIDES="
 	libxaw$secondaryArchSuffix = $portVersion
+	lib:libxaw$secondaryArchSuffix
 	lib:libxaw6$secondaryArchSuffix = $libVersionCompat6
 	lib:libxaw7$secondaryArchSuffix = $libVersionCompat7
 	"
```

FWIW, the reason why I had to add 2 symlinks is explained here:
* **_Before_** the call to `prepareInstalledDevelLibs` we have these symlinks in `$libDir`:
```
libXaw.so -> libXaw7.so
libXaw.so.6 -> libXaw6.so.6
libXaw.so.7 -> libXaw7.so.7
libXaw6.so -> libXaw6.so.6.0.1
libXaw6.so.6 -> libXaw6.so.6.0.1
libXaw7.so -> libXaw7.so.7.0.0
libXaw7.so.7 -> libXaw7.so.7.0.0
```
* **_After_** the call to `prepareInstalledDevelLibs` we only have these (in `$libDir`):
```
libXaw.so -> libXaw7.so
libXaw.so.6 -> libXaw6.so.6
libXaw.so.7 -> libXaw7.so.7
```
* This is because `prepareInstalledDevelLibs` moves these 4 symlinks to `$developLibDir`:
```
libXaw6.so -> libXaw6.so.6.0.1
libXaw6.so.6 -> libXaw6.so.6.0.1
libXaw7.so -> libXaw7.so.7.0.0
libXaw7.so.7 -> libXaw7.so.7.0.0
```
* So this is why I had to add these symlinks in `$libDir` (after the call to `prepareInstalledDevelLibs`):
```
libXaw6.so.6 -> libXaw6.so.6.0.1
libXaw7.so.7 -> libXaw7.so.7.0.0
```

In other words, we are safe with this, thanks to these 2 "chains" of symlinks:
*  `libXaw.so.6` ➡️ `libXaw6.so.6` ➡️ `libXaw6.so.6.0.1`
*  `libXaw.so.7` ➡️ `libXaw7.so.7` ➡️ `libXaw7.so.7.0.0`


So, @pulkomandy and @alaviss, from your last comments I guess the recipe is OK now :)

PR #2146, opened 2018-01-21
Bringing MailNews closer to build

humdingerb:
Doesn't build yet. I'm not good at this... if anyone wants to continue.

Building fails with:

configure: error: Library requirements (libIDL-2.0 >= 0.8.0) not met;
consider adjusting the PKG_CONFIG_PATH environment variable if your
libraries are in a nonstandard prefix so pkg-config can find them.

FYI, libidl-0.8.14.recipe declares:

PROVIDES_devel="
	libidl${secondaryArchSuffix}_devel = $portVersion
	cmd:libIDL_config_2$secondaryArchSuffix = $portVersion
	devel:libIDL_2$secondaryArchSuffix = 0.0.0 compat >= 0
	"

### DIFF
--- a/mail-client/mailnews/mailnews-2.0.0.25.recipe
+++ b/mail-client/mailnews/mailnews-2.0.0.25.recipe
@@ -1,62 +1,138 @@
-SUMMARY="Fork of Mozilla Thunderbird for Haiku"
+SUMMARY="A fork of Mozilla's email client Thunderbird"
 DESCRIPTION="Thunderbird is an open source mail client. \
 It achieves balance between ease of use and customization, catering to the \
-needs of both casual and power users."
+needs of both casual and power users. So does its fork MailNews."
 HOMEPAGE="https://github.com/mmadia/bezilla"
-SOURCE_URI="git+https://github.com/mmadia/bezilla.git#686254"
-REVISION="1"
-LICENSE="MPL v1.1"
 COPYRIGHT="1995-2009 Mozilla Developers and Contributors"
+LICENSE="MPL v1.1"
+REVISION="2"
+gitSrcRevision="6862542e81d0407f8fbdb8cf96a48fdc3156e902"
+SOURCE_URI="$HOMEPAGE/archive/$gitSrcRevision.tar.gz"
+CHECKSUM_SHA256="ecdb6e15bd3a48e4b3e036e7fe3c12341c36ddef621403c558a15441cf364d79"
+SOURCE_DIR="bezilla-$gitSrcRevision"
+PATCHES="mailnews-$portVersion.patchset"
 
-ARCHITECTURES="x86_gcc2 x86 ?arm"
-SECONDARY_ARCHITECTURES="x86_gcc2 x86 ?arm"
+ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
-	mailnews$secondaryArchSuffix = $portVersion
-	cmd:thunderbird$secondaryArchSuffix = $portVersion
-	app:MailNews$secondaryArchSuffix = $portVersion
+	mailnews = $portVersion
+	cmd:thunderbird = $portVersion
+	app:MailNews = $portVersion
 	"
-
 REQUIRES="
-	haiku$secondaryArchSuffix
+	haiku
+	lib:libglib_2.0
+	lib:libIDL_2
+	lib:libintl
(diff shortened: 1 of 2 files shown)


diversys:
My recent commits to BeZilla recipe should be copied here as the code base is the same.
https://github.com/haikuports/haikuports/search?q=bezilla&type=Commits&utf8=%E2%9C%93

humdingerb:
Thanks @diversys, that's bringing it quite a bit further along. I updated the PR, but I'm not good at this... :)

diversys:
Looks good. Does it built now on x86_gcc2? BeZilla (firefox-bin binary) fails to link on x86_64.

humdingerb:
Unfortunately, no. See the changed commit message (I shouldn't have that there...):

> Still doesn't build, stops with:
> mv: cannot stat '/packages/mailnews-2.0.0.25-2/.self/lib/thunderbird-2.0.0.25pre': No such file or directory
> Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 1
> Error: Build has failed - stopping.

diversys:
This is caused by this line in INSTALL() section
`mv $libDir/thunderbird-2.0.0.25pre $appsDir/MailNews`
https://github.com/haikuports/haikuports/pull/2146/files#diff-cbf7e92213cba377798b329993915941L70

Are you sure $libDir/thunderbird-2.0.0.25pre gets created by make install?
Check work-*/packaging folder.

humdingerb:
> Are you sure that $libDir/thunderbird-2.0.0.25pre folder gets created by make install command?

I am not. That's from the original recipe...

> Check work-*/packaging folder.

In there is only "packaging/mailnews/settings/" and  "packaging/mailnews_devel/settings/"

extrowerk:
If it is empty, then the $prefix is wrong ir the build systembl expects it differently.

humdingerb:
> Does it prints during INSTALL() where it puts the files?

As I commented above:

> mv: cannot stat '/packages/mailnews-2.0.0.25-2/.self/lib/thunderbird-2.0.0.25pre': No such file or directory
> Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 1
> Error: Build has failed - stopping.

As I clearly don't know what I'm doing, I offer this PR for adoption. I only created the PR in the first place, because I thought someone else wouldn't have to start from scratch again...
Feel free to close.

extrowerk:
@humdingerb : I meant something what `make install` prints, like:
`/bin/install -c -m 644 something`

As i know no other way to debug it post-mortem, if it prints the path where the stuff gots installed, that would help a lot. If you already deleted the work folder, one have to recompile it. But i hope you didn't did that.

But it seems to be a simple linuxy hardcoded install-prefix, let me explain, why i think that:


1) The "mv" command fails, but the mv command is AFTER the `make  install`, so it had to suceed, so something got installed somewhere.
2) You told, the packaging folder is empty except the settings folders (it generated by haikuporter).
3) A misconfigured program can install his files to a `/some/strange/linux/path`. The install process will finish happily without any error (IDK why, maybe a limitation in chroot), but the packaging folder will be empty, as the files got installed into a different prefix.

So the problem is simple `prefix-misconfiguration` i suppose.
The configure script in your recipe using `--prefix=$prefix`.
So the question is: if that not the right way, then how one supposed to define the install prefix?

fbrosson:
When I need to investigate on what `make install` does, I insert a  `false` right after it, to make `INSTALL()` fail.
```diff
 INSTALL()
 {
 	make install
+	false
```
Then I go to `work-*/packaging/mailnews/` and see what is there.
BTW, you can also move the `false` line to the end of `INSTALL()` and then check that you have in `work-*/packaging/mailnews_devel/` what you wanted.

pulkomandy:
you can also use haikuporter -E to enter the chroot and investigate from there.

fbrosson:
Yep :)
But there is a downside with  `-E`: you generally have very few commands available (as we know, most commands which are not in BUILD_PREREQUIRES are not available in the chroot.) 
So I only use `-E` if I have to investigate issues related to `$portPackageLinksDir`.

humdingerb:
Updated the PR. Package builds now!  Stupid mistake in the ./configure line:
`--enable-application=mail` instead of `--enable-application=browser`... Duh!

Thanks for all the feedback though!

PR #2148, opened 2018-01-23
New recipes and updates

khallebal:
I hope everything is alright this time? please let review.


fbrosson:
Sorry, when I wrote that it would be difficult to split the PR, I was refering to the the previous context, where you were using your master branch.

It would be better, imo, to have one PR per recipe. Well, let's see if this is what @korli was suggesting when he wrote that it would be better to split the PR.

fbrosson:
You can keep this one open and use it for one of your recipes.

khallebal:
Ok i'll stop making changes for now, i'll resume the work tommrow.

PR #2150, opened 2018-01-24
dev-libs/xapian: bump to 1.2.25, add debuginfo package.

fbrosson:
* Drop patch as it was applied upstream.
* Add x86_64, marked as untested.
* Add `TEST()`.
* Mark x86 as untested because `TEST()` KDLs a lot.
* `DESCRIPTION`: fix typo and rewrap.
* `HOMEPAGE`: switch to https.
* `BUILD_PREREQUIRES`: sort.

Notes: Version 1.2.25 builds fine but is only the latest oldstable.
The latest stable release, 1.4.5, does not build on any arch.


fbrosson:
Shall I add the recipe for 1.4.5 (with all architectures disabled, of course) in case someone would like to investivate why the latest stable does not build?

PR #2157, opened 2018-01-25
curl: bump to 7.58.0, fix TEST, mark as tested.

fbrosson:


fbrosson:
99% of the tests are OK on all architectures I could test: x86_gcc2, x86 (secondary), and x86_64.

Shall I mark `arm` as tested and remove `curl-7.54.1.recipe` ?

korli:
Should be OK, but beware that x86_gcc2 still uses an old version: https://git.haiku-os.org/haiku/tree/build/jam/repositories/HaikuPorts/x86_gcc2#n30
It should be updated from time to time, or when binary compatibility is broken. The Haiku buildbot continues to build against the old curl devel package... This can lead to problems eventually.
Upload of packages through the git repo, is now probably broken, because of the Gerrit switch. Let's see what the team has foreseen for that.

fbrosson:
OK, thanks @korli.
Let's keep 7.54.1 then, until [x86_gcc2](https://github.com/haiku/haiku/blob/master/build/jam/repositories/HaikuPorts/x86_gcc2#L30) gets updated.
BTW, the only test that curl 7.58.0 faills (on x86_gcc2 and x86 (2nd arch)) is test**116** ([`tests/data/test116`](https://github.com/curl/curl/tree/master/tests/data/test116)), but it happens to be the same test that is already failing on x86 2nd arch with curl 7.54.1.
On x86_gcc2 (primary arch), however, 7.54.1 succeeds on 731 tests only, out of 847. So I guess it would be a good idea to make the bump on the default image. (I've fixed TEST locally to test 7.54.1.)

On **x86_64**, test**116** is also failing. But there is another test which is also failing on that arch: test**1061** ([`tests/data/test1061`](https://github.com/curl/curl/tree/master/tests/data/test1061)). Actually, this test seems to hang forever.

And there also seems to be a few other tests which crash on x86_64 only:
* [`tests/data/test31`](https://github.com/curl/curl/tree/master/tests/data/test1031)
* [`tests/data/test46`](https://github.com/curl/curl/tree/master/tests/data/test46)
* [`tests/data/test61`](https://github.com/curl/curl/tree/master/tests/data/test61)
* [`tests/data/test1415`](https://github.com/curl/curl/tree/master/tests/data/test1415)

But, when we click "Terminate", these tests continue and their result are OK.

I'll try to check if these tests were already producing the same effects with 7.54.1 on x86_64. So if there are no regressions I guess it will be OK to merge (and, later, to update the default images so that they all include 7.58.0.)

fbrosson:
I've finished testing 7.57.0 and 7.54.1 on x86_64. It turns out that there are not even a single regression in **7.58.0**, so we can safely mark that release as tested on all architectures :)

BTW, to the above list of tests that crash with 7.58.0 on x86_64, there's one more test, [`tests/data/test506`](https://github.com/curl/curl/tree/master/tests/data/test506), which crashed 7.54.1 on x86_64.

Ver | Arch | _Total_ | _#(OK)_ | _#(KO)_ | Tests failing
--- | --- | --- | --- | --- | ---
7.54.1 | x86_gcc2 | _847_ | _731_ | _116_ | 3 9 10 31 33 35 38 39 44 46 58 61 71 73 88 92 95 97 99 110 **116** 135 141 154 156 158 163 166 171 173 174 175 176 177 180 181 186 188 189 194 208 245 246 256 258 259 268 277 281 285 286 506 508 513 515 542 544 545 551 552 553 554 568 572 587 805 912 913 1001 1002 1007 1011 1012 1015 1030 1036 1040 1041 1042 1043 1044 1051 1052 1053 1054 1055 1057 1064 1065 1070 1071 1075 1076 1105 1129 1130 1131 1133 1284 1285 1315 1325 1332 1402 1404 1406 1411 1415 1434 1517 1525 1526 1527 1531 1533 1541
7.54.1 | x86 2nd | _830_ | _829_ | _1_ | **116**
7.57.0 | x86_gcc2 | _908_ | _907_ | _1_ | **116**
**7.58.0** | x86_gcc2 | _917_ | _916_ | _1_ | **116**
**7.58.0** | x86 2nd | _900_ | _899_ | _1_ | **116**
**7.58.0** | x86_64 | _917_ | _915_ | _2_ | **116** **1061**

Issue #2135, opened 2018-01-18
[META] Documentation for various variables and new HaikuPorts practices


alaviss:
As the GCI ended, I realize there were many undocumented practices at HaikuPorts, namely:

- [ ] `libVersion`, `libVersionCompat` for library version tracking
- [ ] `commandBinDir` and friends to strip `secondaryArchSuffix` from binaries for secondary architectures
- [ ] Add `~<vcs>` to ports that doesn't use an official release but instead uses a newer version (but not tagged) from \<vcs\>
- [ ] Whether to ship debuginfo with binaries
- [ ] Styling for `HOMEPAGE` variable

There were undocumented input variables:
- [ ] `cmakeDirArgs`, a convinience variable for cmake ports that uses GNUInstallDirs or simillar modules (#2074)
- [ ] Various variables that relates to `SECONDARY_ARCHITECTURES`, namely: `secondaryArchSuffix`, `effectiveTargetArchitecture` (#1854)

Also, there's a lack of guidelines for packages of various languages:
- [ ] python setuptools-based packages, currently for packaging we use constructs like [this][0]
- [ ] go-based packages, currently we've one pending (#2079)

[0]: https://github.com/haikuports/haikuports/blob/3b5a0e2373b350b51794b8116d5cca191f3cf060/dev-python/markdown/markdown-2.6.9.recipe#L29

pulkomandy:
On Thu, Jan 18, 2018 at 01:56:30AM +0000, alaviss wrote:
> As the GCI ended, I realize there were many undocumented practices at HaikuPorts, namely:

We have documented only things related to the "core" of haikuporter.
Some of the things you list should be added there:

> - [ ] Add `~<vcs>` to ports that doesn't use an official release but instead uses a newer version (but not tagged) from \<vcs\>

I think this is already somewhat documented in the version comparison
spec (but not in the haikuporter wiki, you have to dig into the HPKG
file format spec to see how version compare works). Basically, if you
add a .something to a recipe, it is considered a newer version
(so 1.4.2 > 1.4). But if you add a ~something, it is considered a
pre-release (so 1.4~beta < 1.4)

Using ~git is actually not a very good idea, because then you can't
change the version if you point to a different git commit. We should do
it like debian, that is 1.4~20180118 or similar, to be completely safe.

> 
> - [ ] `libVersion`, `libVersionCompat` for library version tracking
[...]

PR #2163, opened 2018-01-28
Colm+ragel

extrowerk:


dev-util/colm/colm-0.13.0.5.recipe
```diff
@@ -0,0 +1,79 @@
...
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandSuffix=$secondaryArchSuffix
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
```
alaviss: What do you think about this variation? This one will disable the suffix whenever it couldn't be built on the main architecture.

```sh
commandBinDir=$binDir
if [ -z "${ARCHITECTURES##*!$targetArchitecture*}" ]; then
	commandSuffix=
	commandBinDir=$prefix/bin
fi

File: dev-util/colm/colm-0.13.0.5.recipe [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+	"
+
+defineDebugInfoPackage colm$secondaryArchSuffix \
+	$commandBinDir/colm \
+	$libDir/libcolm-$portVersion.so
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	autoreconf -i
```
alaviss: `autoreconf` will call `libtoolize` automatically

```diff
-	libtoolize --force --copy --install

File: dev-util/colm/colm-0.13.0.5.recipe [hunk 3]
```diff
@@ -0,0 +1,79 @@
...
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	colm$secondaryArchSuffix = $portVersion
+	cmd:colm$commandSuffix = $portVersion compat >= 1.13
+	lib:libcolm_$portVersion$secondaryArchSuffix = $portVersion compat >= 1.13
```
alaviss: I think `compat` version for the library quite useless since the version is already in it's name. Also, the lib doesn't seem to provide any kind of ABI stability

```diff
+binVersionCompat="$portVersion compat >= ${portVersion%.*.*}"
 
 PROVIDES="
-	cmd:colm$commandSuffix = $portVersion compat >= 1.13
-	lib:libcolm_$portVersion$secondaryArchSuffix = $portVersion compat >= 1.13
+	cmd:colm$commandSuffix = $binVersionCompat
+	lib:libcolm_$portVersion$secondaryArchSuffix = $portVersion

File: dev-util/colm/colm-0.13.0.5.recipe [hunk 4]
```diff
@@ -0,0 +1,79 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	colm${secondaryArchSuffix}_devel = $portVersion
+	devel:libcolm$secondaryArchSuffix = $portVersion compat >= 1.13
```
alaviss:
```diff
-	devel:libcolm$secondaryArchSuffix = $portVersion compat >= 1.13
+	devel:libcolm$secondaryArchSuffix = $portVersion

File: dev-util/colm/colm-0.13.0.5.recipe [hunk 5]
```diff
@@ -0,0 +1,79 @@
+SUMMARY="COmputer Language Manipulation"
+DESCRIPTION="Colm is a programming language designed for the analysis and \
+transformation of computer languages. Colm is influenced primarily by TXL. It \
+is in the family of program transformation languages."
+HOMEPAGE="https://www.colm.net/open-source/colm/"
```
alaviss: `HOMEPAGE` seems to lack an unified style. There are recipes with `/` at the end, and there are also recipe without it at the end (#2135)

dev-util/ragel/ragel-7.0.0.10.recipe
```diff
@@ -0,0 +1,88 @@
...
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandSuffix=$secondaryArchSuffix
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
```
alaviss: Suggested variation

```sh
commandBinDir=$binDir
if [ -z "${ARCHITECTURES##*!$targetArchitecture*}" ]; then
	commandSuffix=
	commandBinDir=$prefix/bin
fi

File: dev-util/ragel/ragel-7.0.0.10.recipe [hunk 7]
```diff
@@ -0,0 +1,88 @@
...
+	"
+
+defineDebugInfoPackage ragel$secondaryArchSuffix \
+	$commandBinDir/ragel \
+	$libDir/libfsm.so.0.0.0
+
+BUILD()
+{
+	libtoolize --force --copy --install
+	autoreconf -i
```
alaviss: `autoreconf` will call `libtoolize` automatically

```diff
-	libtoolize --force --copy --install

File: dev-util/ragel/ragel-7.0.0.10.recipe [hunk 8]
```diff
@@ -0,0 +1,88 @@
...
+fi
+
+PROVIDES="
+	ragel$secondaryArchSuffix = $portVersion
+	cmd:ragel$commandSuffix = $portVersion compat >= 7.0
+	lib:libfsm$secondaryArchSuffix = $portVersion compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	colm$secondaryArchSuffix
```
alaviss: You should be more specific. Does this port depend on `colm` as a whole or only `lib:libcolm`?

dev-util/ragel/ragel-7.0.0.10.recipe
```diff
@@ -0,0 +1,88 @@
...
+SECONDARY_ARCHITECTURES="?x86"
+
+if [ "$targetArchitecture" != x86_gcc2 ]; then
+	commandSuffix=$secondaryArchSuffix
+	commandBinDir=$binDir
+else
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
```
alaviss:
```diff
+binVersionCompat="$portVersion compat >= ${portVersion%.*.*}"
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```

dev-util/ragel/ragel-7.0.0.10.recipe
```diff
@@ -0,0 +1,88 @@
...
+PROVIDES="
+	ragel$secondaryArchSuffix = $portVersion
+	cmd:ragel$commandSuffix = $portVersion compat >= 7.0
+	lib:libfsm$secondaryArchSuffix = $portVersion compat >= 0
```
alaviss:
```diff
-	cmd:ragel$commandSuffix = $portVersion compat >= 7.0
+	cmd:ragel$commandSuffix = $binVersionCompat
-	lib:libfsm$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libfsm$secondaryArchSuffix = $libVersionCompat
```

dev-util/ragel/ragel-7.0.0.10.recipe
```diff
@@ -0,0 +1,88 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	colm$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	ragel${secondaryArchSuffix}_devel = $portVersion
+	devel:libfsm$secondaryArchSuffix = $portVersion compat >= 0
```
alaviss:
```diff
-	devel:libfsm$secondaryArchSuffix = $portVersion compat >= 0
+	devel:libfsm$secondaryArchSuffix = $libVersionCompat

File: dev-util/ragel/ragel-7.0.0.10.recipe [hunk 12]
```diff
@@ -0,0 +1,88 @@
...
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	"
+
+defineDebugInfoPackage ragel$secondaryArchSuffix \
+	$commandBinDir/ragel \
+	$libDir/libfsm.so.0.0.0
```
alaviss:
```diff
-	$libDir/libfsm.so.0.0.0
+	$libDir/libfsm.so.$libVersion

File: dev-util/colm/colm-0.13.0.5.recipe (hunk 5 above)
Comment (fbrosson) [reply dev-util/colm/colm-0.13.0.5.recipe]: Here the / is perfect :)
The rule is that if you pass an URI to `curl --head` which is missing the trailing slash, then, if the server redirects you to the "corrected" URI, i.e. the same URI but with a slash appended at the end of the line, then this means we should have the trailing slash for the URI in `HOMEPAGE`.

Sometimes the URI does not need the slash, but the server won't complain if you add one.
That's the case for e.g. `https://github.com/haikuports/haikuports`. But since GitHub does not send a recirect for it, it is better to not add the slash in URIs for any project hosted on GitHub.

File: dev-util/colm/colm-0.13.0.5.recipe (hunk 1 above)
Comment (fbrosson) [reply dev-util/colm/colm-0.13.0.5.recipe]: Hmm, it's not very eye-friendly.
What I recommend is the more traditional style:
```
commandBinDir=$binDir
then
	commandBinDir=$prefix/bin
	commandSuffix=
fi
```

File: dev-util/ragel/ragel-7.0.0.10.recipe (hunk 6 above)
Comment (fbrosson) [reply dev-util/ragel/ragel-7.0.0.10.recipe]: What about this variation?
```
commandBinDir=$binDir
then
	commandBinDir=$prefix/bin
	commandSuffix=
fi
```
Remark: the commandBinDir "trick" should **_only_** be used **_if_** `ARCHITECTURES` contains `!x86_gcc2`.

File: dev-util/ragel/ragel-7.0.0.10.recipe (hunk 8 above)
Comment (extrowerk) [reply dev-util/ragel/ragel-7.0.0.10.recipe]: Just lib:libcolm

File: dev-util/colm/colm-0.13.0.5.recipe (hunk 1 above)
Comment (alaviss) [reply dev-util/colm/colm-0.13.0.5.recipe]: Given that probably only `x86_gcc2` would need this, I think it's fine to only use this for said architecture.

BTW, should we add this to HaikuPorter itself? Since we don't want those suffixes for binaries when a port won't build for the main architecture anyway.

Comment (fbrosson) [reply dev-util/colm/colm-0.13.0.5.recipe]: 1) +1, because x86_gcc2 is, indeed, the only arch for which we use `commandBinDir`.
(We would have been able to use it for old apps which only build gcc2, if we had x86_gcc2 as second arch for x86_64 primary arch, but we don't build images with that combination.)

2) I'd say it is better to leave that outside of HaikuPorter. This gives more flexibility, in particular if we want to allow users of x86_gcc2 hybrid to have the `/system/bin/foobar` from the primary arch package and the `/system/bin/x86/foobar` from the secondary arch. We generally choose to drop the latter when the former works fine, but we could have situations where we're not sure the runtime built with gcc2 works perfectly.

Comment (pulkomandy) [reply dev-util/colm/colm-0.13.0.5.recipe]: In fact there are special cases already. For example:
 * ffmpeg builds with both archs but we ship the binaries from the x86 version, because they work better.
 * For gcc (and most tools in the build toolchain), we do want to ship both versions.

If there was an automated solution to this, we would not have introduced an explicit $secondaryArchSuffix at all (at least for binaries).

File: dev-util/ragel/ragel-7.0.0.10.recipe (hunk 6 above)
Comment (fbrosson) [reply dev-util/ragel/ragel-7.0.0.10.recipe]: > Remark: the commandBinDir "trick" should **_only_** be used **_if_** `ARCHITECTURES` contains `!x86_gcc2`.

Hmm, the truth is that we **_could_** use it **_even_** if  `ARCHITECTURES` contains `x86_gcc2` or `?x86_gcc2`, but we would need to add **_this_** after the `REQUIRES`:
```
CONFLICTS="
	ragel
	"
fi
```
Important: In the above snippet it's really `ragel` that we need to have in `CONFLICTS`. Because putting `ragel$secondaryArchSuffix` would make ragel_x86 conflict with itself ;-)
(The purpose is to make the secondary arch package conflict with the primary arch package.)

Comment (pulkomandy) [reply dev-util/ragel/ragel-7.0.0.10.recipe]: Well, that is getting into nonsense territory. The reason we introduced the $secondaryArchSuffix is precisely to not have this kind of conflicts and allow packages to be installed side by side. Why would one insist so hard on removing that feature?

Comment (fbrosson) [reply dev-util/ragel/ragel-7.0.0.10.recipe]: I'm not saying we should do so. I'm just saying that, in theory, we could do that. But I do not recommend using the commandBinDir "trick" for recipes which are not broken on a primary arch.

File: dev-util/colm/colm-0.13.0.5.recipe [hunk 13]
```diff
@@ -0,0 +1,77 @@
...
+COPYRIGHT="2006-2016 Adrian Thurston"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://www.colm.net/files/colm/colm-$portVersion.tar.gz"
+CHECKSUM_SHA256="33e624677176958eaad76ebe6c391a68a0b4728fec8cc039efa1316f525f408c"
+
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
```
fbrosson: @extrowerk, could you have a look at the sources and see if colm can be built with gcc2 ?
If it cannot, then you should mark x86_gcc2 as broken:
```diff
-ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```

BTW, I just did some metrics, and it turns out that there are many more recipes using

```
commandBinDir=$binDir
	commandSuffix=
	commandBinDir=$prefix/bin
fi
```
than recipes using
```
commandBinDir=$binDir
	commandBinDir=$prefix/bin
	commandSuffix=
fi
```
So if you swap the order of `commandSuffix` and `commandBinDir` (to have `commandSuffix` before `commandBinDir`) it would be cool :)
Thanks!

BTW, if colm can be built with gcc2 then you should drop the bloc with `commandSuffix` and `commandBinDir`.

korli:
Please consider fixing: http://vmpkg.haiku-os.org/master/x86_64/buildruns/966/builds/9996.log

PR #2164, opened 2018-01-29
sys-block/parted: add debuginfo package, mark as untested.

fbrosson:
* Mark as untested on all architectures, as parted does not seem to be working correctly on Haiku at the moment.
* Switch `SOURCE_URI` to `ftpmirror.gnu.org`, using https.
* Drop libtool files as well as the unneeded `$infoDir/dir`.

### DIFF
--- a/sys-block/parted/parted-3.2.recipe
+++ b/sys-block/parted/parted-3.2.recipe
@@ -5,13 +5,13 @@ data on hard disks and disk imaging."
 HOMEPAGE="https://www.gnu.org/software/parted/"
 COPYRIGHT="1999-2014 Free Software Foundation Inc."
 LICENSE="GNU GPL v3"
-REVISION="2"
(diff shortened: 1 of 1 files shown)


sys-block/parted/parted-3.2.recipe
```diff
@@ -21,12 +21,21 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
 	commandBinDir=$prefix/bin
 fi
 
+libpartedVersion="2.0.1"
+libparted_fs_resizeVersion="0.0.1"
```
alaviss: How about using shorter names?

```diff
-libpartedVersion="2.0.1"
-libparted_fs_resizeVersion="0.0.1"
+libVersion="2.0.1"
+libFsResizeVersion="0.0.1"

Issue #2167, opened 2018-01-31
Update libedit to current 2017-03-29-3.1 from 2015-03-21-3.1


Vidrep:

korli:
As this is a system package, change with care.

fbrosson:
➡️ #2168

PR #2168, opened 2018-02-01
libedit: bump to 20170329-3.1, add debuginfo package.

fbrosson:
Also add a tiny gcc2 patch and drop docs on secondary arch.

Remark: `--enable-widec` is now enabled by default in configure.

### DIFF
--- a/sys-libs/libedit/libedit-20170329_3.1.recipe
+++ b/sys-libs/libedit/libedit-20170329_3.1.recipe
@@ -3,21 +3,24 @@ DESCRIPTION="This is an autotool- and libtoolized port of the NetBSD Editline li
 (libedit). This Berkeley-style licensed command line editor library provides \
 generic line editing, history, and tokenization functions, similar to those \
 found in GNU Readline."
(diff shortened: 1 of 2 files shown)


fbrosson:
Thanks @alaviss.

FWIW, I'm using this build of libedit since before I opened this PR and it seems to be working fine.

@korli, may I add you as a reviewer? I read [your comment](https://github.com/haikuports/haikuports/issues/2167#issuecomment-361924754) about this package, so I'd like to have your green lights before merging. Thanks!

BTW, we currently have 2 recipes in master: `libedit-2014_10_30_3.1.recipe and libedit-2015_03_21_3.1.recipe`.
This PR drops the former, keeps the latter (unmodified), and adds `libedit-20170329_3.1.recipe`.

Remark: The small change in the naming convention is handy because it follows upstream's naming and, consequently, allows to use `$portVersion` instead of hardcoded versioning.

waddlesplash:
Sounds fine anyway. Merging.

PR #2169, opened 2018-02-03
gptfdisk: bump to 1.0.3, make TEST() succeed on x86_64.

fbrosson:


fbrosson:
gptfdisk seems to also be working fine on x86_gcc2 but for some unknown reason when gdisk's stdout is not a terminal the output is truncated just before the `Partition table scan:` line.
(This only happens on x86_gcc2.)

Anyway, this issue does not seem to have any impact on the runtime when it is ran interactively or when it's output is not redirected.

Shall I add support for secondary architectures and mark x86_gcc2 as not tested?

PR #2178, opened 2018-02-03
Boost: version boost

extrowerk:


dev-libs/boost/boost166-1.66.0.recipe
```diff
@@ -0,0 +1,180 @@
...
+		--enable-parallel-mark \
+		inlining=on \
+		threading=multi \
+		variant=release \
+		link=static,shared \
+		runtime-link=shared \
+		--without-python \
+		install
+
+	prepareInstalledDevelLibs libboost_atomic libboost_chrono \
```
fbrosson: Given the huge number of libs, you might wish to replace lines 159 to 172 by these 2 lines:
```
	prepareInstalledDevelLibs `echo "$PROVIDES" | sed -n \
		-e "s/lib:\(.*\)$secondaryArchSuffix =.*/\1/p"`
```

extrowerk: Nope, thats incomplete. Tested with PROVIDES_Devel, but that also not the complete list.

One have to merge PROVIDES and PROVIDES_devel, eliminate the duplicates and sort it in abc. Then it should work.

fbrosson: Good catch. I had not noticed that the `PROVIDES` and `PROVIDES_devel` did not match exactly.

It seems there are 31 `lib:lib*` in `PROVIDES` and 33 `devel:lib*` in `PROVIDES_devel`. Well, since the call to `prepareInstalledDevelLibs` mentions these 33 libs, this variant should work:
```
	prepareInstalledDevelLibs `echo "$PROVIDES_devel" | sed -n \
		-e "s/devel:\(.*\)$secondaryArchSuffix =.*/\1/p"`
```

FWIW, the 2 "extra" libs that are only in `PROVIDES_devel` are `libboost_exception` and `libboost_test_exec_monitor`.

extrowerk: Thanks it worked.

PR #2180, opened 2018-02-04
Benchmark : version bump

DethroTull:
Previous recipe was for version 1.2.0, and according to https://repology.org/metapackages/?inrepo=haikuports_master&outdated=True, benchmark was outdated so I wanted to update in order to get my feet wet with the haikuports process.


Begasus:
You should rename the topic to include the name of the project "benchmark" in this case ;)

dev-libs/benchmark/benchmark-1.3.0.recipe
```diff
@@ -0,0 +1,70 @@
...
+INSTALL()
+{
+	cd build
+	make install
+
+	mkdir -p $includeDir
+	mv $prefix/include/* $includeDir/
+	rm -rf $prefix/include/
+
+	prepareInstalledDevelLibs libbenchmark
```
fbrosson:
```diff
-	prepareInstalledDevelLibs libbenchmark
+	prepareInstalledDevelLib libbenchmark
```
(prepareInstalledDevelLib**s** is required when we have more than one lib. 
Here, prepareInstalledDevel**Lib** is enough and will do the same work :)

dev-libs/benchmark/benchmark-1.3.0.recipe
```diff
@@ -0,0 +1,70 @@
+SUMMARY="A microbenchmark support library"
+DESCRIPTION="A library to support the benchmarking of functions, similar to unit-tests."
+HOMEPAGE="https://github.com/google/benchmark/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/google/benchmark/"
+HOMEPAGE="https://github.com/google/benchmark"
```

dev-libs/benchmark/benchmark-1.3.0.recipe
DethroTull: Very good to know!  Thank you for the explanation.

DethroTull: Ok, I've obviously done something incorrect here.  I'm going to remove this pull request and try again.

PR #2184, opened 2018-02-04
Glib2: bump

extrowerk:


dev-libs/glib/glib2-2.54.3.recipe
```diff
@@ -41,11 +40,11 @@ PROVIDES="
 	cmd:glib_compile_schemas$secondaryArchSuffix
 	cmd:gresource$secondaryArchSuffix
 	cmd:gsettings$secondaryArchSuffix
-	lib:libgio_2.0$secondaryArchSuffix = 0.5400.2 compat >= 0
-	lib:libglib_2.0$secondaryArchSuffix = 0.5400.2 compat >= 0
-	lib:libgmodule_2.0$secondaryArchSuffix = 0.5400.2 compat >= 0
-	lib:libgobject_2.0$secondaryArchSuffix = 0.5400.2 compat >= 0
-	lib:libgthread_2.0$secondaryArchSuffix = 0.5400.2 compat >= 0
+	lib:libgio_2.0$secondaryArchSuffix = 0.5400.3 compat >= 0
```
fbrosson: Shall we add a `libVersion` and its `libVersionCompat`?
```diff
 ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="?x86"
+
+libVersion="0.5400.3"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	glib2$secondaryArchSuffix = $portVersion compat >= 2
-	lib:libgio_2.0$secondaryArchSuffix = 0.5400.3 compat >= 0
+	lib:libgio_2.0$secondaryArchSuffix = $libVersionCompat
```
(Of course, the purpose of this change is to replace all other occurences of `0.5400.3 compat >= 0` by `$libVersionCompat`.)

dev-libs/glib/glib2-2.54.3.recipe
```diff
@@ -65,48 +64,62 @@ PROVIDES_devel="
...
+	$binDir/gio-querymodules \
+	$binDir/glib-compile-resources \
+	$binDir/glib-compile-schemas \
+	$binDir/gresource \
+	$binDir/gsettings \
+	$libDir/libgio-2.0.so.0 \
+	$libDir/libglib-2.0.so.0 \
+	$libDir/libgmodule-2.0.so.0 \
+	$libDir/libgobject-2.0.so.0 \
+	$libDir/libgthread-2.0.so.0
```
fbrosson: If I'm not mistaken, it would be better to pass the names of the real files, not the symlinks.
```diff
-	$libDir/libgio-2.0.so.0 \
+	$libDir/libgio-2.0.so.$libVersion \
-	$libDir/libglib-2.0.so.0 \
+	$libDir/libglib-2.0.so.$libVersion \
-	$libDir/libgmodule-2.0.so.0 \
+	$libDir/libgmodule-2.0.so.$libVersion \
-	$libDir/libgobject-2.0.so.0 \
+	$libDir/libgobject-2.0.so.$libVersion \
-	$libDir/libgthread-2.0.so.0
+	$libDir/libgthread-2.0.so.$libVersion
```

dev-libs/glib/glib2-2.54.3.recipe
```diff
@@ -65,48 +64,62 @@ PROVIDES_devel="
...
 	glib2$secondaryArchSuffix == $portVersion base
-	devel:libffi$secondaryArchSuffix
-	devel:libiconv$secondaryArchSuffix
-	devel:libintl$secondaryArchSuffix
-	devel:libpcre$secondaryArchSuffix
-	devel:libz$secondaryArchSuffix
-	cmd:perl
 	haiku$secondaryArchSuffix
-	lib:libffi$secondaryArchSuffix
+	cmd:perl
```
fbrosson: I think it would be better to move `cmd:perl` to `BUILD_PREREQUIRES`.
(But I'm not sure about that, as I might be missing something.)

Hmm, I also think you don't need to mention any `lib:`in `REQUIRES_devel`, since these are provided by devel packages which already depend on their respective base packages.
In other words, if you put e.g. `devel:libffi$secondaryArchSuffix` in a `REQUIRES_devel` then you don't need to also put `lib:libffi$secondaryArchSuffix` since the former automatically pulls the latter.

extrowerk: Tested already before i sent my PR, not worked, hp complains about non-declaredd dependencies, so i had to add them back. (The earlier recipe sports them too)

fbrosson: Hmm, you are right. `lib:libintl` needs to be added explicitly. That's because it is provided by the `gettext_libintl` package, while `devel:libintl` is provided by the `gettext` package.

But I think you might be able to get rid of all the other `lib:` lines in your `REQUIRES_devel` and `BUILD_REQUIRES`.

dev-libs/glib/glib2-2.54.3.recipe
```diff
@@ -23,12 +22,15 @@ COPYRIGHT="1995-1997  Peter Mattis, Spencer Kimball and Josh MacDonald
...
 
 ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="?x86_gcc2 ?x86"
+SECONDARY_ARCHITECTURES="?x86"
+
+libVersion="0.5400.3"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
 	glib2$secondaryArchSuffix = $portVersion compat >= 2
```
alaviss: You could drop `compat >= 2`, as it's already part of the package name

```diff
- 	glib2$secondaryArchSuffix = $portVersion compat >= 2
+ 	glib2$secondaryArchSuffix = $portVersion

PR #2192, opened 2018-02-05
Chocolate-Doom : version bump

DethroTull:
Updated chocolate-doom to most recent version.  I updated the libSDL entries from SDL1 to SDL2 respectively as configure would fail otherwise.  It compiled and installed properly.  I was able to play the x86_64 version.


games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,76 @@
+SUMMARY="Doom source port that accurately reproduces the experience of Doom"
+DESCRIPTION="Chocolate Doom is a source port of the game Doom, by id Software. \
+The purpose of Chocolate Doom is to be as compatible as possible with the \
+original DOS version of Doom."
+HOMEPAGE="https://www.chocolate-doom.org/"
+COPYRIGHT="Id Software Inc. 1993-1996
+	Simon Howard
+	James Haley
+	Samuel Villarreal"
```
alaviss: This should be updated to conform to the Packaging Policy. Thanks to folks at Debian, [here][0] is the list of copyright holders and their associated years.

```diff
-COPYRIGHT="Id Software Inc. 1993-1996
-	Simon Howard
-	James Haley
-	Samuel Villarreal"
+COPYRIGHT="1993-1996 Id Software, Inc
+	1993-2008 Raven Software
+	1996 Rogue Entertainment / Velocity, Inc.
+	1996-2013 Free Software Foundation, Inc
+	1999 Thomas A. Fine
+	2002-2003 Marcel Telka
+	2002-2010 The DOSBox Team
+	2005-2014 Simon Howard
+	2005-2014 Andrey Budko
+	2006 Ben Ryves
+	2008 GhostlyDeath
+	2008 David Flater
+	2010 James Haley
+	2010 Samuel Villarreal
+	2011 Team Eternity
+	2014 Fabian Greffrath
+	2001-2008 Contributors to the Freedoom project"
```

[0]: http://metadata.ftp-master.debian.org/changelogs/main/c/chocolate-doom/chocolate-doom_3.0.0-3_copyright

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+LICENSE="GNU GPL v2"
```
alaviss: There're [a few more licenses](http://metadata.ftp-master.debian.org/changelogs/main/c/chocolate-doom/chocolate-doom_3.0.0-3_copyright)

```diff
-LICENSE="GNU GPL v2"
+LICENSE="GNU GPL v2
+	BSD (3-clause)"

File: games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe [hunk 3]
```diff
@@ -0,0 +1,76 @@
...
+	cmd:chocolate_doom$secondaryArchSuffix = $portVersion
+	cmd:chocolate_doom_setup$secondaryArchSuffix = $portVersion
+	cmd:chocolate_heretic_setup$secondaryArchSuffix = $portVersion
+	cmd:chocolate_heretic$secondaryArchSuffix = $portVersion
+	cmd:chocolate_hexen_setup$secondaryArchSuffix = $portVersion
+	cmd:chocolate_hexen$secondaryArchSuffix = $portVersion
+	cmd:chocolate_server$secondaryArchSuffix = $portVersion
+	cmd:chocolate_setup$secondaryArchSuffix = $portVersion
+	cmd:chocolate_strife_setup$secondaryArchSuffix = $portVersion
+	cmd:chocolate_strife$secondaryArchSuffix = $portVersion
```
alaviss: Since this port don't work on x86_gcc2, you may want to consider using [`commandBinDir`][0] to strip the `_x86` suffix on x86_gcc2.

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+SOURCE_URI="https://www.chocolate-doom.org/downloads/$portVersion/chocolate-doom-$portVersion.tar.gz"
+CHECKSUM_SHA256="73aea623930c7d18a7a778eea391e1ddfbe90ad1ac40a91b380afca4b0e1dab8"
+SOURCE_DIR="chocolate-doom-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	chocolate_doom$secondaryArchSuffix = $portVersion
+	cmd:chocolate_doom$secondaryArchSuffix = $portVersion
```
alaviss: I believe some of these are GUI applications. If that's the case, you would want to create a resources file and install the binary to `$appsDir`

DethroTull: Should I create a resources file for each executable, seeing that they're all for separate games?

alaviss: You could make a file containing placeholders (ie. `@MAJOR@`, `@DESC@`, `@ICON@`) then fill them in during packaging with `sed`.

alaviss: Also, if they're separate games, you should consider splitting the package into smaller ones, each containing a different version of the game (assuming they don't depend on each other).

DethroTull: I did find other recipe and resource files that did what you said with the placeholders and implemented it in my recipe.

I had a few issues with putting the files into $appDir, however.  First it removed the apps from the global path (which I'm guessing could be fixed with some script to add $appsDir to Path), but also you need to specify a wad file as an argument, which made it not launch when clicking from a deskbar symlink.  

Also, I added the defineDebugInfoPackage info the way the page you linked me told me to, but when I did it also gave me an error once it was done compiling saying it couldn't find the '/chocolate[-,_]doom' file.  If I removed that section, it compiled and installed. 

Also, I noticed the source itself includes a png to use as an icon, but even with manually copying it beforehand to the additional-files dir, adding it to my recipe file, and adding code to add the png to the executable, nothing happened.  I searched all over, but could not find a way to convert the png to a byte string used in the resource file's 'resource vector_icon'.  Icon-O-Matic would not open the file correctly.    

Thank you for the nudges in the right direction.  I've learned more than I could have ever learned to with just your suggestions!

DethroTull: Oh, and from what I saw of the output from the compile, it does not appear they are separate games, but more so each executable runs the iwad files from their respective games (doom, heretic, hexen, strife).  I can't tell you exactly, but each game might be a copy from the main executable.  The difference is in the respective setup files which have options specific to the game it's configuring.   

After creating the resource file for the main chocolate-doom executable, it did not add the resource info to the other executables (as would be expected), but I feel they should all have their own resource files)

fbrosson: I think I know why haikuporter dropped the resources when you had the `defineDebugInfoPackage`. It's (probably) because you are using the latest stable release of haikuporter.

BTW, in case you would like to use `defineDebugInfoPackage` with a package which creates runtimes that have resources, you should use a git revision of haikuporter that includes [this commit](https://github.com/haikuports/haikuporter/commit/c4c7fe9307a7f35ac05a02aa34bcb9b527c03e45).

(Maybe we should tag the current HEAD of HaikuPorter as release "1.2.3".)

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+	cp chocolate-doom \
+		chocolate-doom-setup \
+		chocolate-hexen \
+		chocolate-hexen-setup \
+		chocolate-heretic \
+		chocolate-heretic-setup \
+		chocolate-server \
+		chocolate-setup \
+		chocolate-strife-setup \
+		chocolate-strife $binDir
```
fbrosson: When you add the [commandBinDir](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) bloc, you might wish to change `INSTALL()` to:
```diff
 INSTALL()
 {
 	cd src
-	mkdir -p $binDir
+	install -d $commandBinDir
+	install -t $commandBinDir \
-	cp chocolate-doom \
+		chocolate-doom \
 		chocolate-doom-setup \
 		chocolate-hexen \
 		chocolate-hexen-setup \
 		chocolate-heretic \
 		chocolate-heretic-setup \
 		chocolate-server \
 		chocolate-setup \
 		chocolate-strife-setup \
-		chocolate-strife $binDir
+		chocolate-strife
 }
```

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
fbrosson: Yep, that's a very good idea.
BTW, it might be nice to include in this PR the removal of the old recipe: `chocolate_doom-2.3.0.recipe`.
(If 3.0.0 works fine then we can drop 2.3.0.)

DethroTull:
I updated my haikuporter to use the commit listed in the comments here, and it still does not attach the png image properly.  I am at a loss.  I've left it in the additional-files directory in case someone else is willing to try.  The rest of the resource file is added, however.

mmlr:
Haiku simply doesn't support PNG icons. It supports old style bitmap icons at some scales and the Haiku Vector Icon Format (HVIF) which can be generated with Icon-o-Matic and is what the vicn resources use. To create a vector icon, the PNG would need to be traced to vector paths in some program (Inkskape for example, or autotrace on the command line) and converted to a SVG, cleaned up and imported into I-o-M. This usually produces rather messy output if one isn't used to cleaning the traced bitmaps up, so trying to find an SVG version for the icon directly makes more sense. As for the PNG, as it won't work directly in any casr, I'd prefer it not to be put in at all and an issue for a task to make the icon be created instead.

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,128 @@
...
+	$commandBinDir/chocolate-doom$commandSuffix \
+	$commandBinDir/chocolate-doom-setup$commandSuffix \
+	$commandBinDir/chocolate-heretic-setup$commandSuffix \
+	$commandBinDir/chocolate-heretic$commandSuffix \
+	$commandBinDir/chocolate-hexen-setup$commandSuffix \
+	$commandBinDir/chocolate-hexen$commandSuffix \
+	$commandBinDir/chocolate-server$commandSuffix \
+	$commandBinDir/chocolate-setup$commandSuffix \
+	$commandBinDir/chocolate-strife-setup$commandSuffix \
+	$commandBinDir/chocolate-strife$commandSuffix
```
fbrosson: There is a small **_latent_** error here, which does not get noticed in Haiku x86_gcc2 hybrid (i.e. with x86_gcc2 primary arch with x86 secondary arch), but which **_would_** be noticed in custom hybrid builds of Haiku such as x86_64 primary arch with x86 secondary arch.

```diff
-defineDebugInfoPackage chocolate_doom \
+defineDebugInfoPackage chocolate_doom$secondaryArchSuffix \
-	$commandBinDir/chocolate-doom$commandSuffix \
+	$commandBinDir/chocolate-doom \
-	$commandBinDir/chocolate-doom-setup$commandSuffix \
+	$commandBinDir/chocolate-doom-setup \
-	$commandBinDir/chocolate-heretic-setup$commandSuffix \
+	$commandBinDir/chocolate-heretic-setup \
-	$commandBinDir/chocolate-heretic$commandSuffix \
+	$commandBinDir/chocolate-heretic \
-	$commandBinDir/chocolate-hexen-setup$commandSuffix \
+	$commandBinDir/chocolate-hexen-setup \
-	$commandBinDir/chocolate-hexen$commandSuffix \
+	$commandBinDir/chocolate-hexen \
-	$commandBinDir/chocolate-server$commandSuffix \
+	$commandBinDir/chocolate-server \
-	$commandBinDir/chocolate-setup$commandSuffix \
+	$commandBinDir/chocolate-setup \
-	$commandBinDir/chocolate-strife-setup$commandSuffix \
+	$commandBinDir/chocolate-strife-setup \
-	$commandBinDir/chocolate-strife$commandSuffix
+	$commandBinDir/chocolate-strife
```

Explanation: We should not use:
* `$secondaryArchSuffix` in the lines where `$binDir` appears, nor
* `$commandSuffix` in the lines where `$commandBinDir` appears.

This is because :
1) without the commandBinDir/commandSuffix, the runtimes we have in `$binDir` do not end with any suffix.
2) with the commandBinDir/commandSuffix, the runtimes we have in `$commandBinDir` (i.e. `$prefix/bin`) do not end with any suffix.

fbrosson: BTW, the suffix is unknown to the build system.
What haikuporter does, when we build a recipe with the defaut `$binDir` (i.e. without using a `commandBinDir`) for a secondary arch, say, x86, is that it creates in `$prefix/bin` a symlink for each runtime found in `$binDir` (i.e. `$prefix/bin/x86/`) after the build.
So, if the build creates `$prefix/bin/x86/foobar` then haikuporter will create in `$prefix/bin/` a symlink pointing to `x86/foobar` and will name that symlink `foobar-x86`.
That's why there is no suffix in any runtime found in `$prefix/bin/x86/`.

Suffixes **_only_** exist in `$prefix/bin/` for symlinks created **_by_** haikuporter **_itself_**.
When we use the commandBinDir/commandSuffix "trick" in a recipe which does not build for the x86_gcc2 primary arch, we have no suffixes at all.

But if someone builds a custom hybrid Haiku, with e.g. x86_64 as primary arch and x86 as secondary arch, then haikuporter would accept to build chocolate_doom_x86.
In such an environment, `BUILD()` and `INSTALL()` would both run correctly, but when haikuporter would try to create the debuginfo package, it would not find `$prefix/bin/x86/chocolate-doom_x86` (nor `$prefix/bin/x86/chocolate-doom-x86`).
As a matter of fact, what we would get would be the runtimes (e.g. `chocolate-doom`) in `$prefix/bin/x86/` and the symlinks (e.g. `chocolate-doom-x86 -> x86/chocolate-doom`) in `$prefix/bin/`.

So if you wish to create a debuginfo package for chocolate_doom then `defineDebugInfoPackage` requires:
* `chocolate_doom$secondaryArchSuffix` for its very first parameter,
* `$commandBinDir/chocolate-doom`, `$commandBinDir/chocolate-doom-setup`, and so on, for each of the runtimes created.

DethroTull:
@mmlr that would definitely explain why I couldn't get it to work.  Thank you for the explanation.

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,128 @@
...
+	2014, Fabian Greffrath"
+LICENSE="GNU GPL v2
+	BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://www.chocolate-doom.org/downloads/$portVersion/chocolate-doom-$portVersion.tar.gz"
+CHECKSUM_SHA256="73aea623930c7d18a7a778eea391e1ddfbe90ad1ac40a91b380afca4b0e1dab8"
+SOURCE_DIR="chocolate-doom-$portVersion"
+ADDITIONAL_FILES="
+	chocolate_doom.rdef.in
+	chocolate_doom.png
```
fbrosson:
```diff
-ADDITIONAL_FILES="
-	chocolate_doom.rdef.in
-	chocolate_doom.png
-	"
+ADDITIONAL_FILES="chocolate_doom.rdef.in"
```

games-fps/chocolate-doom/chocolate_doom-3.0.0.recipe
```diff
@@ -0,0 +1,128 @@
+SUMMARY="Doom source port that accurately reproduces the experience of Doom"
+DESCRIPTION="Chocolate Doom is a source port of the game Doom, by id Software. \
+The purpose of Chocolate Doom is to be as compatible as possible with the \
+original DOS version of Doom."
+HOMEPAGE="https://www.chocolate-doom.org/"
+COPYRIGHT="1993-1996, Id Software, Inc
+	1993-2008, Raven Software
+	1996, Rogue Entertainment / Velocity, Inc.
+	1996-2013, Free Software Foundation, Inc
```
fbrosson: You forgot the dot after "Inc" at lines 6 and 9.

BTW, we also prefer to drop the commas after the copyright years:
```diff
-COPYRIGHT="1993-1996, Id Software, Inc
+COPYRIGHT="1993-1996 Id Software, Inc.
-	1993-2008, Raven Software
+	1993-2008 Raven Software
-	1996, Rogue Entertainment / Velocity, Inc.
+	1996 Rogue Entertainment / Velocity, Inc.
-	1996-2013, Free Software Foundation, Inc
+	1996-2013 Free Software Foundation, Inc.
-	1999, Thomas A. Fine
+	1999 Thomas A. Fine
-	2002-2003, Marcel Telka
+	2002-2003 Marcel Telka
-	2002-2010, The DOSBox Team
+	2002-2010 The DOSBox Team
-	2005-2014, Simon Howard
+	2005-2014 Simon Howard
-	2005-2014, Andrey Budko
+	2005-2014 Andrey Budko
-	2006, Ben Ryves
+	2006 Ben Ryves
-	2008, GhostlyDeath
+	2008 GhostlyDeath
-	2008, David Flater
+	2008 David Flater
-	2010, James Haley, Samuel Villarreal
+	2010 James Haley, Samuel Villarreal
-	2011, Team Eternity
+	2011 Team Eternity
-	2014, Fabian Greffrath"
+	2014 Fabian Greffrath"
```

scottmc: What's the status on this one?

Begasus:
@DethroTull are you still looking into this?

Begasus:
As @scottmc already mentioned, there is already a version bump merged for this PR, aside from the rdef file I don't think this will add anything new and could be closed without merging?

Begasus:
Superseded by https://github.com/haikuports/haikuports/pull/3435 ... closing

PR #2200, opened 2018-02-09
mbedtls: bump to 2.7.0.

fbrosson:


net-libs/mbedtls/mbedtls-2.7.0.recipe
```diff
@@ -75,9 +83,9 @@ REQUIRES="
 
 PROVIDES_devel="
 	mbedtls${secondaryArchSuffix}_devel = $portVersion
-	devel:libmbedcrypto$secondaryArchSuffix = 0 compat >= 0
-	devel:libmbedtls$secondaryArchSuffix = 10 compat >= 10
-	devel:libmbedx509$secondaryArchSuffix = 0 compat >= 0
+	devel:libmbedcrypto$secondaryArchSuffix = $libcryptoVersionCompat
+	devel:libmbedtls$secondaryArchSuffix = $libtlsVersionCompat
+	devel:libmbedx509$secondaryArchSuffix = $libx509VersionCompat
```
Begasus: Do we use *Compat* here too? I thought it was only needed for "lib:lib*"

fbrosson: I also thought, **_some time ago_**, that we should only use `compat` for `lib:lib*` lines.
But later I understood that the `compat` can also be used for `devel:lib*` lines, although this is not automatic.
If I understand correctly, we should only avoid the `compat` when we want to make sure that any external recipe depending on a `devel:libmbed*` would use an explicit version of mbedtls's `devel:libmbed*`.

@alaviss, do you know whether or not we should use  compats in this recipe, and where? Thanks!

FWIW, mbedtls does **_not_** use the `.so.x.y.z` notation for its libs. Also, there are no symlinks.

pulkomandy:
> FWIW, mbedtls does **_not_** use the `.so.x.y.z` notation for its libs. Also, there are no symlinks.

Then there should be no version or compat entries for the provides.

But this should be checked carefully with upstream, because it means all
versions of their lib will stay the same and they can never make ABI
changes (well, they can, but it will be annoying for us).

In some cases (eg. lua) we patched build systems ourselves to add a
soname to libraries, so we could have proper provides/requires and allow
ABI changes between versions.

So if the soname (the so.x.y.z) is not present, we should:
1) Contact the upstream author to ask them what their plan is for ABI
   support:
   - They plan to always maintain old ABIs in new versions of the lib (ok)
   - They plan to make a libfoo2.so when they change ABI (also ok)
   - They don't care/know about it (bad)

In the 3rd case,
2) Patch the buildsystem to include a soname in the library and add the
approrpiate symlinks and compat entries.

But in cases a and b:
2) Do not include versions or compat to the lib entries. The lib has no
versionning then.

fbrosson:
These are the sonames we have with the default build for mbedtls 2.7.0 vs 2.6.0:

mbedtls version | crypto | tls | x509
-- | -- | -- | --
2.7.0 | libmbedcrypto.so.**1** | libmbedtls.so.10 | libmbedx509.so.0
2.6.0 | libmbedcrypto.so.**0** | libmbedtls.so.10 | libmbedx509.so.0

Since there are currently no recipes which depend on mbedtls, may I just drop all the compats (in both `PROVIDES` and `PROVIDES_devel`) and leave the burden of contacting upstream to someone who has more experience with this topic?

net-libs/mbedtls/mbedtls-2.7.0.recipe
alaviss: From [BuildingPackages][0]

> The `version_ref` after the "compat"/"compatible" string specifies the oldest version the resolvable is backwards compatible with.

So no, we don't use `compat`, only `lib<name>Version`

[0]: https://github.com/alaviss/haiku/blob/wiki-fixes/docs/develop/packages/BuildingPackages.rst#the-packageinfo

PR #2203, opened 2018-02-10
Qemu 2.11.1

extrowerk:
100% more icon and debug package


app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -125,6 +138,26 @@ INSTALL()
...
+
+	if [ $effectiveTargetArchitecture == x86_64 ]; then
+		addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
+		addResourcesToBinaries qemu.rdef $binDir/qemu-system-x86_64
+	else
+		addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
+	fi
+
 	# provide convenience symlink to just "qemu"
 	if [ $effectiveTargetArchitecture == x86_64 ]; then
```
fbrosson: What you have works fine, but I'd like to suggest a small variant:
```diff
 	$portDir/additional-files/qemu.rdef.in > qemu.rdef
 
+	addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
-	if [ $effectiveTargetArchitecture == x86_64 ]; then
+	if [ "$effectiveTargetArchitecture" = x86_64 ]; then
-		addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
 		addResourcesToBinaries qemu.rdef $binDir/qemu-system-x86_64
-	else
-		addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
 	fi
 
 	# provide convenience symlink to just "qemu"
-	if [ $effectiveTargetArchitecture == x86_64 ]; then
+	if [ "$effectiveTargetArchitecture" = x86_64 ]; then
 		ln -s qemu-system-x86_64 $binDir/qemu
 	else
 		ln -s qemu-system-i386 $binDir/qemu
 fi
```
(Reminder: `=` is slightly better than `==` because the latter only works with bash, while the former works with both bash and sh.)

fbrosson: I think it might be nice to use a [commandBinDir](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable).

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -105,16 +109,25 @@ BUILD_PREREQUIRES="
...
+	"$binDir"/qemu-system-i386
+
 BUILD()
 {
-	export CFLAGS="-lbsd -lnetwork"
+	export CFLAGS="-lbsd -lnetwork \
+		-I`finddir B_SYSTEM_HEADERS_DIRECTORY`/capstone"
 	runConfigure --omit-buildspec ./configure \
-		--disable-stack-protector \
+		--iasl=`finddir B_SYSTEM_BIN_DIRECTORY`/iasl \
```
alaviss: The configure script already have [a default value for iasl][0]. As long as `iasl` could be found in `PATH`, it should work.

```diff
-		--iasl=`finddir B_SYSTEM_BIN_DIRECTORY`/iasl \
```

[0]: https://github.com/qemu/qemu/blob/v2.11.0/configure#L251

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -69,13 +71,15 @@ REQUIRES="
...
 	lib:libusb_1.0$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
 	devel:libbz2$secondaryArchSuffix
+	devel:libcapstone$secondaryArchSuffix
 	devel:libcurl$secondaryArchSuffix
 	devel:libfdt$secondaryArchSuffix
 	#devel:libgcrypt$secondaryArchSuffix
```
alaviss: You should check if this line could be removed.

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -7,12 +7,12 @@ achieves very good performance."
 HOMEPAGE="http://wiki.qemu.org/"
 COPYRIGHT="2003-2017 Fabrice Bellard"
 LICENSE="GNU GPL v2"
```
alaviss: There might be [more licenses and copyright holders][0]. Thanks Debian folks for sorting them out.

[0]: http://metadata.ftp-master.debian.org/changelogs/main/q/qemu/qemu_2.11+dfsg-1_copyright

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -105,16 +109,25 @@ BUILD_PREREQUIRES="
...
 	cmd:make
 	cmd:find
+	cmd:iasl$secondaryArchSuffix
 	cmd:pkg_config$secondaryArchSuffix
 	cmd:pod2man
 	cmd:python
 	"
 
+defineDebugInfoPackage qemu$secondaryArchSuffix \
+	"$binDir"/qemu-system-x86_64 \
```
alaviss: The x86 package won't have this

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -0,0 +1,180 @@
...
+	cmd:find
+	cmd:iasl$secondaryArchSuffix
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:pod2man
+	cmd:python
+	"
+
+defineDebugInfoPackage qemu$secondaryArchSuffix \
+	"$binDir"/qemu-system-x86_64 \
+	"$binDir"/qemu-system-i386
```
alaviss:
```diff
-	"$binDir"/qemu-system-x86_64 \
-	"$binDir"/qemu-system-i386
+	"$commandBinDir"/qemu-system-x86_64 \
+	"$commandBinDir"/qemu-system-i386

File: app-emulation/qemu/qemu-2.11.0.recipe [hunk 7]
```diff
@@ -0,0 +1,180 @@
...
+	local LONG_INFO="$SUMMARY"
+	sed \
+		-e "s|@APP_SIGNATURE@|$APP_SIGNATURE|" \
+		-e "s|@MAJOR@|$MAJOR|" \
+		-e "s|@MIDDLE@|$MIDDLE|" \
+		-e "s|@MINOR@|$MINOR|" \
+		-e "s|@LONG_INFO@|$LONG_INFO|" \
+		$portDir/additional-files/qemu.rdef.in > qemu.rdef
+
+	addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
```
alaviss:
```diff
-	addResourcesToBinaries qemu.rdef $binDir/qemu-system-i386
+	addResourcesToBinaries qemu.rdef $commandBinDir/qemu-system-i386

File: app-emulation/qemu/qemu-2.11.0.recipe [hunk 8]
```diff
@@ -0,0 +1,180 @@
...
+	if [ "$effectiveTargetArchitecture" = x86_64 ]; then
+		addResourcesToBinaries qemu.rdef $binDir/qemu-system-x86_64
```
alaviss:
```diff
-		addResourcesToBinaries qemu.rdef $binDir/qemu-system-x86_64
+		addResourcesToBinaries qemu.rdef $commandBinDir/qemu-system-x86_64

File: app-emulation/qemu/qemu-2.11.0.recipe [hunk 9]
```diff
@@ -0,0 +1,180 @@
...
+	fi
+
+	# provide convenience symlink to just "qemu"
+	if [ "$effectiveTargetArchitecture" = x86_64 ]; then
+		ln -s qemu-system-x86_64 $binDir/qemu
+	else
+		ln -s qemu-system-i386 $binDir/qemu
```
alaviss:
```diff
-		ln -s qemu-system-x86_64 $binDir/qemu
+		ln -s qemu-system-x86_64 $commandBinDir/qemu
 	else
-		ln -s qemu-system-i386 $binDir/qemu
+		ln -s qemu-system-x86_64 $commandBinDir/qemu

File: app-emulation/qemu/qemu-2.11.0.recipe [hunk 10]
```diff
@@ -0,0 +1,180 @@
...
+	cmd:python
+	"
+
+defineDebugInfoPackage qemu$secondaryArchSuffix \
+	"$binDir"/qemu-system-x86_64 \
+	"$binDir"/qemu-system-i386
+
+BUILD()
+{
+	export CFLAGS="-I`finddir B_SYSTEM_HEADERS_DIRECTORY`/capstone"
```
alaviss: Also, the configure script [uses pkg-config to infer capstone include dirs][0], so you should not need to manually specify it. If it doesn't work, consider fixing capstone recipe instead.

```diff
-	export CFLAGS="-I`finddir B_SYSTEM_HEADERS_DIRECTORY`/capstone"
```

[0]: https://github.com/qemu/qemu/blob/v2.11.0/configure#L4487

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -0,0 +1,180 @@
...
+
+	runConfigure --omit-dirs "binDir sbinDir" \
+		--omit-buildspec ./configure \
+		--bindir="$commandBinDir" --sbindir="$commandBinDir" \
+		--iasl=`finddir B_SYSTEM_BIN_DIRECTORY`/iasl \
```
alaviss: The configure script already have [a default value for iasl][0]. As long as `iasl` could be found in `PATH`, it should work.

```diff
-		--iasl=`finddir B_SYSTEM_BIN_DIRECTORY`/iasl \
```

[0]: https://github.com/qemu/qemu/blob/v2.11.0/configure#L251

humdingerb:
> Please remove the .iom binary file.

I'm for keeping Icon-O-Matic files:
* You can edit it directly without having to extract the HVIF  from the binary.
* It allows naming of objects that you don't get when editing the HVIF.

app-emulation/qemu/qemu-2.11.0.recipe
extrowerk: The pkg-config passes the "develop/header" folder, not a capstone subfolder.
On linux it passes the include folder, not a subfolder.

So the build system using some magic somewhere, but i had no luck to nail it down. And as qemu still broken, i think i already invested way too much work into it. So it is wontfix for now.

app-emulation/qemu/patches/qemu-2.11.0.patchset
```diff
@@ -234,5 +234,55 @@ index 2e8b43b..adcc732 100644
...
+ 
+diff --git a/util/oslib-posix.c b/util/oslib-posix.c
+index bc58448..405c85f 100644
+--- a/util/oslib-posix.c
++++ b/util/oslib-posix.c
+@@ -38,6 +38,7 @@
+ #include <libgen.h>
+ #ifndef __HAIKU__
+ #include <sys/signal.h>
++#include <bsd/stdlib.h>
```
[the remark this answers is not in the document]
korli: This patch doesn't make much sense anyway: it's a patch for non-haiku platforms...

app-emulation/qemu/qemu-2.11.0.recipe
korli: FWIW haikuporter requires bash syntax anyway, it's kind of a moot point to bring this over and over again.

fbrosson: The reason why I recommend using sh syntax instead of bash syntax whenever it is possible (and does not cost anything) is that:
1) nobody can tell that haikuporter will always require bash;
2) if we ever choose to switch to another shell for haikuporter, if that other shell is compatible with sh but not with bash, then avoiding bashisms now will save us work and decrease git noise.

I'm not converting any already written recipes. I'm just making suggestions for recipes that are not merged yet.

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -0,0 +1,179 @@
...
+	cmd:qemu_system_ppc$commandSuffix = $portVersion
+	cmd:qemu_system_ppc64$commandSuffix = $portVersion
+	cmd:qemu_system_ppcemb$commandSuffix = $portVersion
+	cmd:qemu_system_s390x$commandSuffix = $portVersion
+	cmd:qemu_system_sh4$commandSuffix = $portVersion
+	cmd:qemu_system_sh4eb$commandSuffix = $portVersion
+	cmd:qemu_system_sparc$commandSuffix = $portVersion
+	cmd:qemu_system_sparc64$commandSuffix = $portVersion
+	cmd:qemu_system_unicore32$commandSuffix = $portVersion
+	cmd:qemu_system_x86_64$commandSuffix = $portVersion
```
fbrosson: Suggestion:
```
case "$effectiveTargetArchitecture" in
	x86) qemuArchitectures="i386";;
	x86_64) qemuArchitectures="i386 x86_64";;
	arm|ppc) qemuArchitectures="$effectiveTargetArchitecture";;
esac

PROVIDES="
	qemu$secondaryArchSuffix = $portVersion
	cmd:qemu$commandSuffix = $portVersion
	cmd:qemu_img$commandSuffix = $portVersion
	cmd:qemu_io$commandSuffix = $portVersion
	"
for i in $qemuArchitectures; do
PROVIDES="$PROVIDES
	cmd:qemu_system_$i$commandSuffix = $portVersion
	"
done
```

app-emulation/qemu/qemu-2.11.0.recipe
```diff
@@ -0,0 +1,179 @@
...
+	cmd:make
+	cmd:find
+	cmd:iasl$secondaryArchSuffix
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:pod2man
+	cmd:python
+	"
+
+defineDebugInfoPackage qemu$secondaryArchSuffix \
+	"$commandBinDir"/qemu-system-x86_64 \
```
fbrosson:
```diff
 defineDebugInfoPackage qemu$secondaryArchSuffix \
-	"$commandBinDir"/qemu-system-x86_64 \
-	"$commandBinDir"/qemu-system-i386
+	"`echo "$PROVIDES" | sed -n \
+		-e "s|cmd:\(.*\)$commandSuffix = .*|$commandBinDir/\1|p" \
+	`"
```

extrowerk: No, provides have more cmd Than in the package.

fbrosson: The above comment assumes `PROVIDES` is defined with [this suggestion](#discussion_r167442676).
(But I need to fix the sed command above.)

fbrosson: > No, provides have more cmd Than in the package.

I have built qemu on x86_64, and, not counting the qemu symlink, the only runtimes I get in `/system/bin/` are:
```
qemu-img
qemu-io
qemu-system-i386
qemu-system-x86_64
```

I guess the other `qemu-system-*` need some extra dependencies to be built.
Do you obtain other runtimes (which I did not get)?
(If not, we should either drop the extra `cmd:` lines or comment them out.)

extrowerk: No, but korli and mmlr worked on this recipe, they defined it so, and i will do everything, but i won't do that.
The recipe is still disabled on every arch and it have almost no practical use, so i don't want to change the provides sections.

fbrosson: OK, no problem.
(I did not know that the `cmd:` lines were from korli and mmlr.)
So, of course, that's OK to keep them.

app-emulation/qemu/qemu-2.11.0.recipe
waddlesplash: Eh. Fine as is, I think.

fbrosson:
You might wish to try `2.11.1` which was released 3 days ago.

extrowerk:
@fbrosson: with maximal respect: no. qemu is just broken on Haiku, i try to keep the recipe in working state but i don't want to follow every minor release. This is ok to merge as is.

fbrosson:
I wouldn't say the PR is OK as is, mainly because 2.11.1 is out, but I don't object a merge.

extrowerk:
I assume it is ok now after the bump.

PR #2204, opened 2018-02-10
imagemagick, bump version

Begasus:


fbrosson:
Since you are adding `imagemagick7-7.0.7.22.recipe` and `imagemagick-7.0.7.22.patchset`, it would be nice to remove the `imagemagick7-7.0.7.17.recipe` and  `imagemagick-7.0.7.17.patchset` files, in order to have only one 7.0.7.x.

Begasus:
I wanted to wait until this got approved :) Tests are the same as for previous version

Begasus:
PS, not tested on x86_64 ...

fbrosson:
FWIW, here's a summary of the tests results for imagemagick7-7.0.7.22 on x86 secondary arch:

TOTAL: | 86
-- | --
PASS: | 79
FAIL: | 7

Failed tests | Error message
------------------------- | --
tests/wandtest.tap&nbsp;1 | tests/wandtest.c main 5321 non-conforming drawing primitive definition **text** @ error/draw.c/DrawImage/3284
Magick++/tests/tests.tap&nbsp;10 Magick++/demo/demos.tap&nbsp;2<br/>Magick++/demo/demos.tap&nbsp;3<br/>Magick++/demo/demos.tap&nbsp;5<br/>Magick++/demo/demos.tap&nbsp;6<br/>Magick++/demo/demos.tap&nbsp;7 | Caught exception: lt-button: unable to read font **helvetica** @ error/annotate.c/RenderFreetype/1388



BTW, I just did some investigation (with find, grep, etc...) to see if we have any recipes which depend on any of the `cmd:` lines in imagemagick.
The full list of `cmd:` items provided by imagemagick7 are: animate compare composite conjure convert display identify import magick magick_script mogrify montage stream
Some of these appear in grep results ("compare" and "convert") for some recipes (in app-crypt/gnupg/, dev-libs/caffe/, media-libs/leptonica/ and sys-apps/dtc/), but after a closer look, it turns out that these are not the same command names.

The **conclusion** is that we can safely convert this recipe to use the commandBinDir & commandSuffix, without even having to change any other recipe.

@Begasus, I haven't tested the recipe on x86_64 yet, but I'll do so asap. Once I have checked this, I can either:
* merge your PR as it is now, and then make the changes in this recipe to make it use the [commandBinDir](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) & commandSuffix variables (if nobody objects, of course), or:
* leave that to you if you agree.

(There will be less noise on our Git logs if we choose the second option :)

media-gfx/imagemagick/imagemagick7-7.0.7.22.recipe
```diff
@@ -63,14 +70,32 @@ PROVIDES_devel="
...
+	$binDir/identify \
+	$binDir/import \
+	$binDir/magick \
+	$binDir/magick-script \
+	$binDir/mogrify \
+	$binDir/montage \
+	$binDir/stream \
+	$libDir/libMagick++-7.Q16HDRI.so.${libMagickVersion%%.*} \
+	$libDir/libMagickCore-7.Q16HDRI.so.${libMagickCoreVersion%%.*} \
+	$libDir/libMagickWand-7.Q16HDRI.so.${libMagickWandVersion%%.*}
```
fbrosson: I'd recommend this variant:
```diff
-	$libDir/libMagick++-7.Q16HDRI.so.${libMagickVersion%%.*} \
+	$libDir/libMagick++-7.Q16HDRI.so.$libMagickVersion \
-	$libDir/libMagickCore-7.Q16HDRI.so.${libMagickCoreVersion%%.*} \
+	$libDir/libMagickCore-7.Q16HDRI.so.$libMagickCoreVersion \
-	$libDir/libMagickWand-7.Q16HDRI.so.${libMagickWandVersion%%.*}
+	$libDir/libMagickWand-7.Q16HDRI.so.$libMagickWandVersion
```
(This has the advantage of producing an error if we forget to update any of the `lib*Version` variables.)

Begasus: The lib versions for the debuginfopackage rely on the previous, it didn't work with "4.0.0" it needed "4" ...

Begasus:
This is getting complicated, we have a "secondaryArchSuffix" for a reason, for some I can understand the need for commandBinDir etc, but this one shouldn't produce an error?

media-gfx/imagemagick/imagemagick7-7.0.7.22.recipe
fbrosson: Well, I tested both of my builds (x86 secondary arch and x86_64) with the above suggestion, and everything was OK.

Begasus: Let me check to be sure ...

Begasus:
It's complicated if a binary uses $prefix and you "need" to use commandSuffix, I think in this case it's not needed?

Begasus:
I should rephrase it, it's complicated if a binary uses $binDir and one try's to push it to use $commandSuffix ...

media-gfx/imagemagick/imagemagick7-7.0.7.22.recipe
Begasus: You were right, updated ...

fbrosson:
No problem. Let's forget about the commandBinDir and commandSuffix variables.
The recipe is perfectly OK without these. Let's merge it!
(I might try do add these when a new version becomes available.)

PR #2213, opened 2018-02-13
libatomic_ops: bump version to 7.6.2

ivmai:


sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,66 @@
...
+	cmd:make
+	cmd:sed
+	cmd:libtool$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	runConfigure ./configure --disable-dependency-tracking \
+			--enable-shared \
+			--with-gnu-ld
```
fbrosson:
```diff
 	runConfigure ./configure --disable-dependency-tracking \
-			--enable-shared \
+		--enable-shared \
-			--with-gnu-ld
+		--with-gnu-ld
```

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,66 @@
...
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:gawk
+	cmd:grep
+	cmd:make
+	cmd:sed
+	cmd:libtool$secondaryArchSuffix
```
fbrosson:
```diff
BUILD_PREREQUIRES="
+	cmd:gawk
 	cmd:gcc$secondaryArchSuffix
+	cmd:grep
+	cmd:ld$secondaryArchSuffix
-	cmd:gawk
-	cmd:grep
+	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	cmd:sed
-	cmd:libtool$secondaryArchSuffix
 	"
```
(`libtool` -> `libtoolize`, plus alphabetical sorting)

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,66 @@
+SUMMARY="An atomic memory update operations portable implementation"
+DESCRIPTION="Semi-portable access to hardware-provided \
+atomic memory update operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/ivmai/libatomic_ops/"
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
```

BTW, in `DESCRIPTION`, you might wish to move the words `atomic memory update` (currently in the beginning of line 3) to the end of line 2.
(This would be more consistent with our recommended "80-chars limit" _philosophy_...)

ivmai:
All done.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -1,7 +1,6 @@
 SUMMARY="An atomic memory update operations portable implementation"
-DESCRIPTION="Semi-portable access to hardware-provided \
-atomic memory update operations on a number architectures."
-HOMEPAGE="https://github.com/ivmai/libatomic_ops/"
+DESCRIPTION="Semi-portable access to hardware-provided atomic memory update operations on a number architectures."
```
fbrosson: Oops, what I had in mind was:
```
DESCRIPTION="Semi-portable access to hardware-provided atomic memory update \
operations on a number architectures." 
```
I like to use diffs (with red and green lines) but I don't know why I did not use it to comment about `DESCRIPTION`.
Well, I guess I was affraid of being too picky... :D

Anyway, thanks for the updates so far.
(I'll test the recipe asap.)

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,65 @@
...
+	make install
+
+	fixPkgconfig
+
+	prepareInstalledDevelLibs \
+			libatomic_ops libatomic_ops_gpl
+
+	# devel package
+	packageEntries devel \
+			$developDir $dataDir
```
fbrosson:
```diff
	packageEntries devel \
-			$developDir $dataDir
+		$developDir
```
With `libatomic_ops-7.4.4.recipe` we do get a `$dataDir`, with these contents:
```
data/libatomic_ops/COPYING
data/libatomic_ops/LICENSING.txt
data/libatomic_ops/README.md
data/libatomic_ops/README.txt
data/libatomic_ops/README_malloc.txt
data/libatomic_ops/README_stack.txt
data/libatomic_ops/README_win32.txt
```
But with `libatomic_ops-7.6.2.recipe` the build does not install these.

Indeed, one of the changes upstream made to the main `Makefile.am` is:
```diff
---libatomic_ops-7.4.4/Makefile.am
+++libatomic_ops-7.6.2/Makefile.am
@@ -6,8 +6,19 @@
 pkgconfig_DATA = pkgconfig/atomic_ops.pc
 noinst_DATA = pkgconfig/atomic_ops-uninstalled.pc
 
-dist_pkgdata_DATA = COPYING README.md
+if ENABLE_DOCS
+dist_doc_DATA = COPYING README.md
+endif
 
 EXTRA_DIST = autogen.sh
```

The good news is that the `README.md` file, with 4 other `README_*.txt` are now in `$docDir`.
So all you need to do is drop the `$dataDir` from line 64, as I suggested in the diff on the top of this message.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,65 @@
...
+SOURCE_URI="https://github.com/ivmai/libatomic_ops/releases/download/v$portVersion/libatomic_ops-$portVersion.tar.gz"
+CHECKSUM_SHA256="219724edad3d580d4d37b22e1d7cb52f0006d282d26a9b8681b560a625142ee6"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	libatomic_ops$secondaryArchSuffix = $portVersion
+	lib:libatomic_ops$secondaryArchSuffix = 2.1.1 compat >= 1
+	lib:libatomic_ops_gpl$secondaryArchSuffix = 2.1.2 compat >= 1
```
fbrosson:
```diff
-	lib:libatomic_ops$secondaryArchSuffix = 2.1.1 compat >= 1
+	lib:libatomic_ops$secondaryArchSuffix = 1.1.1 compat >= 1
-	lib:libatomic_ops_gpl$secondaryArchSuffix = 2.1.2 compat >= 1
+	lib:libatomic_ops_gpl$secondaryArchSuffix = 1.1.2 compat >= 1
```

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,65 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libatomic_ops$secondaryArchSuffix_devel = $portVersion
+	devel:libatomic_ops$secondaryArchSuffix = 2.1.1 compat >= 1
+	devel:libatomic_ops_gpl$secondaryArchSuffix = 2.1.2 compat >= 1
```
fbrosson:
```diff
-	devel:libatomic_ops$secondaryArchSuffix = 2.1.1 compat >= 1
+	devel:libatomic_ops$secondaryArchSuffix = 1.1.1 compat >= 1
-	devel:libatomic_ops_gpl$secondaryArchSuffix = 2.1.2 compat >= 1
+	devel:libatomic_ops_gpl$secondaryArchSuffix = 1.1.2 compat >= 1
```

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
fbrosson: Hmm, @ivmai, I've just realized that **_you_** are the upstream author! This is great!

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,65 @@
...
+		--with-gnu-ld
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	fixPkgconfig
```
fbrosson: * The `COPYING` file that gets installed in `$docDir` is almost the same as the `/system/data/licenses/GNU GPL v2` we have, so maybe we could replace it by a relative symlink.
* It might be a good idea to move the readme files to the `$developDocDir`.
* We like to drop the libtool files.
* We usually call `fixPkgconfig` right after `prepareInstalledDevelLibs`.

```diff
 	make install
 
-	fixPkgconfig
+	rm -f $docDir/COPYING
+	ln -r -s $dataDir/licenses/"GNU GPL v2" $docDir/COPYING
+	mkdir -p $developDocDir
+	mv $docDir/README* $developDocDir
+
+	# drop libtool files
+	rm -f $libDir/libatomic_ops*.la
 
 	prepareInstalledDevelLibs \
 		libatomic_ops \
 		libatomic_ops_gpl
+	fixPkgconfig
```

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,65 @@
+SUMMARY="An atomic memory update operations portable implementation"
+DESCRIPTION="Semi-portable access to hardware-provided atomic memory update operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
+COPYRIGHT="2003-2011 Hewlett-Packard Development Company, L.P."
+LICENSE="GNU GPL v2"
```
fbrosson: It looks like this would be more appropriate.
```diff
-LICENSE="GNU GPL v2"
+LICENSE="MIT
+	GNU GPL v2"
```

waddlesplash: Or the other way around, perhaps...

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
fbrosson: @waddlesplash: Ok, thanks!

@ivmai: Here's the new diff, to address @waddlesplash's comment.
```diff
 	make install
 
-	fixPkgconfig
+	mkdir -p $developDocDir
+	mv $docDir/README* $developDocDir
+
+	# remove libtool files
+	rm -f $libDir/libatomic_ops*.la
 
 	prepareInstalledDevelLibs \
 		libatomic_ops \
 		libatomic_ops_gpl
+	fixPkgconfig
```
This is almost the same diff as in my previous comment. The only difference is that it does not replace the COPYING file by a symlink.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
alaviss: Shouldn't we use [`libVersionCompat`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-libversioncompat-user-variable)?

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
alaviss: You could just remove the file. The GNU GPL v2 doesn't require one to carry a copy of the license text with the package.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
alaviss: The main library (`libatomic_ops`) is MIT licensed, so I would prefer having MIT as the first entry.

Also, don't forget this little [custom license][0]

[0]: https://github.com/ivmai/libatomic_ops/blob/v7.6.2/doc/LICENSING.txt#L43

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,65 @@
+SUMMARY="An atomic memory update operations portable implementation"
+DESCRIPTION="Semi-portable access to hardware-provided atomic memory update operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
+COPYRIGHT="2003-2011 Hewlett-Packard Development Company, L.P."
```
alaviss: Missing holders:
- [1991-1994 Xerox Corporation][0]
- [1996-1999 Silicon Graphics][0]
- [2016 Marek Vasut][1]
- [2005, 2007 Thiemo Seufer][2]
- [2007 NEC LE-IT][3]

[0]: https://github.com/ivmai/libatomic_ops/blob/v7.6.2/src/atomic_ops/sysdeps/gcc/generic-arithm.template#L2
[2]: https://github.com/ivmai/libatomic_ops/blob/v7.6.2/src/atomic_ops/sysdeps/gcc/mips.h#L2
[1]: https://github.com/ivmai/libatomic_ops/blob/v7.6.2/src/atomic_ops/sysdeps/gcc/nios2.h#L2
[3]: https://github.com/ivmai/libatomic_ops/blob/v7.6.2/src/atomic_ops/sysdeps/armcc/arm_v6.h#L2

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,78 @@
+SUMMARY="An atomic memory update operations portable implementation"
+DESCRIPTION="Semi-portable access to hardware-provided atomic memory update \
+operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
+COPYRIGHT="1991-1994 by Xerox Corporation.  All rights reserved.
+	1996-1999 by Silicon Graphics.  All rights reserved.
+	2003-2011 Hewlett-Packard Development Company, L.P.
+	2005,2007 Thiemo Seufer
+	2007 by NEC LE-IT
```
fbrosson: We usually drop the "All rights reserved", although I don't know if we should continue doing so.

```diff
-COPYRIGHT="1991-1994 by Xerox Corporation.  All rights reserved.
+COPYRIGHT="1991-1994 by Xerox Corporation
-	1996-1999 by Silicon Graphics.  All rights reserved.
+	1996-1999 by Silicon Graphics
 	2003-2011 Hewlett-Packard Development Company, L.P.
 	2005,2007 Thiemo Seufer
-	2007 by NEC LE-IT
+	2007 NEC LE-IT
 	2016 Marek Vasut"
```

ivmai:
Dear colleagues,

Thank you for your suggestions, I have followed them (except for  "Shouldn't we use libVersionCompat?" one).
Please let me know if something more should be changed.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,78 @@
...
+SOURCE_URI="https://github.com/ivmai/libatomic_ops/releases/download/v$portVersion/libatomic_ops-$portVersion.tar.gz"
+CHECKSUM_SHA256="219724edad3d580d4d37b22e1d7cb52f0006d282d26a9b8681b560a625142ee6"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	libatomic_ops$secondaryArchSuffix = $portVersion
+	lib:libatomic_ops$secondaryArchSuffix = 1.1.1 compat >= 1
+	lib:libatomic_ops_gpl$secondaryArchSuffix = 1.1.2 compat >= 1
```
fbrosson: You might wish to define 4 custom variables (like `lib{ops,gpl}Version{,Compat}`) and use them later. Here's a suggestion:
```diff
 SECONDARY_ARCHITECTURES="x86"
+
+libopsVersion="1.1.1"
+libgplVersion="1.1.2"
+
+libopsVersionCompat="$libopsVersion compat >= ${libopsVersion%%.*}"
+libgplVersionCompat="$libgplVersion compat >= ${libgplVersion%%.*}"
+
 PROVIDES="
 	libatomic_ops$secondaryArchSuffix = $portVersion
-	lib:libatomic_ops$secondaryArchSuffix = 1.1.1 compat >= 1
+	lib:libatomic_ops$secondaryArchSuffix = $libopsVersionCompat
-	lib:libatomic_ops_gpl$secondaryArchSuffix = 1.1.2 compat >= 1
+	lib:libatomic_ops_gpl$secondaryArchSuffix = $libgplVersionCompat
 	"
```

ivmai:
@alaviss, could you guide me about libVersionCompat change please?

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,78 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libatomic_ops$secondaryArchSuffix_devel = $portVersion
+	devel:libatomic_ops$secondaryArchSuffix = 1.1.1 compat >= 1
+	devel:libatomic_ops_gpl$secondaryArchSuffix = 1.1.2 compat >= 1
```
fbrosson:
```diff
 PROVIDES_devel="
 	libatomic_ops$secondaryArchSuffix_devel = $portVersion
-	devel:libatomic_ops$secondaryArchSuffix = 1.1.1 compat >= 1
+	devel:libatomic_ops$secondaryArchSuffix = $libopsVersionCompat
-	devel:libatomic_ops_gpl$secondaryArchSuffix = 1.1.2 compat >= 1
+	devel:libatomic_ops_gpl$secondaryArchSuffix = $libgplVersionCompat
 	"
```

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
fbrosson: BTW, these custom variables are handy if/when we create the [debuginfo](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#debuginfo-package) package. For this recipe we could insert this block before the `BUILD()` function:
```diff
 	cmd:make
 	cmd:sed
 	"
+
+defineDebugInfoPackage libatomic_ops$secondaryArchSuffix \
+        "$libDir"/libatomic_ops.so.$libopsVersion \
+        "$libDir"/libatomic_ops_gpl.so.$libgplVersion
+
 BUILD()
 {
```

Does anyone know if, for this particular recipe, it would be a good idea to create the debuginfo package?
(Those contributors who think it's a good idea, please add a 👍 to this message. Those who think the opposite, please add a 👎 , of course. Thanks!)
(Of course, if anyone wishes to write a real comment, please do so!)

ivmai:
> cosmetic edits
> suggestion to use custom variables 

Done.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,84 @@
+SUMMARY="An atomic memory update operations portable implementation"
+DESCRIPTION="Semi-portable access to hardware-provided atomic memory update \
+operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
+COPYRIGHT="1991-1994 Xerox Corporation
+	1996-1999 Silicon Graphics
+	2003-2011 Hewlett-Packard Development Company, L.P.
+	2005,2007 Thiemo Seufer
```
fbrosson: I forgot to spot the missing space here, but since there are a few other changes, could you fix this? Thanks!
```diff
-	2005,2007 Thiemo Seufer
+	2005, 2007 Thiemo Seufer
```
(Well, although we have several recipes with a similar missing space, most recipes have that space, so it's better to add it here too.)

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+	"
+REQUIRES_devel="
+	libatomic_ops$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gawk
```
fbrosson:
```diff
 BUILD_PREREQUIRES="
+	cmd:find
 	cmd:gawk
```
(It looks like `find` is needed for running `make check`.)

BTW, could you please add a `TEST()` function, after the `INSTALL()` function? Thanks!
```diff
 	packageEntries devel \
 		$developDir
 }
+
+TEST()
+{
+	make check
+}
```

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
+COPYRIGHT="1991-1994 Xerox Corporation
+	1996-1999 Silicon Graphics
+	2003-2011 Hewlett-Packard Development Company, L.P.
+	2005,2007 Thiemo Seufer
+	2007 NEC LE-IT
+	2016 Marek Vasut"
+LICENSE="MIT
+	GNU GPL v2"
```
alaviss: The [custom license][0] should also be included.

To add a custom license for a port, create a folder named "licenses" next to the recipe, and put the license text to a file named after the license. For this case, you would want to put the [license text][0] to `sys-libs/libatomic_ops/licenses/Boehm`

```diff
 LICENSE="MIT
+	Boehm
 	GNU GPL v2"
```

[0]: https://github.com/ivmai/libatomic_ops/blob/v7.6.2/doc/LICENSING.txt#L43

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+CHECKSUM_SHA256="219724edad3d580d4d37b22e1d7cb52f0006d282d26a9b8681b560a625142ee6"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libopsVersion="1.1.1"
+libgplVersion="1.1.2"
+
+libopsVersionCompat="$libopsVersion compat >= ${libopsVersion%%.*}"
+libgplVersionCompat="$libgplVersion compat >= ${libgplVersion%%.*}"
```
alaviss: There should be no space between `libVersion` and `libVersionCompat` as in the [Guidelines][0].

Also, as a suggestion, we usually use camelCase for user variables, so I suggest we stick to that with this recipe.

```diff
-libopsVersion="1.1.1"
-libgplVersion="1.1.2"
-
-libopsVersionCompat="$libopsVersion compat >= ${libopsVersion%%.*}"
-libgplVersionCompat="$libgplVersion compat >= ${libgplVersion%%.*}"
+libVersion="1.1.1"
+libGplVersion="1.1.2"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+libGplVersionCompat="$libGplVersion compat >= ${libGplVersion%%.*}
```

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+	cmd:grep
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	"
+
+BUILD()
+{
+	runConfigure ./configure --disable-dependency-tracking \
```
alaviss: Looks like this port make use of autoconf. Could you check if the library was built with debuginfo? If so we need a debuginfo package as @fbrosson [suggested](https://github.com/haikuports/haikuports/pull/2213#discussion_r168295424)

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
fbrosson: Shall we add a small exception, to allow an empty line between the `libVersion`s and the `libVersionCompat`s, **_if_** there is more than one shared library?
(I think it looks nicer with that space, but I'm also OK without it.)

ivmai:
All done (copyright space fix, copyright addition, vars camel case, TEST, debug-info).

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,94 @@
...
+	cmd:grep
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:sed
+	"
+
+defineDebugInfoPackage libatomic_ops$secondaryArchSuffix \
+	"$libDir"/libatomic_ops.so.$libVersion \
+	"$libDir"/libatomic_ops_gpl.so.libGplVersion
```
fbrosson:
```diff
-	"$libDir"/libatomic_ops_gpl.so.libGplVersion
+	"$libDir"/libatomic_ops_gpl.so.$libGplVersion
```

ivmai:
> "$libDir"/libatomic_ops_gpl.so.libGplVersion

Fixed.

sys-libs/libatomic_ops/libatomic_ops-7.6.2.recipe
```diff
@@ -0,0 +1,94 @@
...
+operations on a number architectures."
+HOMEPAGE="https://github.com/ivmai/libatomic_ops"
+COPYRIGHT="1991-1994 Xerox Corporation
+	1996-1999 Silicon Graphics
+	2003-2011 Hewlett-Packard Development Company, L.P.
+	2005, 2007 Thiemo Seufer
+	2007 NEC LE-IT
+	2016 Marek Vasut"
+LICENSE="MIT
+	Boehm
```
fbrosson: Hmm, I just noticed that the contents of this file are not the Boehm license. It just mentions that license.
If we keep a `Boehm` in `LICENSE` then this file will be in `/system/data/licenses/Boehm`, potentially hiding a correct copy of the real Boehm license.
So, since that file is the same as `doc/LICENSING.txt`, I'd suggest that we remove `Boehm` from the `LICENSE` variable and, instead, only keep one copy of it, which is already in `/system/documentation/packages/libatomic_ops/LICENSING.txt` when the package is installed.

```diff
 LICENSE="MIT
-	Boehm
	GNU GPL v2"
```
@alaviss, would you be OK with this?

(Sorry for not having noticed this earlier.)
(Everything else is OK :)

pulkomandy: The LICENSE field should try to be correct. So if some license is used in the code, it should be mentionned here. And of course that means the license file should be there. It can be a symlink if said license is already included in the package, however.

So:

    ln -s $documentationDir/LICENSING.txt $dataDir/licenses/Boehm

ivmai:
How about renaming "Boehm" to e.g. "BoehmLibatomicOps" file?

ivmai:
I removed the license file and added a symlink on install: e6a3e14

ivmai:
Error: No match found for license BoehmLibatomicOps

ivmai:
I took the license from http://www.hboehm.info/gc/license.txt

fbrosson:
The recipe builds fine on all supported architectures: x86_64 and x86 (both primary and secondary).
And they all pass all tests :)
```
PASS: test_atomic
PASS: test_atomic_generalized
PASS: test_stack
PASS: test_malloc
PASS: test_atomic_pthreads
Testsuite summary for libatomic_ops 7.6.2
# TOTAL: 5
# PASS:  5
```
And, for me, everything is OK, so I'll approuve the PR.
(I'll let @pulkomandy and @alaviss confirm that they agree with me, in partucular about the way the recipe handles the licensing topic.)

ivmai:
Thank you.
Now I will look at dev-libs/boehm-gc upgrading to 7.6.4

PR #2218, opened 2018-02-17
Guitar: new recipe

extrowerk:


dev-vcs/guitar/guitar-0.2.0~git.recipe
```diff
@@ -0,0 +1,84 @@
...
+* Open Source, Free Software
+* Run on Windows, macOS and Linux
+* Written in C++
+* Powered by Qt 5"
+HOMEPAGE="https://soramimi.github.io/Guitar/"
+COPYRIGHT="2017 S.Fuchita"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="828a6f934d86f8eb42a81c12fdb6ae52063a6835"
+SOURCE_URI="https://github.com/soramimi/guitar/archive/$srcGitRev.tar.gz"
```
fbrosson:
```diff
-SOURCE_URI="https://github.com/soramimi/guitar/archive/$srcGitRev.tar.gz"
+SOURCE_URI="https://github.com/soramimi/Guitar/archive/$srcGitRev.tar.gz"
```

fbrosson: I **_do_** know that it also works with a lower case "g".
I'm just suggesting this to because the official name is "Guitar" (and also because we have an upper-case "G" in both `SOURCE_DIR` and `SOURCE_FILENAME`.)

dev-vcs/guitar/guitar-0.2.0~git.recipe
```diff
@@ -0,0 +1,84 @@
...
+CHECKSUM_SHA256="dd6d5f54856dbb6911993344bad2e684564654a04ddf01ee1e016d7018654178"
+SOURCE_DIR="Guitar-$srcGitRev"
+SOURCE_FILENAME="Guitar-$portVersion.tar.gz"
```
fbrosson:
```diff
-SOURCE_FILENAME="Guitar-$portVersion.tar.gz"
+SOURCE_FILENAME="Guitar-$portVersion-$srcGitRev.tar.gz"
```

extrowerk:
Done, thanks for pointing them out.

pulkomandy:
The use of tilde in the version makes it so the git version is compared "less than" to the stable one. Otherwise you would use a dot or an underscome.

dev-vcs/guitar/guitar-0.2.1~git.recipe
```diff
@@ -0,0 +1,85 @@
...
+SOURCE_DIR="Guitar-$srcGitRev"
+SOURCE_FILENAME="Guitar-$portVersion-$srcGitRev.tar.gz"
+ADDITIONAL_FILES="guitar.rdef.in"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	guitar$secondaryArchSuffix = $portVersion
+	app:guitar
```
fbrosson:
```diff
-	app:guitar
+	app:Guitar
```

dev-vcs/guitar/guitar-0.2.1~git.recipe
```diff
@@ -0,0 +1,85 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	cmd:git
```
fbrosson: The app builds and seems to works fine. Tested on x86_64 and x86 second arch.

```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
+	cmd:file
 	cmd:git
```
This is because the app asks for the paths of both the `git` and `file` commands.

BTW, I'd also recommend this change in `PROViDES`:
```diff
-	app:guitar
+	app:Guitar
```

dev-vcs/guitar/guitar-0.2.1~git.recipe
```diff
@@ -0,0 +1,86 @@
...
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Network$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
```
fbrosson:
```diff
-	cmd:git
```
I think that `cmd:git` is only needed in `REQUIRES`.
I'll kick a build without it in `BUILD_PREREQUIRES` to check this.

fbrosson: I can confitm that you can remove `cmd:git` from `BUILD_PREREQUIRES`. (I even recommend doing so.)
The resulting package is identical to the one we get if we have `cmd:git` in `BUILD_PREREQUIRES`.
(Of course there is one difference, but it is just in the metadata in `.PackageInfo` file.)

PR #2238, opened 2018-02-21
check: bump to 0.12.0, bring back 0.10.0 for x86_gcc2.

fbrosson:
0.10.0 is the last release that works on x86_gcc2 primary arch.


fbrosson:
I'm using `REVISION="4"` in `check-0.10.0.recipe` because that recipe has already existed in the repo, until it got bumped from `0.10.0` to `0.11.0`.
BTW, the bump was OK for x86_64 but not for x86_gcc2 as `0.11.0` does not build on x86_gcc2 primary arch.

Also, the (gcc2) patch for `0.11.0` was no longer required, so I'm "recycling" it for `0.10.0`, with a simple "backwards" rename, from `check-0.11.0.patch` to `check-0.10.0.patch`.

pulkomandy:
I had removed the gcc2 version because there was nothing still using it,
IIRC. There is no need to keep everything available for gcc2, unless
there is something that depends on it, right?

fbrosson:
The following recipes are marked as tested on x86_gcc2 and have `lib:libcheck` in their `BUILD_REQUIRES`:
* dev-libs/libwapcaplet/libwapcaplet-0.3.0.recipe
* dev-libs/libwapcaplet/libwapcaplet-0.4.0.recipe

PR #2244, opened 2018-02-23
Add workflow diagram

scottmc:
This was creating during GCI 2017 by student Bach Nguyen


fbrosson:
@scottmc, I might be wrong, but I think that it would be better to put the files `HaikuPorts-HaikuArchives-WorkflowDiagram.png` and `HaikuPorts-HaikuArchives-WorkflowDiagram.svg` in the wiki or the website.

fbrosson:
I think so. And I'd say that we would just need to treat the svg file as if it was an "ordinary" image.

scottmc:
Yeah, the intention is the store them in the repo, perhaps in a subdirectory, and then add the png version to the wiki where it fits.  Only reason for putting the .svg in as well is if we need to modify the graphic later.

fbrosson:
Yep, it looks like it is possible to add an image. The "toolbar" has:
h1 h2 h3 "link icon" "image icon" B(old) I(talics) ...
So I guess clicking on the "image icon" will trigger an "Open..." in the browser.

scottmc:
Maybe @waddlesplash can suggest a good place to store these images so that they can be added to the wiki page for both HaikuPorts and HaikuArchives?  Perhaps haike/website?  or elsewhere?

fbrosson:
I suspect that it is not possible to insert an orphan image in the wiki.
It looks like the editor only allows to embed the image when editing (or creating) a page.

So, as you suggested, it might be easier to store the image somewhere in [haiku/website](https://github.com/haiku/website).

waddlesplash:
It is possible to embed orphan images in the wiki, because the wiki is just another git repository. Clone it, add the images there, and that's that. Don't put it in master.

PR #2246, opened 2018-02-24
RocksnDiamonds, bump version

Begasus:


pulkomandy:
SDL1 is more stable on Haiku than SDL2, so there is no hurry in migrating.

PR #2247, opened 2018-02-24
Guitar: git required as BUILD_PREREQ TOO

extrowerk:


fbrosson:
Sorry @extrowerk, I was sure it was not needed, but I guess I made a mistake during my test:
I had first built it with the `cmd:git` in `BUILD_PREREQUIRES`, and then kicked a new build, without the `cmd:git`. I guess I forgot to clean the work directory before the second build :(
(I'll check this.)

fbrosson:
Well, I just checked again (on both x86_64 and x86 second arch), and guitar does **_not_** need any `cmd:git` in `BUILD_PREREQUIRES`. It's only needed in `REQUIRES`.

What error did you get when you tried to build guitar without the `cmd:git` in `BUILD_PREREQUIRES`?

extrowerk:
http://kitchen.haiku-os.org/master/x86_64/?buildrunDir=998&viewMode=expanded
http://kitchen.haiku-os.org/master/x86_gcc2/?buildrunDir=845&viewMode=expanded

extrowerk:
make uses the prepare.rb ruby script to generate the version.c. To do so, it using git, that's why it is required. Delete your work folder and it will fail to regenerate it.

fbrosson:
`prepare.rb` only fails if it tries to use git.
And it only tries to use git if it finds a `.git` directory.

This means that:
1) The builders should use haikuporter's `--no-git-repo` (or `-G`) switch.
2) guitar does not need `cmd:git` in `BUILD_PREREQUIRES`.

BTW, I have these shell aliases:
```
alias hp='/bin/haikuporter              -G --get-dependencies --no-source-packages'
alias hpg='/bin/haikuporter                --get-dependencies --no-source-packages'
```
and I only use hpg when I know that I'll need to make changes to a patchset.
For guitar I used hp (i.e. with `-G`)

The reason why I'm saying that the builders should call haikuporter with `-G` is that for some recipes (like this one) there is no way to prevent the build from failing if it finds the `.git` directory and tries to use git.

Well, at least we found the explanation ;)

PR #2245, opened 2018-02-24
 RetroArch: bump version to 1.7.1

gouchi:
RetroArch should be updated to version 1.7.1 as the [Haiku port has been restored](https://github.com/libretro/RetroArch/blob/master/CHANGES.md#171). Please confirm it builds with HaikuPorter first.

PR #2248, opened 2018-02-25
RetroArch 1.7.1 recipe + FBAlpha 0.2.97.42 recipe

kwyxz:
What this does:
- build RetroArch and FBAlpha core packages
- configure global settings retroarch.cfg with default Haiku-friendly paths
- ship with visual assets for RetroArch's GUI
- include an .info file for FBAlpha core

I did not mean to cut PR #2245 short, but have been working on these for a few days now. 
Please consider this pull request. Thanks!


games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
```diff
@@ -0,0 +1,44 @@
+SUMMARY="A port of Final Burn Alpha to the libretro architecture"
+DESCRIPTION="Final Burn Alpha is a multi-systems Arcade Emulator most popular for \
+emulating Neo-Geo, Capcom, Konami, and Cave games. It is developed by the Final Burn \
+team and originated from FinalBurn by Dave."
```
fbrosson:
```diff
-DESCRIPTION="Final Burn Alpha is a multi-systems Arcade Emulator most popular for \
+DESCRIPTION="Final Burn Alpha is a multi-systems Arcade Emulator most popular \
-emulating Neo-Geo, Capcom, Konami, and Cave games. It is developed by the Final Burn \
+for emulating Neo-Geo, Capcom, Konami, and Cave games. It is developed by the \
-team and originated from FinalBurn by Dave."
+Final Burn team and originated from FinalBurn by Dave."
```
(We like to have lines of up to 80 chars (including the backslash) but no more.)

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
```diff
@@ -0,0 +1,44 @@
...
+HOMEPAGE="https://github.com/libretro/fbalpha"
+COPYRIGHT="2000-2018 the Final Burn team, the libretro team"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/libretro/fbalpha/archive/099c24b166f08b671f703051de1daffb2e6d5eb7.tar.gz"
```
fbrosson:
```diff
+srcGitRev="099c24b166f08b671f703051de1daffb2e6d5eb7"
-SOURCE_URI="https://github.com/libretro/fbalpha/archive/099c24b166f08b671f703051de1daffb2e6d5eb7.tar.gz"
+SOURCE_URI="https://github.com/libretro/fbalpha/archive/$srcGitRev.tar.gz"
```

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
```diff
@@ -0,0 +1,44 @@
...
+CHECKSUM_SHA256="2ac4a4cf35aaf3380195b4804d9574be33ae52f4f95b7bd3ab8be450de5c8673"
+SOURCE_DIR="fbalpha-099c24b166f08b671f703051de1daffb2e6d5eb7"
```
fbrosson:
```diff
-SOURCE_DIR="fbalpha-099c24b166f08b671f703051de1daffb2e6d5eb7"
+SOURCE_DIR="fbalpha-$srcGitRev"
```

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
```diff
@@ -0,0 +1,44 @@
...
+PATCHES="fbalpha_libretro-$portVersion.patchset"
+
+ARCHITECTURES="x86_64"
```
fbrosson:
```diff
-ARCHITECTURES="x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
```diff
@@ -0,0 +1,44 @@
...
+emulating Neo-Geo, Capcom, Konami, and Cave games. It is developed by the Final Burn \
+team and originated from FinalBurn by Dave."
+HOMEPAGE="https://github.com/libretro/fbalpha"
+COPYRIGHT="2000-2018 the Final Burn team, the libretro team"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/libretro/fbalpha/archive/099c24b166f08b671f703051de1daffb2e6d5eb7.tar.gz"
+CHECKSUM_SHA256="2ac4a4cf35aaf3380195b4804d9574be33ae52f4f95b7bd3ab8be450de5c8673"
+SOURCE_DIR="fbalpha-099c24b166f08b671f703051de1daffb2e6d5eb7"
+PATCHES="fbalpha_libretro-$portVersion.patchset"
```
fbrosson: It would be nice to replace the patchset by a file in `haikuports/games-emulation/fbalpha_libretro/additional-files/` that we could name `fbalpha_libretro.info.in`.
```diff
-PATCHES="fbalpha_libretro-$portVersion.patchset"
+ADDITIONAL_FILES="fbalpha_libretro.info.in"
```
That file would be almost identical to the `fbalpha_libretro.info` file, except for one line:
```diff
-display_version = "v0.2.97.42"
+display_version = "@DISPLAY_VERSION@"
```
Then in `BUILD()` you would only need this change:
```diff
 BUILD()
 {
+	sed -e "s/@DISPLAY_VERSION@/v${portVersion%~.*}/" \
+		$portDir/additional-files/fbalpha_libretro.info.in \
+		> fbalpha_libretro.info
 	make -f makefile.libretro $jobArgs
 }
```

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
```diff
@@ -0,0 +1,44 @@
...
+BUILD()
+{
+	make -f makefile.libretro $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $libDir/libretro
+	cp -a fbalpha_libretro.info $libDir/libretro/
+	cp -a fbalpha_libretro.so $libDir/libretro/
```
fbrosson: Suggestion:
```diff
 INSTALL()
 {
-	mkdir -p $libDir/libretro
+	install -d "$libDir"/libretro
-	cp -a fbalpha_libretro.info $libDir/libretro/
-	cp -a fbalpha_libretro.so $libDir/libretro/
+	install -t "$libDir"/libretro \
+		fbalpha_libretro.info \
+		fbalpha_libretro.so
 }
```

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,60 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	"
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg keep-old
+"
+
+BUILD()
+{
+	./fetch-submodules.sh
```
alaviss: I recommend adding `SOURCE_URI_` for each submodules, as it would allow one to build this package when a proxy is used. This is only a suggestion, so it's not strictly required.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,60 @@
+SUMMARY="The reference frontend of libretro"
+DESCRIPTION="RetroArch is a frontend for emulators, game engines and media players. \
+It enables you to run classic games on a wide range of computers and consoles \
+through its slick graphical interface. It has advanced features like shaders, \
+netplay, rewinding, next-frame response times, and more!"
```
alaviss: `DESCRIPTION` should be wrapped to fit into 80-column.

```diff
-DESCRIPTION="RetroArch is a frontend for emulators, game engines and media players. \
-It enables you to run classic games on a wide range of computers and consoles \
-through its slick graphical interface. It has advanced features like shaders, \
-netplay, rewinding, next-frame response times, and more!"
+DESCRIPTION="RetroArch is a frontend for emulators, game engines and \
+media players. It enables you to run classic games on a wide range of \
+computers and consoles through its slick graphical interface. It has advanced \
+features like shaders, netplay, rewinding, next-frame response times, and more!"
```

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,60 @@
...
+	cmd:gcc$secondaryArchSuffix
+	"
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg keep-old
+"
+
+BUILD()
+{
+	./fetch-submodules.sh
+	./configure --prefix=$installDestDir --global-config-dir=$settingsDir --bindir=$binDir --datarootdir=$dataRootDir --docdir=$docDir --mandir=$manDir --with-assets_dir=$dataDir
```
fbrosson:
```diff
+PATCH()
+{
+	libretroDir="`finddir B_SYSTEM_DIRECTORY`/$relativeLibDir/libretro"
+	retroarchDataDir="`finddir B_SYSTEM_DATA_DIRECTORY`/retroarch"
+	sed -i \
+		-e "s|^# \(libretro_directory =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(libretro_info_path =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(content_directory =\)|\1 \"$retroarchDataDir/\"|;" \
+		retroarch.cfg
+}
+
 BUILD()
 {
-	./fetch-submodules.sh
-	./configure --prefix=$installDestDir --global-config-dir=$settingsDir --bindir=$binDir --datarootdir=$dataRootDir --docdir=$docDir --mandir=$manDir --with-assets_dir=$dataDir
+	runConfigure --omit-dirs "binDir libExecDir infoDir" ./configure \
+		--bindir="$appsDir" --global-config-dir="$settingsDir" \
+		--with-assets_dir=$dataDir
 	make $jobArgs
 }
```
Regarding the `fetch-submodules.sh` script, it cannot be called by haikuporter because all the downloads need to be done by haikuporter itself.
This is going to be a bit complex, but we'll need to create as many `SOURCE_URI_2`, `SOURCE_URI_3`, etc... and their `CHECKSUM_SHA256_2`, `CHECKSUM_SHA256_3`, etc.. as there are submodules to download. Then we will have to use `$sourceDir2`, `$sourceDir3`, etc... to either create symlinks or copy these directories to the build directory.
(I'll guide you with this.)

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,60 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	addAppDeskbarSymlink $binDir/retroarch RetroArch
```
fbrosson:
```diff
-	addAppDeskbarSymlink $binDir/retroarch RetroArch
+	addAppDeskbarSymlink "$appsDir"/retroarch RetroArch
+	install -T retroarch.cfg $dataDir/retroarch/retroarch.cfg.default
```

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,60 @@
...
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:which
+	cmd:git
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	"
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg keep-old
+"
```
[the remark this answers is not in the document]
alaviss: Please note that `USER_SETTINGS_FILES` is only used for files that are not included in the package. Consult [BuildingPackages#The .PackageInfo][0] for more details

>      

> `user-settings-files`: A list of user settings file infos. The list elements must have the following format:
>
>       user_settings_file_info     ::= path [ "directory" | "template" template_path ]
>
> `path` is the relative path of the settings file or directory, starting with "settings/". It is not included in the package. However, if `template_path` is specified, it is a path to a file included in the package that can serve as a template for the settings file. It doesn't imply any automatic action on package activation, though. If the "directory" keyword is given, the path refers to a settings directory (typical when a program creates multiple settings files).


[0]: https://github.com/alaviss/haiku/blob/wiki-fixes/docs/develop/packages/BuildingPackages.rst#the-packageinfo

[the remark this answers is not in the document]
fbrosson: I might be wrong, but what I have understood from the recipes I've seen so far is that `USER_SETTINGS_FILES` is used to inform pkgman that a given config in `/system/settings/` has a a read-only template. Here we could use `/system/data/retroarch/retroarch.cfg.default` for that default.
BTW, I did not kick a build yet, so I don't know if `INSTALL()` will install the `retroarch.cfg`. In case it does not, then all we need to do is add a `install -t "$settingsDir" retroarch.cfg` in `INSTALL()`.

If we're not allowed to use `USER_SETTINGS_FILES` for config files that we install, then there are cannot provide templates for them. My undestanding is that the template provides sane defaults that the user can copy in case the config does not work.

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
kwyxz: My mistake, I thought this only applied to the text within the description. Will fix that.

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
kwyxz: Oh that's fancy. Thanks.

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git.recipe
kwyxz: Sure. Will do.

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
kwyxz: Cool, thanks. Out of curiosity, is `runConfigure` the preferred option still when the `./configure` script was not generated by autotools? RetroArch's `./configure` script is entirely custom.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,60 @@
...
+	app:retroarch$secondaryArchSuffix = $portVersion
+	cmd:retroarch$secondaryArchSuffix = $portVersion
+	cmd:retroarch_cg2glsl$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libsdl2_2.0$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	ffmpeg$secondaryArchSuffix
```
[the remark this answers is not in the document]
kwyxz: This is legacy from the old retroarch recipe, I will look into that.

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
kwyxz: Hello folks,

To make things clear, `/system/settings/retroarch.cfg` is indeed a read-only skeleton file, used by retroarch upon its first run as a template to generate a user-specific settings file in `$XDG_CONFIG_HOME/retroarch/retroarch.cfg`

games-emulation/retroarch/retroarch-1.7.1.recipe
kwyxz: I hesitated between $binDir and $appDir for that one. Is there a fundamental difference? The docs I've read were a bit vague on this.

kwyxz: Also I'm not sure about the .default for the default cfg, as this file really is meant to be read by retroarch as a template when no user-specific cfg file exists.

games-emulation/retroarch/retroarch-1.7.1.recipe
kwyxz: Sure, I'll start looking into this and will come back to you if I can't figure it out :D

kwyxz: Also to keep it simple we could also just package these separately.

games-emulation/retroarch/retroarch-1.7.1.recipe
kwyxz: Seems like replacing addAppDeskbarSymlink by $appsDir breaks the Menu link (even though I replaced --bindir properly)

games-emulation/retroarch/retroarch-1.7.1.recipe
fbrosson: No problem:)
BTW, in case you have a look at [haikuports/net-im/telegram-desktop](https://github.com/haikuports/haikuports/tree/master/net-im/telegram-desktop) please use `srcGitRev_2`, `srcGitRev_3`, etc... instead of `COMMIT`, which we try to avoid because we have far more recipes with `srcGitRev` than with `COMMIT`.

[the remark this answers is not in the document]
kwyxz: Here's how it looks like now
````
runConfigure --omit-dirs "binDir dataDir includeDir infoDir libDir \
                libExecDir localStateDir oldIncludeDir sbinDir sharedStateDir \
                sysconfDir" ./configure --bindir="$appsDir" \
                --global-config-dir="$settingsDir" \
                --with-assets_dir="$dataDir"
````

File: games-emulation/retroarch/retroarch-1.7.1.recipe (hunk 10 above)
Comment (fbrosson) [reply games-emulation/retroarch/retroarch-1.7.1.recipe]: Having a template config in `/system/data/retroarch/` is not really required, but it can be handy in case the user makes changes to `/system/settings/retroarch.cfg` and forgets to make a backup of it. (Although the app does not modify `/system/settings/retroarch.cfg`, the user can still change it, and even break it.)
So the template config serves both as a backup config for 1.7.1 and as a reference when 1.7.2 becomes available, so that the auto-merge feature can compute the diff between the template config of 1.7.2 vs 1.7.1 and apply that diff to the user's config in `/system/settings/`.
(Of course if the template config does not change between 1.7.1 and 1.7.2 then the `auto-merge` will behave like a `keep-old`.)

Regarding the `$appsDir` vs `$binDir` the rule is that command line runtimes (with no GUI) should be in `$binDir`, while GUI programs are better in `$appsDir`. But I guess there can be exceptions.
BTW, `/boot/system/apps` is in PATH, so it is even possible to start retroarch from the command line.

Comment (kwyxz) [reply games-emulation/retroarch/retroarch-1.7.1.recipe]: Sounds great, thanks. I've fixed most of the issues reported, I'm now fiddling with the extra SOURCE_URIs to replace the fetch script.

**Comment by kwyxz** [comment]
Done changes, @alaviss @fbrosson please let me know what else could be done better :) thanks!

File: games-emulation/retroarch/retroarch-1.7.1.recipe [hunk 13]
```diff
@@ -0,0 +1,87 @@
...
+		-e "s|^# \(libretro_info_path =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(content_directory =\)|\1 \"$retroarchDataDir/\"|;" \
+		-e "s|^# \(assets_directory =\)|\1 \"$retroarchDataDir/assets/\"|;" \
+		retroarch.cfg
+}
+
+BUILD()
+{
+	ln -sfn $sourceDir2/retroarch-assets-$srcGitRev_2 media/assets
+	runConfigure --omit-dirs "binDir dataDir includeDir infoDir libDir libExecDir localStateDir oldIncludeDir sbinDir sharedStateDir sysconfDir" \
```
fbrosson: Hmm, after all, it might be better to avoid the `runConfigure`.
So, here are the 2 possibilities:
* With `runConfigure`:
```
	runConfigure --omit-dirs "binDir dataDir includeDir infoDir libDir \
		libExecDir localStateDir oldIncludeDir sbinDir sharedStateDir \
		sysconfDir" \
		./configure --bindir="$appsDir" \
		--global-config-dir="$settingsDir" \
		--with-assets_dir="$dataDir"
```
* Without `runConfigure`:
```
	./configure --prefix="$prefix" --global-config-dir="$settingsDir" \
		--bindir="$appsDir" --datarootdir="$dataRootDir" \
		--docdir="$docDir" --mandir="$manDir" \
		--with-assets_dir="$dataDir"
```
Yep, I think the second option is slightly better. What do you think?

Note: I haven't tested the recipe yet, so there could be errors I haven't noticed.

kwyxz: For what it's worth, I also believe that, since RetroArch's ./configure is entirely custom and not using autotools at all, the second option is simpler hence slightly better.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,87 @@
...
+BUILD_PREREQUIRES="
+        cmd:gcc$secondaryArchSuffix
+        cmd:git
+	cmd:make
+        cmd:pkg_config$secondaryArchSuffix
+	cmd:which
+	"
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg keep-old
+"
```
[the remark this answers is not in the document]
kwyxz: I'm getting this upon package install:
````
[system] warning: package retroarch-1.7.1-1-x86_64.hpkg: "/boot/system/settings/retroarch.cfg" cannot be updated automatically, since it doesn't have a SYS:PACKAGE attribute. Please perform the update manually if needed.
````
Is it normal? It's just a warning so I don't worry much, but just want to make sure it's fine.

**Comment by kwyxz** [comment]
Pushed new commit following recommended changes, only left runConfigure as is, until @alaviss gives his opinion. Thanks again for taking the time to review this. Hopefully when I start packaging the 30+ cores I will know what I'm doing :p

File: games-emulation/retroarch/retroarch-1.7.1.recipe [hunk 15]
```diff
@@ -0,0 +1,91 @@
...
+REVISION="1"
+SOURCE_URI="https://github.com/libretro/RetroArch/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="e9189c7da6b2ca0142528eca14f26104ee322f052f8609203d90b60938d8d3f3"
+SOURCE_FILENAME="retroarch-$portVersion.tar.gz"
+SOURCE_DIR="RetroArch-$portVersion"
+
+srcGitRev_2="505ad6205e5fd8a155daad02ab39953a0b15b0a9"
+SOURCE_URI_2="https://github.com/libretro/retroarch-assets/archive/$srcGitRev_2.tar.gz"
+CHECKSUM_SHA256_2="de37772b00b755b80aafe5d4df0713c0fe6b0e5a79017d94e43f279e798aa85d"
+SOURCE_FILENAME_2="retroarch-assets-$srcGitRev_2.tar.gz"
```
fbrosson: Oops, you ate the `-$portVersion` :)
Well, this is OK as is, but what I'd recomment is:
```diff
-SOURCE_FILENAME_2="retroarch-assets-$srcGitRev_2.tar.gz"
+SOURCE_FILENAME_2="retroarch-assets-$portVersion-$srcGitRev_2.tar.gz"
```
I agree that this is not required, but it is handy if one wants to manually remove an outdated tarball, since it is much easier fo figure out which tarballs are outdated and which are not. And this (currently) only adds 6 chars to the name of the tarball.

Everything else looks OK, except that I forgot to point that we have recently chosen to recommend putting `GLOBAL_WRITABLE_FILES` and `USER_SETTINGS_FILES` **_before_** `PROVIDES`:
```diff
 SECONDARY_ARCHITECTURES="x86"
+
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg auto-merge
+	"
+USER_SETTINGS_FILES="
+	$relativeSettingsDir/retroarch.cfg template \
+		$relativeDataDir/retroarch/retroarch.cfg.default
+	"
+
 PROVIDES="
```
We have several recipes which have them after `BUILD_PREREQUIRES`, but we have many more which have it before `PROVIDES`. So now, everytime we can "fix" this without too much git noise, we try to do so :)

[HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering)

fbrosson: This topic a not easy, and I'm not even sure I already understood it correctly :)
* [`GLOBAL_WRITABLE_FILES`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#global_writable_files) is used to let the system know that a package may create files in `/system/settings/` (or any subdirectory of it.) Mentioning a config file here does not mean the package comes with that file. It only means the package may create that config file.
* [`USER_SETTINGS_FILES`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#user_settings_files) is used when the package comes with a config file, and also allows to specify the default config. BTW, if the `template` keyword is used then when a package is installed for the very first time, the config should be identical to the template, which is generally in a subdirectory of `$dataDir/` (i.e. `/system/data/<subdir>/`).

kwyxz:
@alaviss ?

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42~git.recipe
```diff
@@ -0,0 +1,50 @@
...
+for emulating Neo-Geo, Capcom, Konami, and Cave games. It is developed by the \
+Final Burn team and originated from FinalBurn by Dave."
+HOMEPAGE="https://github.com/libretro/fbalpha"
+COPYRIGHT="2000-2018 the Final Burn team, the libretro team"
+LICENSE="GNU GPL v3"
+REVISION="1"
+srcGitRev="099c24b166f08b671f703051de1daffb2e6d5eb7"
+SOURCE_URI="https://github.com/libretro/fbalpha/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="2ac4a4cf35aaf3380195b4804d9574be33ae52f4f95b7bd3ab8be450de5c8673"
+SOURCE_FILENAME="fbalpha-$portVersion~git.tar.gz"
```
fbrosson:
```diff
-SOURCE_FILENAME="fbalpha-$portVersion~git.tar.gz"
+SOURCE_FILENAME="fbalpha-$portVersion-$srcGitRev.tar.gz"
```
(The `~git` is already in `$portVersion` so we don't need to add another one here.)

fbrosson:
I think the :+1: @alaviss added to [my comment](#discussion_r170481350) applies to this sentence:

> Yep, I think the second option is slightly better. What do you think?

But let's wait for his confirmation.

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
fbrosson: I guess it's not normal, but I'm not sure because I never got this message.
It might as well be a consequence of an incorrect settings for `GLOBAL_WRITABLE_FILES` and/or `USER_SETTINGS_FILES`.

Did you install the hpkg with `pkgman install retroarch-1.7.1-1-x86_64.hpkg` or did you copy the hpkg to `/system/packages/` ?
Both work, but I prefer using pkgman (instead of cp) because when the installation requires older versions to be uninstalled or when there are conflicts, pkgman asks the user what to do on the console, whereas if you install by copying the hpkg, the questions are displayed in the GUI.

Also, you cannot re-install the exact same version&revision of a package if it is already installed.
If there is no revision bump, then you need to uninstall the previous hpkg (with `pkgman uninstall retroarch`) and then re-install your new hpkg.

A workaround that I use for packages which I cannot uninstall (without having to accept to uninstall all the packages that depend on it) is that I keep a copy of the previous hpkg. So I first install an old hpkg and then I install the new one.
(BTW, there is currently no way (to my knowledge) to have pkgman replace a package by a new one with the same version and revision. We can cheat with `touch -r` and `cp -p` but after this we need to reboot.)

gouchi:
@kwyxz @fbrosson @alaviss Thank you !

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42~git.recipe
kwyxz: Ok, I'm still trying to understand what the policy is here :D I guess we can use only ~git and update the package revision if we ever refresh it with the same version but a more recent git revision.

fbrosson: We use `~git.recipe` (instead of `.recipe`) whenever we have a recipe which uses a srcGitRev to point to some git revision that is not available in a stable release.

When you run `haikuporter fbalpla_libretro-0.2.97.42~git` (or just  `haikuporter fbalpla_libretro` if that's the most recent version), haikuporter defines several shell variables, among which you have portVersion which is defined to the string `0.2.97.42~git`. This string is the name of the recipe file without the `.recipe` suffix and without every letter from the beginning up to the first dash. (It drops the `fbalpla_libretro-`.)
That's why the `~git` becomes part of the `$portVersion`.
Here, with a `SOURCE_FILENAME="fbalpha-$portVersion-$srcGitRev.tar.gz"` the tarball will be called `fbalpha-0.2.97.42~git-099c24b166f08b671f703051de1daffb2e6d5eb7.tar.gz`.

fbrosson: > I guess we can use only ~git and update the package revision if we ever refresh it with the same version but a more recent git revision.

Exactly.

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
kwyxz: I installed it with `pkgman` and I believe I have been getting this message because `retroarch` was already installed.
Good to know however that `pkgman` won't overwrite an already existing package with the same revision.
I just tried again and here goes:
````
~> pkgman uninstall retroarch
  in system:
    uninstall package retroarch-1.7.1-3
Continue? [yes/no] (yes) : yes[system] Applying changes ...
[system] Changes applied. Old activation state backed up in "state_2018-02-27_16:05:50"
[system] Cleaning up ...
[system] Done.
~> pkgman install /boot/home/haikuports/packages/retroarch-1.7.1-1-x86_64.hpkg
100% repochecksum-1 [65 bytes]
Validating checksum for Haiku...done.
100% repocache-2 [3.65 KiB]
Validating checksum for Haiku...done.
100% repochecksum-1 [64 bytes]
Validating checksum for HaikuPorts...done.
100% repocache-2 [676.07 KiB]
Validating checksum for HaikuPorts...done.
  in system:
    install package retroarch-1.7.1-1 from local file
Continue? [yes/no] (yes) : [system] Applying changes ...
[system] Changes applied. Old activation state backed up in "state_2018-02-27_16:06:21"
[system] Cleaning up ...
[system] Done.
````
Looks like we're good.
Again I would like to reiterate my thanks for all your patient help :)

**Comment by kwyxz** [comment]
Posted commit reverting changes done to `./configure` section. Crossing fingers for this to be the last :D

**Comment by kwyxz** [comment]
NARRATOR: it wasn't.

File: games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42~git.recipe (hunk 16 above)
Comment (kwyxz) [reply games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42~git.recipe]: Hmmm in this case it's a bit different then, here 0.2.97.42 is the latest stable release, but the source code on Git is more recent. Does naming it 0.2.97.42~git still work?

File: games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42~git.recipe [hunk 17]
```diff
@@ -9,8 +9,8 @@ REVISION="1"
 srcGitRev="099c24b166f08b671f703051de1daffb2e6d5eb7"
 SOURCE_URI="https://github.com/libretro/fbalpha/archive/$srcGitRev.tar.gz"
 CHECKSUM_SHA256="2ac4a4cf35aaf3380195b4804d9574be33ae52f4f95b7bd3ab8be450de5c8673"
-SOURCE_FILENAME="fbalpha-$portVersion~git.tar.gz"
-SOURCE_DIR="fbalpha-$srcGitRev"
+SOURCE_FILENAME="fbalpha-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="fbalpha-$portVersion-$srcGitRev"
```
fbrosson: Oops, SOURCE_DIR was correct before your last update.
```diff
-SOURCE_DIR="fbalpha-$portVersion-$srcGitRev"
+SOURCE_DIR="fbalpha-$srcGitRev"
```
(As you know, SOURCE_DIR is the name of the top directory of the tarball, and the tarball we're using here is a snapshot. So no `$portVersion` here. Just the `$srcGitRev` :)

kwyxz: Daaaaaamn ok ok fixing that whoops

fbrosson:
Regarding the modules like the retroarch-assets, if these are just data files, it might be worth creating a separate recipe, e.g. `retroarch_assets-20180226.recipe`, in the same directory as retroarch or in a separate directory.
retroarch_assets would set `ARCHITECTURES="any"` and retroarch would depend on retroarch_assets.

kwyxz:
I thought about this when I stripped down the `./fetch-submodules.sh` script, and the reason why I thought I'd keep the assets with this package (and took out everything else the script fetches) is that upstream's `make install` actually installs them if they're present. Can a recipe be parsed by Haikuports automatically from another one so we can still enforce using the provided `make install`?

fbrosson:
> Can a recipe be parsed by Haikuports automatically from another one

Nope. Although haikuporter can build several recipes at once, it parses them one by one.
What can be done, however, is make retroarch depend on retroarch_assets, so we would first "build" retroarch_assets and then retroarch itself.

What are the contents of the retroarch-assets? Does it need to build anything? Or does it only contain files that need to be copied to some directories?
Would a retroarch build be different with vs without the retroarch-assets?
In other words, is it possible to build retroarch without the assets, and then install the 2 packages (retroarch and retroarch_assets) and obtain the same functionality as if retroarch had been built with the assets right from the beginning?

Well, may we can start with everything in a single recipe for now, as it is easier.

kwyxz:
> Nope. Although haikuporter can build several recipes at once, it parses them one by one.
> What can be done, however, is make retroarch depend on retroarch_assets, so we would first "build" retroarch_assets and then retroarch itself.

Ok. I am running some tests right now, if I can find a solution that works and that does not look too horrible I will submit yet another commit :p 

> What are the contents of the retroarch-assets? Does it need to build anything? Or does it only contain files that need to be copied to some directories?
> Would a retroarch build be different with vs without the retroarch-assets?
> In other words, is it possible to build retroarch without the assets, and then install the 2 packages (retroarch and retroarch_assets) and obtain the same functionality as if retroarch had been built with the assets right from the beginning?

It's totally possible to build retroarch without the assets and then copy them later. They are only a bunch of PNG icons. I did ask the Libretro team last week if they'd rather have the assets provided with the same package or in a separate one, and they felt like having the assets already installed is much better for the user (basically it shows icons on the RetroArch UI, while without the assets they're only ugly black squares). It's a commodity, but a rather important one.

However having a dependency on retroarch_assets from retroarch is a solution to that. So that would work for everyone :)

kwyxz:
I just found out I had a basic misunderstanding of what SOURCE_DIR does. Will need to review this in both fbalpha and retroarch recipes.

fbrosson:
There is some Haiku-specific things to keep in mind if we write a separate recipe for the assets.
It's that the prefix of each package is unique to that package.
This means retroarch would probably need to use something we call _Package Links_.
(You can search for portPackageLinksDir in work-1.7.1/wrapper-script.)

To get an idea of how this works, you can look at the recipe we have for curl. It depends on another package, ca_root_certificates, and uses `$portPackageLinksDir` in `BUILD()` to "compute" the path to the directory that holds the files from the ca_root_certificates package.

There is also a simpler method: we can tell retroarch that the files it needs to read are in "`finddir B_SYSTEM_DATA_DIRECTORY`/retroarch/` instead of `$dataDir/retroarch/`. Using `$dataDir/retroarch/` would not allow retroarch to access the files from retroarch_assets, while the "`finddir B_SYSTEM_DATA_DIRECTORY`/retroarch/` would return `/boot/system/data/retroarch/` and would probably be OK.

BTW, I don't know if, in some situations, it is required to use the Package Links method.

kwyxz:
I splitted the assets into their own recipe. @fbrosson what do you think? Thanks :)

games-emulation/retroarch/retroarch_assets-20180129.recipe
```diff
@@ -0,0 +1,28 @@
...
+PROVIDES="
+	retroarch_assets = $portVersion
+	"
+
+INSTALL()
+{
+	retroarchDataDir="$dataDir/retroarch/assets/"
+	mkdir -p "$retroarchDataDir"
+	cp -ar * "$retroarchDataDir"
+	rm -rf "$retroarchDataDir"/Makefile \
```
fbrosson: Since the tarball is quite big (around 86 MiB), I'd like to suggest a small variant, but I don't know if it is acceptable because it does something (`rm ../../flag.unpack ../../flag.build`) which violates an implicit rule. This variant works like a charm, but we need to wait for the green lights from at least a few key contributors.

```diff
 PROVIDES="
 	retroarch_assets = $portVersion
 	"
+REQUIRES="
+	haiku
+	"

 INSTALL()
 {
 	retroarchDataDir="$dataDir/retroarch/assets/"
 	mkdir -p "$retroarchDataDir"
-	cp -ar * "$retroarchDataDir"
-	rm -rf "$retroarchDataDir"/Makefile \
-		"$retroarchDataDir"/configure \
-		"$retroarchDataDir"/src
+	rm ../../flag.unpack ../../flag.build Makefile configure
+	rm -rf src
+	mv * "$retroarchDataDir"
 }
```

If this variant is not acceptable, then the only changes I see are:
* adding a `BUILD()` that would `rm configure Makefile` and `rm -rf src`.
* adjusting `INSTALL()` accordingly.
* Adding `REQUIRES` as suggested above.
* addressing the `LICENSE` topic... which does not currently mention the real licenses used by the assets.

kwyxz: The actual license is CC-BY-4.0 according to github, which is not in the list of included licenses :(

````
Error: Error: No match found for license CC-BY-4.0
Error: Error: Valid license filenames included with Haiku are:
Error: Error: Anti-Grain Geometry, Apache v2, Artistic, Artistic v2.0, BSD (2-clause), BSD (3-clause), BSD (4-clause), Be Sample Code License, Bitstream Charter, CDDL v1, CQuantizer, Curl, DEC, Flex, FreeType, GNU GPL font exception, GNU GPL v1, GNU GPL v2, GNU GPL v3, GNU LGPL v2, GNU LGPL v2.1, GNU LGPL v3, ICU, Info-ZIP, Intel (2xxx firmware), Intel (ACPICA), Intel (firmware), JPEG, Jam, JasPer v2, LibJPEG, LibPNG, MAPM, MIT, MIT (no promotion), MPL v1.1, MPL v2.0, Marvell (firmware), OpenGroup, OpenSSH, OpenSSL, PCRE, Public Domain, Python, Ralink (firmware), SGI Free B, SIL Open Font License v1.1, SQLite, Speex, UIUC, Vim, WebKit Apple, Zlib, bzip2
````

Comment (kwyxz) [reply games-emulation/retroarch/retroarch_assets-20180129.recipe]: According to [this page](https://wiki.creativecommons.org/wiki/ShareAlike_compatibility) CC-BY-4.0 is one-way compatible with the GNU GPL v3 (see the bottom of the page) so we should be fine with the GPL v3.

Comment (kwyxz) [reply games-emulation/retroarch/retroarch_assets-20180129.recipe]: I kept CC-BY-4.0 commented out to reflect this in my latest commit.

Comment (fbrosson) [reply games-emulation/retroarch/retroarch_assets-20180129.recipe]: Yep, and that's not funny...
Whenever a recipe has one or more licenses that are not included in `haiku-r1~alpha4_pm_hrev*-*-*.hpkg`, the missing licenses need to be added to the recipe's `licenses` subdirectory.
So, for retroarch_assets, we need to put in `haikuports/games-emulation/retroarch/licenses/` a copy of the COPYING file (from the retroarch-asset tarball) renamed to CC-BY-4.0.
BTW, since we'll have `LICENSE="CC-BY-4.0"`, the license will be available  in `/boot/system/data/licenses/` (when the package is installed) so we can remove the COPYING file before the installation to `$dataDir/retroarch/` (just like we removed the Makefile and configure files.)

But if, in the future, the `CC-BY-4.0` file is added to haiku, then it will be OK to remove it from `haikuports/games-emulation/retroarch/licenses/`.
However, this also means that the package would no longer contain a copy of the license, so, after all, it might be better to keep the COPYING file.
(BTW, we really need to decide what to do with all these popular licenses (e.g. GPL), because we are not embedding them in the package.)

Comment (kwyxz) [reply games-emulation/retroarch/retroarch_assets-20180129.recipe]: Ok I've made a new push following these recommendations

File: games-emulation/retroarch/retroarch_assets-20180129.recipe [hunk 19]
```diff
@@ -0,0 +1,35 @@
...
+		"$retroarchDataDir"/configure \
+		"$retroarchDataDir"/COPYING \
+		"$retroarchDataDir"/src
+}
+
+INSTALL()
+{
+	retroarchDataDir="$dataDir/retroarch/assets/"
+	mkdir -p "$retroarchDataDir"
+	cp -ar * "$retroarchDataDir"
```
fbrosson: Hmm, I think my idea of adding a `BUILD()` to remove some files and the src tree was not bright :)
That's because removing source files in the build directory might have side effect on the `--extract-patchset` feature of haikuporter.

So, since you made a small mistake in your current `BUILD()` (`retroarchDataDir` is not defined, so the rm would not delete any files), it gives me a chance to suggest another variant:
```diff
-BUILD()
-{
-	rm -rf "$retroarchDataDir"/Makefile \
-		"$retroarchDataDir"/configure \
-		"$retroarchDataDir"/COPYING \
-		"$retroarchDataDir"/src
-}
-
 INSTALL()
 {
-	retroarchDataDir="$dataDir/retroarch/assets/"
+	retroarchDataDir="$dataDir"/retroarch/assets
 	mkdir -p "$retroarchDataDir"
-	cp -ar * "$retroarchDataDir"
+	# copy everything but configure, COPYING, Makefile and src
+	cp -ar [abd-rt-z]* "$retroarchDataDir"
+	rm -f "$retroarchDataDir"/xmb/convert.sh
 }
```

fbrosson:
@kwyxz, in case stables releases of retroarch-assets become available, and if the version scheme chosen for it is roughly synchronised with the versions of retroarch, we would run into trouble if we keep retroarch_assets in `games-emulation/retroarch`.
So maybe it would be better to move `retroarch_assets-20180129.recipe` to `games-emulation/retroarch-assets/` (as well as `CC-BY-4.0` to `games-emulation/retroarch-assets/licenses/`, of course.)
(Yep, dashes are allowed in the names of the directories. The only place where they are forbidden is in the "basename" of the recipe files.)

Directory for the  retroarch_assets recipe | Pros | Cons
-- | -- | --
`games-emulation/retroarch/` | same dir as the retroarch recipe | risky if stable releases of retroarch-assets become available
`games-emulation/retroarch-assets/` | 100% safe | separate directories

Shall we choose the safe directory before the merge?
(Sorry for not having though about this before the split...)

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42~git.recipe
fbrosson: There is currently no tarball for v0.2.97.42 although we can see this commit: [Bump to v0.2.97.42](https://github.com/libretro/fbalpha/commit/ed1e7061cb4500ac57b3f7aa4245de66b28c26e9).

This means we should either :
* drop the `~git` (in the name of the recipe) but keep the srcGitRev and increment `REVISION` everytime we update srcGitRev, until 0.2.97.43 becomes available, or
* rename the recipe to `fbalpha_libretro-0.2.97.43~git.recipe`.

> Does naming it 0.2.97.42~git still work?

As you suspected, it's not recommended to keep the `~git` (after `0.2.97.42` in the name of the recipe) if that version is already officially out.
It would work, but if a stable tarball for v0.2.97.42 is made available with contents older than the git revision we have in srcGitRev then it would be misleading, as we would be using something newer than the official realease.

Recipe name | Pros | Cons
-- | -- | --
fbalpha_libretro-0.2.97.42~git.recipe | - | _"Hey, HaikuPorts thinks `v0.2.97.42` is not yet available!!!"_
fbalpha_libretro-0.2.97.42.recipe | _"HaikuPorts knows `v0.2.97.42` is available."_ | _"Is this really `v0.2.97.42` or is it `v0.2.97.43~git`?"_
fbalpha_libretro-0.2.97.43~git.recipe | More appropriate. | _"Why does the app reports `v0.2.97.43` if that release is not out?"_

Well, we could choose `fbalpha_libretro-0.2.97.43~git.recipe` and change `BUILD()` a little to have the app report `v0.2.97.43~git`(instead of `v0.2.97.43`):
```diff
-	sed -e "s/@DISPLAY_VERSION@/v${portVersion%~.*}/" \
+	sed -e "s/@DISPLAY_VERSION@/v$portVersion/" \
 		$portDir/additional-files/fbalpha_libretro.info.in \
 		> fbalpha_libretro.info
```
What do you think?

games-emulation/retroarch/retroarch_assets-20180129.recipe
kwyxz: My only issue with this variant is that it will need to be updated if there are extra folders added by upstream in the media/assets directory that start with letters c or s.

kwyxz:
I'm fine with having retroarch_assets in their own directory, most of the work is already done.
Since we're starting to have a lot of commits, should I close this PR and rebase? This will also allow me to create separate PRs for all recipes.

kwyxz:
Regarding versioning and git... I am not 100% comfortable with using versions that are ahead of upstream. I've looked at what Debian does, and it seems like they use "+" when they want to indicate that the git snapshot is actually more current than the stable version (while ~ would indicate the git snapshot, as you said, predates the version). They seem to agree that versioning should be upstream's job, so unless there is a mutual agreement with upstream on what version is going to be next, the package maintainer should not make assumptions : https://lists.debian.org/debian-mentors/2011/03/msg00326.html

games-emulation/retroarch/retroarch_assets-20180129.recipe
fbrosson: You are right. Your first `INSTALL()` is better. Even if it copies src and then removes it, that's still better than having unneeded patterns.

Regarding the `xmb/convert.sh` shell script, which converts some svg to png, we don't have inkscape but we might, some day, find some way to use [media-libs/libsvgtiny](https://github.com/haikuports/haikuports/tree/master/media-libs/libsvgtiny) for doing the conversion.
I guess this conversion should be done at build time.
Well, I guess retroarch does complain if it only finds the svg.

fbrosson:
That's a good idea, to use a +. We can probably use something like `fbalpla_libretro-0.2.97.42+20180224.recipe` with the current srcGitRev or `fbalpla_libretro-0.2.97.42+20180227.recipe` if you update srcGitRev to the latest upstream git revision.

Regarding the rebase and PR split, that's up to you.
BTW, we also like to have separate PRs for each recipe, but that's also OK to use one PR for recipes that are related.
So, if you would like to close this PR and open 3 new PRs, that's perfectly OK for me.

BTW, shall we use dashes (instead of underscores) in the names of the directories?
* games-emulation/retroarch/retroarch-1.7.1.recipe
* games-emulation/retroarch-assets/retroarch_assets-20180129.recipe
* games-emulation/fbalpha-libretro/fbalpha_libretro-0.2.97.42+20180224.recipe

games-emulation/retroarch/retroarch_assets-20180129.recipe
kwyxz: I will revert the changes to INSTALL()
Regarding convert.sh it's not necessary to retroarch. It's a convenience script for icon designers, that is not needed by the end user. Same goes with NPMApng2PMApng.py and I will have them removed as well.

kwyxz:
Regarding directory naming I'm fine with using dashes. Upstream does not have any strict naming (ie the fbalpha libretro core repository is simply called "fbalpha" while the mame2010 one is "mame2010-libretro" and fceumm is "libretro-fceumm") however the libretro-super script (where I'm pulling the info files from) does try to name things in a more coherent way (libretro-*). 

I'll use dashes with my new PRs.

Thanks again for all your patient advice and insight :)

fbrosson:
You're welcome!
Thank _you_ for contributing ;)

PR #2253, opened 2018-02-25
graphviz, bump version, fix build

Begasus:


Begasus:
Thanks to @alaviss for the waitpid problem :)

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,165 @@
...
+liblab_gamutVersion="1.0.0"
+libpathplanVersion="4.0.0"
+libxdotVersion="4.0.0"
+libcdtVersionCompat="$libcdtVersion compat >= ${libcdtVersion%%.*}"
+libcgraphVersionCompat="$libcgraphVersion compat >= ${libcgraphVersion%%.*}"
+libgvcVersionCompat="$libgvcVersion compat >= ${libgvcVersion%%.*}"
+libgvprVersionCompat="$libgvprVersion compat >= ${libgvprVersion%%.*}"
+liblab_gamutVersionCompat="$liblab_gamutVersion compat >= ${liblab_gamutVersion%%.*}"
+libpathplanVersionCompat="$libpathplanVersion compat >= ${libpathplanVersion%%.*}"
+libxdotVersionCompat="$libxdotVersion compat >= ${libxdotVersion%%.*}"
```
fbrosson: Suggesion:
```diff
+graphvizLibs="\
+	libcdt \
+	libcgraph \
+	libgvc \
+	libgvpr \
+	liblab_gamut  \
+	libpathplan \
+	libxdot \
+	"
 libcdtVersion="5.0.0"
 libcgraphVersion="6.0.0"
 libgvcVersion="6.0.0"
 libgvprVersion="2.0.0"
 liblab_gamutVersion="1.0.0"
 libpathplanVersion="4.0.0"
 libxdotVersion="4.0.0"
-libcdtVersionCompat="$libcdtVersion compat >= ${libcdtVersion%%.*}"
-libcgraphVersionCompat="$libcgraphVersion compat >= ${libcgraphVersion%%.*}"
-libgvcVersionCompat="$libgvcVersion compat >= ${libgvcVersion%%.*}"
-libgvprVersionCompat="$libgvprVersion compat >= ${libgvprVersion%%.*}"
-liblab_gamutVersionCompat="$liblab_gamutVersion compat >= ${liblab_gamutVersion%%.*}"
-libpathplanVersionCompat="$libpathplanVersion compat >= ${libpathplanVersion%%.*}"
-libxdotVersionCompat="$libxdotVersion compat >= ${libxdotVersion%%.*}"
+for i in $graphvizLibs; do
+	eval "${i}VersionCompat=\"\$${i}Version compat >= \${${i}Version%%.*}\""
+done
```

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,165 @@
...
+	make install
+
+	#remove libtool files
+	rm $libDir/*.la
+	rm $libDir/graphviz/*.la
+
+	prepareInstalledDevelLibs libcdt \
+		libcgraph \
+		libgvc \
+		libgvpr \
```
fbrosson:
```diff
-	prepareInstalledDevelLibs libcdt \
-		libcgraph \
-		libgvc \
-		libgvpr \
-		liblab_gamut  \
-		libpathplan \
-		libxdot
+	prepareInstalledDevelLibs $graphvizLibs
```

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,165 @@
...
+		liblab_gamut  \
+		libpathplan \
+		libxdot
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
```
fbrosson:
```diff
 	packageEntries devel \
-		$developDir
+		$developDir \
+		$manDir/man3
```
BTW, I looks like there are also some docs that could be moved to `$developDocDir`.

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: Had been looking into that, wasn't quite ready with that yesterday so I thought I'd post what is there already :)

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: Yep, need to look into that too :)

media-gfx/graphviz/patches/graphviz-2.40.1.patchset
```diff
@@ -0,0 +1,120 @@
...
+--- a/lib/sfio/sfhdr.h
++++ b/lib/sfio/sfhdr.h
+@@ -906,7 +906,9 @@ extern "C" {
+ #endif
+ 
+     extern time_t time(time_t *);
+-    extern int waitpid(int, int *, int);
++#ifndef waitpid
++    extern pid_t waitpid(pid_t, int *, int);
++#endif
```
alaviss: Did you merge both patches together? Only one of them should be necessary.

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,165 @@
...
+	cmd:bcomps$secondaryArchSuffix
+	cmd:ccomps$secondaryArchSuffix
+	cmd:circo$secondaryArchSuffix
+	cmd:cluster$secondaryArchSuffix
+	#cmd:diffimg$secondaryArchSuffix
+	cmd:dijkstra$secondaryArchSuffix
+	cmd:dot_builtins$secondaryArchSuffix
+	cmd:dot$secondaryArchSuffix
+	cmd:dot2gxl$secondaryArchSuffix
+	#cmd:dotty$secondaryArchSuffix
```
alaviss:
```diff
-	#cmd:dotty$secondaryArchSuffix

File: media-gfx/graphviz/graphviz-2.40.1.recipe [hunk 6]
```diff
@@ -0,0 +1,165 @@
...
+	cmd:gv2gxl$secondaryArchSuffix
+	cmd:gvcolor$secondaryArchSuffix
+	cmd:gvgen$secondaryArchSuffix
+	cmd:gvpack$secondaryArchSuffix
+	cmd:gvpr$secondaryArchSuffix
+	cmd:gvmap$secondaryArchSuffix
+	cmd:gvmap.sh$secondaryArchSuffix
+	cmd:gxl2dot$secondaryArchSuffix
+	cmd:gxl2gv$secondaryArchSuffix
+	#cmd:lneato$secondaryArchSuffix
```
alaviss:
```diff
-	#cmd:lneato$secondaryArchSuffix

File: media-gfx/graphviz/graphviz-2.40.1.recipe [hunk 7]
```diff
@@ -0,0 +1,165 @@
...
+libxdotVersionCompat="$libxdotVersion compat >= ${libxdotVersion%%.*}"
+
+PROVIDES="
+	graphviz$secondaryArchSuffix = $portVersion compat >= 2
+	cmd:acyclic$secondaryArchSuffix
+	cmd:bcomps$secondaryArchSuffix
+	cmd:ccomps$secondaryArchSuffix
+	cmd:circo$secondaryArchSuffix
+	cmd:cluster$secondaryArchSuffix
+	#cmd:diffimg$secondaryArchSuffix
```
alaviss:
```diff
-	#cmd:diffimg$secondaryArchSuffix

File: media-gfx/graphviz/graphviz-2.40.1.recipe [hunk 8]
```diff
@@ -0,0 +1,165 @@
...
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:freetype_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	./autogen.sh
+	runConfigure ./configure CFLAGS=-D_BSD_SOURCE \
```
alaviss: `CPPFLAGS` should be used for definitions. They're used by both C and C++ compilers, and would avoid overriding vendor defaults `CFLAGS`

```diff
-	runConfigure ./configure CFLAGS=-D_BSD_SOURCE \
+	runConfigure ./configure CPPFLAGS=-D_BSD_SOURCE \

File: media-gfx/graphviz/graphviz-2.40.1.recipe [hunk 9]
```diff
@@ -0,0 +1,165 @@
...
+bioinformatics,  software engineering, database and web design, machine \
+learning, and in visual interfaces for other technical domains.
+The Graphviz layout programs take descriptions of graphs in a simple text \
+language, and make diagrams in useful formats, such as images and SVG for web \
+pages, PDF or Postscript for inclusion in other documents; or display in an \
+interactive graph browser. (Graphviz also supports GXL, an XML dialect.)
+Graphviz has many useful features for concrete diagrams, such as options for \
+colors, fonts, tabular node layouts, line styles, hyperlinks, rolland custom \
+shapes."
+HOMEPAGE="http://www.graphviz.org/"
```
alaviss:
```diff
-HOMEPAGE="http://www.graphviz.org/"
+HOMEPAGE="https://www.graphviz.org/"

File: media-gfx/graphviz/graphviz-2.40.1.recipe [hunk 10]
```diff
@@ -0,0 +1,165 @@
...
+COPYRIGHT="1994-2010 AT&T Corp."
+LICENSE="CPL v1.0"
+REVISION="1"
+SOURCE_URI="http://www.graphviz.org/pub/graphviz/stable/SOURCES/graphviz.tar.gz"
+CHECKSUM_SHA256="ca5218fade0204d59947126c38439f432853543b0818d9d728c589dfe7f3a421"
+SOURCE_FILENAME="graphviz-$portVersion.tar.gz"
+PATCHES="graphviz-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2"
```
alaviss: Maybe this should be tested for `x86_64` as well?

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,165 @@
...
+The Graphviz layout programs take descriptions of graphs in a simple text \
+language, and make diagrams in useful formats, such as images and SVG for web \
+pages, PDF or Postscript for inclusion in other documents; or display in an \
+interactive graph browser. (Graphviz also supports GXL, an XML dialect.)
+Graphviz has many useful features for concrete diagrams, such as options for \
+colors, fonts, tabular node layouts, line styles, hyperlinks, rolland custom \
+shapes."
+HOMEPAGE="http://www.graphviz.org/"
+COPYRIGHT="1994-2010 AT&T Corp."
+LICENSE="CPL v1.0"
```
alaviss: The license is now [named `EPL v1.0`][0]

[0]: https://gitlab.com/graphviz/graphviz/blob/stable_release_2.40.1/LICENSE#L1

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,165 @@
...
+learning, and in visual interfaces for other technical domains.
+The Graphviz layout programs take descriptions of graphs in a simple text \
+language, and make diagrams in useful formats, such as images and SVG for web \
+pages, PDF or Postscript for inclusion in other documents; or display in an \
+interactive graph browser. (Graphviz also supports GXL, an XML dialect.)
+Graphviz has many useful features for concrete diagrams, such as options for \
+colors, fonts, tabular node layouts, line styles, hyperlinks, rolland custom \
+shapes."
+HOMEPAGE="http://www.graphviz.org/"
+COPYRIGHT="1994-2010 AT&T Corp."
```
alaviss: I think this would need an update.

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: These commands don't seem to be part of the package?

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: Need to push that one :)

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: Haven't checked x86_64 yet ...

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: Leftover, need to change

media-gfx/graphviz/patches/graphviz-2.40.1.patchset
Begasus: So the #ifndef waitpid/#endif shouldn't be needed?

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: PS, in the package there is man1 man3 and man7, would it be ok to use 
$manDir/man{1,3,7} here?

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: My bad, misread these here (those can be removed yes) ...

media-gfx/graphviz/graphviz-2.40.1.recipe
fbrosson: man1 and man7 should should stay in the base package.
The only man section that needs to go to the devel package man3.

media-gfx/graphviz/graphviz-2.40.1.recipe
fbrosson: It builds fine on x86_64, but I did not manage to build on x86_gcc2 (primary arch).
@Begasus, did you succeed in building it on x86_gcc2 primary arch?
I think what we'll need is:
```
```

Begasus: Bugger .. forgot to update that line too, if it works for x86 I'll update it :)

media-gfx/graphviz/graphviz-2.40.1.recipe
Begasus: https://www.graphviz.org/license/ (this one still calls it CPL v1.0)?

fbrosson:
graphviz is failing to build (for me) on x86 second arch. I can only build it on x86_64.

Begasus:
I'll try again this weekend (if I get around to it), but got it up and build earlier this week ...

Begasus:
Still builds ok for me (for x86 on x86_gcc2) ... 
grabbing graphviz_x86-2.40.1-1-x86_gcc2.hpkg and moving it to /boot/home/haikuports/packages/graphviz_x86-2.40.1-1-x86_gcc2.hpkg

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,207 @@
...
+	lib:libfreetype$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libfontconfig$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	graphviz${secondaryArchSuffix}_devel = $portVersion
```
fbrosson: > Still builds ok for me (for x86 on x86_gcc2) ...

@Begasus: I haven't investigated yet why it's not building for me on x86, but since it is building fine for you and since the recipe looks OK (and builds fine for me on x86_64), I'd say we can probably merge it.

Just one thing before the merge. @alaviss, should we add a compat for the _devel package?
```diff
 PROVIDES_devel="
-	graphviz${secondaryArchSuffix}_devel = $portVersion
+	graphviz${secondaryArchSuffix}_devel = $portVersionCompat
```
Thanks!

Begasus:
Just checked vala38 (v 0.38.8) to be sure, builds ok here (for x86 on x86_gcc2) (tests for vala are ok too) ...

media-gfx/graphviz/graphviz-2.40.1.recipe
alaviss: I would prefer not having them at all. As our policy requires explict dependencies declaration, soversion becomes the prefered number to compare against. So unless there exists any port uses this information, I would recommend against it.

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,207 @@
...
+	lib:liblab_gamut$secondaryArchSuffix = $liblab_gamutVersionCompat
+	lib:libpathplan$secondaryArchSuffix = $libpathplanVersionCompat
+	lib:libxdot$secondaryArchSuffix = $libxdotVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libexpat$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
```
fbrosson:
```diff
-	lib:libpng$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
```

media-gfx/graphviz/graphviz-2.40.1.recipe
```diff
@@ -0,0 +1,207 @@
...
+	"
+REQUIRES_devel="
+	graphviz$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libexpat$secondaryArchSuffix
+	devel:libfreetype$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
```
fbrosson:
```diff
-	devel:libpng$secondaryArchSuffix
+	devel:libpng16$secondaryArchSuffix
```

alaviss:
@fbrosson 
> Suggestion: libpng -> libpng16

You wouldn't want to bump the numbers manually when libpng 1.7 is out. Why isn't 1.6 the default version already?

fbrosson:
The problem here is that both libpng and libpng16 provide lib:libpng.
So if someone has chosen to install libpng and libpng_devel, and then build a recipe which depends on libpng then the resulting package will depend on libpng instead of libpng16.
The main reason for having libpng12, libpng(15) and libpng16 is that some programs could build fine against an old version but not against a newer version, so we have been updating 3 parallel branches of it.
BTW, since a few years we have been forcing the dependency on libpng16.
But maybe we could retire libpng(15), I mean, bump libpng to 1.6.x, and make it replace libpng16.

That said, I think it is safer to stick to libpng16 and if any program needs libpng17 then we would use libpng17 instead of libpng16.

alaviss:
We should retire libpng 1.5. As shown by the [ABI tracker](https://abi-laboratory.pro/tracker/timeline/libpng/), we can have high confidence that most programs that builds on libpng 1.5 would also do with libpng 1.6.

pulkomandy:
On Sun, Mar 04, 2018 at 05:28:55AM +0000, alaviss wrote:
> We should retire libpng 1.5. As shown by the [ABI tracker](https://abi-laboratory.pro/tracker/timeline/libpng/), we can have high confidence that most programs that builds on libpng 1.5 would also do with libpng 1.6.

The plan is to retire it eventually, but we need to be sure that nothing
uses it anymore. This includes software not hosted by haikuports (in one
of the various other repos or not in an hpkg), so it is hard to do.

Next time we have a major ABI breakage and need to rebuild everything
anyways may be a good time.

Begasus:
@alaviss there is a test suite but it requires ksh93

mmlr:
Please note that the new graphviz build apparently contains a bug where it does a ```poll``` claiming to supply a huge amount of FDs. Due to a bug in Haiku this currently currently triggers an overflow and KDLs on x86_64. This is the reason for the build failures on the builder. I have identified the problem, but it will take some time to implement and deploy the fix.

fbrosson:
OK, thanks @mmlr for the info.
(My x86_64 VM was running with 6 GiB of RAM, so this is probably the reason why the build was successful :)

Begasus:
I've managed to build it in VBox with 2GiB or RAM (both arches), but maybe it's a good thing it revealed a bug :)

mmlr:
It's an uint32 overflow in create_area_etc() possibly due to uninitialized memory used in the application. It isn't related to the amount of available RAM. It will only KDL on x86_64 because the cleared size is given in size_t which is 64 bits there, while a truncated area with a size of the lower 32 bits is actually created. But even if it is a bug in graphviz, it's indeed always a good thing when this kind of thing is revealed.

mmlr:
Yes, it will probably KDL randomly (I haven't fully investigated the cause, as mentioned I suspect uninitialized memory). So it doesn't just fail the build, but it will actually crash the builder, so it's somewhat of a pressing issue.

PR #2256, opened 2018-02-26
vala, bump to version 0.40.4

Begasus:
Requires graphviz: https://github.com/haikuports/haikuports/pull/2253


dev-lang/vala/patches/vala38-0.38.8.patchset
```diff
@@ -0,0 +1,107 @@
...
+Subject: remove control-flow/bug736774-2.vala test
+
+
+diff --git a/tests/Makefile.am b/tests/Makefile.am
+index 4bae01f..5c490ba 100644
+--- a/tests/Makefile.am
++++ b/tests/Makefile.am
+@@ -127,7 +127,6 @@ TESTS = \
+ 	control-flow/bug665904.vala \
+ 	control-flow/bug691514.vala	\
```
fbrosson:
```diff
- 	control-flow/bug691514.vala	\
+ 	control-flow/bug691514.vala \
```
(near the end of the line: space -> tab)

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+	2006-2018 Raffaele Sandrini"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/vala/${portVersion%.*}/vala-$portVersion.tar.xz"
+CHECKSUM_SHA256="2fa746b51cd66e43577d1da06a80b708c2875cadaafee77e9700ea35cf23882c"
+SOURCE_DIR="vala-$portVersion"
+PATCHES="vala38-$portVersion.patchset"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```
fbrosson:
```diff
-libVersion="0.0.0"
-libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
-
 ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
 SECONDARY_ARCHITECTURES="?x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
```

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libdbus_1$secondaryArchSuffix
+	lib:libffi$secondaryArchSuffix
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libgvc$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	"
+
```
fbrosson: Among the 70 recipes in dev-python, 22 have one or more `REPLACES`, and most of these 22 recipes do have a blank line before their `REPLACES`, but these are special because they follow a `for .. in .. do .. done` to handle our multiple python versions.

If we ignore the dev-python category, all other recipes that have a `REPLACES` do not have an empty line before it.

So I'd say that we don't need a blank line between `REQUIRES` and `REPLACES`, just like we don't have a blank line between `PROVIDES` and `REQUIRES`.

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+	$binDir/vapigen \
+	$binDir/vala-${portVersion%.*} \
+	$binDir/valac-${portVersion%.*} \
+	$binDir/valadoc-${portVersion%.*} \
+	$binDir/vapicheck-${portVersion%.*} \
+	$binDir/vapigen-${portVersion%.*}
+
+PATCH()
+{
+		sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
```
fbrosson:
```diff
-		sed -i -e 's/-lm/-lroot/' tests/testrunner.sh 
+	sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
```

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+
+	#devel and common packages
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	#tests run ok up to
+	#runtime_loader: Cannot open file libvaladoc-0.38.so.0: No such file or directory
```
fbrosson: I haven't checked this yet, but I'm quite sure it would work with something like:
```diff
 TEST()
 {
-	#tests run ok up to
-	#runtime_loader: Cannot open file libvaladoc-0.38.so.0: No such file or directory
+	LIBRARY_PATH="$sourceDir/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
 	make check
 }
```
Note: This assumes that the share lib is built in a `.libs` subdirectory of the build directory, so you might need to replace `.libs` by something else if that's not the case.

Remark: With the complicated line I'm suggesting, the body of `TEST()` is equivalent to:
* `LIBRARY_PATH="$sourceDir/.libs" make check` if `LIBRARY_PATH` is an empty string, and to
* `LIBRARY_PATH="$sourceDir/.libs:$LIBRARY_PATH" make check` if `LIBRARY_PATH` is not an empty string.

(But it is better to use the complicated line because it handles both cases.)

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+PATCHES="vala38-$portVersion.patchset"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+PROVIDES="
+	vala$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson:
```diff
-	vala$secondaryArchSuffix = $portVersion compat >= 0
+	vala$secondaryArchSuffix = $portVersionCompat
```

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+	lib:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	"
+
+REPLACES="
+	vala${secondaryArchSuffix}_common
+	"
+
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
```
fbrosson:
```diff
-	vala${secondaryArchSuffix}_devel = $portVersion compat >= 0
+	vala${secondaryArchSuffix}_devel = $portVersionCompat
```

dev-lang/vala/vala38-0.38.8.recipe
Begasus: I think the previous version 0.36.12 is still being developed, would it be wise to also include vala in the REPLACES section for this version? (don't know if both can exist at  the same time ...

fbrosson: We don't need to have `REPLACES` for this purpose, because pkgman knows that:
* it needs to uninstall `0.36.12` if you try to install `0.38.8`, and
* it needs to uninstall `0.38.8` if you try to install `0.36.12`.

fbrosson: `REPLACES` is only used if the basename of a package changes. Here, both recipes provide the same package name: vala.

Oops, I just noticed that you named your recipe `vala38-0.38.8.recipe`. With this name you would need to have a different `PROVIDES`:
```diff
-	vala$secondaryArchSuffix = $portVersionCompat
+	vala38$secondaryArchSuffix = $portVersionCompat
```
and use a `CONFLICTS`.
`REPLACES` are stronger than `CONFLICTS` because once you ask pkgman to install something, it automatically uninstalls any package which is in the `REPLACES`.
`REPLACES` should only be used when you wish to retire an old package name or get rid of old subpackages.

For vala, you can probably rename `vala38-0.38.8.recipe` to `vala-0.38.8.recipe` and let the users choose the version they want.
BTW, if `0.38.8` is not yet well tested then it is better to mark it as untested. That's OK because `0.36.12` will be available from the binary repos.

dev-lang/vala/vala38-0.38.8.recipe
Begasus: Shared libs seem to be installed at "install" time, so not present in the build section .. ?

fbrosson: I haven't tested the recipe yet, but normally, after the call to `make` in `BUILD()`, you should have everything, including the 2 shared libs:
* libvala-0.38.so.0.38.8
* libvaladoc-0.38.so.0.38.8

They might be in a hidden directory (e.g. `.libs`), so the best way to find them is:
```
cd work-0.38.8/sources/vala-0.38.8
find . -name "libvala*.so.*"
```
And `INSTALL()` will not remove them from the build directory, so they will be available to `TEST()` if you use the right directory for the definition of `LIBRARY_PATH`.

Begasus: find . -name "libvala*.so.* doesn't show anything in the $sourceDir ...

fbrosson: Well, that's strange, because with vala-0.36.12 (from your other PR, #1920, which I have already tested) I did find a shared lib, `libvala-0.36.so.0.0.0`, in `work-0.36.12/sources/vala-0.36.12/codegen/.libs/` iirc.

Do you get a `work-0.38.8/sources/vala-0.38.8/codegen/.libs/` directory with this PR?

Begasus: Required lib seems to be in /boot/home/haikuports/dev-lang/vala/work-x86-0.38.8/sources/vala-0.38.8/libvaladoc/.libs (didn't uncheck hide dot files), but even then find didn't seem to catch it ...

fbrosson: I think you forgot to protect the `libvala*.so*` parameter in your find command.
(find will **_never_** miss a file if it exists, unless your command is not correct, or your filesystem is corrupt :)

Anyway, this should allow `make check` to find the `libvaladoc.so`:
```
TEST()
{
	LIBRARY_PATH="$sourceDir/libvaladoc/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
	make check
}
```

Begasus:
Updated, seems it needed both vala and valadoc libs, tests run ok now :)

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+
+	#devel and common packages
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	LIBRARY_PATH="$sourceDir/libvaladoc/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
+	LIBRARY_PATH="$sourceDir/vala/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
```
fbrosson: I'm not sure this would work. What do you think of the following variant?
```diff
 TEST()
 {
-	LIBRARY_PATH="$sourceDir/libvaladoc/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
-	LIBRARY_PATH="$sourceDir/vala/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
+ 	libvalaDir="$sourceDir"/vala/.libs
+ 	libvaladocDir="$sourceDir"/libvaladoc/.libs
+	LIBRARY_PATH="$libvalaDir:$libvaladocDir${LIBRARY_PATH:+:$LIBRARY_PATH}" \
 	make check
 }
```

Begasus: It worked just now, but I can check the variant :)

Begasus:
Requires https://github.com/haikuports/haikuports/pull/2253

dev-lang/vala/vala38-0.38.8.recipe
Begasus: Both seem to be ok (for secondaryArch on x86_gcc2), if you prefer to use your variant I'm glad to update it to that one ... but maybe someone could check it also on x86_64? ...

fbrosson: I've built graphviz on x86_64 (a few days ago) but it fails to build on x86 second arch, so I can't try to build vala-0.38.8 on x86 second arch.
Anyway, when I try to build vala-0.38.8 on x86_64 I get this:
```
checking for glib-2.0 >= 2.40.0 gobject-2.0 >= 2.40.0... yes
checking for gmodule-2.0 >= 2.40.0... yes
checking for libgvc >= 2.16... yes
checking for CGRAPH... no
...
make  all-am
make[3]: Entering directory '/sources/vala-0.38.8/libvaladoc'
  CCLD     libvaladoc-0.38.la
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/5.4.0/../../../../x86_64-unknown-haiku/bin/ld: cannot find -lgvc
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/5.4.0/../../../../x86_64-unknown-haiku/bin/ld: cannot find -lcgraph
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/5.4.0/../../../../x86_64-unknown-haiku/bin/ld: cannot find -lcdt
Makefile:992: recipe for target 'libvaladoc-0.38.la' failed
make[3]: *** [libvaladoc-0.38.la] Error 1
```
I can't investigate at the moment but I'll try to do so soon.

fbrosson: Regarding the `LIBRARY_PATH` in `TEST()`, you are right. Both work.
But you might wish to simplify your code a little:
```
	LIBRARY_PATH="$sourceDir/libvaladoc/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
	LIBRARY_PATH="$sourceDir/vala/.libs:$LIBRARY_PATH" \
 	make check
```
You can also keep what you have right now, but there are 2 small downsides:
1) it is not obvious to the reader that `LIBRARY_PATH` will have `vala/.libs` before `libvaladoc/.libs`.
2) it is not very useful to have the `${LIBRARY_PATH:+:$LIBRARY_PATH}` in each line, as it is only needed in the first line.

For these reasons I would recommend using the variant I suggested above, because it's easier to understand what it does. Also, it only has one `${LIBRARY_PATH:+:$LIBRARY_PATH}`.

There are other variants we could use:
```
	LIBRARY_PATH="$sourceDir/vala/.libs:$sourceDir/libvaladoc/.libs${LIBRARY_PATH:+:$LIBRARY_PATH}" \
	make check
```
Pros: we don't need to define valaDir and libvaladocDir.
Cons: the line is around 100 chars.
But I think it would be OK, here, to have a line > 80 chars.
So this is what I'd recommend, after all, if nobody objects having a long line here.

And here's a final variant:
```
	LIBRARY_PATH="$sourceDir/vala/.libs:$sourceDir/libvaladoc/.libs\
${LIBRARY_PATH:+:$LIBRARY_PATH}" \
	make check
```
It's ugly but at least it fits in less than 80 chars.

Begasus: It requires graphviz (hence my last comment here - and I've posted about that for the previous version for vala) :)
Just build graphviz 2.40.1 on x86_64 and then vala 0.38.8, both build ok ... (just need to run the test for vala, but I don't think that one will be a problem)

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+compared to applications and libraries written in C."
+HOMEPAGE="https://wiki.gnome.org/Projects/Vala"
+COPYRIGHT="2006-2018 Jürg Billeter
+	2006-2018 Raffaele Sandrini"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/vala/${portVersion%.*}/vala-$portVersion.tar.xz"
+CHECKSUM_SHA256="2fa746b51cd66e43577d1da06a80b708c2875cadaafee77e9700ea35cf23882c"
+SOURCE_DIR="vala-$portVersion"
+PATCHES="vala38-$portVersion.patchset"
```
fbrosson: You should either:
* fix the `PROVIDES`, `PROVIDES_devel`, and `defineDebugInfoPackage` (by adding a "38" at the appropriate places), and add a `CONFLICTS`, or
* rename the recipe to `vala-0.38.8.recipe` (and the patch to `vala-0.38.8.patchset`, of course) and make this change:
```
-SOURCE_DIR="vala-$portVersion"
-PATCHES="vala38-$portVersion.patchset"
+PATCHES="vala-$portVersion.patchset"
```

It seems vala-0.36.12 is almost OK and maybe we should merge it before this one.

dev-lang/vala/vala38-0.38.8.recipe
Begasus: Tests run ok on x86_gcc2 (for x86) and x86_64 ...

fbrosson: That's what I did. I have even installed graphviz 2.40.1 but for some reason unknown to me I  vala 0.38.8 fails to build.

The tests for vala 0.36.12 are all OK on all architectures. The only change we might have to do is moving some files to the devel package.

dev-lang/vala/vala38-0.38.8.recipe
Begasus: Right there, probably be something for this weekend, don't know if I get around to it in the next days :)

dev-lang/vala/vala38-0.38.8.recipe
Begasus: I'll wait until someone steps up here to put it in the right sentence before pushing an update :)

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+	$binDir/vapigen \
+	$binDir/vala-${portVersion%.*} \
+	$binDir/valac-${portVersion%.*} \
+	$binDir/valadoc-${portVersion%.*} \
+	$binDir/vapicheck-${portVersion%.*} \
+	$binDir/vapigen-${portVersion%.*}
+
+PATCH()
+{
+	sed -i -e 's/-lm/-lroot/' tests/testrunner.sh
```
alaviss: Is this still necessary? IIRC, @waddlesplash have added empty `libm.a` to avoid this kind of patching. Also, `libroot` is linked by default as it's the libc anyway, but I could be wrong.

dev-lang/vala/vala38-0.38.8.recipe
```diff
@@ -0,0 +1,131 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:autoreconf
+	cmd:awk
+	cmd:bison
+	cmd:cmp
+	cmd:diff
+	cmd:flex
```
alaviss:
```diff
 	cmd:diff
+	cmd:find
 	cmd:flex

File: dev-lang/vala/vala38-0.38.8.recipe [hunk 12]
```diff
@@ -0,0 +1,131 @@
...
+	devel:libcgraph$secondaryArchSuffix
+	devel:libdbus_1$secondaryArchSuffix
+	devel:libglib_2.0$secondaryArchSuffix
+	devel:libgvc$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libintl$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
```
alaviss: These don't appear to be used anywhere in the recipe

```diff
-	cmd:autoconf
-	cmd:automake

File: dev-lang/vala/vala38-0.38.8.recipe [hunk 13]
```diff
@@ -0,0 +1,131 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f $libDir/*.la
+	rm -f $libDir/*/*.la
+	rm -f $libDir/*/*/*/*.la
```
alaviss: A cleaner solution

```diff
-	rm -f $libDir/*.la
-	rm -f $libDir/*/*.la
-	rm -f $libDir/*/*/*/*.la
+	find "$libDir" -name '*.la' -delete

File: dev-lang/vala/vala38-0.38.8.recipe (hunk 10 above)
Comment (fbrosson) [reply dev-lang/vala/vala38-0.38.8.recipe]: If it was just a `'s/-lm//'` then it would have been possible to drop the it.
But here, someone would have to check that it would still work without the `-lroot`.

**Comment by Begasus** [comment]
Updated with given comments, bumped to latest release, cleanup (all test passed for secondaryArch (x86)) ...

File: dev-lang/vala/vala-0.40.1.recipe [hunk 14]
```diff
@@ -0,0 +1,121 @@
...
+CHECKSUM_SHA256="79b544305640b00010008cad0df1a1742363f636e4a7e3d410f651edc728f605"
+SOURCE_DIR="vala-$portVersion"
+PATCHES="vala-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
```
fbrosson:
```diff
-portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
```

dev-lang/vala/vala-0.40.1.recipe
```diff
@@ -0,0 +1,121 @@
...
+
+PROVIDES="
+	vala$secondaryArchSuffix = $portVersionCompat
```
fbrosson:
```diff
 PROVIDES="
-	vala$secondaryArchSuffix = $portVersionCompat
+	vala$secondaryArchSuffix = $portVersion
```

dev-lang/vala/vala-0.40.1.recipe
```diff
@@ -0,0 +1,121 @@
...
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libgvc$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	"
+REPLACES="
+	vala${secondaryArchSuffix}_common
+	"
+
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersionCompat
```
fbrosson:
```diff
 PROVIDES_devel="
-	vala${secondaryArchSuffix}_devel = $portVersionCompat
+	vala${secondaryArchSuffix}_devel = $portVersion
```

dev-lang/vala/vala-0.40.1.recipe
Begasus: Will look into it :+1:

Begasus: Should I still drop portVersionCompat also?

fbrosson: Yep, we don't need it.
This is because it is better to declare the dependencies on lib versions rather than on package versions.

Begasus: Bumped :)

dev-lang/vala/vala-0.40.1.recipe
```diff
@@ -0,0 +1,120 @@
...
+	cmd:vala_gen_introspect_$portVers$secondaryArchSuffix
+	cmd:valac$secondaryArchSuffix
+	cmd:valac_$portVers$secondaryArchSuffix
+	cmd:valadoc$secondaryArchSuffix
+	cmd:valadoc_$portVers$secondaryArchSuffix
+	cmd:vapigen$secondaryArchSuffix
+	cmd:vapigen_$portVers$secondaryArchSuffix
+	lib:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
+	lib:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
+	"
```
fbrosson:
```diff
 	lib:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
 	lib:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
+	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
+	devel:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
 	"
```
This is what we did for vala `0.36.12` and I think we should do so here too.

dev-lang/vala/vala-0.40.1.recipe
```diff
@@ -0,0 +1,120 @@
...
+
+PROVIDES_devel="
+	vala${secondaryArchSuffix}_devel = $portVersion
+	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
+	devel:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	vala$secondaryArchSuffix == $portVersion base
+	"
```
fbrosson:
```diff
-
-PROVIDES_devel="
-	vala${secondaryArchSuffix}_devel = $portVersion
-	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
-	devel:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
-	"
-REQUIRES_devel="
-	vala$secondaryArchSuffix == $portVersion base
-	"
 ```

File: dev-lang/vala/vala-0.40.1.recipe [hunk 19]
```diff
@@ -0,0 +1,120 @@
...
+	make install
+
+	find "$libDir" -name '*.la' -delete
+
+	prepareInstalledDevelLibs libvala-$portVers libvaladoc-$portVers
+	fixPkgconfig
+
+	#devel and common packages
+	packageEntries devel \
+		$developDir
```
fbrosson:
```diff
-
-	#devel and common packages
-	packageEntries devel \
-		$developDir
```

dev-lang/vala/vala-0.40.1.recipe
alaviss: Why are we merging `_devel` to `vala`? I think I have missed the discussions about this.

alaviss: Sounds fine enough to me. Debian appears to split the package, but I believe it's only due to their policy to split things apart. If `vala` is not really usable without the `_devel` part, then I also believe we shouldn't split it apart.

dev-lang/vala/vala-0.40.1.recipe
```diff
@@ -0,0 +1,120 @@
...
+	2006-2018 Raffaele Sandrini"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/vala/${portVersion%.*}/vala-$portVersion.tar.xz"
+CHECKSUM_SHA256="79b544305640b00010008cad0df1a1742363f636e4a7e3d410f651edc728f605"
+SOURCE_DIR="vala-$portVersion"
+PATCHES="vala-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
```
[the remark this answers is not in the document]
Begasus: Not sure if you installed the latest version for Graphviz (as it deppends on it)?

dev-lang/vala/vala-0.40.1.recipe
Begasus: It was still something that has been there in the past (before we removed the _devel package for vala IIRC)

fbrosson: Yep, it was there since the very first time the recipe was added. That was for vala `0.30.1`.

BTW, I'll try to build vala-0.40.3.
```diff
-CHECKSUM_SHA256="79b544305640b00010008cad0df1a1742363f636e4a7e3d410f651edc728f605"
+CHECKSUM_SHA256="6d4f1f0b2edaa0d4aa96f72735a9845e6b1acf80a0a2ae494b5d43e07833119e"
```

Begasus: I'll try it here also, shouldn't be a real change from the version at hand ...

Begasus:
bumbped to 0.40.3, build is ok (and test) for x86 (secondaryArch on x86_gcc2) ... removed the devel package as requested ...

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	find "$libDir" -name '*.la' -delete
+}
```
fbrosson: Hmm, if we drop the
```diff
	prepareInstalledDevelLibs libvala-$portVers libvaladoc-$portVers
	fixPkgconfig
```
then I think we also need to drop the `devel:lib` lines in `PROVIDES`.
Reminder: to check this you need to pass `-S` to haikuporter.

alaviss: Well we do need both lines, as our pkgconfig and linker is configured to search within `B_SYSTEM_DEVELOP_LIB_DIRECTORY`

fbrosson: Yep, I agree.

(BTW, in [this comment](https://github.com/haikuports/haikuports/pull/2256#discussion_r180072230) I had only suggested the removal of the `packageEntries devel $developDir`.)

So, @Begasus, don't forget to restore the 
```
	prepareInstalledDevelLibs libvala-$portVers libvaladoc-$portVers
	fixPkgconfig
```
Thanks!

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+	2006-2018 Raffaele Sandrini"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/vala/${portVersion%.*}/vala-$portVersion.tar.xz"
+CHECKSUM_SHA256="6d4f1f0b2edaa0d4aa96f72735a9845e6b1acf80a0a2ae494b5d43e07833119e"
+SOURCE_DIR="vala-$portVersion"
+PATCHES="vala-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
```
fbrosson:
```diff
-ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
I have tested vala 0.40.3 on all archictectures, (including x86 primary), and all tests are OK:<details>
```
make[4]: Entering directory '/sources/vala-0.40.3/tests'
TEST: Building...                                                      OK
...
All 474 tests passed
TEST: Building...                                                      OK
  /nullability/bug611223:                                              OK
All 1 tests passed
...
make[6]: Entering directory '/sources/vala-0.40.3/valadoc/tests'
TEST: Building...                                                      OK
...
All 8 tests passed
```

BTW, the build failures I had previously on x86_64 were due to a buggy build of graphviz.
After I installed `graphviz{,_devel}` from the repo I was able to successfully build vala and pass all the tests.
</details>

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+	lib:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
+	devel:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
+	devel:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libcdt$secondaryArchSuffix
+	lib:libcgraph$secondaryArchSuffix
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libgvc$secondaryArchSuffix
```
fbrosson:
```diff
 	lib:libglib_2.0$secondaryArchSuffix
+	lib:libgmodule_2.0$secondaryArchSuffix
+	lib:libgobject_2.0$secondaryArchSuffix
 	lib:libgvc$secondaryArchSuffix
```
This is not strictly required because `libgobject-2.0` and `libgmodule-2.0` are being pulled by `libglib-2.0`.
But it is better to mention these libs because they are needed by `lib/valadoc/doclets/*/*.so`.

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+REPLACES="
+	vala${secondaryArchSuffix}_common
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libcdt$secondaryArchSuffix
+	devel:libcgraph$secondaryArchSuffix
+	devel:libglib_2.0$secondaryArchSuffix
+	devel:libgvc$secondaryArchSuffix
```
fbrosson:
```diff
 	devel:libglib_2.0$secondaryArchSuffix
+	devel:libgmodule_2.0$secondaryArchSuffix
+	devel:libgobject_2.0$secondaryArchSuffix
 	devel:libgvc$secondaryArchSuffix
```
This is not strictly required because `libgobject-2.0` and `libgmodule-2.0` are being pulled by `libglib-2.0`.
But it is better to mention these libs because they are needed for building `lib/valadoc/doclets/*/*.so`.

dev-lang/vala/vala-0.40.3.recipe
fbrosson:
```diff
 	find "$libDir" -name '*.la' -delete
+
+	prepareInstalledDevelLibs libvala-$portVers libvaladoc-$portVers
+	fixPkgconfig
 }
```
(as discussed)

dev-lang/vala/vala-0.40.3.recipe
Begasus: Ok, I'll change this later, have seen them passing by but thought pulling glib in would be enough for these :)

fbrosson: Yep, I think it is (slightly) better to include them, but if other contributors think it's OK to only mention libglib2 then that's OK for me.

dev-lang/vala/vala-0.40.3.recipe
Begasus: Just a notice, I always use -S with haikuporter :)

Begasus: I probably won't hurt to include it anyway :) (running fixPkgconfig to make sure everything is correct) :)

dev-lang/vala/vala-0.40.3.recipe
Begasus: I think libintl isn't needed too, autoreconf doesn't use it (not using gettext) ...

fbrosson: I don't know if libintl is required for the build to succeed, but I can say that the shared libs we obtain (when we build vala with `devel:libintl$secondaryArchSuffix` in `BUILD_REQUIRES`) all need libintl:
```bash
cd /system
for i in $(package list -p /system/packages/vala-0.40.3-1-x86_64.hpkg | grep '\.so$' ) ; do objdump -x $i | grep intl && echo $i; done
  NEEDED               libintl.so.8
develop/lib/libvala-0.40.so
  NEEDED               libintl.so.8
develop/lib/libvaladoc-0.40.so
  NEEDED               libintl.so.8
lib/vala-0.40/libvalaccodegen.so
  NEEDED               libintl.so.8
lib/valadoc/doclets/devhelp/libdoclet.so
  NEEDED               libintl.so.8
lib/valadoc/doclets/gtkdoc/libdoclet.so
  NEEDED               libintl.so.8
lib/valadoc/doclets/html/libdoclet.so
```

Begasus: In this PR 0.36.12 wasn't in my branch, so I can't remove it from this branch ...

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,106 @@
...
+additional runtime requirements and without using a different ABI \
+compared to applications and libraries written in C."
+HOMEPAGE="https://wiki.gnome.org/Projects/Vala"
+COPYRIGHT="2006-2018 Jürg Billeter
+	2006-2018 Raffaele Sandrini"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/vala/${portVersion%.*}/vala-$portVersion.tar.xz"
+CHECKSUM_SHA256="6d4f1f0b2edaa0d4aa96f72735a9845e6b1acf80a0a2ae494b5d43e07833119e"
+SOURCE_DIR="vala-$portVersion"
```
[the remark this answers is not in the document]
fbrosson: Good catch.

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,112 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage vala$secondaryArchSuffix \
+	$binDir/vala \
+	$binDir/vala-$portVers \
+	$binDir/valac \
```
fbrosson:
```diff
 defineDebugInfoPackage vala$secondaryArchSuffix \
-	$binDir/vala \
 	$binDir/vala-$portVers \
-	$binDir/valac \
 	$binDir/valac-$portVers \
-	$binDir/valadoc \
 	$binDir/valadoc-$portVers \
-	$binDir/vapigen \
 	$binDir/vapigen-$portVers \
 	$libDir/libvala-$portVers.so.$libVersion \
 	$libDir/libvaladoc-$portVers.so.$libVersion
```

BTW, I lost my (previously working) copy of `vala-0.40.3.recipe` so I had to take yours, and now I'm getting some errors. I'm quite sure, however, that I had written comments for each difference between your's and mine.

(I've recently updated to hrev51880 but I don't think this is related. I'll kick a fresh build and investigate.)

dev-lang/vala/vala-0.40.3.recipe
fbrosson: No problem. After all, that's a good idea to keep a recipe for vala `0.36.x` because that's a LTS branch.

Begasus: Ok, so if this is ok could this be merged, or do you still have problems building it?

fbrosson: I did not succeed yet, but I did not investigate much...

dev-lang/vala/vala-0.40.3.recipe
```diff
@@ -0,0 +1,112 @@
...
+programming language features to GNOME developers without imposing any \
+additional runtime requirements and without using a different ABI \
+compared to applications and libraries written in C."
+HOMEPAGE="https://wiki.gnome.org/Projects/Vala"
+COPYRIGHT="2006-2018 Jürg Billeter
+	2006-2018 Raffaele Sandrini"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/vala/${portVersion%.*}/vala-$portVersion.tar.xz"
+CHECKSUM_SHA256="6d4f1f0b2edaa0d4aa96f72735a9845e6b1acf80a0a2ae494b5d43e07833119e"
```
fbrosson:
```diff
-CHECKSUM_SHA256="6d4f1f0b2edaa0d4aa96f72735a9845e6b1acf80a0a2ae494b5d43e07833119e"
+CHECKSUM_SHA256="379354a2a2f7ee5c4d6e0f5e88b0e32620dcd5f51972baf6d90d9f18eb689198"
```
(That's in case you would like to update to vala `0.40.4` which was released yesterday.)

PR #2259, opened 2018-03-01
retroarch-assets: assets for the RetroArch frontend

kwyxz:
This package installs graphical assets to use with the RetroArch frontend.
No stable release exist as of now, hence the versioning using git. 
The retroarch package depends on this as, without the assets, the interface is not really usable.


games-emulation/retroarch-assets/retroarch_assets-git20180129.recipe
```diff
@@ -0,0 +1,34 @@
...
+
+INSTALL()
+{
+	retroarchDataDir="$dataDir/retroarch/assets/"
+	mkdir -p "$retroarchDataDir"
+	cp -ar * "$retroarchDataDir"
+	rm -rf "$retroarchDataDir"/Makefile \
+		"$retroarchDataDir"/configure \
+		"$retroarchDataDir"/COPYING \
+		"$retroarchDataDir"/src \
```
fbrosson: May I suggest a variant I just found?
```diff
-	rm -rf "$retroarchDataDir"/Makefile \
-		"$retroarchDataDir"/configure \
-		"$retroarchDataDir"/COPYING \
-		"$retroarchDataDir"/src \
-		"$retroarchDataDir"/xmb/NPMApng2PMApng.py \
-		"$retroarchDataDir"/xmb/convert.sh
+	cd "$retroarchDataDir"
+	rm -rf src
+	rm -f configure COPYING Makefile
+	rm -f xmb/convert.sh xmb/NPMApng2PMApng.py
```
Pros: 4 lines instead of 6, and no more unaligned `$retroarchDataDir`.
Cons: 3 rm instead of just 1.

Regarding the name of the recipe, I wonder if it would not be better to use `retroarch_assets-0~git.recipe`.

**Pros:**
* no need to rename the recipe at every update.
* A simple `REVISION` bump would be enough, everytime we update srcGitRev.
* Safer, in case a stable release becomes available with a string  < "git".

**Cons:**
* We are using a 0 as "next target release", but this looks fair enough, because there won't ever be such a version. BTW, we already use `-0~git.recipe` for some recipes.

fbrosson: Oops, if we do rename the recipe then we also need to include `$srcGitRev` in `SOURCE_FILENAME`.
```diff
-SOURCE_FILENAME="retroarch-assets-$portVersion.tar.gz"
+SOURCE_FILENAME="retroarch-assets-$portVersion-$srcGitRev.tar.gz"
```
(But I still think it's a good idea to rename the recipe.)

fbrosson: I forgot to mention that if we drop the `20180129` (in the name of the recipe) then we have to include a `-$srcGitRev` in `SOURCE_FILENAME`.
So maybe we can just replace `.git20180129` by `0~git20180129` in the name of the recipe.
Anyway, I see 2 possibilities:

`portVersion` | `SOURCE_FILENAME`
--- | ---
`0~git20180129` | `"retroarch-assets-$portVersion.tar.gz"`
`0~git` | `"retroarch-assets-$portVersion-$srcGitRev.tar.gz"`

(As we know, the `portVersion` derives from the name of the recipe, `retroarch_assets-$portVersion.recipe`.)

pulkomandy: Keeping the date in the portversion makes more sense, if the sources are different the portversion should change. This way we can use the revision for its original meaning.

fbrosson: That sound fine.
BTW, do you think it would be a good idea to change SOURCE_FILENAME to `"retroarch-assets-${portVersion#0~git}.tar.gz"` ?

kwyxz: Updated recipe with the cosmetic change for the deletions
Is there a consensus regarding the naming? Is the current one fine in the end?

fbrosson: For the naming I'd recommend `retroarch_assets-0~git20180129.recipe`.
That's safer than `retroarch_assets-git20180129.recipe`, which would be considered as "greater" than anything with a numerical portVersion, as weel as anything with a portVersion starting with any letter between a and f.

games-emulation/retroarch-assets/retroarch_assets-git20180129.recipe
```diff
@@ -0,0 +1,32 @@
...
+	haiku
+	"
+
+INSTALL()
+{
+	retroarchDataDir="$dataDir/retroarch/assets/"
+	mkdir -p "$retroarchDataDir"
+	cp -ar * "$retroarchDataDir"
+	cd "$retroarchDataDir"
+	rm -rf src
```
fbrosson:
```diff
-	rm -rf src 
+	rm -rf src
```
(trailing whitspace at the end of the line)

kwyxz: Whoops! Thanks. Fixed that and renamed recipe.

games-emulation/retroarch-assets/retroarch_assets-git20180129.recipe
alaviss: Can we just remove the `git` altogether? That would allow us to handle the case when upstream decides to switch vcs, and overall more consistant versioning across ports.

fbrosson: @alaviss, did you prefer `retroarch_assets-20180129.recipe` or `retroarch_assets-0~20180129.recipe` over `retroarch_assets-0~git20180129.recipe` ?
I think that having `retroarch_assets-0~git20180129.recipe` has two advantages;
1) It makes it clear that we're using a git snapshot.
2) it is flexible enough, thanks to the `0~`, to allow a switch to any new versioning scheme.

alaviss: I would prefer `retroarch_assets-0~20180129git`. This approach is flexible enough that

This way, we can easily swap back and forth official pre-release (`alpha`, `beta`, `rc` suffixes) and snapshot (`vcs` suffix) and would still have the version compared correctly.

Cons is that when official pre-release is out (not many use that versioning scheme), we would have to stick with dates attached.

fbrosson: Actually, with your suggestion (`retroarch_assets-0~20180129git`), and thanks to the `-0~` we can:
* switch to alpha, beta, rc, etc... as long as we keep or increment the date.
* switch to different versioning schemes as often as needed, if we accept to replace `-0~` by `0.1~`, then `0.2~`, `0.3~`... everytime a new versioning scheme would be seen as a backward move if we kept the `-0~` unchanged.
* drop the date (and the `git`) once, by also dropping the `0~` (but this would cut the 2 previous possibilities.)

PR #2260, opened 2018-03-01
retroarch: bump to 1.7.1

kwyxz:
With this update:
- bump to latest RetroArch version
- dependency to ffmpeg libs detailed
- dependency to retroarch_assets added
- default retroarch.cfg file copied in the right place


kwyxz:
Do we need anything to get this moving forward? @fbrosson ? I'm starting to work on additional cores following your guidelines with the FB Alpha ones but they won't be useful if we don't bump RetroArch :D

fbrosson:
Sorry, I forgot to say that the recipe seems OK to me. I have even been able to build (and install) retroarch, retroarch_assets and fbalpha_libretro, all on x86_64.
When I launch retroarch, if finds the shared lib provided by fbalpha_libretro.
Then I get a full screen graphical view with some icons, and I can move the mouse, but I can't select anything.
I wonder if changing this line in `BUILD()` would make any change:
```diff
-		--with-assets_dir="$dataDir"
+		--with-assets_dir="`finddir 
+			B_SYSTEM_DATA_DIRECTORY`/retroarch/assets"
```
Well, I don't think it would make any difference.

kwyxz:
I'll have to test mouse actions, I always use the keyboard / controller to navigate through the menus. Do these work for you at least?

fbrosson:
I tried to use the keyboard but it did not seem to have any effect.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libsdl2$secondaryArchSuffix
+	devel:libgl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
```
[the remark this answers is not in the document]
kwyxz: Good catch, git is not required anymore as we're not running the fetch-submodules script.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg auto-merge
+	"
+
+USER_SETTINGS_FILES="
+	$relativeSettingsDir/retroarch.cfg template \
+		$relativeDataDir/retroarch/retroarch.cfg.default
+	"
```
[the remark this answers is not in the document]
kwyxz: Ugh I am getting very confused. Is it acceptable if I simply remove this section?

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+		-e "s|^# \(libretro_directory =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(libretro_info_path =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(content_directory =\)|\1 \"$retroarchDataDir/\"|;" \
+		-e "s|^# \(assets_directory =\)|\1 \"$retroarchDataDir/assets/\"|;" \
+		retroarch.cfg
+}
+
+BUILD()
+{
+	./configure --prefix="$prefix" --global-config-dir="$settingsDir" \
```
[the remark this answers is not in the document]
kwyxz: Gotcha.

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
fbrosson: Good catch.
If it builds without `cmd:git` (and **_without_** passing `--no-git-repo` to haikuporter) then it would be nice to drop the `cmd:git` here.

(I don't know if that's the case here, but some recipes like dev-vcs/guitar fail to build without `cmd:git` because their build process looks for a `.git` directory, and if it finds that directory but cannot find git itself, it fails to build. This could be fixed by having the builders pass `--no-git-repo` to haikuporter to prevent the creating of the `.git` directory. Until then, we need to keep the _not-really-needed_ `cmd:git` in these recipes.)

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
alaviss: Sorry for polluting this PR, it's more or less a proposal. Just keep it as is until further notice. On the other hand, I should probably move this to Trac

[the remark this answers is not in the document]
kwyxz: Oh no worries, I am just trying to make things right :) but as all this is new to me I'm getting a bit lost amidst all the different keywords.

kwyxz:
Thanks @alaviss, I've done the changes.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -73,7 +72,7 @@ PATCH()
 
 BUILD()
 {
-	./configure --prefix="$prefix" --global-config-dir="$settingsDir" \
+	./configure --prefix="$prefix" --global-config-dir="$sysConfDir" \
```
fbrosson:
```
Warning: POLICY ERROR: Invalid top-level package entry "etc"
Warning: POLICY ERROR: Package declares non-existent global writable file "settings/retroarch.cfg" as included
Warning: Error: packaging policy violation(s) in strict mode
```
Let me check if this was caused by this recent change.

fbrosson:
```diff
-sysConfDir
+sysconfDir
```
(`s/C/c/`)

fbrosson: FWIW:
```
cd games-emulation/retroarch/work-1.7.1
grep -E '^settingsDir=|^sysconfDir=' wrapper-script 
settingsDir='/packages/retroarch-1.7.1-1/.self/settings'
sysconfDir='/packages/retroarch-1.7.1-1/.settings'
```
(I haven't figured yet if there are any differences. Both work fine.)

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,88 @@
...
+		-e "s|^# \(libretro_directory =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(libretro_info_path =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(content_directory =\)|\1 \"$retroarchDataDir/\"|;" \
+		-e "s|^# \(assets_directory =\)|\1 \"$retroarchDataDir/assets/\"|;" \
+		retroarch.cfg
+}
+
+BUILD()
+{
+	./configure --prefix="$prefix" --global-config-dir="$sysConfDir" \
```
fbrosson:
```diff
-	./configure --prefix="$prefix" --global-config-dir="$sysConfDir" \
+	./configure --prefix="$prefix" --global-config-dir="$sysconfDir" \
```
(Everything else is OK.)

fbrosson:
> I'll have to test mouse actions, I always use the keyboard / controller to navigate through the menus. Do these work for you at least?

I just understood how the menu works with the keyboard. Now I can navigate in the menus :tada:

games-emulation/retroarch/retroarch-1.7.1.recipe
kwyxz: I am more and more confused :'D https://github.com/haikuports/haikuports/pull/2260#discussion_r171910057

kwyxz: Though considering https://github.com/haikuports/haikuports/pull/2260#discussion_r171954430 I would be in favor of reverting to settingsDir - as we're looking for the global configuration directory here, so /system/settings

games-emulation/retroarch/retroarch-1.7.1.recipe
kwyxz: My guess is that it's caused by this

+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg auto-merge
+	"

fbrosson:
I'm trying to build with SDL 1.2 (instead of 2.0) to see if it works better.

kwyxz:
@alaviss what is left for me to do to get this merged? Thank you :)

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
alaviss: Please apply the changes above.

In case you're wondering, `sysconfDir` refers to the relocatable **global** `settings` directory, and `settingsDir` refer to the **package's** `settings` directory, which only contains the original version that comes with the package.

fbrosson:
It would be nice to use https (instead of http) for `core_updater_buildbot_assets_url`.
But it seems retroarch does not yet support SSL/TLS. Maybe introducing support for libcurl could be easier than adding support for libssl or gnutls in the existing network-related code.
Remark: This is much easier to say than to do.

Also, it seems we should set `core_assets_directory` to a directory where we have write access.

kwyxz:
The url for `core_updater_buildbot_assets_url` comes from upstream, I'd rather leave it as is.

kwyxz:
I have a question @fbrosson, this morning I did a new build of retroarch and noticed when I installed it it's pulling fbalpha_libretro from haikuports... possibly because we're using SUPPLEMENTS? I don't think we want this to happen, we want libretro cores to depend on retroarch, but installing retroarch should not automatically pull every core.

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
alaviss: I've discussed with @pulkomandy on IRC and it appears that our use of `USER_SETTINGS_FILES` is incorrect, not only for this package but many more. I will open an issue discussing this problem.

For now, please change the path to the configuration file that is created by retroarch in `/home/user/config`. The template should be `$sysconfDir/retroarch.conf`.

Apologizes, more explainations would come later.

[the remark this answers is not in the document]
kwyxz: I'm not sure I understand 100% what you mean, would commenting out the whole USER_SETTINGS_FILES sections work out?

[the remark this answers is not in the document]
alaviss: @kwyxz Please check #2270 for more details

alaviss:
Please use [`addOnsDir`][0] for libretro cores folders, as they more or less are add-ons to libretro

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#environment-variables

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
kwyxz: Thanks for writing up on this @alaviss, I pushed a new commit.

kwyxz:
Ok I'll be updating the cores with addOnsDir too.

kwyxz:
Ok so now the changes done to the recipe had an impact to the default retroarch.cfg file as it doesn't include any of the changes made by the sed command any more...

alaviss:
I don't think `auto-merge` is working yet, so you may want to remove `/system/settings/retroarch.cfg` first.

kwyxz:
It is *very* puzzling. The `retroarch.cfg` file in `work-1.7.1/sources` **is** modified properly with the right paths. So is the default file copied over in /system/data/retroarch.
But /system/settings/retroarch.cfg now looks completely untouched and I don't know which commit broke that :(

alaviss:
I believe the absent of the file was due to the use of `sysconfDir` causes the file to be copied there (therefore only exists in chroot `/system/settings`). Manually copying the file to `settingsDir` might solve the problem.

kwyxz:
May I ask what the benefit of using `sysconfDir` instead of `settingsDir` with `./configure` is? Since we need to manually install the config file that previously was copied over by `make install`?

kwyxz:
I give up for today.
```
INSTALL()
{
        make install
        addAppDeskbarSymlink "$appsDir"/retroarch RetroArch
        install -m 0755 -d "$dataDir"/retroarch
        install -m 0644 -T retroarch.cfg "$settingsDir"/retroarch.cfg
        install -m 0644 -T retroarch.cfg "$dataDir"/retroarch/retroarch.cfg.default
}
```
Making sure none are there and the package is not installed:
```
user@shredder ~ $ ls /system/settings/retroarch.cfg /system/data/retroarch/retroarch.cfg.default
ls: cannot access '/system/settings/retroarch.cfg': No such file or directory
ls: cannot access '/system/data/retroarch/retroarch.cfg.default': No such file or directory
user@shredder ~ $ pkgman uninstall retroarch
*** Failed to find a match for "retroarch": Name not found
```
Building then installing the package and checking on the files:

user@shredder ~ $ hp -v retroarch
	updating dependency infos of retroarch-1.7.1
----------------------------------------------------------------------
games-emulation::retroarch-1.7.1
	/boot/home/haikuports/games-emulation/retroarch/retroarch-1.7.1.recipe
[...]
waiting for build package retroarch-1.7.1-1 to be deactivated
grabbing retroarch-1.7.1-1-x86_64.hpkg and moving it to /boot/home/haikuports/packages/retroarch-1.7.1-1-x86_64.hpkg
```
Installing the package and checking on the config files:
```
user@shredder ~ $ md5sum /system/settings/retroarch.cfg /system/data/retroarch/retroarch.cfg.default 
a7bef273ea5ac5fa48e8cc102a909df8  /system/settings/retroarch.cfg
03709cc381fbca6c695633e7be04af3a  /system/data/retroarch/retroarch.cfg.default
```
There **has** to be something obvious I am missing, but I just don't see it.

**Comment by fbrosson** [comment]
@kwyxz
> I did a new build of retroarch and noticed when I installed it it's pulling fbalpha_libretro from haikuports... possibly because we're using SUPPLEMENTS? I don't think we want this to happen, we want libretro cores to depend on retroarch, but installing retroarch should not automatically pull every core.

Sorry about this. I did not know that `SUPPLEMENTS` had this effect. I though it would make pkgman suggest, but not automatically install. (Well, I haven't checked this yet.)

BTW, when I merged bsnes_libretro (about 2 hours ago) I had not yet seen your comment... But we can remplace the `SUPPLEMENTS` by a `REQUIRES` if you wish.


Regarding the retroarch.cfg (as well as any configuration files), there is a special directory that `pkgman uninstall` does not clean: `/system/packages/administrative/writable-files/`
So, if you want to uninstall retroarch and bring it to the state it was before the installation, you have to do these steps in this order:
```
pkgman uninstall retroarch
rm /system/settings/retroarch.cfg
rm -rf /system/packages/administrative/writable-files/retroarch-1.7.1-1
```
Note: the subdirectories of `/system/packages/administrative/writable-files/` should not be removed if the corresponding package and all its settings have not already been removed.
That said, if I'm not mistake, subdirectories associated to older versions of a package may be removed.

**Comment by fbrosson** [comment]
@alaviss & @kwyxz, I thought the /system/add-ons/` was intended to be used only by packages closely related to the system.
But maybe I'm wrong. Or, in case I'm right, we could still see if this there is a consensus to start using it for packages not related to the system.

**Comment by alaviss** [comment]
@fbrosson Qt and KF5 are the heavy users. BlogPositive also use it for plugins as well. And I believe none of them are closely related to the system in anyway possible.

**Comment by fbrosson** [comment]
OK for me. Let's use it for retroarch too!
(If anyone objects, please comment asap. Thanks!)

File: games-emulation/retroarch/retroarch-1.7.1.recipe [hunk 6]
```diff
@@ -0,0 +1,88 @@
...
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:which
+	"
+
+PATCH()
+{
+	libretroDir="`finddir B_SYSTEM_DIRECTORY`/$relativeLibDir/libretro"
```
fbrosson: If we switch from `/system/lib/libretro/` to `/system/add-ons/libretro/` then:
```diff
-	libretroDir="`finddir B_SYSTEM_DIRECTORY`/$relativeLibDir/libretro"
+	libretroDir="`finddir B_SYSTEM_DIRECTORY`/$relativeAddOnsDir/libretro"
```

kwyxz:
Thanks @fbrosson removing the subfolder in /system/packages/administrative/writable-files/ saved my sanity! Pushing the latest commit.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+GLOBAL_WRITABLE_FILES="
+	$relativeSettingsDir/retroarch.cfg auto-merge
+	"
+
+USER_SETTINGS_FILES="
+	$relativeSettingsDir/retroarch/retroarch.cfg template \
+		$sysconfDir/retroarch.cfg
```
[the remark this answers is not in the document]
kwyxz: Hi @fbrosson 
I've been using `hp -v` since the very beginning, this alias is set as such 
```
hp='haikuporter -S -j4 --get-dependencies --no-source-packages'
```
With the current code `retroarch` builds and installs just fine :)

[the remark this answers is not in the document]
fbrosson: :+1: for using `-S` ;)

Regarding the template, I'm a bit worried by the path we have after the `template` keyword. Here, we are passing the absolute path that is equivalent to the path of the config itself.
My understanding is that what we pass as template should be a relative path to a file in a read-only directory like `$dataDir/retroarch/retroarch.cfg`.

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,89 @@
...
+	"
+
+PATCH()
+{
+	libretroDir="`finddir B_SYSTEM_DIRECTORY`/$relativeAddOnsDir/libretro"
+	retroarchDataDir="`finddir B_SYSTEM_DATA_DIRECTORY`/retroarch"
+	sed -i \
+		-e "s|^# \(libretro_directory =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(libretro_info_path =\)|\1 \"$libretroDir/\"|;" \
+		-e "s|^# \(content_directory =\)|\1 \"$retroarchDataDir/\"|;" \
```
fbrosson: I've just managed to use the _"Online Updater"_ → _"Content Downloader"_ menu, but I had to change the `content_directory` in my `retroarch.cfg`. So I guess we should change `PATCH()` a little. Here's a suggestion:
```diff
 	libretroDir="`finddir B_SYSTEM_DIRECTORY`/$relativeAddOnsDir/libretro"
-	retroarchDataDir="`finddir B_SYSTEM_DATA_DIRECTORY`/retroarch"
+	assetsDir="`finddir B_SYSTEM_DATA_DIRECTORY`/retroarch/assets"
+	contentDir="`finddir B_USER_DATA_DIRECTORY`/retroarch"
 	sed -i \
 		-e "s|^# \(libretro_directory =\)|\1 \"$libretroDir/\"|;" \
 		-e "s|^# \(libretro_info_path =\)|\1 \"$libretroDir/\"|;" \
-		-e "s|^# \(content_directory =\)|\1 \"$retroarchDataDir/\"|;" \
-		-e "s|^# \(assets_directory =\)|\1 \"$retroarchDataDir/assets/\"|;" \
+		-e "s|^# \(assets_directory =\)|\1 \"$assetsDir/\"|;" \
+		-e "s|^# \(content_directory =\)|\1 \"$contentDir/\"|;" \
 		retroarch.cfg
```
And if retroarch does not create the content directory itself (either at startup or when the user chooses to download something) then we will also need to add a post-install action to create it.

alaviss: A small concern of mine regarding the use of `finddir` for global directories is that the resulting package won't use the correct resources when installed per-user. The assets folder then would be `/home/user/config/data/retroarch/assets`, and not `/system/data/retroarch/assets`, which is hard-coded into the configs.

It's possible to install a package per-user, see [Infrastructure].

Also, as described by [DirectoryStructure], `/home/user/config/data` is not writable, so that's something to keep in mind.

[Infrastructure]: https://git.haiku-os.org/haiku/tree/docs/develop/packages/Infrastructure.rst#n60
[DirectoryStructure]: https://git.haiku-os.org/haiku/tree/docs/develop/packages/DirectoryStructure.rst

games-emulation/retroarch/retroarch-1.7.1.recipe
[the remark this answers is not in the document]
alaviss: @fbrosson They're not equivalent, see [Infrastructure](https://git.haiku-os.org/haiku/tree/docs/develop/packages/Infrastructure.rst#n171)

games-emulation/retroarch/retroarch-1.7.1.recipe
fbrosson: Good catch, thanks!
Shall we use
```diff
-	contentDir="`finddir B_USER_DATA_DIRECTORY`/retroarch"
+	contentDir="`finddir B_SYSTEM_NONPACKAGED_DATA_DIRECTORY`/retroarch"
```
This would result in `contentsDir=/boot/system/non-packaged/data/retroarch/`.

BTW, the assets directory would still be `/boot/system/data/retroarch/assets/`.

kwyxz: Wouldn't this defeat the purpose though? Isn't the point of packaging to not use /non-packaged/?

kwyxz: (honest question, maybe there's a meaning behind /non-packaged/ that I am not aware of)

fbrosson: Here the `contentDir` is just a path that we are storing in the default config.
The build process (of the retroarch recipe) does not create anything in there.
If we used `finddir B_SYSTEM_DATA_DIRECTORY`/retroarch"` for the `content_directory` config parameter then retroarch would be unable to write anything in there (as this directory is read-only.)

kwyxz: I see your point. It makes sense to configure the content directory to a writable location for users to download things through the online updater.

I have the feeling though that a per-user package installation is a very specific use case, where it makes way more sense for the user to modify directly the retroarch.cfg file created upon the first run of retroarch than to use the default paths.

fbrosson: I think I have a nice alternative:
```
	contentDir="`finddir B_SYSTEM_VAR_DIRECTORY`/retroarch"
```
This has several advantages:
* It is writable :)
* `finddir B_SYSTEM_VAR_DIRECTORY` yelds `/boot/system/var` and this is common to all users.
* An `mkdir -p $sharedStateDir/retroarch` will create the empty directory at installation time.
* If that directory already exists before installation then its contents are kept.

(I'm not sure yet about the last two items, so I'll give it a try.)

kwyxz: Cool! Waiting for your feedback before I push a new commit then.

fbrosson: I have updated my comment [here](https://github.com/haikuports/haikuports/pull/2260#discussion_r172115179).
With these settings I can  choose _Online Updater_ → _Content Downloader_ and see my downloads land in `/system/var/retroarch/assets/`.
I still have to check whether or not we need a post-install to create some directories.

BTW, I haven't investigated yet the topic about the template config. I'll do that tomorrow ;-)

Regading the `SUPPLEMENTS` I did a test and I can confirm that pkgman does not allow for a fine grained selection. If the user asks to install retroarch, then it does ask for confirmation before install... but there is not way to say we only want to install retroarch. Maybe we should add a `--ignore-supplements` to pkgman ;)

alaviss: @fbrosson I think `--ignore-supplements` should be added and be the default. This really reminded me about `apt-get` family where I always have to pass `--no-install-recommends` to avoid unneeded packages.

kwyxz: I pushed a commit with your changes @fbrosson - thanks!

fbrosson:
I'd be more comfortable with this last change:
```diff
 USER_SETTINGS_FILES="
 	$relativeSettingsDir/retroarch/retroarch.cfg template \
-		$sysconfDir/retroarch.cfg
+		$relativeDataDir/retroarch/retroarch.cfg.default
```
Because I think that using `$sysconfDir/retroarch.cfg` as the template is wrong.
Well, even if what I'm suggesting is not correct, I think it is better than what we have right now.
Shall we add this change before the merge and fix this recipe later, once we have a better understanding of the "template" topic?

(I'm worried by the current output of `package list -i retroarch-1.7.1-1-x86_64.hpkg` which says that the template for the `retroarch.cfg` file is.. a the file itself. To convince yourself about this, you can `ls -l` the path provided, then edit `/system/settings/retroarch.cfg` and then repeat the `ls -l`. You will notice that the template has changed. That's not normal.)

games-emulation/retroarch/retroarch-1.7.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+		--docdir="$docDir" --mandir="$manDir" \
+		--with-assets_dir="$dataDir"
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	addAppDeskbarSymlink "$appsDir"/retroarch RetroArch
+	install -m 0755 -d "$dataDir"/retroarch
```
fbrosson: We **_might_** need this:
```diff
-	install -m 0755 -d "$dataDir"/retroarch 
+	install -m 0755 -d "$dataDir"/retroarch "$sharedStateDir"/retroarch
```
That will make the package have a `var/retroarch/` directory. But I think this won't be enough.
We will probably need to add something like `create-retroarch-directories.sh` in `additional-files`.
(That's what we do for `www-servers/lighttpd` with a `create-lighttpd-directories.sh`.)
But there might be a better method. (I think we should investigate before the merge.)

games-emulation/retroarch/retroarch-1.7.1.recipe
fbrosson: What a coincidence, @alaviss! I also don't like to install packages that are only recommended.
So, in case you did not find yet a workaround for the `--no-install-recommends`, here's how I "fixed" this issue. I just created a file named `/etc/apt/apt.conf.d/no-recommended.conf` and placed this line in it:
```
Apt::Install-Recommends "false";
```

pulkomandy: You are mistaken about what "supplements" is. The goal is not to provide some optional extension to a package. The goal is to do a reverse "requires".

This is used for example for the qthaikuplugins package, which provides the rendering engine for Qt. Without this package it is not possible to use Qt at all. However, it is a plug-in and could be replaced, so Qt does not know that it depends on it.

From the documentation of "supplements":

    this package will automatically be selected for installation if the supplemented entities are already installed.

The use of this is quite rare, usually software is built with known dependencies and REQUIRES can be used and works fine. Only in some rare cases a SUPPLEMENTS is needed. But when it is needed, the relationship is as strong as REQUIRES, and not something you should be able to disable with a command line flag.

We don't have anything similar to apt "recommends" scheme, and we don't plan to add such a thing, because no one likes it :)

pulkomandy: What you describe with plug-ins has already been done, for example for some times we had mesa depending on either mesa_swrast or mesa_swpipe renderers. Using REQUIRES with an agreed upon dependency name works just fine here.

Back to the case at hand: to me it makes perfect sense that the retroarch plugins require the core. What would you do with the plugins if the core is not installed? There is a dependency, just like if retroarch was a library and all the cores would be executables depending on it. It just happens to use different technical choices, making the shared code an executable and the specific one libraries. But that is irrelevant to the package manager. For any engine to be usable you need retroarch to be installed. So they should depend on it. If that was not the case, we would have spared ourselves the effort of manually writing REQUIRES/SUPPLEMENTS at all and would let the package manager guess from executable and libraries inside the packages directly. We added a separate dependency system precisely because this wasn't flexible enough to cover some unusual cases.

fbrosson: > to me it makes perfect sense that the retroarch plugins require the core. What would you do with the plugins if the core is not installed?

I think what you mean by "core" is "engine". In libretro speak, a "core" is what we are calling a plugin.

Anyway, the fact that a plugin is useless without an engine does not mean that the plugin itself requires the engine. It's the **_user_** who needs the engine, not the plugin.

That said, I'm OK with having a `REQUIRES="retroarch"` in each plugin.

I only wish the implementation of `SUPPLEMENTS` was a bit more flexible so that we would be able to replace the `REQUIRES="retroarch"` by a `SUPPLEMENTS="retroarch"` without having all the plugins installed when the user installs retroarch. But I can live with this :)

alaviss:
@fbrosson 
> (I'm worried by the current output of `package list -i retroarch-1.7.1-1-x86_64.hpkg` which says that the template for the `retroarch.cfg` file is the file itself. To convince yourself about this, you can `ls -l` the path provided, then edit `/system/settings/retroarch.cfg` and then repeat the `ls -l`. You will notice that the template has changed. That's not normal.)

Please note that entries in `USER_SETTINGS_FILES` are files relative to `/home/user/config`. So the entry in `USER_SETTINGS_FILES` is actually

    /home/user/config/settings/retroarch/retroarch.cfg template /system/settings/retroarch.cfg

(this is just a clarifying figure)

I told @kwyxz to set the template path to `sysconfDir/retroarch.cfg` because retroarch will copy `/system/settings/retroarch.cfg` to `/home/user/config/settings/retroarch/retroarch.cfg` in the first run.

P/s: If you found `USER_SETTINGS_FILES` confusing in #2270, please help me clarify the explanation there. Thanks.

games-emulation/retroarch/retroarch-1.7.1.recipe
alaviss: Creating the directory there works, just remember to specify it in `GLOBAL_WRITABLE_FILES`

fbrosson:
You are right. Until now I had not correctly understood how `USER_SETTINGS_FILES` was meant to be used.
(When I started using it I made the wrong assumption that the recipes that were already using it were correct, but that was not the case.)

Before the merge, however, I've got a few questions:
1) if you delete your `/boot/home/config/settings/retroarch/retroarch.cfg` does not exist when we launch retroarch then it is created, but it is quite different from the `/boot/system/config/settings/retroarch.cfg`.
2) the relative path of `/boot/system/config/settings/retroarch.cfg` to `/boot/system/config/` is `settings/`, while the relative path of `/boot/home/config/settings/retroarch/retroarch.cfg` to `/boot/home/config/` is `settings/retroarch/`. Hence this question: is it OK to have the system-wide `retroarch.cfg` directly in `settings/` while the user's `retroiarch.cfg` is in `settings/retroarch/` ?
3) Do we still need to create `"$dataDir"/retroarch/retroarch.cfg.default`? If yes, how do we tell the system that this is the default /system/config/settings/retroarch.cfg` ? If not, where may a user look for the template config for the system-wide config?

pulkomandy:
There is nothing to tell the system about for configuration templates.

The use cases for settings files are:
- The application creates the settings itself on first start, and manages them (updates, et): this is the preferred behavior. USER_SETTINGS_FILE can be set to let the package manager know that the file belongs to the application (even if it is not in the package) and could be deleted with it (I think not done yet, similar to "purge" in apt). In that case, we may use the "template" later for a "restore default settings" operation or to make the initial setup for a new user, however there is nothing like that in place yet (and we can add the information when we have something using it)

- The application does NOT create or migrate the settings file on its own (not good, but some UNIX apps are so made). This is handled only for GLOBAL settings file because the problem is usually happening with servers and the like (even on UNIX, user applications know how to create their own files in the user directory). In that case, the package manager is responsible for migrating the files from a version to another, hence we hint it with a strategy (keep-old, merge, ...). I think the "merge" strategy is not implemented yet and will work just like keep-old for now.

So, in UNIX words:
- GLOBAL_WRITABLE_FILES is /etc, and the package manager handles creation and updates of the files.
- USER_SETTINGS_FILES is ~/.config, and the package manager does not manage it for now.

kwyxz:
So... are we good, guys? :)

fbrosson:
OK for me.
We might need other minor changes like setting some default parameters that define directory paths, but I guess this can be done later.

I guess everyone is OK for the merge. Shall we merge?

fbrosson:
I was preparing a commit summary:
```
retroarch: bump to 1.7.1 (#2260)

* Define USER_SETTINGS_FILES and USER_SETTINGS_FILES.
* Set sane default paths for various configuration parameters
  to allow e.g. the Online downloader to work.
```
when I remembered that we haven't investigated yet the topic about creating the directories:
```
cache_directory = "/boot/system/cache/retroarch/"
content_directory = "/boot/system/var/retroarch/"
```
It would be nice to do that before the merge.

I also noticed that the previous version depends on ffmpeg.
Hence this question: do we really want to drop it for 1.7.1?

Thanks for all your great contributions around retroarch!
BTW, please do not hesitate to make suggestions for the commit summary.

kwyxz:
Well, technically this version still more or less depends on ffmpeg:

```
+	lib:libavcodec$secondaryArchSuffix
+	lib:libavformat$secondaryArchSuffix
+	lib:libavdevice$secondaryArchSuffix
+	lib:libswresample$secondaryArchSuffix
+	lib:libavresample$secondaryArchSuffix
+	lib:libavutil$secondaryArchSuffix
+	lib:libswscale$secondaryArchSuffix
```
I just made it more granular upon @alaviss ' request

fbrosson:
My mistake. (I had not noticed that you had replaced ffmpeg by the `lib:libav*`!!!)
That's excellent!

alaviss:
@fbrosson 
>     * Define USER_SETTINGS_FILES and USER_SETTINGS_FILES.

> I remembered that we haven't investigated yet the topic about creating the directories

`install -d`, then specify them in `GLOBAL_WRITABLE_FILES` should work.

fbrosson:
Good news: Everything is working fine.
(I was not sure retroarch would create the `/boot/system/var/retroarch/` directory by itself, but it does so when we choose _"Online Updater"_ → _"Content Downloader"_.)

I also found out that we don't need to create the `/boot/system/cache/retroarch/` directory.
retroarch does not create it, but it does not seem to need it either.
In `PATCH()` we change the default config to set `cache_directory = "/boot/system/cache/retroarch/"`. 
It might be unneeded, but I think it is safer to keep it, as it seems to be a sane default.

**So, for me the recipe is OK for a merge :)**

Sorry for the delay, and thanks again for your great contribution with RetroArch and all its cores!

alaviss:
Oops, I totally forgot that we should specify any global writable directory that we use in `GLOBAL_WRITABLE_FILES` (files/directories that do not exist in the package can also be specified). It's currently just metadata so it's not required, but it could be used to clean unneeded files upon package uninstallation by `pkgman` in the future.

fbrosson:
Me too :)

kwyxz:
I will keep that in mind for the next bump. Thank you @fbrosson and @alaviss for your very patient help with my humble packaging beginnings! Now that Retroarch is merged I can seriously work on getting as many cores packaged as possible. We've already done the work upstream to get most of them to build already so the packaging should be easy now that I've got recipes I can use as templates.

PR #2261, opened 2018-03-01
fbalpha_libretro: Final Burn Alpha libretro core

kwyxz:
This is a package for fbalpha_libretro, a Final Burn Alpha libretro core to use with RetroArch.
Final Burn Alpha is an Arcade and multi-system emulator. The package requires retroarch to be installed.

This is a development version from git, based of stable version 0.2.97.42

@fbrosson I tried using "+" as a separator for versions + git revision but unfortunately haikuporter considers it's an illegal character. I ended up using an extra dot, hopefully it should do the trick without breaking too many guidelines!


games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git20180226.recipe
```diff
@@ -0,0 +1,50 @@
...
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	fbalpha_libretro$secondaryArchSuffix = $portVersion
+	lib:fbalpha_libretro$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	retroarch$secondaryArchSuffix
```
fbrosson:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
+	"
+SUPPLEMENTS="
 	retroarch$secondaryArchSuffix
 	"
```
(Very few recipes have a [`SUPPLEMENTS`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#supplements) keyword, but this one is a very good example.)

BTW, it builds fine on x86 (both primary and secondary), and x86_64.

kwyxz: Interesting feature!

kwyxz:
Uh, it seems like the system I pushed the latest commit from has a completely wrong time and date XD

games-emulation/fbalpha_libretro/fbalpha_libretro-0.2.97.42.git20180226.recipe
fbrosson: Indeed :)

Oops, you forgot to remove the `retroarch$secondaryArchSuffix` from `REQUIRES`.
One of the nice things about `SUPPLEMENTS` is that it allows to create a sort of relation between fbalpha_libretro and retroarch without creating any dependency (on any direction.)

kwyxz: Oh so REQUIRES is not necessary when SUPPLEMENTS is present, I see. Lemme fix that.

PR #2268, opened 2018-03-03
3dengine_libretro: a 3D engine for Libretro

kwyxz:
This 3D engine is a tech demo of what is possible with Libretro in terms of 3D development.


fbrosson:
Builds fine on all architectures.

I can load the core with retroarch, but I haven't found yet how to use any of the cores I have built.
When I choose _Resume_ (or even _Restart_) from the _Quick Menu_ (after having loaded a core), the only thing I get is a dark background (RGB=0x001000) where I can move the mouse and see the cursor move, but nothing else happens.

Cores built so far:
- [x] fbalpha_libretro
- [x] bsnes_libretro
- [x] 3dengine_libretro
- [x] 2048_libretro

kwyxz:
3D Engine is more of a demo / proof-of-concept than an actual core. 2048 however should show you the 2048 game grid when you select "Start Core" after loading it. FBAlpha and Bsnes however require you to load up a game ROM (that we can't distribute for obvious reasons).

fbrosson:
OK, thanks.
With 2048 I do have a "Start Core" but I need to check if my build of retroarch is OK because at the moment when I choose "Start Core" I get two Terminal windows and retroarch crashes.

kwyxz:
Hu yeah, that is weird, it works fine for me. So do the other cores when I start a ROM.

kwyxz:
Updated REQUIRES and removed SUPPLEMENTS

fbrosson:
It would be nice if all contributors merging PRs with more than one commit would take care of cleaning the detailed commit history by editing the modifiable text box that appears when we click on the "Squash and merge" button. GitHub allows this, so I think we should all use this possibility.

(For this commit, for instance, the only interesing line (in the commit message) is the very first one. The other four lines could have been discarded as they are not relevant at all.)

For the paranoid contributors, there is an obvious workaround: squash your commits on your local branch and then `git push` them with `--force`. But I do think that it is better to only squash the commits when the PR is merged... provided the cleanup is done by the contributor merging the PR ;)

Now, after having said this, I'd like to add that I do agree that having a few extra lines in the git log won't hurt anyone. So no worries if this happens again :)

PR #2285, opened 2018-03-05
libunistring: bump to 0.9.9, add debuginfo.

fbrosson:
* Fix compat.
* Move `$infoDir` and html doc to the devel package.
* Drop documentation on second arch.
* Fix LICENSE.
* Switch `HOMEPAGE` & `SOURCE_URI` to https.


dev-libs/libunistring/libunistring-0.9.9.recipe
```diff
@@ -0,0 +1,86 @@
...
+REQUIRES_devel="
+	libunistring$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libiconv$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
```
alaviss:
```diff
-	cmd:autoconf
+	cmd:autoreconf

File: dev-libs/libunistring/patches/libunistring-0.9.9.patchset [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+diff --git a/configure.ac b/configure.ac
+index 38fe5dd..5a822de 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -23,7 +23,7 @@ AC_CONFIG_AUX_DIR([build-aux])
+ . $srcdir/version.sh
+ gl_INIT_PACKAGE([libunistring], [$VERSION_NUMBER])
+ AM_INIT_AUTOMAKE([silent-rules dist-xz])
+-AM_CONFIG_HEADER([config.h])
++AC_CONFIG_HEADERS([config.h])
```
alaviss: Is this even required?

dev-libs/libunistring/libunistring-0.9.9.recipe
fbrosson: Good catch, thanks @alaviss!
Hopefully this does not (currently) have any impact.

Regarding the patchset, you are also right. We can drop it. (I have checked this on each arch, including x86 second arch.
The resulting shared lib is identical.
For the static lib, the checksums do not match but this is due to the fact that, unlike Debian, we haven't implemented reproductible builds yet.

PR #2287, opened 2018-03-06
bluemsx_libretro: MSX/Coleco/SG-1000 emulator core for the libretro API

kwyxz:
This is a libretro core of BlueMSX, originally an emulator of MSX/MSX2 systems, that is now also able to emulate the Coleco Colecovision and the Sega SG-1000 systems.


app-emulation/bluemsx_libretro/bluemsx_libretro-2.9.0.20180219git.recipe
```diff
@@ -0,0 +1,52 @@
+SUMMARY="MSX/Coleco/SG-1000 emulator core for the libretro API"
+DESCRIPTION="This is a libretro port of BlueMSX, originally an emulator of \
+MSX/MSX2 computers. BlueMSX is now able to also emulate the Colecovision and \
+the Sega SG-1000 systems."
+HOMEPAGE="http://bluemsx.msxblue.com/index.htm"
```
fbrosson:
```diff
-HOMEPAGE="http://bluemsx.msxblue.com/index.htm"
+HOMEPAGE="http://bluemsx.msxblue.com/"
```
(This is recommended because the website might choose to switch from static html files to somthing else.)
(Of course this is possible (and OK) because the website provides the same content and because it does not send a 301 redirect.)

kwyxz: Sure thing. Fixed in latest commit.

PR #2291, opened 2018-03-06
libconfig: bump to 1.7.2, add debuginfo & SOURCE_FILENAME.

fbrosson:
Also move $libDir/cmake to the devel package.


dev-libs/libconfig/libconfig-1.7.2.recipe
```diff
@@ -47,6 +51,9 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+defineDebugInfoPackage libconfig$secondaryArchSuffix \
+	"$libDir"/libconfig.so.$libVersion
```
alaviss: Are you missing `libconfig++` here?

PR #2294, opened 2018-03-08
fdk-aac: bump to 0.1.6, move from libfdk_aac/ to fdk-aac/.

fbrosson:


fbrosson:
The recipe for the previous release, 0.1.5, was added to `media-libs/libfdk_aac/` although the [GCI task](https://codein.withgoogle.com/tasks/6011364982128640/) suggested to add it to `media-libs/fdk-aac/` (to follow [Gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/media-libs/fdk-aac)).
But nobody noticed it was not using the recommended path when the task was [validated](https://codein.withgoogle.com/dashboard/task-instances/6406800649748480/) and merged in #1878.

@alaviss, @waddlesplash and others:
Should we keep the recipe in `media-libs/libfdk_aac/` or do we move (and rename) it to `media-libs/fdk-aac/` to follow Gentoo's directory naming?

(I have assumed it is better to move, and have therefore added a `REPLACES`, but I can drop the second commit if you prefer to stick to the path we have right now.)

waddlesplash:
Moving is good, yes.

alaviss:
It was originally in `fdk_aac`. IIRC I renamed it to `libfdk_aac` due to the fact that this port only contains the library (Gentoo package may optionally include the sample encode/decoder).

With that said, I believe it's fine with either names.

fbrosson:
OK, I see. Yes, both are fine, but having it in fdk-aac is slightly better, even if we just have the library.
(If Gentoo had both a libfdk-aac and a fdk-aac, then yes, it would make much more sense to stay in libfdk_aac or libfdk-aac.)
So, no regrets for the move :)

pulkomandy:
On Sat, Mar 10, 2018 at 04:29:27PM +0000, fbrosson wrote:
> OK, I see. Yes, both are fine, but having it in fdk-aac is slightly better, even if we just have the library.
> (If Gentoo had both a libfdk-aac and a fdk-aac, then yes, it would make much more sense to stay in libfdk_aac or libfdk-aac.)
> So, no regrets for the move :)

We try to match Gentoo in terms of repo organization, but not
necessarily in terms of package names. So the simplest fix would be to
move, but not rename the recipe (avoids messing with REPLACES, etc).

PR #2295, opened 2018-03-08
New recipe for lightsoff v1.1

humdingerb:


haiku-games/lightsoff/lightsoff-1.1.recipe
```diff
@@ -0,0 +1,56 @@
...
+BUILD()
+{
+	make -C src OBJ_DIR=objects $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $instDir $instDir/docs $instDir/media
+	cp src/objects/$APP $instDir
+	cp -r docs $instDir
```
fbrosson: With the current recipe we have these files/dirs in `$appsDir/LightsOff/`:
```
apps/LightsOff/LightsOff
apps/LightsOff/docs/
apps/LightsOff/docs/GettingStarted.html
apps/LightsOff/media/
apps/LightsOff/media/altwin.wav
apps/LightsOff/media/click.wav
apps/LightsOff/media/fanfare.wav
```
I think it would be better to put the html doc in `$docDir` and the wav media in `$dataDir/LightsOff/`.

**Pros:**
* The html file would be in the recommended directory tree.
* The wav files would be in a directory meant to be used for data files.
* The app would be launchable from the command line with a simple `LightsOff` since `/boot/system/apps` is in the `PATH`.

**Cons:**
* This might need some extra changes in the sources.

That said, I don't object a merge as is.

BTW, it would be nice to drop the old recipe, `lightsoff-1.0.recipe`.

humdingerb:
Whenever the "all-in-one-folder" vs. "spread-to-several-folders" debate comes up, people can't agree. :) I tend to be in team "all-in-one" for native GUI apps, BTW, but if we ever had a consensus one way or the other, I'd go along with whatever were decided.

As there's already a mix of both, maybe we can keep this one as is. It does simplify the code to only have to look in the app's subfolder, and a code change would require again another tag. I also don't see many people starting a game from Terminal.

I'd like to keep the previous version around, in case people would like to be able to quickly compare versions. Will delete v1.0 when v1.2 comes out.

fbrosson:
No problem :)
BTW, I have tested the game on x86_64 and x86_gcc2 and everything looks OK.

humdingerb:
Great!

> BTW, I have tested the game on x86_64 and x86_gcc2 and everything looks OK.

Me, too. :)
Thanks to Owen for all the bug fixes and new features!

pulkomandy:
Actually the right way would be to embed the sounds inside the executable as resources. Possibly the html as well but that would be a little more difficult (it should work to pass it to Web+ as a data: uri, maybe)

humdingerb:
At least wrt to documentation, I'd rather have a separate html file. People can access it without launching the app, can create symlinks to it, bookmark it. People may love the "click.wav and re-use it as their Vision notification... :)
Generally, I don't see much benefit of doing away with separate files.

PR #2301, opened 2018-03-09
cap32_libretro: a libretro port of the Caprice32 Amstrad CPC emulator

kwyxz:
This is the libretro core for Caprice 32, the Amstrad CPC emulator that is able to faithfully imitate the CPC464, CPC664, and CPC6128 models.


app-emulation/cap32_libretro/cap32_libretro-4.2.0_20171109.recipe
```diff
@@ -0,0 +1,54 @@
+SUMMARY="A port of Caprice32, an Amstrad CPC emulator, to the libretro API"
+DESCRIPTION="Caprice32 is a software emulator of the Amstrad CPC 8bit home \
+computer series. The emulator faithfully imitates the CPC464, CPC664, and \
+CPC6128 models."
+HOMEPAGE="http://sourceforge.net/projects/caprice32/"
```
fbrosson:
```diff
-HOMEPAGE="http://sourceforge.net/projects/caprice32/"
+HOMEPAGE="https://sourceforge.net/projects/caprice32/
+	https://github.com/libretro/libretro-cap32"
-COPYRIGHT="1997-2018, Ulrich Doewich, dantoine the libretro team"
+COPYRIGHT="1997-2018 Ulrich Doewich, dantoine the libretro team"
```
Yep, it is possible to have more than one `HOMEPAGE`.
I don't know, however, if that's a good idea, for this recipe.
Anyway, if you keep just the SF homepage, there is a small change: `s/http/https/`

kwyxz: Hmm sure why not but we've never done this before for the other cores?
Pushed recipe with "https"

pulkomandy:
On Sat, Mar 10, 2018 at 06:04:26AM +0000, Benjamin FRANCOIS wrote:
> +DESCRIPTION="Caprice32 is a software emulator of the Amstrad CPC 8bit home \
> +computer series. The emulator faithfully imitates the CPC464, CPC664, and \
> +CPC6128 models."

I'm sorry, but I would remove "faithfully" here :o)

Having worked on a fork of Caprice32 and after that switching to the
much more accurate ACE emulator, both available for Haiku, I can tell
that the original Caprice32 isn't really that accurate as an emulator.

(it's just a personal opinion, feel free to ignore it)

fbrosson:
Here's a possible variant:
```diff
-computer series. The emulator faithfully imitates the CPC464, CPC664, and \
-CPC6128 models."
+computer series. The emulator tries to faithfully imitate the CPC464, CPC664, \
+and CPC6128 models."
```
But please feel free to stick to keep or drop the `faithfully`.
I've though about this in case it could be a sort of compromise between keeping vs dropping the word `faithfully`, but this does not mean it is a better solution.

kwyxz:
I don't have any opinion on that topic, if Caprice 32 is lacking in terms of accuracy I have no problem with dropping "faithfully" :D

fbrosson:
I'm OK with both keeping and dropping the "faithfully", but if we keep it and it's not really true then some users might be disappointed.
That said, I don't have an objective opinion here because I haven't tried this core :D

So, shall we merge the recipe as is?
Or shall we wait a little to see if someone else has an opinion regarding the "faithfully"?

kwyxz:
I dropped "faithfully", that should make it easier to get a merge :D

app-emulation/cap32_libretro/cap32_libretro-4.2.0_20171109.recipe
```diff
@@ -0,0 +1,53 @@
+SUMMARY="A port of Caprice32, an Amstrad CPC emulator, to the libretro API"
+DESCRIPTION="Caprice32 is a software emulator of the Amstrad CPC 8bit home \
+computer series. The emulator imitates the CPC464, CPC664, and CPC6128 models."
+HOMEPAGE="https://sourceforge.net/projects/caprice32/"
+COPYRIGHT="1997-2018, Ulrich Doewich, dantoine the libretro team"
```
fbrosson: Do we get rid of two commas?
```diff
-computer series. The emulator imitates the CPC464, CPC664, and CPC6128 models."
+computer series. The emulator imitates the CPC464, CPC664 and CPC6128 models."
 HOMEPAGE="https://sourceforge.net/projects/caprice32/"
-COPYRIGHT="1997-2018, Ulrich Doewich, dantoine the libretro team"
+COPYRIGHT="1997-2018 Ulrich Doewich, dantoine the libretro team"
```
I'm not 100% sure about the first one, i.e. before the "and".
BTW, in `COPYRIGHT`, is "dantoine the libretro team" OK?

kwyxz: dantoine is a separate copyright holder
I fixed that

PR #2303, opened 2018-03-10
dev-libs/libmspack: bump to 0.6alpha, add debuginfo.

fbrosson:
* Add `cmd:` entries for the runtimes added in this release.
* Call `cabd_test` in `TEST()`. (`ALL 194 TESTS PASSED.`)
* Add patch from Gentoo to fix `cabd_test`.
* Compute arguments being passed to `defineDebugInfoPackage`.

### DIFF
--- a/dev-libs/libmspack/libmspack-0.6~alpha.recipe
+++ b/dev-libs/libmspack/libmspack-0.6~alpha.recipe
@@ -13,27 +13,43 @@ version of any particular compressor or decompressor, use the \
 mspack_version() function. The UNIX library interface version is defined as \
 the highest-versioned library component."
 HOMEPAGE="https://www.cabextract.org.uk/libmspack/"
-COPYRIGHT="2000-2015 Stuart Caie"
+COPYRIGHT="2000-2017 Stuart Caie"
 LICENSE="GNU LGPL v2"
-REVISION="2"
+REVISION="1"
 SOURCE_URI="$HOMEPAGE/libmspack-${portVersion/\~/}.tar.gz"
-CHECKSUM_SHA256="8967f275525f5067b364cee43b73e44d0433668c39f9376dfff19f653d1c8110"
+CHECKSUM_SHA256="1edbee82accb28e679ab538f803aab7a5a569e4102ccf1715b462b1bd915f921"
 SOURCE_DIR="libmspack-${portVersion/\~/}"
+
+SOURCE_FILENAME_2="libmspack-${portVersion/\~/_}-fix-tests.patch"
+srcGitRev_2="831379877f8c646c5643dd98305afdb5a6b4ca6c"
+SOURCE_URI_2="https://gitweb.gentoo.org/repo/gentoo.git/plain/dev-libs/libmspack/files/$SOURCE_FILENAME_2?id=$srcGitRev_2#noarchive"
+CHECKSUM_SHA256_2="ab1bbe6b469316c320507f1b13aee36fe06b3732949c52b8e5e0f8dd4c3cf8a1"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86_gcc2 x86"
 
+libVersion="0.1.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libmspack$secondaryArchSuffix = $portVersion
-	lib:libmspack$secondaryArchSuffix = 0.1.0 compat >= 0
+	lib:libmspack$secondaryArchSuffix = $libVersionCompat
 	"
+if [ -z "$secondaryArchSuffix" ]; then
+PROVIDES="$PROVIDES
+	cmd:cabrip = $portVersion
+	cmd:chmextract = $portVersion
+	cmd:msexpand = $portVersion
+	cmd:oabextract = $portVersion
+	"
+fi
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
 
 PROVIDES_devel="
 	libmspack${secondaryArchSuffix}_devel = $portVersion
-	devel:libmspack$secondaryArchSuffix = 0.1.0 compat >= 0
+	devel:libmspack$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES_devel="
 	libmspack$secondaryArchSuffix == $portVersion base
@@ -52,7 +68,24 @@ BUILD_PREREQUIRES="
 	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
 	"
+
+debugList=(
+	"$libDir"/libmspack.so.$libVersion
+	)
+if [ -z "$secondaryArchSuffix" ]; then
+	for i in `echo "$PROVIDES" | sed -n -e \
+			"s/^[\t]*cmd:\(.*\)$secondaryArchSuffix =.*/\1/p;"`; do
+		debugList+=("$binDir"/$i)
+	done
+fi
+defineDebugInfoPackage libmspack$secondaryArchSuffix "${debugList[@]}"
+
+PATCH()
+{
+	patch -p1 -i "$sourceDir2"/$SOURCE_FILENAME_2
+}
 
 BUILD()
 {
@@ -74,6 +107,8 @@ INSTALL()
 		cp -r doc/* $developDocDir
 		rm $developDocDir/Doxyfile*
 		rm $developDocDir/Makefile*
+	else
+		rm -rf "$binDir"
 	fi
 
 	rm $libDir/libmspack.la
@@ -88,4 +123,6 @@ INSTALL()
 TEST()
 {
 	make check
+	cd test
+	./cabd_test
 }


dev-libs/libmspack/libmspack-0.6~alpha.recipe
```diff
@@ -13,27 +13,43 @@ version of any particular compressor or decompressor, use the \
...
+REVISION="1"
 SOURCE_URI="$HOMEPAGE/libmspack-${portVersion/\~/}.tar.gz"
-CHECKSUM_SHA256="8967f275525f5067b364cee43b73e44d0433668c39f9376dfff19f653d1c8110"
+CHECKSUM_SHA256="1edbee82accb28e679ab538f803aab7a5a569e4102ccf1715b462b1bd915f921"
 SOURCE_DIR="libmspack-${portVersion/\~/}"
 
+SOURCE_FILENAME_2="libmspack-${portVersion/\~/_}-fix-tests.patch"
+srcGitRev_2="831379877f8c646c5643dd98305afdb5a6b4ca6c"
+SOURCE_URI_2="https://gitweb.gentoo.org/repo/gentoo.git/plain/dev-libs/libmspack/files/$SOURCE_FILENAME_2?id=$srcGitRev_2#noarchive"
+CHECKSUM_SHA256_2="ab1bbe6b469316c320507f1b13aee36fe06b3732949c52b8e5e0f8dd4c3cf8a1"
```
alaviss: You could grab the upstream commit instead: https://github.com/kyz/libmspack/commit/deb65056f436a4ce06946c4dde270f533e3cab4d

dev-libs/libmspack/libmspack-0.6~alpha.recipe
```diff
@@ -13,27 +13,43 @@ version of any particular compressor or decompressor, use the \
...
+	lib:libmspack$secondaryArchSuffix = $libVersionCompat
 	"
+if [ -z "$secondaryArchSuffix" ]; then
+PROVIDES="$PROVIDES
+	cmd:cabrip = $portVersion
+	cmd:chmextract = $portVersion
+	cmd:msexpand = $portVersion
+	cmd:oabextract = $portVersion
+	"
+fi
```
alaviss: I believe we should also have a policy for when should we just get rid of the `cmd:`s

dev-libs/libmspack/libmspack-0.6~alpha.recipe
```diff
@@ -88,4 +123,6 @@ INSTALL()
 TEST()
 {
 	make check
+	cd test
+	./cabd_test
```
alaviss: Is `cabd_test` not automatically run by `make check`?

dev-libs/libmspack/libmspack-0.6~alpha.recipe
fbrosson: Well, I believe we do have such a policy. It's only that it's not documented ;)
From what I have understood, here are the rules:
* If a recipe builds on all primary architectures then it is recommended to drop the `cmd:` on secondary arch builds.
* If, however, the execution speed is important or if the runtimes could have unnoticed bugs (when built with gcc2) then we keep the `cmd:` for all builds.

dev-libs/libmspack/libmspack-0.6~alpha.recipe
fbrosson: Since the fix is only for the tests, I'd rather stick to the stable release.
Dropping these 4 lines and the `PATCH()` function (once the next release is out) will be better than switching to a snapshot and, later, switching back to a stable release.
IMHO the only benefit in swithing to a snapshot would be to get rid of the `PATCH()` function and the `*_2` variables, but we know that these are only temporary.

dev-libs/libmspack/libmspack-0.6~alpha.recipe
fbrosson: Nope.
(I just checked this again to make sure.)

BTW, Are OK with sticking to the stable release and keeping Gentoo's patch?

dev-libs/libmspack/libmspack-0.6~alpha.recipe
alaviss: I mean downloading the commit (add `.patch` to the end of commit uri) then do `git am` to the source then `haikuporter -e`

Pros:
- No need to fetch an external patch

Cons:
- Bigger diff

An alternative is to fetch the patch directly from upstream.

fbrosson: Hmm, the "grab the patch from Gentoo" method is more generic than the `git am` + `haikuporter -e` method, and imho it has the big advantage of adding a constant number of lines, no matter how big the patch is.
The other advantage is that it makes it clear that the patch is not from us. This also means we credit Gentoo for their nice work :)

pulkomandy:
On Sun, Mar 11, 2018 at 01:31:52AM -0800, fbrosson wrote:
> Well, I believe we do have such a policy. It's only that it's not documented ;)
> From what I have understood, here are the rules:
> * If a recipe builds on all primary architectures then it is recommended to drop the `cmd:` on secondary arch builds.
> * If, however, the execution speed is important or if the runtimes could have unnoticed bugs (when built with gcc2) then we keep the `cmd:` for all builds.

There is no rule because common sense should be used.

What I would do is:
* If the recipe is built for secondary arch already (eg. because it
  contains libs), consider using the binaries from the secondary arch.
  There is no reason to keep using gcc2 here, so let's start moving away
  from it. In that case, the commands should be built by the secondary
  recipe, but moved to the main "bin" directory
* If the recipe does not use the secondary arch and builds fine with
  gcc2 without efforts (no patches needed), it's reasonable to keep
  using gcc2 as long as it works.

There is the special case of needing to keep both. This is the case for
example for gcc (because the executables are very different and used for
different purposes), sdl-config, pkg-config and similar tools. In that
case the commands are kept in both primary and secondary.

However, this makes the recipes more complex (especially in the first
case), so I'm fine with not forcing people to do it that way.

pulkomandy:
On Sun, Mar 11, 2018 at 10:48:30AM +0000, fbrosson wrote:
> fbrosson commented on this pull request.
> 
> 
> 
> >  SOURCE_DIR="libmspack-${portVersion/\~/}"
>  
> +SOURCE_FILENAME_2="libmspack-${portVersion/\~/_}-fix-tests.patch"
> +srcGitRev_2="831379877f8c646c5643dd98305afdb5a6b4ca6c"
> +SOURCE_URI_2="https://gitweb.gentoo.org/repo/gentoo.git/plain/dev-libs/libmspack/files/$SOURCE_FILENAME_2?id=$srcGitRev_2#noarchive"
> +CHECKSUM_SHA256_2="ab1bbe6b469316c320507f1b13aee36fe06b3732949c52b8e5e0f8dd4c3cf8a1"
> 
> Hmm, the "grab the patch from Gentoo" method is more generic than the `git am` + `haikuporter -e` method, and imho it has the big advantage of adding a constant number of lines, no matter how big the patch is.
> The other advantage is that it makes it clear that the patch is not from us. This also means we credit Gentoo for their nice work :)

You can keep ownership when you create patchsets from haikuporter (git
--set-author or the like). I'm uncomfortable with relying on the gentoo
repo staying online so we can build things. I would prefer to have a
copy of the patches in our own repo instead (converted as a patchset).

fbrosson:
I'm quite sure that **_if_** Gentoo's repo has ever been unavailable, it was certainly for a very short time.
And, if I'm not mistaken, as soon as this recipe will be built by the build servers, a copy of the patch will be available on Haiku's cache.
So I think that trying to avoid Gentoo's repo is not necessary.
Moreover, the patch here is just for the tests, so even if Gentoo's repo becomes unavailable for an unacceptable long period, the workaround would be easy: comment out the parts that pull the patch.
IMHO we can live with this very small risk. There is no need to add a patchset that will be removed with the next release. Don't you agree?

pulkomandy:
On Sun, Mar 11, 2018 at 12:15:13PM +0000, fbrosson wrote:
> I'm quite sure that **_if_** Gentoo's repo has ever been unavailable, it was certainly for a very short time.
> And, if I'm not mistaken, as soon as this recipe will be built by the build servers, a copy of the patch will be available on Haiku's cache.
> So I think that trying to avoid Gentoo's repo is not necessary.
> Moreover, the patch here is just for the tests, so even if Gentoo's repo becomes unavailable for an unacceptable long period, the workaround would be easy: comment out the parts that pull the patch.
> IMHO we can live with this very small risk. There is no need to add a patchset that will be removed with the next release. Don't you agree?

In that case I would do nothing and wait for the next release :)

fbrosson:
0.6alpha fixes two vulnerabilities, as stated in § `Security vulnerabilities in libmspack` (see https://www.cabextract.org.uk/libmspack/):
* [CVE-2017-11423](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-11423)
* [CVE-2017-6419](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-6419)

Moreover, since the problem in this release is not in the software itself but in its test suite, the next release might not be available very soon.
So I believe it would be better to merge this PR without waiting for the next release.

fbrosson:
Shall I merge?

PR #2304, opened 2018-03-10
craft_libretro: a Minecraft clone libretro core

kwyxz:
A very performant Minecraft clone written in just a few thousand lines of C using modern OpenGL (shaders). Online multiplayer support is included using a Python-based server. Many people have used this as a base upon which a more complex game was built.


games-misc/craft_libretro/craft_libretro-1.0_20180210.recipe
```diff
@@ -0,0 +1,55 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	retroarch$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	mesa${secondaryArchSuffix}_devel
```
fbrosson: It builds fine, but retroarch crashes (on my system) when I try to load the core.
I tried with the following changes but it still crashes:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	retroarch$secondaryArchSuffix
+	lib:libcurl$secondaryArchSuffix
 	lib:libGL$secondaryArchSuffix
+	lib:libglew$secondaryArchSuffix
+	lib:libneon$secondaryArchSuffix
+	lib:libsqlite3$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libzstd$secondaryArchSuffix
 	"
```
```diff
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	mesa${secondaryArchSuffix}_devel
+	devel:libcurl$secondaryArchSuffix
+	devel:libGL$secondaryArchSuffix
+	devel:libglew$secondaryArchSuffix
+	devel:libneon$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libzstd$secondaryArchSuffix
	"
```
I'm suggesting libz because I read it in the Makefile.libretro iirc.
I'm suggesting libglew, libcurl and libssl because I read it in the README.md iirc.
I'm suggesting libzstd because the Debugger asked me to locate libzstd_debuginfo. But maybe this does not mean that the core needs libzstd.

```diff
-	make -f Makefile.libretro $jobArgs
+	make -f Makefile.libretro $jobArgs HAVE_LIBCURL=1 HAVE_NEON=1 DEBUG=1
```
Remark: Most of the dependencies I have added are not used:
```
objdump -x /system/add-ons/libretro/craft_libretro.so | grep NEEDED
  NEEDED               libnetwork.so
  NEEDED               libcurl.so.4
  NEEDED               libGL.so.1
```

Anyway, it might be worth applying these changes:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	retroarch$secondaryArchSuffix
+	lib:libcurl$secondaryArchSuffix
 	lib:libGL$secondaryArchSuffix
 	"
```
```diff
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	mesa${secondaryArchSuffix}_devel
+	devel:libcurl$secondaryArchSuffix
+	devel:libGL$secondaryArchSuffix
```
```diff
-	make -f Makefile.libretro $jobArgs
+	make -f Makefile.libretro $jobArgs HAVE_LIBCURL=1
```

(I should update my system because I'm not using the latest hrev...)

kwyxz: Interesting, it crashes for me too, but it worked a few weeks ago when I tested it to see if some changes needed to be done upstream. I'll look into that.

kwyxz: Even weirder: the last commit upstream is... to make the core build and work on Haiku https://github.com/libretro/Craft/commit/7421f7921a7a52b65100e7c2189aa764cdd5de60

kwyxz: Ok so I just built the core directly without using haikuporter and it works, just doing make -f `Makefile.libretro`. This is puzzling to say the least.

fbrosson: I think this could be due to a missing dependency.
If you run
```
objdump -x /system/add-ons/libretro/craft_libretro.so | grep NEEDED
```
with the build which works fine, do you get the same results as with the build which crashes?
Do you get this?
```
  NEEDED               libnetwork.so
  NEEDED               libGL.so.1
```

kwyxz: Fixed it. I noticed during build that Makefile.common was making calls to git. 
Adding GIT_VERSION as an argument to make gets the core to work for me.
New commit pushed.

kwyxz: So I'm guessing the chroot haikuporter uses to build is missing git, which caused this.

fbrosson: Exact.
BTW, now I can load the core, but later, when I choose "Start Core", retroarch crashes.

kwyxz: Ha. That is something else... it works fine here, I can get into the game and play.

fbrosson: I guess the problem is in my system.
I'll approve the PR. May I let you merge it yourself?
(That would be strange if I merge the PR, since the core crashes retroarch on my system.)

BTW, when you merge, may I suggest that you drop the _optional extended description_ and only keep the commit title?
(If you don't see an "Add an optional extended description..." in the text zone above the "Confirm squash and merge" button, this means there is at least one space (or any other character) in this field, and if you don't remove everything then the title will be followed by a "...")

Regarding the commit "title", I try to keep it <= 69 chars, including the ` (#xxxx)`.

PR #2305, opened 2018-03-10
libgit2: bump to 0.26.2, add debuginfo, drop (broken) TEST().

fbrosson:


alaviss:
Why are the `TEST()` broken? And you should add `SOURCE_FILENAME` to rename the archive.

fbrosson:
Good catch, thanks for spotting the missing `SOURCE_FILENAME`.
(I had though about it but later forgot to add it :(

I'm trying to see if we can rescue the `TEST()` function :)
(It might not be "broken", after all. I wrote this because there are no "test" nor "check" targets, but there must be some way to make use of all the files in the `tests` subdirectory.)

dev-libs/libgit2/libgit2-0.26.2.recipe
```diff
@@ -12,19 +12,23 @@ DESCRIPTION="The library provides:
...
 SOURCE_URI="https://github.com/libgit2/libgit2/archive/v$portVersion.tar.gz"
-CHECKSUM_SHA256="6a62393e0ceb37d02fe0d5707713f504e7acac9006ef33da1e88960bd78b6eac"
+CHECKSUM_SHA256="747b47d5b02a2387ff81301c694763785181b895690b6eb91ed1ae4b7904307b"
+SOURCE_FILENAME="libgit2-$portVersion.tar.gz"
 
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%.*}"
```
fbrosson: The name of the symlink is not `libgit2.so.0` but `libgit2.so.26`:
```
~> ls -go /system/lib/libgit2.so.26 
lr-xr-xr-x 1 0 Mar 11 13:53 /system/lib/libgit2.so.26 -> libgit2.so.0.26.2

~> readelf -d /system/lib/libgit2.so.26 | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libgit2.so.26]
```
Should I change `libVersionCompat`?
```diff
-libVersionCompat="$libVersion compat >= ${libVersion%.*}"
+libVersionCompat="$libVersion compat >= 26"
```

Remarks:
* libgit2 is used by `kde-frameworks/ktexteditor/` and by 2 of the recipes of `dev-lang/rust/`: `1.21~dev` and `1.22~dev`
* I have already rebuilt libgit2 with these changes:
  * in `BUILD()`: `cmake --build .` instead of `make $jobArgs`
  * in `INSTALL()`: `cmake --build . --target install` instead of `make install`

in the hope that it would create `libgit2.so.26.2` but it still creates libgit2.so.0.26.2`.

Or should we rename the symlink to `libgit2.so.0.26` (and keep `libVersionCompat` unchanged?

fbrosson: > Or should we rename the symlink to libgit2.so.0.26 (and keep libVersionCompat unchanged?

Hmm, I think the answer to my own question is: no, because the symlink needs to match the soname.
So I guess what we need to do is:
* `libVersionCompat="$libVersion compat >= 26"`
* bump `REVISION` for the 3 recipes that need libgit2, once we have merged this PR.

Is this correct?

alaviss: Sure, please make sure that the recipes you mentioned were built by our buildbots before bumping the revision.

fbrosson:
Well, there's a small problem. I can build ktexteditor (against libgit2-0.26.2), but I can't install it, because of the **_nothing provides lib:libgit2>=0.26.2_**:
```
~> package list -i /system/packages/libgit2-0.26.2-1-x86_64.hpkg | grep lib:libgit2
	provides: lib:libgit2 = 0.26.2 (compatible >= 26)

~> package list -i ktexteditor-5.39.0-1-x86_64.hpkg | grep libgit2
	requires: lib:libgit2>=0.26.2

~> pkgman install ktexteditor-5.39.0-1-x86_64.hpkg 
Encountered problems:
problem 1: nothing provides lib:libgit2>=0.26.2 needed by ktexteditor-5.39.0-1
  solution 1:
    - do not install "pkg:ktexteditor-5.39.0-1-x86_64"
Please select a solution, skip the problem for now or quit.
select [1/s/q]: 
```
The "fix" is easy:
After I revert my previous commit with this change:
```diff
-libVersionCompat="$libVersion compat >= 26"
+libVersionCompat="$libVersion compat >= ${libVersion%.*}"
```
(which will set `libVersionCompat` to `0.26.2 compat >= 0.26`)
I can install ktextktexteditor:
```
~> pkgman install ktexteditor*.hpkg
  in system:
    install package ktexteditor-5.39.0-1 from local file
    install package ktexteditor_devel-5.39.0-1 from local file
Continue? [yes/no] (yes) : [system] Applying changes ...
```
Hopefully, this is similar to the declaration we have right now for `libgit2-0.26.0`.

BTW, this means we won't need to touch any of the 3 recipes which depend on libgit2.

PR #2308, opened 2018-03-11
Openblas: Fix pkgconfig, add debuginfo

extrowerk:


sci-libs/openblas/openblas-0.2.20.recipe
```diff
@@ -64,6 +69,9 @@ INSTALL()
 	prepareInstalledDevelLibs libopenblas libopenblasp-r$portVersion
 
 	cp -P $developLibDir/libopenblas.so.0 $libDir
+
+	fixPkgconfig
```
fbrosson: I'd move the `fixPkgconfig` just below the `prepareInstalledDevelLibs`.

BTW, I'm wondering if the `cp` is really needed.

PR #2312, opened 2018-03-11
jpeg: Fixpkgconfig, cleanup

extrowerk:


media-libs/jpeg/jpeg-9c.recipe
```diff
@@ -58,9 +58,7 @@ BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	"
 BUILD_PREREQUIRES="
-	cmd:aclocal
-	cmd:autoconf
-	cmd:automake
```
fbrosson: `cmd:aclocal` and `cmd:automake` are both in the automake package, and the autoconf package (which currently provides `cmd:autoreconf` does not depend on automake, so I think the only `cmd:` you can safely remove here is `cmd:autoconf`.
In other words, I think you should keep `cmd:aclocal` and `cmd:automake`, so that `cmd:autoreconf` will find them.
(I haven't checked this yet, so I might be wrong.)

media-libs/jpeg/jpeg-9c.recipe
```diff
@@ -59,8 +59,8 @@ BUILD_REQUIRES="
 	"
 BUILD_PREREQUIRES="
 	cmd:aclocal
-	cmd:autoconf
 	cmd:automake
```
alaviss:
```diff
-	cmd:aclocal
-	cmd:automake

fbrosson: @alaviss, are you sure about this?
I might be wrong, but I think we need to keep them, as I wrote [here](https://github.com/haikuports/haikuports/pull/2312#discussion_r173657673).
What do you think?

alaviss: It's up to `cmd:autoreconf` to defines what it needs, that's the entire meaning of the dependency system. IIRC we've discussed about this before.

fbrosson: Yep, I know.
I might be wrong, but I think it would only be OK to drop the `cmd:aclocal` and `cmd:automake` in jpeg's `BUILD_PREREQUIRES` if they are in autoconf's `REQUIRES`. But they are not there.
Maybe we should add `cmd:aclocal` and `cmd:automake` to autoconf's `REQUIRES`, but I guess there must be a reason why they are not there.
So, until we change autoconf's `REQUIRES` I believe we should `cmd:aclocal` and `cmd:automake` in every `BUILD_PREREQUIRES` that also has `cmd:autoreconf`.

Remark: I might be missing something.

pulkomandy: The rules are still the same:
- If the *recipe* or the *buildsystem* runs aclocal or autoreconf directly (for example from a bootstrap.sh or autogen.sh or similar script), then it should be in the BUILD_PREREQUIRES
- If the recipe just calls autoconf and *autoconf* decides on its own to run aclocal or autoreconf, then it means that autoconf depends on aclocal or autoreconf, and should REQUIRE them.

fbrosson: OK, thanks.
If I'm not mistaken, autoconf does not call aclocal nor automake.
auto**re**conf, however, may call autoheader, aclocal, automake, autopoint and libtoolize.
So if we want to apply the rules, we should either:

1) add `cmd:autoheader`, `cmd:aclocal`, `cmd:automake`, `cmd:autopoint` and `cmd:libtoolize` to the autoconf's `REQUIRES`, or

2) move `cmd:autoreconf` to a subpackage and add the `cmd:*` mentioned above to that subpackage's `REQUIRES`.

So, should we do that? Or do we continue with our current exception, and keep `cmd:aclocal` and `cmd:automake` in each recipe that calls autoreconf?

pulkomandy: I would start with solution 1, and consider splitting the package if we notice it is annoying (for example if autoreconf is rarely used and people just wanted to run autoconf, and they are annoyed that they have to install all these dependencies they won't use)

This is an improvement because if in later autoconf version something changes (a requirement is added or removed), we have just one recipe to change, not everything that uses autoreconf.

fbrosson: I agree that this would (probably) be an improvement, but it might also have more downsides than advantages, so I'd rather leave this change to someone else :)

BTW, until such a change occurs, I think we should keep `cmd:aclocal` and `cmd:automake` in jpeg's `BUILD_PREREQUIRES`.

alaviss: > BTW, until such a change occurs, I think we should keep `cmd:aclocal` and `cmd:automake` in jpeg's `BUILD_PREREQUIRES`.

Let's simplify this a bit. Is it possible to build the package with those two left out? If yes, please remove them.

About changes to `autoconf`, as only `cmd:autoreconf` appears to require those `cmd:`s, I think we should consider splitting the package.

fbrosson: > Let's simplify this a bit.

I'm OK with removing `cmd:aclocal` and `cmd:automake`, but imho we can wait until these are added as dependencies of `cmd:autoreconf`.
The fact that autoreconf does not need the tools from the automake package for **_this_** particular release of jpeg does not mean that it will also be the case for the next releases, so later we might have to either add back `cmd:aclocal` and `cmd:automake` to jpeg's `BUILD_PREREQUIRES` or to speed up the changes/split in autoconf.

Until we have fixed the dependencies of `cmd:autoreconf` I think we should just replace the `cmd:autoconf` by `cmd:autoreconf` on those recipes which call autoreconf.

There's something else we could do: focus this PR on one change only: adding the call to `fixPkgconfig` in `INSTALL()`.

**Pros:**
* Recipe fixed with a minimalistic change.
* No risk that a future release becomes broken.

**Cons:**
* Another change might be required once the dependencies of `cmd:autoreconf` are fixed. (The change would be required if we split autoconf; it would be recommended if we don't split autoconf.)

If we drop `cmd:aclocal` and `cmd:automake` before the dependencies of `cmd:autoreconf` are fixed then the Pros/Cons are:
**Pros:** One commit.
**Cons:** Until we fix the dependencies of `cmd:autoreconf` we are not assured that jpeg's `BUILD_PREREQUIRES` will still work for the next releases.

Now, having said all this, if this release builds without the `cmd:aclocal` and `cmd:automake` then I'm OK with their removal. (It's just that it's not my prefered solution.)

@extrowerk, please feel free to drop them. And sorry for all these long discussions, btw :)

PR #2319, opened 2018-03-12
unrar: bump to 5.6.1, build sfx & libunrar, shrink patchset.

fbrosson:


app-arch/unrar/unrar-5.6.1.recipe
```diff
@@ -32,6 +32,8 @@ ARCHITECTURES="x86_gcc2 x86 x86_64"
 PROVIDES="
 	unrar = $portVersion
 	cmd:unrar = $portVersion
+	lib:libunrar = $portVersion
+	devel:libunrar = $portVersion
```
alaviss: Why isn't `devel:libunrar` resides in it's own package?

fbrosson: Yep, you are right. Keeping the static lib in the base package (instead of in a `unrar_devel` subpackage) is against the general recommendation.
I did so because I think that creating an unrar_devel package just for the static lib would have been worse than putting the static lib in the base package.
BTW, the main advantage of keeping everything in the base package is that we can decide to drop the static lib at any time, without having to insert any `REPLACES`.
Indeed, if I had created a devel package, then we would be forced to include a `REPLACES="unrar_devel"` if we later decide to retire that subpackage.

BTW, I thought about removing the static lib, but then I thought that it would be OK to keep it in the base package.

Well, I see 3 scenari:
1) We keep the static lib in the base package.
2) We drop the static lib.
3) We move the static lib to a devel subpackage.

I'm OK with 1 and 2. Scenario 3 is not my favorite. Upstream does not even make it clear which header files would be required to use the lib.
BTW, my first intention was to provide the SFX file, and then I thought that also providing the shared lib would be cool. And I ended with also providing the static lib :)

FWIW, the SFX file (which allows to create self extracting archives with the `cat` command, as we know) works like a charm :)

fbrosson: I just realized that there are no downsides with creating an `unrar_devel` package for the static lib.

In case there is a consensus for scenari 2 or 3, shall I wait untl next release becomes available?

pulkomandy:
On Fri, Mar 16, 2018 at 10:56:24AM -0700, fbrosson wrote:
> Well, I see 3 scenari:
> 1) We keep the static lib in the base package.

Static libs are useful to developers only, so if we ship them, we put
them in the _devel package.

> 2) We drop the static lib.

If no one needs it, then that is the right thing to do

> 3) We move the static lib to a devel subpackage.
> 
> I'm OK with 1 and 2. Scenario 3 is not my favorite. Upstream does not even make it clear which header files would be required to use the lib.

Same as for the shared lib, right?

fbrosson:
> > 2) We drop the static lib.

> If no one needs it, then that is the right thing to do

I'm OK with dropping the static lib.

> Same as for the shared lib, right?

Well, until now nobody has been able to use `libunrar.so` because we never shipped it before.
So I'd keep it in the base package.

BTW, we have at least one recipe which provides `devel:lib*` entries in its base package. That's wireshark:
```
~> package list -i /system/packages/wireshark-2.4.2-1-x86_64.hpkg | grep devel:
	provides: devel:libwireshark = 10.0.2 (compatible >= 10)
	provides: devel:libwiretap = 7.0.2 (compatible >= 7)
	provides: devel:libwscodecs = 1.1.0 (compatible >= 1)
	provides: devel:libwsutil = 8.0.0 (compatible >= 8)

~> package list -p /system/packages/wireshark-2.4.2-1-x86_64.hpkg | grep develop/lib/lib
develop/lib/libwireshark.so
develop/lib/libwireshark.so.10
develop/lib/libwireshark.so.10.0.2
develop/lib/libwiretap.so
develop/lib/libwiretap.so.7
develop/lib/libwiretap.so.7.0.2
develop/lib/libwscodecs.so
develop/lib/libwscodecs.so.1
develop/lib/libwscodecs.so.1.1.0
develop/lib/libwsutil.so
develop/lib/libwsutil.so.8
develop/lib/libwsutil.so.8.0.0
```
So, shall we go for scenario 2?

PR #2320, opened 2018-03-12
easyrpg_libretro: an RPG Maker game player using the libretro API

kwyxz:
EasyRPG Player is a game interpreter to play RPG Maker 2000, 2003 and EasyRPG games. It uses the LCF parser library (liblcf) to read RPG Maker game data.


games-engines/easyrpg_libretro/easyrpg_libretro-0.5.1.0_20180216.recipe
```diff
@@ -0,0 +1,55 @@
...
+
+PROVIDES="
+	easyrpg_libretro$secondaryArchSuffix = $portVersion
+	addon:easyrpg_libretro$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	retroarch$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	"
```
fbrosson: It builds fine, and I can also select _"RPG Maker 2000/2003 (EasyRPG)"_ with _"Load Core"_ in retroarch, but the _"Start Core"_ menu does not appear with this core.

Trying to Investigate:
```
~> objdump -x /system/add-ons/libretro/easyrpg_libretro.so | grep NEEDED
  NEEDED               libiconv.so.2
  NEEDED               libstdc++.so.6
  NEEDED               libgcc_s.so.1
```
Trying a small change in the recipe:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	retroarch$secondaryArchSuffix
+	lib:libgcc_s$secondaryArchSuffix
 	lib:libiconv$secondaryArchSuffix
+	lib:libstdc++$secondaryArchSuffix
 	"
```
Result: not better :(

BTW, on Haiku, libgcc_s and libstdc++ are provided by the gcc_syslibs package.
(`pkgman search lib:libgcc_s` and `pkgman search lib:libstdc++` shows that gcc_syslibs is the package they belong to.)

Hmm, among all our recipes, the only one which has `lib:libgcc` (in its `REQUIRES`) is blender. And there are absolutely no recipes with `lib:libstdc++` (in its requires too).

So I guess we don't need to add them here. (But I might be wrong.)
I'm writing this in case someone thinks we should start mentioning these dependencies, if that's appropriatte, of course.

fbrosson:
@kwyxz, I guess this core does not depend on any external file. Am I right?
Does it work on your system when you build both retroarch and the core with haikuporter, of course?
(Well, I guess it does.)
If it does then I can merge.

BTW, I might have an incorrect retroarch.cfg.
Could you please try retroarch with the default config, available in `/boot/system/data/retroarch/retroarch.cfg.default`), after having made a backup of both `/boot/system/settings/retroarch.cfg` and `/boot/home/config/settings/retroarch/retroarch.cfg` ?

I wonder if the default retroarch.cfg is missing one or more important parameters.
Thanks!

kwyxz:
This core _does_ depend on external files: it is an RPG Maker Game _player_, hence you need to load an RPG Maker Game through "Load Content" once you've loaded the core :) as far as I can tell it works fine here

fbrosson:
OK, great. Let's merge then!

Issue #2270, opened 2018-03-04
USER_SETTINGS_FILES is being used incorrectly


alaviss:
This issue exists to clarify the exact uses of `USER_SETTINGS_FILES` as well as a tracker for ports that use this variable wrongly

According to https://github.com/haiku/haiku/commit/fe707a23fc6dc0bccb3e80201bf3e4146dcde74e

> #### hpkg format: Add attributes for declaring settings files 
>
> Global and user settings files can be declared. For global ones an
update policy can be specified. If not specified, the settings file is
not included in the package, but created by the program (or user) later.
If an update type is specified, it defines what to do with the settings
file when updating the package to a newer version.
>
> User settings files are never included in the package; they are always
created by the program or the user. If the package contains a template/
default settings file, it can be declared, but for informative purposes
only.

`global-settings-files` was later renamed into `global-writable-files` in https://github.com/haiku/haiku/commit/fe707a23fc6dc0bccb3e80201bf3e4146dcde74e

This means that `user-settings-files` refer to settings files that reside in the user home directory (`/home/user/settings`), **not** the ones in global settings directories (`/home/user/config/settings/global`, `/system/settings`)

A popular use case of `user-settings-files` in HaikuPorts appears to be a way of "informing" pkgman about read-only templates for global settings files , and as @fbrosson [put it][0]:

> I might be wrong, but what I have understood from the recipes I've seen so far is that `USER_SETTINGS_FILES` is used to inform pkgman that a given config in `/system/settings/` has a a read-only template. Here we could use `/system/data/retroarch/retroarch.cfg.default` for that default.

Recipe that use `USER_SETTINGS_FILES` wrongly:
- [ ] [app-crypt/johntheripper]
- [ ] [app-text/mandoc]
- [ ] [net-mail/dovecot]
- [ ] [net-misc/openssh]

Possible wrong usage:
- [app-crypt/p11_kit]
- [app-shells/bash_completion]
- [dev-util/ccache]
- [net-misc/geoipupdate]
- [net-misc/wget]
- [net-nds/openldap]
- [net-vpn/tor]
- [www-client/lynx]
- [www-servers/lighthttpd]

[0]: https://github.com/haikuports/haikuports/pull/2248#discussion_r170460360
[app-crypt/johntheripper]: https://github.com/haikuports/haikuports/blob/07b75bf701e63e6e09445c08a07c4e5f7c70d87b/app-crypt/johntheripper/johntheripper-1.8.0.recipe#L69
[...]

PR #2321, opened 2018-03-12
lighttpd: bump to 1.4.49.

fbrosson:


alaviss:
We should consider using `addOnsDir` for lighthttpd modules.

Also, please check if the use of `USER_SETTINGS_FILES` here is correct. It's a suspect in #2270

fbrosson:
If we change the `lib:mod_*` into `addon:mod_*` then we would also need to move the modules to `$addOnsDir/lighttpd/`, but this does not seem to be a good idea because on most OSes the modules are under the lighttpd subdirectory of `$prefix/lib/`.

That said, if there is some consensus to move to the `$addOnsDir` then I'm OK with it.

Regarding the `USER_SETTINGS_FILES`, I will try to see what can be done.
But if you already have any hints on how this should be changed, they are very welcome ;)

alaviss:
> If we change the `lib:mod_*` into `addon:mod_*` then we would also need to move the modules to `$addOnsDir/lighttpd/`, but this does not seem to be a good idea because on most OSes the modules are under the lighttpd subdirectory of `$prefix/lib/`.
>
> That said, if there is some consensus to move to the `$addOnsDir` then I'm OK with it.

We're not other OSes. We should do what is correct, that is, to put addons where they should be.

> Regarding the `USER_SETTINGS_FILES`, I will try to see what can be done.
But if you already have any hints on how this should be changed, they are very welcome ;)

If it's confirmed that `lighthttpd` doesn't read configuration files within the user home directory (`/home/user/config/settings`), then `USER_SETTINGS_FILES` should be removed, and for every file in `USER_SETTINGS_FILES` that is not in `GLOBAL_WRITABLE_FILES`, please add the path there and left out the "action" (`keep-old`, `auto-merge`, etc.).

fbrosson:
> We're not other OSes. We should do what is correct, that is, to put addons where they should be.

:+1:

Currently, the paths we put in lighttpd's default config are meant for a system-wide installation.
I don't know if this would be a good idea, nor if it would be worth, but I imagine we could replace these absolute paths by relative ones, so that the package would also be usable when installed in the user's space.
I guess this would be tricky. Do you think we should try that? Or just decide that lighttpd is only usable when installed system-wide?

alaviss:
> Currently, the paths we put in lighttpd's default config are meant for a system-wide installation.
I don't know if this would be a good idea, nor if it would be worth, but I imagine we could replace these absolute paths by relative ones, so that the package would also be usable when installed in the user's space.
I guess this would be tricky. Do you think we should try that? Or just decide that lighttpd is only usable when installed system-wide?

The only directory constant that we could depends on is `sysconfDir`. For others, unfortunately, we can only obtain them reliably through `find_path()`, which mean we would need to patch `lighttpd` for that. 

I'd say we should replace `constantSystemSettingsDir` with `sysconfDir`, and left the others as is for now.

Also, don't forget to reset `REVISION`.

fbrosson:
Thanks for the hint about the `sysconfDir`. I'll dig this.

BTW, I'm quite sure it is possible to configure lighttpd without any absolute paths.
If this is doable then we can create a package that can work no matter how it is installed, system-wide or user-only.

> Also, don't forget to reset `REVISION`.

Good catch, thanks!

fbrosson:
After all, although modules and plugins are both shared libs, a module is _not_ a plugin.
BTW, for Apache, for instance, we are not declaring the modules as plugins.

alaviss:
> After all, although modules and plugins are both shared libs, a module is not a plugin.

A module **is** a plugin. They extend the core functions of the program, and can be loaded dynamically. This is expressed throughout lighttpd source, check any `mod*` source file and you will find `plugin` everywhere.

> BTW, for Apache, for instance, we are not declaring the modules as plugins.

Then I believe we have to update Apache to declare modules as plugins as well, as Apache modules works the same way as lighttpd modules

fbrosson:
Well, I'm not against the change. I just think this is a polical issue, and that there are more downsides than advantages.
BTW, I'm not sure but I think lighttpd is used for testing https://github.com/haiku/webkit, and that it references some modules with their current installation directory, so changing lighttpd would require to change some files in webkit's test suite.

Anyway, if you would like to make the changes for lighttpd and/or apache, please do not hesitate. Other contributors might even agree with this idea.
(As said earlier, I don't disagree. I just think it would be extra work for no real benefit.)

PR #2327, opened 2018-03-14
dosbox_libretro: a port of dosbox to the libretro API

kwyxz:
This is the libretro core for DosBox, the famous DOS emulator.
I also took the liberty of moving the original "dosbox" port from games-emulation to app-emulation as I feel it corresponds better to where it should be, please let me know if this was a bad idea.


pulkomandy:
> I also took the liberty of moving the original "dosbox" port from games-emulation to app-emulation as I feel it corresponds better to where it should be, please let me know if this was a bad idea.

We try to follow Gentoo's organization, even when they make strange
choices.

http://gpo.zugaina.org/games-emulation/dosbox

So, please don't move things around :)

kwyxz:
Fair enough, I will move dosbox back then. It's weird to me to see mednafen in app-emulation and dosbox in games-emulation though :/

fbrosson:
Thanks @kwyxz for your PR!
It builds fine and I can choose _"Load Core"_.
When I choose _"Start Core"_, however, it makes retroarch crash on my system.
(I think I need to update my system to avoid this crash.)
I guess it works for you.  If it does then I can merge :)

kwyxz:
Yeah it works and opens a nice DOS prompt for me. I'm kind of concerned that so many cores seem to crash your system though :/

fbrosson:
No problem. I'm quite sure I just need to `pkgman update` my system.

PR #2329, opened 2018-03-15
fmsx_libretro: a port of fMSX to the libretro API

kwyxz:
Written by legendary emulator author Marat Fayzullin, fMSX was among the first, if not the very first, MSX emulator for x86 computers, its development started in 1993. This is a libretro port of this emulator.


fbrosson:
Thanks @kwyxz for this PR!
It builds fine, and I can load the core on retroarch, but, surprisingly, it shows the _"Start Core"_ menu although I did not download any external file. (When I select it, retroarch crashes.)
Of course, I'm perfectly ok with being able to select the _"Start Core"_ menu even if the roms are not available.
I'm just reporting this in case the _"Start Core"_ menu should only be shown if the firmware files have been found.
Shall I merge anyway?

kwyxz:
Yeah the core should be able to boot the system BIOS files if they are present when the core is started, even if no disk is loaded. If I select Start Core here without any BIOS files I get a black screen and nothing happens.

PR #2330, opened 2018-03-15
freeintv_libretro: a Mattel Intellivision core for libretro

kwyxz:
This emulator seems to have been written with libretro in mind from the ground up. Quoting the author: "FreeIntv is a libretro core Intellivision Emulator, intended for use with RetroPie on a Raspberry Pi 3. The primary goal was to allow my wife to play Astrosmash without needing to use a keyboard." :-) 

The core builds and works fine with Haiku too.


games-emulation/freeintv_libretro/freeintv_libretro-20180214.recipe
```diff
@@ -0,0 +1,52 @@
+SUMMARY="A Mattel Intellivision emulator using the libretro API"
+DESCRIPTION="FreeIntv is a libretro emulation core for the Mattel \
+Intellivision designed to be compatible with joypads from the SNES era \
+forward even if they originally required a number pad."
```
fbrosson: Everything is OK!

Well, if you could move _backwards_ the word "forward" it would excellent :smile:
```diff
-Intellivision designed to be compatible with joypads from the SNES era \
+Intellivision designed to be compatible with joypads from the SNES era forward \
-forward even if they originally required a number pad."
+even if they originally required a number pad."
```
(Yep, the line will be exactly 80 chars, including the backslash.)

(This does not make a real difference. It's just that it would be better to "fix" this before the merge rather than leaving it as is.)
Thanks!

kwyxz: Sure. Done!

PR #2334, opened 2018-03-16
curl: bump to 7.59.0, depend on lib{brotlidec,idn2,psl}.

fbrosson:
Added comments after the definition of `TEST_PREREQUIRES`. They should be removed if haikuports/haikuporter/issues/151 is implemented.

### DIFF
--- a/net-misc/curl/curl-7.59.0.recipe
+++ b/net-misc/curl/curl-7.59.0.recipe
@@ -11,7 +11,7 @@ COPYRIGHT="1996-2018 Daniel Stenberg"
 LICENSE="Curl"
 REVISION="1"
 SOURCE_URI="https://curl.haxx.se/download/curl-$portVersion.tar.bz2"
-CHECKSUM_SHA256="1cb081f97807c01e3ed747b6e1c9fee7a01cb10048f1cd0b5f56cfe0209de731"
+CHECKSUM_SHA256="b5920ffd6a8c95585fb95070e0ced38322790cb335c39d0dab852d12e157b5a0"
 ADDITIONAL_FILES="curl.rdef"
 
 ARCHITECTURES="x86_gcc2 ?x86 x86_64 ?arm"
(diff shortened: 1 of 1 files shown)


fbrosson:
Do you all agree with the new dependencies?

Usefulness | Comments
-- | --
lib**psl** is very welcome. | libpsl passes all tests on all architectures where the tests can be built.
lib**idn2** is welcome. | libidn2 is not available on x86_gcc2, but it passes all tests on all other architectures.
lib**brotli**dec is a nice to have. | brotli passes all tests on all architectures.

I see 4 scenari:
1) We add lib**psl**, lib**idn2** and lib**brotli**dec. (This is what we have right now in this PR.)
2) We only add libpsl and libidn2.
3) We only add libpsl.
4) We don't add any of these dependencies.

I think scenario 1 is fine, although I'm OK with all others, including scenario 4. Any thoughts?
Please let me know if you would prefer any other scenario. Thanks!

mmlr:
Please note that libcurl is a dependency of the main Haiku package. Adding more dependencies to curl will further widen the indirect dependencies needed to build Haiku and should therefore be considered wisely.

fbrosson:
OK, I see. Shall I update this PR for scenario 4 right now?
Or do we wait a little before we decide what is the best scenario?

Would it make sense to add a libcurl**3** recipe with the most recent release of curl which provides libcurl.so.**3** and use that as the build dependency for Haiku?
This would allow to add extra dependencies to curl 7.59.0.
Or would this be worse than all the other scenari?

alaviss:
@fbrosson Can you provides details on what each of those added dependencies do?

fbrosson:
* libpsl would be useful with privacy protection when accepting coockies. https://rockdaboot.github.io/libpsl/
* libidn2 would allow curl to handle Internationalized Domain Names (IDN). https://gitlab.com/libidn/libidn2
* libbrotlidec would enable curl to receive data compressed with brotli. https://github.com/google/brotli

None of these are really needed, but since they all seem to work fine, it might be worth adding them.
Well, this is a political question. Do we want Haiku to depend on these packages?
I don't have any opinion here. I'm just suggesting, in case this could be a good idea.

mmlr:
Splitting out a libcurl package would not help, as the dependencies are most probably linked into that one. A libcurl-minimal could be provided that conflicts with the full one. The question is whether the added features are relevant for the builtin use of libcurl in Haiku. At least the package kit uses it as far as I remember.

I also have not checked if any of those are already dependencies of Haiku for other reasons.

fbrosson:
OK. Shall I update the PR for scenario 4, then?
Or shall we give ourselves more time to decide whether or not it would be worth adding some (or all) of the suggested dependencies?

BTW, libpsl, libidn2 and brotli were all added during our last GCI :)

fbrosson:
We could probably add libidn2, but let's keep things simple, after all, and go for scenario 4 ;)

PR #2336, opened 2018-03-16
Libjpegturbo : new recipe

extrowerk:


media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,129 @@
...
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libjpegVersion="62.2.0"
+libjpegVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+libturbojpegVersion="0.1.0"
+libturboVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```
fbrosson:
```diff
 libjpegVersion="62.2.0"
-libjpegVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+libjpegVersionCompat="$libjpegVersion compat >= ${libVersion%%.*}"
 libturbojpegVersion="0.1.0"
-libturboVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+libturbojpegVersionCompat="$libturbojpegVersion compat >= ${libVersion%%.*}"
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,129 @@
...
+	jpeg_turbo${secondaryArchSuffix}_devel = $portVersion
+	devel:libjpeg$secondaryArchSuffix = $libjpegVersion
+	devel:libturbojpeg$secondaryArchSuffix = $libturbojpegVersion
+	"
+REQUIRES_devel="
+	jpeg_turbo$secondaryArchSuffix == $portVersion base
+	"
+
+PROVIDES_tools="
+	jpeg_tools = $portVersion
```
fbrosson:
```diff
 PROVIDES_tools="
-	jpeg_tools = $portVersion
+	jpeg_turbo${secondaryArchSuffix}_tools = $portVersion
```

fbrosson: BTW, it seems we also need a
```
CONFLICTS="
	jpeg$secondaryArchSuffix
	"
```

```
CONFLICTS_devel="
	jpeg${secondaryArchSuffix}_devel
	"
```

```
CONFLICTS_tools="
	jpeg${secondaryArchSuffix}_tools
	"
```

(Each of these 3 blocs should be added just below `REQUIRES`, `REQUIRES_devel` and `REQUIRES_tools`).

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+of making high-speed JPEG compression/decompression technology available to \
+a broader range of users and developers."
+HOMEPAGE="https://www.libjpeg-turbo.org/"
+COPYRIGHT="1991-2018 Thomas G. Lane, Guido Vollbeding"
+LICENSE="BSD (3-clause)
+	IJG"
+REVISION="1"
+SOURCE_URI="https://github.com/libjpeg-turbo/libjpeg-turbo/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="1a17020f859cb12711175a67eab5c71fc1904e04b587046218e36106e07eabde"
+SOURCE_DIR="libjpeg-turbo-$portVersion"
```
fbrosson:
```diff
+SOURCE_FILENAME="libjpeg-turbo-$portVersion.tar.gz"
 SOURCE_DIR="libjpeg-turbo-$portVersion"
```
(Everything else looks OK now, but I haven't tested the recipe yet.)

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+PROVIDES="
+	jpeg_turbo$secondaryArchSuffix = $portVersion
+	lib:libjpeg$secondaryArchSuffix = $libjpegVersion
+	lib:libturbojpeg$secondaryArchSuffix = $libturbojpegVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+CONFLICTS="
+	jpeg$secondaryArchSuffix
```
alaviss: Shouldn't this be `lib:libjpeg`?

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+libjpeg-turbo was originally based on libjpeg/SIMD, an MMX-accelerated \
+derivative of libjpeg v6b developed by Miyasaka Masaru. The TigerVNC and \
+VirtualGL projects made numerous enhancements to the codec in 2009, and in \
+early 2010, libjpeg-turbo spun off into an independent project, with the goal \
+of making high-speed JPEG compression/decompression technology available to \
+a broader range of users and developers."
+HOMEPAGE="https://www.libjpeg-turbo.org/"
+COPYRIGHT="1991-2018 Thomas G. Lane, Guido Vollbeding"
+LICENSE="BSD (3-clause)
+	IJG"
```
alaviss: We have a similiar license around called [Jpeg], should we use the same name for this port?

[Jpeg]: https://github.com/haikuports/haikuports/blob/c9e40d638522692285e21d08ee91c8371d7abe87/media-libs/jpeg/licenses/JPEG#L101

alaviss: The source is also licensed under the [Zlib license](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/1.5.3/LICENSE.md)

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+big-endian pixel buffers (RGBX, XBGR, etc.), as well as a full-featured Java \
+interface.
+libjpeg-turbo was originally based on libjpeg/SIMD, an MMX-accelerated \
+derivative of libjpeg v6b developed by Miyasaka Masaru. The TigerVNC and \
+VirtualGL projects made numerous enhancements to the codec in 2009, and in \
+early 2010, libjpeg-turbo spun off into an independent project, with the goal \
+of making high-speed JPEG compression/decompression technology available to \
+a broader range of users and developers."
+HOMEPAGE="https://www.libjpeg-turbo.org/"
+COPYRIGHT="1991-2018 Thomas G. Lane, Guido Vollbeding"
```
alaviss: There are more holders, please inspect the source code for a complete list

**TODO**: Make a copyright extractor script.

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+CONFLICTS_tools="
+	jpeg${secondaryArchSuffix}_tools
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
```
alaviss:
```diff
-	cmd:autoconf
+	cmd:autoreconf

File: media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe [hunk 8]
```diff
@@ -0,0 +1,138 @@
...
+		--bindir $commandBinDir \
+		--enable-shared
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm $libDir/libjpeg.la
```
alaviss: Does `libturbojpeg` also have an libtool archive?

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
fbrosson: Hmm, **_if_** we do switch from `jpeg$secondaryArchSuffix` to `lib:libjpeg$secondaryArchSuffix` **_then_** we should probably **_also_** include `lib:libturbojpeg$secondaryArchSuffix`.
```diff
 CONFLICTS="
-	jpeg$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libturbojpeg$secondaryArchSuffix
 	"
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+PROVIDES_devel="
+	jpeg_turbo${secondaryArchSuffix}_devel = $portVersion
+	devel:libjpeg$secondaryArchSuffix = $libjpegVersion
+	devel:libturbojpeg$secondaryArchSuffix = $libturbojpegVersion
+	"
+REQUIRES_devel="
+	jpeg_turbo$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_devel="
+	jpeg${secondaryArchSuffix}_devel
```
fbrosson: I'm not sure this is a good idea. I'm just adding this comment in case we change `CONFLICTS`.
```diff
CONFLICTS_devel="
-	jpeg${secondaryArchSuffix}_devel
+	lib:libjpeg$secondaryArchSuffix
+	lib:libturbojpeg$secondaryArchSuffix
 	"
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,138 @@
...
+	cmd:rdjpgcom$commandSuffix = $portVersion
+	cmd:tjbench$commandSuffix = $portVersion
+	cmd:wrjpgcom$commandSuffix = $portVersion
+	"
+REQUIRES_tools="
+	haiku$secondaryArchSuffix
+	jpeg_turbo$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_tools="
+	jpeg${secondaryArchSuffix}_tools
```
fbrosson: I'm not sure this is a good idea. I'm just adding this comment in case we change `CONFLICTS` and `CONFLICTS_devel`.
```diff
 CONFLICTS_tools="
-	jpeg${secondaryArchSuffix}_tools
+	cmd:cjpeg$commandSuffix = $portVersion
+	cmd:djpeg$commandSuffix = $portVersion
+	cmd:jpegtran$commandSuffix = $portVersion
+	cmd:rdjpgcom$commandSuffix = $portVersion
+	cmd:tjbench$commandSuffix = $portVersion
+	cmd:wrjpgcom$commandSuffix = $portVersion
```

After all, shall we just keep the current `CONFLICTS`, `CONFLICTS_devel` and `CONFLICTS_tools` ?

It makes more sense to have only the package names. In case we ever change the jpeg package, then we would just need to adjust jpeg_turbo.
I'm not confortable with having `CONFLICTS` using `lib:libjpeg` and `CONFLICTS_tools` not using `cmd:` entries.
We should probably choose the same style for all the `CONFLICTS*` in a recipe.

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
alaviss: I think we should only defines `CONFLICTS` for stuff that conflicts with what currently in Haiku
Ports. There might be reasons behind why `PROVIDES` doesn't get translated automatically to `CONFLICTS`

fbrosson: :+1:
So, shall we keep the `CONFLICTS*` as we have them right now?
(That's enough to make sure we don't co-install jpeg_turbo and jpeg.)

fbrosson: I guess the reason why `PROVIDES` does not get translated automatically to `CONFLICTS` is to permit overlaps, in some cases, when having these overlaps is not a big issue.

BTW, I'm starting to think that we should (probably) recommend sticking to package names when we define any `CONFLICTS*`.

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+		--bindir $commandBinDir \
+		--enable-shared
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm $libDir/{libjpeg,libturbojpeg}.la
```
fbrosson: What do you think of this variant?
```diff
-	rm $libDir/{libjpeg,libturbojpeg}.la
+	rm -f "$libDir"/lib*jpeg.la
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+	2011 Siarhei Siamashka
+	2013 Linaro Limited
+	2014 MIPS Technologies, Inc.
+	2015 Google, Inc.
+	2015 Matthieu Darbois"
+LICENSE="BSD (3-clause)
+	IJG
+	Zlib"
+REVISION="1"
+SOURCE_URI="https://github.com/libjpeg-turbo/libjpeg-turbo/archive/$portVersion.tar.gz"
```
fbrosson: https://www.libjpeg-turbo.org/ has a link called "Official Binaries and Source Tarballs" which points to SourceForge.
Gentoo also uses the tarball from SF. So maybe we should use them too.
```diff
-SOURCE_URI="https://github.com/libjpeg-turbo/libjpeg-turbo/archive/$portVersion.tar.gz"
+SOURCE_URI="https://downloads.sourceforge.net/libjpeg-turbo/libjpeg-turbo-$portVersion.tar.gz"
-CHECKSUM_SHA256="1a17020f859cb12711175a67eab5c71fc1904e04b587046218e36106e07eabde" 
+CHECKSUM_SHA256="b24890e2bb46e12e72a79f7e965f409f4e16466d00e1dd15d93d73ee6b592523"
```
BTW, if you switch to the official tarball from SF we won't need the `SOURCE_FILENAME` anymore :)

Remark: I did not test yet if both archives are equivalent. I guess they are, but I need to kick a build.

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+CHECKSUM_SHA256="1a17020f859cb12711175a67eab5c71fc1904e04b587046218e36106e07eabde"
+SOURCE_FILENAME="libjpeg-turbo-$portVersion.tar.gz"
+SOURCE_DIR="libjpeg-turbo-$portVersion"
+PATCHES="libjpeg_turbo-$portVersion.patchset"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64 arm"
```
fbrosson:
```diff
-ARCHITECTURES="?x86_gcc2 x86 x86_64 arm"
+ARCHITECTURES="x86_gcc2 x86 x86_64 ?arm"
 SECONDARY_ARCHITECTURES="x86"
-
-commandSuffix=$secondaryArchSuffix
-commandBinDir=$binDir
-if [ "$targetArchitecture" = x86_gcc2 ]; then
-	commandSuffix=
-	commandBinDir=$prefix/bin
-fi
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+CONFLICTS_devel="
+	jpeg${secondaryArchSuffix}_devel
+	"
+
+PROVIDES_tools="
+	jpeg_turbo${secondaryArchSuffix}_tools = $portVersion
+	cmd:cjpeg$commandSuffix = $portVersion
+	cmd:djpeg$commandSuffix = $portVersion
+	cmd:jpegtran$commandSuffix = $portVersion
+	cmd:rdjpgcom$commandSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES_tools="
 	jpeg_turbo${secondaryArchSuffix}_tools = $portVersion
-	cmd:cjpeg$commandSuffix = $portVersion
-	cmd:djpeg$commandSuffix = $portVersion
-	cmd:jpegtran$commandSuffix = $portVersion
-	cmd:rdjpgcom$commandSuffix = $portVersion
-	cmd:tjbench$commandSuffix = $portVersion
-	cmd:wrjpgcom$commandSuffix = $portVersion
+	cmd:cjpeg$secondaryArchSuffix = $portVersion
+	cmd:djpeg$secondaryArchSuffix = $portVersion
+	cmd:jpegtran$secondaryArchSuffix = $portVersion
+	cmd:rdjpgcom$secondaryArchSuffix = $portVersion
+	cmd:tjbench$secondaryArchSuffix = $portVersion
+	cmd:wrjpgcom$secondaryArchSuffix = $portVersion
 	"
```
(`$commandSuffix` -> `$secondaryArchSuffix`)

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+	"
+
+defineDebugInfoPackage jpeg_turbo$secondaryArchSuffix \
+	$libDir/libjpeg.so.$libjpegVersion \
+	$libDir/libturbojpeg.so.$libturbojpegVersion
+
+BUILD()
+{
+	autoreconf -fi
+	runConfigure --omit-dirs binDir ./configure \
```
fbrosson:
```diff
-	runConfigure --omit-dirs binDir ./configure \
-		--bindir $commandBinDir \
-		--enable-shared
+	runConfigure ./configure
```
Why? Because `[default=yes]` for `--enable-shared` as we can see with:
```
./configure --help | grep -- --enable-shared
  --enable-shared[=PKGS]  build shared libraries [default=yes]
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,149 @@
...
+
+	prepareInstalledDevelLibs libjpeg libturbojpeg
+
+	# devel package
+	packageEntries devel \
+		$developDir
+
+	# tools package
+	packageEntries tools \
+		$commandBinDir \
```
fbrosson:
```diff
	packageEntries tools \
-		$commandBinDir \
+		$binDir \
 		$documentationDir
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
alaviss: Better yet

    rm -f "$libDir"/*.la

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,139 @@
...
+LICENSE="BSD (3-clause)
+	IJG
+	Zlib"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/libjpeg-turbo/libjpeg-turbo-$portVersion.tar.gz"
+CHECKSUM_SHA256="b24890e2bb46e12e72a79f7e965f409f4e16466d00e1dd15d93d73ee6b592523"
+SOURCE_DIR="libjpeg-turbo-$portVersion"
+PATCHES="libjpeg_turbo-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
```
fbrosson:
```diff
-ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
+ARCHITECTURES="?x86_gcc2 ?x86 ?x86_64 ?arm"
```
I'd recommend marking all architectures as untested because when I restarted my x86_gcc2 system after having replaced `jpeg{,_tools,_devel}` by `libjpeg_turbo{,_tools,_devel}` Tracked was not able to start.
(I did not try rebooting my other Haiku installs (x86 and x86_64) so we might not have the problem on these architectures.

Anyway, unless someone investigates this problem on each arch, I think we should mark everything as untested.

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -0,0 +1,139 @@
...
+	"
+
+defineDebugInfoPackage jpeg_turbo$secondaryArchSuffix \
+	$libDir/libjpeg.so.$libjpegVersion \
+	$libDir/libturbojpeg.so.$libturbojpegVersion
+
+BUILD()
+{
+	autoreconf -fi
+	runConfigure --omit-dirs binDir ./configure
```
fbrosson: You forgot this tiny change:
```diff
-	runConfigure --omit-dirs binDir ./configure
+	runConfigure ./configure
```

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
alaviss: It's expected. `jpeg` is a dependency of `haiku`, and since `libjpeg_turbo` provides different soversion, it's not a surprise that things wouldn't work. To fix this, we would need to rebuild Haiku with `libjpeg_turbo`.

fbrosson: OK, thanks.
Well, I think it is safer to stick to jpeg until the core developers decide to switch to libjpeg_turbo.

alaviss: Now I've got to think if we even have to declare this package as conflicts with `jpeg`, given the different soname they carry.

For `_devel`, however, `CONFLICTS` still remains necessary.

fbrosson:
The archives from GitHub and SF are almost identical. The main difference is that the tarball from SF has several `Makefile.in` files, while that of GitHub does not.
Of course, all the sources are identical.
So I have tested the resulting package built with the official sources from SF.
Everything is OK on:
- [x] x86_64
- [ ] x86_gcc2
- [ ] x86

Once I have finished the tests on all architectures, shall I merge the PR?

fbrosson:
`GREAT SUCCESS!` on all architectures :smile:
Let's merge!

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
fbrosson: Sorry @alaviss, I had not noticed your comment.
You are right that both packages don't overlap.
FWIW:

package | jpeg | libjpeg_turbo
-- | -- | --
symlink | libjpeg.so.9 | libjpeg.so.62
shared lib | libjpeg.so.9.3.0 | libjpeg.so.62.2.0

Anyway, it might be safer to keep libjpeg_turbo marked as untested until everyone is OK with enabling it.

alaviss: I would say we enable it and drop the conflicts for `jpeg`. This way, we would allow one to build pkgs with `libjpeg_turbo` and install it alongside their `jpeg`-based Haiku.

pulkomandy:
On Sun, Mar 18, 2018 at 08:43:32AM -0700, fbrosson wrote:
> I'd recommend marking all architectures as untested because when I restarted my x86_gcc2 system after having replaced `jpeg{,_tools,_devel}` by `libjpeg_turbo{,_tools,_devel}` Tracked was not able to start.
> (I did not try rebooting my other Haiku installs (x86 and x86_64) so we might not have the problem on these architectures.
> 
> Anyway, unless someone investigates this problem on each arch, I think we should mark everything as untested.

That's not untested. That's very clearly tested, and broken!

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
fbrosson: Well, anyone willing to install libjpeg_turbo can do it locally. It's easy to do for those users who know what they are doing.

I'd say that having this small barrier is nice, because if anyone could install libjpeg_turbo without being aware of the possible consequences, then we would end up with 2 groups of users: those who have switched, and those who have not.
In such a world, whenever we need to test a package which depends on libjpeg, we would need to check that it's OK with both the _standard_ libjpeg (from the jpeg package) and with the alternate libjpeg (from libjpeg_turbo).

So I think we should only mark libjpeg_turbo as tested and drop the conflicts for its base package after one user, at least, has had the chance to rebuild (and test) everything against the libjpeg from libjpeg_turbo. Of course, we would also need to make sure Haiku's core developers are OK with the switch.

There's something else we need to check: will libjpeg_turbo build on arm and ppc? If not, maybe we should tweak libjpeg_turbo's pkconfig files and install libjpeg with an alternate name, so that packages which only build on `x86{,_gcc2,_x86}` and depend on libjpeg would be able to **_opt-in_** the dependency on libjpeg_turbo.

Now, I'm not against a switch. I just wish we don't do that before we are sure it won't break anything.
Regading the advances on Haiku for arm & ppc, I hope the possible libjpeg switch will not be an issue for these architectures.

fbrosson:
@pulkomandy, I have to admit that you are right.
Even if rebuilding Haiku against libjpeg_turbo's libjpeg might fix the issue we have with Tracker crashing at startup on a system where jpeg has been replaced by libjpeg_turbo, it's not done yet, so I guess we should mark libjpeg_turbo as broken.
So, shall I mark all architectures as broken?

extrowerk:
Maybe we should just adjust the ABI version: https://github.com/libjpeg-turbo/libjpeg-turbo/blob/master/configure.ac#L153

Which jpeg ABI does the current libjpeg provides? I fear it is 9, but it seems libjpeg-turbo provides only 6-7-8 ABI compatibility.

alaviss:
@extrowerk
It's 9, unfortunately. However, that also means we don't have to declare incompatibility with `libjpeg`.

Hopefully:

    libjpeg$secondaryArchSuffix == 8

should cover `CONFLICTS` if anyone still using the old jpeg 8

fbrosson:
package | symlink | shared lib
-- | -- | --
jpeg | `libjpeg.so.9` | `libjpeg.so.9.3.0`
libjpeg_turbo | `libjpeg.so.62` | `libjpeg.so.62.2.0`

As @alaviss suggested, dropping the `CONFLICTS` (on the base package only) might fix the issue we have with Tracker. I'll try that soon. But I don't know if it would be OK to mark libjpeg_turbo as tested, since it might cause other issues if we don't also rebuild Haiku against libjpeg_turbo's libjpeg.

pulkomandy:
On Mon, Mar 19, 2018 at 09:43:14AM +0000, alaviss wrote:
> @extrowerk
> It's 9, unfortunately. However, that also means we don't have to declare incompatibility with `libjpeg`.
> 
> Hopefully:
> 
>     libjpeg$secondaryArchSuffix == 8
> 
> should cover `CONFLICTS` if anyone still using the old jpeg 8

Yes, that should be fine. People can then use the _turbo version to
build their software if they want to. And someone could make a
libjpegturbo_translator (it should just be a rebuild of the existing
translator linked against the new lib) so that native apps can benefit
from it.

fbrosson:
:arrow_right: #2347

PR #2338, opened 2018-03-16
fuse_libretro: a port of the FUSE emulator to libretro

kwyxz:
FUSE is a ZX Spectrum emulator that can emulate nearly every Spectrum model on the market. This is the libretro port

screenshot_haiku_2018-03-16_16 15 12


app-emulation/fuse_libretro/fuse_libretro-1.1.1_20180310.recipe
```diff
@@ -0,0 +1,52 @@
+SUMMARY="A ZX Spectrum emulator core for libretro"
+DESCRIPTION="Fuse (the Free Unix Spectrum Emulator) was originally, and somewhat unsurprisingly, a ZX Spectrum emulator for Unix. However, it has now also been ported to libretro."
```
fbrosson:
```diff
-DESCRIPTION="Fuse (the Free Unix Spectrum Emulator) was originally, and somewhat unsurprisingly, a ZX Spectrum emulator for Unix. However, it has now also been ported to libretro."
+DESCRIPTION="Fuse (the Free Unix Spectrum Emulator) was originally, and \
+somewhat unsurprisingly, a ZX Spectrum emulator for Unix. However, it has now \
+also been ported to libretro."
```
BTW, **for the commit summary**, it would be nice to shrink it a little bit to make it <= 69 chars, including the ` (#2338)`.
If we use `fuse_libretro: a port of the FUSE Spectrum emulator to libretro (#2338)` then GitHub will only hide the `38)` unless we click on the "...", in which case it would be shown on the next line.

So, for the commit summary, what do you think of this?
```diff
-fuse_libretro: a port of the FUSE Spectrum emulator to libretro (#2338)
+fuse_libretro: a port of the ZX Spectrum emulator to libretro (#2338)
```

BTW, for your next PRs, you should probably replace "Initial commit" with a nice commit summary :)
Of course, this does not mean all our commits should be limited ot just 69 chars, so you are welcome to write longer commit summaries, as long as the commit title is <= 69 chars.
BTW, this is **_not_** a requirement. It's **_just_** a recommendation.

PR #2340, opened 2018-03-17
Libtorrent-Rasterbar: bump, cleanup, debuginfo

extrowerk:


net-libs/libtorrent-rasterbar/libtorrent_rasterbar-1.1.6.recipe
```diff
@@ -63,6 +66,9 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+defineDebugInfoPackage libtorrent_rasterbar$secondaryArchSuffix \
+	$libDir/libtorrent-rasterbar.so.$libVersion
+
 BUILD()
 {
 	sed -i 's/-msse4.2//' configure.ac
```
alaviss: This should be moved to `PATCH()`. Also, please check if it's even necessary, I didn't find any `-msse4.2` in `configure.ac`.

PR #2344, opened 2018-03-18
libvorbis: bump to 1.3.6, add debuginfo & TEST.

fbrosson:
* Switch `HOMEPAGE` and `SOURCE_DIR` to https.
* Drop the docs on 2nd arch.
* Preserve the subdirectories of `$developDocDir` instead of installing everything in a single directory.


media-libs/libvorbis/libvorbis-1.3.6.recipe
```diff
@@ -58,15 +70,29 @@ BUILD_PREREQUIRES="
...
+		doc/Makefile.am
+}
+
 BUILD()
 {
 	libtoolize -fci
 	mkdir -p m4
 	aclocal --install -I m4
 	autoconf
 	automake
```
alaviss: `autoreconf` could be used here to shorten the recipe.

fbrosson: Well, I did try using autoreconf, but it did not work as expected, so I decided that the best thing do do was not to shorten the recipe but to limit the changes to the minimum.
BTW, I already forgot what errors I bumped into when I tried using autoreconf.

PR #2347, opened 2018-03-19
libjpeg-turbo: drop conflicts for the base package.

fbrosson:


media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
```diff
@@ -55,6 +55,8 @@ REQUIRES="
 	haiku$secondaryArchSuffix
 	"
 
+ARCHITECTURES_devel="?x86_gcc2 ?x86 ?x86_64 ?arm"
+SECONDARY_ARCHITECTURES_devel="?x86"
```
fbrosson: Should I drop the "untested" flag for the devel subpackage?
```diff
-ARCHITECTURES_devel="?x86_gcc2 ?x86 ?x86_64 ?arm"
-SECONDARY_ARCHITECTURES_devel="?x86"
```

fbrosson: Or would it be better to just drop the `CONFLICTS` for the base package (as I did in the very first commit of this PR) and keep the untested flag for everything, i.e. `libjpeg_turbo{,_tools,_devel}`?
In other words, should I completely revert the second commit of this PR?

fbrosson:
Sorry @Begasus for the "late" review request. I was sure I had included you in the first round, but it was not the case...

media-libs/libjpeg-turbo/libjpeg_turbo-1.5.3.recipe
fbrosson: I dropped the second commit. Let's keep the "untested" flag for now.
(We can change it to "tested" later.)

PR #2349, opened 2018-03-20
gme_libretro: a libretro port of the Game Music Emu library

kwyxz:
Game_Music_Emu emulates game music in several popular file formats. It allows playout of sound files as if they were being issued by the original hardware.


kwyxz:
I get no error when I load the core... did you end up updating your Haiku system?

fbrosson:
> I get no error when I load the core...

Cool. If both retroarch and the core you are using were built with haikuporter then I guess we can merge. (I guess they are.)

> did you end up updating your Haiku system?

Not yet... I guess this explains my error.

kwyxz:
Yeah both retroarch and the core I use are from the packages built through Haikuporter

fbrosson:
Whoops, I have updated my system to the latest hrev, but I still get the error when I choose _Load Core_:
```bash
~> retroarch
resolve symbol "fseeko64" returned: -2147478780
Unlocking BLocker with sem 52603 from wrong thread 1057, current holder -1 (see issue #6400).
~> uname -a
Haiku dual 1 hrev51856 Mar 17 2018 14:24:19 x86_64 x86_64 Haiku
```
(The error is not shown on the GUI but, as you can see, on the console from where I launch retroarch.)

All the packages I have installed were built by Haiku's build servers.

Do you get the `resolve symbol "fseeko64"` error if you also use the packages from Haiku's build servers?
In case you get the error, does this prevent using the core?

(BTW, I did not manage to start any core, not even DOSBox. I tried starting the DOSBox core with retroarch under Haiku x86_64, both in a VM (QEMU+Xen+QXL) and on a real hardware. I get the same errors. I suspect that I am missing something related to SDL. But don't worry about this. It's probably unrelated to retroarch. I'll open a separate issue once I have gathered enough information.)

kwyxz:
```
Do you get the resolve symbol "fseeko64" error if you also use the packages from Haiku's build servers?
```

No, I do get the semaphore messages, but I did not get this fseeko64 error.

fbrosson:
That's puzzling. Thanks for the info :)
Anyway, sooner or later we will hear from other users ;)

PR #2367, opened 2018-03-22
Guile: new recipe

extrowerk:


dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
+SUMMARY="GNU Ubiquitous Intelligent Language for Extensions"
+DESCRIPTION="Guile is a library designed to help programmers create flexible \
+applications. Using Guile in an application allows programmers to write \
+plug-ins, or modules (there are many names, but the concept is essentially \
+the same) and users to use them to have an application fit their needs."
+HOMEPAGE="https://www.gnu.org/software/guile/"
+COPYRIGHT="1993-2018 Aubrey Jaffer, George Carrette, et al."
+LICENSE="GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="https://ftp.gnu.org/gnu/guile/guile-2.2.3.tar.gz"
```
fbrosson: Suggestion:
```diff
-SOURCE_URI="https://ftp.gnu.org/gnu/guile/guile-2.2.3.tar.gz"
+SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.gz
+	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.gz"
```
(Notice that the "/gnu/" is only required for the second URI. It is not needed for the first one.)

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libguileVersion="1.3.0"
+libguileVersionCompat="$libguileVersion compat >= ${libguileVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
+
```
fbrosson: Suggestion:
```diff
 portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
+portVers="${portVersion%.*}"
```
Well, we don't really need this, but it can be handy since we can later replace each occurence of `2.2` by `$portVers`.

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+PROVIDES="
+	guile$secondaryArchSuffix = $portVersionCompat
+	cmd:guild$commandSuffix = $portVersion
+	cmd:guile$commandSuffix = $portVersion
+	cmd:guile_config$commandSuffix = $portVersion
+	cmd:guile_snarf$commandSuffix = $portVersion
+	cmd:guile_tools$commandSuffix = $portVersion
+	lib:libguile_2.2$secondaryArchSuffix = $libguileVersionCompat
```
fbrosson:
```diff
-	lib:libguile_2.2$secondaryArchSuffix = $libguileVersionCompat
+	lib:libguile_$portVers$secondaryArchSuffix = $libguileVersionCompat
```

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+	lib:libintl$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libreadline$secondaryArchSuffix
+	lib:libunistring$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	guile${secondaryArchSuffix}_devel = $portVersionCompat
+	devel:libguile_2.2$secondaryArchSuffix = $libguileVersionCompat
```
fbrosson:
```diff
-	devel:libguile_2.2$secondaryArchSuffix = $libguileVersionCompat
+	devel:libguile_$portVers$secondaryArchSuffix = $libguileVersionCompat
```

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+	cmd:gperf
+	cmd:grep
+	cmd:make
+	cmd:ld
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	"
+
+defineDebugInfoPackage guile$secondaryArchSuffix \
+	"$libDir"/libguile-2.2.so.$libguileVersion
```
fbrosson:
```diff
-	"$libDir"/libguile-2.2.so.$libguileVersion
+	"$libDir"/libguile-$portVers.so.$libguileVersion
```

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+
+BUILD()
+{
+	CFLAGS="-D_BSD_SOURCE" LDFLAGS="-lbsd -lnetwork" \
+		runConfigure --omit-dirs binDir ./configure \
+			--with-threads
```
fbrosson:
```diff
 	CFLAGS="-D_BSD_SOURCE" LDFLAGS="-lbsd -lnetwork" \
-		runConfigure --omit-dirs binDir ./configure \
+	runConfigure --omit-dirs binDir ./configure \
-			--with-threads
+		--with-threads
```
(I'd say that removing one tab before the `runConfigure` (and the next line too) makes the bloc look nicer because the first line in `BUILD()` is just setting environment variables for `runConfigure`.)

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f "$libDir"/lib*.la
+	prepareInstalledDevelLibs libguile-2.2
```
fbrosson:
```diff
-	prepareInstalledDevelLibs libguile-2.2
+	prepareInstalledDevelLibs libguile-$portVers
```

dev-scheme/guile/guile-2.2.3.recipe
alaviss: Just an addition to @fbrosson suggestion

```diff
-	CFLAGS="-D_BSD_SOURCE" LDFLAGS="-lbsd -lnetwork" \
+	CPPFLAGS="-D_BSD_SOURCE" LIBS="-lbsd -lnetwork" \

File: dev-scheme/guile/guile-2.2.3.recipe [hunk 8]
```diff
@@ -0,0 +1,111 @@
...
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f "$libDir"/lib*.la
+	prepareInstalledDevelLibs libguile-$portVers
```
fbrosson: The recipe is almost perfect and Guile builds fine on all architectures :)
I just found a few minor things: We have a static lib (and its libtool file) in the base package:
```
package list -p guile-2.2.3-1-x86_64.hpkg  | grep -E '\.l?a$'
lib/guile/2.2/extensions/guile-readline.a
lib/guile/2.2/extensions/guile-readline.la
```
So I'm trying this change:
```diff
 	rm -f "$libDir"/lib*.la
+	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.la
-	prepareInstalledDevelLibs libguile-$portVers
+	prepareInstalledDevelLibs \
+		libguile-$portVers \
+		guile/$portVers/extensions/guile-readline
```
I don't know if it's OK to pass (`$libDir`-relative) paths to `prepareInstalledDevelLibs`.
That's what I'm going to test right now.

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,111 @@
...
+	fixPkgconfig
+	packageEntries devel $developDir
```
fbrosson:
```
package list -p guile-2.2.3-1-x86_64.hpkg | grep aclocal
data/aclocal
data/aclocal/guile.m4
```
I'm not sure about this, but maybe we should move `/system/data/aclocal/guile.m4` to the devel package.
```diff
-	packageEntries devel $developDir
+	packageEntries devel \
+		$dataDir/aclocal
+		$developDir
```
@alaviss, what do you think?

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,118 @@
...
+	lib:libintl$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libreadline$secondaryArchSuffix
+	lib:libunistring$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	guile${secondaryArchSuffix}_devel = $portVersionCompat
+	cmd:guile_config$commandSuffix = $portVersion
```
fbrosson: I wonder if it would not be better to install `guile-config` in `$binDir`.
```diff
-	cmd:guile_config$commandSuffix = $portVersion
+	cmd:guile_config$secondaryArchSuffix = $portVersion
```

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,118 @@
...
+
+	rm -f "$libDir"/lib*.la
+	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.la
+
+	prepareInstalledDevelLib libguile-$portVers
+
+	fixPkgconfig
+
+	packageEntries devel \
+		"$commandBinDir"/guile-config \
```
fbrosson: I wonder if it would not be better to install `guile-config` in `$binDir`.
```diff
+	if [ -n "$secondaryArchSuffix" ]; then
+		mkdir -p "$binDir"
+		mv "$binDir"/guile-config "$binDir"
+	fi
+
 	packageEntries devel \
-		"$commandBinDir"/guile-config \
+		"$binDir"/guile-config \
 		"$dataDir"/aclocal \
 		"$developDir"
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rmdir "$binDir"
+	fi
```

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,118 @@
...
+	# No MADV_DONTNEED, yet
+	sed --in-place '/HAVE_SYS_MMAN_H/d' config.h
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
```
fbrosson:
```diff
	make install
+	rm -f "$libDir"/charset.alias
```

dev-scheme/guile/guile-2.2.3.recipe
fbrosson: Sorry, there was a small error in my previous comment. It's fixed now.
(`s/binDir/commandBinDir/` in the line with the `mv`)

dev-scheme/guile/guile-2.2.3.recipe
```diff
@@ -0,0 +1,118 @@
...
+	rm -f "$libDir"/lib*.la
+	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.la
```
fbrosson:
```
package list -p guile_x86-2.2.3-1-x86_gcc2.hpkg | grep /extensions/
lib/x86/guile/2.2/extensions/guile-readline.a
lib/x86/guile/2.2/extensions/guile-readline.so
lib/x86/guile/2.2/extensions/guile-readline.so.0
lib/x86/guile/2.2/extensions/guile-readline.so.0.0.0
```
We can't move `guile-readline.a` to the devel package with `prepareInstalledDevelLib` because it would not move it to `$developLibDir/guile/2.2/extensions/` but to `$developLibDir/`.
(It seems it's only appropriate to call `prepareInstalledDevelLib` for libs that are in `$libDir`, not for libs which are in a subdirectory.)
So, instead of dropping just the libtool file, shall we also drop the static lib?
```diff
-	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.la
+	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.*a
```
(This is not the main lib, `libguile-2.2.a` which `prepareInstalledDevelLib` already moves to the devel package.)

alaviss: I believe using `find` would be much simpler in this case. It would also handle the case where more extensions are added.

```diff
-	rm -f "$libDir"/lib*.la
-	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.la
+	find -name '*.la' -type f -delete

File: dev-scheme/guile/guile-2.2.3.recipe (hunk 11 above)
Comment (alaviss) [reply dev-scheme/guile/guile-2.2.3.recipe]: @fbrosson Or maybe you could just remove `--omit-dirs binDir` up there

File: dev-scheme/guile/guile-2.2.3.recipe (hunk 13 above)
Comment (fbrosson) [reply dev-scheme/guile/guile-2.2.3.recipe]: Hmm, there are only 2 libtool files, so I think it is nicer without the find command.

BTW, I'm suggesting one `rm` to delete the libtool file of the main lib (`libguile-2.2.la`), and another `rm` to delete both the libtool file of the extension lib (`guile-readline.la`) as well as the static lib itself, `guile-readline.a`:
```bash
	rm -f "$libDir"/lib*.la
	rm -f "$libDir"/guile/$portVers/extensions/guile-readline.*a
```

alaviss: It's nice until they add more, but I guess that could be for later when they actually do so.

About the static lib, I don't know how guile work, but maybe the static lib is necessary some how?

dev-scheme/guile/guile-2.2.3.recipe
alaviss: Oh, sorry, I didn't notice that.

dev-scheme/guile/guile-2.2.3.recipe
fbrosson: The extension available in our build of guile is currently present as a static lib, `guile-readline.a` as well as a shared lib, `guile-readline.so.0.0.0`.
You are probably right that guile might need that extension's static lib when guile is told to compile some program. But since it also comes with the shared lib, I believe guile can be told to create a compiled runtime that dynamically links to the shared lib.

I therefore see 3 scenari:
1) we keep that static lib in the base package. That's OK for me, although I think it would be better in the devel package.
2) we move the static lib to the devel package, but in that case we would have to create the symbolic links just like `prepareInstalledDevelLib` does when it is given the (base)name of a shared lib in `$libDir`. We could try this, but we won't be sure it is OK until we have used guile to compile something that needs the `guile-readline.a` static lib. I'm OK with this option, but it will require some investigation.
3) we choose to drop the static lib. I think this is the best scenario because it is easy to "implement" and is also easy to revert, in case we find out that it's not possible to compile a program against `guile-readline.so.0` if the corresponding static lib is missing. But this would be weird.

So I belive the best scenari are 3 and 1. The PR is currently doing scenario 1. Shall we try scenario 3?

To address your comment about future extensions, shall we use the following commands?
```bash
	rm -f "$libDir"/lib*.la
	rm -f "$libDir"/guile/$portVers/extensions/*.la
	rm -f "$libDir"/guile/$portVers/extensions/*.a
```
(Of course, the last command should be dropped if we stay in scenario 1.)

PR #2370, opened 2018-03-23
Add libxft-2.3.2 recipe

nikolaszimmermann:
I need Xft as a dependency for a complex package that I'm trying to port. I'm aware that there's no X11 server on Haiku, but nonetheless Xft seems to compile.

### DIFF
--- a/x11-libs/libxft/libxft-2.3.2.recipe
+++ b/x11-libs/libxft/libxft-2.3.2.recipe
@@ -1,0 +1,99 @@
+SUMMARY="FreeType-based font drawing library for X"
+DESCRIPTION="Xft provides a client-side font API for X applications, making \
+the FreeType font rasterizer available to X clients. Fontconfig is used for \
+font specification resolution. Where available, the RENDER extension handles \
+glyph drawing; otherwise, the core X protocol is used."
+HOMEPAGE="https://www.x.org/"
+COPYRIGHT="2000-2003 Keith Packard"
+LICENSE="libXft"
+REVISION="1"
+SOURCE_URI="https://www.x.org/releases/individual/lib/libXft-$portVersion.tar.bz2"
+CHECKSUM_SHA256="f5a3c824761df351ca91827ac221090943ef28b248573486050de89f4bfcdc4c"
+SOURCE_DIR="libXft-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	libxft$secondaryArchSuffix = $portVersion
+	lib:libxft$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libfontconfig$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libx11$secondaryArchSuffix
+	lib:libxrender$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libxft${secondaryArchSuffix}_devel = $portVersion
+	devel:libxft$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	libxft$secondaryArchSuffix == $portVersion base
+	devel:libx11$secondaryArchSuffix
+	devel:libxau$secondaryArchSuffix
+	devel:libxcb$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:kbproto$secondaryArchSuffix
+	devel:libfontconfig$secondaryArchSuffix
+	devel:libpthread_stubs$secondaryArchSuffix
+	devel:libx11$secondaryArchSuffix
+	devel:libxrender$secondaryArchSuffix
+	devel:renderproto$secondaryArchSuffix
+	devel:util_macros$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libxft$secondaryArchSuffix \
+	"$libDir"/libXft.so.$libVersion
+
+BUILD()
+{
+	autoreconf -i
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	# remove libtool library file
+	rm -f "$libDir"/libXft.la
+
+	prepareInstalledDevelLib libXft
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf "$documentationDir"
+		maybe_manDir=
+	else
+		maybe_manDir="$manDir"
+	fi
+
+	packageEntries devel \
+		"$developDir" \
+		${maybe_manDir:+"$maybe_manDir"}
+}
+
+TEST()
+{
+	make check
+}
--- a/x11-libs/libxft/licenses/libXft
+++ b/x11-libs/libxft/licenses/libXft
@@ -1,0 +1,21 @@
+
+Copyright © 2001,2003 Keith Packard
+
+Permission to use, copy, modify, distribute, and sell this software and its
+documentation for any purpose is hereby granted without fee, provided that
+the above copyright notice appear in all copies and that both that
+copyright notice and this permission notice appear in supporting
+documentation, and that the name of Keith Packard not be used in
+advertising or publicity pertaining to distribution of the software without
+specific, written prior permission.  Keith Packard makes no
+representations about the suitability of this software for any purpose.  It
+is provided "as is" without express or implied warranty.
+
+KEITH PACKARD DISCLAIMS ALL WARRANTIES WITH REGARD TO THIS SOFTWARE,
+INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS, IN NO
+EVENT SHALL KEITH PACKARD BE LIABLE FOR ANY SPECIAL, INDIRECT OR
+CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM LOSS OF USE,
+DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR OTHER
+TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+PERFORMANCE OF THIS SOFTWARE.
+


x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+SOURCE_URI="https://www.x.org/releases/individual/lib/libXft-$portVersion.tar.gz"
+CHECKSUM_SHA256="26cdddcc70b187833cbe9dc54df1864ba4c03a7175b2ca9276de9f05dce74507"
+SOURCE_DIR="libXft-$portVersion"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+PROVIDES="
+	libxft$secondaryArchSuffix = $portVersion
+	lib:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
```
alaviss: For easier maintainance, we recommend the use of [`libVersionCompat`][0], which would help reduce duplicates, and allow for smaller diff per version bump.

```diff
 
+libVersion="4.11.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libxft$secondaryArchSuffix = $portVersion
-	lib:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
+	lib:libxft$secondaryArchSuffix = $libVersionCompat
```

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-libversioncompat-user-variable

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+	haiku$secondaryArchSuffix
+	lib:libfontconfig$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libxrender$secondaryArchSuffix
+	lib:libx11$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libxft${secondaryArchSuffix}_devel = $portVersion
+	devel:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
```
alaviss:
```diff
-	devel:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
+	devel:libxft$secondaryArchSuffix = $libVersionCompat

File: x11-libs/libxft/libxft-2.3.2.recipe [hunk 3]
```diff
@@ -0,0 +1,99 @@
...
+SOURCE_DIR="libXft-$portVersion"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+PROVIDES="
+	libxft$secondaryArchSuffix = $portVersion
+	lib:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
+	"
+
```
alaviss: The newline here should be removed, as required by the [Guidelines][0]

```diff
-
```

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libfontconfig$secondaryArchSuffix
+	devel:libx11$secondaryArchSuffix
+	devel:util_macros$secondaryArchSuffix
+	devel:kbproto$secondaryArchSuffix
+	devel:libpthread_stubs$secondaryArchSuffix
+	devel:libxrender$secondaryArchSuffix
+	devel:renderproto$secondaryArchSuffix
+	"
+
```
alaviss: The newline here should be removed, as required by the [Guidelines][0]

```diff
-
```

[0]: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+	libxft$secondaryArchSuffix = $portVersion
+	lib:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
+	"
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libfontconfig$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libxrender$secondaryArchSuffix
+	lib:libx11$secondaryArchSuffix
```
alaviss: Alphabetical sorting is very welcome here

```diff
+	lib:libx11$secondaryArchSuffix
 	lib:libxrender$secondaryArchSuffix
-	lib:libx11$secondaryArchSuffix

File: x11-libs/libxft/libxft-2.3.2.recipe [hunk 6]
```diff
@@ -0,0 +1,99 @@
...
+
+PROVIDES_devel="
+	libxft${secondaryArchSuffix}_devel = $portVersion
+	devel:libxft$secondaryArchSuffix = 4.11.0 compat >= 4
+	"
+REQUIRES_devel="
+	libxft$secondaryArchSuffix == $portVersion base
+	devel:libx11$secondaryArchSuffix
+	devel:libxcb$secondaryArchSuffix
+	devel:libxau$secondaryArchSuffix
```
alaviss:
```diff
+	devel:libxau$secondaryArchSuffix
 	devel:libxcb$secondaryArchSuffix
-	devel:libxau$secondaryArchSuffix

File: x11-libs/libxft/libxft-2.3.2.recipe [hunk 7]
```diff
@@ -0,0 +1,99 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libfontconfig$secondaryArchSuffix
+	devel:libx11$secondaryArchSuffix
+	devel:util_macros$secondaryArchSuffix
+	devel:kbproto$secondaryArchSuffix
+	devel:libpthread_stubs$secondaryArchSuffix
+	devel:libxrender$secondaryArchSuffix
+	devel:renderproto$secondaryArchSuffix
```
alaviss:
```diff
-	devel:libfontconfig$secondaryArchSuffix
-	devel:libx11$secondaryArchSuffix
-	devel:util_macros$secondaryArchSuffix
 	devel:kbproto$secondaryArchSuffix
+	devel:libfontconfig$secondaryArchSuffix
 	devel:libpthread_stubs$secondaryArchSuffix
+	devel:libx11$secondaryArchSuffix
 	devel:libxrender$secondaryArchSuffix
 	devel:renderproto$secondaryArchSuffix
+	devel:util_macros$secondaryArchSuffix

File: x11-libs/libxft/libxft-2.3.2.recipe [hunk 8]
```diff
@@ -0,0 +1,99 @@
...
+
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:automake
+	cmd:autoconf
+	cmd:autoreconf
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
```
alaviss:
```diff
+	cmd:autoconf
+	cmd:automake
+	cmd:autoreconf
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
 	cmd:libtoolize$secondaryArchSuffix
-	cmd:automake
-	cmd:autoconf
-	cmd:autoreconf
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix

File: x11-libs/libxft/libxft-2.3.2.recipe [hunk 9]
```diff
@@ -0,0 +1,99 @@
...
+	"
+
+BUILD()
+{
+	libtoolize --force --copy --install
```
alaviss: `autoreconf` will automatically call `libtoolize` as needed, so an explicit call should not be necessary

```diff
-	libtoolize --force --copy --install
```

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+The development work is being done in conjunction with the freedesktop.org \
+community. \
+The X.Org Foundation is the educational non-profit corporation whose \
+Board serves this effort, and whose Members lead this work."
+HOMEPAGE="https://www.x.org/"
+COPYRIGHT="1989-95 GROUPE BULL
+	1998 Arnaud LE HORS
+	1996 Lorens Younes"
+# TODO: check all the licenses in the COPYING file in the next century.
+LICENSE="MIT"
```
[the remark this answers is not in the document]
fbrosson: Since it is a custom license, I'd call it libXft.
We have done this for a few other recipes:
* x11-libs/libxcomposite/licenses/libXcomposite
* x11-libs/libxcursor/licenses/libXcursor
* x11-libs/libxdamage/licenses/libXdamage
* x11-libs/libxfixes/licenses/libXfixes
* x11-libs/libxrender/licenses/libXrender

(Each of these files is a copy of the `COPYING` file from the tarball.)

nikolaszimmermann:
Thanks for the comments! I will address them soon.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	# remove libtool library file
+	rm $libDir/libXft.la
+
+	prepareInstalledDevelLibs \
```
fbrosson: Suggestion:
```diff
-	prepareInstalledDevelLibs \
-		libXft
+	prepareInstalledDevelLib libXft
-
 	fixPkgconfig
```
(Dropping the "s" in prepareInstalledDevelLib**s** is not required, but since we know that this recipe has only one lib and is likely to never have more than one lib, it would be slightly better to call `prepareInstalledDevelLib` instead of `prepareInstalledDevelLibs`.)

BTW, I have a few recipes which I have started some time ago but I need to do some polishing before I push them:
* x11-libs/libxscrnsaver
* x11-proto/printerproto
* x11-libs/libxp

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
+SUMMARY="X.Org Xft library"
```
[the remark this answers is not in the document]
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+DESCRIPTION="The X.Org project provides an open source implementation of the \
+X Window System. \
+The development work is being done in conjunction with the freedesktop.org \
+community. \
+The X.Org Foundation is the educational non-profit corporation whose \
+Board serves this effort, and whose Members lead this work."
```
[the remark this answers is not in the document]
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+HOMEPAGE="https://www.x.org/"
+COPYRIGHT="1989-95 GROUPE BULL
+	1998 Arnaud LE HORS
+	1996 Lorens Younes"
```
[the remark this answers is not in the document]
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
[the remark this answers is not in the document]
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+# TODO: check all the licenses in the COPYING file in the next century.
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://www.x.org/releases/individual/lib/libXft-$portVersion.tar.gz"
+CHECKSUM_SHA256="26cdddcc70b187833cbe9dc54df1864ba4c03a7175b2ca9276de9f05dce74507"
```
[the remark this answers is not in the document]
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed - I removed more places where newlines were present that violate the guidelines.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Okay, I've removed this call. (Copy & paste left-over from another X11 recipe).

nikolaszimmermann:
I will try to incorporate the feedback in following PRs, thanks for the detailed comments - very helpful.
To ease the reviewing process, I will fix all comments in a separated git commit and squash them later once the review is finished.

If you prefer a different workflow, let me know.

nikolaszimmermann:
alaviss, you asked for debuginfo declarations - I'm not fully sure I understand the need for them in release builds. Please enlighten me.

fbrosson:
The workflow you plan to use is perfect.
That said, if you would like to use `git --amend`, that would also be OK, so please do not hesitate to use any workflow you wish.

fbrosson:
Regarding the debuginfo package, we've been adding them mainly for 2 reasons:
1) It is better than simply stripping the runtimes,
2) users can install the debuginfo packages at any time, or not install them at all.

When I add a debuginfo package in a recipe I look at the size of the files in the debuginfo package (with e.g. `package list libxft_debuginfo-2.3.2-1_x86_64.hpkg`) to see if the files are not empty skeletons. If they are big enough, i.e. at least a few kilobytes, then I know it (probably) makes sense to create the debuginfo package.
@alaviss, would you say that packages compiled without `-g` produce unusable debuginfo packages?
We had a discussion on that topic sometime ago but I think we did not really decide what was the best practice.

alaviss:
@fbrosson Yes, when packages are not compiled with `-g`, no debuginfo is generated. However, since we don't enforce any `CFLAGS`, the package build system is free to choose their own, which sometimes includes `-g`

The best way to check whether a package is compiled with debuginfo is to look at it's gcc invocation. For `autohell`-based packages, run `make V=1` and you will be able to see how the build system invoke gcc. If `-g` is found then it's safe to define debuginfo package.

For `cmake`-based, `make VERBOSE=1` should do the trick.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+extension handles glyph drawing; otherwise, the core X protocol is used."
+HOMEPAGE="https://www.x.org/"
+COPYRIGHT="2000-2003 Keith Packard"
+LICENSE="libXft"
+REVISION="1"
+SOURCE_URI="https://www.x.org/releases/individual/lib/libXft-$portVersion.tar.bz2"
+CHECKSUM_SHA256="f5a3c824761df351ca91827ac221090943ef28b248573486050de89f4bfcdc4c"
+SOURCE_DIR="libXft-$portVersion"
+
+ARCHITECTURES="?x86_gcc2 x86 x86_64"
```
fbrosson: You can mark x86_gcc2 as tested, as the recipe builds fine on that arch:
```diff
-ARCHITECTURES="?x86_gcc2 x86 x86_64"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```

BTW, some contributors like to get rid of the `x86` in `ARCHITECTURES` and of the `x86_gcc2` in `SECONDARY_ARCHITECTURES` but I prefer to keep them if I notice that the recipe builds on these architectures.
FWIW:
I only drop a `x86` (or `?x86`) in `ARCHITECTURES` when the recipe is broken on that architecture.
I only drop a `x86_gcc2` (or `?x86_gcc2`) in ``SECONDARY_ARCHITECTURES`` is when the recipe is broken on that architecture.
And when I find that a recipe builds but I can't test it then I just mark it as untested.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+libVersion="4.11.0"
```
fbrosson:
```diff
-libVersion="4.11.0"
+libVersion="$portVersion"
```
That's not very common, but the version of the shared lib is the same as that of the package.

alaviss:
@nikolaszimmermann For most packages, "release builds" is just disabling assertion/debugging code, which is great, since you don't want applications to start crashing over a small bump in release build. 

Debuginfo, however, is different. It's main purpose is to let you have pretty and meaningful stacktrace when crash happens. If you happens to carry around the exact source, you could also get the exact line that trigger the crash, what variables where present, etc. It is generated by gcc when `-g` flags is used, which is common when no `CFLAGS` is defined. However, debuginfo tends to be quite heavy, so it's stripped and placed into an another package, which users could install at anytime.

I've explained above on how to check if a package is built with debug info. For more information about the subject, you should take a look at tħe Gentoo article on [stacktraces](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Backtraces)

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+	make install
+
+	# remove libtool library file
+	rm $libDir/libXft.la
+
+	prepareInstalledDevelLib libXft
+	fixPkgconfig
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		rm -rf $binDir $documentationDir
```
fbrosson:
```diff
 	if [ -n "$secondaryArchSuffix" ]; then
-		rm -rf $binDir $documentationDir
+		rm -rf "$documentationDir"
+		maybe_manDir=
+	else
+		maybe_manDir="$manDir"
 	fi
```
(We can only put `$manDir/man3` in the devel package if we still have `$documentationDir`.)

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+	fi
+
+	packageEntries devel \
+		$developDir
```
fbrosson:
```diff
	packageEntries devel \
-		$developDir
+		"$developDir" \
+		${maybe_manDir:+"$maybe_manDir"}
```

If you prefer to use this simpler variant:
```
	packageEntries devel \
		$developDir \
		$maybe_manDir
```
that's OK.

The reason why prefer the first suggestion is to have a more flexible recipe which would still work even if a user chooses to:
a) have spaces or tabs in the path to the haikuports tree (that's the `TREE_PATH` parameter in `~/config/settings/thaikuports.conf`), and:
b) pass `--nochroot` to haikuporter when building this recipe. (This option is currently broken, but I'd not assume that it will always be broken.)

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,92 @@
...
+	cmd:automake
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
```
fbrosson:
```diff
 	cmd:pkg_config$secondaryArchSuffix
 	"
+
+defineDebugInfoPackage libxft$secondaryArchSuffix \
+	"$libDir"/libXft.so.$libVersion
+
 BUILD()
```

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Quoting makes sense, I've adapt the recipe. Thanks.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

nikolaszimmermann:
I think I addressed all comments, will push a squashed version now, ready to merge, if there are no further comments.

Thanks for the detailed comments, that makes it easy to get started for Haiku newcomers.

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
+SUMMARY="FreeType-based font drawing library for X"
+DESCRIPTION="Xft provides a client-side font API for X applications, \
+making the FreeType font rasterizer available to X clients. Fontconfig \
+is used for font specification resolution. Where available, the RENDER \
+extension handles glyph drawing; otherwise, the core X protocol is used."
```
fbrosson:
```diff
-DESCRIPTION="Xft provides a client-side font API for X applications, \
+DESCRIPTION="Xft provides a client-side font API for X applications, making \
-making the FreeType font rasterizer available to X clients. Fontconfig \
+the FreeType font rasterizer available to X clients. Fontconfig is used for \
-is used for font specification resolution. Where available, the RENDER \
+font specification resolution. Where available, the RENDER extension handles \
-extension handles glyph drawing; otherwise, the core X protocol is used."
+glyph drawing; otherwise, the core X protocol is used."
```
This is not strictly required, but it would be nice to fill the lines as much as possible (i.e. to have lines <= 80 chars, including the `\`). Thanks!

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+	devel:renderproto$secondaryArchSuffix
+	devel:util_macros$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:automake
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
```
fbrosson:
```diff
- 	cmd:libtoolize$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
+ 	cmd:libtoolize$secondaryArchSuffix
```
(sorting :)

x11-libs/libxft/libxft-2.3.2.recipe
```diff
@@ -0,0 +1,99 @@
...
+	runConfigure ./configure
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	# remove libtool library file
+	rm $libDir/libXft.la
```
fbrosson: Suggestion:
```diff
-	rm $libDir/libXft.la
+	rm -f "$libDir"/libXft.la
```

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed.

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Fixed, sorry, missed it before :-)

x11-libs/libxft/libxft-2.3.2.recipe
nikolaszimmermann: Good catch.

nikolaszimmermann:
Thanks @fbrosson!

PR #2376, opened 2018-03-24
jbigkit, bump version, remove old bep files

Begasus:


media-libs/jbigkit/patches/jbigkit-2.1.patchset
```diff
@@ -0,0 +1,78 @@
...
+
+
+diff --git a/libjbig/Makefile_so b/libjbig/Makefile_so
+new file mode 100644
+index 0000000..a48ddb4
+--- /dev/null
++++ b/libjbig/Makefile_so
+@@ -0,0 +1,40 @@
++# This Makefile builds a shared version of the library, 
++# libjbig.so.2.0, with soname libjbig.so.2,
```
fbrosson: `2.0` → `2.1` :smiley:

Begasus: Bugger, missed that one ;)

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,87 @@
+SUMMARY="A lossless image compression library"
+DESCRIPTION="JBIG-KIT is a software implementation of the JBIG1 data \
+compression standard (ITU-T T.82), which was designed for bi-level image \
+data, such as scanned documents. This library is available in portable C \
+code. It is widely used in fax products, printer firmware, printer drivers, \
+document management systems and imaging software."
+HOMEPAGE="http://www.cl.cam.ac.uk/~mgk25/jbigkit/"
```
alaviss:
```diff
-HOMEPAGE="http://www.cl.cam.ac.uk/~mgk25/jbigkit/"
+HOMEPAGE="https://www.cl.cam.ac.uk/~mgk25/jbigkit/"

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+COPYRIGHT="1995-2014 Markus Kuhn"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://www.cl.cam.ac.uk/~mgk25/download/jbigkit-$portVersion.tar.gz"
```
alaviss:
```diff
-SOURCE_URI="http://www.cl.cam.ac.uk/~mgk25/download/jbigkit-$portVersion.tar.gz"
+SOURCE_URI="https://www.cl.cam.ac.uk/~mgk25/download/jbigkit-$portVersion.tar.gz"

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 4]
```diff
@@ -0,0 +1,87 @@
+SUMMARY="A lossless image compression library"
+DESCRIPTION="JBIG-KIT is a software implementation of the JBIG1 data \
+compression standard (ITU-T T.82), which was designed for bi-level image \
+data, such as scanned documents. This library is available in portable C \
+code. It is widely used in fax products, printer firmware, printer drivers, \
+document management systems and imaging software."
+HOMEPAGE="http://www.cl.cam.ac.uk/~mgk25/jbigkit/"
+COPYRIGHT="1995-2014 Markus Kuhn"
```
alaviss:
```diff
-COPYRIGHT="1995-2014 Markus Kuhn"
+COPYRIGHT="1995-2014 Markus Kuhn
+	1989, 1991 Jef Poskanzer"

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 5]
```diff
@@ -0,0 +1,87 @@
...
+LICENSE="GNU GPL v2"
```
alaviss: There is also this little license embedded at the end of `pbm.5`

```nroff
Copyright (C) 1989, 1991 by Jef Poskanzer.
.\" Permission to use, copy, modify, and distribute this software and its
.\" documentation for any purpose and without fee is hereby granted, provided
.\" that the above copyright notice appear in all copies and that both that
.\" copyright notice and this permission notice appear in supporting
.\" documentation.  This software is provided "as is" without express or
.\" implied warranty.

**Comment by Begasus** [comment]
Anything I could do here to finish this up?

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 6]
```diff
@@ -0,0 +1,88 @@
...
+INSTALL()
+{
+	mkdir -p $binDir $includeDir $libDir $developLibDir
+
+	# install the tools
+	cd pbmtools
+	cp jbgtopbm $binDir
+	cp jbgtopbm85 $binDir
+	cp pbmtojbg $binDir
+	cp pbmtojbg85 $binDir
```
alaviss: Shall we use this shorter variant?

```diff
-	cp jbgtopbm $binDir
-	cp jbgtopbm85 $binDir
-	cp pbmtojbg $binDir
-	cp pbmtojbg85 $binDir
+	install -t "$binDir" -m 755 jbgtopbm{,85} pbmtojbg{,85}

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 7]
```diff
@@ -0,0 +1,88 @@
...
+
+	# install headers and libs
+	cd ../libjbig
+	cp *.h $includeDir
+	cp libjbig* $libDir
```
alaviss: Well, this is mainly for consistency :)

```diff
-	cp *.h $includeDir
-	cp libjbig* $libDir
+	install -t "$includeDir" -m 644 *.h
+	install -t "$libDir" -m 644 libjbig*

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 8]
```diff
@@ -0,0 +1,88 @@
...
+BUILD()
+{
+	make all $jobArgs
+	cd libjbig
+	make -f Makefile_so $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $binDir $includeDir $libDir $developLibDir
```
alaviss:
```diff
-	mkdir -p $binDir $includeDir $libDir $developLibDir
+	install -d -m 755 "$binDir" "$includeDir" "$libDir" "$developLibDir"

File: media-libs/jbigkit/jbigkit-2.1.recipe [hunk 9]
```diff
@@ -0,0 +1,88 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:make
+	"
+defineDebugInfoPackage jbigkit \
+	$libDir/libjbig.so \
+	$libDir/libjbig85.so
```
alaviss: Shouldn't this be done on the main file instead of the symlink? And aren't both `libjbig` and `libjbig85` the same file?

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: Thanks for the pointers, haven't used these kind of versions of "install" yet, will look into it :)

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: Have to check this one ...

fbrosson: I think both work, but it's better, imo, to specify the file instead of the symlink.

```diff
 defineDebugInfoPackage jbigkit \
-	$libDir/libjbig.so \
+	"$libDir"/libjbig.so.$libVersion \
-	$libDir/libjbig85.so
+	"$libDir"/libjbig85.so.$libVersion
```
The main advantage is that if we forget to update the `libVersion` then haikuporter will complain, so we have a chance to notice it before merging.

BTW, if I'm not mistaken, we should add support for secondary architectures because this recipe provides a lib. But maybe there are exceptions.

media-libs/jbigkit/jbigkit-2.1.recipe
fbrosson: What do you think of this (neutral) variant?
```diff
-	install -t "$includeDir" -m 644 *.h
+	install -m 0644 -t "$includeDir" *.h
-	install -t "$libDir" -m 644 libjbig*
+	install -m 0644 -t "$libDir" libjbig*
```

media-libs/jbigkit/jbigkit-2.1.recipe
alaviss: Well, we should add it anyway as lots of progs out there won't compile with gcc2.

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,88 @@
...
+HOMEPAGE="https://www.cl.cam.ac.uk/~mgk25/jbigkit/"
+COPYRIGHT="1995-2014 Markus Kuhn
+	1989, 1991 Jef Poskanzer"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://www.cl.cam.ac.uk/~mgk25/download/jbigkit-$portVersion.tar.gz"
+CHECKSUM_SHA256="de7106b6bfaf495d6865c7dd7ac6ca1381bd12e0d81405ea81e7f2167263d932"
+PATCHES="jbigkit-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 ?x86_64"
```
fbrosson: @Begasus, if you add support for secondary architectures to this recipe, could you please include `?x86` as primary arch and `?x86_gcc2` as secondary arch?

In other words, instead of making this change:
```diff
 ARCHITECTURES="x86_gcc2 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
```
could you please make this one:
```diff
-ARCHITECTURES="x86_gcc2 ?x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 ?x86"
```
or, if you notice that jbigkit builds fine with gcc5:
```diff
-ARCHITECTURES="x86_gcc2 ?x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
```
This does not cost much, and it is better, imo, than simply dropping x86 as primary arch and/or x86_gcc2 as secondary arch.

* Having `?x86` as primary arch means: _"We don't want to upset users of 32 bit x86 processors"_.
* Not having it means _"We don't care if you are not running x86_64."_
* Having `?x86_gcc2` as secondary arch means: _"If we create an hybrid `x86_64`+`x86_gcc2` then running BeOS apps using x86_gcc2 secondary arch on x86_64 will become a possibility, with some hacks."_
* Not having it means: _"Once we drop x86_gcc2 you will have to keep an old x86_gcc2 system if you wish to run any BeOS app on Haiku."_

I guess that adding x86_gcc2 secondary arch to x86_64 would (probably) not be an easy task, and I also know that this might even never happen. But since having `?x86` in `ARCHITECTURES` and `?x86_gcc2` in `SECONDARY_ARCHITECTURES` costs nothing, I think we should have them.

Thanks (and sorry for discussing this in your PR :)

Begasus: No problem for the discussion ;) it helps to clear out things, ps recipe isn't tested yet for x86 and x86_64 (but that will be in a bit) :)

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: This doesn't seem to work?

```
objcopy: '/packages/jbigkit_x86-2.1-1/.self/lib/x86/libjbig.so.2.1': No such file
```

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+	cmd:pbmtojbg85$secondaryArchSuffix
+	lib:libjbig$secondaryArchSuffix = $libVersionCompat
+	lib:libjbig85$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	jbigkit_devel$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES_devel="
-	jbigkit_devel$secondaryArchSuffix = $portVersion
+	jbigkit${secondaryArchSuffix}_devel = $portVersion
```

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+defineDebugInfoPackage jbigkit$secondaryArchSuffix \
```
fbrosson:
```diff
+
 defineDebugInfoPackage jbigkit$secondaryArchSuffix \
```
(inserting an empty line.)

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+	"$libDir"/libjbig.so \
+	"$libDir"/libjbig85.so
+
+BUILD()
+{
+	make all $jobArgs
+	cd libjbig
+	make -f Makefile_so $jobArgs
```
fbrosson: If you like using `make -C <directory>`, here's a neutral variant:
```diff
-	cd libjbig
-	make -f Makefile_so $jobArgs
+	make -C libjbig -f Makefile_so $jobArgs
```
(As said, this won't make any difference, so the only "advantage" of this variant is that you save a "cd" and a newline ;)

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+}
+
+INSTALL()
+{
+	install -d -m 755 $binDir $includeDir $libDir $developLibDir
```
fbrosson: Suggestion (neutral variant, of course):
```diff
-	install -d -m 755 $binDir $includeDir $libDir $developLibDir
+	install -m 0755 -d "$binDir" "$includeDir" "$libDir" "$developLibDir"
```

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+
+	# install the tools
+	cd pbmtools
+	install -t "$binDir" -m 755 jbgtopbm{,85} pbmtojbg{,85}
```
fbrosson: Suggestion (neutral variant, of course):
```diff
-	install -t "$binDir" -m 755 jbgtopbm{,85} pbmtojbg{,85}
+	install -m 0755 -t "$binDir" jbgtopbm{,85} pbmtojbg{,85}
```

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+	# install headers and libs
+	cd ../libjbig
+	install -m 0644 -t "$includeDir" *.h
+	install -m 0644 -t "$libDir" libjbig*
+
+	prepareInstalledDevelLibs libjbig libjbig85
+
+	packageEntries devel \
+		$developLibDir \
+		$includeDir
```
fbrosson: I guess this variant would to the same job:
```diff
 	packageEntries devel \
-		$developLibDir \
-		$includeDir
+		"$developDir"
```
(But I did not check yet.)

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+}
+
+TEST()
+{
+	cd libjbig
+	make test
```
fbrosson: If you like using `make -C <directory>`, here's a neutral variant:
```diff
-	cd libjbig
-	make test
+	make -C test libjbig
```
(As said, this won't make any difference, so the only "advantage" of this variant is that you save a "cd" and a newline ;)

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: Bumps head * ;)

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: Still a no go: 

```
objcopy: error: the input file '/packages/jbigkit-2.1-1/.self/lib/libjbig.so.2.1' is empty
```

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: This variant didn't work:

```
install: failed to access '-m': No such file or directory
```

Begasus:
Checked and tested on all arch's with success ...

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: My bad, didn't include the "" ...

fbrosson:
On Debian, the man5 pages for pbm and pgm are provided by the netpbm package.
But it looks like prebuilt hpkgs of netpbm are not available, so I guess it's OK to keep the man5 in jbigkit.

BTW, I found a few minor issues, so I'll add another review asap.
I'm writting anothe

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,88 @@
...
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+libVersion=$portVersion
+libVersionCompat="$libVersion compat >= 2"
+
+PROVIDES="
+	jbigkit$secondaryArchSuffix = $portVersion
+	cmd:jbgtopbm$secondaryArchSuffix = $portVersion
+	cmd:jbgtopbm85$secondaryArchSuffix = $portVersion
```
fbrosson:
```diff
 PROVIDES="
 	jbigkit$secondaryArchSuffix = $portVersion
-	cmd:jbgtopbm$secondaryArchSuffix = $portVersion
-	cmd:jbgtopbm85$secondaryArchSuffix = $portVersion
-	cmd:pbmtojbg$secondaryArchSuffix = $portVersion
-	cmd:pbmtojbg85$secondaryArchSuffix = $portVersion
 	lib:libjbig$secondaryArchSuffix = $libVersionCompat
 	lib:libjbig85$secondaryArchSuffix = $libVersionCompat
 	"
+if [ -z "$secondaryArchSuffix" ]; then
+PROVIDES="$PROVIDES
+	cmd:jbgtopbm = $portVersion
+	cmd:jbgtopbm85 = $portVersion
+	cmd:pbmtojbg = $portVersion
+	cmd:pbmtojbg85 = $portVersion
+	"
+fi
```

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,88 @@
...
+BUILD()
+{
+	make all $jobArgs
+	make -C libjbig -f Makefile_so $jobArgs
+}
+
+INSTALL()
+{
+	install -m 0755 -d "$binDir" "$includeDir" "$libDir" \
+		"$developLibDir" "$manDir/man5"
```
fbrosson:
```diff
 INSTALL()
 {
-	install -m 0755 -d "$binDir" "$includeDir" "$libDir" \
-		"$developLibDir" "$manDir/man5"
-
-	# install the tools
-	cd pbmtools
-	install -m 755 -t "$binDir" jbgtopbm{,85} pbmtojbg{,85}
-	# install man pages
-	install -m 0644 -t "$manDir/man5" pbm.5 pgm.5
+	if [ -z "$secondaryArchSuffix" ]; then
+		install -m 0755 -d "$binDir" "$manDir"/man5
 
+		# install the tools
+		pushd pbmtools
+		install -m 0755 -t "$binDir" jbgtopbm{,85} pbmtojbg{,85}
+
+		# install man pages
+		install -m 0644 -t "$manDir"/man5 pbm.5 pgm.5
+		popd
+	fi
 
 	# install headers and libs
-	cd ../libjbig
+	install -m 0755 -d "$libDir" "$developLibDir" "$includeDir"
+	cd libjbig
 	install -m 0644 -t "$includeDir" *.h
 	install -m 0644 -t "$libDir" libjbig*
```

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,88 @@
...
+
+	prepareInstalledDevelLibs libjbig libjbig85
+
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make -C libjbig test
```
fbrosson:
```diff
-	make -C libjbig test
+	make test
```
(There's a test target in the main Makefile, and it will call make test in libjbig and then in pbmtools.)
I did not try this, so maybe I should not be making this suggestion :smile:

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: Question :) What if I only install the secondaryArch package? the cmd's are not installed then?

fbrosson: Yeah, that's a good question.

There are 4 scenari for recipes which build on all primary architectures
1) We install everything, on both primary and secondary arch.
2) We install everything on primary architectures only, and on secondary arch we only install the libs. This is OK when speed is not important when running the `cmd:` runtimes.
3) On x86_gcc2 primary arch we only install the lib, and we rename the runtimes from the secondary arch to drop the x86. This is nice if speed is important, or if the runtimes compiled with gcc2 don't work as well as those compiled with gcc5. Otherwise, imo, this is not as nice as scenario 2.
4) Same as 3) but without the renaming.
5) A "mix" of 1) and 3) but moving the `cmd:` to a subpackage. That's what I did for zstd. (I did so because one of the runtimes did not build with gcc2.)

I'd say that if speed is not important then scenario 2 is the best choice because the recipe does not have to make any test about a specific arch. So the recipe is quite simple with this scenario. But I'm OK with any other scenario.

fbrosson: I have to admit that scenario 1 has a big advantage: simplicity. So I'm OK with it, unless other contributors think scenario 2 is more appropriate for this recipe.

media-libs/jbigkit/jbigkit-2.1.recipe
Begasus: Works, only needed to add one cmd for the test ...

media-libs/jbigkit/patches/jbigkit-2.1.patchset
```diff
@@ -0,0 +1,80 @@
...
++# CFLAGS = -O2 -W -Wno-unused-result
++CFLAGS = -O -g -W -ansi -pedantic # -DDEBUG
+ 
+ export CC CFLAGS
+ 
+-- 
+2.16.2
+
+
+From 6b76dac0aaa996bdb546ccc51f113ec8bcc719d6 Mon Sep 17 00:00:00 2001
```
fbrosson: Since you are the author of this patchset and since it is not yet in master, I think it would be nice to squash the three commits in it.

BTW, if you follow the suggestion I made in point 3) of [my previous message](https://github.com/haikuports/haikuports/pull/2376#pullrequestreview-110255091), then this will fix the issue spotted in point 1) (of the same message.)
Reminder: if you wish to use the `-g` in the `CFLAGS` then you should drop the `-c` in `CFLAGS=-fPIC -g -c -Wall`. (That's because you can't pass both `-c` and `-o` in the same call.)

Another option is to not use the `-g`, but in that case it would be better to not create the debuginfo package.

media-libs/jbigkit/patches/jbigkit-2.1.patchset
```diff
@@ -0,0 +1,80 @@
...
++SHELL=/bin/sh
++CC=gcc
++CFLAGS=-fPIC -g -c -Wall
++
++OBJS=  jbig.o  \
++      jbig_ar.o    \
++      jbig85.o
++
++all: $(OBJS)
++	$(CC) -shared -Wl,-soname,libjbig.so -o libjbig.so.2.1 $(OBJS)
```
fbrosson: Hmm, I forgot to mention something: if you declare a `compat >= 2` then I think you should use `-Wl,-soname,libjbig.so.2` instead of `-Wl,-soname,libjbig.so`.
And maybe you would not even need to create one of the 2 symlinks: `libjbig.so ->libjbig.so.2.1` in `$libDir`. Having only `libjbig.so.2 ->libjbig.so.2.1` will probably be OK if I'm not mistaken. In `$developLibDir`, however, I guess you need both symlinks.

cc: @alaviss, as he is an expert.

Begasus: Updated, I'll wait for the reviews :) thanks so far!

Begasus:
No idea there :/

Begasus:
jikes!

Begasus:
Should be ok, or do I still need to rebase? (did a checkout between 2 branches earlier instead from master to jbigkit :/ )

media-libs/jbigkit/jbigkit-2.1.recipe
```diff
@@ -0,0 +1,103 @@
...
+	if [ -z "$secondaryArchSuffix" ]; then
+		install -m 0755 -d "$binDir" "$manDir"/man5
+
+		# install the tools
+		pushd pbmtools
+		install -m 0755 -t "$binDir" jbgtopbm{,85} pbmtojbg{,85}
+
+		# install man pages
+		install -m 0644 -t "$manDir"/man5 pbm.5 pgm.5
+		popd
```
fbrosson: I had recently suggested to replace a `cd somedir` + `cd ..` by a `pushd somedir` but this was a very bad idea because `pushd` and `popd` are not available on all shells. (I won't suggest this anymore.)
So, shall we revert to `cd pbmtools` & `cd ..` ?
```diff
	if [ -z "$secondaryArchSuffix" ]; then
+		cd pbmtools
		install -m 0755 -d "$binDir" "$manDir"/man5
 		# install the tools
-		pushd pbmtools
 		install -m 0755 -t "$binDir" jbgtopbm{,85} pbmtojbg{,85}
 		# install man pages
 		install -m 0644 -t "$manDir"/man5 pbm.5 pgm.5
-		popd
+		cd ..
 	fi
```

Begasus:
builds ok on x86_gcc2 and _x86, tests are ok for the libraries, pbmtools tests fail because it can't find the libraries :/

Begasus:
The created links from the makefile were put as full libraries (no symlinks), so I linked them in the recipe 
This is what it looked like before: https://ibb.co/mbd3nU

PR #2397, opened 2018-03-27
mrboom_libretro: a port of MrBoom to the libretro API

kwyxz:
MrBoom is a Bomberman clone, originally created for MS-DOS, then entirely rewrote from ASM to C++ and ported to the Libretro architecture. This allowed porting it to a wide variety of different architectures and operating systems. Also @fbrosson it is one of these cores that do not require any additional files so you should be able to just load the core and start playing right away :-)


frranck:
Hi @kwyxz there's a new version with music; would mind updating it?

kwyxz:
Working on it. New version requires some patching upstream (I'd rather avoid patches in haikuport whenever possible).

PR #2405, opened 2018-03-31
Alembic 1.7.7

extrowerk:
Let's assume it wil lwork on x86 too.


media-libs/alembic/alembic-1.7.7.recipe
```diff
@@ -6,15 +6,29 @@ of scenes into baked geometry is exactly analogous to the distillation of \
...
+	2016-2018 Tim Yanalunas
+	2016-2018 Blizzard Entertainment, Inc.
+	2018 J Cube, Inc."
 LICENSE="ALEMBIC"
-REVISION="3"
+REVISION="1"
 SOURCE_URI="https://github.com/alembic/alembic/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256='ab0c727bfc4aedfe81c2365e604fb457056dc66909575ccfe19ed406e67c002e'
+CHECKSUM_SHA256='831b016afa9e66187b66819a02c0bd198b0f30e501a3552d0634796728cf1bff'
 PATCHES="alembic-$portVersion.patchset"
```
fbrosson:
```diff
 CHECKSUM_SHA256='831b016afa9e66187b66819a02c0bd198b0f30e501a3552d0634796728cf1bff'
+SOURCE_FILENAME="alembic-$portVersion.tar.gz"
 PATCHES="alembic-$portVersion.patchset"
```

BTW, for the `ILMBASE_INC_DIR` definition below, I think we will need to change it a little to make it work for secondary architectures. (I'll try to check this asap.)

media-libs/alembic/alembic-1.7.7.recipe
```diff
@@ -79,6 +93,9 @@ BUILD()
 {
 	mkdir -p build
 	cd build
+
+	export ILMBASE_INC_DIR="`finddir B_SYSTEM_HEADERS_DIRECTORY`/OpenEXR"
```
fbrosson: I did not check yet, but it looks like it should be:
```diff
-	export ILMBASE_INC_DIR="`finddir B_SYSTEM_HEADERS_DIRECTORY`/OpenEXR"
+	export ILMBASE_INC_DIR="`finddir B_SYSTEM_DIRECTORY`/$relativeIncludeDir/OpenEXR"
```

fbrosson: I have built your recipe on x86_64 and x86 secondary arch. (Will try on x86 primary arch asap.)

* On x86_64 everything is OK: `100% tests passed, 0 tests failed out of 63`
* On x86 secondary arch `BUILD()` works fine but `INSTALL()` fails. The headers are installed in the correct directories (thanks to the above change), but we need to tell cmake where the libraries should be installed.

I think it won't be too hard to find the correct parameters to pass to cmake. I'll leave that to you, but I can try to help you if you wish.

Hmm, maybe we just need to add:
```
	export ILMBASE_LIB_DIR="`finddir B_SYSTEM_DIRECTORY`/$relativeLibDir/OpenEXR"
```

extrowerk:
@fbrosson : can you post somewhere your x86 secondary arch buildlog? It would help a lot. Thanks. (WP no inline comment, sorry)

fbrosson:
I did not keep the build log, so I need to kick another build.
BTW, I'll try adding these lines after the call to cmake:
```
		-DBIN_INSTALL_DIR="$binDir" \
		-DINCLUDE_INSTALL_DIR="$includeDir" \
		-DLIB_INSTALL_DIR="$libDir"
```

BTW, don't you have a x86_gcc2 hybrid install? (Contributors are supposed to test their recipes on each architecture that is officially supported... ;)

fbrosson:
Right now I'm trying this:
```bash
PATCH()
{
		sed -i \
			-e "s|\(DESTINATION\) lib|\1 $relativeLibDir|;" \
			-e "s|\(ConfigPackageLocation\) lib/cmake/|\1 $relativeLibDir/cmake/|;" \
			lib/Alembic/CMakeLists.txt
	fi
}
```
But there's something else that needs to be taken care of: the `$binDir`.
Well, a (more or less dirty) workaround is to use a commandBinDir.

media-libs/alembic/alembic-1.7.7.recipe
```diff
@@ -6,15 +6,29 @@ of scenes into baked geometry is exactly analogous to the distillation of \
...
 SOURCE_URI="https://github.com/alembic/alembic/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256='ab0c727bfc4aedfe81c2365e604fb457056dc66909575ccfe19ed406e67c002e'
+CHECKSUM_SHA256='831b016afa9e66187b66819a02c0bd198b0f30e501a3552d0634796728cf1bff'
 PATCHES="alembic-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
```
fbrosson:
```diff
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+        commandBinDir=$prefix/bin
+fi
+
PROVIDES="
```
(This should do the trick, with the [`PATCH()`](https://github.com/haikuports/haikuports/pull/2405#issuecomment-377707251) function and the alternate definition of [`ILMBASE_INC_DIR`](https://github.com/haikuports/haikuports/pull/2405/files#r178426291).)

extrowerk:
Nope, i'm absolutely against secondary arches and thing like that.
Unneeded abstractions.

extrowerk:
So let's say it isn't supported on secondary arches.

media-libs/alembic/alembic-1.7.7.recipe
fbrosson: The changes to make this recipe work for x86 secondary arch are not very complicated, so here they are, in case you change your mind.
If you don't wish to mark x86 secondary arch as tested, that's OK. I will approve your PR even if you keep the `?x86` :)

```diff
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+SECONDARY_ARCHITECTURES="x86"
 
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandBinDir=$prefix/bin
+fi
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
```

```diff
 	cmd:abctree
-	lib:libalembic$secondaryArchSuffix = $portVersion compat >= 1
+	lib:libalembic$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES="
```

```diff
 PROVIDES_devel="
 	alembic${secondaryArchSuffix}_devel = $portVersion
-	devel:libalembic$secondaryArchSuffix = $portVersion compat >= 1
+	devel:libalembic$secondaryArchSuffix = $libVersionCompat
 	"
```

```diff
 defineDebugInfoPackage alembic$secondaryArchSuffix \
 	$libDir/libAlembic.so.$portVersion \
-	$binDir/abcconvert \
-	$binDir/abcdiff \
-	$binDir/abcecho \
-	$binDir/abcechobounds \
-	$binDir/abcls \
-	$binDir/abcstitcher \
-	$binDir/abctree
+	$commandBinDir/abcconvert \
+	$commandBinDir/abcdiff \
+	$commandBinDir/abcecho \
+	$commandBinDir/abcechobounds \
+	$commandBinDir/abcls \
+	$commandBinDir/abcstitcher \
+	$commandBinDir/abctree
+
+PATCH()
+{
+	if [ -n "$secondaryArchSuffix" ]; then
+		sed -i \
+			-e "s|\(DESTINATION\) lib|\1 $relativeLibDir|;" \
+			-e "s|\(ConfigPackageLocation\) lib/cmake/|\1 $relativeLibDir/cmake/|;" \
+			lib/Alembic/CMakeLists.txt
+	fi
+}
 
 BUILD()
 {
 	mkdir -p build
 	cd build
 
-	export ILMBASE_INC_DIR="`finddir B_SYSTEM_HEADERS_DIRECTORY`/OpenEXR"
+	export ILMBASE_INC_DIR="`finddir B_SYSTEM_DIRECTORY`/$relativeIncludeDir/OpenEXR"
```

fbrosson:
This [comment](https://github.com/haikuports/haikuports/pull/2405#discussion_r178439281) shows the changes you might wish to apply to have the recipe build on a secondary arch.

fbrosson:
Cool! Thanks!
I thought `LIBRARY_INSTALL_DIR` was not enough and that we needed to use the `PATCH()` function instead, but maybe I am wrong. So I'll check your recipe right now.

media-libs/alembic/alembic-1.7.7.recipe
```diff
@@ -67,21 +90,25 @@ BUILD_PREREQUIRES="
...
 {
 	mkdir -p build
 	cd build
+
+	export ILMBASE_INC_DIR="`finddir B_SYSTEM_DIRECTORY`/$relativeIncludeDir/OpenEXR"
+
 	cmake .. \
 		-DUSE_HDF5=ON \
 		-DALEMBIC_LIB_USES_BOOST=ON \
+		-DLIBRARY_INSTALL_DIR="$relativeLibDir" \
```
fbrosson: We really need to drop this line:
```diff
-		-DLIBRARY_INSTALL_DIR="$relativeLibDir" \
```
and, instead, add the `PATCH()` function. Otherwise `INSTALL()` fails with this error:
```
mv: cannot stat '/packages/alembic_x86-1.7.7-1/.self/lib/x86/libAlembic.*': No such file or directory
```
Here's the `PATCH()` function that will make `INSTALL()` happy:
```bash
PATCH()
{
		sed -i \
			-e "s|\(DESTINATION\) lib|\1 $relativeLibDir|;" \
			-e "s|\(ConfigPackageLocation\) lib/cmake/|\1 $relativeLibDir/cmake/|;" \
			lib/Alembic/CMakeLists.txt
	fi
}
```

media-libs/alembic/alembic-1.7.7.recipe
```diff
@@ -6,15 +6,38 @@ of scenes into baked geometry is exactly analogous to the distillation of \
...
 SOURCE_URI="https://github.com/alembic/alembic/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256='ab0c727bfc4aedfe81c2365e604fb457056dc66909575ccfe19ed406e67c002e'
+CHECKSUM_SHA256='831b016afa9e66187b66819a02c0bd198b0f30e501a3552d0634796728cf1bff'
+SOURCE_FILENAME="alembic-$portVersion.tar.gz"
 PATCHES="alembic-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+SECONDARY_ARCHITECTURES="x86"
+
```
fbrosson:
```diff
- 
```
(There's a white space in line 33.)
Everything else is OK :)

PR #2407, opened 2018-03-31
Links : bump, debuginfo, enable some libs

extrowerk:


www-client/links/patches/links-2.15.patchset
```diff
@@ -1,23 +1,38 @@
-From 9b313c4abd8ae832743f9ce70aa728a67e69656e Mon Sep 17 00:00:00 2001
-From: Jerome Duval <jerome.duval@gmail.com>
-Date: Sat, 15 Jul 2017 14:22:23 +0200
-Subject: apply patch from 2.8
+From 44e3ea63ba59013cd233d256a50421963e62bc61 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zolt=C3=A1n=20Mizsei?= <zmizsei@extrowerk.com>
```
fbrosson: The patchset for version 2.14 is made of two patches, both from @korli. You have merged them into a patchset made of a single patch.
Are you sure he is OK with your dropping the original authorship?
I've compared both patchsets with Emacs and it looks like there aren't so many changes between the patchsets of versions `2.14` vs `2.15`.
So maybe you should keep the original authorship, unless @korli is OK with the new authorship, of course.

PR #2411, opened 2018-04-02
Add recipe for BeScreenCapture 2.3.2

jackburton79:


fbrosson:
Thanks @jackburton79 for this update. I'll test the recipe asap.
BTW, would you like to drop `bescreencapture-2.3.recipe`?
(We would still have the recipes for several older recipes, as well as for 2.3.1, so dropping 2.3 and adding 2.3.2 will reduce the git diff  :)
Thanks!

jackburton79:
Yep. That's fine for me.

Il lun 2 apr 2018, 13:08 fbrosson <notifications@github.com> ha scritto:

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+to any media format supported in Haiku.
+BeScreenCapture can record either the entire screen, or just a section you \
+select."
+
+SUMMARY_inputfilter="Shortcut handler for BeScreenCapture"
+DESCRIPTION_inputfilter="Input Server Addon for BeScreenCapture. Allows the \
+user to launch BeScreenCapture and start/stop recording using a keyboard \
+combination (CTRL-COMMAND-SHIFT + R)."
+
+HOMEPAGE="https://github.com/jackburton79/bescreencapture/releases"
```
fbrosson: I'm OK with the current `HOMEPAGE`, but I thought you might wish to change it to its `..`:
```diff
-HOMEPAGE="https://github.com/jackburton79/bescreencapture/releases"
+HOMEPAGE="https://github.com/jackburton79/bescreencapture"
```

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+COPYRIGHT="2014-2018 Stefano Ceccherini"
+LICENSE="BSD (3-clause)
+	MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v${portVersion}.tar.gz"
```
fbrosson: Cosmetic suggestion:
```
-SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v${portVersion}.tar.gz"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v$portVersion.tar.gz"
```

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
```diff
@@ -0,0 +1,72 @@
...
+CHECKSUM_SHA256="c38d0860e97ab2b4070a0d62c0ea554e09328601f253d6c46b87dede7485830f"
+SOURCE_FILENAME="bescreencapture-${portVersion}.tar.gz"
+SOURCE_DIR="bescreencapture-${portVersion}"
```
fbrosson: We don't need to define `SOURCE_FILENAME` because haikuporter already sets it to a default value, which is what we have here. So you can drop it. (Well, you can also keep it, but it's nicer to drop it :)
```
-SOURCE_FILENAME="bescreencapture-${portVersion}.tar.gz"
-SOURCE_DIR="bescreencapture-${portVersion}"
+SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"
```
(For the curly braces, that's the same remark: we can keep them, or drop them. Please feel free to do whatever you prefer.)

Regarding the removal of the recipe for version `2.32`, as well as for the changes suggested in this review, I guess you know how to update the PR. But please feel free to ask if you need help.

jackburton79: What about SOURCE-DIR ?
Is it needed ? I see you removed SOURCE_DIR and kept SOURCE_FILENAME instead.

fbrosson: Thanks for the update!

* `SOURCE_DIR` is not required for this recipe because the name of the directory at the top level of the tarball is `bescreencapture-2.3.2`, and this is exactly what haikuporter uses as the default value for `SOURCE_DIR`.
In other words, we can drop the `SOURCE_DIR`, or keep it. Both would work fine.
(But we prefer to drop it, since it is not required.)

* Regarding the `SOURCE_FILENAME`, although the recipe works without it, we prefer to have it, so that haikuporter will rename the tarball to something that looks nicer than just `v2.3.2.tar.gz`.

So once you have dropped the `SOURCE_DIR` and added the `SOURCE_FILENAME`:
```diff
-SOURCE_DIR="bescreencapture-${portVersion}"
+SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"
```
the PR will be ready for a merge :)

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
```diff
@@ -11,13 +11,12 @@ user to launch BeScreenCapture and start/stop recording using a keyboard \
...
+COPYRIGHT="2014-2018 Stefano Ceccherini"
 LICENSE="BSD (3-clause)
 	MIT"
 REVISION="1"
-SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v${portVersion}.tar.gz"
-CHECKSUM_SHA256="373169858436b2aa626264911beed6f800011adbf443e9972e8af822fc17cf79"
-SOURCE_FILENAME="bescreencapture-${portVersion}.tar.gz"
+SOURCE_URI="https://github.com/jackburton79/bescreencapture/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="c38d0860e97ab2b4070a0d62c0ea554e09328601f253d6c46b87dede7485830f"
 SOURCE_DIR="bescreencapture-${portVersion}"
```
fbrosson: Oops, my [previous message](https://github.com/haikuports/haikuports/pull/2411#discussion_r179114557) is hidden.
```diff
+SOURCE_FILENAME="bescreencapture-$portVersion.tar.gz"
-SOURCE_DIR="bescreencapture-${portVersion}"
```
(The recipe will be ready for a merge after this change.)

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
pulkomandy: A note on why SOURCE_FILENAME is important: our builders can store a mirror of all sources we use when they build things, for future reference. It is much easier to handle that mirror if sourcefiles are properly named, and it also may avoid name conflicts (which will happen sooner or later if only version numbers are used to name the source archives).

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
jackburton79: Done.
What's still missing in the recipe, is the placement of the Readme file. Is there an example t o look at, for this ?

fbrosson:
@jackburton79, if you would like to add the `README.md` file, we have many recipes which could serve as an example. (All our `*_libretro-*.recipe` recipes do that.)

For your convenience, I suggested a variant in [this comment](https://github.com/haikuports/haikuports/pull/2411#discussion_r179209009). But if you prefer to use `mkdir` and `cp`, that's also OK.
(Well, I think using `install -m 0755 -d "$docDir"` + `install -m 0644 -t "$docDir" README.md` is nicer :)
Thanks!

BTW, in case you prefer to not install the `README.md` file, that's OK too.

haiku-apps/bescreencapture/bescreencapture-2.3.2.recipe
```diff
@@ -60,10 +59,15 @@ BUILD()
...
-	cp -a inputfilter/objects/BeScreenCaptureInputFilter $addOnsDir/input_server/filters
+	install -d "$appsDir"
+	install -t "$appsDir" objects/BeScreenCapture
+
+	install -d "$addOnsDir"/input_server/filters
+	install -t "$addOnsDir"/input_server/filters \
+		inputfilter/objects/BeScreenCaptureInputFilter
+
+	install -d "$docDir"
+	install -t "$docDir" README.md
```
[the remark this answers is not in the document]
alaviss: Well, from [`install(1)`](https://man.openbsd.org/install.1)

> **-m** _mode_ Specify an alternate _mode_. The default mode is set to rwxr-xr-x (0755). The specified mode may be either an octal or symbolic value; see chmod(1) for a description of possible mode values.

> So, would you recommend the following changes?

Sure, after all, we don't want an executable `README.md` file :)

The main reason why I usually use `-m755` is for consistency with lines that requires `-m644`, which appears somewhat frequent and have become a muscle memory of mine :smile:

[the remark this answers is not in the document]
fbrosson: I don't know how I managed to not notice this before!
So thanks for your remark :)
Now I know we should always use `-m`.
(Everything is OK, so we can merge.)

PR #2412, opened 2018-04-02
GemRB: bump

extrowerk:


games-engines/gemrb/gemrb-0.8.5.recipe
```diff
@@ -0,0 +1,83 @@
...
+Engine (the one of Baldur's Gate, Icewind Dale, Planescape: Torment, ...) \
+to Linux/Unix, MacOS X and Windows with some enhancements. Would \
+you like to create a game like Baldur's Gate?"
+HOMEPAGE="https://github.com/gemrb/gemrb"
+COPYRIGHT="2003-2017 The GemRB Team"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/gemrb/gemrb/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="d927e1eebd13a4643db363087fc61c5d2242e8a614da261f15bf4e2b1d94965c"
+SOURCE_DIR="gemrb-$portVersion"
```
fbrosson:
```diff
+SOURCE_FILENAME="gemrb-$portVersion.tar.gz"
-SOURCE_DIR="gemrb-$portVersion"
```
(No need for a `SOURCE_DIR`.)

games-engines/gemrb/gemrb-0.8.5.recipe
```diff
@@ -0,0 +1,83 @@
...
+PATCHES="gemrb-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	gemrb = $portVersion
```
fbrosson:
```diff
 PROVIDES="
-	gemrb = $portVersion
+	gemrb = $portVersion$secondaryArchSuffix
```
Other changes are required to have a correct support for secondary architectures.
(You either need to add the `$secondaryArchSuffix` to the `cmd:` items, or define and use a `commandBinDir`.

games-engines/gemrb/gemrb-0.8.5.recipe
```diff
@@ -0,0 +1,83 @@
...
+Engine (the one of Baldur's Gate, Icewind Dale, Planescape: Torment, ...) \
+to Linux/Unix, MacOS X and Windows with some enhancements. Would \
+you like to create a game like Baldur's Gate?"
+HOMEPAGE="https://github.com/gemrb/gemrb"
+COPYRIGHT="2003-2017 The GemRB Team"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/gemrb/gemrb/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="d927e1eebd13a4643db363087fc61c5d2242e8a614da261f15bf4e2b1d94965c"
+PATCHES="gemrb-$portVersion.patchset"
```
fbrosson: Oops, you forgot to insert the `SOURCE_FILENAME`:
```diff
+SOURCE_FILENAME="gemrb-$portVersion.tar.gz"
 PATCHES="gemrb-$portVersion.patchset"
```
According to [wiki/HaikuPorter-BuildRecipes#source_filename](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#source_filename), this is a possibility, but there's another page which says that a recipe should define a `SOURCE_FILENAME` in order to rename the tarball if the original name does not include the basename of the package (or at least something close enough to it).
(Sorry, I can't find that page right now...)

games-engines/gemrb/gemrb-0.8.5.recipe
```diff
@@ -0,0 +1,84 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	# Just icons
+	rm -rf $prefix/share
```
alaviss: Shouldn't we embed the icon to the executable instead of dropping it like this?

games-engines/gemrb/patches/gemrb-0.8.5.patchset
```diff
@@ -0,0 +1,47 @@
...
++		set(LAYOUT "haiku")
+ 	elseif (APPLE)
+ 		set(LAYOUT "bundle")
+ 		# favor mac frameworks over unix libraries
+@@ -111,6 +113,24 @@ elseif (${LAYOUT} MATCHES "fhs")
+ 	SET_PATH( SVG_DIR ${CMAKE_INSTALL_PREFIX}/share/icons/hicolor/scalable/apps )
+ 	SET_PATH( MENU_DIR ${CMAKE_INSTALL_PREFIX}/share/applications )
+ 	SET_PATH( EXAMPLE_CONF_DIR ${SYSCONF_DIR} )
++elseif (${LAYOUT} MATCHES "haiku")
++	SET_PATH( LIB_DIR ${CMAKE_INSTALL_PREFIX}/lib/gemrb )
```
alaviss: Would this work with secondaryArch?

extrowerk:
> Hmm, iirc, Icon-O-Matic has recently gained command line support, hasn't it?

But it still can't handle png, ico, icns and stuff like that.
Direct conversion for SVG is not a right way as hvif can't handle 
everything like svg.

fbrosson:
OKay, thanks.
Maybe we could use tools from the netpbm package, with sed/cut/tr/awk/perl, to create an rdef file. But this seems quite complicated right now ;)

pulkomandy:
HVIF is a binary and tightly-packed format. It is unlikely that you can generate it without dedicated tools.

However, we still support bitmap icons, and these should be possible to generate. But is it worth it when it is easy enough to draw a nice and native HVIF version instead?

games-engines/gemrb/patches/gemrb-0.8.5.patchset
alaviss: @extrowerk Please add detection for x86 secondary arch as well, according to the patch, the libs will ends up at `$prefix/lib` for secondary architecture, which may not be desirable.

PR #2417, opened 2018-04-03
git: add recipe for 2.17.0 but mark it as untested.

fbrosson:
~Also comment out the dependency of `git_svn` on the missing `svn_alien` in order to be able to use haikuporter's `--test` feature after it was taught to examine the `TEST_REQUIRES` variable.~

### DIFF
--- a/dev-vcs/git/git-2.17.0.recipe
+++ b/dev-vcs/git/git-2.17.0.recipe
@@ -8,22 +8,22 @@ It outclasses SCM tools like Subversion, CVS, Perforce, and ClearCase with \
 features like cheap local branching, convenient staging areas, and multiple \
 workflows."
 HOMEPAGE="https://git-scm.com/"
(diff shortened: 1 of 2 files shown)


dev-vcs/git/git-2.17.0.recipe
```diff
@@ -97,6 +97,7 @@ REQUIRES_email="
 	"
 REQUIRES_svn="
 	haiku
+#	alien_svn
```
fbrosson: @pulkomandy, may I merge this PR with the `alien_svn` commented out?
(I have marked git 2.17.0 as untested on all architectures, althought it there are no regressions compared to 2.16.3 and 2.16.2.)
If I uncomment this line then I cannot test the recipe, as explained in https://github.com/haikuports/haikuporter/issues/151#issuecomment-377724701

mmlr:
I have fixed the alien_svn build on gcc >= 5 this morning, so no commenting out should be needed.

PR #2424, opened 2018-04-06
Ncurses: bump

extrowerk:


sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,223 @@
+SUMMARY="New curses library"
+HOMEPAGE="http://www.gnu.org/software/ncurses/ncurses.html"
+SOURCE_URI="http://ftp.gnu.org/pub/gnu/ncurses/ncurses-$portVersion.tar.gz"
```
fbrosson: Shall we switch to https and prepend the URI of the mirror redirector?
```diff
-SOURCE_URI="http://ftp.gnu.org/pub/gnu/ncurses/ncurses-$portVersion.tar.gz"
+SOURCE_URI="https://ftpmirror.gnu.org/ncurses/ncurses-$portVersion.tar.gz
+	https://ftp.gnu.org/gnu/ncurses/ncurses-$portVersion.tar.gz"
```
Notice: The `/gnu/` is not required for the first URI (i.e. the one with the `ftpmirror.gnu.org`) but it still required for the second URI (i.e. the one with the `ftp.gnu.org`).

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,223 @@
...
+fi
+
+BUILD_REQUIRES="
+	"
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:gcc$secondaryArchSuffix
+	cmd:grep
+	cmd:ld$secondaryArchSuffix
+	cmd:libtool
```
fbrosson:
```diff
-	cmd:libtool
+	cmd:libtoolize$secondaryArchSuffix
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,224 @@
...
+	https://ftp.gnu.org/gnu/ncurses/ncurses-$portVersion.tar.gz"
+CHECKSUM_SHA256="aa057eeeb4a14d470101eff4597d5833dcef5965331be3528c08d99cebaa0d17"
+SOURCE_DIR="ncurses-$portVersion"
+LICENSE="MIT"
+COPYRIGHT="1998-2017 Free Software Foundation, Inc."
+REVISION="1"
+ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PATCHES="ncurses-6.1.patchset"
```
fbrosson: Shall we sort some variables and introduce a few custom variables?
```diff
 HOMEPAGE="https://www.gnu.org/software/ncurses/ncurses.html"
+COPYRIGHT="1998-2017 Free Software Foundation, Inc."
+LICENSE="MIT"
+REVISION="1"
 SOURCE_URI="https://ftpmirror.gnu.org/ncurses/ncurses-$portVersion.tar.gz
 	https://ftp.gnu.org/gnu/ncurses/ncurses-$portVersion.tar.gz"
 CHECKSUM_SHA256="aa057eeeb4a14d470101eff4597d5833dcef5965331be3528c08d99cebaa0d17"
 SOURCE_DIR="ncurses-$portVersion"
-LICENSE="MIT"
-COPYRIGHT="1998-2017 Free Software Foundation, Inc."
-REVISION="1"
+PATCHES="ncurses-$portVersion.patchset"
+
 ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
 SECONDARY_ARCHITECTURES="x86_gcc2 x86"
  
-PATCHES="ncurses-6.1.patchset"
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%.*}"
 
 PROVIDES="
-	ncurses6$secondaryArchSuffix = $portVersion compat >= 6.1
+	ncurses6$secondaryArchSuffix = $portVersionCompat
-	lib:libform$secondaryArchSuffix = $portVersion compat >= 6.1
+	lib:libform$secondaryArchSuffix = $libVersionCompat
```
(same changes for the other `lib:` lines.)

Note: I'm not sure about `libVersionCompat`. Maybe we should use 
```
libVersionCompat="$libVersion compat >= ${libVersion%.*}"
```
instead of
```
```

extrowerk:
@fbrosson : I'm not sure, I don't really know, what everything could be break with your changes, maybe my compat change is enough to make a complete disaster. Let's wait for other opinions.

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+The ncurses code was developed under GNU/Linux. It has been in use for some \
+time with OpenBSD as the system curses library, and on FreeBSD and NetBSD as \
+an external package. It should port easily to any ANSI/POSIX-conforming UNIX. \
+It has even been ported to OS/2 Warp!
+The distribution includes the library and support utilities, including a \
+terminfo compiler tic, a decompiler infocmp, clear, tput, tset, and a \
+termcap conversion tool captoinfo. Full manual pages are provided for the \
+library and tools."
+HOMEPAGE="https://www.gnu.org/software/ncurses/ncurses.html"
+COPYRIGHT="1998-2017 Free Software Foundation, Inc."
```
fbrosson:
```diff
-COPYRIGHT="1998-2017 Free Software Foundation, Inc."
+COPYRIGHT="1998-2018 Free Software Foundation, Inc."
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://ftpmirror.gnu.org/ncurses/ncurses-$portVersion.tar.gz
+	https://ftp.gnu.org/gnu/ncurses/ncurses-$portVersion.tar.gz"
+CHECKSUM_SHA256="aa057eeeb4a14d470101eff4597d5833dcef5965331be3528c08d99cebaa0d17"
+SOURCE_DIR="ncurses-$portVersion"
+PATCHES="ncurses-6.1.patchset"
```
fbrosson:
```diff
-PATCHES="ncurses-6.1.patchset"
+PATCHES="ncurses-$portVersion.patchset"
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+
+ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= 6.1"
+portVersionCompat="$libVersionCompat"
```
fbrosson:
```diff
 libVersion="$portVersion"
 libVersionCompat="$libVersion compat >= 6.1"
-portVersionCompat="$libVersionCompat"
+portVersionCompat="$portVersion compat >= 5"
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+
+PROVIDES="
+	ncurses6$secondaryArchSuffix = $portVersionCompat
```
fbrosson:
```diff
 PROVIDES="
-	ncurses6$secondaryArchSuffix = $portVersionCompat
+	ncurses6$secondaryArchSuffix = $portVersion
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+	lib:libncursesw$secondaryArchSuffix = $libVersionCompat
+	lib:libncurses++$secondaryArchSuffix = $libVersionCompat
+	lib:libncurses++w$secondaryArchSuffix = $libVersionCompat
+	lib:libpanel$secondaryArchSuffix = $libVersionCompat
+	lib:libpanelw$secondaryArchSuffix = $libVersionCompat
+	"
+if [ -z "$secondaryArchSuffix" ]; then
+	PROVIDES="$PROVIDES
+		cmd:captoinfo = $portVersion compat >= 5
+		cmd:clear = $portVersion compat >= 5
```
fbrosson:
```diff
 if [ -z "$secondaryArchSuffix" ]; then
 	PROVIDES="$PROVIDES
-		cmd:captoinfo = $portVersion compat >= 5
-		cmd:clear = $portVersion compat >= 5
-		cmd:infocmp = $portVersion compat >= 5
-		cmd:infotocap = $portVersion compat >= 5
-		cmd:reset = $portVersion compat >= 5
-		cmd:tabs = $portVersion compat >= 5
-		cmd:tic = $portVersion compat >= 5
-		cmd:toe = $portVersion compat >= 5
-		cmd:tput = $portVersion compat >= 5
-		cmd:tset = $portVersion compat >= 5
+		cmd:captoinfo = $portVersionCompat
+		cmd:clear = $portVersionCompat
+		cmd:infocmp = $portVersionCompat
+		cmd:infotocap = $portVersionCompat
+		cmd:reset = $portVersionCompat
+		cmd:tabs = $portVersionCompat
+		cmd:tic = $portVersionCompat
+		cmd:toe = $portVersionCompat
+		cmd:tput = $portVersionCompat
+		cmd:tset = $portVersionCompat
 		"
 fi
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+		cmd:tset = $portVersion compat >= 5
+		"
+fi
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+if [ -n "$secondaryArchSuffix" ]; then
+	REQUIRES="$REQUIRES
+		ncurses6 == $portVersion base
+		"
```
fbrosson:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
-if [ -n "$secondaryArchSuffix" ]; then
-	REQUIRES="$REQUIRES
-		ncurses6 == $portVersion base
-		"
-fi
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+fi
+
+PROVIDES_devel="
+	ncurses6${secondaryArchSuffix}_devel = $portVersionCompat
+	cmd:ncurses6_config${secondaryArchSuffix} = $libVersionCompat
+	cmd:ncursesw6_config${secondaryArchSuffix} = $libVersionCompat
```
fbrosson:
```diff
 PROVIDES_devel="
-	ncurses6${secondaryArchSuffix}_devel = $portVersionCompat
+	ncurses6${secondaryArchSuffix}_devel = $portVersion
-	cmd:ncurses6_config${secondaryArchSuffix} = $libVersionCompat
-	cmd:ncursesw6_config${secondaryArchSuffix} = $libVersionCompat
+	cmd:ncurses6_config$secondaryArchSuffix = $libVersionCompat
+	cmd:ncursesw6_config$secondaryArchSuffix = $libVersionCompat
-	devel:libform${secondaryArchSuffix} = $libVersionCompat
-	devel:libformw${secondaryArchSuffix} = $libVersionCompat
-	devel:libmenu${secondaryArchSuffix} = $libVersionCompat
-	devel:libmenuw${secondaryArchSuffix} = $libVersionCompat
-	devel:libncurses${secondaryArchSuffix} = $libVersionCompat
-	devel:libncursesw${secondaryArchSuffix} = $libVersionCompat
-	devel:libncurses++${secondaryArchSuffix} = $libVersionCompat
-	devel:libncurses++w${secondaryArchSuffix} = $libVersionCompat
-	devel:libpanel${secondaryArchSuffix} = $libVersionCompat
-	devel:libpanelw${secondaryArchSuffix} = $libVersionCompat
+	devel:libform$secondaryArchSuffix = $libVersionCompat
+	devel:libformw$secondaryArchSuffix = $libVersionCompat
+	devel:libmenu$secondaryArchSuffix = $libVersionCompat
+	devel:libmenuw$secondaryArchSuffix = $libVersionCompat
+	devel:libncurses$secondaryArchSuffix = $libVersionCompat
+	devel:libncursesw$secondaryArchSuffix = $libVersionCompat
+	devel:libncurses++$secondaryArchSuffix = $libVersionCompat
+	devel:libncurses++w$secondaryArchSuffix = $libVersionCompat
+	devel:libpanel$secondaryArchSuffix = $libVersionCompat
+	devel:libpanelw$secondaryArchSuffix = $libVersionCompat
 	"
```

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,222 @@
...
+	devel:libmenuw${secondaryArchSuffix} = $libVersionCompat
+	devel:libncurses${secondaryArchSuffix} = $libVersionCompat
+	devel:libncursesw${secondaryArchSuffix} = $libVersionCompat
+	devel:libncurses++${secondaryArchSuffix} = $libVersionCompat
+	devel:libncurses++w${secondaryArchSuffix} = $libVersionCompat
+	devel:libpanel${secondaryArchSuffix} = $libVersionCompat
+	devel:libpanelw${secondaryArchSuffix} = $libVersionCompat
+	"
+REQUIRES_devel="
+	ncurses6${secondaryArchSuffix} == $portVersion base
```
fbrosson:
```diff
 REQUIRES_devel="
-	ncurses6${secondaryArchSuffix} == $portVersion base
+	ncurses6$secondaryArchSuffix == $portVersion base
 	"
```

sys-libs/ncurses/ncurses6-6.1.recipe
fbrosson: Remark: I'm not sure about this:
```diff
-	cmd:ncurses6_config${secondaryArchSuffix} = $libVersionCompat
-	cmd:ncursesw6_config${secondaryArchSuffix} = $libVersionCompat
+	cmd:ncurses6_config$secondaryArchSuffix = $portVersion
+	cmd:ncursesw6_config$secondaryArchSuffix = $portVersion
```

sys-libs/ncurses/ncurses6-6.1.recipe
fbrosson: Oops, sorry @extrowerk for having written the above comment.
I though the dependency was there for no reason, but I just saw this comment in `BUILD`:
```
	# For the secondary architecture we don't want the programs nor the
	# terminfo DB. The main package will depend on the primary package which
	# provides all this.
```
So I guess you will need to revert the change.
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
+if [ -n "$secondaryArchSuffix" ]; then
+	REQUIRES="$REQUIRES
+		ncurses6 == $portVersion base
+		"
+fi
```

Hmm, that said, maybe that's not such a bad idea. Do we really need to make the base package for the secondary arch depend on that of the primary arch? Well, I don't know. Let's see what the other contributors think.
(Feel free to revert the change when you wish.)

sys-libs/ncurses/ncurses6-6.1.recipe
```diff
@@ -0,0 +1,217 @@
...
+CHECKSUM_SHA256="aa057eeeb4a14d470101eff4597d5833dcef5965331be3528c08d99cebaa0d17"
+SOURCE_DIR="ncurses-$portVersion"
+PATCHES="ncurses-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64 arm"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= 6.1"
+portVersionCompat="$portVersion compat >= 5"
```
alaviss:
```diff
-portVersionCompat="$portVersion compat >= 5"
+portVersionCompat="$portVersion compat >= 6"
```

fbrosson: For `ncurses6-6.0` the `cmd:` entries have `compat >= 5`, so I thought we should keep them, but I might be wrong.

alaviss: The meaning of `compat` is to signify the latest version that the versioned entry is backward compatible to, so I think using `compat >= 6` should be safer, but I don't really know for sure if it's totally backward compatible with version 5.

fbrosson: I used `find` with `-exec grep -E -l 'cmd:captoinfo|cmd:clear|cmd:infocmp|cmd:infotocap|cmd:reset|cmd:tabs|cmd:tic|cmd:toe|cmd:tput|cmd:tset' {} \;` on all of our recipes and the only result (other than the recipes for ncurses) is `dev-util/fenris/fenris-0.07.recipe` which has a `cmd:tput` in `BUILD_PREREQUIRES` (but without any condition on the version.)

@extrowerk, @alaviss: This means you were both right: We can use `compat >= 6`.
(Sorry for my previous suggestions to use `compat >= 5` :)

```diff
-portVersionCompat="$portVersion compat >= 5"
+portVersionCompat="$portVersion compat >= 6"
```

waddlesplash:
@extrowerk can you try installing a built version of this and see if it breaks the whole system (or even lets you install it?) If it's fine, then I'll merge.

extrowerk:
@waddlesplash ruby, libedit and possibly other things requires lib:libncurses>=6.0

waddlesplash:
The ABI laboratory says that only 2 symbols are gone in 6.1 vs 6.0, and these are noted in the release notes as being "deprecated since 2004." So, it's probably safe to merge without checking literally everything. :)

waddlesplash:
... oh, but you've declared the comatibility to be `>= 6.1`. Why? The soname has not changed.

PR #2425, opened 2018-04-07
gnu-efi: bump to 3.0.8, update COPYRIGHT.

fbrosson:


fbrosson:
FWIW, I have already tested this. Just like release `3.0.4`, release `3.0.8` also allows to build the `haiku_loader.efi` on x86_64.
BTW, I have built it from freshly configured build trees, using:
```
jam -q -sHAIKU_BOOT_PLATFORM=efi haiku_loader.efi
```
from a Haiku x86_64 system, and then from a Debian GNU/Linux 9 (stretch) system.

Of course, I also tried booting Haiku (on an EFI-enabled QEMU VM) with the resulting `haiku_loader.efi` which lands in  `objects/haiku/x86_64/release/system/boot/efi/`.

Everything looks OK. I mean, Haiku x86_64 boots fine on my VM when I use the EFI loader built with gnu-efi 3.0.8.

(Remark: I have even replaced the "Return to main menu" by a "Return to Main Menu" in `src/system/boot/platform/efi/video.cpp` to be absolutely sure that the loader I was loading was the one built with gnu-efi 3.0.8, although this verification was not really needed, as I knew I had copied the file at the correct location in my ESP partition :)

**Important:** I had to make this change:
```diff
-	gnu_efi_kernel-3.0.4-2
+	gnu_efi_kernel-3.0.8-1
```
in line 8 of [haiku/build/jam/repositories/HaikuPorts/x86_64](https://github.com/haiku/haiku/blob/d3bdde3594411407b163e862442a84b04f174898/build/jam/repositories/HaikuPorts/x86_64#L8) in my local haiku repo, before running jam.
(This means we will need to apply the above change to this file after this PR gets merged.)

BTW, I did not read the docs yet to learn what I should do to ask for write access to the haiku repo but I will try to :) So, if anyone would like to apply this cnange this would be cool ;) Thanks!

fbrosson:
I've just tested the `haiku_loader.efi` on real hardware and it boots fine :tada:
So, let's merge!

PR #2429, opened 2018-04-08
mlpack: bump

extrowerk:


sci-libs/mlpack/mlpack-3.0.0.recipe
```diff
@@ -0,0 +1,200 @@
...
+	rmdir $prefix/include
+
+	prepareInstalledDevelLib libmlpack
+
+	fixPkgconfig
+
+	packageEntries cli \
+		$binDir/mlpack_adaboost \
+		$binDir/mlpack_approx_kfn \
+		$binDir/mlpack_cf \
```
fbrosson: Since all the `cmd:` entries are in `PROVIDES_cli`, you don't need to list them here.
You can therefore replace the full list by`"$binDir"` after  `packageEntries cli`.
```diff
 	packageEntries cli \
+		"$binDir"
-		$binDir/mlpack_adaboost \
-		$binDir/mlpack_approx_kfn \
-		$binDir/mlpack_cf \
```
...

PR #2432, opened 2018-04-09
ranger: Initial recipe

tts2k:


app-misc/ranger/ranger-1.9.1.recipe
```diff
@@ -0,0 +1,89 @@
+SUMMARY="A simple, vim-like file manager"
+DESCRIPTION="Ranger is a console file manager that gives you greater flexibility and a good overview of your files without having to leave your console. It visualizes the directory tree in two dimensions: the directory hierarchy on one, lists of files on the other, with a preview to the right so you know where you'll be going."
```
fbrosson:
```diff
-DESCRIPTION="Ranger is a console file manager that gives you greater flexibility and a good overview of your files without having to leave your console. It visualizes the directory tree in two dimensions: the directory hierarchy on one, lists of files on the other, with a preview to the right so you know where you'll be going."
+DESCRIPTION="Ranger is a console file manager that gives you greater \
+flexibility and a good overview of your files without having to leave your \
+console. It visualizes the directory tree in two dimensions: the directory \
+hierarchy on one, lists of files on the other, with a preview to the right so \
+you know where you'll be going."
```

tts2k: oh crap, i totally forgot about this

app-misc/ranger/ranger-1.9.1.recipe
```diff
@@ -0,0 +1,93 @@
...
+
+PYTHON_PACKAGES=(python python3)
+PYTHON_VERSIONS=(2.7 3.6)
+for i in "${!PYTHON_PACKAGES[@]}"; do
+pythonPackage=${PYTHON_PACKAGES[i]}
+pythonVersion=${PYTHON_VERSIONS[$i]}
+eval "PROVIDES_${pythonPackage}=\"\
+	${portName}_$pythonPackage = $portVersion\n\
+	cmd:ranger = $portVersion\n\
+	cmd:rifle = $portVersion\
```
alaviss: A problem here is that ranger provides some `cmd:`s. It's currently possible to install both `python3` and `python` version, and the one installed later(?) will have it's `cmd:`s take priority. Normally this should be fine, but since ranger is capable of loading python plugins, I think we should either:

- Only package the `cmd:`s for a specific python version. I would vote for python2 in this case, since it works equally well for all architectures
- Further splitting the package with `_python_cmd` and `_python3_cmd`

fbrosson: With a few `cmd:`-style conflcts we can fix this. Here are the changes, with a few cosmetic neutral changes too :)
```diff
 for i in "${!PYTHON_PACKAGES[@]}"; do
 pythonPackage=${PYTHON_PACKAGES[i]}
 pythonVersion=${PYTHON_VERSIONS[$i]}
-eval "PROVIDES_${pythonPackage}=\"\
-	${portName}_$pythonPackage = $portVersion\n\
-	cmd:ranger = $portVersion\n\
-	cmd:rifle = $portVersion\
-	\"; \
-REQUIRES_$pythonPackage=\"\
-	haiku\n\
-	cmd:python$pythonVersion\
-	\""
+eval "\
+	PROVIDES_$pythonPackage=\"\n\
+		${portName}_$pythonPackage = $portVersion\n\
+		cmd:ranger = $portVersion\n\
+		cmd:rifle = $portVersion\n\
+		\"; \
+	REQUIRES_$pythonPackage=\"\n\
+		haiku\n\
+		cmd:python$pythonVersion\n\
+		\"; \
+	CONFLICTS_$pythonPackage=\"\n\
+		cmd:ranger\n\
+		cmd:rifle\n\
+		\"; \
+	"
 BUILD_REQUIRES="$BUILD_REQUIRES
 	setuptools_$pythonPackage"
```
(I've just tested and it works fine.)
What do you think?
(

alaviss: `ranger` could be used as a python module, so I don't think we should not forbid one from coding with `python3` while running `ranger` with `python2` plugins. But I don't think anyone would have that use case anyway, so this seem fair enough to me.

As of styling, how about this?

```sh
eval "PROVIDES_$pythonPackage='
	${portName}_$pythonPackage = $portVersion
	cmd:ranger = $portVersion
	cmd:rifle = $portVersion
	'"
eval "REQUIRES_$pythonPackage='
	haiku
	cmd:python$pythonVersion
	'"
eval "CONFLICTS_$pythonPackage='
	cmd:ranger
	cmd:rifle
	'"

fbrosson: That's a good idea, for the style.

For the `CONFLICTS` I must say that I'm not as confortable with my own suggestion now that you have broght to my attention that it would not allow some use cases.
The most flexible solution would be to have the conflicts for the _python_cmd, and allow the modules of each python subpackage to be co-installed.
But this would mean that we would have one base package and 4 subpackages...
I'm OK with this but I guess some users would prefer the current layout.

Shall we keep this PR open for 1 or 2 days, so that we can all think about it and see what @tts2k and the other contributors think about this?

Hmm, we might workaround the flexibility issue by switching to a layout similar to the one we have for youtube-dl. But don't think this would be worth for this particular package.

alaviss: Personally I think that we should sacrifice some flexibility for users to make maintainance easier. 

After a few thoughts, I believe your suggestion is pretty good as the usecase I mentioned probably never exists. Most Linux distros only package for a specific version of python, and I haven't seen anyone complaining about that yet. Also, if one really needs those use cases, they should be capable of modifying our recipe, or they could just convert whatever plugin they're using to python3 anyway.

fbrosson: Yeah, I guess you are right, @alaviss. Let's go for your style of the conflicts I suggested.
@tts2k, what do you think?

tts2k: I'm ok with @alaviss style.

PR #2433, opened 2018-04-09
zsh: bump to 5.5.1.

fbrosson:


fbrosson:
Good catch. I had not noticed that a `USER_SETTINGS_FILES` should be used in this recipe.

Regarding a patch to make zsh use `find_directory`, thisi an excellent idea.
(Unfortunately I don't have the time to do that myself, and I am sure you would implement this faster and better than me.)
So, would you like to grab my branch and open another PR to write a nice patch ?

I did not spend much time on this PR. I just updated the checksum, refreshed the patchset and ran the tests on x86_64 and x86, so I'm OK with closing this PR. Well, I have also built zsh 5.4.2 to see if the tests were behving the same way. They all look fine at the beginning, during a few minutes, but then they hang, using all the available CPU. I'm confident that zsh 5.5 is working at least as well as 5.4.2, although I did not play a lot with it.

alaviss:
Sure, I will work on this when I have time. You don't really have to close this PR, I will work on a patch then push it to your branch :)

fbrosson:
Cool, that's an excellent idea.
BTW, I've just changed the authorship of the commit, so that you will appear as the contributor of this PR. After all, you will be doing the most difficult part, so it is fair to have you as the author of the commit.
BTW, I've also rebased against master.

alaviss:
Turns out movings zsh dots were as simple as modifying an enviroment variable. But then the user custom configurations would not be applied due to the fact that the files are now moved elsewhere in the new version. I believe we should have some kind of notice so that the user would know to relocate their configs accordingly. But how should we do it?

pulkomandy:
We did move bash config and history using post install scripts, perhaps look there to see how it was done (you may have to dig old versions of the recipe). If you want to display a message or ask the user what to do, you could use the "alert" command from the script.

fbrosson:
Hmm, what about creating symlinks at the new location, pointing to the old locations? We could do this in a post-install script. That would be safe and we would not upset any user.

fbrosson:
d790719d5ddae9e15cff8ae89eab5df84c701fc4 `bash: also migrate ~/.profile to ~/config/settings `

fbrosson:
I see 3 scenari:

1) We write a post-install script which carefully tests all possible cases, and does what the users should do, i.e. move their settings to the new place. And we could use the `alert` command to let the user know the script has moved the files.

2) We create symlinks from the new location to the old location, and use `alert` to display a warning at install and at every boot, until the users remove the symlinks and moves their config to the right place.

3) Same as 2, but without any alert.

2 is safer than 1, but it is also more complicated and more annoying for the user.

pulkomandy:
I would go with 1. You make "all possible cases" sound complex but really it's just about checking if there is already a config/settings/zsh and doing nothing in that case. Otherwise we just move the files there.

And we are still not in beta, so we can occasionally break things.

An alert at every boot? Seriously? That would make me consider uninstalling the software.

fbrosson:
> We could also left the old configs as is and only apply the new defaults to brand new installations

@alaviss, do you mean: _"Unconditionally ignore the settings in the old location, and only read the settings in the new location (if they exist)."_ ? Yep, that could be an option.

However, I think I'd follow @pulkomandy's arguments for scenario 1. (Thanks @pulkomandy!)

What do you think? Well, please feel free to go for any scenario, including your own suggestion.
After all, you are going to do the difficult part, so the choice is yours ;)

alaviss:
> @alaviss, do you mean: "Unconditionally ignore the settings in the old location, and only read the settings in the new location (if they exist)." ? Yep, that could be an option.

I mean: "If settings is found in the old location, read them instead of the one in the new location".

Why is this possible? It's because zsh store the dotfiles at `$ZDOTDIR`, which defaults to `$HOME` if unset, so we would just need to add the required `export` to `/system/settings/zsh/zshenv`, which is sourced everytime zsh starts, and zsh will source files from the `$ZDOTDIR` of our choice. Better yet, since `zshenv` is a zsh script, we could just add a conditional not to set the variable we found `.z*` files lying around in `$HOME`.

alaviss:
Of course, there's a downside to the `$ZDOTDIR` approach: Files in `$ZDOTDIR` would still needs to be prefixed with a `.`(dot)

Although in theory, we could patch this, I would recommend against doing so as the `$ZDOTDIR` behavior is well documented, any sane scripts/tutorial out there would look for `.z*`(with the dot prefix) instead of `z*`(without the dot prefix) within that folder.

fbrosson:
All this is not trivial, but I'm sure you will do a great job.
(But remember: there is no urgency. So don't forget to sleep :)

alaviss:
So lots has been done, but I've not tested the recipe anywhere other than x86_64.

Some zsh tests are still failing :(, here are some infos about them:
- Some tests just hang/fail in chroot, run fine outside
- Most of utf-8 tests fail

fbrosson:
Your recipe builds fine on x86_64 and x86.
And with a few gcc2 patches (in 4 files) it also builds on x86_gcc2 :)

I did not test anything related to the settings, but your recipe looks good to me.

BTW, the tests seem to hang when they reach:
```
./X03zlebindkey.ztst: starting.
```
But this was already the case with the recipe we initialy had in this PR, as well as for zsh-5.4.2 if I'm not mistaken.

On x86, zsh also seems to be working fine, but when I run zsh I get this line:
```
~> zsh
/packages/zsh-5.5-1/.settings/zsh/zshenv:3: wait failed: no such process
\<> 
```
Moreover, when I un `hp --test zsh` the tests do not seem to work at all:
```
./ztst.zsh:108: wait failed: no such file or directory
./runtests.zsh:12: wait failed: no error
make[1]: *** [Makefile:188: check] Error 1
make[1]: Leaving directory '/sources/zsh-5.5/Test'
```
So maybe we should mark x86 as untested.
That said, this is probably due to the fact that my x86 system is not up to date.

Regarding the x86_gcc2 arch, I can push a commit (later) to this PR to update the patchset so that we can have zsh on that arch if you wish. I can also do that in a separate PR; we could also leave x86_gcc2 without zsh.

alaviss:
You should just push the commits here as well :) For x86, we should mark it as broken instead.

fbrosson:
Okay, I'll push them here :)

Regarding the x86 arch, since zsh builds (and even works) on that arch, I would prefer to have a `?x86` rather than a `!x86`.
(BTW, I'll make that tiny change when I push the gcc2 patch.)

fbrosson:
zsh 5.5 is _not_ broken on x86. I can even use it.
(What's broken is the test suite.)
Imho a `?x86` is better than a `!x86`. And I believe that the reason why the test suite is failing on that arch is unrelated to the arch itself but to the (old) hrev I'm running on my x86 VM.

fbrosson:
That's it. I've just pushed the patch to make zsh build on x86_gcc2.


I've been thinking about adding a `REQUIRES_doc="zsh == $portVersion base"` (with the correct 3-line style.)
**Pros:** If we `pkgman uninstall zsh` then `zsh_doc` will also be uninstalled. If we `pkgman update zsh` then `zsh_doc` will also updated.
**Cons:** With this dependency it won't be possible to install `zsh_doc` without also installing `zsh`. But I guess this won't be a big issue.

alaviss:
The recipe used to contain a `sed` call to disable the usage of `wait3`. I removed it to see if any prob occurs, but haven't found any so far. Probably Haiku used to have bugs with `wait3`, so you could try to add that line back and see if the situation improves ;)

fbrosson:
Good catch.
Now I understand why the test suite was OK on x86 when I created the PR, as I had not removed the sed command.

Well, we _could_ add the `sed --in-place '/HAVE_WAIT3/d' config.h` only for x86, but this would probably not be a good idea because it seems it is not needed on recent hrevs. What do you think?
(Hmm, maybe it can help with the tests. We should probably check this.)

(I'll try it locally, on my old x86, and then on the other architectures.)

alaviss:
It makes no difference for the tests for me, that's why I removed it. If it's still not working for newer hrev of `x86`, then you should open a ticket on Trac.

Please also note that some tests just refuse to run within the chroot, so it might be a good idea to `cd` to the work directory and run `make check` there :)

fbrosson:
Thanks for the hint about running the tests without the chroot. I'll give it a try.

Regarding the x86 arch, my hrev is not _newer_. It is _older_. (Actually, it is _much_ older.)

alaviss:
@fbrosson Have you re-run the tests? It would be nice if we have some logs around, as to aid future contributors who would opt to fix them.

fbrosson:
Yep, I ran the tests 3 days ago, with the `sed --in-place '/HAVE_WAIT3/d' config.h` and for each architecture.
With the sed fix, the tests on my old x86 behave the same as on x86_gcc2 and x86_64.

BTW, I saw that bash 5.5.1 was just released (2 days ago iirc), so I'll give it a try.
I did not bump the recipe (nor the patchset) to 5.5.1 yet, but once this is done I can push a commit to this PR if you agree. BTW, if you wish, I can also rebase my zsh branch against master.

(When I update a PR that is based on a PR whose branch is based on an old HEAD, I tend to duplicate my haikuports tree (without the download subdirectories, of course) in order to be able to `git checkout` _old_ branches without disturbing haikuporter.)

pulkomandy:
There were some fixes to the wait*() functions in Haiku so maybe this  HAVE_WAIT3 hack can be removed because modern versions of Haiku don't have the problem anymore?

fbrosson:
OK, thanks @pulkomandy for the info about the `HAVE_WAIT3`.
(I had indeed added it back, commented out, but I'll remove it completely.)

fbrosson:
@alaviss, I've pushed a tiny change: `move zprofile and zshenv to additional-files.`
I think it is a small improvement, but since the commit will be yours, I can revert it if you prefer the previous version.

app-shells/zsh/zsh-5.5.1.recipe
```diff
@@ -92,6 +96,7 @@ BUILD()
 		--with-tcsetpgrp \
 		--enable-zsh-secure-free
 
+	sed --in-place '/HAVE_WAIT3/d' config.h
```
alaviss: I thought you decided to remove this? Is there any reason why this is coming back?

alaviss:
@fbrosson 

> @alaviss, I've pushed a tiny change: move zprofile and zshenv to additional-files.
I think it is a small improvement, but since the commit will be yours, I can revert it if you prefer the previous version.

Thanks! The reason why those where in the recipe is that they were small, then I forgot to move them out when I added the migration path :smile:

fbrosson:
> If possible, could you please file an issue for the testsuite failures?

Do you mean on the upstream project? Or in Haiku?
BTW, I think you would do this better than me ;)

alaviss:
> It will be fun because the squashed commit will be yours, so you will be approving a PR which will push your own commit :)

I usually prefer a rebase-merge over squashing, since commit messages and authorship are correctly preserved. So if you want to do a rebase-merge, it would be nice to also drop unneeded commits (i.e the add/remove `HAVE_WAIT3` commits can be dropped), which you can perform with `git rebase -i`, then you will have to do a force-push.

With that said, I'm also fine with squashing, so please just merge with the way you find comfortable.

alaviss:
> Do you mean on the upstream project? Or in Haiku?
BTW, I think you would do this better than me ;)

In HaikuPorts, of course :) I'm a bit busy rn so I can't do it (yet). Thanks!

fbrosson:
(OK for the issue in HaikuPorts. I can create it. But since there is no urgency, I can let you open it later if you agree.)

For the merge, the only thing we loose with the "Squash and merge" method is the original GPG signature.
If I merge it with this method, the commit won't appear as _Verified_ because I'm not the author of the to-be-squashed commit.
If you merge it, with the same method, it will probably appear as  _Verified_ (because the contributor doing the merge -you- will also be the author of the to-be-squashed commit.)
Yeah, I know, the GPG signature won't be your yours but the generic one, from GitHub.

So, would you like to use the "Squash and merge" method?
(Don't forget to remove non usefull comments (I mean, the comments about the fixes/updates) in the detailed commit message.)

alaviss:
> Yeah, I know, the GPG signature won't be your yours but the generic one, from GitHub.

This could be walked around if you merge from the command line like I did for #2026 (although I accidentally made some noise as well). Since I have found myself some free time I can do a rebase-merge now. Here is how I'm going to do it:

- Pull the latest changes and rebase with `master`
- Remove unneeded commits
- Force push for you to review the commit timeline (note, because of the nature of force-push you will have to reclone the branch once I've pushed)
- Merge when we both agree on the timeline and contents of each commits (merging will have to be done by me).

Of course, if you prefer squash-merge, then I can also do that. I usually like rebase-merge as it will preserve each commit and their timeline, giving readers a view on how things has been developed, and why, combined with the ease of reverting any undesirable changes.

fbrosson:
Well, I'm not very fond of the "Rebase and merge" method. Imo a squashed commit is better because it adds less noise. 
For merging PR 2026 (i.e. pkgconf and friends) it was nice to use the "Rebase and merge" method because we wanted to keep the 1 commit for each of the 4 recipes.

The other use case for the "Rebase and merge" method is that, if I'm not mistaken, it allows to preserve your original GPG signature. But I think this method requires that you first rebase your commit against master. And, as we know, we also need to add the ` (#2433)` in the commit title _before_ the merge.

To tell the truth, I'm a bit curious about how GitHub will show this PR if you merge it with the "Squash and merge" method. But again, I'm okay with any method of your choice :)

**Regarding the migration scripts, I did not test them.** I only installed zsh and tried a simple `for i in 1 2 3; do echo $i; done` to see if zsh was "working". Yeah, that's a really basic test :smile:

fbrosson:
> This could be walked around if you merge from the command line

Hmm, I tried a merge from the command line a very long time ago, but it was not a great success :(
The result was 9763e2ed998965886827ba7efd04329fb3873dd1 with a duplicate (but empty) commit: f33fe87b7dccceea79263e9f18d3ea5581a81eaa. (I later apologized, 3 commits later, in b3e8768b0d90b54bf68c0b13a84e6bc3fb053f51 :)

My understanding is that if we want to have a PR be shown as merged we need to merge it on GitHub's web interface, using any of the available methods.

pulkomandy:
I'm going to sound annoying but I would appreciate if the Gerrit flow was used instead. It allows to review commits one by one, and ensures we keep a clean history (and allows for easily merging only parts of a PR).

I will be experimenting with gerrithub on my personal repos, to see if that is a workable solution. And I'm not going to spend any effort trying to switch haikuports to Gerrit, because I spent enough energy doing it for Haiku itself :)

fbrosson:
Hmm, I thought gerrit was only used for the repos that we mirror to https://github.com/haiku

So, should we start using gerrit for haikuports too?
Are there any plans to stop accepting PRs on GitHub, and only accepting changes in gerrit?
(I think this is not going to attract potential contributors.)

alaviss:
@fbrosson 
> To tell the truth, I'm a bit curious about how GitHub will show this PR if you merge it with the "Squash and merge" method. But again, I'm okay with any method of your choice :)

Sure, then I'm going with my plan :) The commits should now reflect what they do, and show a clean history of how things were done. Reviewing by commits is, if I'm not mistaken, the Gerrit way. I would appreciate an another pair of eyes looking at the commits though, and if everything looks nice, I will merge them to `master`.

Adding the PR number might not be needed, since checking what PR was a commit came from can easily be done in the web interface, by clicking on the commit.

> My understanding is that if we want a PR to be shown as merged then we need to merge it on GitHub's web interface, using any of the available methods.

That was what I thought, until I merge #2026 with `git merge` and `git push`, and Github correctly show that the PR is merged :)

fbrosson:
> Sure, then I'm going with my plan :)

:+1:

> Adding the PR number might not be needed, since checking what PR was a commit came from can easily be done in the web interface, by clicking on the commit.

I might be wrong, but I think you can't find the PR from the commit if you don't have a reference to the PR in the commit message.


> That was what I thought, until I merge #2026 with git merge and git push, and GitHub correctly show that the PR is merged :)

I see. I guess the reason why it did not work _for me_ was because the hashes (on the branch the PR was attached to, vs what I pushed to master) were not the same. I guess what I did wrong was that I did not rebase the branch against master before the merge from the command line.
So thanks for the info!

fbrosson:
After all, we can try without adding the reference to this PR!

Regarding the move to the `additional-files`, are you sure you don't want to squash the `zsh: move zprofile and zshenv to additional-files` with `app-shells/zsh: various improvements` ?

Hmm, maybe you don't like it enough to squash it with your commit :smile:
(In that case, please do not hesitate to drop the `zsh: move zprofile and zshenv to additional-files`.)

alaviss:
@fbrosson 

Then I would have to repatch the zsh bump commit, so I would like to avoid that :smile: 

It worked :) Open a commit and you should see the PR number in them :D

fbrosson:
Correct!

PR #2438, opened 2018-04-10
mozc: bump version, Support x86_gcc2 environment

hanya:
Now mozc port can be built on x86_gcc2 environment.


app-i18n/mozc/mozc-2.23.2785.102.4.recipe
```diff
@@ -66,15 +79,22 @@ INSTALL()
...
 	mkdir -p $MozcDataDir/zinnia
 	mkdir -p $MozcDataDir/images
 	mkdir -p $dataDir/locale/catalogs/x-vnd.Mozc-InputMethod
 	mkdir -p $dataDir/locale/catalogs/x-vnd.Mozc-MozcTool
-	mkdir -p $addOnsDir/input_server/methods
-
-	cp $MozcOutDir/mozc $addOnsDir/input_server/methods/mozc
+	mkdir -p $InputMethodDir
 
+	cp $MozcOutDir/mozc $InputMethodDir/mozc
```
fbrosson: So, here's a variant which works the same way as your original recipe. I don't know if this variant is an improvement. I'm OK with your recipe. I'm just suggesting this variant to see what you and the other contributors think about it.
```diff
 	cp $MozcOutDir/mozc $InputMethodDir/mozc
-if [ "$targetArchitecture" == x86_gcc2 ]; then
-	cp $MozcOutDir/mozc_task $MozcDataDir/mozc_task
-fi
+	if [ -n "$secondaryArchSuffix" ]; then
+		mkdir -p $addOnsDir/../input_server/methods
+		ln -r -s $InputMethodDir/mozc \
+			$addOnsDir/../input_server/methods/mozc${secondaryArchSuffix/_/-} 
+	fi
+	if [ "$targetArchitecture" = x86_gcc2 ]; then
+		cp $MozcOutDir/mozc_task $MozcDataDir/mozc_task
+	fi
 	cp $MozcOutDir/mozc_server $MozcDataDir/mozc_server
```

The result of the above code, on x86_gcc2, is that instead of having just one file:
```
add-ons/input_server/methods/mozc
```
we have a symlink and the file itself:
```
add-ons/input_server/methods/mozc-x86 -> ../../x86/input_server/methods/mozc
add-ons/x86/input_server/methods/mozc
```

I'm suggesting this just to avoid putting x86 code in a directory meant for x86_gcc2.
Again, I don't know if this is an improvement.
And I'm OK with your current recipe.

hanya: I'm sorry, I should add some comment about building this package on gcc2 environment.

The mozc can not be build with gcc2 so we have to build mozc with gcc5 but we can not 
link input method to the input server built by gcc2.
I have to split the code into two parts. The one is input server add-on built with gcc2 
even x86(gcc5) is choosen as secondary architecture to build the package. For this way, 
cmd:g++ (without $secondaryArchSuffix) is required if $targetArchitecture is "x86_gcc2". 
The another is mozc_task which executes mozc built with gcc5. This part is built as single add-on on x86_64 environment.
The add-on and the mozc_task interprocess each other by the Message system.

Therefore, mozc file for the input method should be placed under non x86 location 
since it is built by gcc2. I could not obtain non x86 add-ons directory by a variable so I choose 
$addOnsDir/../input_server/methods way. But I'm not sure there is a correct way to obtain it.

fbrosson: Hmm, I'm thinking about the gcc2 / gcc5 mix. Maybe we should move the gcc2 part into a separate tiny x86_gcc2 primary arch package and make the big x86 secondary arch package depend on the tiny package.
**Pros:** we don't mix gcc2 code with gcc5 code.
**Cons:** too much work for no valid reason.

Well, splitting is probably a very bad idea. Let's stick to what we have right now.

PR #2445, opened 2018-04-14
fixed building of the projectconceptor plugins 

Paradoxianer:
this makes projectconceptor usable
alos set to lates commit since it contains some spellcorrections

### DIFF
--- a/haiku-apps/projectconceptor/projectconceptor-0.1.0.recipe
+++ b/haiku-apps/projectconceptor/projectconceptor-0.1.0.recipe
@@ -11,10 +11,9 @@ LICENSE="BSD (3-clause)
 	MIT
 	Open Tracker License
 	Zlib"
-REVISION="1"
-srcGitRev="8192c2e9f0c46d4dc316e20b2d3d6ecfe81110ba"
-SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
-CHECKSUM_SHA256="d76718776fe85d49a6c6137fff0189abff786c3402d2ec197d811391c016e6f2"
+REVISION="2"
+srcGitRev="4bcf8c73c05213322de347f08a2db6696d97a005"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz" CHECKSUM_SHA256="7b80d51bb6ab1ec2a3f22290b3e35dfcc9f667c652ca5c01eb826d86a21abb52"
 SOURCE_FILENAME="$APP-$portVersion-$srcGitRev.tar.gz"
 SOURCE_DIR="$APP-$srcGitRev"
 
@@ -40,7 +39,8 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	make -C src OBJ_DIR=objects $jobArgs
+	make app  -C src OBJ_DIR=objects $jobArgs
+	make plugins -C src OBJ_DIR=objects $jobArgs
 }
 
 INSTALL()


haiku-apps/projectconceptor/projectconceptor-0.1.0.recipe
```diff
@@ -11,10 +11,9 @@ LICENSE="BSD (3-clause)
 	MIT
 	Open Tracker License
 	Zlib"
-REVISION="1"
-srcGitRev="8192c2e9f0c46d4dc316e20b2d3d6ecfe81110ba"
-SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
-CHECKSUM_SHA256="d76718776fe85d49a6c6137fff0189abff786c3402d2ec197d811391c016e6f2"
+REVISION="2"
+srcGitRev="4bcf8c73c05213322de347f08a2db6696d97a005"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz" CHECKSUM_SHA256="7b80d51bb6ab1ec2a3f22290b3e35dfcc9f667c652ca5c01eb826d86a21abb52"
```
fbrosson:
```diff
-SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz" CHECKSUM_SHA256="7b80d51bb6ab1ec2a3f22290b3e35dfcc9f667c652ca5c01eb826d86a21abb52"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="7b80d51bb6ab1ec2a3f22290b3e35dfcc9f667c652ca5c01eb826d86a21abb52"
```
BTW, it also builds on x86 and the app seems to be working fine.
```diff
-ARCHITECTURES="x86_gcc2 !x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 !x86_64"
```

(Well, on x86 it crashes when I quit the app, but this might be caused by my x86 system being quite old.)

haiku-apps/projectconceptor/projectconceptor-0.1.0.recipe
```diff
@@ -40,7 +39,8 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	make -C src OBJ_DIR=objects $jobArgs
+	make app  -C src OBJ_DIR=objects $jobArgs
+	make plugins -C src OBJ_DIR=objects $jobArgs
```
fbrosson: You might wish to combine these two lines:
```diff
-	make app  -C src OBJ_DIR=objects $jobArgs
-	make plugins -C src OBJ_DIR=objects $jobArgs
+	make -C src OBJ_DIR=objects app plugins $jobArgs
```
(I've built with both variants and they produce identical packages.)

Paradoxianer: ok thanks.. because the other way around (withou first building app target) somehow most of the plugins didnt compiled. (i eg was missing Grapheditor... which is the main editor)

haiku-apps/projectconceptor/projectconceptor-0.1.0.recipe
Paradoxianer: Actually i fixed (the build) yesterday .. because since graphviz is only x86 it will be the main version anyway... (graphviz was a planned backend / editorplugin for this ... years and years ago ;-D)

Paradoxianer:
since i am not so firm with haikuports and pullrequest in general .. do i need to change this and submit it new?? or how do i proceed now?

haiku-apps/projectconceptor/projectconceptor-0.1.0.recipe
fbrosson: You lost me :confused:

(FWIW, we have a [recipe for graphviz 2.40.1](https://github.com/haikuports/haikuports/blob/68bd91f476304d1d12989c797eb36dee5d77d209/media-gfx/graphviz/graphviz-2.40.1.recipe) which works on x86 and x86_64.)

Paradoxianer:
That would be sooo nice

fbrosson:
I just noticed that the compilation is done with `-g`, so I'll add a debuginfo package.

Paradoxianer:
Thanks i can disable debug in the makefile.. but since its alpha.. i guess this is the way to go for the moment.
But somehow the build failed.. any idea?
http://kitchen.haiku-os.org/master/x86_gcc2/buildruns/1034/builds/27486.log

Paradoxianer:
I think we need to leave the two make files lines, because i cheated on building i "fake" the App as a lib so the app needs to exist bevore building the plugins, since they are linked against the the app. At least i changed this and now my build isnt broken anymore. - need to fix this when i have time a sort everything out which needs to go in a seperated projectcocneptor lib

fbrosson:
OK, I think I know what is wrong: With the current makefile we cannot use -jN (if N>1).
Maybe we should change the makefile:
```diff
 app: gitupdate pkgupdate app/makefile FORCE
 	cd app && make
 
-plugins: plugins/makefile FORCE
+plugins: app plugins/makefile FORCE
 	cd plugins && make
```

Obviously, an easier workaround is to drop the `%jobArgs` in the recipe. (I'll take care of this in a couple of minnutes.)

The reason why I did not see this is because my VMs have only one virtual CPU.
I'm sorry about this, btw...

Paradoxianer:
Its ok ... i know the makefile solution is a really really bad work around .. thats why i have an issu on this opend... but... as my makefile already produced a really nice working hpkg there was no reason to  target this.. (as other tasks are more important for me)

Paradoxianer:
👍

PR #2452, opened 2018-04-15
ncmpcpp: bump, debuginfo

extrowerk:


media-sound/ncmpcpp/ncmpcpp-0.8.2.recipe
```diff
@@ -106,3 +109,8 @@ INSTALL()
 
 	strip$secondaryArchSuffix $binDir/ncmpcpp
```
fbrosson: You can't have both the debuginfo and the call to strip.
(If we add the debuginfo package then haikuporter will strip the runtime for us.)

media-sound/ncmpcpp/ncmpcpp-0.8.2.recipe
```diff
@@ -82,6 +81,9 @@ USER_SETTINGS_FILES="
 	settings/ncmpcpp/error.log
```
alaviss: I don't really believe the log file is a setting file :smile: 

```diff
-	settings/ncmpcpp/error.log

PR #2459, opened 2018-04-17
Poppler: bump

extrowerk:


app-text/poppler/poppler-0.63.0.recipe
```diff
@@ -0,0 +1,433 @@
...
+ 	1998-2003 Daniel Veillard
+ 	2011 Daiki Ueno
+ 	2005 Dan Sheridan
+ 	2011 Daniel Glöckner
+ 	2013 Daniel Kahn Gillmor
+ 	2011 Daniel Richard G.
+ 	2009 Darren Kenny
+ 	2009-2010 David Benjamin
+ 	2006 David Faure
+ 	1997, Derek B. Noonburg
```
fbrosson:
```diff
- 	1997, Derek B. Noonburg
+ 	1997 Derek B. Noonburg
```
BTW, if what you have right now in `COPYRIGHT` is the result of a grep then you are probably getting names from files such as `INSTALL` that mention authors which sometimes don't have much to do with the project itself.

In other words, if all these names come from source or header files, then that's fine.
If, however, they come from generic readme files or from generated files which might include names of people who maintain the tools used to generate them, then these names should probably be skipped.

app-text/poppler/poppler-0.63.0.recipe
```diff
@@ -0,0 +1,433 @@
...
+	devel:libpoppler$secondaryArchSuffix = $libVersionCompat
+	devel:libpoppler${secondaryArchSuffix}_cpp = $libCppVersionCompat
+	devel:libpoppler${secondaryArchSuffix}_glib = $libGlibVersionCompat
+	devel:libpoppler${secondaryArchSuffix}_qt5 = $libQt5VersionCompat
+	"
+REQUIRES_devel="
+	haiku${secondaryArchSuffix}_devel
+	poppler$secondaryArchSuffix == $portVersion base
+	lib:libpoppler${secondaryArchSuffix}_glib
+	lib:libpoppler${secondaryArchSuffix}_qt5
```
fbrosson:
```
	lib:libpoppler${secondaryArchSuffix}_glib
	lib:libpoppler${secondaryArchSuffix}_qt5
```
I guess these should not be in `REQUIRES_devel` but in `PROVIDES`, otherwise you would not have their `devel:` counterparts in `PROVIDES_devel`.
That said, I might be missing something because I did not test the recipe.

app-text/poppler/poppler-0.63.0.recipe
```diff
@@ -0,0 +1,433 @@
...
+	devel:libfontconfig$secondaryArchSuffix
+	devel:libfreetype$secondaryArchSuffix
+	devel:libgio_2.0$secondaryArchSuffix
+	devel:libglib_2.0$secondaryArchSuffix
+	devel:libgobject_2.0$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libintl$secondaryArchSuffix
+	devel:libjpeg$secondaryArchSuffix
+	devel:liblcms2$secondaryArchSuffix
+	#devel:libnss3$secondaryArchSuffix
```
fbrosson:
```diff
-	#devel:libnss3$secondaryArchSuffix
+#	devel:libnss3$secondaryArchSuffix
```
(That's more eye-friendly.)

app-text/poppler/poppler-0.63.0.recipe
```diff
@@ -0,0 +1,433 @@
...
+		-DCMAKE_BUILD_TYPE=Release \
+		-DENABLE_XPDF_HEADERS=ON
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	prepareInstalledDevelLibs libpoppler \
```
fbrosson: The current style would be perfect for a `defineDebugInfoPackage` or for a `packageEntries`, but for a `prepareInstalledDevelLibs` I'd recommend:
```diff
-	prepareInstalledDevelLibs libpoppler \
+	prepareInstalledDevelLibs \
+		libpoppler \
 		libpoppler-cpp \
 		libpoppler-glib \
 		libpoppler-qt5
```
(Imo this style variant is slitghly more appropriate, but I'm OK with the current style.)

app-text/poppler/poppler-0.63.0.recipe
scottmc: I see no point in including all of the authors and /or contributors.  In the early days of Haikuports we had a rule to crop it at 4 or 5 and perhaps add "et all" or to the end.
Since this is a fork of another project, I would credit the original project, the author of the new project and the maintain of the current  project and then and the others as a single group, including the year ranges.  Assume any year ranges prior to the fork belong to the original project.

fbrosson: :+1:

BTW, I might have added more than 4 or 5 names in some recipes, but the contributor who most likes complete lists is... @alaviss :smile:

alaviss: I never liked `COPYRIGHT`. As far as I know, we and Debian-based distributions are the only ones who specify them. So if we wanted to include those, I would like to make sure that the list is correct, and complete.

With that said, I would not object if we add an guideline to keep the list short, it would greatly reduce the effort of playing with regexes and maintaining the list, which I would greatly appreciate :smile:.

pulkomandy: We should just follow what the upstream says for copyright. If it requires grepping through their files to find out, it's their problem and we should ask them to fix that.
I'm fine with using "1997-2018 Poppler contributors" or something similar, but we should check if the upstream is ok with that wording.

app-text/poppler/poppler-0.63.0.recipe
fbrosson: Sorry, I had not noticed that the style for `prepareInstalledDevelLibs` was already in `poppler-0.62.0.recipe`. So there is no need to change this.

app-text/poppler/poppler-0.63.0.recipe
korli: Actually devel:libpoppler${secondaryArchSuffix}_glib and devel:libpoppler${secondaryArchSuffix}_qt5 are usually symlinks to shared libraries, which need to be installed so that the devel symlinks can work. Hence the presence of lib:libpoppler${secondaryArchSuffix}_glib and lib:libpoppler${secondaryArchSuffix}_qt5 in REQUIRES_devel.

korli:
Please note there are a number of packages requiring the previous version of poppler. Be sure to update them accordingly when this PR is merged.

fbrosson:
Thanks @korli for the info.
(BTW, I did not plan to merge this PR myself because I know it requires quite some work to make sure everything is OK.)

app-text/poppler/poppler-0.63.0.recipe
alaviss: I think they should be

```diff
-	lib:libpoppler${secondaryArchSuffix}_glib
-	lib:libpoppler${secondaryArchSuffix}_qt5
+	poppler_glib$secondaryArchSuffix == $portVersion
+	poppler_qt5$secondaryArchSuffix == $portVersion
```

instead.

app-text/poppler/poppler-0.63.0.recipe
```diff
@@ -0,0 +1,433 @@
...
+	lib:libtiff$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	poppler${secondaryArchSuffix}_devel = $portVersion
+	devel:libpoppler$secondaryArchSuffix = $libVersionCompat
+	devel:libpoppler${secondaryArchSuffix}_cpp = $libCppVersionCompat
+	devel:libpoppler${secondaryArchSuffix}_glib = $libGlibVersionCompat
+	devel:libpoppler${secondaryArchSuffix}_qt5 = $libQt5VersionCompat
```
alaviss: Shall we made this small change?

```diff
-	devel:libpoppler${secondaryArchSuffix}_cpp = $libCppVersionCompat
-	devel:libpoppler${secondaryArchSuffix}_glib = $libGlibVersionCompat
-	devel:libpoppler${secondaryArchSuffix}_qt5 = $libQt5VersionCompat
+	devel:libpoppler_cpp$secondaryArchSuffix = $libCppVersionCompat
+	devel:libpoppler_glib$secondaryArchSuffix = $libGlibVersionCompat
+	devel:libpoppler_qt5$secondaryArchSuffix = $libQt5VersionCompat
```

This is because the library name is actually `libpoppler_{glib,qt5}`, so I believe users would more likely to type 

    pkgman install devel:libpoppler_glib_x86

than

    pkgman install devel:libpoppler_x86_glib

fbrosson: From what I have seen so far, `devel:libpoppler${secondaryArchSuffix}_cpp` is correct, and `devel:libpoppler_cpp${secondaryArchSuffix}` would be wrong.

Some recipes (like some we have for very old releases gcc) did use the second style (i.e. the one you are suggesting), but most recent recipes which could use the old stlyle have been switched to the current style, so I guess the switch was done for some probably good reason.

scottmc:
@extrowerk can you get this one wrapped up soon?

extrowerk:
I have some days taken off, from morning i will have more time to get my open PRs in better shape.

-- Z.

2018.09.25. dátummal, 21:05 időpontban Scott McCreary <notifications@github.com> írta:

alaviss:
A newer version (0.66.0) is already in master

PR #2464, opened 2018-04-19
Libretro port of Nestopia

kwyxz:
This is the libretro port of Nestopia, one of the most accurate NES emulators available


games-emulation/nestopia_libretro/nestopia_libretro-1.49_20180419.recipe
```diff
@@ -0,0 +1,55 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	sed -e "s/@DISPLAY_VERSION@/v${portVersion/_/-}/" \
+		$portDir/additional-files/nestopia_libretro.info.in \
+		> nestopia_libretro.info
+	cd libretro/
```
fbrosson: For a `mv`, having a slash can be nice, e.g. `mv file directory/`.
But for a `cd libretro/` the slash does not bring much added value, so I'd drop it.

That said, if you wish to keep the slash in `cd libretro/` then that's OK for me.

Hmm, instead of dropping the slash, we could drop the whole `cd libretro/` and use `make`'s `-C` option:
```diff
-	cd libretro/
-	make GIT_VERSION=${srcGitRev:0:7} $jobArgs
+	make -C libretro GIT_VERSION=${srcGitRev:0:7} $jobArgs
```
What do you think? (Please feel free to choose any variant or keep the recipe as is.)

fbrosson:
@kwyxz, I had to shrink `SUMMARY` a little bit, so I rephrased it.
I hope it is OK. Please let me know if you would prefer something else. Thanks!

Regarding the `make -C libretro`, I assumed you would be OK, and pushed the change. I can revert it if you wish, of course.

kwyxz:
Yes! It's fine, thank you.

PR #2470, opened 2018-04-23
UDIS86: enable for x86_64

extrowerk:
All tests pass.

### DIFF
--- a/dev-libs/udis86/udis86-1.7.2.recipe
+++ b/dev-libs/udis86/udis86-1.7.2.recipe
@@ -33,6 +33,9 @@ BUILD_PREREQUIRES="
 	cmd:yasm
 	"
 
+defineDebugInfoPackage udis86 \
+	"$binDir"/udcli
+
 BUILD()
 {
 	libtoolize --force --copy --install
--- a/dev-libs/udis86/udis86-1.7.2.recipe
+++ b/dev-libs/udis86/udis86-1.7.2.recipe
@@ -6,11 +6,11 @@ HOMEPAGE="http://udis86.sourceforge.net/"
 COPYRIGHT="2002-2012, Vivek Thampi"
 LICENSE="BSD (2-clause)"
 REVISION="1"
-SOURCE_URI="https://github.com/vmt/udis86/archive/v1.7.2.tar.gz"
+SOURCE_URI="https://github.com/vmt/udis86/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="43567f7e12168943c5b5ffb3d3f5b7a33cb36328f8938a993458f3ded0ba5779"
 PATCHES="udis86-$portVersion.patchset"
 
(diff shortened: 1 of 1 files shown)


dev-libs/udis86/udis86-1.7.2.recipe
```diff
@@ -6,11 +6,11 @@ HOMEPAGE="http://udis86.sourceforge.net/"
 COPYRIGHT="2002-2012, Vivek Thampi"
 LICENSE="BSD (2-clause)"
 REVISION="1"
-SOURCE_URI="https://github.com/vmt/udis86/archive/v1.7.2.tar.gz"
+SOURCE_URI="https://github.com/vmt/udis86/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="43567f7e12168943c5b5ffb3d3f5b7a33cb36328f8938a993458f3ded0ba5779"
```
fbrosson:
```diff
 SOURCE_URI="https://github.com/vmt/udis86/archive/v$portVersion.tar.gz" 
 CHECKSUM_SHA256="43567f7e12168943c5b5ffb3d3f5b7a33cb36328f8938a993458f3ded0ba5779"
+SOURCE_FILENAME="udis86-$portVersion.tar.gz"
 PATCHES="udis86-$portVersion.patchset"
```
(It would be nice to have a `SOURCE_FILENAME`.)

dev-libs/udis86/udis86-1.7.2.recipe
```diff
@@ -43,6 +43,7 @@ BUILD()
 INSTALL()
 {
 	make install
+	rm $libDir/libudis86.la
```
fbrosson: May I suggest a variant for the line you are adding?
```diff
-	rm $libDir/libudis86.la
+	rm -f "$libDir"/libudis86.la
```

dev-libs/udis86/udis86-1.7.2.recipe
```diff
@@ -6,11 +6,12 @@ HOMEPAGE="http://udis86.sourceforge.net/"
 COPYRIGHT="2002-2012, Vivek Thampi"
 LICENSE="BSD (2-clause)"
 REVISION="1"
```
fbrosson:
```diff
-REVISION="1"
+REVISION="2"
```
(I just realised it would be nice to bump `REVISION` in order to really get rid of the libtool file in the x86_gcc2 package.)

fbrosson:
`-g` is used during the compilation, so maybe we should define a debuginfo package:
```diff
 	cmd:yasm
 	"
 
+defineDebugInfoPackage udis86 \
+	"$binDir"/udcli
 
 BUILD()
 {
```

Or maybe this is not a good idea, because there is only one runtime and it's not that big: 384 KiB.
So please feel free to not add the debuginfo package.
(Does someone have an opinion about this question?)

PR #2472, opened 2018-04-24
GCC : enable libstdcxx-filesystem-ts

extrowerk:
It builds fine, but i still have to define the provides, move things and stuff.

### DIFF
--- a/sys-devel/gcc/gcc-5.4.0_2016_06_04.recipe
+++ b/sys-devel/gcc/gcc-5.4.0_2016_06_04.recipe
@@ -124,7 +124,6 @@ BUILD()
 		--with-bug-url=http://dev.haiku-os.org/ \
 		--with-default-libstdcxx-abi=gcc4-compatible \
 		--enable-libssp \
-		--enable-libstdcxx-filesystem-ts \
 		$additionalConfigureFlags
 
 	make $jobArgs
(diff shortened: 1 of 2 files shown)


extrowerk:
It builds fine, but i still have to define the provides, move things and stuff.

sys-devel/gcc/gcc-5.4.0_2016_06_04.recipe
```diff
@@ -124,6 +124,7 @@ BUILD()
 		--with-bug-url=http://dev.haiku-os.org/ \
 		--with-default-libstdcxx-abi=gcc4-compatible \
 		--enable-libssp \
+		--enable-libstdcxx-filesystem-ts \
```
alaviss: Isn't that's the default? From gcc [docs](https://gcc.gnu.org/onlinedocs/libstdc++/manual/configure.html)

> `--enable-libstdcxx-filesystem-ts`[default]
>
>    Build `libstdc++fs.a` as well as the usual libstdc++ and libsupc++ libraries. This is enabled by default on select POSIX targets where it is known to work and disabled otherwise.

extrowerk:
@alaviss: nope, it isn't enabled on Haiku in default.

fbrosson:
@extrowerk, I guess you should _also_ add:
* `lib:libstdc++fs$secondaryArchSuffix = $portVersion compat >= 5` to `PROVIDES_syslibs`
* `devel:libstdc++fs$secondaryArchSuffix = $portVersion compat >= 5` to `PROVIDES_syslibs_devel`

alaviss:
@fbrosson According to the docs the output library is a static one, so I think we only need the `devel:` for gcc

korli:
Better patch gcc to support Haiku: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/acinclude.m4#L4291 
I'm supposing you tested the support on Haiku and that's why you want to enable it in the recipe, am I right?

extrowerk:
@korli : yep, it is actually required for waifu2x. I compiled gcc with my changes, installed and built waifu2x with it. Works nicely (on x86_64 for sure)

I'm informed, it is an experimental thing, but afaik nothing will just ad-hoc depend on it and as the provided lib a static one, it couldn't generate too much problems for us.

extrowerk:
@korli : done.

waddlesplash:
Enabled in GCC 7.3, so closing this.

PR #2489, opened 2018-04-28
MPD: bump

extrowerk:


extrowerk:
The random crash still exist. @mmu_man, any idea?

extrowerk:
@mmuman for reference: https://github.com/MusicPlayerDaemon/MPD/issues/270

PR #2492, opened 2018-04-28
libstaroffice: bump to 0.0.6, add debuginfo.

fbrosson:


app-text/libstaroffice/libstaroffice-0.0.6.recipe
```diff
@@ -2,22 +2,33 @@ SUMMARY="StarOffice import library"
...
-SOURCE_URI="http://github.com/fosnola/libstaroffice/releases/download/$portVersion/libstaroffice-$portVersion.tar.bz2"
-CHECKSUM_SHA256="86f396ae9a25a84cef4f45356132b283863a4657dcaff61eff4fd9f5bf3a2c08"
+SOURCE_URI="https://github.com/fosnola/libstaroffice/releases/download/$portVersion/libstaroffice-$portVersion.tar.bz2"
+CHECKSUM_SHA256="cc0f2d43d856e917c34711fbfde60114920f032713c4ce70bd535e07800f3c7d"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="?x86"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
```
KapiX: Why do you create `commandSuffix`? It works exactly like `secondaryArchSuffix`, unless I'm missing something.

fbrosson: That's part of the [commandBinDir](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) trick to have the runtimes installed in `/system/bin/` instead of `/system/bin/x86/` when we build on x86 secondary arch.

PR #2496, opened 2018-04-28
libabw: bump to 0.1.2, add debuginfo.

fbrosson:


app-text/libabw/libabw-0.1.2.recipe
```diff
@@ -57,25 +68,36 @@ BUILD_PREREQUIRES="
...
+	"$commandBinDir"/abw2text \
+	"$libDir"/libabw-$portVers.so.$libVersion
+
 BUILD()
 {
 	libtoolize --force --copy --install
 	aclocal
 	autoheader
 	automake -a -c --foreign
 	autoconf
```
alaviss: Shall we simplify the recipe with `autoreconf`?

```diff
-	libtoolize --force --copy --install
-	aclocal
-	autoheader
-	automake -a -c --foreign
-	autoconf
+	autoreconf -vfi
```

fbrosson: Yeah, I thought about it, but left it unchanged for a completely unrelated reason:
Git computes a _similarity index_ to decide how it should display the changes.
If it is >= 50% then it shows:
```
#   renamed:    libabw-0.1.1.recipe -> libabw-0.1.2.recipe
```
If it is < 50% then it shows:
```
#   deleted:    libabw-0.1.1.recipe
#   new file:   libabw-0.1.2.recipe
```
With the `commandBinDir` change, I got below 50%, so I had to give up a couple of cosmetic changes here and there, otherwise Git would have considered that the commit was a "deleted"+"new file"  instead of a "renamed" file.

We could do the change you suggested, but I wouldn't bother :)
BTW, we could even choose to not switch to `autoreconf` and see these lines as museum stuff :smile:

PR #2503, opened 2018-04-29
HSTR: new recipe

extrowerk:


app-shells/hstr/hstr-1.24.recipe
```diff
@@ -0,0 +1,72 @@
...
+	cmd:automake
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage hstr$secondaryArchSuffix \
+	"$binDir"/hh$secondaryArchSuffix
```
fbrosson:
```diff
-	"$binDir"/hh$secondaryArchSuffix
+	"$commandBinDir"/hh
```
You should also define the [`commandBinDir`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) variable and use it where appropriate, in particular in the call to `runConfigure`.

app-shells/hstr/patches/hstr-1.24.patchset
```diff
@@ -0,0 +1,86 @@
...
+index 80b18d2..5141663 100644
+--- a/src/hstr.c
++++ b/src/hstr.c
+@@ -22,6 +22,8 @@
+ #include <locale.h>
+ #ifdef __APPLE__
+ #include <curses.h>
++#elif defined(__HAIKU__)
++#include <ncurses.h>
+ #elif defined(__FreeBSD__)
```
alaviss: You could do an `||` with the `defined(__FreeBSD__)` below as well, which IMO, would be cleaner

app-shells/hstr/patches/hstr-1.24.patchset
```diff
@@ -0,0 +1,86 @@
...
++++ b/src/hstr_history.c
+@@ -55,7 +55,11 @@ char *get_history_file_name()
+     if(!historyFile || strlen(historyFile)==0) {
+         char *home = getenv(ENV_VAR_HOME);
+         historyFile = malloc(strlen(home) + 1 + strlen(FILE_DEFAULT_HISTORY) + 1);
+-        strcat(strcat(strcpy(historyFile, home), "/"), FILE_DEFAULT_HISTORY);
++        #ifndef __HAIKU__
++            strcat(strcat(strcpy(historyFile, home), "/"), FILE_DEFAULT_HISTORY);
++        #else
++            strcat(strcat(strcpy(historyFile, home), "/config/settings/"), FILE_DEFAULT_HISTORY);
```
alaviss: Shall we instead use [`find_directory()`][0]?

[0]: https://www.haiku-os.org/docs/api/FindDirectory_8h.html#af95c55d472448ab7af9f32c33aeea71e

app-shells/hstr/patches/hstr-1.24.patchset
fbrosson: Hmm, the diff would be _smaller_, but I wouldn't say that it would be _cleaner_.
Keeping a separate `#elif defined(__HAIKU__)` has one tiny advantage, of course: if the set of headers needed for this purpose ever changes between Haiku and FreeBSD, the "impact" would be isolated.

(If the project already had a few OSes grouped by categories (e.g. BSD*, Linux, etc...) then things would be different.)

PR #2514, opened 2018-04-30
QWinFF: new recipe

extrowerk:


media-video/qwinff/patches/qwinff-0.2.1.patchset
```diff
@@ -0,0 +1,478 @@
+From 6febcd95e5edc992baccddcdc78a56560fbe58d3 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zolt=C3=A1n=20Mizsei?= <zmizsei@extrowerk.com>
+Date: Mon, 30 Apr 2018 11:10:53 +0200
+Subject: Applied patches from rpmfusion
```
alaviss: When applying patches from other distributions that are git commits, please use `git am <path to patch>` to apply them, as the command will correctly preserve ownership and the commit message of the patches.

extrowerk:
@alviss: the rpmfusion patches doesn't apply cleanly, it was the best was i could.

PR #2516, opened 2018-04-30
S25rttr 0.8.2 18

extrowerk:
Still needs love, as it can't find the current executable after start.
The place where the user needs to copy the original datafiles needs to be writeable.


games-strategy/s25rttr/patches/s25rttr-0.8.2~git-source2.patchset
```diff
@@ -0,0 +1,22 @@
+From abd96f51fbb493d4dba37da30d178c906432e064 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zolt=C3=A1n=20Mizsei?= <zmizsei@extrowerk.com>
+Date: Mon, 30 Apr 2018 11:46:48 +0200
+Subject: Haiku is little endian
```
fbrosson: Haiku is little endian on x86{,_gcc2,_86} and arm, but it is big endian on ppc.

OK, we don't have ppc in `ARCHITECTURES` for this recipe, so you can probably just change the comment:
```diff
-Subject: Haiku is little endian
+Subject: Haiku is little endian on x86*
```

PR #2517, opened 2018-04-30
CutyCapt: new recipe

extrowerk:


net-print/cutycapt/cutycapt-1.0~git.recipe
```diff
@@ -0,0 +1,76 @@
...
+BUILD()
+{
+	qmake .
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $commandBinDir
+	cp CutyCapt $commandBinDir
```
fbrosson: Suggestion:
```diff
-	mkdir -p $commandBinDir
+	install -d "$commandBinDir"
-	cp CutyCapt $commandBinDir
+	install -t "$commandBinDir" CutyCapt
```

alaviss:
Since this is a new recipe, should we use the `<version>_<date>` versioning format? For this recipe it would be `1.0_20130715`.

PR #2519, opened 2018-04-30
aobook: enable secondary arch

hanya:


app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,47 +4,58 @@ of Japanese-language literature."
...
+	cmd:sed
 	"
 
 USER_SETTINGS_FILES="
 	settings/aobook directory
 	"
 
+PATCH()
+{
+if [ "$effectiveTargetArchitecture" = x86 ]; then
```
fbrosson: I guess `PATCH`
1) is only needed on x86 secondary arch
2) changes nothing on x86 primary arch

So, if I'm not mistaken, you can:
```diff
-if [ "$effectiveTargetArchitecture" = x86 ]; then
+if [ -n "$secondaryArchSuffix" -a "$effectiveTargetArchitecture" = x86 ]; then
 	FreeTypeInc="`findpaths -e -a x86 B_FIND_PATH_HEADERS_DIRECTORY freetype2`"
 	sed -i 's,$(shell findpaths -e B_FIND_PATH_HEADERS_DIRECTORY freetype2)',$FreeTypeInc, \
 		mlib/Makefile.haiku src/Makefile.haiku
 fi
```

Alternatively, if I'm not mistaken, you could:
```diff
-if [ "$effectiveTargetArchitecture" = x86 ]; then
+if [ -n "$secondaryArchSuffix" ]; then
-	FreeTypeInc="`findpaths -e -a x86 B_FIND_PATH_HEADERS_DIRECTORY freetype2`"
+	FreeTypeInc="`findpaths -e -a "$effectiveTargetArchitecture" \
+		B_FIND_PATH_HEADERS_DIRECTORY freetype2`"
 	sed -i 's,$(shell findpaths -e B_FIND_PATH_HEADERS_DIRECTORY freetype2)',$FreeTypeInc, \
 		mlib/Makefile.haiku src/Makefile.haiku
 fi
```
Note: I did not try any of my suggestions.

hanya: $effectiveTargetArchitecture value could be passed to the findpaths for both primary and secondary architecture.
It looks I can remove architecture check to make the recipe simple.

app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,47 +4,57 @@ of Japanese-language literature."
...
-	lib:libiconv
-	lib:libjpeg
-	lib:libpng
-	lib:libz
+	haiku$secondaryArchSuffix
+	lib:libfontconfig$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
```
fbrosson:
```diff
-	lib:libpng$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
```
(This is prefered.)

app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,47 +4,57 @@ of Japanese-language literature."
...
-	devel:libiconv
-	devel:libjpeg
-	devel:libpng
-	devel:libz
+	haiku${secondaryArchSuffix}_devel
+	devel:libfontconfig$secondaryArchSuffix
+	devel:libfreetype$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libjpeg$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
```
fbrosson:
```diff
-	devel:libpng$secondaryArchSuffix
+	devel:libpng16$secondaryArchSuffix
```
(This is prefered.)

app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,47 +4,57 @@ of Japanese-language literature."
...
 
 USER_SETTINGS_FILES="
 	settings/aobook directory
 	"
 
+PATCH()
+{
+	FreeTypeInc="`findpaths -e -a "$effectiveTargetArchitecture" \
+		B_FIND_PATH_HEADERS_DIRECTORY freetype2`"
+	sed -i 's,$(shell findpaths -e B_FIND_PATH_HEADERS_DIRECTORY freetype2)',$FreeTypeInc, \
```
fbrosson: What do you think of this variant:
```diff
 PATCH()
 {
-	FreeTypeInc="`findpaths -e -a "$effectiveTargetArchitecture" \
-		B_FIND_PATH_HEADERS_DIRECTORY freetype2`"
+if [ -n "$secondaryArchSuffix" ]; then
-	sed -i 's,$(shell findpaths -e B_FIND_PATH_HEADERS_DIRECTORY freetype2)',$FreeTypeInc, \
-		mlib/Makefile.haiku src/Makefile.haiku
+	sed -i -e "s,\(shell findpaths -e \)\(B_FIND_PATH_HEADERS_DIRECTORY freetype2)\),\1-a \$(ARCH) \2," \
 		mlib/Makefile.haiku src/Makefile.haiku
+fi
 }
```
BTW, since haikuporter calls `setarch`, you could probably use `$(shell getarch)` in the 2 `Makefile.haiku` files and use the result of that command and pass it to 
`shell findpaths -e -a $(ARCH) B_FIND_PATH_HEADERS_DIRECTORY freetype2`

Hmm, you could even use this variant:
```
PATCH()
{
	sed -i -e "s,\(shell findpaths -e \)\(B_FIND_PATH_HEADERS_DIRECTORY freetype2)\),\1-a \$(ARCH) \2," \
 		mlib/Makefile.haiku src/Makefile.haiku
}
```
and append ` ARCH=$effectiveTargetArchitecture` at the end of each call to `make`.

app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,54 +4,62 @@ of Japanese-language literature."
...
+		mlib/Makefile.haiku src/Makefile.haiku
+}
+
 BUILD()
 {
 	cd mlib
-	make -f Makefile.haiku $jobArgs OBJ_DIR=objects
+	make -f Makefile.haiku $jobArgs OBJ_DIR=objects ARCH=$effectiveTargetArchitecture
 
 	cd ../src
```
fbrosson: The recipe builds fine and everything seems to work fine (although I did not load any file with the app.)
So I we can merge your PR.

But before doing so, I'd like to mention something a noticed when I tried a variant for `BUILD()`:
```bash
BUILD()
{
	make -C mlib -f Makefile.haiku $jobArgs OBJ_DIR=objects \
		ARCH=$effectiveTargetArchitecture
	make -C src -f Makefile.haiku $jobArgs OBJ_DIR=objects \
		ARCH=$effectiveTargetArchitecture default catalogs bindcatalogs
}
```
What I noticed is that with this variant the app as bigger than when I build it with the recipe in this PR.
When I build aobook on x86 secondary arch with my variant, /system/apps/Aobook is around 526 KiB, instead of 490 KiB.
It's a bit puzzling, but I have to admit that I did not investigate a lot :)

Anyway, I can merge your PR as soon as you wish.

hanya: I have to pass the architecture when catalogs is build to make freetype header is found. 
The difference of the binary file is the fail of the binding the catalog file generated by the catalogs.

fbrosson: Cool, thanks for the investigation!
So, this means my last variant was a good idea :)

BTW, I've just tested one more variant (based on this one) which seems to be working fine.
You can drop the 2 ` ARCH=$effectiveTargetArchitecture` in `BUILD` and the whole `PATCH` function if you apply these small changes upstream, in `mlib/Makefile.haiku` and `src/Makefile.haiku`:
```diff
 LIBPATHS = 
 
+ARCH = $(shell getarch)
+
 #	Additional paths to look for system headers. These use the form
 #	"#include <header>". Directories that contain the files in SRCS are
 #	NOT auto-included here.
 SYSTEM_INCLUDE_PATHS = \
-	$(shell findpaths -e B_FIND_PATH_HEADERS_DIRECTORY freetype2) \
+	$(shell findpaths -e -a $(ARCH) B_FIND_PATH_HEADERS_DIRECTORY freetype2) \
```

Shall we merge the recipe as is, or would you prefer to implement this "last minute" variant before the merge?
(It's perfectly OK to merge this PR as is and postpone this last minute suggestion to a future release. Both scenari are perfectly OK.)

hanya:
I've added the architecture for findpaths command in the upstream. 
And also I tried to support Noto CJK font, default font for CJK people on Haiku OS.

app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,41 +4,43 @@ of Japanese-language literature."
 HOMEPAGE="https://osdn.net/projects/aobook/"
 COPYRIGHT="2014-2017 Azel"
```
fbrosson:
```diff
-COPYRIGHT="2014-2017 Azel"
+COPYRIGHT="2014-2018 Azel"
```

app-text/aobook/aobook-1.0.3.recipe
```diff
@@ -4,41 +4,43 @@ of Japanese-language literature."
...
 LICENSE="BSD (3-clause)"
-REVISION="2"
-localRevision="3"
+REVISION="3"
+localRevision="4"
 SOURCE_URI="https://github.com/hanya/aobook-haiku/archive/v$portVersion-$localRevision.tar.gz"
```
fbrosson: I've been thinking about something we could do: rename the recipe to `aobook-1.0.3_4.recipe`.
It would allow to "insert" the `localRevision` inside the `portVersion`. This would require the following cosmetic changes:
```diff
-REVISION="3"
+REVISION="1"
-localRevision="4"
-SOURCE_URI="https://github.com/hanya/aobook-haiku/archive/v$portVersion-$localRevision.tar.gz"
+REVISION="1"
+SOURCE_URI="https://github.com/hanya/aobook-haiku/archive/v${portVersion/_/-}.tar.gz"
 CHECKSUM_SHA256="f840a98b0f0c1f3fa31ae19498bc9051106e068b9af95eb4065643693f7f1e07"
+SOURCE_FILENAME="aobook-haiku-${portVersion/_/-}.tar.gz"
-SOURCE_DIR="aobook-haiku-$portVersion-$localRevision"
+SOURCE_DIR="aobook-haiku-${portVersion/_/-}"
```

I think this could be a good idea. What do you think?

(Well, the `SOURCE_FILENAME` is not a requirement for changing the name of the recipe but a general recommendation.)

In case you prefer to keep the internalRevision, that's also OK.
(In that case, could you please insert a `SOURCE_FILENAME="aobook-haiku-$portVersion-$localRevision.tar.gz"` between `CHECKSUM_SHA256` and `SOURCE_DIR` ?)

PR #2524, opened 2018-05-01
QtChan: new recipe

extrowerk:


www-aps/qtchan/qtchan-0.95.recipe
```diff
@@ -0,0 +1,87 @@
...
+
+INSTALL()
+{
+	cd build
+	mkdir -p "$appsDir"
+	install -t "$appsDir" qtchan
+
+	mv "$appsDir"/qtchan "$appsDir"/QtChan
+
+# TODO: needs an icon
```
fbrosson: You could add a tiny `additional-files/qtchan.rdef.in` with something like
```

resource app_signature "application/x-vnd.qtchan";
resource app_flags B_MULTIPLE_LAUNCH;

resource app_version {
	major = @MAJOR@,
	middle = @MIDDLE@,
	minor = @MINOR@,
	variety = B_APPV_FINAL,
	internal = 0,
	short_info = "qtchan",
	long_info = "@LONG_INFO@"
};
```
and uncomment this whole section, with one extra line:
```
		-e "s|@LONG_INFO@|$SUMMARY|" \
```
This would make it slightly easier for another contributor to add an icon.
What do you think?

www-aps/qtchan/patches/qtchan-0.95.patchset
```diff
@@ -0,0 +1,60 @@
...
++++ b/mainwindow.cpp
+@@ -63,7 +63,11 @@ MainWindow::MainWindow(QWidget *parent) :
+ 	connect(&settingsView,&Settings::update,[=](QString field, QVariant value){
+ 		if(field == "use4chanPass" && value.toBool() == true){
+ 			QSettings settings(QSettings::IniFormat,QSettings::UserScope,"qtchan","qtchan");
+-			QString defaultCookies = QDir::homePath() + "/.config/qtchan/cookies";
++			#ifndef __HAIKU__
++				QString defaultCookies = QDir::homePath() + "/.config/qtchan/cookies";
++			#else
++				QString defaultCookies = QDir::homePath() + "/config/settings/qtchan/cookies";
```
alaviss: Shouldn't [`find_directory()`][0] be used instead of hardcoding paths? Also, for cookies, I think `B_USER_CACHE_DIRECTORY` or `B_USER_VAR_DIRECTORY` would be a better place to put them into

[0]: https://www.haiku-os.org/docs/api/FindDirectory_8h.html#af95c55d472448ab7af9f32c33aeea71e

www-aps/qtchan/qtchan-0.95.recipe
```diff
@@ -0,0 +1,87 @@
...
+* open files in OS's default application
+* search and filter."
+HOMEPAGE="https://github.com/siavash119/qtchan"
+COPYRIGHT="2017 siavash119"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/siavash119/qtchan/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="b86fc331d9d91ad5d1fd15ac404e79e65084d1db58598861f95aa76917170721"
+SOURCE_FILENAME="qtchan-$portVersion.tar.gz"
+SOURCE_DIR="qtchan-$portVersion"
```
alaviss: It's unnecessary since that's the default already

```diff
-SOURCE_DIR="qtchan-$portVersion"
```

www-aps/qtchan/qtchan-0.95.recipe
```diff
@@ -0,0 +1,87 @@
...
+PATCHES="qtchan-$portVersion.patchset"
+#ADDITIONAL_FILES="qtchan.rdef.in"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qtchan$secondaryArchSuffix = $portVersion
+	app:OtterBrowser$secondaryArchSuffix = $portVersion
```
alaviss: Otter?

www-aps/qtchan/qtchan-0.95.recipe
```diff
@@ -0,0 +1,87 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	mkdir -p "$appsDir"
+	install -t "$appsDir" qtchan
+
+	mv "$appsDir"/qtchan "$appsDir"/QtChan
```
alaviss: Shall we use this shorter variant?

```diff
-	mkdir -p "$appsDir"
-	install -t "$appsDir" qtchan
-
-	mv "$appsDir"/qtchan "$appsDir"/QtChan
+	install -D -m 755 qtchan "$appsDir/QtChan"
```

alaviss:
Also, shouldn't the folder be `www-apps` instead of `www-aps`?

www-aps/qtchan/patches/qtchan-0.95.patchset
fbrosson: Yep, it would be nice to use `find_directory()`.

Regarding the cookies, I'd say that it is OK (and even better imo) to keep them in `~/config/settings/qtchan/cookies/`.
Keeping them in `B_USER_CACHE_DIRECTORY` or `B_USER_VAR_DIRECTORY` would mean that other programs would possibly notice them, while having them in a dedicated directory (like the one we have right now in the recipe) is a little bit safer.

alaviss: Oh, I meant `B_USER_CACHE_DIRECTORY/qtchan/cookies` instead.

fbrosson: Well, even if it's in a subdirectory of `B_USER_CACHE_DIRECTORY` I would still think that the cookies would be better in `~/config/settings/qtchan/cookies/`.
My point of view is that cookies are not really cache data. They are user data created by an application.
Cached data, on the other hand, can be restored from an external source. Cookies can only be "recreated" (but not always restored to what the user had previously), provided the user submits the appropriate inputs, e.g. credentials.

Now, if more contributors think `B_USER_CACHE_DIRECTORY` would be better than `~/config/settings/qtchan/cookies/` then I'm OK with your suggestion :)

alaviss: I've also looked at WebPositive [source][0] and found that cookies are stored in `B_USER_SETTINGS_DIRECTORY`, so I agree that's where we should store the cookies.

[0]Y http://xref.plausible.coop/source/xref/haiku/src/apps/webpositive/BrowserApp.cpp#90

www-apps/qtchan/patches/qtchan-0.95.patchset
```diff
@@ -0,0 +1,60 @@
...
+@@ -63,7 +63,11 @@ MainWindow::MainWindow(QWidget *parent) :
+ 	connect(&settingsView,&Settings::update,[=](QString field, QVariant value){
+ 		if(field == "use4chanPass" && value.toBool() == true){
+ 			QSettings settings(QSettings::IniFormat,QSettings::UserScope,"qtchan","qtchan");
+-			QString defaultCookies = QDir::homePath() + "/.config/qtchan/cookies";
++			#ifndef __HAIKU__
++				QString defaultCookies = QDir::homePath() + "/.config/qtchan/cookies";
++			#else
++				QString defaultCookies = QDir::homePath() + "/config/settings/qtchan/cookies";
++			#endif
```
alaviss: Since QtChan is an Qt5 program, you could change the patch to make use Qt's [`QStandardPaths::AppConfigLocation`](https://doc.qt.io/qt-5/qstandardpaths.html), which would returns the location of the configuration folder (it uses `find_directory()` internally for Haiku), with an upside of being easily upstreamable (it's cross-platform, and IMO that's how it should be done).

PR #2525, opened 2018-05-02
util-linux: bump to 2.32, build lib{blkid,fdisk,uuid}.

fbrosson:
~Also build blkid, hexdump and uuidgen.~

Also build blkid, fdisk, hexdump and uuidgen, but put fdisk in a subpackage, marked as untested, as the geometry autodetection is not working, so the default values are not useful at all.
fdisk is working fine but it does not know how to inform Haiku that a partition table was modified. Workaround: reboot.
TODO: teach fdisk and partprobe how to talk to Haiku's kernel.

### DIFF
--- a/sys-apps/util-linux/patches/util_linux-2.32.patchset
+++ b/sys-apps/util-linux/patches/util_linux-2.32.patchset
@@ -131,17 +131,18 @@ From: fbrosson <fbrosson@localhost>
 Date: Tue, 1 May 2018 23:37:15 +0000
 Subject: define {makedev,major,minor} macros in blkidP.h for libblkid
 
+
 diff --git a/libblkid/src/blkidP.h b/libblkid/src/blkidP.h
-index 22c9856..66130a3 100644
+index 22c9856..1a08190 100644
 --- a/libblkid/src/blkidP.h
 +++ b/libblkid/src/blkidP.h
 @@ -341,6 +341,10 @@ UL_DEBUG_DECLARE_MASK(libblkid);
  #define UL_DEBUG_CURRENT_MASK	UL_DEBUG_MASK(libblkid)
  #include "debugobj.h"
  
-+#define makedev(ma, mi) ((dev_t)((ma) << 8) | (mi))
-+#define major(devno) ((devno) >> 8)
-+#define minor(devno) ((devno) & 0xff)
++#define makedev(ma, mi) ((dev_t)(((((unsigned int)(ma)) << 8) & 0xff00) | (mi)))
++#define major(devnum) (((unsigned int)((devnum) & 0xff00) >> 8))
++#define minor(devnum) (unsigned int)((devnum) & 0xffff00ff)
 +
  extern void blkid_debug_dump_dev(blkid_dev dev);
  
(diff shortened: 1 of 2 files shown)


sys-apps/util-linux/patches/util_linux-2.32.patchset
```diff
@@ -130,3 +125,43 @@ index e1c4059..ac553bf 100644
...
+index 22c9856..719bd83 100644
+--- a/libblkid/src/blkidP.h
++++ b/libblkid/src/blkidP.h
+@@ -341,6 +341,10 @@ UL_DEBUG_DECLARE_MASK(libblkid);
+ #define UL_DEBUG_CURRENT_MASK	UL_DEBUG_MASK(libblkid)
+ #include "debugobj.h"
+ 
++extern unsigned int major(dev_t devno);
++extern unsigned int minor(dev_t devno);
++extern dev_t makedev(unsigned int ma, unsigned int mi);
```
fbrosson:
```diff
++extern unsigned int major(dev_t devno);
++extern unsigned int minor(dev_t devno);
++extern dev_t makedev(unsigned int ma, unsigned int mi)
```
Would it be acceptable to move these declarations (or variants of it) to the system headers?

sys-apps/util-linux/patches/util_linux-2.32.patchset
```diff
@@ -130,3 +125,43 @@ index e1c4059..ac553bf 100644
...
+index 056df2b..3cbe939 100644
+--- a/libblkid/src/dev.c
++++ b/libblkid/src/dev.c
+@@ -15,6 +15,10 @@
+ 
+ #include "blkidP.h"
+ 
++unsigned int major(dev_t devno) { return devno>>8; }
++unsigned int minor(dev_t devno) { return devno & 0xff; }
++dev_t makedev(unsigned int ma, unsigned int mi) { return (dev_t)(ma<<8) | mi; }
```
fbrosson:
```diff
++unsigned int major(dev_t devno) { return devno>>8; }
++unsigned int minor(dev_t devno) { return devno & 0xff; }
++dev_t makedev(unsigned int ma, unsigned int mi) { return (dev_t)(ma<<8) | mi; }
```
Would it be acceptable to move these functions (or variants of it) to one of the system shared libraries or to libbsd, libgnu or anything more appropriate ?

Would someone be willing to take care of it?
@alaviss, since you are an expert, what would you recommend? Thanks!

alaviss:
Is `blkid` working?

And I don't think enabling components that rely on the Linux kernel is a good idea. Tools such as `hexdump` are nice, but `libfdisk` and `libblkid` requires the Linux kernel and it would be a better idea if we make our own version instead.

fbrosson:
Yep, blkid is working :)
I don't know if libfdisk is usable, but one way to check this is to add it :)

sys-apps/util-linux/patches/util_linux-2.32.patchset
alaviss: A quick look shows that those are common POSIX extensions that exists in the BSDs and Linux, so I would recommend placing them in `libroot`, implemented as macros. In fact, they are already in [Haiku's ntfs driver][0], so we just need to move them out.

[0]: http://xref.plausible.coop/source/xref/haiku/src/add-ons/kernel/file_systems/ntfs/libntfs/config.h#312

alaviss:
Please note that `blkid` is also [provided by][0] e2fsprogs, so we would have to decide which `blkid` to use.

> I don't know if libfdisk is usable, but one way to check this is to add it :)

Add the `fdisk` tool as well, then we can test `libfdisk` ;)

[0]: https://github.com/haikuports/haikuports/blob/a2467a63876cb0c3ab4bb2e28d4ddd60017a5ee7/sys-fs/e2fsprogs/e2fsprogs-1.43.9.recipe#L14

sys-apps/util-linux/patches/util_linux-2.32.patchset
fbrosson: Cool, thanks! I'll try with Haiku's ntfs driver.

fbrosson:
Yep, I had already prepared a recipe for e2fsprogs 1.44.1 with some changes: #2526

Regarding fdisk and libdisk, I already tried to build fdisk in util-linux but I think I had to disable it for some reason. I'll check this again.

fbrosson:
I have changed the patchset to use macros (instead of fake functions), but if anyone would like to add `makedev`, `major` and `minor` to libroot then I will happily update this PR to make use of it :)

alaviss:
> But if anyone would like to add `makedev`, `major` and `minor` to libroot

Do it @fbrosson

fbrosson:
@alaviss

> Do it @fbrosson

I created a patch for it: https://review.haiku-os.org/c/haiku/+/271
But I think it won't be merged. So this means this PR can be merged as is :smile:

kallisti5:
This transitions any consumers of libuuid over to the util_linux, meaning we now have to ship util_linux in our images and build-packages.

Since the tools included with util_linux are "non-functional" stubs... this is going to lead to confusion.

PR #2546, opened 2018-05-10
C-Blosc : bump

extrowerk:


dev-libs/c_blosc/patches/c_blosc-1.14.3.patchset
```diff
@@ -0,0 +1,25 @@
+From d014d1b144f867dbfc158eea1f45d26f9a35bf58 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Zolt=C3=A1n=20Mizsei?= <zmizsei@extrowerk.com>
+Date: Thu, 7 Dec 2017 21:27:35 +0100
+Subject: [PATCH] Build fix
+
+---
+ tests/test_common.h | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
```
fbrosson: Suggestion:
```diff
-Subject: [PATCH] Build fix
+Subject: Build fix
 
----
- tests/test_common.h | 2 +-
- 1 file changed, 1 insertion(+), 1 deletion(-)
```
I guess you are using `git format-patch` to create the patchset.

The suggestion above is the result of `haikuporter -e c_blosc` (`-e` is the short option equivalent to `--extract-patchset`).
Since most (if not all) contributors use haikuporter's `-e` option (which we call after a `-B` with a copy of the patchset from the previous release), it would be nice if you also used it.
(I guess the recommended method to create patchsets used in recipes is to use haikuporter.)

If you would like to try, the process is the same, except that, in the end, instead of calling `git format-patch` you call `haikuporter -e c_blosc`.

dev-libs/c_blosc/c_blosc-1.14.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+
+	# move include dir to correct location
+	mkdir -p $(dirname $includeDir)
+	mv $prefix/include $includeDir
+
+	if [ -n "$secondaryArchSuffix" ]; then
+		# On secondary arch libs should go to $libDir
+		# instead of $prefix/lib
+		mkdir -p $libDir
+		mv $prefix/lib/* $libDir
```
fbrosson: This does not work on x86 secondary arch because (on that arch) `$libDir` is the same as `$prefix/lib/x86`, so this command would move x86 into itself.
I found a dirty workaround: Pass
```
		-DLIB_SUFFIX=${secondaryArchSuffix:+/$effectiveTargetArchitecture} \
```
to cmake and only move what is left:
```diff
-		# On secondary arch libs should go to $libDir
-		# instead of $prefix/lib
+		# On secondary arch the pkgconfig subdir should also go to $libDir
+		# instead of $prefix/lib
-		mkdir -p $libDir
-		mv $prefix/lib/* $libDir
+		mv "$prefix"/lib/pkgconfig "$libDir"
```

I did find how to tell cmake that it should also install the pkgconfig file in the right directory.
(I'll leave that to you, if you want to investigate further ;)

dev-libs/c_blosc/c_blosc-1.14.3.recipe
```diff
@@ -0,0 +1,109 @@
...
+{
+	mkdir -p haiku_build
+	cd haiku_build
+
+	cmake .. \
+		-DPREFER_EXTERNAL_LZ4=ON \
+		-DPREFER_EXTERNAL_SNAPPY=ON \
+		-DPREFER_EXTERNAL_ZLIB=ON \
+		-DPREFER_EXTERNAL_ZSTD=ON \
+		$cmakeDirArgs
```
fbrosson:
```diff
 	cmake .. \
 		-DPREFER_EXTERNAL_LZ4=ON \
 		-DPREFER_EXTERNAL_SNAPPY=ON \
 		-DPREFER_EXTERNAL_ZLIB=ON \
 		-DPREFER_EXTERNAL_ZSTD=ON \
+		-DLIB_SUFFIX=$secondaryArchSubDir \
 		$cmakeDirArgs
```
(That's even better than what I suggested earlier, but I still need to check that it works as expected.)

fbrosson:
Shall we also drop `c_blosc-1.12.1.recipe` (and its `c_blosc-1.12.1.patchset`) and keep `c_blosc-1.13.7.recipe` ?
(Having `1.13.7` and `1.14.3` is probably enought. We probably don't need to keep `1.12.1`.)
(BTW, you _don't_ need to drop `1.13.7`. Dropping _only_ `1.12.1` is enough imo.)

PR #2547, opened 2018-05-10
Libgit2 0.27.0

extrowerk:


dev-libs/libgit2/libgit2-0.27.0.recipe
```diff
@@ -0,0 +1,104 @@
...
+
+defineDebugInfoPackage libgit2$secondaryArchSuffix \
+	"$libDir"/libgit2.so.$libVersion
+
+BUILD()
+{
+	mkdir -p build && cd build
+
+	OPTIONS=
+	if [ $effectiveTargetArchitecture == x86_gcc2 ] ; then
```
fbrosson: Suggestion:
```diff
-	if [ $effectiveTargetArchitecture == x86_gcc2 ] ; then
+	if [ $effectiveTargetArchitecture = x86_gcc2 ] ; then
```
(Yep, both work, but since `==` is a bash-only syntax, while `=` works with both `bash` and `sh`, I would recommend using the most portable syntax.)

Issue #2543, opened 2018-05-08
Apache: Update to latest version


kenmays:

kenmays:
Can close this when updated package approved.

PR #2549, opened 2018-05-11
Apache: bump, provides fix, debuginfo, bindir, general cleanup

extrowerk:
Closes https://github.com/haikuports/haikuports/issues/2543


www-servers/apache/apache-2.4.33.recipe
```diff
@@ -0,0 +1,309 @@
...
+	"
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:pcre_config
```
fbrosson:
```diff
-	cmd:pcre_config
+	cmd:pcre_config$secondaryArchSuffix
```
BTW, shall we drop `2.4.29`? Nobody will use this once `2.4.33` is merged.
Moreover, dropping the old recipe allows to have a much smaller `git log` :)

www-servers/apache/apache-2.4.33.recipe
```diff
@@ -0,0 +1,309 @@
...
+#	devel:libldap$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize
```
fbrosson: Oops, I just noticed I missed this one :)
```diff
-	cmd:libtoolize
+	cmd:libtoolize$secondaryArchSuffix
```

BTW, regarding the previous recipe (`2.4.29`), don't you want to remove it?
If `2.4.33` works, then we don't need `2.4.29`.
And if `2.4.33` does not work, then it would be better to fix it before merging.

In other words, it would be nice to see if this release is working fine, even with a minimalistic config.
(If it looks OK then we can drop the previous recipe. If not, then we can't merge this as is.)

PR #2553, opened 2018-05-12
gifsicle: bump to 1.91, strip runtimes.

fbrosson:


media-gfx/gifsicle/gifsicle-1.91.recipe
```diff
@@ -41,12 +42,17 @@ PATCH()
...
+		# from function scale_image_threaded in src/xform.c
+		sed -i -e "/HAVE___SYNC_ADD_AND_FETCH/ s/1/0/" config.h
+	fi
 	make
 }
 
 INSTALL()
 {
-	make install
+	make install-strip
```
alaviss: Shouldn't we use `defineDebugInfoPackage` instead of dropping debuginfo by stripping?

fbrosson: Yep, I've been thinking about this, but I did not dare to add a debuginfo package because the package does not have any library.
But I can think of a good reason to add it: it makes it (slightly) easier to for a user to investigate possible bugs, as the user would just need to install the debuginfo package instead of having to rebuild everything.

So I'll add the debuginfo package right now ;)

alaviss: Tagging #2135, as we still don't have an official consensus on what should be done with debug info in binaries

pulkomandy:
On Sun, May 13, 2018 at 06:52:56AM -0700, fbrosson wrote:
> Yep, I've been thinking about this, but I did not dare to add a debuginfo package because the package does not have any library.
> But I can think of a good reason to add it: it makes it (slightly) easier to for a user to investigate possible bugs, as the user would just need to install the debuginfo package instead of having to rebuild everything.

This is useless. If someone needs to debug an executable, the goal is to
fix the bug so sooner or later building from source will be needed.

Debuginfo for libraries is convenient because you use it not to debug
the library, but to debug software using the library.

Please do not clutter the repository with useless debuginfo packages.

fbrosson:
Before I revert to `install-strip`, I have one question:
What if the user who suspects a possible bug is not an expert and does not wish (or have the rights) to send the input file?
If he is not an expert he might just give up rebuilding the package. He might even choose to not report anything.
Making the debuginfo package available through HaikuDepot/pkgman gives a friendly signal: _"Haiku provides debuginfo packages so that users can easily report bugs without having to rebuild anything."_

So, shall I proceed and revert to `install-strip`, or do we keep the debuginfo package?

alaviss:
@pulkomandy

> This is useless. If someone needs to debug an executable, the goal is to fix the bug so sooner or later building from source will be needed.

It's not. For end-users, with the debuginfo, the Debug Report generated would be more detailed, which would greatly assists the developers to pinpoint and identify the exact cause of the problem.

@fbrosson
> So, shall I proceed and revert to `install-strip`, or do we keep the debuginfo package?

IMO, I would like to keep the debuginfo package. But we should reach an consensus on this first.

pulkomandy:
On Sun, May 13, 2018 at 07:15:42AM -0700, fbrosson wrote:
> Before I revert to `install-strip`, I have one question:
> What if the user who suspects a possible bug is not an expert and does not wish (or have the rights) to send the input file?
> If he is not an expert he might just give up rebuilding the package. He might even choose to not report anything.
> Making the debuginfo package available through HaikuDepot/pkgman gives a friendly signal: _"Haiku provides debuginfo packages so that users can easily report bugs without having to rebuild anything."_
> 
> So, shall I proceed and revert to `install-strip`, or do we keep the debuginfo package?

If there is a bug in the program with specific input files, the user
will have to report it to upstream, not us. And the first thing they'll
ask is "try with the latest code in our git/snv/whatever repo, because
surely they don't want to be debugging something that has already been
fixed.

Moreover, an unexperimented user will not know what a debuginfo package
is, and that it should be installed to generate a better debug report
(and they should not need to know).

Debuginfo packages are useful in one single, specific case. As I
mentioned above: when a developer debugs his own program and needs to
look at what's happening inside a library. And we should provide
debuginfo packages only when they are fulfilling this need.

fbrosson:
I agree that it is very unlikely that an unexperienced user would hit a bug while using gifsicle on Haiku. But if this ever happens, then:
a) if we don't have a debuginfo then we will probably never hear about it.
b) if we have a debuginfo then if the user chooses to report the bug, even if he has never used a debuginfo package, the instructions would be quite simple, so the user would probably be happy to send a detailed bug report.

@alaviss
> IMO, I would like to keep the debuginfo package. But we should reach an consensus on this first.

:+1: But I don't think we have convinced @pulkomandy ;)
So I guess I'll drop the debuginfo.

(@pulkomandy, in case you change your mind please let us know.)

fbrosson:
@pulkomandy
That's it. I've dropped the second commit, so we don't have the debuginfo package anymore ;)

pulkomandy:
On Sun, May 13, 2018 at 08:35:25AM -0700, fbrosson wrote:
> I agree that it is very unlikely that an unexperienced user would hit a bug while using gifsicle on Haiku. But if this ever happens, then:
> a) if we don't have a debuginfo then we will probably never hear about it.
> b) if we have a debuginfo then if the user chooses to report the bug, even if he has never used a debuginfo package, the instructions would be quite simple, so the user would probably be happy to send a detailed bug report.

I don't understand. The debuginfo package does not make it any easier to
submit a bugreport. It only makes the bugreport possibly easier to
interpret, but as a Haikuport contributor I don't care. What I will do
regardless is:

1) Ask the user to reproduce this on another platform. If it crashes
there too, tell the user to report to upstream

2) Otherwise, look carefully at our patches to see where things may have
gone wrong. Usually I can guess without a debuginfo package (we still
get a backtrace), or otherwise ask the user to send me their image in
this case, or I can send them a debug build in case they can't send me
the image.

So, I don't see how the debuginfo package makes any difference. The user
can still submit a bugreport and attach a debugger log, even if there is
no debuginfo package. And as I said, they probably don't know what
debuginfo packages are and don't have them installed, so the first
debugreport we'll get will be without the debuginfo package.


I keep my advice: debuginfo packages for binaries are cluttering the
repo, and do not add any value, or not enough to justify listing them in
the available package list for all users. Please don't do it.

fbrosson:
OK, that's fine for me. Let's merge!

@alaviss, yep, the HaikuPorts community probably needs to discuss about consensus building :smile:

PR #2555, opened 2018-05-13
LevelDB: new recipe

extrowerk:


extrowerk:
All tests are passing.

dev-libs/leveldb/leveldb-1.20.recipe
```diff
@@ -0,0 +1,80 @@
...
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage leveldb$secondaryArchSuffix \
+	"$libDir"/libleveldb.so.1
```
fbrosson: Suggestion:
```diff
-	"$libDir"/libleveldb.so.1
+	"$libDir"/libleveldb.so.$libVersion
```

This neutral change will allow to detect if the name  of the shared lib differs from what we currently expect.
Yep, this is probably never going to happen, but in case it does, haikuporter will complain and we will then notice the change.

dev-libs/leveldb/leveldb-1.20.recipe
```diff
@@ -0,0 +1,80 @@
...
+not have a relational data model and it does not support SQL queries. Also, \
+it has no support for indexes. Applications use LevelDB as a library, as it \
+does not provide a server or command-line interface."
+HOMEPAGE="http://leveldb.org/"
+COPYRIGHT="2011 The LevelDB Authors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/google/leveldb/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="f5abe8b5b209c2f36560b75f32ce61412f39a2922f7045ae764a2c23335b6664"
+PATCHES="leveldb-$portVersion.patchset"
```
alaviss:
```diff
 CHECKSUM_SHA256="f5abe8b5b209c2f36560b75f32ce61412f39a2922f7045ae764a2c23335b6664"
+SOURCE_FILENAME="leveldb-$portVersion.tar.gz"
 PATCHES="leveldb-$portVersion.patchset"
```

dev-libs/leveldb/leveldb-1.20.recipe
```diff
@@ -0,0 +1,80 @@
...
+DESCRIPTION="LevelDB stores keys and values in arbitrary byte arrays, and \
+data is sorted by key. It supports batching writes, forward and backward \
+iteration, and compression of the data via Google's Snappy compression \
+library.
+LevelDB is not an SQL database. Like other NoSQL and Dbm stores, it does \
+not have a relational data model and it does not support SQL queries. Also, \
+it has no support for indexes. Applications use LevelDB as a library, as it \
+does not provide a server or command-line interface."
+HOMEPAGE="http://leveldb.org/"
+COPYRIGHT="2011 The LevelDB Authors"
```
alaviss:
```diff
-COPYRIGHT="2011 The LevelDB Authors"
+COPYRIGHT="2017 The LevelDB Authors"
```

dev-libs/leveldb/leveldb-1.20.recipe
extrowerk: I think i had a reason for that as it was not able to find the .so with $libVersion.
Can you maybe test your suggestion?

fbrosson: Sorry, I assumed your libVersion was correct. But it seems it is not.
If it was correct, haikuporter would find the lib (and would not complain.)
To find out the "full" name of the lib, you can `ls dev-libs/leveldb/work-1.20/packaging`/leveldb/lib/libleveldb.so.1.*`

dev-libs/leveldb/leveldb-1.20.recipe
```diff
@@ -0,0 +1,80 @@
...
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage leveldb$secondaryArchSuffix \
+	"$libDir"/libleveldb.so.$libVersionCompat
```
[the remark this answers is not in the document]
fbrosson: You really have to change `"$libDir"/libleveldb.so.$libVersionCompat` to `"$libDir"/libleveldb.so.$libVersion` and set the right value for `libVersionCompat`, because `compat >= 1` is _not_ part of the filename.

You could also revert to `"$libDir"/libleveldb.so.1`, but you would still have to set the right value for `libVersionCompat`, otherwise your recipe will report an incorrect version.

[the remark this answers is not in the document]
fbrosson: This is puzzling: According to archlinux, the name of the lib is not `libleveldb.so.1` but `libleveldb.so.1.20`, so I don't understand why you got an error when you had `libVersion="1.20"`.
Source: https://www.archlinux.org/packages/extra/x86_64/leveldb/files/

dev-libs/leveldb/leveldb-1.20.recipe
```diff
@@ -0,0 +1,81 @@
...
+REVISION="1"
+SOURCE_URI="https://github.com/google/leveldb/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="f5abe8b5b209c2f36560b75f32ce61412f39a2922f7045ae764a2c23335b6664"
+SOURCE_FILENAME="leveldb-$portVersion.tar.gz"
+PATCHES="leveldb-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="1"
```
fbrosson: Shouldn't this be `libVersion="1.20"` (or, even better,  `libVersion="$portVersion"`) ?
Yeah, I know I already commented about this.
BTW, I just added another comment [here](https://github.com/haikuports/haikuports/pull/2555#discussion_r188390049).

extrowerk: HP packages the libleveldb.so.1.20 into the develop package, so the debuginfo part cannot find it as it doesn't search in the packaging/develop subfolder.
No idea if it is ok or not.

fbrosson: Hmm, I think the recipe is missing a call to `fixPkgconfig` after the call to `prepareInstalledDevelLibs`.

extrowerk: Sadly that doesn't help.

alaviss: It should be `libVersion="$portVersion"` IMO, based on [upstream](https://github.com/google/leveldb/blob/v1.20/Makefile#L124)

> HP packages the libleveldb.so.1.20 into the develop package, so the debuginfo part cannot find it as it doesn't search in the packaging/develop subfolder.
No idea if it is ok or not.

According to the [`Makefile`](https://github.com/google/leveldb/blob/v1.20/Makefile#L130), the `libleveldb.so.1` and `libleveldb.so` were symlinked to `libleveldb.so.1.20`. This means that the package could not be used without the `_devel` version installed, so not OK from me. That also means `prepareInstalledDevelLibs` might need to be fixed up.

pulkomandy: Indeed not ok, and this is exacly what prepareInstalledDevelLibs is supposed to handle. In libs there should be only the .1.20 file, and in develop/lib there should be 3 symlinks to it: .so, .so.1 and .so.1.20.

The recipe should install everything in lib and let prepareInstalledDevelLibs handle this. So how did we end with an actual .so.1.20 file in developLibs in the first place?

extrowerk: It is magic, better to say: idk. I can think that the compat string is wrong and HP picks the wrong file or symlink, but i'm not sure how does it works under the hood.

dev-libs/leveldb/leveldb-1.20.recipe
```diff
@@ -0,0 +1,81 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	install -d -m 755 "$libDir" "$includeDir" "$developDocDir"
+
+	cp -R include/* "$includeDir"
+	cp out-static/*.a "$libDir"
+	cp out-shared/libleveldb.so* "$libDir"
```
fbrosson:
```diff
-	cp out-shared/libleveldb.so* "$libDir"
+	cp --no-dereference out-shared/libleveldb.so* "$libDir"
```
This tiny change allows to set `libVersion="$portVersion"`.

dev-libs/leveldb/leveldb-1.20.recipe
fbrosson: The fix is simple: [pass `--no-dereference`](#r188610794) to `cp`  when copying the shared lib and its symlinks.

extrowerk: Oh, snap!

PR #2564, opened 2018-05-17
lz4: bump to 1.8.2, add TEST, move some patches to PATCH.

fbrosson:


fbrosson:
FWIW, I've opened lz4/lz4#540 so maybe we will be able to drop our tiny patchset with the next release :)

PR #2556, opened 2018-05-14
Yaml_CPP: uniform, rename, introduce conflicts

extrowerk:

PR #2565, opened 2018-05-17
openxcom: Update icon and revision number bump.

BeOSRefugee:
Updated icon to latest version from upstream and incremented revision number.


games-strategy/openxcom/openxcom-1.0.recipe
```diff
@@ -7,7 +7,7 @@ community."
 HOMEPAGE="http://openxcom.org/"
 COPYRIGHT="2010-2014 OpenXcom Developers"
 LICENSE="GNU GPL v3"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://codeload.github.com/SupSuper/OpenXcom/tar.gz/v$portVersion"
```
fbrosson: Hmm, I just noticed we could also improve the recipe a little bit. Since you are already editing it, could you please apply the following small changes?
```diff
 REVISION="2"
-SOURCE_URI="https://codeload.github.com/SupSuper/OpenXcom/tar.gz/v$portVersion"
+SOURCE_URI="https://github.com/SupSuper/OpenXcom/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="45acb280010a01d60506b1c5f2951ae501c012cc6161aac470bd15c1e6981246"
+SOURCE_FILENAME="OpenXcom-$portVersion.tar.gz"
 SOURCE_DIR="OpenXcom-$portVersion"
```
(That's because the name of the tarball which gets download does not end with the appropriate suffix. The name of the file we get is just `v1.0`, with no suffix at all.)
Alternatively, I can push this to your PR if you agree.
Thanks!

BeOSRefugee: Sure, although it won't build yet - the 6.x version of yaml-cpp doesn't work with openxcom. @extrowerk submitted PR that fixes the issue, then I'm going to do a PR with all of his changes to openxcom, which fix a whole bunch of other problems and add nightly builds. The rdef file is pretty much the only file that isn't changed, so I figured I'd use it as a test run for my first PR.

KapiX: It's not yaml-cpp that's broken, it's openxcom's detection. Explained here:
https://github.com/haikuports/haikuports/pull/2556#issuecomment-414123309

There is a possibility to fix this in Haiku, but I don't see a reason to do that. yaml-cpp does not intentionally expose 'null.h' in global search path, so breakage should be fixed in apps. That this doesn't fail on Linux is not an argument.

waddlesplash:
Obsolete.

Issue #2568, opened 2018-05-18
libvpx: update to current


kenmays:
libvpx 1.7.0 or higher required for Qt WebEngine 5.10.1.
Enhancement: add svc_context.h to devel package.

PR #2571, opened 2018-05-19
LibVPX4: bump

extrowerk:
Closes: https://github.com/haikuports/haikuports/issues/2568

### DIFF
--- a/media-libs/libvpx/libvpx4-1.7.0.recipe
+++ b/media-libs/libvpx/libvpx4-1.7.0.recipe
@@ -1,0 +1,91 @@
+SUMMARY="WebM VP8 video codec library"
+DESCRIPTION="The WebM VP8 SDK allows you to integrate your applications with \
+the VP8 video codec, a high quality, royalty free, open source codec deployed \
+on millions of computers and devices worldwide.
+
+This distribution of the WebM VP8 Codec SDK includes the following support:
+- WebM VP8 Encoder
+- WebM VP8 Decoder"
+HOMEPAGE="http://www.webmproject.org"
+COPYRIGHT="2009-2012 Google Inc.,
+	The Mozilla Foundation and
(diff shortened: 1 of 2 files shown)


fbrosson:
Oops, sorry, I had not noticed your PR.

extrowerk:
@fbrosson ?

fbrosson:
I started working on this recipe and had almost finished it when I assigned myself to #2571. At that moment I had not noticed you had already opened this PR.)
(The only thing which I had not finished was the very last chunk of the patchset and the `CONFLICTS_devel` topic.)

Well, when I noticed some errors in your PR and also that your patchset was a bit different from the one I wrote, closely based on that of `1.6.1`, I decided to finish my recipe and open a PR.

I'll add a [`Co-authored-by:`](https://help.github.com/articles/creating-a-commit-with-multiple-authors/) with your `users.noreply.github.com` e-mail address so that, if you agree, we can switch to my PR.

fbrosson:
The alternate PR is #2573. I'll only merge it if you are OK with it, of course.
BTW, if you prefer, instead of switching to #2573, I can update #2571 with my changes but keeping you as unique author or adding myself as co-author to the commit. Hmm, after all, this last scenari seems to be the best option. What do you think?

extrowerk:
let's @fbrosson have all the fun :)

kenmays:
@fbrosson - Update didn't include this header: svc_context.h

fbrosson:
Hmm, we need to add `--enable-vp9` to have that file installed.
Let me check and I'll fix this asap.

fbrosson:
Passing `--enable-vp9` was not enough. We also need to pass `CONFIG_SPATIAL_SVC=yes`.
I'll open a PR with this change so that other contributors can comment on this.

fbrosson:
→ #2596

PR #2573, opened 2018-05-19
libvpx: bump to 1.7.0.

fbrosson:
It is safe to drop `libvpx4-1.6.1` because no package is using it at the moment: `games-fps/eduke32`, `media-libs/{gd,gst_plugins_good}`, `media-video/{ffmpeg,handbrake}` and `sci-visualization/gnuplot` are all using `libvpx3-1.5.0`.

Co-authored-by: @extrowerk

### DIFF
--- a/media-libs/libvpx/libvpx-1.7.0.recipe
+++ b/media-libs/libvpx/libvpx-1.7.0.recipe
@@ -6,25 +6,37 @@ on millions of computers and devices worldwide.
 This distribution of the WebM VP8 Codec SDK includes the following support:
 - WebM VP8 Encoder
 - WebM VP8 Decoder"
-HOMEPAGE="http://www.webmproject.org"
+HOMEPAGE="https://www.webmproject.org/"
 COPYRIGHT="2009-2012 Google Inc.,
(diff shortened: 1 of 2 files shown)


fbrosson:
A few recipes depend libvpx, currently provided by `libvpx4-1.6.1.recipe`:
* games-fps/eduke32
* media-video/ffmpeg
* media-video/handbrake
* media-video/vlc
* net-libs/tox
* sci-visualization/gnuplot

So I'm wondering if it would be desirable to try to build them with `libvpx5.so` (provided by `libvpx-1.7.0`) and, if all these packages build and work fine with `libvpx5` we could, maybe, remove `libvpx4-1.6.1.recipe` and its patchset. This would make the `git log -p` much smaller.
~Shall we try this?~

Hmm, there's a much better option: we could drop `libvpx3-1.5.0.recipe` and its patchset, `libvpx-1.5.0.patchset`. This will be faster because we only need to check the packages which were built against `libvpx3.so` and were not rebuilt after `libvpx4.1.6.1` was added.
I'll check this asap.

PR #2575, opened 2018-05-19
mini-xml, fix 2.6 recipe

Begasus:


fbrosson:
> set 2.9 as broken

You can drop this part of the commit message, because `2.9` is already [marked as broken](https://github.com/haikuports/haikuports/blob/master/dev-libs/mini-xml/mini_xml-2.9.recipe#L14):
```bash
grep ARCHITECTURES mini_xml-2.9.recipe
ARCHITECTURES="!x86_gcc2 !x86 !x86_64"
SECONDARY_ARCHITECTURES="!x86_gcc2 !x86"
```

PR #2582, opened 2018-05-20
EDK2: new recipe

extrowerk:


sys-boot/edk2/edk2-2018.recipe
```diff
@@ -0,0 +1,81 @@
...
+CHECKSUM_SHA256="a37aa6dc7a4326b562ce03e225afd96a0cd1dc847d9aeddfd867ce6b2e7cc7bc"
+SOURCE_DIR="edk2-vUDK$portVersion"
+PATCHES="edk2-$portVersion.patchset"
+DISABLE_SOURCE_PACKAGE="yes"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	edk2 = $portVersion compat >= 0
```
fbrosson:
```diff
 PATCHES="edk2-$portVersion.patchset"
-DISABLE_SOURCE_PACKAGE="yes"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
+DISABLE_SOURCE_PACKAGE="yes"
+
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+fi
 
 PROVIDES="
-	edk2 = $portVersion compat >= 0
+	edk2$secondaryArchSuffix = $portVersion compat >= 0
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
 	"
```

Remark: for `DISABLE_SOURCE_PACKAGE`, we have about 88 recipes which define this variable, and it seems the majority has it just after `ARCHITECTURES`, but I'm OK with keeping it after `PATCHES` too.

sys-boot/edk2/edk2-2018.recipe
```diff
@@ -0,0 +1,81 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libuuid$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:iasl$secondaryArchSuffix
```
fbrosson:
```diff
-	cmd:iasl$secondaryArchSuffix
+	cmd:iasl$commandSuffix
```

sys-boot/edk2/edk2-2018.recipe
```diff
@@ -0,0 +1,81 @@
...
+	cmd:sed
+	"
+PATCH()
+{
+	local ACTIVE_PLATFORM="MdeModulePkg/MdeModulePkg.dsc"
+	local TARGET="RELEASE" # DEBUG, RELEASE, NOOPT
+	local TARGET_ARCH="X64" # IA32, IPF, X64, EBC, ARM or AArch64
+	local TOOL_CHAIN_TAG="GCC5"
+
+	if [ "$targetArchitecture" = x86_gcc2 ]; then
```
fbrosson: In case we need `TARGET_ARCH="IA32"` for x86 (both primary and secondary) then I would recommend this:
```diff
-	if [ "$targetArchitecture" = x86_gcc2 ]; then
+	if [ "$effectiveTargetArchitecture" = x86 ]; then
 		local TARGET_ARCH="IA32" # IA32, IPF, X64, EBC, ARM or AArch64
 	fi
```
as it would also work for x86 primary arch.
Testing `[ "$targetArchitecture" = x86_gcc2 ]` only catches the x86 secondary arch of x86_gcc2.

sys-boot/edk2/edk2-2018.recipe
```diff
@@ -0,0 +1,81 @@
...
+	. edksetup.sh
+	local EDK_TOOLS_PATH=`pwd`/BaseTools
+	. edksetup.sh BaseTools
+	build
+}
+
+INSTALL()
+{
+	mkdir -p $dataDir/edk2
+	find Build -type f -name "*.efi" -exec cp "{}" $dataDir/edk2 \;
```
fbrosson: Suggestion:
```diff
-	mkdir -p $dataDir/edk2
+	mkdir -p "$dataDir"/edk2
-	find Build -type f -name "*.efi" -exec cp "{}" $dataDir/edk2 \;
+	find Build -type f -name "*.efi" -exec cp {} "$dataDir"/edk2 \;
```
Hmm, you can keep the quotes around `{}` if you wish, but they are not required.
Yeah, I know, I put quotes in `libVersion="x.y.z"` but this is because I like to do so :)
Anyway, I'm OK with `"{}"` ;)

sys-boot/edk2/edk2-2018.recipe
```diff
@@ -0,0 +1,81 @@
...
+
+	sed -i "s|gcc-ar|ar|g" BaseTools/Conf/tools_def.template
+}
+
+BUILD()
+{
+	make -C BaseTools
+	. edksetup.sh
+	local EDK_TOOLS_PATH=`pwd`/BaseTools
+	. edksetup.sh BaseTools
```
fbrosson: Suggestion:
```diff
-	. edksetup.sh
+	. ./edksetup.sh
-	local EDK_TOOLS_PATH=`pwd`/BaseTools
+	local EDK_TOOLS_PATH="$sourceDir"/BaseTools
-	. edksetup.sh BaseTools
+	. ./edksetup.sh BaseTools
```
Yep, I know `.` is in `PATH`, so this change is not currently required.
But I'd still recommend using an explicit `./edksetup.sh` because it would allow to remove the `.` from the default `PATH` if we ever decide to do so.

sys-boot/edk2/edk2-2018.recipe
extrowerk: What did you meant here?

fbrosson: We need to define `commandSuffix` in edk2's recipe because one of its dependencies, `cmd:iasl` is provided by iasl as `cmd:iasl$commandSuffix`.
(I agree, this is a bit annoying, but the only way to avoid this would be to give up using `commandSuffix`.)

extrowerk: I still can't get your point. In BUILD_REQ i have cmd:iasl$secondaryArchSuffix

fbrosson: If you try to build edk2_x86 on x86_gcc2 it will say it can't find `cmd:iasl_x86`.
That's because the package for iasl provides `cmd:iasl` on x86_gcc2 instead of `cmd:iasl_x86`.

Of course, you could mention `cmd:iasl` instead of `cmd:iasl$secondaryArchSuffix` and it would work on x86_gcc2 but it would not work if you tried to build edk2_x86 on x86_64. So what you need here is exactly `cmd:iasl$commandSuffix` as this is the only expression that is valid in all situations.

sys-boot/edk2/edk2-2018.recipe
```diff
@@ -0,0 +1,89 @@
...
+	. ./edksetup.sh
+	local EDK_TOOLS_PATH="$sourceDir"/BaseTools
+	. ./edksetup.sh BaseTools
+	build
+}
+
+INSTALL()
+{
+	mkdir -p "$dataDir"/edk2
+	find Build -type f -name "*.efi" -exec cp "{}" "$dataDir"/edk2 \;
```
fbrosson:
```diff
-	mkdir -p $dataDir/edk2
+	install -m 0755 -d "$dataDir"/edk2
-	find Build -type f -name "*.efi" -exec cp "{}" $dataDir/edk2 \;
+	cd Build
+	install -m 0644 -t "$dataDir"/edk2 */RELEASE*/*/*.efi
```
There is only one directory which matches this pattern: `$sourceDir/Build/*/RELEASE*/*/`
And that directory has all the `*.efi` we want to install. There are other subdirectories, with a deeper level, but they contain variants (e.g. debug) or copies of the files in the directory mentioned above, but these variants have the same names as the `*.efi` files in that main directory.
So using a `find` to copy everything would lead each target file being overwritten several times. And the last copies could even be from the debug variants.

BTW, the build fails with `gcc 5.4` on x86 (both primary and secondary), even if we define `HOST_ARCH=IA32`. But I should probably upgrade to `gcc 7.3` and try again.

alaviss: @fbrosson If you need to upgrade, then the recipe should declare dependency on newer gcc

fbrosson: `gcc 7.3.0` did not help. So I guess we don't need to declare any condition for a particular version of gcc.
But maybe the problem is somewhere else:
```
/boot/system/develop/tools/x86/bin/../lib/gcc/i586-pc-haiku/7.3.0/../../../../i586-pc-haiku/bin/ld: unrecognised emulation mode: elf_i386
Supported emulations: elf_i386_haiku
make: *** [/sources/edk2-vUDK2018/Build/MdeModule/RELEASE_GCC5/IA32/MdeModulePkg/Application/HelloWorld/HelloWorld/DEBUG/HelloWorld.dll] Error 1
GNUmakefile:385: recipe for target '/sources/edk2-vUDK2018/Build/MdeModule/RELEASE_GCC5/IA32/MdeModulePkg/Application/HelloWorld/HelloWorld/DEBUG/HelloWorld.dll' failed
```
Why does it try to build a **_DLL_**!!! See the `HelloWorld.dll` at the end of the last line.
I guess some Makefiles need to be changed or skipped.

This is puzzling, however, because on x86_64 I don't get this error. And I do get a `HelloWorld.efi`:
```bash
package list -p edk2-2018-1-x86_64.hpkg | grep -i hello
data/edk2/HelloWorld.efi
```

extrowerk: Thanks for the input, but i don't care about 32 bit.

fbrosson: I think it might be a good idea to mark x86 as untested instead of broken:
```diff
-ARCHITECTURES="!x86_gcc2 !x86 x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
-SECONDARY_ARCHITECTURES="!x86"
+SECONDARY_ARCHITECTURES="?x86"
```
With just this extra change (neutral for x86_64) in `BUILD`:
```diff
-	make -C BaseTools
+	if [ "$effectiveTargetArchitecture" = x86 ]; then
+		HOST_ARCH=IA32
+	fi
+	make -C BaseTools ${HOST_ARCH:+HOST_ARCH=$HOST_ARCH}
```
the build starts for x86 (primary and secondary).
This would allow owners of 32-bit EFI systems to investigate and, maybe, fix the build on x86 :)
What do you think?
(That said, I'm OK with keeping the recipe as is, i.e. with x86 marked as broken.)

Issue #438, opened 2016-01-26
add ballsmacker


scottmc:
goes in games-sports/ballsmacker

Ballsmacker is an OpenGL virtual pool game.
Ballsmacker is a Free virtual pool game rendered in OpenGL. Currently British pool is supported; there are plans to support many different billiards games. At the moment the game can be played with two players taking turns on one computer; in the future networked multiplayer will be supported.
http://ballsmacker.sourceforge.net/

[ballsmacker-1.0.0.bep.txt](https://github.com/haikuports/haikuports/files/105672/ballsmacker-1.0.0.bep.txt)

PR #2583, opened 2018-05-20
ballsmacker, new recipe

Begasus:
Solves issue https://github.com/haikuports/haikuports/issues/438

### DIFF
--- a/games-sports/ballsmacker/ballsmacker-1.0.0.recipe
+++ b/games-sports/ballsmacker/ballsmacker-1.0.0.recipe
@@ -1,0 +1,72 @@
+SUMMARY="An OpenGL virtual pool game"
+DESCRIPTION="Ballsmacker is a Free virtual pool game rendered in OpenGL. \
+Currently British and American pool are supported; there are plans to \
+support many different billiards games. At the moment the game can be played \
+with two players taking turns on one computer; in the future networked \
+multiplayer will be supported."
+HOMEPAGE="http://ballsmacker.sourceforge.net/"
+COPYRIGHT="2004 Simon Howard"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/ballsmacker/ballsmacker-$portVersion.tar.gz"
+CHECKSUM_SHA256="5eae4bc75fb9248ebba43cde312e5f22f889f600da7def23307be75df3130a94"
+
+ARCHITECTURES="?x86_gcc2 ?x86_64 ?x86"
+SECONDARY_ARCHITECTURES="?x86"
+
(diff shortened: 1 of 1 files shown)


Begasus:
Builds, installs for x86_gcc2 and x86 (secondaryArchSuffix), but crashes (as reported at the issue) ...

games-sports/ballsmacker/ballsmacker-1.0.0.recipe
```diff
@@ -0,0 +1,72 @@
...
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	ballsmacker$secondaryArchSuffix = $portVersion
+	cmd:ballsmacker$commandSuffix = $portVersion
```
Begasus: If we are on the subject "" for libDir etc, would it be wise to also use ... cmd:ballsmacker"$commandSuffix" = $portVersion here? Or even include "$portVersion" here?

fbrosson: I agree that this is a bit subjective.
What I can say is that:
* `$commandSuffix` will never include any special character, so not using "" for it is OK and will always be.
* `$portVersion` is also quite safe, as it is computed from the name of the recipe. And since we don't use white spaces or tabs in the names of our recipes, we can safely use `$portVersion` without any "".
* `$prefix`, `$libDir`, `$binDir`,`$commandBinDir` and all other similar variables are **_currently_** safe, but if haikuporter's `--nochroot` ever gets a fix that would allow to build a recipe without entering a chroot, and **if** a user is "crazy" enough to have his haikuports tree in a path which has one or more white spaces or tabs, then, in that situation, using this variables without any quotes would lead to errors.

There is an exception, however: When we define `commandBinDir=$binDir` we don't need the quotes, even if `$binDir` has spaces. This is because the shell evaluates `$binDir` and behaves like if there were quotes around it.
But when we call `runConfigure`, passing `--bindir=$commandSuffix` (instead of `--bindir="$commandSuffix"`) would not be safe in the scenario described above.

Now, given that nobody needs to use haikuporter's `--nochroot` feature, we don't need to fix it. So it is likely that the quotes we are adding (in new recipes or in lines that are already being changed for other reasons) will never be needed.

My rationale is that if using quotes does not add noise to the `git log -p` then I think it is slightly better to use them. But if it's adding noise, then we can stick to non quoted variables.

Here, we have a new recipe. So using quotes from the beginning is nice because it means that we would never need to add them later.

So, as a conclusion I can say that I'll probably continue using/adding quotes when it does not add noise to the repo, but, at the same time, I'm OK with any contributor's preferences.

PR #2585, opened 2018-05-20
LibUniBreak: new recipe

extrowerk:


dev-libs/libunibreak/libunibreak-4.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+DESCRIPTION="An implementation of the line breaking and word breaking \
+algorithms as described in Unicode Standard Annex 14 and Unicode Standard \
+Annex 29."
+HOMEPAGE="http://vimgadgets.sourceforge.net/libunibreak/"
+COPYRIGHT="2016 Andreas Röver"
+LICENSE="Zlib"
+REVISION="1"
+SOURCE_URI="https://github.com/adah1972/libunibreak/archive/libunibreak_4_0.tar.gz"
+CHECKSUM_SHA256="8ea270d2d9dea6b633f7de1a4a624163c3549b4b6f5031f08c6b96d704d58aa4"
+SOURCE_DIR="libunibreak-libunibreak_4_0"
```
fbrosson: Suggestion:
```diff
-SOURCE_URI="https://github.com/adah1972/libunibreak/archive/libunibreak_4_0.tar.gz"
+SOURCE_URI="https://github.com/adah1972/libunibreak/archive/libunibreak_${portVersion//./_}.tar.gz"
 CHECKSUM_SHA256="8ea270d2d9dea6b633f7de1a4a624163c3549b4b6f5031f08c6b96d704d58aa4"
-SOURCE_DIR="libunibreak-libunibreak_4_0"
+SOURCE_DIR="libunibreak-libunibreak_${portVersion//./_}"
```
Well, maybe this is not a good idea. After all, if there are no more releases since 2016, then we can probably stick to the hard coded `4_0`.
(Feel free to choose whatever you think is best.)

dev-libs/libunibreak/libunibreak-4.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+{
+	make install
+
+	# remove libtool library files
+	rm $libDir/*.la
+	rm $libDir/liblinebreak.a
+
+	# prepare develop/lib
+	prepareInstalledDevelLibs libunibreak
+	ln -s $developLibDir/libunibreak.a $developLibDir/liblinebreak.a
```
fbrosson:
```diff
-	ln -s $developLibDir/libunibreak.a $developLibDir/liblinebreak.a
+	ln -s libunibreak.a $developLibDir/liblinebreak.a
```
Alternatively, you could use `ln -r -s` and keep the same arguments.
BTW, maybe we don't need to symlink `libunibreak.a`.

dev-libs/libunibreak/libunibreak-4.0.recipe
alaviss: Upstream also provide an another tarball: https://github.com/adah1972/libunibreak/releases/download/libunibreak_4_0/libunibreak-4.0.tar.gz

This tarball has a different checksum and appears to contains docs as well, so you might want to use it instead.

dev-libs/libunibreak/libunibreak-4.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage libunibreak$secondaryArchSuffix \
+	"$libDir"/libunibreak.so.$libVersion
+
+BUILD()
+{
+	./autogen.sh
+	runConfigure ./configure
```
alaviss:
```diff
-	./autogen.sh
-	runConfigure ./configure
+	runConfigure ./autogen.sh
```

This is because `autogen.sh` [runs configure automatically](https://github.com/adah1972/libunibreak/blob/23aa7415254420de0aa4186e72f602257762c9ec/autogen.sh#L12)

dev-libs/libunibreak/libunibreak-4.0.recipe
extrowerk: According my tests we do need the a symlink.

fbrosson: OK.
(But it would still be nice to make it a _relative_ symlink.)

PR #2587, opened 2018-05-20
gnuplot: bump to 5.0.7, switch to libvpx.so.5.

fbrosson:


fbrosson:
Oops, i forgot to reset `REVISION` to `1`...
When I noticed it it was too late.
Anyway, it works the same way :)

PR #2591, opened 2018-05-21
python3: fix #2558 'str' has no attribute 'append'

### LINKED ISSUE #2558: python ctypes.util 'str' object has no attribute 'append'
https://github.com/haikuports/haikuports/blob/e3878f5ecd45ef70fadb1477808c6109cfded91a/dev-lang/python/patches/python3_x86-3.6.3.patchset#L20

should be `directory += "/x86"` or better `directory = os.path.join(directory, "x86")`

found this while installing something via pip

irtusb:
#2558 

append method does not exist for strings in python


korli:
Please bump the recipe revision.

irtusb:
this way?

PR #2600, opened 2018-05-24
angband, bump version

Begasus:
Builds ok on x86 (secondaryArch), not tested (yet) on x86_x64, fix for gcc2 would be minor (mostly parse errors) ...

### DIFF
--- a/games-roguelike/angband/angband-4.1.2.recipe
+++ b/games-roguelike/angband/angband-4.1.2.recipe
@@ -1,0 +1,113 @@
+SUMMARY="A free, single-player dungeon exploration game"
+DESCRIPTION="Angband is a free, single-player dungeon exploration game. \
+You play an adventurer: seeking riches, fighting monsters, and preparing \
+for a final battle with Morgoth, the Lord of Darkness."
+HOMEPAGE="http://rephial.org/"
+COPYRIGHT="1985 Robert Alan Koeneke
+	1989 James E. Wilson
+	1990 Alex Cutler and Andy Astrand
+	1991-1992 Sean Marsh, Geoff Hill, Charles Teague
+	1993 Charles Swiger
+	1994-1999 Ben Harrison
+	2000-2014 Robert Ruehlmann"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://rephial.org/downloads/4.1/angband-$portVersion.tar.gz"
+CHECKSUM_SHA256="30bc0979e0845cdc43de2a8f65c4d54d03d24d402b32b8589fbbc368ccfa0e2a"
+PATCHES="angband-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+GLOBAL_WRITABLE_FILES="
+	settings/angband/customize/font-gcu.prf keep-old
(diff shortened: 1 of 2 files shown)


games-roguelike/angband/patches/angband-4.1.2.patchset
```diff
@@ -0,0 +1,36 @@
...
+diff --git a/src/z-rand.c b/src/z-rand.c
+index 57f0e4b..526af98 100644
+--- a/src/z-rand.c
++++ b/src/z-rand.c
+@@ -537,7 +537,9 @@ void rand_fix(u32b val)
+ 	rand_fixval = val;
+ }
+ 
++#ifndef __HAIKU__
+ int getpid(void);
```
alaviss: Actually, just changing `int` to `pid_t` should solve the problem for us and possibly others, as that's the [correct signature][0] for the function :)

```diff
-+#ifndef __HAIKU__
- int getpid(void);
+-int getpid(void);
++pid_t getpid(void);
-+#endif
```
[0]: http://pubs.opengroup.org/onlinepubs/9699919799/functions/getpid.html

games-roguelike/angband/angband-4.1.2.recipe
```diff
@@ -0,0 +1,112 @@
...
+	settings/angband/gamedata/slay.txt keep-old
+	settings/angband/gamedata/store.txt keep-old
+	settings/angband/gamedata/terrain.txt keep-old
+	settings/angband/gamedata/trap.txt keep-old
+	settings/angband/gamedata/vault.txt keep-old
+	settings/angband/gamedata/world.txt keep-old
+	"
+
+ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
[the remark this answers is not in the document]
Begasus: Started out to try to fix them, using commandBinDir would be the next step if no one disapproves :)

games-roguelike/angband/patches/angband-4.1.2.patchset
```diff
@@ -0,0 +1,35 @@
...
+diff --git a/configure.ac b/configure.ac
+index 94ef3fe..66fdf2c 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -98,7 +98,7 @@ CFLAGS="$CFLAGS -DHAVE_CONFIG_H"
+ CPPFLAGS="$CPPFLAGS -I." 
+ 
+ if test "$GCC" = "yes"; then
+-	CFLAGS="$CFLAGS -W -Wall -Wextra -Wno-unused-parameter -pedantic"
++	CFLAGS="$CFLAGS -W -Wall"
```
alaviss: Why was the warning flags removed? Also, `-pedantic` is necessary for strict ISO C compilance, so I would prefer to keep it.

Begasus: I used some of the patch from previous version, probably due to not being able to build it on x86_gcc2 at that moment ... 
https://github.com/haikuports/haikuports/blob/75a3cf5384f650242f4b8a1094ab305d33ed99e7/games-roguelike/angband/patches/angband-3.5.0.patchset

Begasus: This isn't needed indeed, I think they were needed to get the previous build for x86_gcc2 working ....

PR #2601, opened 2018-05-24
OpenBLAS: bump

extrowerk:
WIP


sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
+SUMMARY="An optimized BLAS library"
+DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD\
+ version."
```
fbrosson:
```diff
-DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD\
+DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD \
- version."
+version."
```

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+HOMEPAGE="http://www.openblas.net/"
+COPYRIGHT="2011-2018 The OpenBLAS Project"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://github.com/xianyi/OpenBLAS/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="cf51543709abe364d8ecfb5c09a2b533d2b725ea1a66f203509b21a8e9d8f1a1"
+SOURCE_DIR="OpenBLAS-$portVersion"
```
fbrosson:
```diff
+SOURCE_FILENAME="OpenBLAS-$portVersion.tar.gz"
 SOURCE_DIR="OpenBLAS-$portVersion"
```
([HaikuPorter-Guidelines#archive-name](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#archive-name))

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+PATCHES="openblas-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
fbrosson:
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
```
(No need to mention broken arches on `SECONDARY_ARCHITECTURES`)

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+
+PROVIDES="
+	openblas$secondaryArchSuffix = $portVersion
+	lib:libopenblas$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libopenblasp_r${portVersion}$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson:
```diff
-	lib:libopenblasp_r${portVersion}$secondaryArchSuffix = $portVersion compat >= 0
+	lib:libopenblasp_r$portVersion$secondaryArchSuffix = $portVersion compat >= 0
```
(You can drop the `{}` in here.)

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgfortran$secondaryArchSuffix
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	openblas${secondaryArchSuffix}_devel = $portVersion
+	devel:libopenblas$secondaryArchSuffix = $portVersion compat >= 0
+	devel:libopenblasp_r${portVersion}$secondaryArchSuffix = $portVersion compat >= 0
```
fbrosson:
```diff
-	devel:libopenblasp_r${portVersion}$secondaryArchSuffix = $portVersion compat >= 0
+	devel:libopenblasp_r$portVersion$secondaryArchSuffix = $portVersion compat >= 0
```
(You can drop the `{}` in here.)

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+	cmd:gfortran$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
+	cmd:make
+	cmd:perl
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage openblas$secondaryArchSuffix \
+	"$libDir"/libopenblas.so.0 \
+	"$libDir"/libopenblasp-r${portVersion}.so
```
fbrosson:
```diff
-	"$libDir"/libopenblasp-r${portVersion}.so
+	"$libDir"/libopenblasp-r$portVersion.so
```
(You can drop the {} in here.)

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+
+BUILD()
+{
+	# DYNAMIC_ARCH is x86 only
+	# PPC & others should specify TARGET=
+	target=
+	[ "$targetArchitecture" == x86 ] && target=NORTHWOOD
+	[ "$targetArchitecture" == x86_64 ] && target=PRESCOTT
```
fbrosson:
```diff
-	target=
-	[ "$targetArchitecture" == x86 ] && target=NORTHWOOD
-	[ "$targetArchitecture" == x86_64 ] && target=PRESCOTT
+	case "$effectiveTargetArchitecture" in
+		x86) target=NORTHWOOD;;
+		x86_64) target=PRESCOTT;;
+		*) target=;;
+	esac
```

Without this change, `BUILD()` would fail as soon as it reaches a test that returns false.

sci-libs/openblas/openblas-0.3.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+	make MAKE_NB_JOBS=${jobArgs#-j} NO_LAPACKE=1 NO_AFFINITY=1 NO_WARMUP=1 \
+		NUM_THREADS=64 DYNAMIC_ARCH=1 USE_OPENMP=1 TARGET=$target
```
fbrosson:
```diff
-		NUM_THREADS=64 DYNAMIC_ARCH=1 USE_OPENMP=1 TARGET=$target
+		NUM_THREADS=64 DYNAMIC_ARCH=1 USE_OPENMP=1 ${target:+TARGET=$target}
```
(This change allows to only pass `TARGET=$target` if `$target` is not an empty string.)

Well, maybe this suggestion is not needed at all.

sci-libs/openblas/patches/openblas-0.3.0.patchset
```diff
@@ -0,0 +1,196 @@
...
+diff --git a/driver/others/memory.c b/driver/others/memory.c
+index ef328b9..6a5c843 100644
+--- a/driver/others/memory.c
++++ b/driver/others/memory.c
+@@ -108,6 +108,10 @@ USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ #include <sys/resource.h>
+ #endif
+ 
++#ifdef OS_HAIKU
++#include <posix/unistd.h>
```
alaviss:
```diff
-+#include <posix/unistd.h>
++#include <unistd.h>
```

extrowerk:
Main problem to solve before cosmetic fixes:

runtime_loader: /boot/system/lib/libgfortran.so.4.0.0: Could not resolve symbol '__cpu_model'

@waddlesplash: ping, something really broken here. The current HDF5 recipe also not buildable because this.

extrowerk:
Tests:
- [ ] dev-lang/r
- [ ] dev-libs/caffe
- [ ] dev-python/numpy
- [ ] media-gfx/opentoonz
- [ ] media-libs/opencv
- [ ] sci-libs/armadillo
- [ ] sci-libs/arpack
- [ ] sci-libs/blaze
- [ ] sci-libs/cblas_reference
- [ ] sci-libs/dlib
- [ ] sci-libs/lapack
- [ ] sci-libs/mlpack
- [ ] sci-libs/mlpack
- [ ] sci-mathematics/octave

sci-libs/openblas/openblas2-0.2.20.recipe
```diff
@@ -25,12 +30,15 @@ REQUIRES="
...
+	openblas2${secondaryArchSuffix}_devel = $portVersionCompat
+	devel:libopenblas$secondaryArchSuffix = $libVersionCompat
+	devel:libopenblasp_r$portVersion$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES_devel="
-	openblas$secondaryArchSuffix == $portVersion base
+	openblas2$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_devel="
+	openblas3${secondaryArchSuffix}_devel
```
fbrosson: I might be wrong, but it looks like both `openblas2-0.2.20.recipe` and `openblas3-0.3.1.recipe` provide `lib:libopenblas$secondaryArchSuffix`, so this means their base packages will overlap.

Now, if we tweak these recipes to allow these 2 base packages to co-exist, then I'd recommend a variant for the `CONFLICTS_devel`:

First, a small remark; Although we can :
* put a :
```
CONFLICTS_devel="
	openblas3${secondaryArchSuffix}_devel
	"
```
in  `openblas2-0.2.20.recipe`
* and, at the same time, put a :
```
CONFLICTS_devel="
	openblas2${secondaryArchSuffix}_devel
	"
```
in `openblas3-0.3.1.recipe`

it is enough to have only one `CONFLICTS_devel`, in any of these recipes.
But since the first `0.3.x` was released after the first `0.2.x`, I'd recommend keeping only the  `CONFLICTS_devel` in `openblas3-0.3.1.recipe`.

sci-libs/openblas/openblas2-0.2.20.recipe
```diff
@@ -1,22 +1,27 @@
...
-DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD\
- version."
+DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD \
+version."
 HOMEPAGE="http://www.openblas.net/"
 COPYRIGHT="2011-2017 The OpenBLAS Project"
 LICENSE="BSD (3-clause)"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="http://github.com/xianyi/OpenBLAS/archive/v$portVersion.tar.gz"
```
fbrosson: `s/http/https/`

BTW, although I'm OK with the revision bump, I think other contributors might prefer a reset because we have "renames" in `PROVIDES` and `PROVIDES_devel`.

Also, are you sure openblas2 and openblas3 don't overlap?

extrowerk: They overlap for sure. What should we do then?

fbrosson: There are 2 solutions:

A) We just need to add a
```
CONFLICTS="
	openblas2$secondaryArchSuffix
	"
```
to `openblas3-0.3.1.recipe`.

Pros:
This would allow to have prebuilt packages for both openblas2 and openblas3.

Cons:
The "cost" for this is that we have two recipes, and that upgrades won't be automatic, unless we add a `REPLACES` to one of the recipes.

B) The other way to handle this is to drop openblas `0.2.x` and only keep `0.3.1`, in `openblas-0.3.1.recipe`, i.e. without the rename from "openblas" to "openblas3".
Pros:
No need for any `CONFLICTS{,_devel}` or `REPLACES`.
Cons:
`0.2.x` won't be available.

(I don't know which is best.)

extrowerk: I rebuilt all the dependent packages with openblas3, there was no problems.
I'm not against dropping the old recipe.
@korli: any input?

sci-libs/openblas/openblas2-0.2.20.recipe
```diff
@@ -1,22 +1,27 @@
...
-DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD\
- version."
-HOMEPAGE="http://www.openblas.net/"
+DESCRIPTION="OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD \
+version."
+HOMEPAGE="https://www.openblas.net/"
 COPYRIGHT="2011-2017 The OpenBLAS Project"
 LICENSE="BSD (3-clause)"
-REVISION="2"
+REVISION="3"
```
alaviss: You may want to reset `REVISION` instead.

sci-libs/openblas/openblas2-0.2.20.recipe
alaviss: I usually like to drop the old version, but it seems like Haiku has [an interest in keeping old software running][0]. So I think we should keep the old version until security problems start to surface.

[0]: https://github.com/haikuports/haikuports/issues/2788#issuecomment-404202842

pulkomandy:
On Tue, Jul 24, 2018 at 03:55:47PM +0000, alaviss wrote:
> I usually like to drop the old version, but it seems like Haiku has
> [an interest in keeping old software running][0]. So I think we should keep
> the old version until security problems start to surface.

You are exagerating my opinion here.

There is an interest in keeping BeOS software running (what would be the
point in being compatible otherwise?). There is also an interest in
keeping some useful Haiku apps running (because otherwise we would
always get complains that "the OS is nice, but there are no apps". Let's
try to keep what we have running.

But, there is no need to keep old versions of a specific piece of
software running, when there are newer versions available. In the case
of OpenBLAS, I don't even think it got much use at all from 3rd party
software yet? We can enjoy the fact that we are still not in beta and
remove it before we commit to some ABI stability, then.

alaviss:
@pulkomandy 

> Let's try to keep what we have running.

Thanks for the clarification :)

@extrowerk 

If you're sure that you have revbumped and tested all package that depends on OpenBLAS, then it's fine to drop the older version. While I've never looked at 3rd party repos, I think no one actually uses OpenBLAS anyway.

I'm not certain how we should merge the PRs in case we do drop the old package, since I don't know how the builders work. But I would assume that they will build the new OpenBLAS first then start building the rest in the queue, so merge this PR then merge the revbumps in quick succession.

waddlesplash:
> build the new OpenBLAS first then start building the rest in the queue, so merge this PR then merge the revbumps in quick succession.

Nope, the build order is not defined unless there are explicit dependencies. If we want to have implicit dependencies, we will need to merge this, wait for the builders to begin, and then merge the others so the next run will rebuild them against the new OpenBLAS. See https://github.com/haikuports/haikuporter/issues/158.

fbrosson:
Dependencies | Pros | Cons
-- | -- | --
Implicit | Already there, and working quite well. | We have to wait for the builders to finish their work before we revbump the other recipes.
Explicit | No need to schedule any build. | We'll need to update the dependencies.

I think we can continue with implicit dependencies, but I'm OK with switching to explicit ones.

BTW, we can probably do a smooth transition, i.e. not changing all recipes at the same time,, in case we choose to use explicit dependencies.
Also, we can even keep implicit dependencies for some recipes, and switch to explicit ones for other recipes, if nobody objects such a "mix" :)

pulkomandy:
The "waiting for builders" is a brittle point. For example when we bootstrap a new architecture or when we need to rebuild a lot of things because of major ABI changes (this will happen on the road to R2), we can't really handle this by a dozen of commits bumping revisions of small subsets of packages. "Pro": reliable builds, making sure a given revision of the package is matched with a precise version of the dependencies, accross all architectures (nothing guarantees this with implicit dependencies).

There is already a script to automate the revision bump, it should just be extended to also bump the dependency constraint. No extra work for developers, then. In that case, there is no "Con".

Of course there is no need to change all the recipes at once.

sci-libs/openblas/openblas3-0.3.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	openblas3$secondaryArchSuffix = $portVersionCompat
```
fbrosson:
```diff
 PROVIDES="
-	openblas3$secondaryArchSuffix = $portVersionCompat
+	openblas$secondaryArchSuffix = $portVersionCompat
```

sci-libs/openblas/openblas3-0.3.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+	lib:libopenblasp_r$portVersion$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libgfortran$secondaryArchSuffix
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	openblas3${secondaryArchSuffix}_devel = $portVersionCompat
```
fbrosson:
```diff
 PROVIDES_devel="
-	openblas3${secondaryArchSuffix}_devel = $portVersionCompat
+	openblas${secondaryArchSuffix}_devel = $portVersionCompat
```

sci-libs/openblas/openblas3-0.3.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+	devel:libopenblas$secondaryArchSuffix = $libVersionCompat
+	devel:libopenblasp_r$portVersion$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	openblas3$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_devel="
+	openblas2${secondaryArchSuffix}_devel
```
fbrosson:
```diff
-CONFLICTS_devel="
-	openblas2${secondaryArchSuffix}_devel
-	"
```

sci-libs/openblas/openblas3-0.3.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:gfortran$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
+	cmd:make
+	cmd:perl
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage openblas3$secondaryArchSuffix \
```
fbrosson:
```diff
-defineDebugInfoPackage openblas3$secondaryArchSuffix \
+defineDebugInfoPackage openblas$secondaryArchSuffix \
```

Also, don't forget to rename the files:
openblas**3**-0.3.1.{recipe,patchset} :arrow_right: openblas-0.3.1.{recipe,patchset}.

sci-libs/openblas/openblas3-0.3.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+version."
+HOMEPAGE="http://www.openblas.net/"
+COPYRIGHT="2011-2018 The OpenBLAS Project"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://github.com/xianyi/OpenBLAS/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="1f5e956f35f3acdd3c74516e955d797a320c2e0135e31d838cbdb3ea94d0eb33"
+SOURCE_FILENAME="OpenBLAS-$portVersion.tar.gz"
+SOURCE_DIR="OpenBLAS-$portVersion"
+PATCHES="openblas3-$portVersion.patchset"
```
fbrosson:
```diff
-PATCHES="openblas3-$portVersion.patchset"
+PATCHES="openblas-$portVersion.patchset"
```
(I missed this one :)

sci-libs/openblas/openblas3-0.3.1.recipe
```diff
@@ -0,0 +1,94 @@
...
+	lib:libgomp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	openblas3${secondaryArchSuffix}_devel = $portVersionCompat
+	devel:libopenblas$secondaryArchSuffix = $libVersionCompat
+	devel:libopenblasp_r$portVersion$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	openblas3$secondaryArchSuffix == $portVersion base
```
fbrosson:
```diff
 REQUIRES_devel="
-	openblas3$secondaryArchSuffix == $portVersion base
+	openblas$secondaryArchSuffix == $portVersion base
```
(I missed this one too :)

korli:
FWIW using "--update-dependencies" the build order is defined, haikuporter will rebuild a dependency in the queue before any depending ports, when the available dependency hpkg isn't anymore compatible with the "current" recipe.

fbrosson:
> Rename patchset

Don't forget to also rename the recipe:
`openblas3-0.3.1.recipe` :arrow_right: `openblas-0.3.1.recipe`

extrowerk:
Can you do it please, i am awk for this Weekend.

-- Z.

2018.07.27. dátummal, 17:30 időpontban fbrosson <notifications@github.com> írta:

fbrosson:
That's it.
Warning: I did **_not_** test the recipe.

extrowerk:
Builds, packages fine, tests are broken, no idea why.

PR #2608, opened 2018-05-27
hdf5: revbump needed after gcc bump from 5.4.0 to 7.3.0.

fbrosson:
hdf5 needs to be rebuilt against `libgfortran.so.4` (from `gcc 7.3.0`) instead of `libgfortran.so.3` (from `gcc 5.4.0`.)

### DIFF
--- a/sci-libs/hdf5/hdf5-1.10.1.recipe
+++ b/sci-libs/hdf5/hdf5-1.10.1.recipe
@@ -6,7 +6,7 @@ HOMEPAGE="http://www.hdfgroup.org/HDF5/"
 COPYRIGHT="2011-2016 The HDF Group,
 	The Board of Trustees of the University of Illinois."
 LICENSE="BSD (3-clause)"
-REVISION="1"
+REVISION="2"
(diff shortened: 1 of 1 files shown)


fbrosson:
It looks like hdf5 won't build with `gfortran 7.3.0`, so let's not rebuild it.

Users who have hdf5 and wish to upgrade to an hrev > 51972 will have to either uninstall hdf5 or give up the update.
I'm not opening an issue about this because I don't need hdf5 and don't have enough time to investigate.

alaviss:
Then shouldn't a version requirement for `gfortran` (`< 7.3`) be applied here? An issue should be opened to track packages that doesn't build with newer `gfortran`, so contributors can dive into and investigate.

fbrosson:
We could, indeed, add a version requirement, but I don't know if this would be the best way to handle this because gfortran `5.4.0` is no longer available from the binary repos.
I guess we should first try to patch hdf5 to make it build with gfortran `7.3.0`. If it turns out that such a patch would be too big or hard to maintain, then, yes, adding a build dependency on an old version of gfortran would be fine.

BTW, I think @extrowerk can tell us whether or not it would be hard to patch hdf5 to make it build with gfortran `7.3.0` ;)

PR #2611, opened 2018-05-27
bzip2: add debuginfo, install {bunzip2,bzcat} using symlinks

fbrosson:
... instead of making copies of bzip2.

Also, install `{bzcmp,bzegrep,bzfgrep,bzless}` using relative symlinks.

### DIFF
--- a/app-arch/bzip2/bzip2-1.0.6.recipe
+++ b/app-arch/bzip2/bzip2-1.0.6.recipe
@@ -8,9 +8,10 @@ similar to those of GNU Gzip."
 HOMEPAGE="http://www.bzip.org/"
 COPYRIGHT="1996-2010 Julian R Seward"
 LICENSE="bzip2"
-REVISION="6"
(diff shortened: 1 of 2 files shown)


fbrosson:
FWIW, _"Rebase and merge"_ **_always_** does the rebase, _even_ if it is _not_ required.

This means the _"Rebase and merge"_ method drops any GPG signature.
This also means that, _currenly,_ the only way to preserve a personal GPG signature embedded in a commit attached to a PR is to `git merge` that PR's branch to one's local master and then `git push` that local master to the remote repo.

alaviss:
I know, that's why I prefer merging from command line :)

Of course, it should be noted that rebasing should be done when appropriate to avoid merge commits, which are ugly and not allowed within HaikuPorts (the "Merge" button was disabled for this repo because of that).

Note that you can rebase locally to preserve _your_ GPG key (and **only** yours, others will be overrided automatically)

PR #2612, opened 2018-05-27
testdisk: select the libuuid from util-linux.

fbrosson:
Also update to the latest snapshot.

### DIFF
--- a/app-admin/testdisk/testdisk-7.1~git.recipe
+++ b/app-admin/testdisk/testdisk-7.1~git.recipe
@@ -26,61 +26,72 @@ non-booting drive which can then be sent to a tech for further analysis. \
 Those more familiar with such procedures should find TestDisk a handy \
 tool in performing onsite recovery."
 HOMEPAGE="https://www.cgsecurity.org/wiki/TestDisk"
-COPYRIGHT="1998-2008 Christophe GRENIER"
+COPYRIGHT="1998-2018 Christophe GRENIER"
 LICENSE="GNU GPL v2"
-REVISION="3"
-gitCommit="d0c0bf21ddadd60c7bb2188a66d39039d5cd2f9c"
-SOURCE_URI=" https://git.cgsecurity.org/cgit/testdisk/snapshot/testdisk-$gitCommit.tar.gz"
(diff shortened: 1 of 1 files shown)


app-admin/testdisk/testdisk-7.1~git.recipe
```diff
@@ -96,10 +107,9 @@ BUILD()
 {
 	export CXXFLAGS="-std=c++11"
 	autoreconf -fi
-	runConfigure --omit-dirs "binDir" ./configure \
-		--bindir=$prefix/bin \
+	runConfigure --omit-dirs binDir ./configure --bindir="$commandBinDir" \
 		--disable-stack-protector
-	make $jobArgs
+	make $jobArgs LDFLAGS="-lintl"
```
alaviss: Why is gettext being forcefully linked here? Also, for adding library to the linking process, please use the `LIBS` variable instead (ie. `LIBS="-lintl"`), as it will appends the said library to the end of linker flags, so that flags used by the build system are applied (ie. `--as-needed` requires libs to be passed after the switch)

The environment variable should be declared prior to `runConfigure`, so the build system can pick it up and process it accordingly.

fbrosson: I had an error about a symbol that could not be resolved, and I figured out that I could get rid of this error by just adding `-lintl`.
But there might very well be a better fix. (I'll try to investigate.)

app-admin/testdisk/testdisk-7.1~git.recipe
```diff
@@ -25,62 +25,73 @@ TestDisk can be used to collect detailed information about a \
 non-booting drive which can then be sent to a tech for further analysis. \
 Those more familiar with such procedures should find TestDisk a handy \
 tool in performing onsite recovery."
-HOMEPAGE="http://www.cgsecurity.org/wiki/TestDisk"
-COPYRIGHT="1998-2008 Christophe GRENIER"
+HOMEPAGE="https://www.cgsecurity.org/wiki/TestDisk"
```
fbrosson: For the record, the tiny change in 71d61a58ce32eb0f19963fa1ecc87e480bbf1be1 was already in this PR.
I've already fixed the conflict on my local branch, but I'll finish #2614 before updating this one.

PR #2614, opened 2018-05-27
util-linux: fix INSTALL when ALLOW_UNTESTED=no

fbrosson:
~util-linux: Do not define `*ARCHITECTURES_fdisk` as their content upsets the build servers.~

util-linux: fix `INSTALL` when `ALLOW_UNTESTED=no`

* While at it, do not export `CFLAGS` and `LDFLAGS`. Instead, pass `CPPFLAGS` and `LDFLAGS` to both `autogen.sh` and `configure`.
* Now that we don't override the project's default `CFLAGS` anymore we can add the debuginfo package.
    
**Warning:** If you call haikuporter after the recipe was last modified and then change `ALLOW_UNTESTED` to `yes`, the `.DependencyInfo` file for the fdisk subpackage will be missing.
Workaround: touch the recipe before calling haikuporter. Of course, this is required only once, that is, if you have recently changed `ALLOW_UNTESTED` to `yes`.

### DIFF
--- a/sys-apps/util-linux/util_linux-2.32.recipe
+++ b/sys-apps/util-linux/util_linux-2.32.recipe
@@ -6,7 +6,7 @@ LICENSE="GNU GPL v2
 	GNU LGPL v2
 	BSD (4-clause)
 	Public Domain"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://www.kernel.org/pub/linux/utils/util-linux/v2.32/util-linux-$portVersion.tar.xz"
 CHECKSUM_SHA256="6c7397abc764e32e8159c2e96042874a190303e77adceb4ac5bd502a272a4734"
 SOURCE_DIR="util-linux-$portVersion"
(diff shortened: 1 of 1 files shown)


fbrosson:
> util-linux: Do not define `{SECONDARY_,}ARCHITECTURES_fdisk` as it upsets the build servers.

@mmlr, do you know if this is true?
(In other words, shall I merge this PR?)

The build servers did not process 971393f641683409afffc1f9e669cd6a42de0539 nor 7c61af65208f3094f685b8e07d66bb669ce071b5, so there are no binary packages for `util-linux 2.32-1` in the repo.
(The build servers also did not build `bzip2-1.0.6-7`, bumped with 13818325f80881d4d4e847d1132229113596daf3.)

BTW, #2526 also needs libuuid and libblkid.

Thanks!

mmlr:
The buildmaster uses HaikuPorter, so it should not come to a different result than what you get locally. If it is buildable with HaikuPorter then the buildmaster and builders should be able to build it. It is however configured more strictly, i.e. no untested ports.

The reported error was ```packageEntries: error: "fdisk" doesn't seem to be a valid package suffix.``` This is a build time error, not something that happens at parse time. It indicates that the ```_fdisk``` subpackage was not properly defined by the recipe keys. In this case, the subpackage is probably discarded due to the architectures being marked untested. In that case, the build process obviously must not try to then package something into that subpackage, it has to do that conditionally for the architectures where it is available. Lifting the architecture restriction is the other way to solve it.

The disk of the buildmaster ran full, that's why it didn't continue processing new recipes. I've cleaned out some 40GiBs of obsolete packages I kept from the shared packages directories and it now has ample space again (and sprung into action again due to that).

fbrosson:
> The disk of the buildmaster ran full, that's why it didn't continue processing new recipes. I've cleaned out some 40GiBs of obsolete packages I kept from the shared packages directories and it now has ample space again (and sprung into action again due to that)

Cool, thanks for the info!

alaviss:
@fbrosson 

Your commit does a bit more that what the summary said, which might confuse some people when they do a `git blame` on the file. I suggest you split the commits next time, or use a summary like:

    util-linux: improvements and bug fixes

With that said, I should really stop asking contributors to fix recipe bugs whenever they bump one 😄, or at least ask them make separate commits for the changes, but that might be too much.

fbrosson:
Well, since the bug was only affecting those users who have `ALLOW_UNTESTED=no` I thought this information was more relevant in the commit summary than mentionning the addition of the debuginfo package.

Regarding the separate commits for fixing the bug and for adding the debuginfo package, since these two changes were orthogonal, I felt it was not worth making a separate commit just for adding the debuginfo.

(But I'll think about this a bit more from now on :)

PR #2616, opened 2018-05-28
parted: drop LDFLAGS, pass CPPFLAGS to runConfigure.

fbrosson:


fbrosson:
`parted`'s configure script has an `--enable-read-only` switch, but we're not using it because it's not very useful.
Should I add it and mark the recipe as tested?
Or is it fine as is, without the `--enable-read-only` ?
BTW, I did not dare to mark the recipe as tested because `parted` can erase data if used inappropriately.

sys-block/parted/parted-3.2.recipe
```diff
@@ -102,13 +102,12 @@ defineDebugInfoPackage parted$secondaryArchSuffix \
 BUILD()
 {
 	autoreconf
+	CPPFLAGS="-Wimplicit-fallthrough=0" \
+	LDFLAGS="-lblkid -luuid -lintl" \
```
alaviss:
```diff
-	LDFLAGS="-lblkid -luuid -lintl" \
+	LIBS="-lblkid -luuid -lintl" \
```

This still doesn't seem right however, the build system should correctly link these libraries as needed without having us to append it here.

fbrosson: Oops, I forgot to fix this.

alaviss:
> BTW, I did not dare to mark the recipe as tested because parted can erase data if used inappropriately.

That's the end-user problem, it will be our problem if parted erases data when it shouldn't

sys-block/parted/parted-3.2.recipe
alaviss: Can you check if it's possible to drop some `LIBS`? After all, they should not be necessary. The build system should detect whether they should be linked or not.

fbrosson: Good catch. (It looks like we don't need to specify `LIBS`.) Thanks!
I guess I added it because it was not working without it, but maybe the following change is what made it work without the `LIBS`:
```diff
-	cmd:gettext
+	cmd:gettext$secondaryArchSuffix
```

alaviss:
@fbrosson Can you file an issue for this on HaikuPorts? Thanks.

fbrosson:
I'd rather leave that to you or to another kernel developer :)

PR #2633, opened 2018-06-01
xapian: select the libuuid from util-linux, fix build for version 1.4.5.

fbrosson:


dev-libs/xapian/xapian-1.2.25.recipe
```diff
@@ -48,21 +47,29 @@ PROVIDES="
...
 
 PROVIDES_devel="
 	xapian${secondaryArchSuffix}_devel = $portVersion
 	devel:libxapian$secondaryArchSuffix = $libVersionCompat
+	cmd:xapian_config$secondaryArchSuffix = $portVersion
 	"
 REQUIRES_devel="
 	xapian$secondaryArchSuffix == $portVersion base
 	"
+if [ -n "$secondaryArchSuffix" -a "$commandBinDir" != "$prefix"/bin ]; then
```
alaviss:
```diff
-if [ -n "$secondaryArchSuffix" -a "$commandBinDir" != "$prefix"/bin ]; then
+if [ -n "$secondaryArchSuffix" -a "$commandBinDir" == "$binDir" ]; then
```

The less hardcoding, the better :)

dev-libs/xapian/xapian-1.2.25.recipe
```diff
@@ -48,21 +47,29 @@ PROVIDES="
...
+CONFLICTS_devel="
+	xapian_devel
```
alaviss: Sorry, but I don't understand this. Why would `xapian${secondaryArchSuffix}_devel` conflicts with `xapian_devel`?

dev-libs/xapian/xapian-1.2.25.recipe
```diff
@@ -74,12 +81,17 @@ BUILD_PREREQUIRES="
 	cmd:sed
 	"
 
-defineDebugInfoPackage xapian$secondaryArchSuffix \
-	"$libDir"/libxapian.so.$libVersion
+debugList=("$libDir"/libxapian.so.$libVersion)
+for i in `echo "$PROVIDES" | sed -n -e \
+		"s/^[\t]*cmd:\(.*\)$commandSuffix =.*/\1/p;" | sed -e "s/_/-/g"`; do
```
alaviss: I'm not really a fan of using non-builtin when dealing with metadata though, as you can't be sure that external cmds exist, but I think it's fine here.

dev-libs/xapian/xapian-1.2.25.recipe
fbrosson: Yeah, I do agree that `"$prefix"/bin` is a hardcoded path, but since we already have it earlier, using `[ -n "$secondaryArchSuffix" -a "$commandBinDir" != "$prefix"/bin ]` instead of `[ -n "$secondaryArchSuffix" -a "$commandBinDir" = "$binDir" ]` makes the test easier to translate to english.
The sentence I would translate this to is:
_Are we building for a secondary arch and are we installing on something other than `"$prefix"/bin`?_

After all, we want a conflicts _only_ if these two conditions are both met:
* we are building for a secondary arch
* we did not change the default definition for `commandBinDir`.

dev-libs/xapian/xapian-1.2.25.recipe
fbrosson: This conflicts is needed because we have moved `$libDir/cmake/xapian/*` to `$dataDir/cmake/Modules/xapian/*`.
This is not an issue on x86 secondary arch **_if_** the primary arch is x86_gcc2.

But since x86 secondary arch will become available for a x86_64 hybrid, we need to make sure that, on that hybrid arch, `xapian_x86_devel` and `xapian_devel` will never be activated together.
That's because these packages have overlapping `$dataDir/cmake/Modules/xapian/*` files which have conflicting contents.

dev-libs/xapian/xapian-1.2.25.recipe
fbrosson: After all, since I had to fix a typo, I switched the test to use your variant :)

dev-libs/xapian/xapian-1.2.25.recipe
fbrosson: Yeah, I also try to avoid unusual blocs like this one, but here, as you already noticed, its fine because there are a lot of `cmd:` items, so duplicating them here would be worse than having this tiny block.

dev-libs/xapian/xapian-1.2.25.recipe
```diff
@@ -89,18 +101,31 @@ INSTALL()
...
+	fi
+
 	prepareInstalledDevelLib libxapian
 	fixPkgconfig
 
-	mkdir -p $dataDir/cmake/Modules
-	mv $libDir/cmake/xapian $dataDir/cmake/Modules/xapian
+	# move $libDir/cmake/xapian/* to $dataDir/cmake/Modules/xapian/*
+	mkdir -p "$dataDir"/cmake
+	mv "$libDir"/cmake "$dataDir"/cmake/Modules
```
alaviss: I've only noticed this just now, but why was this moved?

fbrosson: I don't know.
The recipe was created in 1439375f3ae420f8eb3a6ce82b0373db9d487ee4 with this move:
https://github.com/haikuports/haikuports/blob/1439375f3ae420f8eb3a6ce82b0373db9d487ee4/dev-libs/xapian/xapian-1.2.21.recipe#L82

If you think we should keep it in `$libDir` and if there are no downsides to this, then that's OK for me. 
(I can't change it myself because I'm not a cmake expert, so I might miss something.)

PR #2637, opened 2018-06-02
botan, bump version

Begasus:


dev-libs/botan/botan-2.6.0.recipe
```diff
@@ -0,0 +1,93 @@
...
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python
+	"
+
+PATCH()
+{
+	sed -i -e "/^all\>/ s/-pthread -fstack-protector/-fno-stack-protector/" src/build-data/cc/gcc.txt
```
alaviss: I think we only need to remove `-pthread` now, as the stack protector is fixed.

dev-libs/botan/botan-2.6.0.recipe
```diff
@@ -0,0 +1,93 @@
+SUMMARY="A cryptography library written in C++"
+DESCRIPTION="This is the best option for crypthography in C++"
+HOMEPAGE="https://botan.randombit.net/"
+COPYRIGHT="1999-2018 The Botan Authors"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/randombit/botan/archive/$portVersion.tar.gz"
```
fbrosson: Suggestion:
```diff
-SOURCE_URI="https://github.com/randombit/botan/archive/$portVersion.tar.gz"
+SOURCE_URI="https://botan.randombit.net/releases/Botan-$portVersion.tgz"
+SOURCE_DIR="Botan-$portVersion"
```
Differences:
* The `SOURCE_DIR` is:
  * `botan-2.6.0` for the tarball on GitHub,
  * `Botan-2.6.0` for the tarball [advertised](https://botan.randombit.net/#release-downloads) on the project's site,

* The contents are identical for all the files, except for `src/build-data/version.txt`:
```diff
diff -u botan-2.6.0/src/build-data/version.txt  Botan-2.6.0/src/build-data/version.txt
--- botan-2.6.0/src/build-data/version.txt	2018-04-10 13:17:29.000000000 +0000
+++ Botan-2.6.0/src/build-data/version.txt	2018-04-10 09:17:29.000000000 +0000
@@ -5,6 +5,6 @@
 release_so_abi_rev = 5
 
 # These are set by the distribution script
-release_vc_rev = None
-release_datestamp = 0
-release_type = 'unreleased'
+release_vc_rev = 'git:d7d080992372cf4fbd569cce1d8cd6aa7599fa0d'
+release_datestamp = 20180410
+release_type = 'release'
```
BTW, if we keep the `SOURCE_URI` (i.e. using the tarball from GitHub) then we would have to add a `SOURCE_FILENAME="botan-$portVersion.tar.gz"`, whereas with the tarball from the project's site we don't need to add that variable. (Well, we need to add `SOURCE_DIR`, but imo having a `release_type = 'release'`in `src/build-data/version.txt` is better than having a `release_type = 'unreleased'` :)

dev-libs/botan/botan-2.6.0.recipe
```diff
@@ -0,0 +1,93 @@
+SUMMARY="A cryptography library written in C++"
+DESCRIPTION="This is the best option for crypthography in C++"
```
pulkomandy: I would prefer a more neutral description.

dev-libs/botan/botan-2.6.0.recipe
Begasus: Need to check this, not a big sed guru ;)

dev-libs/botan/botan-2.6.0.recipe
Begasus: I'll update with the given comments :) (ps, most of them were left overs from previous version) :)

dev-libs/botan/botan-2.6.0.recipe
fbrosson: It builds fine with
```
	sed -i -e "/^all\>/ s/-pthread//" src/build-data/cc/gcc.txt
```
but maybe we can even drop the `PATCH()` function completely.

dev-libs/botan/botan-2.6.0.recipe
fbrosson: Cool!
Yep :)
BTW, shall we drop the previous recipe?

dev-libs/botan/botan-2.6.0.recipe
Begasus: I think they already included it in the source "all!haiku -> "-pthread"", could be this patch isn't needed anymore ...

Begasus: Should this still be needed to use sed @fbrosson ?

dev-libs/botan/botan-2.6.0.recipe
Begasus: Once this one is ok I'm fine with it :)

dev-libs/botan/botan-2.6.0.recipe
fbrosson: (I've commented out the `PATCH()` function to make sure we can drop the `sed` command (and the whole `PATCH()` function.)

Begasus: Same here, build is still running though ...

Begasus: Looks like the PATCH() function isn't needed anymore, untill someone steps up I've removed it from the recipe, builds ok for secondaryarch (didn't check x86_64 yet so best someone would check this before merging) :)

fbrosson: I kicked 3 parallel builds: botan_x86 on x86_gcc2, and botan on x86 and x86_64.
When I try to test any of these builds I get:
```
Testing ...
runtime_loader: Cannot open file libbotan-2.so.5: No such file or directory
Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 3
Error: Test has failed - stopping.
```
We can avoid this error with this change:
```diff
 TEST()
 {
+	LIBRARY_PATH="$sourceDir${LIBRARY_PATH:+:$LIBRARY_PATH}" \
 	./botan-test
 }
```
Unfortunately, we now get a crash with it:
```
Testing ...
Testing Botan 2.6.0 (release, dated 20180410, revision git:d7d080992372cf4fbd569cce1d8cd6aa7599fa0d, distribution unspecified)
Starting tests drbg_seed:1534AB7A5ABD1248
block:
/port.recipe: line 101: 28865 Kill Thread
```
Well, anyway, the suggested change in `TEST` (i.e. inserting the line which changes `LIBRARY_PATH`) is a good thing to do.

Begasus: With the suggestion you made it seems like the tests are running ok here ... 
'''
Testing Botan 2.6.0 (release, dated 20180410, revision git:d7d080992372cf4fbd569cce1d8cd6aa7599fa0d, distribution unspecified)
  Starting tests drbg_seed:1534AC7A982411A8
  block:
  AES-128 ran 11580 tests in 157.09 msec all ok
  AES-192 ran 13500 tests in 140.75 msec all ok
  AES-256 ran 15420 tests in 172.17 msec all ok
'''

Begasus: Ending in ... 
```
Entropy source system_rng ran 9 tests in 370.67 msec 1 FAILED
  Failure 1: Entropy source system_rng lzma exception while compressing with error lzma memory allocation failed ...

 Tests complete ran 2621414 tests in 511.63 sec 1 tests failed
  Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 1
  Error: Test has failed - stopping
```
Maybe libzma is missing for TEST?

fbrosson: Good idea. We might solve this with:
```diff
 	cmd:python
 	"
 
+TEST_REQUIRES="
+	devel:liblzma$secondaryArchSuffix
+	"
+
 BUILD()
 {
```

Begasus: No luck there, still the same ...

Begasus: Well, you are the developer ;)

Begasus:
Pushed as far as I got .... test still fails :/

dev-libs/botan/botan-2.6.0.recipe
```diff
@@ -60,16 +66,11 @@ BUILD_PREREQUIRES="
...
 	"
 
-PATCH()
-{
-	sed -i -e "/^all\>/ s/-pthread -fstack-protector/-fno-stack-protector/" src/build-data/cc/gcc.txt
-}
-
 BUILD()
 {
 	libtoolize --force --copy --install
```
alaviss: It doesn't look like they're using GNU autohell anymore. So I think we can remove this line.

Begasus: @alaviss could you push this into this PR? It looks like it's not needed any more, but my local branch is out of sync with this one :)
PS, test still fail on Entropy ... other than that it builds fine

PR #2653, opened 2018-06-06
libzmf: enable secondary arch, switch SOURCE_URI to tar.xz.

fbrosson:


fbrosson:
Builds fine on x86 second arch.
I'll fix `COPYRIGHT` before merging.

Issue #2657, opened 2018-06-06
libreoffice: x86_64 package missing dependencies


Briseur:
It won't start without installing libboost 165 and there's an error message complaining about it but syslog also report missing OpenCL, It complains about some vclplugin too but I guess that it isn't ported yet..
libboost 165 package name is quite obvious and so with a little effort average user can fix that but it isn't the case for OpenCL and there's no error message about it only a line in syslog.

kallisti5:
confirmed here.  boost163 comes pre-installed. It appears the pre-installed boost163 is enough to fill the dependency for libboost, however it should be locked to boost165

diversys:
Should be fixed in 19971303a0fc8c4f47793cb0c9e335105996b3c1.

PR #2661, opened 2018-06-07
LibreOffice: fix build on secondary arch. Minor changes.

fbrosson:
* Depend on `libboost* >= 1.65.0` as suggested by @kallisti5 in #2657.
* Add missing `devel:liboost_*` to `BUILD_REQUIRES`.
* Do not use `--with-distro=LibreOfficeHaiku` but, instead, copy upstream's `distro-configs/LibreOfficeHaiku.conf` to the recipe.
* Add `--with-system-libepubgen` to the config options.

### DIFF
--- a/app-office/libreoffice/libreoffice-6.2.0.0~git.recipe
+++ b/app-office/libreoffice/libreoffice-6.2.0.0~git.recipe
@@ -17,7 +17,7 @@ and Open Source office suite on the market:
 HOMEPAGE="https://www.libreoffice.org/"
 COPYRIGHT="2000-2018 LibreOffice contributors"
 LICENSE="MPL v2.0"
-REVISION="2"
+REVISION="3"
 COMMIT="100b6a229e0ab9888578c138cd38424d16dec608"
 SOURCE_URI="https://github.com/LibreOffice/core/archive/$COMMIT.tar.gz"
 CHECKSUM_SHA256="21639e6388bdda4d4aca82e3957e44d6fe1b20fb36e32ec9d64394be84b552c6"
@@ -28,6 +28,13 @@ ADDITIONAL_FILES="libreoffice.rdef.in"
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+fi
+
+boostMinimumVersion=1.65.0
+
 PROVIDES="
 	libreoffice$secondaryArchSuffix = $portVersion
 	app:LibreOffice$secondaryArchSuffix = $portVersion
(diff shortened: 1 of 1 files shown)


fbrosson:
I have dropped the `--with-distro=LibreOfficeHaiku` switch as [suggested](https://github.com/haikuports/haikuports/commit/2d5e2e29f39eed7004c95debab7abfe4f51c3994#commitcomment-29282161) by @KapiX but if you prefer to restore it and only add those options (e.g. `--with-system-libepubgen`) which are not already in `distro-configs/LibreOfficeHaiku.conf` then I can update the PR.

Imo, although the list of options is quite big, it can be nice to have it in the recipe in order to make it easier to track future changes.

KapiX:
Change the vendor. We're not TDF.

czw., 7 cze 2018 o 17:03 fbrosson <notifications@github.com> napisał(a):

app-office/libreoffice/libreoffice-6.2.0.0~git.recipe
```diff
@@ -189,10 +202,79 @@ BUILD()
...
+		--disable-gio \
+		--disable-randr \
+		--disable-gstreamer-0-10 \
+		--disable-cups \
+		--disable-ccache \
+		--disable-postgresql-sdbc \
+		--disable-lotuswordpro \
+		--disable-firebird-sdbc \
+		\
+		--with-vendor="The Document Foundation" \
```
diversys: Should be --with-vendor="HaikuPorts" as this is what porduces the binary.

fbrosson:
Should I sort alphabetically the options passed to `runConfigure ./autogen.sh`, well, at least a little bit?

BTW, I did not kick a build yet on my system. Shall I do a "blind"  merge?

fbrosson:
Oops, you were faster than me. I was changing the commit message to:
```
* Depend on boost >= 1.65.0 as suggested by @kallisti5 in #2657.
* Add missing devel:lib{cairo,boost_*} to BUILD_REQUIRES.
* Do not use --with-distro=LibreOfficeHaiku but, instead, copy
  upstream's distro-configs/LibreOfficeHaiku.conf to the recipe.
* Add --with-system-libepubgen to config the options.
```
But that's OK.

extrowerk:
6 hours? Omg!
I'm thiking about how the programs will evolve during my life....
We should start to think about some incremental build system wich would take the latest build and recompile just the diff.

I know about make, i mean something different here: packages should know what was the last git commit, and at a new build it should check and rebuild just the absolut minimal changes, then replace the affected parts.
No idea, how realistic this btw

pulkomandy:
On Thu, Jun 07, 2018 at 08:58:25AM -0700, miqlas wrote:
> 6 hours? Omg!
> I'm thiking about how the programs will evolve during my life....
> We should start to think about some incremental build system wich would take the latest build and recompile just the diff.
> 
> I know about make, i mean something different here: packages should know what was the last git commit, and at a new build it should check and rebuild just the absolut minimal changes, then replace the affected parts.
> No idea, how realistic this btw

This is exactly what "make" does, but we force a clean build on purpose
because quite often make misses something. A clean build is
unfortunatele necessary for a reproductible packaging.

extrowerk:
I know how it iscurrently working, i'm just brainatorming about alternative ways.

-- Z.

2018.06.07. dátummal, 19:10 időpontban Adrien Destugues <notifications@github.com> írta:

alaviss:
@extrowerk 

Use ccache. I don't know if we ported it yet though.

PR #2677, opened 2018-06-10
camlp5, bump version

Begasus:


Begasus:
Looks like it should be ok, build is starting fine, but my system here keeps freezing when trying to build (not related to Haiku), if someone could check this out it would be fine :)

alaviss:
> Looks like it should be ok, build is starting fine, but my system here keeps freezing when trying to build (not related to Haiku), if someone could check this out it would be fine :)

Chances are we have to update our OCaml. I once tried, but was hit with too many issues so I abandoned it.

Begasus:
The build didn't produce any errors so far here ... just that my laptop doesn't seem to cope with the build ...

Begasus:
Finaly my system went through all the way, builds ok for x86_gcc2 (haven't checked others yet) :)

PR #2690, opened 2018-06-16
unrtf, lcms, libopenmpt, mpg321, dvdauthor: add commandSuffix.

fbrosson:
Adding `commandSuffix` is neutral for all architectures, except for x86 secondary arch on a x86_64/x86 hybrid system, which is not yet available, so we can keep `REVISION` unchanged.


fbrosson:
It is OK as long as x86 is not available as a secondary arch for x86_64.
But once this becomes available, not having the `$commandSuffix` would break `haikuporter -S mpg321_x86` on x86_64.

PR #2692, opened 2018-06-16
lighttpd: fix PROVIDES*, add debuginfo & commandSuffix.

fbrosson:
* The `lib:mod_*` in `PROVIDES*` were missing a `$secondaryArchSuffix`.
* Use `"$(getPackagePrefix mod_geoip)"` in `defineDebugInfoPackage`.

* Do not use getPackagePrefix for now, because `getPackagePrefix` makes our scripts on travis-ci.org complain:
```
Checking ports of: lighttpd-1.4.49
lighttpd-1.4.49 [www-servers]
packageEntries: error: "mod_geoip" doesn't seem to be
a valid package suffix.
```
So, instead, use a little hack.

### DIFF
--- a/www-servers/lighttpd/lighttpd-1.4.49.recipe
+++ b/www-servers/lighttpd/lighttpd-1.4.49.recipe
@@ -9,7 +9,7 @@ solution for every server that is suffering load problems."
 HOMEPAGE="https://www.lighttpd.net/"
 COPYRIGHT="2003-2018 lighttpd authors"
 LICENSE="BSD (3-clause)"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-$portVersion.tar.xz"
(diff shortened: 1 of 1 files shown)


fbrosson:
Maybe, but I did not find anything better :(

BTW, I guess you meant `getPackagePrefix` instead of `get_package_prefix`.

fbrosson:
Without the second commit we would have these warnings everytime travis examines **_any_** PR, as we can see with the [previous report](https://travis-ci.org/haikuports/haikuports/builds/393000942?utm_source=github_status&utm_medium=notification).
Hopefully, after the second commit, we can see that the [last report](https://travis-ci.org/haikuports/haikuports/builds/393003181?utm_source=github_status&utm_medium=notification) only shows the warnings about the `"$(getPackagePrefix fortran)"` and `"$(getPackagePrefix syslibs)"` we have in 
https://github.com/haikuports/haikuports/blob/3fc8d3030fc21190f1291f50ccd27a4160a906a4/sys-devel/gcc/gcc-7.3.0_2018_05_01.recipe#L164

PR #2695, opened 2018-06-16
TBB: bump version

extrowerk:


dev-cpp/tbb/tbb-2018.4.recipe
```diff
@@ -8,7 +8,7 @@ COPYRIGHT="2005-2018 Intel Corporation"
 LICENSE="Apache v2"
 REVISION="2"
```
fbrosson:
```diff
-REVISION="2"
+REVISION="1"
```
(Shall we reset `REVISION`?)

PR #2700, opened 2018-06-17
gsl: bump to 2.5, fix PROVIDES{,_devel}, add commandBinDir.

fbrosson:
`{lib,devel}:libgsl` had a wrong compat, `>= 0` instead of `>= 23`, so we will need to rebuild calligra, krita and xaos.

* calligra: revbump required after the change in gsl's `PROVIDES`.
* krita: revbump required after the change in gsl's `PROVIDES`.
* xaos: revbump required after the change in gsl's `PROVIDES`.


fbrosson:
It looks like it was a very bad idea to push the revbumps of calligra, krita and xaos immediately after the commit for gsl.
I though the build servers would examine each commit and pipe the jobs in the same order. But I guess implementing this would have downsides.
What I should have done was:
1) push the commit for gsl-2.5 only.
2) wait until the build servers have built it.
3) run `pkgman update gsl` and see if revbumping calligra, krita and xaos was really required.
4) revbump those recipes which would no longer work with the new release of gsl.

Here it seems I made 2 mistakes:
1) I assumed the revbumps were required, but a `package list -i /system/packages/calligra-3.1.0-1-x86_64.hpkg` showed that it depended on `lib:libgsl>=23.0.0` (and not on `lib:libgsl>=0` as I thought it would, given gsl's incorrect `PROVIDES*`) so I guess revbumping calligra was not needed.
2) calligra is not building because it can't find `libflake.so`. From http://kitchen.haiku-os.org/master/x86_64/buildruns/1569/builds/10929.log:
```
[ 20%] Building CXX object libs/flake/CMakeFiles/flake.dir/flake_autogen/mocs_compilation.cpp.o
[ 20%] Linking CXX shared library libflake.so
[ 20%] Built target flake
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 2
```
Conclusion: I wish I had not pushed any revbump...

PR #2703, opened 2018-06-17
GCC: enable ObjC

extrowerk:


extrowerk:
It should be ok now, builds, packages fine.

extrowerk:
It can compile a helloworld just fine, but running the generated binary results "segmentation violation"
http://chunk.io/miqlas/f9d316c6ffa842a490e1203c7b9f87cf

extrowerk:
But ObjFW built correctly and all test passes.

jessicah:
As mentioned on IRC, I don't think we should ship libobjc as part of the package, since a) not recommended to use, and b) don't actually work on Haiku. Shipping broken libraries is just silly.

alaviss:
@jessicah 

We can only hope that no one uses libobjc in that case. But sadly, tutorials showing how to compile objc code with gcc often passes `-lobjc`, so we should support that.

I found this library [gnustep/libobjc2](https://github.com/gnustep/libobjc2) that claims to be a drop-in replacement for gcc's `libobjc`, so with a couple symlinks it should be fine

PR #2705, opened 2018-06-18
libusbmuxd: use autoreconf, add debuginfo & commandSuffix.

fbrosson:
* `BUILD` was no longer working. We could have replaced `automake` by `automake --add-missing` but it was easier to switch to `autoreconf`.
* Switch `HOMEPAGE` and `SOURCE_URI` to https.

### DIFF
--- a/app-pda/libusbmuxd/libusbmuxd-1.0.10.recipe
+++ b/app-pda/libusbmuxd/libusbmuxd-1.0.10.recipe
@@ -1,29 +1,33 @@
 SUMMARY="A client library to multiplex connections from and to iOS devices"
 DESCRIPTION="libusbmuxd is a client library to multiplex connections from and \
 to iOS devices by connecting to a socket provided by a usbmuxd daemon."
-HOMEPAGE="http://www.libimobiledevice.org/"
+HOMEPAGE="https://www.libimobiledevice.org/"
 COPYRIGHT="2013-2014 Nikias Bassen
 	2009-2014 Martin Szulecki"
 LICENSE="GNU GPL v2
 	GNU LGPL v2.1"
 REVISION="2"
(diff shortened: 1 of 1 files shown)


app-pda/libusbmuxd/libusbmuxd-1.0.10.recipe
```diff
@@ -1,29 +1,33 @@
...
 
+libVersion="4.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
 PROVIDES="
 	libusbmuxd$secondaryArchSuffix = $portVersion compat >= 1
-	lib:libusbmuxd$secondaryArchSuffix = 4.0.0 compat >= 4
-	cmd:iproxy = $portVersion compat >= 1
+	lib:libusbmuxd$secondaryArchSuffix = $libVersionCompat
+	cmd:iproxy$commandSuffix = $portVersion compat >= 1
```
alaviss: Shouldn't `cmd` be placed before `lib`?

fbrosson: Yeah, we try to, but I guess we tend to be less strict here, because we have several recipes for which we only add the `cmd:` entries when we're not building for a secondary arch. So, for these recipes we need to start with the `lib:` entries.
Well, here it's not the case, so I believe the only reason we have put `lib:libusbmuxd` before the `cmd:iproxy` was that the `lib:libusbmuxd` is "more" important than the `cmd:iproxy`.

I think this is OK to keep this as is, and only be strict in new recipes.
That said, we can always swap the `lib:` and the `cmd:` the next time we update this recipe.
BTW, I should have done so when I added the `$commandSuffix`...

PR #2706, opened 2018-06-18
isl, bump version

Begasus:


dev-libs/isl/isl-0.19.recipe
```diff
@@ -0,0 +1,89 @@
...
+
+PROVIDES_devel="
+	isl${secondaryArchSuffix}_devel = $portVersion
+	devel:libisl$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	isl$secondaryArchSuffix == $portVersion base
+	"
+CONFLICTS_devel="
+	isl13${secondaryArchSuffix}_devel
```
fbrosson: @Begasus, I don't know if there is a strict rule for this, but my understanding is that we prefer to put a `CONFLICTS isl` in isl13's recipe, instead of a `CONFLICTS isl13` in isl's recipe.
(That's what I have learned from @korli.)
So, I think we should remove the `CONFLICTS` in `dev-libs/isl/isl-0.19.recipe` and, instead, add this one in `isl13-0.14.recipe`:
```
CONFLICTS_devel="
	isl${secondaryArchSuffix}_devel
	"
```
What's nice about this is that a single `CONFLICTS` can handle everything, if put in the right recipe.

Begasus: I already thought it looked strange in here, will remove it when I get around to it :)

Begasus: Why do we even have isl13 (for version 0.14)?

Begasus: Ok, but is anything relying on soname 13 for isl? The only thing I see using isl directly in the recipe is cloog (which builds fine with latest isl version) ...

fbrosson: Good catch.
In that case we can probably remove `dev-libs/isl/isl13-0.14.recipe`.
(I guess anyone building anything outside haikuports and linking against `libisl.so.13` provided by `isl13-0.14.recipe` would be able to build against any newer libisl.)

Begasus: I'll keep it in for now, until someone comes up with a reason to keep it in :)

PR #2709, opened 2018-06-19
vba_next_libretro: VBA-M fork optimized for libretro

kwyxz:
This is a libretro-optimized fork of VisualBoyAdvance-M, a Nintendo Game Boy Advance emulator.


games-emulation/vba_next_libretro/vba_next_libretro-1.0.2_20180603.recipe
```diff
@@ -0,0 +1,52 @@
+SUMMARY="An optimized VisualBoyAdvance-M core for the libretro API"
+DESCRIPTION="VBA-Next is a fork of VBA-M, itself a fork of VisualBoyAdvance, \
+a popular Nintendo Game Boy Advance emulator. VBA Next was developed by the \
+libretro team as an optimized version of VBA-M."
+HOMEPAGE="https://github.com/libretro/vba-next/"
```
fbrosson: Thanks @kwyxz for this PR!

I haven't built the recipe yet (but will do so in a few minutes).
Everything looks OK, except that it would be slightly better to drop the trailing slash here:
```diff
-HOMEPAGE="https://github.com/libretro/vba-next/"
+HOMEPAGE="https://github.com/libretro/vba-next"
```
That's because without the trailing we don't get a "302 redirect", so this means the slash is not useful.
Of course, for other sites such as SF, when we get a "302 redirect" with a slash appended, then yes, we like to have the trailing slash:)

(I can push this tiny change to your PR if you wish.)

kwyxz:
Please do!

-- 
Apologies for the top posting and the horrible quoting : sent from an iPhone.

games-emulation/vba_next_libretro/vba_next_libretro-1.0.2_20180603.recipe
fbrosson: > That's because without the trailing we don't get a "302 redirect", so this means the slash is not useful.

What I meant here is: `curl --head https://github.com/libretro/vba-next` does not send a redirect, so we can drop the trailing slash we had in the previous definition of `HOMEPAGE`, i.e. before my tiny update :)

PR #2713, opened 2018-06-19
Perl: File::Next

extrowerk:


waddlesplash:
This can be installed via CPAN, yes? So why make a recipe for it, unless something in HaikuPorts requires it?

extrowerk:
@waddlesplash : ack needs it.

PR #2725, opened 2018-06-23
audiofile, fix build

Begasus:


Begasus:
@diversys thanks for the pointer to the gentoo patch :+1:

media-libs/audiofile/audiofile-0.3.6.recipe
```diff
@@ -30,9 +30,10 @@ Supported compression formats:
...
 COPYRIGHT="1998-2013 Michael Pruett, Chris Pirazzi, Scott Porter, Doug Scott"
 LICENSE="GNU LGPL v2.1"
-REVISION="4"
+REVISION="5"
 SOURCE_URI="https://audiofile.68k.org/audiofile-0.3.6.tar.gz"
 CHECKSUM_SHA256="cdc60df19ab08bfe55344395739bb08f50fc15c92da3962fac334d3bff116965"
+PATCHES="audiofile-$portVersion.patchset"
 
 ARCHITECTURES="x86 x86_64"
 SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
fbrosson:
```diff
-PATCHES="audiofile-$portVersion.patchset"
+
+SOURCE_FILENAME_2="audiofile-0.3.6-gcc6-build-fixes.patch"
+srcGitRev_2="1709d2d431a493deea7765902011730e04fab466"
+SOURCE_URI_2="https://gitweb.gentoo.org/repo/gentoo.git/plain/media-libs/audiofile/files/$SOURCE_FILENAME_2?id=$srcGitRev_2#noarchive"
+CHECKSUM_SHA256_2="c088e405ae54bc560e7d203cc5a18edb829b555e675c363d789b0f3f2b4d5d0c"
 
-ARCHITECTURES="x86 x86_64"
-SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```

```diff
 
+PATCH()
+{
+	patch -p1 -i "$sourceDir2"/$SOURCE_FILENAME_2
+}
+
 BUILD()
```

It works like a charm.
Since I've already edited the recipe, I can update your PR :)

fbrosson:
Sorry @Begasus for having updated your PR without waiting for your reply. I don't like being so directive.
But I'm sure you are OK with this ;)

Regarding the small diff between your patchset and Gentoo's patch, here are the resulting differences:
```diff
--- HaikuPorts-imported-patch/audiofile-0.3.6/test/NeXT.cpp
+++ Gentoo-external-patch/audiofile-0.3.6/test/NeXT.cpp
@@ -37,7 +37,7 @@
 
 #include "TestUtilities.h"
 
-const char kDataUnspecifiedLength[] =
+const signed char kDataUnspecifiedLength[] =
 {
 	'.', 's', 'n', 'd',
 	0, 0, 0, 24, // offset of 24 bytes
```

Hmm, we should also fix a small error in `BUILD_PREREQUIRES`:
```diff
-	cmd:gcc$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
```

(I'll do that, since I've already made 2 updates...)
Thanks!

P.S. Imho, we should always use `SOURCE_URI_x` to fetch patches from e.g. Gentoo, because this adds less noise to our repo, and it is also easier to update on our side, should Gentoo update their patches.

fbrosson:
Oops, I'm sorry, I just noticed that the difference was only in `audiofile-0.3.6/test/NeXT.cpp`!
Well, anyway, using an external patch is better than importing a copy of it.

pulkomandy:
Actually, are we sure that:
- Gentoo does not remove old patches?
- Gentoo does not replace old patches with new ones at the same URL?

Well, in either case our build will fail (missing source or SHA256
mismatch). But I don't know if relying on them is a good idea.

If the patch is needed both by Gentoo and Haiku, it may be a good idea
to upstream it anyways!

fbrosson:
> Actually, are we sure that:
> - Gentoo does not remove old patches?
> - Gentoo does not replace old patches with new ones at the same URL? 

Yep, because Gentoo uses git, and because we're using the `?id=` parameter (in our `SOURCE_URI_2`) with the reference for their commit. So, even if they update their patch, we won't have any checksum mismatch because we will continue to point to the version that Gentoo is currently using.

And in case we ever need to resync with a newer version of their patch, all we need to do is update `srcGitRev_2` and `CHECKSUM_SHA256_2`.

fbrosson:
@Begasus, @pulkomandy, shall we merge this?
(Of course, I'll use the "Squash and merge" method, so that we will have only one commit, authored by @Begasus.)

Begasus:
I'm ok with this @fbrosson if it's ok with @pulkomandy

fbrosson:
@pulkomandy:
> If the patch is needed both by Gentoo and Haiku, it may be a good idea to upstream it anyways!

Gentoo has probably already done so.

@Begasus: Thanks!

@pulkomandy: shall we merge?

fbrosson:
<Details>

https://github.com/haikuports/haikuports/blob/d25489d6d4cb82f4ef99be584364838a08a5c471/app-shells/dash/dash-0.5.10.2.recipe#L17
https://github.com/haikuports/haikuports/blob/d25489d6d4cb82f4ef99be584364838a08a5c471/app-shells/dash/dash-0.5.10.2.recipe#L22
https://github.com/haikuports/haikuports/blob/de8f825da846a5b11d68cad8c215feeeaa201b1a/app-text/hunspell/hunspell-1.4.2.recipe#L19
https://github.com/haikuports/haikuports/blob/de8f825da846a5b11d68cad8c215feeeaa201b1a/app-text/hunspell/hunspell-1.5.4.recipe#L20
https://github.com/haikuports/haikuports/blob/1adfc3b06807a94519a4853169aa8cc2205b3ace/dev-libs/geoip/geoip-1.6.12.recipe#L20
https://github.com/haikuports/haikuports/blob/8a4e5324b827a4d574817a2ccbf7dadb1c824e0f/dev-libs/libmspack/libmspack-0.6%7Ealpha.recipe#L25
https://github.com/haikuports/haikuports/blob/fc0936b04009e713f9bf551851781c83f9f23871/media-gfx/png2ico/png2ico-2002.12.08.recipe#L12
https://github.com/haikuports/haikuports/blob/7d01b6e4997f17c66dae2cb416ef4eeb1d4fb2ac/media-libs/tiff/tiff-3.9.7.recipe#L16
https://github.com/haikuports/haikuports/blob/822f5d01afbda2eb7e8ec0a34255a4c60244c20d/sys-fs/squashfs_tools/squashfs_tools-4.3.recipe#L23

</Details>

FWIW, we already have several recipes which download and apply patches from Gentoo, as shown in the above _Details_ bloc.

@pulkomandy, do we have your green lights?

@Begasus: For the commit message, shall we use `audiofile: fix build, using Gentoo's patch.` (or any alternative of your choice)?

fbrosson:
Yep, of course.
Actually, when we use the "Squash and merge" method, we can edit the commit message "on the fly" (i.e. after clicking on "Squash and merge" but before clicking on "Confirm merge"), without even having to update the PR.
(So I'll do that, of course, as well as removing the unneeded text on the "detailed commit message" window, to get rid of the commit messages from my updates.)

fbrosson:
No problem :)
@pulkomandy will come back sooner or later, so let's wait for his opinion.

pulkomandy:
yes it's ok, I was just curious if this was safe.

alaviss:
Usually I would prefer using `git am` to apply the patch, then extract it out as a patchset. That way, the patch authorship is kept, and the recipe won't be cluttered by additional variables.

But I'm fine with either way.

fbrosson:
When I need to update a recipe, say "foobar", which already has a patch in our HaikuPorts tree, from release 1.2.3 to 1.2.4 what I do is:
```
cd patches
cp foobar-1.2.3.patchset foobar-1.2.4.patchset
hp -c foobar-1.2.4
hp -B foobar-1.2.4
hp -e foobar-1.2.4
emacs foobar-1.2.3.patchset foobar-1.2.4.patchset
```
Then, in Emacs I use **Tools** > **Compare (Ediff)** > **Two buffers** to easily revert the unneeded differences (i.e. the random sha1sums and the git versions) in `foobar-1.2.4.patchset` by overwriting them with what we have in `foobar-1.2.3.patchset`. This allows to have a much smaller `git log -p` on the commit I'm preparing.

Of course, the patchset does not apply cleanly on all situations, so when it needs manual edits I just go for it, using `--author` and `--date` options of `git commit` so that the `hp -e foobar-1.2.4` will produce an updated patchset keeping the original authorship.

Now, in cases like this one, where the patch does not exist in our tree, creating a copy of something that already exists somewhere else is not a recommended because it adds noise to our repo.
Using an "external" patch is imho the best method to use patches written by 3rd parties and which are available with a few extra lines in the recipe :)

alaviss:
> Now, in cases like this one, where the patch does not exist in our tree, creating a copy of something that already exists somewhere else is not a recommended because it adds noise to our repo.

There's noise either way :). Either in the recipe or the patchset. I prefer noise where we don't usually look at. When an external patch doesn't cleanly apply, chances are it's no longer required, or there're updated version for it. In such cases, investigating whether the patch is needed should be done. Having commit noise in the patchset is not much of a problem, since it will eventually disappear in the future.

Also, it's generally recommended to host patches in our own infrastructure. That way, when Gentoo changes their infra/or having an outage, we wouldn't end up with an unbuildable port.

fbrosson:
> There's noise either way :)

True, but with an external patch the noise is limited to 10 lines :)
* 4 x 1 line for the definition of `SOURCE_FILENAME_2`, `srcGitRev_2`, `SOURCE_URI_2` and `CHECKSUM_SHA256_2`;
* 4 lines for the `PATCH` function.
* 2 x 1 blank line.

The external patchset has 127 lines, so if we had chosen to make a copy of it in our repo, it would have been over 100 lines.

Regarding the availability of the Gentoo infrastructure, they have probably never had any big downtime.
Moreover, HaikuPorter has it's own cache, so I'm very confident that we won't ever have any issues with downloading any external patch from Gentoo.

OK, they could change the URI of their files, but that would not be a big issue because:
* we have our cache, and it would probably already have a copy of the external patch, so nobody would be blocked;
* it would be easy to update our recipes to use the new URIs, should such a change occur, because we really have very few recipes which use external patches from Gentoo:
  * app-shells/dash/dash-0.5.10.2.recipe
  * app-text/hunspell/hunspell-1.4.2.recipe
  * app-text/hunspell/hunspell-1.5.4.recipe
  * dev-libs/geoip/geoip-1.6.12.recipe
  * dev-libs/libmspack/libmspack-0.6~alpha.recipe
  * media-gfx/png2ico/png2ico-2002.12.08.recipe
  * media-libs/audiofile/audiofile-0.3.6.recipe
  * media-libs/tiff/tiff-3.9.7.recipe
  * sys-fs/squashfs_tools/squashfs_tools-4.3.recipe

PR #2726, opened 2018-06-23
zeromq, bump version

Begasus:


net-libs/zeromq/zeromq-4.2.3.recipe
```diff
@@ -0,0 +1,91 @@
...
+	2012 Spotify AB
+	2013 Ericsson AB
+	2014 AppDynamics Inc.
+	2015 Google, Inc.
+	2015-2016 Brocade Communications Systems Inc."
+LICENSE="GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/zeromq/libzmq/releases/download/v$portVersion/zeromq-$portVersion.tar.gz"
+CHECKSUM_SHA256="8f1e2b2aade4dbfde98d82366d61baef2f62e812530160d2e6d0a5bb24e40bc0"
+SOURCE_URI_2="https://raw.githubusercontent.com/zeromq/cppzmq/a604af9/zmq.hpp#noarchive"
```
Begasus: cppzmq provides a seperate archive at:
https://github.com/zeromq/cppzmq
I've used the latest push for zmq.hpp now (not sure if it's needed to build a seperate package for it?)

net-libs/zeromq/zeromq-4.2.3.recipe
```diff
@@ -0,0 +1,91 @@
...
+CHECKSUM_SHA256_2="02b73219656df3ffe3b231c7d0326e754a738ef2a83c8e94cd1e47d22cdd9535"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
fbrosson:
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64" 
-SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
```
In `SECONDARY_ARCHITECTURES` we don't need to mention those secondary arches which are already marked as broken in `ARCHITECTURES`.
Actually, I think it is even recommended to not mention `!x86_gcc2` in `SECONDARY_ARCHITECTURES`. Having `!x86_gcc2` in `ARCHITECTURES` is sufficient.

BTW, if zeromq can't be easily built on x86_gcc2 primary arch then it might be a good idea to define [`commandBinDir`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) (and `commandSuffix`) and replace `cmd:curve_keygen$secondaryArchSuffix` by `cmd:curve_keygen$commandSuffix`.)

Begasus: checked the build for x86_gcc2 and it failed yes (even with some small changes), I'll set commandSuffix :)

fbrosson:
Thanks for the `commandBinDir` and `commandSuffix`!

To find out whether or not we could drop the previous recipe and its patch, I have built and tested both `4.2.2` and `4.2.3`, each on every architecture.
What I have noticed is that there are no regressions. The 3 tests that fail on `4.2.2` (on each arch) still fail on `4.2.3`:
* `FAIL: tests/test_reqrep_tcp`
* `FAIL: tests/test_inproc_connect`
* `FAIL: tests/test_shutdown_stress`

FWIW, here are the statistics for the tests for each version on each arch:

* `4.2.2`

Architecture | TOTAL | PASS | FAIL
--- | --- | --- | ---
x86 primary | 86 | 83 | 3
x86 secondary | 85 | 82 | 3
x86_64 | 85 | 82 | 3

* `4.2.3`

Architecture | TOTAL | PASS | FAIL
--- | --- | --- | ---
x86 primary | 89 | 86 | 3
x86 secondary | 88 | 85 | 3
x86_64 | 88 | 85 | 3

BTW, whenever we do a version bump for a recipe which we can easily test, it would be nice to do such comparisons, so that we can avoid having more than one version for a given recipe.
(Here, zeromq `4.2.3` is as good as `4.2.2`, so we can drop the old version.)
P.S. I did the test on x86 primary arch, but since this arch is not officially supported, it is OK to not care about x86 primary arch.

So, if you could update your PR to remove `zeromq-4.2.2.recipe` and `patches/zeromq-4.2.2.patchset` it would be great!
I can help you if you need any guidance for this. You can push a new commit or amend the existing commit and push if with `--force`. Both will lead to the same result because we will be doing a "Squash and merge".

(Hmm, if you prefer, I can do the update for you.)

alaviss:
Please open an issue for this port test failures as well. Thanks!

PR #2741, opened 2018-06-26
vifm, bump version

Begasus:


app-misc/vifm/vifm-0.9.1.recipe
```diff
@@ -4,13 +4,12 @@ options/commands/configuration, which also borrows some useful ideas from mutt.
 If you use vi, Vifm gives you complete keyboard control over your files without \
 having to learn a new set of commands."
 HOMEPAGE="https://vifm.info/"
-COPYRIGHT="2017"
+COPYRIGHT="2018"
```
alaviss: I wonder how could haikuporter accept this

Begasus: Thought it was strange too, will look further into this, but I think I didn't see a mention of any copyright holder, ps no error in haikuporter with -S

PR #2747, opened 2018-06-29
popt, fix build

Begasus:
The test still fails, could use some looking into ...


alaviss:
@Begasus 

Please open an issue for the tests failure so that everyone could look into it. Thanks!

Issue #2743, opened 2018-06-26
Emacs: gnutls fails to connect to org-mode.org


tidux:
Steps to reproduce:

1. ` git clone https://github.com/syl20bnr/spacemacs ~/.emacs.d && emacs`

2. When it finishes, close (`C-x C-c` or just killing the Terminal tab) and start `emacs` again.

PR #2753, opened 2018-06-30
emacs: bump

extrowerk:


extrowerk:
fixes https://github.com/haikuports/haikuports/issues/2743

app-editors/emacs/emacs-26.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+{
+	# Hard links not supported on BFS.
+	sed -i 's/ln -f/ln -s/g' src/Makefile.in
+}
+
+BUILD()
+{
+	export LDFLAGS="-lnetwork -lbsd"
+	CFLAGS="-D_BSD_SOURCE"
+	export HOME=/boot/home
```
alaviss: This should be contained inside the build folder. We don't want it to mess with user's home when they build the recipe without chroot.

Also:

```diff
-	export LDFLAGS="-lnetwork -lbsd"
-	CFLAGS="-D_BSD_SOURCE"
+	export LIBS="-Wl,--as-needed -lnetwork -lbsd"
+	export CPPFLAGS="-D_BSD_SOURCE"
```

`-Wl,--as-needed` will avoid linking those libraries if they are unneeded.

app-editors/emacs/emacs-26.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+		rm -rf src/$f
+	done
+
+	make install
+
+	# ctags fix
+	mv $binDir/ctags $binDir/ctags.emacs
+
+	# Cleanup
+	find $dataDir -type f -name "*.el.gz" -delete
```
alaviss: What are those files?

extrowerk: The emacs source code. Emacs provides his source as gzip files. Some packagers creates an own subpackage for them.
They aren't needed for normal use.

PR #2759, opened 2018-07-01
QtDropbox: new recipe

extrowerk:


dev-qt/qtdropbox/qtdropbox-20151109.recipe
```diff
@@ -0,0 +1,77 @@
+SUMMARY="An API for the well known cloud storage service"
+DESCRIPTION="Basically QtDropbox aims to provide an easy to use possibility \
+to access the REST API of Dropbox. All HTTP calls are hidden behind the \
+curtains of neat C++/Qt classes with nice method names and specific uses."
+HOMEPAGE="https://github.com/lycis/QtDropbox/"
```
fbrosson:
```diff
-HOMEPAGE="https://github.com/lycis/QtDropbox/"
+HOMEPAGE="https://github.com/lycis/QtDropbox"
```
If we did get a redirect (to the address that includes a trailing slash) when we request the address without the trailing slash, then yes, it would have been nice to keep the trailing slash.
But here, we don't get redirected, so this means the trailing slash is useless.
(We can therefore drop it.)

dev-qt/qtdropbox/qtdropbox-20151109.recipe
```diff
@@ -0,0 +1,77 @@
...
+to access the REST API of Dropbox. All HTTP calls are hidden behind the \
+curtains of neat C++/Qt classes with nice method names and specific uses."
+HOMEPAGE="https://github.com/lycis/QtDropbox"
+COPYRIGHT="2015 Daniel Eder"
+LICENSE="GNU GPL v3"
+REVISION="1"
+srcGitRev="17ad0070e8157fc973a3f3c47c676e73728c26d0"
+SOURCE_URI="$HOMEPAGE/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="edceab7fdc5deeeb39b2f868ba24ccd2b479013467dad4c81179a412e87bdc6b"
+SOURCE_DIR="QtDropbox-$srcGitRev"
```
alaviss:
```diff
 CHECKSUM_SHA256="edceab7fdc5deeeb39b2f868ba24ccd2b479013467dad4c81179a412e87bdc6b"
+SOURCE_FILENAME="QtDropbox-$portVersion-$srcGitRev.tar.gz"
 SOURCE_DIR="QtDropbox-$srcGitRev"
```

dev-qt/qtdropbox/qtdropbox-20151109.recipe
```diff
@@ -0,0 +1,77 @@
+SUMMARY="An API for the well known cloud storage service"
+DESCRIPTION="Basically QtDropbox aims to provide an easy to use possibility \
+to access the REST API of Dropbox. All HTTP calls are hidden behind the \
+curtains of neat C++/Qt classes with nice method names and specific uses."
+HOMEPAGE="https://github.com/lycis/QtDropbox"
+COPYRIGHT="2015 Daniel Eder"
+LICENSE="GNU GPL v3"
```
alaviss: According to [upstream](https://github.com/lycis/QtDropbox#further-information)

```diff
-LICENSE="GNU GPL v3"
+LICENSE="GNU LGPL v3"
```

dev-qt/qtdropbox/qtdropbox-20151109.recipe
```diff
@@ -0,0 +1,77 @@
...
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Network$secondaryArchSuffix
+	devel:libQt5Xml$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:qmake$secondaryArchSuffix >= 5.7
```
alaviss: Shouldn't this be enough?

```diff
-	cmd:qmake$secondaryArchSuffix >= 5.7
+	cmd:qmake$secondaryArchSuffix >= 5
```

PR #2766, opened 2018-07-01
pciutils: bump

extrowerk:


sys-apps/pciutils/patches/pciutils-3.6.0.patchset
```diff
@@ -0,0 +1,24 @@
...
++++ b/lib/sysdep.h
+@@ -68,6 +68,10 @@ typedef u16 word;
+ #endif
+ #endif
+ 
++#ifdef PCI_OS_HAIKU
++#include <endian.h>
++#endif
++
+ #ifdef PCI_OS_SYLIXOS
```
Begasus: Can't you use #if defined(PCI_OS_SYLIXOS) || defined(PCI_OS_HAIKU)?

fbrosson: Each OS has its dedicated `#ifdef` bloc, so it is better to keep the one for Haiku in a separate bloc too.

sys-apps/pciutils/pciutils-3.6.0.recipe
```diff
@@ -0,0 +1,140 @@
...
+
+- lspci - displays detailed information about all PCI buses and devices in \
+the system
+- setpci - allows reading from and writing to PCI device configuration \
+registers. For example, you can adjust the latency timers with it."
+HOMEPAGE="https://mj.ucw.cz/sw/pciutils/"
+COPYRIGHT="1997-2018 Martin Mares"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="ftp://atrey.karlin.mff.cuni.cz/pub/linux/pci/pciutils-$portVersion.tar.gz"
```
alaviss: The [kernel.org](https://www.kernel.org/pub/software/utils/pciutils/) mirror should also be added, to aid people behind firewalls.

sys-apps/pciutils/pciutils-3.6.0.recipe
```diff
@@ -0,0 +1,140 @@
...
+fi
+
+PROVIDES="
+	pciutils$secondaryArchSuffix = $portVersion
+	cmd:lspci$secondaryArchSuffix = $portVersion
+	cmd:setpci$secondaryArchSuffix = $portVersion
+	"
+if [ "$LIBPCI_SHARED" = yes ]; then
+	PROVIDES="$PROVIDES
+		lib:libpci$secondaryArchSuffix = $portVersion compat >= 3
```
alaviss: Shall we use `libVersionCompat` here?

```sh
libVersion="$portVersion"
```

```diff
-		lib:libpci$secondaryArchSuffix = $portVersion compat >= 3
+		lib:libpci$secondaryArchSuffix = $libVersionCompat
```

sys-apps/pciutils/pciutils-3.6.0.recipe
```diff
@@ -0,0 +1,140 @@
...
+	ARCHITECTURES_data="any"
+	PROVIDES_data="
+		pciutils_data = $portVersion
+		"
+	REQUIRES_data=""
+fi
+
+PROVIDES_devel="
+	pciutils${secondaryArchSuffix}_devel = $portVersion
+	devel:libpci$secondaryArchSuffix = $portVersion compat >= 3
```
alaviss:
```diff
-	devel:libpci$secondaryArchSuffix = $portVersion compat >= 3
+	devel:libpci$secondaryArchSuffix = $libVersionCompat
```

sys-apps/pciutils/pciutils-3.6.0.recipe
```diff
@@ -0,0 +1,140 @@
...
+	fixPkgconfig
+
+	if [ -z "$secondaryArchSuffix" ]; then
+		maybe_manDir_man7=$manDir/man7
+		packageEntries data \
+			$dataDir
+		sed -i -e "s,$dataDir/,/system/data/,g" \
+			$manDir/man8/lspci.8 \
+			$manDir/man8/setpci.8
+		local include_pci=/system/$relativeOldIncludeDir/pci
```
alaviss: Please use the normal `includeDir`. It may look bad, but it will reflect the correct header location in case the user install the package in user prefix.

sys-apps/pciutils/pciutils-3.6.0.recipe
```diff
@@ -0,0 +1,140 @@
...
+			$manDir/man8/lspci.8
+		sed -i -e "s,/usr/include/pci/,$include_pci/,g" \
+			$manDir/man8/setpci.8
+	else
+		maybe_manDir_man7=
+		rm -rf $dataDir $documentationDir
+	fi
+
+	# $dataDir/pciutils/pci.ids is not in pciutils but in pciutils_data.
+	local data_dir=/packages/pciutils_data${dataDir#/packages/pciutils$secondaryArchSuffix}
```
alaviss:
```diff
-	local data_dir=/packages/pciutils_data${dataDir#/packages/pciutils$secondaryArchSuffix}
+	local data_dir=$(getPackagePrefix pciutils_data-$portVersion)/${relativeDataDir}
```

This should looks cleaner :)

sys-apps/pciutils/pciutils-3.6.0.recipe
fbrosson: @alaviss, here `$include_pci` is only used to patch `$manDir/man8/lspci.8` and `$manDir/man8/setpci.8`.
Imo it's OK to have the man pages document the friendly paths.
Shall we keep what we have right now in the recipe about this `include_pci` variable?

sys-apps/pciutils/pciutils-3.6.0.recipe
fbrosson: Hmm, I would be OK with using `getPackagePrefix`, but not until someone has managed to fix haikuporter so that it won't produce the warnings we get in Travis-CI.
We already have `getPackagePrefix` in a few other recipes, and everytime the Travis-CI script is called for any recipe, we see a lot of noise:
<Details>

```
packageEntries: error: "fortran" doesn't seem to be
a valid package suffix.
packageEntries: error: "fortran" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
packageEntries: error: "syslibs" doesn't seem to be
a valid package suffix.
```

alaviss: I think it's only a problem when you put it on the topmost level. Travis certainly won't care about stuff in functions.

sys-apps/pciutils/pciutils-3.6.0.recipe
alaviss: Nice paths are only good for end-user, for devs it's a nightmare when you can't find what you want where you were told. So I don't agree with the nice paths.

fbrosson: Well, I'd say that:
1) even non developers might read the man pages of `lspci` and `setpci`.
2) developers who might install the pciutils package in `/boot/home/config/packages/` (instead of `/boot/system/packages/`) are already aware that anything they would get in `/boot/system/develop/headers/` with "system-wide installs" is to be expected in `/boot/home/config/develop/headers/` with "user-only installs".
3) The real paths are really ugly and do not bring much added value.
4) If I'm not mistaken, the paths shown in the man pages would anyway be wrong if the pciutils package is installed in `/boot/home/config/packages/` instead of `/boot/system/packages/`.

sys-apps/pciutils/pciutils-3.6.0.recipe
```diff
@@ -14,13 +14,17 @@ HOMEPAGE="https://mj.ucw.cz/sw/pciutils/"
 COPYRIGHT="1997-2018 Martin Mares"
 LICENSE="GNU GPL v2"
 REVISION="1"
-SOURCE_URI="ftp://atrey.karlin.mff.cuni.cz/pub/linux/pci/pciutils-$portVersion.tar.gz"
+SOURCE_URI="ftp://atrey.karlin.mff.cuni.cz/pub/linux/pci/pciutils-$portVersion.tar.gz
+	https://mirrors.edge.kernel.org/pub/software/utils/pciutils/pciutils-$portVersion.tar.gz"
```
fbrosson: Oops, the tarball we get from the second URI does not have the same sha256sum:
```
36453246969d1fe2ec8da9db481ed12f288f3e9117d24caf10238cb3cfb8e58f  pciutils-3.6.0.tar.gz
```
This means we can't keep this mirror.
I found https://www.mirrorservice.org/pub/software/utils/pciutils/ but it has the same sha256sum as the above mirror, so we can't use it either.

The only mirror we can use is https://fossies.org/linux/misc/pciutils-3.6.0.tar.gz/

```diff
 SOURCE_URI="ftp://atrey.karlin.mff.cuni.cz/pub/linux/pci/pciutils-$portVersion.tar.gz
-	https://mirrors.edge.kernel.org/pub/software/utils/pciutils/pciutils-$portVersion.tar.gz"
+	https://fossies.org/linux/misc/pciutils-$portVersion.tar.gz"
```

sys-apps/pciutils/pciutils-3.6.0.recipe
alaviss: Guess I'm fine with whatever paths you put there. But why `relativeOldIncludeDir` though?

sys-apps/pciutils/pciutils-3.6.0.recipe
alaviss: Or we can actually migrate to kernel.org mirror, they have `.xz` tarballs ;)

fbrosson: Yup, this might be an option.
That said, there is (or was?) an implicit recommendation to avoid `tar.xz` tarballs for recipes that build on x86_gcc2 (primary arch). This is (or was) supposed to make the recipe friendly to users of x86_gcc2 *non*-hybrid systems, as xz_utils is not available on these systems.

So, unless someone tells us we can ignore this recommendation, I'd keep the `tar.gz` for now.
(Once gcc2 support is gone we will be free to switch to `tar.xz` :)

sys-apps/pciutils/pciutils-3.6.0.recipe
fbrosson: Good question :)
I don't really know the answer...

I guess that when I added the `$relativeOldIncludeDir` in the recipe I was not yet familiar with all the variables haikuporter puts in `work-x86-*/wrapper-script`, so when I looked for a variable defined as `'develop/headers/x86'`, the first one I saw was `relativeOldIncludeDir`.
BTW, I had found it while viewing the file with vi, but, as we can see with a `grep -n`, the `relativeOldIncludeDir` is much closer to the begining of the file than the `relativeIncludeDir`:
```bash
grep -n "'develop/headers/x86'" wrapper-script 
15:relativeOldIncludeDir='develop/headers/x86'
68:relativeIncludeDir='develop/headers/x86'
```
So I guess I had not even found the shorter variable.

So you are perfectly right. I could have used `relativeIncludeDir` instead.

And yes, we can replace the `relativeOldIncludeDir` by a `relativeIncludeDir` :)

scottmc:
What's the status on this one?  Can we get it wrapped up soon?

fbrosson:
Several weeks ago I noticed that `lspci` was crashing on x86_64 (and maybe on x86_gcc2 too.)
I remember that `lspci` did not crash on any arch the last time I had updated this recipe.
And I believe there has been a regression in Haiku.
I did not recheck this recently, so maybe things are better now.
BTW, I currently don't have enough time to investigate or even open an issue, sorry...

diversys:
I created a ticket for that https://dev.haiku-os.org/ticket/14429

scottmc:
lspci issue has been resolved in Haiku, please retry with an updated Haiku.

Issue #2760, opened 2018-07-01
command_not_found package broken?


extrowerk:
On x86_64 i got this output:

https://dev.haiku-os.org/ticket/14222#comment:3

See the attachment.

alaviss:
`GLOBAL_WRITABLE_FILES` is supposed to be the metadata for such purposes, but no one implemented any cleaning up functions for `pkgman` using it AFAIK.

alaviss:
Closed via 728340e427aed9ef190ab6a39115076131b0d275

PR #2776, opened 2018-07-06
app-shells/command_not_found: fix handler

alaviss:
- Properly escaped dollar signs
- Respects command_not_found error code
- Added fallback to avoid deadlock in case command_not_found was
uninstalled

Solves #2760

This is done via a quick web search and is completely untested.


alaviss:
@extrowerk Can you check if this PR solves the issue mentioned above?

alaviss:
Closed via 728340e427aed9ef190ab6a39115076131b0d275

PR #2778, opened 2018-07-07
popt, fix pkg-config file for x86 secondaryArchSuffix

Begasus:


waddlesplash:
Maybe `fixPkgconfig` should do this?

dev-libs/popt/popt-1.16.recipe
```diff
@@ -61,6 +61,10 @@ INSTALL()
 	make install
 
 	rm $libDir/libpopt.la
+if [ -n "$secondaryArchSuffix" ]; then
+	mkdir -p $libDir/pkgconfig
+	mv $prefix/lib/pkgconfig/* $libDir/pkgconfig/
```
alaviss: Oops, there will be an empty `$prefix/lib/pkgconfig` folder in the resulting package. You should add

    rmdir "$prefix/lib/pkgconfig"

to get rid of that folder

PR #2799, opened 2018-07-13
xproto, bump revision after changes to util-macros

Begasus:


Begasus:
So far things are looking good with xorgproto, so I think this PR could be closed without merging later :)
(xproto is included in xorgproto)

PR #2800, opened 2018-07-13
libfontenc, bump revision after changes to util-macros

Begasus:


Begasus:
@fbrosson @alaviss @korli this recipe is mixing up 2 arch's, in this case it works ok (as there are no other xlibs involved), when I try to build libxfont_x86 I end up with 2 arch's also, but then it needs this library (which needs pkg-config-x86) meaning it won't find xproto and fontsproto (any arch)
So maybe it's not a good idea to have the headers "any" arch to begin with?

PR #2804, opened 2018-07-14
fontsproto, bump revision after util-macros changes

Begasus:


Begasus:
So far things are looking good with xorgproto, so I think this PR could be closed without merging later :)
(fontsproto is included in xorgproto)

PR #2806, opened 2018-07-14
xorgproto, new recipe, combine all proto packages in one

Begasus:


alaviss:
This package should have every protos that it provides in `PROVIDES`, if it want to be a drop-in replacement.

Begasus:
@alaviss xorgproto provides them all already atm? Or should they be added explicitly?
In that case you still would need to add "devel:xproto" for instance instead of "devel:xorgproto"?

x11-proto/xorgproto/xorgproto-2018.4.recipe
```diff
@@ -0,0 +1,53 @@
...
+REVISION="1"
+SOURCE_URI="https://xorg.freedesktop.org/archive/individual/proto/xorgproto-$portVersion.tar.bz2"
+CHECKSUM_SHA256="fee885e0512899ea5280c593fdb2735beb1693ad170c22ebcc844470eec415a0"
+PATCHES="xorgproto-$portVersion.patchset"
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	xorgproto = $portVersion
+	devel:xorgproto = $portVersion
```
alaviss: @Begasus 

What I meant is something like this. This way, we don't have to modify other packages, and can easily switch back to separate package, should other contributors dislike the idea :)

```diff
-	devel:xorgproto = $portVersion
+	devel:applewmproto = 1.4.2
+	devel:bigreqsproto = 1.1.2
+	and so on
```
(names and versions are found in `meson.build` file in the tarball)

Begasus: Sounds reasonable, but wouldn't haikuporter pull xproto from the depot atm instead of using the one declared in xorgproto?

alaviss: Unless you have xorgproto installed, that is :)

And I think xorgproto should confilcts with all other protos

PR #2815, opened 2018-07-19
libxcb, bump after changes to util-macros

Begasus:


Begasus:
Checks passed for libxcb and libxcb_x86 on gcc2 hybrid ...

x11-libs/libxcb/libxcb-1.13.recipe
```diff
@@ -118,29 +117,26 @@ PROVIDES_devel="
...
 	cmd:doxygen
+	cmd:gcc$secondaryArchSuffix
 	cmd:libtoolize$secondaryArchSuffix
-	cmd:aclocal
 	cmd:make
-	cmd:gcc$secondaryArchSuffix
-	cmd:ld$secondaryArchSuffix
 	cmd:pkg_config$secondaryArchSuffix
 	cmd:python
+	cmd:xsltproc
```
fbrosson: We need a `$secondaryArchSuffix` here.
```diff
-	cmd:xsltproc
+	cmd:xsltproc$secondaryArchSuffix
```
That's because libxslt's recipe provides it with the suffix:
https://github.com/haikuports/haikuports/blob/8dd83364dfc08075d5ac885ec61e7325361c0180/dev-libs/libxslt/libxslt-1.1.32.recipe#L23

fbrosson: Oops, I just noticed libxslt builds on all architectires, so the rationale in my previous comment is not correct.

The reason why we should probably add the suffix is that it will make sure we don't pull libxslt's libs for a primary arch when we build libxcb for a secondary arch.

pulkomandy:
On Thu, Jul 19, 2018 at 05:15:51PM -0700, fbrosson wrote:
> The reason why we should probably add the suffix is that it will make sure we don't pull libxslt's libs for a primary arch when we build libxcb for a secondary arch.

How would that be a problem? The libs are in different directories so
they wouldn't be found. And even if they were, linking would fail.

xlstproc binary should be packaged for a single architecture. The
secondary arch recipe should drop all binaries.

fbrosson:
I wrote that _we should **probably** add the suffix_ because I think it is safer **_with_** the suffix than **_without_** it.

My undestadning is that :
* it is perfectly OK for a secondary arch build of a package to depend on a `cmd:` provided by a primary arch package, as long as that primary arch does not provide any libs.
* in secondary arch builds (e.g. `libxcb_x86`) we should avoid adding dependencies on primary arch `cmd:` entries **_without_** the suffix if these entries are also available **_with_** the suffix and if that dependency provides libs.

> The libs are in different directories so they wouldn't be found. And even if they were, linking would fail.

If using the suffix does not cost much and allows to avoid any risk of having build failures, why would we not use the suffix?


Regarding the libxslt package, I believe :
* the only reason why we might choose to  drop the `cmd:xlstproc` on secondary arch builds would be to make the secondary arch packages smaller.
* the potential impacts of dropping the `cmd:xlstproc` on secondary arch builds could be worse than the benefits.

There might very well be no such impacts, but at least we don't need to worry about this if we keep the `cmd:xsltproc$secondaryArchSuffix` in libxslt's recipe.

Having said this, I don't object such a change in libxslt. It's just that I don't think this would be an improvement.

pulkomandy:
Let's look at our users.

What they want is a single version of the executable. Why provide two versions if they behave exactly the same?

As a result, bin/x86 should be used only when there is significant differences between the gcc2 and gcc7 version of a package (for example, when we had both ffmpeg 0.10 and ffmpeg_x86 4.x ; or for gcc itself). In most case, there should be a single executable stored in bin/.

This unfolds to two possibilities:
- Either the package can be built with gcc2, and the gcc2 binary can be provided in bin/
- Or, the package cannot be built with gcc2, so the secondary arch binaries are installed in bin/

As a result, bin:+$secondaryArchSuffix is an exceptional case, most often used for build tools: pkg-config, gcc, sdl-config, etc, where we really need both versions to coexist.

fbrosson:
Okay, you have convinced me :)

Begasus:
Thanks for responding, in these cases xsltproc binary is only used (no library is called from this script or any of the other ones), so I think this should be ok (even then ... adding it in or not doesn't really make a difference as far as I can see ...

PR #2820, opened 2018-07-21
libxscrnsaver, bump revision after changes to util-macros and xorgproto

Begasus:


waddlesplash:
Yes, provides and requires are case insensitive, so it's a cosmetic issue.

PR #2821, opened 2018-07-21
libxfixes, bump revision after changes to util-macros and xorgproto

Begasus:


x11-libs/libxfixes/libxfixes-5.0.3.recipe
```diff
@@ -66,8 +65,7 @@ defineDebugInfoPackage libxfixes$secondaryArchSuffix \
 
 BUILD()
 {
-	libtoolize --force --copy --install
-	autoreconf -i
+	autoreconf -vfi
```
Begasus: Should these be needed, in this case libxfixes builds fine without "autoreconf -vfi" and I'm pretty sure it's not needed for the other x recipes?

Begasus: Nevermind, looks like in some cases it's needed :)

PR #2835, opened 2018-07-21
Pygame 1.9.4

extrowerk:


dev-python/pygame/patches/pygame-1.9.4.patchset
```diff
@@ -302,18 +302,18 @@ index c54a93a..1b9d208 100644
  src/scale_mmx64_gcc.c
```
alaviss: Do we still need to drop these `scale` files?

dev-python/pygame/python_pygame-1.9.4.recipe
```diff
@@ -47,8 +49,9 @@ BUILD_REQUIRES="
 	"
```
alaviss: There shouldn't be a newline here.

dev-python/pygame/python_pygame-1.9.4.recipe
```diff
@@ -1,16 +1,16 @@
 SUMMARY="A popular game development module for python"
 DESCRIPTION="PyGame - python bindings to sdl and other libs that facilitate \
 game production."
-HOMEPAGE="http://www.pygame.org/"
+HOMEPAGE="https://www.pygame.org/"
 COPYRIGHT="2000-2004, 2007  Pete Shinners
```
alaviss:
```diff
-COPYRIGHT="2000-2004, 2007  Pete Shinners
+COPYRIGHT="2000-2004, 2007 Pete Shinners
```

dev-python/pygame/python_pygame-1.9.4.recipe
waddlesplash: Huh? Yes there should, this is between PREREQUIRES and BUILD().

dev-python/pygame/patches/pygame-1.9.4.patchset
waddlesplash: Probably.

fbrosson:
FWIW, the default merge method (between _"Squash and merge"_ and _"Rebase and merge"_) varies depending on which merge method one has chosen for the previous merge.
Here, for instance, the _"Rebase and merge"_ method was chosen, so we got: 7de1f05091a2a9ef1a17c7d89893406708c6cffd.
My suggestion: always check the merge method if there is more than one commit in the PR.

@waddlesplash, I saw that you have switched to _"Squash and merge"_ in your next merge, and this is cool! So thanls, and sorry for writing ths little reminder :)

Issue #2598, opened 2018-05-23
ICU: Bump to ICU v61.1


kenmays:
ICU 61 provides full support for the recent Unicode 10.0 release with many new characters and many property improvements. Unicode 10.0 adds 8,518 characters, for a total of 136,690 characters. These additions include 4 new scripts, for a total of 139 scripts, as well as 56 new emoji characters. Locale data is updated to CLDR 33 locale data and adds Google double-conversion library for formatting doubles.  This is the library used in V8 and a number of other projects for converting doubles to decimals. 

Homepage: http://site.icu-project.org/
Src: http://download.icu-project.org/files/icu4c/61.1/icu4c-61_1-src.tgz

Successfully Tested: ICU 61.1 on Haiku x86_64 hrev51971
[icu-61.1-testing_cocobean.txt](https://github.com/haikuports/haikuports/files/2029848/icu-61.1-testing_cocobean.txt)

Recipe/Patches:
[icu-61.1-haiku.zip](https://github.com/haikuports/haikuports/files/2030156/icu-61.1-haiku.zip)

korli:
What about gcc2?

kenmays:
@korli - Give it a test and let me know. I'll set up for that soon.

kenmays:
1. Can enhance ICU 57.1 recipe with tzdata2018e.

extrowerk:
@kenmays : please use pull requests for your work, and do not manually edit the patch files.
Are your recipes tested with haikuporter?

My PR superseeds your patches and recipe, but it would be nice if you would use the infrastructure, as it allows us to give comments, ideas and check your work easily.

extrowerk:
Your gcc2 patchset summary references files, wich aren't patched in your patchset, that's inconsistent and not acceptable.

kenmays:
@extrowerk - Amazing work on bumping ICU. Closing this one!

kenmays:
https://github.com/haikuports/haikuports/pull/5097

PR #2845, opened 2018-07-22
ICU: bump

extrowerk:
Bump, coexistence, updated patchset.

Warning: this build uses the ICU 61+ namespace. To change it to the 60 and earlier namespace consult the "readme.html" in the icu4c folder.

Patch doesn't applies cleanly, some part completly missing, so i kept only what useful.
The packaging stage have a bug somewhere along the devel package requires, help would be really appreciated.

### DIFF
--- a/dev-libs/icu/icu62-62.1.recipe
+++ b/dev-libs/icu/icu62-62.1.recipe
@@ -42,13 +42,13 @@ HOMEPAGE="http://www.icu-project.org"
 COPYRIGHT="1995-2018 IBM Corporation and others."
 LICENSE="ICU"
 REVISION="1"
-SOURCE_URI="https://sourceforge.net/projects/icu/files/ICU4C/62.1/icu4c-62_1-src.tgz"
+SOURCE_URI="https://github.com/unicode-org/icu/archive/release-62-1.tar.gz"
 SOURCE_URI_2="http://www.iana.org/time-zones/repository/releases/tzdata2018e.tar.gz#noarchive"
 SOURCE_URI_3="https://sourceforge.net/projects/icu/files/ICU4C/62.1/icu4c-62_1-data.zip#noarchive"
-CHECKSUM_SHA256="3dd9868d666350dda66a6e305eecde9d479fb70b30d5b55d78a1deffb97d5aa3"
+CHECKSUM_SHA256="e15ffd84606323cbad5515bf9ecdf8061cc3bf80fb883b9e6aa162e485aa9761"
 CHECKSUM_SHA256_2="6b288e5926841a4cb490909fe822d85c36ae75538ad69baf20da9628b63b692e"
 CHECKSUM_SHA256_3="93bcaf58cfa9223972da79cb8e0f94819ace4995db6041b84f4d9bd810246d1d"
-SOURCE_DIR="icu"
+SOURCE_DIR="icu-release-62-1"
 PATCHES="icu-$portVersion.patchset"
 
 ARCHITECTURES="?x86_gcc2 ?x86 ?x86_64 ?arm"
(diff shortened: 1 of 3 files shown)


pulkomandy:
Don't change the namespace back to icu60, this is how ICU handles ABI versions. It makes sure all symbols are different from one version to the other, and as a result it is even possible to link a lib with an old ICU, another lib with a new one, and then use both libs in the same app without problems.

What's in the patches looks ok, but what happened to gcc2 support? Most of the old patch was about that, IIRC (as everything else, we can upstream to ICU).

General comments as I proofread this:
- It seems strange that we use separate data dirs for gcc2/gcc7, yet we use the same path for all ICU versions. How comes the files are interchangeable between ICU 5.4 and 6.2, but not between gcc2 and gcc7 versions?
- It seems strange that $libDir/icu ends up in the devel package. What's in there?

I'm ok to help but your description of the problem isn't very helpful. Do you have a buildlog or something?

extrowerk:
The mentioned files in the patchset does not exist, i grepped the source tree to see if it is moved to somewhere else, but nothing.  Maybe i overlooked it.

The libdir/icu contains: "pkgdata.inc" and "Makefile.inc" files. I think they are correctly delegated to the devel package.

Packaging fixed in the meantime, but the test fails:

```
cnumtst.c: In function 'Test12052_NullPointer':
cnumtst.c:3070:34: error: 'u' undeclared (first use in this function)
     static const UChar input[] = u"199a";
                                  ^
cnumtst.c:3070:34: note: each undeclared identifier is reported only once for each function it appears in
cnumtst.c:3070:35: error: expected ',' or ';' before string constant
     static const UChar input[] = u"199a";
                                   ^~~~~~
```

pulkomandy:
On Sun, Jul 22, 2018 at 07:44:55AM -0700, miqlas wrote:
> The mentioned files in the patchset does not exist, i grepped the source tree to see if it is moved to somewhere else, but nothing.  Maybe i overlooked it.

Yes, maybe the patches need to be re-done elsewhere. Or maybe they
changed their code slightly and it builds again with gcc2.

> The libdir/icu contains: "pkgdata.inc" and "Makefile.inc" files. I think they are correctly delegated to the devel package.

Yes, putting them in the devel package is right, but why are they in
$libDir?

> 
> Packaging fixed in the meantime, but the test fails:
> 
> ```
> cnumtst.c: In function 'Test12052_NullPointer':
> cnumtst.c:3070:34: error: 'u' undeclared (first use in this function)
>      static const UChar input[] = u"199a";

GCC2 does not support this notation. This may work:

    static const UChar input[] = {'1', '9', '9', 'a', 0};

-- 
Adrien.

extrowerk:
I see this, maybe could be helpful:
"=== Important Notes: ===
Data Packaging: library
 This means: ICU data will be linked with ICU. A shared data library will be built. 
 To locate data: ICU will use the linked data library. If linked with the stub library located in stubdata/, the application can use udata_setCommonData() or set a data path to override."

In the meantime i found an another tarball wich contains all the required files, applied the patchset, looks good.

extrowerk:
Tests needs some patching, but that is well beyon my skills.
Other than that it builds and packages fine.

extrowerk:
Closes  #2598

dev-libs/icu/icu62-62.1.recipe
```diff
@@ -0,0 +1,205 @@
...
+	devel:libicudata$secondaryArchSuffix = $libVersionCompat
+	devel:libicui18n$secondaryArchSuffix = $libVersionCompat
+	devel:libicuio$secondaryArchSuffix = $libVersionCompat
+	devel:libicutest$secondaryArchSuffix = $libVersionCompat
+	devel:libicutu$secondaryArchSuffix = $libVersionCompat
+	devel:libicuuc$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	icu62$secondaryArchSuffix == $portVersion base
+	"
```
korli: A CONFLICTS_devel would be needed IMO.

extrowerk:
It needs additional gcc2 patches, thanks @Begasus for testing it.

PR #2846, opened 2018-07-22
libxcursor, bump revision after changes to util-macros and xorgproto

Begasus:


x11-libs/libxcursor/libxcursor-1.1.15.recipe
```diff
@@ -35,30 +38,28 @@ REQUIRES_devel="
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	devel:cursorproto$secondaryArchSuffix
```
Begasus: I can't find any link to cursorproto for this one?

fbrosson: I though I had created a recipe which provided a `devel:cursorproto$secondaryArchSuffix` but it turns out that there isn't any tarball that I could have used in such a recipe, so I guess this line was an accident.
Anyway, without this line the recipe builds fine, so I fixed it in d977b5c2717dbdd33f972040429555bb56d310ff.

x11-libs/libxcursor/libxcursor-1.1.15.recipe
```diff
@@ -20,7 +20,10 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libX11$secondaryArchSuffix
+	lib:libx11$secondaryArchSuffix
+	lib:libxau$secondaryArchSuffix
+	lib:libxcb$secondaryArchSuffix
+	lib:libxdmcp$secondaryArchSuffix
```
fbrosson:
```diff
-	lib:libx11$secondaryArchSuffix
+	lib:libX11$secondaryArchSuffix
-	lib:libxau$secondaryArchSuffix
-	lib:libxcb$secondaryArchSuffix
-	lib:libxdmcp$secondaryArchSuffix
```
(libxcursor does not seem to need `libxau`, `libxcb` or `libxdmcp`.)

For the `libX11` vs `libx11`, I'd keep it as it is currently in master, but I'm OK with your changing it to `libx11` if you wish.

x11-libs/libxcursor/libxcursor-1.1.15.recipe
```diff
@@ -35,30 +38,28 @@ REQUIRES_devel="
...
-	devel:libpthread_stubs$secondaryArchSuffix
-	devel:libX11$secondaryArchSuffix
+	#devel:cursorproto
+	devel:fixesproto
+	devel:kbproto
+	devel:libpthread_stubs
+	devel:libx11$secondaryArchSuffix
+	devel:libxau$secondaryArchSuffix
+	devel:libxcb$secondaryArchSuffix
+	devel:libxdmcp$secondaryArchSuffix
```
fbrosson:
```diff
-	devel:libxau$secondaryArchSuffix
-	devel:libxcb$secondaryArchSuffix
-	devel:libxdmcp$secondaryArchSuffix
```

x11-libs/libxcursor/libxcursor-1.1.15.recipe
Begasus: Ok, I'm not against it, but having a uniform matter of the libs "provided" would be a good idea then, eg ... for libXau for instance, change provides to "lib:libXau" instead of "lib:libxau" (this is one of many cases for the X11 libs ... imho "requirement" should fit the "provides" ...

x11-libs/libxcursor/libxcursor-1.1.15.recipe
Begasus: Thanks :) although now I'm of par with this PR ;)

x11-libs/libxcursor/libxcursor-1.1.15.recipe
Begasus: In my recent changes they require libxau/libxcb/libxdmcp (or should we call them libXau/libXcb/libXdmcp?) :)

x11-libs/libxcursor/libxcursor-1.1.15.recipe
Begasus: Package 'xcb', required by 'x11', not found ...

Begasus:
Should be ok now ...

Begasus:
Will be included in https://github.com/haikuports/haikuports/pull/2902 ...

PR #2850, opened 2018-07-22
libice, bump revision after changes to util-macros and xorgproto

Begasus:
Should we use "lib:libICE" or "lib:libice" here?

### DIFF
--- a/x11-libs/libice/libice-1.0.9.recipe
+++ b/x11-libs/libice/libice-1.0.9.recipe
@@ -7,7 +7,7 @@ a particular message."
 HOMEPAGE="https://www.x.org/releases/individual/lib/"
 COPYRIGHT="1993, 1998 The Open Group"
 LICENSE="MIT (no promotion)"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://www.x.org/releases/individual/lib/libICE-$portVersion.tar.bz2"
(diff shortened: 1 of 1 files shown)


Begasus:
Point taken, but it isn't the first time this has come up, if so, why not use "lib:libX*" instead of "lib:libx*" for all the other recipes then?

alaviss:
We name `lib:` after the library soname, which in this case is `libICE`, so yes, `libICE` :)

Begasus:
My point in another discussion soname for libxau is libXau, so why not change those too?

fbrosson:
_Short answer:_
It's up to you :)
I'd say that as long as it works, you can choose the variant of your choice.

_Long answer:_
I think `libICE`, `libX11` and  `libXau` look nicer than `libice`, `libx11` and  `libxau`, but I'm OK with whatever variant you might choose.
Also, you don't need to change the case of all occurrences of these `lib:lib*`, unless you really want to. You can, for example, "convert" only the lines that you already need to edit.

Begasus:
It doesn't matter what would be nicer :) It would be about setting an example on how to correctly use the sonames as @alaviss mentioned (it has been brought up earlier - eg libSDL/libsdl libGL/libgl etc ... 
It would be good to make some guidance line in what to use (I'm in favor here for libX*)

PR #2851, opened 2018-07-22
libsm, bump revision after changes to util-macros and xorgproto

Begasus:


x11-libs/libsm/libsm-1.2.2.recipe
```diff
@@ -38,18 +38,23 @@ REQUIRES_devel="
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	docbook_xml_dtd
+	docbook_xsl_stylesheets
```
alaviss: `d` is before `h` so...

```diff
-	haiku${secondaryArchSuffix}_devel
	docbook_xml_dtd
	docbook_xsl_stylesheets
+	haiku${secondaryArchSuffix}_devel
```

fbrosson: Yep, but I think there is an implicit rule which allows to put `haiku${secondaryArchSuffix}_devel` first.

Begasus: I would prefer haiku on top too ;)

alaviss: Well, it's fine if y'all prefer that :)

Begasus: Again one of those thing we should have a mutual rule for ;)

PR #2853, opened 2018-07-23
libepoxy, bump version after changes to util-macros

Begasus:
Infected by this update ... https://github.com/haikuports/haikuports/blob/bb0cc2b8eb1525b612507fca88db472248dd5748/app-office/libreoffice/libreoffice-6.2.0.0%7Egit.recipe


alaviss:
Since the library produced would pretty much be the same, so I don't think we need to bump libreoffice :)

PR #2856, opened 2018-07-24
Add recipe for BeScreenCapture 2.3.3

jackburton79:
Update BeScreenCapture to version 2.3.3.
- Fix build with recent compilers.
- Fix unstoppable capture thread on full-disk conditions.
- Disable controls when recording to avoid possible problems.
- Improved error messages.
- Improved drawing of the deskbar replicant.

### DIFF
--- a/haiku-apps/bescreencapture/bescreencapture-2.3.3.recipe
+++ b/haiku-apps/bescreencapture/bescreencapture-2.3.3.recipe
@@ -1,0 +1,76 @@
+SUMMARY="A screen recorder utility"
+DESCRIPTION="BeScreenCapture is a screen recorder utility for Haiku.
+It allows you to record what happens on your screen, then save it \
+to any media format supported in Haiku.
+BeScreenCapture can record either the entire screen, or just a section you \
+select."
+
+SUMMARY_inputfilter="Shortcut handler for BeScreenCapture"
+DESCRIPTION_inputfilter="Input Server Addon for BeScreenCapture. Allows the \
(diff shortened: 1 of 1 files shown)


fbrosson:
Thanks @jackburton79 for updating your great app!

I have already checked that it builds on all architectures, including x86_gcc2.
* It works fine for me on x86_64.
* On x86 most steps are working. It only fails at the very last step: `Could not create clip. General system error`, but this is because my x86 is very old.
* On x86_gcc2, for some reason, unknown to me, the app does not start on that arch. (When I start it from the command line, all I get is `Abort`.)

I've tested version `2.3.2` and found that it has the same issue. But when I had merged that release, in #2411, I had tested the app and noticed that it **did work fine** on x86_64 **and** x86_gcc2.
So I belive the regression is not in the app itself but in Haiku.

Now, since the pre-built package we have right now for `2.3.2` is having the issue described above, I suggest that we merge this PR.

Postponing the merge would not be of any help.

So, shall we merge?

BTW, do we merge this PR as is, i.e. as a "Add recipe for version 2.3.3"? Or do we drop `2.3.2`?
I'm OK with both scenari, although I think we don't really need `2.3.2` since we're adding `2.3.3`.

waddlesplash:
Yep, let's merge.

jackburton79:
> On x86_gcc2, for some reason, unknown to me, the app does not start on that arch. (When I start it from the command line, all I get is Abort.)

I think there could be some problems with ffmpeg. Lately they started building ffmpeg with gcc 7 on x86_gcc2, too.

Issue #1750, opened 2017-10-28
[binutils] 2.28.1 assertion fail


diversys:
Some ports like Qt4, gcc, llvm and quite a few others fail with:
`/boot/system/develop/tools/x86/bin/../lib/gcc/i586-pc-haiku/5.4.0/../../../../i586-pc-haiku/bin/ld: BFD (GNU Binutils) 2.28.1 assertion fail /sources/binutils-2.28.1/bfd/elf32-i386.c:5636`

waddlesplash:
Here's an example, thanks to @bbjimmy: http://coquillemartialarts.com/jim/libqt4-4.8.7.recipe.log

waddlesplash:
Same assert happens with binutils 2.29.

Lan72:
There is a solution https://github.com/haikuports/haikuports/issues/1794
Thanks to **diversys** personally!

waddlesplash:
well, I'm the one that did the fix 😛

Lan72:
**waddlesplash**
Then many thanks to you, as well as all those who helped solve this problem!
This is the foundation for the stable operation of web browsers based on Qt5, and many other applications on x86_gcc2.

waddlesplash:
I guess this isn't an issue anymore.

PR #2860, opened 2018-07-24
Mesa: bump to 17.1.10.

fbrosson:


kenmays:
1. All Mesa library 17.1.10 'make check' tests pass successfully.

kenmays:
Yes, please reopen. Just remove the "--enable-osmesa" in the recipe as it is mainly not used for Haiku.

waddlesplash:
@kenmays OSMesa was explicitly enabled in #2029, so it should stay enabled.

@fbrosson, please incorporate #2070 at the same time, as it fixes https://dev.haiku-os.org/ticket/13191#comment:26 ! :)

fbrosson:
I've just reverted the import of the pach from #2070, to address @waddlesplash's comment, which is still visible in: 
https://github.com/fbrosson/haikuports/commit/da1fe3054e2ff978104593d47567a7b45e9af7cf#commitcomment-29828965

> Actually we decided in the ticket that the correct way to patch would be to make the locker recursive. So, sorry, but maybe back this patch out for now, and I'll try to write a more correct one later (since @kallisti5 is too busy).

No problem :)

extrowerk:
Nothing depends on osmesa, you can remove it.

fbrosson:
@korli, @kallisti5, @waddlesplash and anyone else interesed in Mesa, do I have your green lights for removing `--enable-osmesa` ?

waddlesplash:
It was enabled without much fanfare, and is apparently broken on secondary arch, so I don't see much reason to require lots of fanfare to disable it. :)

fbrosson:
1) In `INSTALL` we currently stript `"$addOnsDir/opengl/Software Pipe"`. Shall I drop that call and, instead, include this add-on in the debuginfo package?
2) It would be easy to only disable OSMesa on secondary arches. Should I do so? Yep, I know it would require a few extra lines in the recipe, but maybe this could be worth the "trouble".

waddlesplash:
> In INSTALL we currently stript "$addOnsDir/opengl/Software Pipe". Shall I drop that call and, instead, include this add-on in the debuginfo package?

> It would be easy to only disable OSMesa on secondary arches. Should I do so? Yep, I know it would require a few extra lines in the recipe, but maybe this could be worth the "trouble".

Eh, let's disable it globally until @kallisti5 can look into it further.

fbrosson:
Oops, I noticed my typo (strip**t**) and, instead of fixing it only in my own message, I fixed it in yours :smile:

fbrosson:
Good news: the reason why all my builds of mesa_x86 (for both `17.1.7` and `17.1.10`, regardless of whether or not I enabled OSMesa) were failing was because I had built and installed `binutils_x86-2.28.1_2017_08_05`. (That version is marked as untested on x86 secondary arch, and now I understand why...)

After I downgraded to `binutils-2.26.1_2016_07_22`, which is marked as tested, all my builds of mesa succeeded.

So, since OSMesa was is guilty, I suggest that we keep it.
Of course, I can remove it if you prefer not to have it.

waddlesplash:
Yes, let's keep it then.

Yeah, that binutils version seems pretty broken on 32 bit (see #1750). I haven't spent the time trying to figure out why. We should try a newer version...

fbrosson:
Cool, thanks!
I'll install the packages and reboot to check that Haiku is happy with it, and then, maybe, I can merge, unless someone wishes to do some more testing.

fbrosson:
Haiku and RetroArch are both happy with Mesa 17.1.10, on all arches.
So let's merge!

PR #2863, opened 2018-07-25
libXt, bump revision after changes to util-macros and xorgproto

Begasus:


alaviss:
I'm wondering if `autoreconf` can be dropped ;)

PR #2867, opened 2018-07-28
libxaw, bump revision after changes to util-macros and xorgproto

Begasus:


x11-libs/libxaw/libxaw-1.0.13.recipe
```diff
@@ -83,7 +88,7 @@ defineDebugInfoPackage libxaw$secondaryArchSuffix \
 
 BUILD()
 {
-	autoreconf -vfi
+	#autoreconf -vfi
```
alaviss: If it's confirmed to be unnecessary, then we should just get rid of this line :)

Begasus: I'm still going over the libs (from the start) checking if aclocal/autoconf is called without autoreconf -vfi, if they are called for (missing) I'm still leaving it in, if nothing complaints about them missing I'll remove them once I get to them :)

Begasus:

PR #2868, opened 2018-07-28
libxrandr, bump revision after changes to util-macros and xorgproto

Begasus:


Begasus:

PR #2871, opened 2018-07-28
libxcm, bump revision after changes to xorg-proto

Begasus:


media-libs/libxcm/libxcm-0.5.4.recipe
```diff
@@ -65,7 +83,7 @@ defineDebugInfoPackage libxcm$secondaryArchSuffix \
 
 BUILD()
 {
-	autoreconf -fi
+	autoreconf -vfi
```
alaviss: Is this necessary?

Begasus: If I'm not mistaken in this case autoreconf is needed ...

Begasus:

PR #2879, opened 2018-07-29
libX11, bump after new dependency for libXdmcp

Begasus:


waddlesplash:
Needs revbump.

Begasus:
Right! Thanks (but also still working on it) :)

Begasus:
These provide the base (together with xorg-proto) for the Xlibs, needs careful checking ;)

waddlesplash:
Ready for a merge?

Begasus:
Not there yet ;)

alaviss:
I think you should split this into multiple commits (not multiple PRs) for easy tracking/reverting :)

Begasus:
I'm not too familiar with tracking/reverting, so if needed you'll here a yell from me :)

x11-libs/libx11/libx11-1.6.5.recipe
```diff
@@ -148,7 +161,7 @@ INSTALL()
 	fixPkgconfig
 
 	if [ -n "$secondaryArchSuffix" ]; then
-		rm -rf $documentationDir
+		rm -rf $documentationDir/man
```
alaviss: You could as well use `manDir` instead

Begasus: I was only following some portions that already included it, in this case it would be removed because $documentationDir is deleted, and yes, $manDir should probably fit better here :)

PR #2892, opened 2018-08-05
MPV: bump

extrowerk:
Builds fine, but complains about the locale, which needs to be C, but i was unable to hack it together. Help needed.


alaviss:
Try to add icu into `BUILD_REQUIRES`, since our locale system depends on it for locales other than C.

extrowerk:
It complains with ICU too:

"# mpv --version
Non-C locale detected. This is not supported.
Call 'setlocale(LC_NUMERIC, "C");' in your code."

alaviss:
Oh, I understand now. MPV only [supports][0] the "C" locale. I think mpv called `setlocale(LC_ALL, "C")`, but it wasn't picked up by our locale system.

The correct walkaround should be to supply a wrapper calling `LC_NUMERIC="C" mpv`.

[0]: https://github.com/mpv-player/mpv/commit/21322a016e74ddf20cffaab5d621db1cf694eaa5

pulkomandy:
The correct way is to open a bugreport at Haiku about setlocale not working.

extrowerk:
Afaik mpv calls no setlocale, but i tried to add it before this line:
https://github.com/mpv-player/mpv/blob/master/player/main.c#L440
AFAIK this is the main().

I did it like:
```setlocale(LC_NUMERIC, "C");```
But no dice. So i assume setlocale is broken in Haiku.

PR #2901, opened 2018-08-09
sys-apps/man-pages-posix: add version 2013.a

alaviss:
Bringing more man pages to Haiku :) These manual pages will greatly aid developers working on Haiku.

Recipe may need some styling adjustment however. And may be check my grammar as well :P


alaviss:
Latest changes:
- Changed "may" to "might", thanks @humdingerb for the input :)

PR #2902, opened 2018-08-09
libX*, bump revision, use correct REQUIRES_devel

Begasus:


x11-libs/libxau/libxau-1.0.8.recipe
```diff
@@ -6,7 +6,7 @@ LibXau is used by both Xlib and XCB."
 HOMEPAGE="https://www.x.org/"
 COPYRIGHT="1988, 1993, 1994, 1998  The Open Group"
 LICENSE="MIT (no promotion)"
-REVISION="5"
+REVISION="7"
```
alaviss: Is that a bit too big for a jump?

Begasus: Seems I needed to update the local master for haikuports too, fix in a minute :)

Begasus:
Could use some feedback on the cleanup for libX11, seems the libraries libXau and libXdmcp are not needed for the libX11 build (and some of the proto headers) ...

alaviss:
If you drop them and it still work, then they should be dropped

Begasus:
I'll leave them in (commented out) for now in case someone decides they are actually needed ...

alaviss:
> I'll leave them in (commented out) for now in case someone decides they are actually needed ...

One could look at the commit to get them back :) So there should be no need to do so

x11-libs/libxau/libxau-1.0.8.recipe
```diff
@@ -54,21 +55,22 @@ defineDebugInfoPackage libxau$secondaryArchSuffix \
...
 	make install
+	install -d -m 755 "$docDir"
+	install -m 644 -t "$docDir" README
 
 	prepareInstalledDevelLib libXau
 	fixPkgconfig
 
 	if [ -n "$secondaryArchSuffix" ]; then
-		rm -rf $documentationDir
+		rm -rf $documentationDir/man
```
alaviss:
```diff
-		rm -rf $documentationDir/man
+		rm -rf $manDir
```

x11-libs/libxcb/libxcb-1.13.recipe
```diff
@@ -171,7 +176,7 @@ INSTALL()
 	fixPkgconfig
 
 	if [ -n "$secondaryArchSuffix" ]; then
-		rm -rf $documentationDir
+		rm -rf $documentationDir/man
```
alaviss:
```diff
-		rm -rf $documentationDir/man
+		rm -rf $manDir
```

Begasus:
Guess these should do it, checked for x86_gcc2 and x86(secondaryArchSuffix) ...

PR #2912, opened 2018-08-12
Bochs x64 support

extrowerk:


app-emulation/bochs/bochs-2.6.9.recipe
```diff
@@ -21,18 +21,31 @@ REQUIRES="
...
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+BUILD_REQUIRES+="
+	devel:libgfortran$secondaryArchSuffix
+	"
+fi
 BUILD_PREREQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	cmd:awk
 	cmd:gcc$secondaryArchSuffix
+	cmd:gfortran$secondaryArchSuffix
```
extrowerk: Well, this is not correct... Needs to be guarded with effectiveTragetArch too.

PR #2924, opened 2018-08-16
hxtools, bump version

Begasus:


dev-util/hxtools/patches/hxtools-20180301.patchset
```diff
@@ -131,5 +97,49 @@ index 2c973a1..88d3a73 100644
...
+
+diff --git a/smm/aumeta b/smm/aumeta
+old mode 100755
+new mode 100644
+index 2b6959d..af76eac
+--- a/smm/aumeta
++++ b/smm/aumeta
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/bin/env perl
```
alaviss: The more "universal" path should be `/usr/bin/env perl`. It's supported by us and some *nix out there with weird `PATH` as well

Begasus: Bumped :)

dev-util/hxtools/hxtools-20180301.recipe
```diff
@@ -63,15 +68,15 @@ years, and each of which seems to be too small to warrants its own project.
 - wktimer(1) — work timer
 "
 HOMEPAGE="http://inai.de/projects/hxtools/"
```
alaviss:
```diff
-HOMEPAGE="http://inai.de/projects/hxtools/"
+HOMEPAGE="https://inai.de/projects/hxtools/"
```

dev-util/hxtools/hxtools-20180301.recipe
```diff
@@ -96,11 +102,15 @@ PROVIDES="
 	cmd:git_track$secondaryArchSuffix
```
alaviss: I think we should use `commandSuffix` for this port. On `x86_gcc2` these commands will not work as expected. They're git plugins, which means they are usually called like this:

    git track # yes, no dash

with `secondaryArchSuffix`, it would be:

    git track-x86

which certainly no one expects.

PR #2925, opened 2018-08-16
libexif, add libVersion(Compat), add defineDebug…

Begasus:
…InfoPackage

### DIFF
--- a/media-libs/libexif/libexif-0.6.21.recipe
+++ b/media-libs/libexif/libexif-0.6.21.recipe
@@ -9,7 +9,7 @@ COPYRIGHT="2012 Lutz Mueller
 	Hubert Figuiere
 	"
 LICENSE="GNU LGPL v2.1"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://sf.net/projects/libexif/files/libexif/0.6.21/libexif-$portVersion.tar.gz"
 CHECKSUM_SHA256="edb7eb13664cf950a6edd132b75e99afe61c5effe2f16494e6d27bc404b287bf"
 if [ $effectiveTargetArchitecture = x86_gcc2 ]; then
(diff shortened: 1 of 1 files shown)


media-libs/libexif/libexif-0.6.21.recipe
```diff
@@ -19,9 +19,12 @@ fi
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion="12.3.3"
```
fbrosson: Suggestion:
```diff
-libVersion="12.3.3"
+libVersion=12.3.3
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```

BTW, for the commit message I'd suggest this variant:
```diff
-libexif, bump revision, use libVersion(ComPat), introduce defineDebugInfoPagackage
+libexif: use libVersion(Compat), add debuginfo package
```
(That's because the `bump revision` is implicit in most cases.)
Feel free to keep the `introduce defineDebugInfoPagackage` as is if you wish.
(I shortened it to make it <= 69 chars.)

Hmm, I did not run
```bash
readelf -S libexif.so.12.3.3 | grep debug
```
to see if the compiler is being called with `-g`, but I guess you already checked this.

Begasus: updated, I didn't use readelf, but did just earlier, can't tell what it says though, I see that the list is shorter from the source lib then the installed lib ...

fbrosson: Oops, I forgot to mention that the `readelf -S libexif.so.12.3.3 | grep debug` should be run on a target which has not been stripped. So, as you probably noticed, you need to either comment-out the `defineDebugInfoPackage` (if you want to run the command on the installed lib), or find the lib in the build tree (and use it for the test.)

fbrosson: funny typo in the commit message ;)
```diff
-libexif, add libVersion(Compat), add defineDebugInfoPagackage
+libexif, add libVersion(Compat), add defineDebugInfoPackage
```

Begasus: figured the stripped lib wouldn’t produce the same output as the one in the build tree :)

PR #2932, opened 2018-08-17
libfm-qt, add libVersion(Compat), add defineDebugInfoPackage

Begasus:


x11-libs/libfm-qt/libfm_qt-0.13.0.recipe
```diff
@@ -74,14 +83,9 @@ INSTALL()
 	cd build
 	make install
 
-	prepareInstalledDevelLibs libfm-qt
+	prepareInstalledDevelLib libfm-qt
 	fixPkgconfig
 
 	packageEntries devel \
 		$developDir
```
fbrosson:
```diff
 	packageEntries devel \
-		$developDir
+		"$developDir"
```
(Many recipes were recently modified, with cosmetic/neutral edits only, so I'd say an extra cosmetic edit here might be acceptable :)

A few weeks ago I would not have suggested this change, but it looks like if we don't start protecting these variables (when we edit a recipe for some reason) then the trend will not even start to change.

Begasus: Need to make myself notice these things :) (on x86_64 atm, will update later)

PR #2945, opened 2018-08-22
WxQT: bump, debuginfo, versioncompat

extrowerk:


extrowerk:
I should disable OGL, right?

waddlesplash:
LGTM

dev-qt/wxqt/wxqt-3.1.1.recipe
```diff
@@ -120,7 +152,7 @@ PATCH()
 
 BUILD()
 {
-	export CXXFLAGS="-std=c++11"
+	export CXXFLAGS="-std=c++11 -fpermissive"
```
alaviss: I'm a bit skeptical about adding `-fpermissive`, as it allows non-conforming code to compile. If upstream doesn't do this by default, then I think the problem is in our patches.

I don't think explicitly setting c++11 is necessary.

waddlesplash: It was when we were on GCC5. And no, the Qt port is a bit experimental, so this is probably ok, I think?

PR #2947, opened 2018-08-23
OpenMPI: Bump, added bison, enabled libtoolize, all test passes

extrowerk:


extrowerk:
Oh, it seems i forgot to add the actual recipe. lemme do it quick..

PR #2951, opened 2018-08-26
ibm_plex, fix checksum

Begasus:

PR #2954, opened 2018-08-26
IBM Plex: updated checksum, revbump

extrowerk:


extrowerk:
Bega was faster: https://github.com/haikuports/haikuports/pull/2951/files

Issue #2933, opened 2018-08-18
[RFC] R1beta1 branch policy


waddlesplash:
So, R1beta1 is fast approaching, and we need to decide how we are going to conduct the `r1beta1` branch here. For Haiku itself, the `r1beta1` branch will probably get only bug fixes as they come, without a lot of major changes that will probably hit master. HaikuPorts should probably do something similar; but the question is exactly what?

I think there are 3 general options:

1. **Security updates only.** This is the "debian"/"ubuntu" model: apply patches to only those packages that have published CVEs and patches for them. Do not bump even for minor bugfixes.
2. **Minor bug-fix updates.** Everything from # 1, but also include minor bug fix updates (e.g. `4.0.1 -> 4.0.2` of some software, assuming it follows Semantic Versioning.) This is what openSUSE does (? I think?)
3. **Moderate feature updates.** Everything from # 2, but also include moderate feature updates (e.g. `4.0.1 -> 4.1.0` of some software, assuming it follows Semantic Versioning.) This is what CentOS does (though they usually mark point releases when they do this.)

I'm heavily in favor of the second option. Thoughts, @fbrosson, @alaviss?

waddlesplash:
oh, and @extrowerk too.

pulkomandy:
On Fri, Aug 17, 2018 at 09:04:41PM -0700, waddlesplash wrote:
> So, R1beta1 is fast approaching, and we need to decide how we are going to conduct the `r1beta1` branch here. For Haiku itself, the `r1beta1` branch will probably get only bug fixes as they come, without a lot of major changes that will probably hit master. HaikuPorts should probably do something similar; but the question is exactly what?

I don't agree on this for Haiku, the way I see it, the beta1 branch
there will eventually lead to R1. If beta1 is a throwaway branch and R1
is built from trunk, then beta1 isn't really a beta.

Anyway, let's focus on Haikuports here :)

> 
> I think there are 3 general options:
> 
> 1. **Security updates only.** This is the "debian"/"ubuntu" model: apply patches to only those packages that have published CVEs and patches for them. Do not bump even for minor bugfixes.
> 2. **Minor bug-fix updates.** Everything from # 1, but also include minor bug fix updates (e.g. `4.0.1 -> 4.0.2` of some software, assuming it follows Semantic Versioning.) This is what openSUSE does (? I think?)
[...]

PR #2958, opened 2018-08-26
OpenSceneGraph: patchset fix

extrowerk:


extrowerk:
https://github.com/haikuports/haikuports/issues/2933#issuecomment-416015069

PR #2961, opened 2018-08-26
net-libs/gnutls: drop older versions

alaviss:
The ABI doesn't seem to be broken between these versions.

https://abi-laboratory.pro/index.php?view=timeline&l=gnutls

Tag #2933


alaviss:
We can keep both, but they conflicts, and `pkgman` will unconditionally pull the later version, since it's compatible with everything built with the old one.

If we want super stability then we can keep 3.5 and drop (or disable) 3.6. We have to decide which one will be installed by default.

fbrosson:
You are right. We need to force one of these releases to be the default. So maybe we should mark `3.6.3` as untested.

waddlesplash:
... gnutls36 was the current version and you disabled it. :-p

alaviss:
@waddlesplash

You could have done a rebase-merge instead.

Reasons for disabling 3.6.3 is presented above ;)

waddlesplash:
Yes, but other packages were already built against 3.6. So that makes more sense to keep.

PR #2965, opened 2018-08-26
Mesa 18.1

kenmays:


waddlesplash:
Does it work?

kenmays:
I've tested the Mesa 18.1.7 libs and everything now compiles cleanly in Haiku. The mesa-swpipe driver component needs approval by kallisti5 as suggested in ticket: https://dev.haiku-os.org/ticket/13191

PR #2981, opened 2018-08-31
R: bump

extrowerk:


scottmc:
Does it pass all of the tests?

extrowerk:
@scottmc : sadly no, but i can't seem to find the reason.

extrowerk:
Disabling cairo and pango doesn't changes anything, test breaks, but there is no clear output, what's the problem is. Here is  the testlog with cyairo and pango enabled:
https://gist.github.com/extrowerk/e18ca0f7bf4be19fd8cb0e0901d36596

PR #2983, opened 2018-09-01
dev-python/requests: requires certifi

alaviss:


alaviss:
I'm gonna update requests instead

PR #2985, opened 2018-09-01
Retroarch: bump to version 1.7.4

kwyxz:
Recipe changes:
- partial revert of #2794 : version number in recipe breaks build (at least for me...)
- bump to new version
- added patch file (patch submitted upstream simultaneously)

Upstream changes relevant to Haiku:
- automatically hide "Configuration Override options" in Quick Menu.
- small Bugfix to not trigger savestate code when pressing Reset.
- enable SSL/TLS support by default for desktop platforms.


games-emulation/retroarch/retroarch-1.7.4.recipe
```diff
@@ -6,11 +6,12 @@ shaders, netplay, rewinding, next-frame response times, and more!"
...
 COPYRIGHT="2010-2018 The RetroArch Team"
 LICENSE="GNU GPL v3"
-REVISION="3"
+REVISION="1"
 SOURCE_URI="https://github.com/libretro/RetroArch/archive/v$portVersion.tar.gz"
-CHECKSUM_SHA256="a60c2244609bb87cdb56dd8e1020c3be757569b5246141328804ebc5574327ea"
+CHECKSUM_SHA256="1c095076e2a9d2ec4e501176b3963ea2a831c1e5f025bb25386491272f14de40"
 SOURCE_FILENAME="retroarch-$portVersion.tar.gz"
 SOURCE_DIR="RetroArch-$portVersion"
+PATCHES="retroarch-1.7.4.patchset"
```
extrowerk: You should use $portVersion istead the version number

kwyxz: I considered it, however this patchset will be gone with the next release. My patch was already accepted upstream, hence there wouldn't be any real benefit. If you really think it's necessary I can still to it though.

fbrosson: I think it can be a goot idea to use `$portVersion` because when `1.7.5` becomes available, it might need another patch (not related to the one you have upstreamed), so this line would already be OK.

BTW, except for some special cases, we  tend to use `$portVersion` :)

fbrosson:
`1.7.4` builds fine on x86_64 and seems to work, although I'm having 2 minor issues:

1) I'm sometimes getting auto-repeats for keys which I'm not holding. But this could be due to my VM.

2) Some icons (in the menus) are replaced by black squares. But this could be due to my retroarch config.
In the first menu, for instance, **_Load Core_**, **_Load content_** and **_Configurations_** have their icons, while **_Online Updater_**, **_Information_**, **_Help_** and **_Quit RetroArch_** have black squares.
In the second menu the only items which have icons are: **_Input_**, **_Configuration_** and **_Directory_**. All other items (e.g. **_Driver_**, **_Video_**, **_Mixer Settings_**) have black squares instead of an icon.

(I'll try to build `1.7.4` on x86_gcc2h with x86 secondary arch.)

kwyxz:
1. Not familiar with this but it could be a bug. Does it happen with specific keys?
2. The icons are definitely a theme issue. Try going to Settings / User Interface and choose a different icon theme. I do not know if there is a consensus on the new default but Retroactive should be pretty complete and not have black squares. I'll check if we should update retroarch-assets as well.

kwyxz:
@fbrosson confirmed theme issue when using theme "Monochrome", this PR solves it #2990 
I could not see it as I am using a different theme that was not missing any art

kwyxz:
I build on x86_64 but I want to make sure what your issue is: does it happen directly with the menu? And when you press several times the right arrow it jumps several menus? As far as I can tell, I am not experiencing this.

fbrosson:
Yep, it's happening with the menu while I navigate with the arrow keys. Most of the time my single presses are correctly precessed as single presses. But occasionally a single press is seen as a press+hold-down. When this happens it's like if I were pressing the key and holding it down forever. I can make it stop by pressing that same key again (or any other key.)

fbrosson:
(Yep, it did not happen with `1.7.3`.)

I updated my Haiku x86_64 to a recent hrev and also switched to QEMU+KVM (instead of QEMU+KVM).
I managed to reproduce the autorepat bug once, but was unable to reproduce it twice.
I suspect that this is a timing issue: two consecutive key presses (with a particular time interval between these two key presses) can be seen as a single key press +  hold down.

Well, since this tiny bug does not seem to be easy to reproduce I suggest that we merge this PR.
(If anyone would prefer to investigate before the merge, please let us know.)

games-emulation/retroarch/patches/retroarch-1.7.4.patchset
```diff
@@ -0,0 +1,48 @@
+From a8b885b119c3824f4ca71e1b5cfa8869436705db Mon Sep 17 00:00:00 2001
+From: Benjamin FRANCOIS <kwyxz@kwyxz.org>
+Date: Sat, 1 Sep 2018 07:29:49 +0000
+Subject: Added simple define to allow build on Haiku
+
+
+diff --git a/deps/mbedtls/entropy_poll.c b/deps/mbedtls/entropy_poll.c
+index a116e60..4a913aa 100644
+--- a/deps/mbedtls/entropy_poll.c
```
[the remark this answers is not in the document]
kwyxz: Hello @alaviss - I asked upstream and even though the mbedtls bundled has been rather heavily modified (see https://github.com/libretro/RetroArch/commits/master/deps/mbedtls) there's according to them a chance it won't pose a problem with Haiku. It however would need to add an option to the ./configure script to enable using the system lib. I'll need time to look into that, unless somebody is willing to give it a try!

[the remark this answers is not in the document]
alaviss: In that case then it's fine to not use the system libraries :) It's recommended to use system libraries when upstream bundles vanilla copies of it's dependencies, but it's certainly not the case here.

fbrosson:
I've just restarted my Linux box with Xen (to be able to run Haiku with QEMU+Xen) and it turns out that it is slighly easier to hit the autorepeat bug with Xen than it is with KVM.

Anyway, since nobody seems to object a merge, let's do it!

PR #2991, opened 2018-09-02
TOR: config folder location

extrowerk:


net-vpn/tor/patches/tor-0.3.3.9.patchset
```diff
@@ -35,5 +35,45 @@ index 9c88d37..74f2e48 100644
...
++++ b/src/or/config.c
+@@ -5024,7 +5024,11 @@ find_torrc_filename(config_line_t *cmd_arg,
+ #ifndef _WIN32
+       char *fn = NULL;
+       if (!defaults_file) {
+-        fn = expand_filename("~/.torrc");
++        #ifndef __HAIKU__
++          fn = expand_filename("~/.torrc");
++        #else
++          fn = expand_filename("~/config/settings/tor/torrc");
```
alaviss: Please consider using `find_directory()` instead.

net-vpn/tor/patches/tor-0.3.3.9.patchset
```diff
@@ -35,5 +35,45 @@ index 9c88d37..74f2e48 100644
...
+         file_status_t hmst = file_status(fn);
+@@ -7749,7 +7753,11 @@ get_data_directory(const char *val)
+ #else /* !(defined(_WIN32)) */
+   const char *d = val;
+   if (!d)
+-    d = "~/.tor";
++    #ifndef __HAIKU__
++      d = "~/.tor";
++    #else
++      d = "~/config/settings/tor";
```
alaviss: IMO `B_USER_VAR_DIRECTORY` might be better suited for storing data from applications.

net-vpn/tor/patches/tor-0.3.3.9.patchset
scottmc: I second that request.  Also I'd change the logic here to ```#ifdef __HAIKU__``` and if not let it fall thru to the else for the defaul case.  This makes it easier for other OSes that might also need to move the settings file, to then add a case for their OS as well.

net-vpn/tor/patches/tor-0.3.3.9.patchset
scottmc: same logic comment as about...  change to ```ifdef __HAIKU__``` and if not let the else be the default.

scottmc:
@fbrosson i see you bumped the version a couple days ago.  can you review this open pull request to see if anything in here should have been included in your update?

net-vpn/tor/patches/tor-0.3.3.9.patchset
fbrosson: Same comments as @alaviss and @scottmc :)

extrowerk:
I never had too much luck using find_directoy(), that's why i never using it.

scottmc:
Ok, then let's close this PR, and instead open a new issue to solve this directory issue using find_directory().  This might be a good GCI task.

PR #2993, opened 2018-09-02
libcdio-paranoia, new recipe

Begasus:


Begasus:
Not tested on x86_64 yet ...

PR #3000, opened 2018-09-04
Vim 8.1 update 

kenmays:
Needs patchset review for haikuport import/regeneration check.
Built and tested vim 8.1 on hrev52300 x86_gcc2.
Migrating all legacy vim patchsets and recipes to current (vim 8.1.0346, Sept 2, 2018).


extrowerk:
Patch doesn't applies cleanly.

PR #3001, opened 2018-09-04
Mplayer: bump to snapshot

extrowerk:
Latest stable version crashes.
I had to mirror the sources on github, as the snapshot aren't versoined and not kept.

Works nicely on x86_64 but untested on x86. But it should build fine.


scottmc:
Please submit the patch upstream as a pull request and follow up on it as needed.

extrowerk:
@scottmc : Did long time ago, but the project is in "sleeping" state.

scottmc:
It looked to me that they were keeping up on pull requests.  Also keep in mind that sometimes smaller pull requests are easy to get added.  So maybe consider breaking it up in smaller parts if it's too much at once.

scottmc:
> @scottmc : Did long time ago, but the project is in "sleeping" state.
> Correction: I sent the patches to one dev, and just checked their ML (what i haven't did before): they would like to see some changes here and there:
> http://lists.mplayerhq.hu/pipermail/mplayer-dev-eng/2017-September/073641.html
> Let's see if i can do it.
> 
> @diversys : afaik the code is from you. Are you OK with license change? (for more info check the ML thread)

sounds like a reasonable change.  be sure to get permissions from all that contributed to the patch.

PR #3004, opened 2018-09-05
Mesa cleanup

kenmays:
Cleans up legacy Mesa recipes/patches not fully tested, not working, or unused on Haiku as of Sept 2018..
Mesa 17.1.10 and Mesa 7.9.2 are tested as working for Haiku R1B1.


extrowerk:
Let's ask @kallisti5 if they really aren't needed.

kallisti5:
toss em.  We had some issues with more recent mesa's not working, but if we have a functional port currently, they really aren't needed.

If we do need them for historical reasons we can get them from the git history.

PR #3009, opened 2018-09-06
curl: use libnghttp2.

fbrosson:


fbrosson:
curl is working like a charm with libnghttp2. Tested with `curl --http2 ...`

Reminder: if this gets merged then nghttp2 will need to be included in the default images.
I'm pretty sure there won't be any downsides but I'll only merge this if everyone is OK with this new dependency.

scottmc:
@korli any issues with this addition as a new dependency?

PR #2768, opened 2018-07-01
libcdio, bump version

Begasus:

PR #3013, opened 2018-09-08
Revbump recipes depending on libcdio (new version)

Begasus:
Depending on https://github.com/haikuports/haikuports/pull/2768


fbrosson:
This is cool! Thanks @Begasus!

Reminder to the member who will merge this PR:

1) Please make sure #2768 has been merged.
(#2768 is not merged yet because we don't know yet if all the recipes being rev-bumped in this one, #3013, will build fine with the new release of libcdio.)
Hmm, @Begasus, in case you (or someone else) have already checked this, then we can merge #2768 :)

2) Please make sure the buildbots have finished building libcdio. In case something went wrong and we merge #3013 without having the new libcdio package then we would need to repeat the same rev-bumps...

3) **Most important:** The prefered merge method for this PR, #3013, is **"Rebase and merge"**. So please **do not** use "Squash and merge" here. Also remember that the default merge method for a given user is the last merge method used by that user. This means anyone doing a **"Rebase and merge"** should remember to select "Squash and merge" for the next PR (if that is what this user wants, of course.)

Begasus:
I've been trying to build avidemux and some of the other ones, but as they require quite some resources I couldn't finish to the end, system keeps freezing up, so it would be nice if someone with the right HW would be able to check this ...

PR #3019, opened 2018-09-09
Guile: bump

extrowerk:


dev-scheme/guile/guile-2.2.4.recipe
```diff
@@ -9,7 +9,7 @@ LICENSE="GNU LGPL v3"
 REVISION="1"
 SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.gz
 	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.gz"
-CHECKSUM_SHA256="87ee07caef33c97ddc74bf3c29ce7628cfac12061f573e4a29a3a1176754610a"
+CHECKSUM_SHA256="33b904c0bf4e48e156f3fb1d0e6b0392033bd610c6c9d9a0410c6e0ea96a3e5c"
```
fbrosson: Suggestion: switch to `tar.xz`
```diff
-SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.gz
+SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.xz
-	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.gz"
+	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.xz"
-CHECKSUM_SHA256="33b904c0bf4e48e156f3fb1d0e6b0392033bd610c6c9d9a0410c6e0ea96a3e5c"
+CHECKSUM_SHA256="d9e8b94af7b206fcf52bec6501b921bd7d0bd7a31fb7e896a35ba8253678e31e"
```
I checked that the tarballs have the exact same contents:
```
cmp <(gzip -cd guile-2.2.4.tar.gz) <(xz -cd guile-2.2.4.tar.xz) ; echo $?
0
```

fbrosson: Hmm, the `tar.lz` is even smaller than the `tar.xz` (8.8 MiB instead of 10MiB), so you can even switch to `tar.lz`.

```diff
-SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.gz
+SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.lz
-	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.gz"
+	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.lz"
-CHECKSUM_SHA256="33b904c0bf4e48e156f3fb1d0e6b0392033bd610c6c9d9a0410c6e0ea96a3e5c"
+CHECKSUM_SHA256="b72724cb59d4cc1eed005a10a960831efe0f1f7bf17cc6a88fbfab0a48b4bf5c"
```
I checked that the tarballs have the exact same contents:
```bash
cmp <(gzip -cd guile-2.2.4.tar.gz) <(lzip -cd guile-2.2.4.tar.lz); echo $?
0
```
(haikuporter can handle `tar.lz` archives if `lzip` is available. And it is on all platforms except x86_gcc2 non hybrid.)

fbrosson:
`2.2.4` builds OK on both x86_64 and x86, and passes most tests, although a few crash.
(I guess `2.2.4` passes more tests than `2.2.3` and crashes less, but I did not check this.)

PR #3039, opened 2018-09-11
libdsk, bump version, add libVersion(Compat) and defineDebugInfoPackage

Begasus:


app-emulation/libdsk/libdsk-1.5.8.recipe
```diff
@@ -12,66 +12,72 @@ controller emulation, thus giving the emulator transparent access to .DSK \
...
+libVersion="4.3.1"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
 	libdsk$secondaryArchSuffix = $portVersion
-	lib:libdsk$secondaryArchSuffix = 4.3.0 compat >= 4
+	lib:libdsk$secondaryArchSuffix = $libVersionCompat
 	"
+if [ -z "$secondaryArchSuffix" ]; then
+	PROVIDES_tools="$PROVIDES
```
fbrosson:
```diff
 if [ -z "$secondaryArchSuffix" ]; then
- 	PROVIDES_tools="$PROVIDES
+	PROVIDES_tools="
+		libdsk${secondaryArchSuffix}_tools = $portVersion
		cmd:apriboot
		cmd:dskconv
```

app-emulation/libdsk/libdsk-1.5.8.recipe
```diff
@@ -12,66 +12,72 @@ controller emulation, thus giving the emulator transparent access to .DSK \
...
+		cmd:dskconv
+		cmd:dskdump
+		cmd:dskform
+		cmd:dskid
+		cmd:dskscan
+		cmd:dsktrans
+		cmd:dskutil
+		cmd:md3serial
+		"
+	REQUIRES_tools="
```
fbrosson: I did not check this, but I guess the tools will use the libs from the base package, so:
```diff
 	REQUIRES_tools="
+		libdsk$secondaryArchSuffix == $portVersion base
 		haiku
```

The rest looks OK, although I did not test the recipe.

BTW, it looks like double quotes are now much more welcome, so I'd recommend these cosmetic changes:
* `$developDir` :arrow_right: `"$developDir"`
* `$binDir` :arrow_right: `"$binDir"`

app-emulation/libdsk/libdsk-1.5.8.recipe
fbrosson: I just realised that we can (and probably should) use the following variant, instead of my first suggestion above:
```diff
 if [ -z "$secondaryArchSuffix" ]; then
- 	PROVIDES_tools="$PROVIDES
+	PROVIDES_tools="
+		libdsk_tools = $portVersion
		cmd:apriboot
		cmd:dskconv
```
(Indeed, we can drop the `${secondaryArchSuffix}` because `PROVIDES_tools` is only being defined if `$secondaryArchSuffix` is empty.)

PR #3060, opened 2018-09-14
jack2, bump version, remove bep file

Begasus:


Begasus:
Build is ok, but when packaging it ends up with:

  mimesetting files for package jack2_x86-1.9.10~git-1-x86_gcc2.hpkg ...
  creating package jack2_x86-1.9.10~git-1-x86_gcc2.hpkg ...
  heap size:                   813074
  TOC size:                      4707
  package attributes size:       1491
  total size:                  813154
  Warning: 'NoneType' object has no attribute 'split'

media-sound/jack2/jack2-1.9.10~git.recipe
```diff
@@ -0,0 +1,110 @@
...
+	python ./waf build
+}
+
+INSTALL()
+{
+	python ./waf build install
+
+	mkdir -p "$includeDir"
+	mv "$prefix"/include/jack/ "$includeDir"
+	rmdir "$prefix"/include
```
Begasus: If I leave it out it ends up with a POLICY ERROR for the "include" folder, keeping it in seems to remove the whole base package in the "packaging" folder, only the devel package is in there then ...

waddlesplash:
We should not port jack2, but instead use libjackcompat. Running a separate media services system that has most of the same features as our built in one would be confusing.

Begasus:
Ok, then this recipe and the bep file could be removed all together?

Begasus:
Marked as "!" atm, feel free to close this one if it's not needed ;)

PR #3070, opened 2018-09-15
Nasm: bump version

kenmays:
Built and tested on hrev52339 x86_gcc2.


alaviss:
Is there any reason why the `rc` version was used instead? I think we would prefer stable version over a release candidate.

kenmays:
Covered small bug fixes to certain instructions after the NASM 2.12.02 release.

alaviss:
Official [NASM 2.13.03][0] was out on the 7th of February this year, please use that instead.

[0]: https://www.nasm.us/pub/nasm/releasebuilds/2.13.03/

kenmays:
Reverted to NASM 2.12.02 - since I've built and tested and built this beforehand on GCC2..

dev-lang/nasm/nasm-2.12.02.recipe
```diff
@@ -7,13 +7,11 @@ designed to be simple and easy to understand, similar to Intel's but less \
 complex. It supports all currently known x86 architectural extensions, and \
 has strong support for macros."
 HOMEPAGE="https://www.nasm.us/"
-COPYRIGHT="1996-2001 Simon Tatham, Julian Hall
-	 2002-2016 H. Peter Anvin, Jim Kukunas, Cyrill Gorcunov, Frank B. \
-Kotler"
+COPYRIGHT="Copyright 1996-2016 the NASM Authors - All rights reserved."
```
fbrosson: We don't include the word `Copyright` nor the `All rights reserved.` :)
```diff
-COPYRIGHT="Copyright 1996-2016 the NASM Authors - All rights reserved."
+COPYRIGHT="1996-2016 the NASM Authors"
```
Hmm, do we keep the `the` before `NASM Authors` ?
(I mean, would `"1996-2016 NASM Authors"`be better?)

scottmc: yes, i'd go with just "1996-2016 NASM Authors"

korli: not resolved

alaviss:
> On x86_gcc2 non hybrid we don't have xz.

I thought we don't support x86_gcc2 non-hybrid? Otherwise, shouldn't `cmd:xz` be available?

pulkomandy:
x86_gcc2 non-hybrid builds are still possible at Haiku side, but the difference is meaningless now that we have package management (basically you can pkgman install haiku_x86 and your build will become an hybrid one. Or even better, pkgman install cmd:xz and let it pull in the dependencies). And even if one doesn't want to do that, they can still download the hpkg file and use it. They just can't build it themselves.

gcc2-only builds are still available for experimentation purposes, but I don't expect haikuports to support them in any way. You can ignore them.

fbrosson:
OK, thanks @pulkomandy. That's good to know!
(So this means we have green lights to switch the `SOURCE_URI` of most recipes (if not all) to `tar.xz`.)

Sorry @kenmays for having reverted your switch to `tar.xz`...

PR #3078, opened 2018-09-18
schismtracker, add autoreconf, add Deskbar symlink

Begasus:


media-sound/schismtracker/schismtracker-20120105.recipe
```diff
@@ -49,9 +49,11 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	aclocal
+	autoreconf -vfi
 	runConfigure ./configure
 	make $jobArgs
+
+	addAppDeskbarSymlink "$binDir"/schismtracker SchismTracker
```
korli: addAppDeskbarSymlink usually goes in INSTALL().

Begasus: jikes ;)

PR #3080, opened 2018-09-18
beshare, bump version

Begasus:


korli:
It should be a version bump (2.33)

https://github.com/HaikuArchives/BeShare/archive/v2.33.tar.gz

Begasus:
Will fix later :)

Begasus:
new tar.gz doesn't include the gcc7 fixes ...

Begasus:
So I could update the version but still use the srcGitRev?

korli:
The recipe doesn't seem to reference the patchset.

Begasus:
> The recipe doesn't seem to reference the patchset.

It wasn't suppose to @korli , it was to point out the wrong version being declared, fixed in the new tag release :)

korli:
thanks!

PR #3089, opened 2018-09-19
hexvexed, bump revision after new beta

Begasus:


Begasus:
Recipe is working ok, and the background is listed in the menu Entry, but when selecting the background doesn't change, but that would be a fix in the source I presume :)

scottmc:
Does it show the red/orange background, or is it just default?

Begasus:
> Does it show the red/orange background, or is it just default?

It keeps with the default, but after a restart of the app you see a (very) quick glimpse of the background red/orange

humdingerb:
The problem is that the BackView still draws the blue colour , so the background image isn't seen.
I submitted a PR that does away with the "None" background colour and always shows a background image. I added a few more of those and have Hexvexed look for those in on eof B_FIND_PATH_DATA_DIRECTORY. The recipe should put those into $dataDir/hexvexed/backgrounds.

The PR is at https://bitbucket.org/scottmc/hexvexed/pull-requests/8

Begasus:
@scottmc  and @humdingerb , did the PR got into the zip, or would it be best to do a srcGitRev recipe to import the latest changes?

scottmc:
try the beta3 for the updated source.

haiku-games/hexvexed/hexvexed-1.0.0.recipe
```diff
@@ -6,11 +6,10 @@ of squares, the tiles are hexagon shaped. It's 1.5x the fun!"
 HOMEPAGE="https://bitbucket.org/scottmc/hexvexed"
 COPYRIGHT="2009-2017 Scott McCreary"
 LICENSE="MIT"
-REVISION="3"
-SOURCE_URI="https://bitbucket.org/scottmc/hexvexed/get/1.0.0-Beta.zip"
-CHECKSUM_SHA256="d6f9509d6fed6fc0967ec2e095c3874dbe01905395592bfa68bffa2a58341299"
-SOURCE_DIR="scottmc-hexvexed-b8a523a7bf04"
-PATCHES="hexvexed-$portVersion.patchset"
+REVISION="4"
+SOURCE_URI="https://github.com/HaikuArchives/HexVexed/archive/1.0.0-Beta2.tar.gz"
```
scottmc: Beta3 now...

Begasus: bumped recipe to beta3, half way there, images are stored in $dataDir/hexvexed/backgrounds (and are present) but still don't show up, if I copy them to ~/config/non-packaged/... it works

Begasus: https://ibb.co/dPsBBe

humdingerb: I guess I'll have to take another look. I have to admit, I didn't try installing HexVexed to /system and if it finds the backgrounds in the data folder there...

Also, shouldn't the settings always be in ~/config/settings/ ?

scottmc:
maybe try it with backgrounds in appsdir/hexvexed/backagrounds instead of datadir?  verify the package works in either home/config and in system

humdingerb:
@scottmc, I pushed a fix for the not-found-backgrounds. Maybe Begasus tries that first, before you doing another tag (if you were to tag...). I tested by building a package, and it works...

Should we have HexVexed load/save settings from ~/config/settings/ instead of the system settings? If so, how important is auto-importing from the old location (it's "just" a game... :) ).

scottmc:
settings right now are mostly to save your previously selected game mode, but in future might also store your best times.  where do most programs put their settings?  I guess just match that and it'd be fine.

humdingerb:
Normally, settings are saved in the user's hierarchy, i.e. ~/config/settings/
If there may be more files in the future, I'd say put it in a "HexVexed" subfolder there. I'll prepare a patch.

scottmc:
sounds good.

Begasus:
Almost there :) HexVexed picks up $datadir/hexvexed/background/* :)
But if I copy the backgrounds to ~/config-nonpackaged/.../* and rename them to *2.png Hexvexed doesn't' pick up the 2 folders, instead uses the one from nonpackaged (and starts with a blank/white background, as it doesn't remember the last location?)

humdingerb:
Backgrounds in non-packaged completely override the packaged backgrounds, they are not "merged". I consider this way of using custom backgrounds more as an easter egg. To be user-friendly, it would have to be set from the GUI, which would require quite some code changes. Maybe if a real coder is found as mentor, this would be a couple of GCI tasks... :)

Begasus:
Ok, the code is working as expected then, @scottmc could you tag again or should I use the srcGitRev checkout in the recipe?

scottmc:
Use srcgitrev to test, and if that works and everything else works as expected, let me know and I'll do fresh tag if one is needed.

Begasus:
> Use srcgitrev to test, and if that works and everything else works as expected, let me know and I'll do fresh tag if one is needed.

Already tested with srcGitRev, works as explained, so it's good to go for now, the other things could be ok for GCI as @humdingerb  replied :)

PR #3105, opened 2018-09-21
alembic, fix recipe for _x86 (secondaryArch)

Begasus:


extrowerk:
Nice!

Begasus:
cmake doesn't work well for secondaryArchSuffix (specialy for the libs)

Begasus:
For the subfolder in $libDir? If it's not created it will give an error ...

PR #3106, opened 2018-09-22
libsoundtouch, bump version

Begasus:


media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -0,0 +1,74 @@
...
+The library additionally supports estimating stable beats-per-minute rates for \
+audio tracks.
+
+- Tempo (time stretch): Changes the sound to play at faster or slower tempo \
+than originally without affecting the sound pitch.
+- Pitch (key) : Changes the sound pitch or key while keeping the original \
+tempo (speed).
+Playback Rate : Changes both tempo and pitch together as if a vinyl disc was \
+played at different RPM rate."
+HOMEPAGE="http://www.surina.net/soundtouch"
```
fbrosson:
```diff
-HOMEPAGE="http://www.surina.net/soundtouch"
+HOMEPAGE="https://www.surina.net/soundtouch/"
```

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -0,0 +1,74 @@
...
+The library additionally supports estimating stable beats-per-minute rates for \
+audio tracks.
+
+- Tempo (time stretch): Changes the sound to play at faster or slower tempo \
+than originally without affecting the sound pitch.
+- Pitch (key) : Changes the sound pitch or key while keeping the original \
+tempo (speed).
+Playback Rate : Changes both tempo and pitch together as if a vinyl disc was \
+played at different RPM rate."
+HOMEPAGE="https://www.surina.net/soundtouch"
```
fbrosson: Oops, you forgot the trailing slash.
```diff
-HOMEPAGE="https://www.surina.net/soundtouch"
+HOMEPAGE="https://www.surina.net/soundtouch/"
```
That's because `curl --head` (or `curl -I`) redirects to the same URI with a trailing slash added, so it is better to add it in the recipe.

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -0,0 +1,74 @@
...
+	make install
+
+	prepareInstalledDevelLibs libSoundTouch
+	fixPkgconfig
+
+	rm -Rf $prefix/doc
+
+	# devel package
+	packageEntries devel \
+		$developDir
```
fbrosson: Suggestion:
```diff
-	rm -Rf $prefix/doc
+	rm -Rf "$prefix"/doc
 
 	# devel package
 	packageEntries devel \
-		$developDir
+		"$developDir"
 }
+
+TEST()
+{
+	make check
+}
```

fbrosson: Yep, I know there are no tests, but there is a `check` target in the `Makefile`, so imho it is slightly better to have a `TEST()`, even if it _currently_ does nothing.

Begasus: Will add :)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -0,0 +1,74 @@
...
+CHECKSUM_SHA256="3fb740c7ac4b304342fbf11a91d14fa3e296269b9fec6266609c4620477d0ad6"
+SOURCE_DIR="soundtouch-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	libsoundtouch$secondaryArchSuffix = $portVersion compat >= 1
+	cmd:soundstretch$secondaryArchSuffix
+	lib:libSoundTouch$secondaryArchSuffix = 1.0.0 compat >= 1
```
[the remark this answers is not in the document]
Begasus: wasn't sure about this, as the soname didn't change between versions?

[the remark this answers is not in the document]
Begasus: PS, why the use of portVersionCompat (I even think the compat >= 1 shouldn't be needed)?
Or would it be required for other recipes depending on compat >=1?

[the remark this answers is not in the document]
fbrosson: Yep, I can confirm that the soname did not change. (It's `libSoundTouch.so.1` for both the old release, `2.0.0`, and the new release, `2.1.0`)
```
$ objdump -x libSoundTouch.so.1.0.0 | grep SONAME
  SONAME               libSoundTouch.so.1
```

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
alaviss: Is there any documentation in that "doc" folder? If so, I think we should package it.

fbrosson: Good question. IDK.

Begasus: Removed, obselte :)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -10,30 +10,33 @@ than originally without affecting the sound pitch.
...
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
-	libsoundtouch$secondaryArchSuffix = $portVersion compat >= 1
-	cmd:soundstretch$secondaryArchSuffix
-	lib:libSoundTouch$secondaryArchSuffix = 1.0.0 compat >= 1
+	libsoundtouch$secondaryArchSuffix = $portVersionCompat
+	cmd:soundstretch$secondaryArchSuffix = 2.1
```
fbrosson:
```diff
-	cmd:soundstretch$secondaryArchSuffix = 2.1
+	cmd:soundstretch$secondaryArchSuffix = $portVersion
```
(or maybe `$portVersionCompat` instead of `$portVersion`, or `$portVersion compat >= 2.1`)

Hmm, BTW, if we're sure that the program cannot be built with gcc2 then it might be a good idea to define a [`commandBinDir`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) and `commandSuffix`.

Begasus: Already changed locally for the portVersion
libsoundtouch failing for gcc2 with "cpu_detect_x86.cpp:77: Invalid token in expression"

fbrosson: > Already changed locally for the portVersion

Oops, we still have `2.1`, while `$portVersion` is `2.1.0`.

Begasus: Now we don't :)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -18,16 +18,23 @@ SOURCE_URI="https://gitlab.com/soundtouch/soundtouch/-/archive/$portVersion/soun
...
+	commandSuffix=
+	commandBinDir="$prefix"/bin
+fi
+
 libVersion="1.0.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 portVersionCompat="$portVersion compat >= 1"
 
-ARCHITECTURES="!x86_gcc2 x86 x86_64"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
fbrosson:
```diff
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
 
 commandSuffix=$secondaryArchSuffix
-commandBinDir="$binDir"
+commandBinDir=$binDir
 if [ "$targetArchitecture" = x86_gcc2 ]; then
 	commandSuffix=
-	commandBinDir="$prefix"/bin
+	commandBinDir=$prefix/bin
 fi
+
 libVersion="1.0.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 portVersionCompat="$portVersion compat >= 1"
- 
-ARCHITECTURES="x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="x86"

 PROVIDES="
```
* [Using the commandBinDir User Variable](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable)
* [HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering)

fbrosson: Remark: If I'm not mistaken, most recipes which define commandBinDir don't use quotes in their definition of `commandBinDir`.
This was true for **all** recipes until recently. I don't know why @korli decided to add quotes there. They are completely unneeded.
And even if "haikuporter --nochroot" is implemented and a user chooses a `TREE_PATH` containing white spaces then the quotes would **still** be unneeded because the shell does not need them when it is doing an assignement which does not include white spaces. The quotes are only required if there are white spaces in the assignement **itself**. If the white spaces are **only** in the variables being refered then the quotes are **not** required.

So, since most recipes currently follow the template we have in the wiki, I suggest that we stick to it.

Begasus: PS, @korli didn't add them, that was my fault :)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -10,30 +10,40 @@ than originally without affecting the sound pitch.
...
+libVersion="1.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+portVersionCompat="$portVersion compat >= 1"
+
 PROVIDES="
-	libsoundtouch$secondaryArchSuffix = $portVersion compat >= 1
-	cmd:soundstretch$secondaryArchSuffix
-	lib:libSoundTouch$secondaryArchSuffix = 1.0.0 compat >= 1
+	libsoundtouch$secondaryArchSuffix = $portVersionCompat
+	cmd:soundstretch$commandSuffix = $portVersion compat >= 2.1
```
fbrosson: I just checked that there are no recipes depending on `cmd:soundstretch` or `libsoundstretch`, so this means that using the same compat string for both the package and the `cmd:soundstretch` would have no impact:
```diff
-portVersionCompat="$portVersion compat >= 1"
+portVersionCompat="$portVersion compat >= 2.1"
 
 PROVIDES="
 	libsoundtouch$secondaryArchSuffix = $portVersionCompat
-	cmd:soundstretch$commandSuffix = $portVersion compat >= 2.1
+	cmd:soundstretch$commandSuffix = $portVersionCompat
```

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -52,24 +62,33 @@ BUILD_PREREQUIRES="
...
-	runConfigure ./configure
+	runConfigure --omit-dirs binDir ./configure --bindir="$commandBinDir"
 	make $jobArgs
 }
 
 INSTALL()
 {
 	make install
 
+	rm -rf "$libDir"/*.la
```
fbrosson:
```diff
- 	rm -rf "$libDir"/*.la
+	rm -f "$libDir"/*.la
```
(We don't need recursion. `-f` is enough.)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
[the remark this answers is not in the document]
Begasus: Shouldn't $portVersionCompat reflect the version of the cmd also then? (like 2.1 instead of 1)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
fbrosson: In 53e790e14e7ad2c6dcc4d97d26cd538fdf3c5035 @korli changed `commandBinDir=$binDir` to `commandBinDir="$binDir"` for:
* [app-arch/zstd/zstd-1.1.2.recipe](https://github.com/haikuports/haikuports/commit/53e790e14e7ad2c6dcc4d97d26cd538fdf3c5035#diff-16d36e0aa549977d29c4ad4a90ad8528R20)
* [app-editors/emacs/emacs-26.1.recipe](https://github.com/haikuports/haikuports/commit/53e790e14e7ad2c6dcc4d97d26cd538fdf3c5035#diff-1ce5c71d39fa3f430738cf5752777e98R27)

I don't remember if there were other similar changes. But I'm quite sure that most recipes are still using a style consitent with the wiki.

fbrosson: More of these in b369b0e338abbafc4dd08c38fc4bf048ab559be4:
* [media-libs/alembic/alembic-1.7.8.recipe](https://github.com/haikuports/haikuports/commit/b369b0e338abbafc4dd08c38fc4bf048ab559be4#diff-fdcaf416fea60f28e463587e604f4250R34)
* [media-libs/lcms/lcms-2.9.recipe](https://github.com/haikuports/haikuports/commit/b369b0e338abbafc4dd08c38fc4bf048ab559be4#diff-6690ae642865718f5ad4d786731c80b4R20)

I really don't understand why he made these edits. They were unneeded and, above all, not consistent with the wiki.

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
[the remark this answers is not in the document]
fbrosson: I don't really know. Well, I'd suggest setting  `portVersionCompat="$portVersin compat >= 2.1"`, like you.
That said, we could use `$portVersion` (instead of `$portVersionCompat`) and drop the definition of `portVersionCompat`.
Any opinions here?

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
fbrosson: Good news: we still have over 120 recipes using the style recommended by the wiki, and only 15 recipes using a non consistent style.

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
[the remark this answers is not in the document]
Begasus: I'm in favour of "portVersionCompat=$portVersion compat >= 2.1" then (as the actual version printed by the binary is 2.1)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -18,23 +18,23 @@ SOURCE_URI="https://gitlab.com/soundtouch/soundtouch/-/archive/$portVersion/soun
...
+	commandBinDir=$prefix/bin
 fi
 
 libVersion="1.0.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
-portVersionCompat="$portVersion compat >= 1"
-
-ARCHITECTURES="x86_gcc2 x86 x86_64"
-SECONDARY_ARCHITECTURES="x86"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
```
fbrosson: If you want to have "compat >= 2.1" then you could use:
```diff
-portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%.*}"
```
or hardcode it. I don't know which is best. Well, using
```
portVersionCompat="$portVersion compat >= 2.1"
```
might not be that bad. Indeed, when  `2.2.x` gets released it could be possible that it would be compatible with `2.1.x`.

Begasus: Fixed, I guess by the time they release 2.2.0 we could delete this one after an updated recipe :)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
```diff
@@ -52,24 +62,33 @@ BUILD_PREREQUIRES="
...
+
 	prepareInstalledDevelLibs libSoundTouch
 	fixPkgconfig
 
-	rm -Rf $prefix/doc
-
 	# devel package
 	packageEntries devel \
-		$developDir
+		"$developDir"
```
fbrosson:
```diff
 	packageEntries devel \
+		"$dataDir" \
 		"$developDir"
```
(We want `data/aclocal/soundtouch.m4` in the devel package.)

media-libs/libsoundtouch/libsoundtouch-2.1.0.recipe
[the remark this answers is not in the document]
fbrosson: IDK. Maybe we could simply drop the `portVersionCompat` and not use any compat for the package and for the `cmd:`.
Pros: No need to check for the compatibility of the runtime.

PR #3108, opened 2018-09-22
LibSSH: new recipe

extrowerk:


net-libs/libssh/libssh-0.8.3.recipe
```diff
@@ -0,0 +1,93 @@
...
+	2007 Daniel Gollub
+	2007-2018 Andreas Schneider
+	2012 Dmitriy Kuznetsov
+	2017 Sartura d.o.o.
+	2017 Jan-Lukas Wynen
+	2018 Anderson Toshiyuki Sasaki
+	2018 by Red Hat, Inc."
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://www.libssh.org/files/0.8/libssh-$portVersion.tar.xz"
```
fbrosson: Suggestion:
```diff
-SOURCE_URI="https://www.libssh.org/files/0.8/libssh-$portVersion.tar.xz"
+SOURCE_URI="https://www.libssh.org/files/${portVersion%.*}/libssh-$portVersion.tar.xz"
```
This is OK because all releases seem to use the same `x.y.z` format, so `${portVersion%.*}` will always be `x.y`.

PR #3119, opened 2018-09-23
[WIP] golang: add recipe for 1.4.3

return:
This PR adds a version of Go that makes it possible to bootstrap any Go toolchain which is written in Go (Go 1.5 and above).

This Golang port is based on the GSoC 2014 Go port which uses the Solaris port, but has been updated to work on newer versions of Go with some catches though (direct syscalls are unsupported, no cgo yet, etc)

One can try to bootstrap a higher Go version using this, but note that since this is still a WIP and some bugs may prevent bootstrapping to complete and I'm still documenting these bugs in several go packages.

I also tested the packages under Haiku x86_64 and ran ./all.bash and all of the results can be found in this [gist](https://gist.github.com/return/e76f5ce30c240a6123e916f39f4b930d).

### DIFF
--- a/dev-lang/golang/golang-1.4.3.recipe
+++ b/dev-lang/golang/golang-1.4.3.recipe
@@ -1,0 +1,73 @@
+SUMMARY="The Go Programming Language"
+DESCRIPTION="The Go Programming Language is an open source project to make \
+programmers more productive.
+
+Go is expressive, concise, clean, and efficient. Its concurrency \
+mechanisms make it easy to write programs that get the most out of multicore \
+and networked machines, while its novel type system enables flexible and \
+modular program construction. Go compiles quickly to machine code yet has the \
+convenience of garbage collection and the power of run-time reflection. It's a \
+fast, statically typed, compiled language that feels like a dynamically typed, \
+interpreted language."
+HOMEPAGE="https://golang.org/go"
+COPYRIGHT="2009-2018 The Go Authors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://dl.google.com/go/go$portVersion.src.tar.gz"
+CHECKSUM_SHA256="9947fc705b0b841b5938c48b22dc33e9647ec0752bae66e50278df4f23f64959"
+SOURCE_DIR="go"
+PATCHES="golang-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 !x86 ?x86_64"
+SECONDARY_ARCHITECTURES="!x86"
+
+PROVIDES="
+	golang$secondaryArchSuffix = $portVersion
+	cmd:go$secondaryArchSuffix = $portVersion
+	cmd:gofmt$secondaryArchSuffix = $portVersion
+	cmd:godoc$secondaryArchSuffix = $portVersion
+	"
+if [ -n "$secondaryArchSuffix" ]; then
+PROVIDES="$PROVIDES
+	cmd:go = $portVersion
+	cmd:gofmt = $portVersion
+	cmd:godoc = $portVersion
+	"
+fi
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
+	"
+
+BUILD()
+{
+	echo "$portVersion" > VERSION
+	export GOROOT_FINAL=/system/$relativeDevelopLibDir/go
+	cd src
+	DISABLE_ASLR=1 ./make.bash
+	cd ..
+	export GOPATH=/tmp
+	export GOROOT="$sourceDir"
+}
+
+INSTALL()
+{
+	mkdir -p $developLibDir/go/
+	cp -r src pkg doc $developLibDir/go/
+	mkdir -p $binDir
+	cp bin/* $binDir/
+}
+
+TEST()
+{
+	cd src
+	DISABLE_ASLR=1 ./all.bash
+}
--- a/dev-lang/golang/patches/golang-1.4.3.patchset
+++ b/dev-lang/golang/patches/golang-1.4.3.patchset
@@ -1,0 +1,7777 @@
+From 709601cc4136cb49e4a6f3f1b2b90058b698d396 Mon Sep 17 00:00:00 2001
+From: Calvin Hill <calvin@hakobaito.co.uk>
+Date: Sun, 23 Sep 2018 00:55:04 +0100
+Subject: cmd/ld: Port cmd and ld to build on Haiku x86_64.
+
+
+diff --git a/include/link.h b/include/link.h
+index 05e117c..50c8120 100644
+--- a/include/link.h
++++ b/include/link.h
+@@ -495,6 +495,7 @@ enum {
+ 	Hdragonfly,
+ 	Helf,
+ 	Hfreebsd,
++	Hhaiku,
+ 	Hlinux,
+ 	Hnacl,
+ 	Hnetbsd,
+diff --git a/src/cmd/5l/asm.c b/src/cmd/5l/asm.c
+index 9c1c04e..7af74ea 100644
+--- a/src/cmd/5l/asm.c
++++ b/src/cmd/5l/asm.c
+@@ -38,6 +38,7 @@
+ 
+ char linuxdynld[] = "/lib/ld-linux.so.3"; // 2 for OABI, 3 for EABI
+ char freebsddynld[] = "/usr/libexec/ld-elf.so.1";
++char haikudynld[] = "XXX";
+ char openbsddynld[] = "XXX";
+ char netbsddynld[] = "/libexec/ld.elf_so";
+ char dragonflydynld[] = "XXX";
+diff --git a/src/cmd/6l/asm.c b/src/cmd/6l/asm.c
+index 18b5aa3..3936b83 100644
+--- a/src/cmd/6l/asm.c
++++ b/src/cmd/6l/asm.c
+@@ -41,6 +41,7 @@
+ 
+ char linuxdynld[] = "/lib64/ld-linux-x86-64.so.2";
+ char freebsddynld[] = "/libexec/ld-elf.so.1";
++char haikudynld[] = "/system/runtime_loader";
+ char openbsddynld[] = "/usr/libexec/ld.so";
+ char netbsddynld[] = "/libexec/ld.elf_so";
+ char dragonflydynld[] = "/usr/libexec/ld-elf.so.2";
+@@ -301,7 +302,15 @@ elfreloc1(Reloc *r, vlong sectoff)
+ 
+ 	case R_PCREL:
+ 		if(r->siz == 4) {
+-			VPUT(R_X86_64_PC32 | (uint64)elfsym<<32);
++			// HACK to fix relocation issue with Haiku //
++			if(HEADTYPE == Hhaiku) {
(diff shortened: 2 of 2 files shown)


waddlesplash:
Have you seen https://github.com/korli/go -- and cc @korli, does that repository actually work?

Direct syscalls being unsupported is a known issue; we haven't locked our syscall numbers yet, similarly to what macOS does here.

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+ARCHITECTURES="!x86_gcc2 !x86 ?x86_64"
+SECONDARY_ARCHITECTURES="!x86"
+
+PROVIDES="
+	golang$secondaryArchSuffix = $portVersion
+	cmd:go$secondaryArchSuffix = $portVersion
+	cmd:gofmt$secondaryArchSuffix = $portVersion
+	cmd:godoc$secondaryArchSuffix = $portVersion
+	"
+if [ "$secondaryArchSuffix" != "" ]; then
```
fbrosson: You might wish to use this variant:
```diff
-if [ "$secondaryArchSuffix" != "" ]; then
+if [ -n "$secondaryArchSuffix" ]; then
```

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+	cmd:go = $portVersion
+	cmd:gofmt = $portVersion
+	cmd:godoc = $portVersion
+	"
+fi
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+BUILD_REQUIRES="
```
fbrosson:
```diff
 REQUIRES="
 	haiku$secondaryArchSuffix
 	"
+
 BUILD_REQUIRES="
```
(We like to have an empty line before `BUILD_REQUIRES`, see [HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering).)

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+
+BUILD()
+{
+	echo "$portVersion" > VERSION
+	export GOROOT_FINAL=/system/develop/lib$secondaryArchSubDir/go
+	cd src
+	DISABLE_ASLR=1 ./make.bash
+	cd ..
+	export GOPATH=/tmp
+	export GOROOT=`pwd`
```
fbrosson: Suggestion:
```diff
-	export GOROOT=`pwd`
+	export GOROOT="$sourceDir"
```

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,80 @@
...
+}
+
+INSTALL()
+{
+	mkdir -p $developLibDir/go/
+	cp -r src pkg doc $developLibDir/go/
+	mkdir -p $binDir
+	cp bin/* $binDir/
+	if [ "$secondaryArchSuffix" != "" ]; then
```
fbrosson: You might wish to use this variant:
```diff
- 	if [ "$secondaryArchSuffix" != "" ]; then
+	if [ -n "$secondaryArchSuffix" ]; then
```

fbrosson: Hmm, haikuporter takes care of creating the symlinks (`bin/foobar-x86 -> x86/foobar`), so you should probably drop this bloc. If you don't want to have `foobar-x86` in `bin/` then you should probably define the [commandBinDir](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) User Variable.
```diff
-	if [ "$secondaryArchSuffix" != "" ]; then
-		cd $binDir/..
-		for a in `ls $binDir`
-		do
-			ln -s .$secondaryArchSubDir/$a $a
-		done
- 	fi
```

return:
@fbrosson Patches applied into this recipe.

@waddlesplash I've seen this port before at first but it never worked when cross-compiling with Linux. I've tried [bootstrapping](https://gist.github.com/return/6e1d24dc67da616b5f828a3b84b12704) it today with Go 1.4.3, and got as far as compiling a few packages but it gets stuck after compiling net/url.

I'm investigating this hang in the bootstrapping process, but it seems that there could be an issue with some of the toolchain tools (asm, compile and link). Maybe someone could have a closer look?

Please add help wanted.

dev-lang/golang/golang-1.4.3.recipe
fbrosson: Oops, you forgot to insert an empty line between the last line of the `REQUIRES` bloc and the `BUILD_REQUIRES`.

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,74 @@
...
+
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
+	"
+
+BUILD()
+{
+	echo "$portVersion" > VERSION
+	export GOROOT_FINAL=/system/develop/lib$secondaryArchSubDir/go
```
fbrosson: You can use this variant, which is equivalent to what you have right now:
```diff
- 	export GOROOT_FINAL=/system/develop/lib$secondaryArchSubDir/go
+ 	export GOROOT_FINAL=/system/$relativeDevelopLibDir/go
```

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,74 @@
...
+	mkdir -p $binDir
+	cp bin/* $binDir/
+}
+
+TEST()
+{
+	cd src
+	DISABLE_ASLR=1 ./all.bash
+}
+
```
fbrosson: No need for this empty line at the end of the recipe :)

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+	cmd:godoc$secondaryArchSuffix = $portVersion
+	"
+if [ -n "$secondaryArchSuffix" ]; then
+	PROVIDES="$PROVIDES
+	cmd:go = $portVersion
+	cmd:gofmt = $portVersion
+	cmd:godoc = $portVersion
+	"
+fi
+
```
Begasus: No empty line here, ps, if this doesn't build for x86_gcc2 couldn't you use commandSuffix?

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+	cp -r src pkg doc $developLibDir/go/
+	mkdir -p $binDir
+	cp bin/* $binDir/
+}
+
+TEST()
+{
+	cd src
+	DISABLE_ASLR=1 ./all.bash
+}
```
fbrosson: Oops, now the last line is missing a line-feed.
What text editor are you using? Pe? Notepad?
These are not vim-friendly... they seem to remove the line-feed at the last line, except if that last line is empty.

Next time I'll think twice before writing a comment like [this one](https://github.com/haikuports/haikuports/pull/3119#discussion_r219911390) :smile:

An easy fix for this is to open the file with `vim`, save it and quit. (`vi golang-1.4.3.recipe` + `ZZ` or `:wq`)

dev-lang/golang/golang-1.4.3.recipe
fbrosson: I think we can bend the rules a little bit regarding the empty line. That's because we have a `if .. then .. fi` just before the `REQUIRES`.
So, imho, the empty line might be ok here. @Begasus, what do you think?

The only change I'd recommend is:
```diff
 if [ -n "$secondaryArchSuffix" ]; then
-	PROVIDES="$PROVIDES
+PROVIDES="$PROVIDES
 	cmd:go = $portVersion
 	cmd:gofmt = $portVersion
 	cmd:godoc = $portVersion
 	"
 fi
```

Begasus: Just my 2 cents, is using $secondaryArchSuffix needed? If so then it's ok by me

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+ARCHITECTURES="!x86_gcc2 !x86 ?x86_64"
+SECONDARY_ARCHITECTURES="!x86"
+
+PROVIDES="
+	golang$secondaryArchSuffix = $portVersion
+	cmd:go$secondaryArchSuffix = $portVersion
+	cmd:gofmt$secondaryArchSuffix = $portVersion
+	cmd:godoc$secondaryArchSuffix = $portVersion
+	"
+if [ -n "$secondaryArchSuffix" ]; then
```
fbrosson: Sorry, I just realized that what we probably need is:
```diff
-ARCHITECTURES="!x86_gcc2 !x86 ?x86_64"
+ARCHITECTURES="!x86_gcc2 ?x86 ?x86_64"
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
-SECONDARY_ARCHITECTURES="!x86"
+SECONDARY_ARCHITECTURES="?x86"
+commandBinDir=$prefix/bin
+fi
 
 PROVIDES="
 	golang$secondaryArchSuffix = $portVersion
-	cmd:go$secondaryArchSuffix = $portVersion
-	cmd:gofmt$secondaryArchSuffix = $portVersion
-	cmd:godoc$secondaryArchSuffix = $portVersion
-	"
-if [ -n "$secondaryArchSuffix" ]; then
-PROVIDES="$PROVIDES
 	cmd:go = $portVersion
 	cmd:gofmt = $portVersion
 	cmd:godoc = $portVersion
 	"
-fi
```

dev-lang/golang/golang-1.4.3.recipe
```diff
@@ -0,0 +1,73 @@
...
+	export GOPATH=/tmp
+	export GOROOT="$sourceDir"
+}
+
+INSTALL()
+{
+	mkdir -p $developLibDir/go/
+	cp -r src pkg doc $developLibDir/go/
+	mkdir -p $binDir
+	cp bin/* $binDir/
```
fbrosson:
```diff
 INSTALL()
 {
-	mkdir -p $developLibDir/go/
+	mkdir -p "$developLibDir"/go/
-	cp -r src pkg doc $developLibDir/go/
+	cp -r src pkg doc "$developLibDir"/go/
-	mkdir -p $binDir
+	mkdir -p "$commandBinDinDir"
-	cp bin/* $binDir/
+	cp bin/* "$commandBinDinDir"
 }
```
(You can discard the cosmetic changes about the quotes if you wish, but I recommend using them in new recipes.)

dev-lang/golang/golang-1.4.3.recipe
fbrosson: Oops, this is not a new recipe, so maybe we should wait for some input from other contributors before we change this.

dev-lang/golang/golang-1.4.3.recipe
fbrosson: Same comment as above.
I guess I should check what we have for the existing recipe (1.3) of golong on x86_64 and golang_x86 on x86_gcc2.

dev-lang/golang/golang-1.4.3.recipe
Begasus: That was what I was referring too :)

kallisti5:
I'm pushing these through.  I'm looking at adding Haiku support to travis-ci since rust uses them.. and their worker code is all golang.  We have zero golang ports at the moment, and later golangs do a bootstrap (go building go)... so this is badly needed.

I'll review any final style issues and cleanup.

PR #3125, opened 2018-09-24
mini-xml, bump version

Begasus:
Marked as "?" because the creation of the documentation still errors out the build, see: https://github.com/haikuports/haikuports/commit/54f1fa8771392f1f23d80b8998fecdd1847b8feb#commitcomment-26342020


Begasus:
> The comment speaks of crashes when testing. Is it what you mean with "the creation of the documentation still errors out"?

Yes, the buildlog for v2.11 is the same as for v2.9 mentioned above

korli:
> Yes, the buildlog for v2.11 is the same as for v2.9 mentioned above

I see your patchset removes "testmxml mxml.xml @MXML_EPUB@" because it breaks the build. so the build isn't broken with the patchset? Does "make check" actually do something?

Begasus:
Build is ok, the make check doesn't do something (so obsolete)

korli:
then TEST() should call "make testmxml"

Begasus:
With textxml it fails (as expected I think) ...

PR #3145, opened 2018-09-25
Slack++ initial recipe to experiment with. runs but is untested.

cb88:


Begasus:
You've got a Trailing whitespace somewhere in there (didn't check) :)

cb88:
Yes, I saw... it isn't at the end of the recipe I'll have to open it up in Pe etc.. when I get a chance to fix it I'll update the PR.

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+DESCRIPTION="slack++ is an ncurses based client the Searchable Log of All Conversation and Knowledge"
+HOMEPAGE="https://github.com/ga2arch/slack"
+COPYRIGHT="2018"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="961dec741764dfb81999703b84989d93ec7a88ee"
+SOURCE_URI="https://github.com/ga2arch/slack/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="9c84b5d0ee7b528edfdde219263dd7c307c6e8cd9aadc491e7148f7457ea30a5"
+SOURCE_DIR="slack-$srcGitRev"
+#PATCHES=""
```
Begasus: No need for empty PATCHES=""

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+
+ARCHITECTURES="x86 !x86_gcc2"
```
Begasus: If it builds with gcc7 you probably could add x86_64 in there too so someone could check it there for you

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+SECONDARY_ARCHITECTURES="x86 !x86_gcc2"
```
Begasus: This would make sense "!x86_gcc2 x86" if x86_64 was added in the previous line

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+
+PROVIDES="cmd:slack++$secondaryArchSuffix = $portVersion"
```
Begasus: I guess you meant:

> PROVIDES="slack++$secondaryArchSuffix = $portVersion
cmd:slack++$secondaryArchSuffix = $portVersion
"

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+
```
Begasus: No empty line between PROVIDES and REQUIRES

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+REQUIRES="
+ 	haiku$secondaryArchSuffix
+	lib:libjsoncpp$secondaryArchSuffix
+	lib:libcurl$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	lib:libcrypto$secondaryArchSuffix
+	"
```
Begasus: Insert an empty line here

cb88: This seems to be more of a style thing perhaps the Gentle intro should be updated to this? the https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines allude to this but yeah it didn't make it clear that whitespace is wanted style consideration.

Begasus: Here is an example :) 
https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#a-simple-example-using-the-recommended-style-for-description

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libjsoncpp$secondaryArchSuffix
+	devel:libcurl$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	devel:libcrypto$secondaryArchSuffix
+	"
+
```
Begasus: No empty line here :)

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+"
```
Begasus:
```diff
-"
+	"

File: net-im/slack++/slack++-0.recipe [hunk 9]
```diff
@@ -0,0 +1,56 @@
...
+
+BUILD()
+{
+	mkdir -p build
+        cd build
+        cmake -G "Unix Makefiles" ../ $cmakeDirArgs
```
Begasus:
```diff
-        cd build
-        cmake -G "Unix Makefiles" ../ $cmakeDirArgs
+		cd build
+		cmake -G "Unix Makefiles" ../ $cmakeDirArgs

Comment (Begasus) [reply net-im/slack++/slack++-0.recipe]: Best practice to use tabs instead of spaces

File: net-im/slack++/slack++-0.recipe (hunk 8 above)
Comment (Begasus) [reply net-im/slack++/slack++-0.recipe]: This needs to be done ....

File: net-im/slack++/slack++-0.recipe [hunk 10]
```diff
@@ -0,0 +1,56 @@
...
+REVISION="1"
+srcGitRev="961dec741764dfb81999703b84989d93ec7a88ee"
+SOURCE_URI="https://github.com/ga2arch/slack/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="9c84b5d0ee7b528edfdde219263dd7c307c6e8cd9aadc491e7148f7457ea30a5"
+SOURCE_DIR="slack-$srcGitRev"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+PROVIDES="slack++$secondaryArchSuffix = $portVersion
+		cmd:slack++$secondaryArchSuffix = $portVersion"
```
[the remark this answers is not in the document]
Begasus: Adjust to the guidelines ...

net-im/slack++/slack++-0.recipe
Begasus: Remove empty line ...

net-im/slack++/slack++-0.recipe
cb88: Sorry brain did not parse what you were showing me.

Begasus: The double quote needs a tab in front of it :)

cb88: Yeah I also fixed the provides which was slightly janky.

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+COPYRIGHT="2018"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="961dec741764dfb81999703b84989d93ec7a88ee"
+SOURCE_URI="https://github.com/ga2arch/slack/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="9c84b5d0ee7b528edfdde219263dd7c307c6e8cd9aadc491e7148f7457ea30a5"
+SOURCE_DIR="slack-$srcGitRev"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
```
Begasus: Insert an emtpy line here, and also
```diff
-SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
```
no need to call if it doesn't build for x86_gcc2 :)

cb88: Yeah but isn't there technically supposed to eventually be x86_gcc2 on x86_64?

Begasus: If it doesn't build on primaryArch it won't either on secondaryArch :) "!x86_gcc2"

fbrosson: The rule is a bit special: We do mention broken arches in `ARCHITECTURES`, but we prefer not to mention them in `SECONDARY_ARCHITECTURES`.

cb88: Ok this makes sense now I think.

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
+SUMMARY="Enterprise chat and services on the command line - slack++"
+DESCRIPTION="slack++ is an ncurses based client the Searchable Log of All Conversation and Knowledge"
+HOMEPAGE="https://github.com/ga2arch/slack"
+COPYRIGHT="2018"
```
alaviss: `COPYRIGHT` has to be in `<year> <copyright holder>` format.

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+LICENSE="GNU GPL v2"
```
alaviss: I'm actually a bit skeptical about this. None of the source files state what they are licensed as, there is no COPYRIGHT/LICENSE file, only an `PKGBUILD` with the `licenses` field set to `GPL`

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+	cd build
+	cmake -G "Unix Makefiles" ../ $cmakeDirArgs
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	mkdir -p $binDir
+	install -t $binDir -s slack++
```
alaviss:
```diff
-	install -t $binDir -s slack++
+	install -t $binDir slack++
```

If the binary contains debugging information, please use `defineDebugInfoPackage` to strip it from the executable.

fbrosson:
Hmm, I just noticed that your PR is based on commits that you pushed to your master branch.
While this is not a problem at all, many users prefer to use [feature-branches](https://github.com/haikuports/haikuports/wiki/DevelopmentModel#what-are-feature-branches).

This allows you to keep your master branch perfectly in sync with the project's master, and even easily refresh your master branch at any time, without having to touch your feature-branches.
That said, you are perfectly free to submit anything from your master branch. Using feature-branches is just a recommendation, and not using them does not have any impact.

net-im/slack++/slack++-0.recipe
cb88: I fail to see why, a copyright or license file is a convention not a mandate.

The PKGBUILD is doing the same thing.

net-im/slack++/slack++-0.recipe
cb88: Well the binary obviously does not contain such as I stripped it :P 

The guidelines say it is optional anyway. If it is needed at a later time it can be added.

cb88:
> 
> 
> Hmm, I just noticed that your PR is based on commits that you pushed to your master branch.
> While this is not a problem at all, many users prefer to use [feature-branches](https://github.com/haikuports/haikuports/wiki/DevelopmentModel#what-are-feature-branches).
> 
> This allows you to keep your master branch perfectly in sync with the project's master, and even easily refresh your master branch at any time, without having to touch your feature-branches.
> That said, you are perfectly free to submit anything from your master branch. Using feature-branches is just a recommendation, and not using them does not have any impact.

Yeah I haven't used git much but definitely a valid point there. I will have to give that a try.

net-im/slack++/slack++-0.recipe
cb88: Asked for clarification here https://github.com/ga2arch/slack/issues/19

cb88: Got feedback and fixed this the source is now GPL v3 explicitly.

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
+SUMMARY="Slack Enterprise chat and services on the command line"
+DESCRIPTION="slack++ is an ncurses based client for the Searchable Log of All Conversation and Knowledge"
```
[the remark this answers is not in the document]
cb88: The haikuports --lint didn't catch this, that's not very helpful?

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,56 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+srcGitRev="f085f2a269d012c876e128db74d350754ec0ca30"
+SOURCE_URI="https://github.com/ga2arch/slack/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="2b42d26328e666dbbd431e8f389234eb19fd56f4ad65997d3d442b0cb96a4b6a"
+SOURCE_DIR="slack-$srcGitRev"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+PROVIDES="
```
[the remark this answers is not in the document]
cb88: I'm not sure how this change to secondary architectures is useful? What I am missing there?

net-im/slack++/slack++-0.recipe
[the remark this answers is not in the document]
fbrosson: Yep, that's because haikuporter is not as picky as it could/should be :)

net-im/slack++/slack++-0.recipe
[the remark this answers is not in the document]
fbrosson: I do agree that it won't make any diference to most users.
But to those users who have built x86_64 hybrid with support for x86 secondary arch, having `slack++_x86` as a buildable recipe is of no use. They already have `slack++` for x86_64 primary arch.
`slack++_x86` is only needed on x86_gcc2 hybrid.

BTW, you can improve the recipe a little bit by [using the commandBinDir User Variable](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable) and making this change in `PROVIDES`:
```diff
- 	cmd:slack++$secondaryArchSuffix = $portVersion
+	cmd:slack++ = $portVersion
```

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,58 @@
...
+SOURCE_URI="https://github.com/ga2arch/slack/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="2b42d26328e666dbbd431e8f389234eb19fd56f4ad65997d3d442b0cb96a4b6a"
+SOURCE_DIR="slack-$srcGitRev"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	slack++$secondaryArchSuffix = $portVersion
+	cmd:slack++$secondaryArchSuffix = $portVersion
```
fbrosson: In case you choose to use the `commandBinDir` variable, the following suggestion overrides my previous one.
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+commandBinDir=$prefix/bin
+fi
+
 PROVIDES="
 	slack++$secondaryArchSuffix = $portVersion
-	cmd:slack++$secondaryArchSuffix = $portVersion
+	cmd:slack++ = $portVersion
```

net-im/slack++/slack++-0.recipe
```diff
@@ -0,0 +1,58 @@
...
+	cd build
+	cmake -G "Unix Makefiles" ../ $cmakeDirArgs
+	make $jobArgs
+}
+
+INSTALL()
+{
+	cd build
+	mkdir -p $binDir
+	install -t $binDir -s slack++
```
fbrosson:
```diff
-	mkdir -p $binDir
+	mkdir -p "$commandBinDir"
-	install -t $binDir -s slack++
+	install -t "$commandBinDir" -s slack++
```

net-im/slack++/slack++-0.recipe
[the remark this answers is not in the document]
KapiX: I don't buy it. We already had similar workarounds, to mark x86_gcc2 as not-broken when build for secondary architectures. Haikuporter was changed to remove this. Why introduce something like this again? What if x86_64 gets support for x86? Are you going to remove it from all recipes?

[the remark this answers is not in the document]
fbrosson: Of course not!
We only need to do this for recipes which do not provide any `lib:lib*` in `PROVIDES`.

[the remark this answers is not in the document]
fbrosson: @cb88, have you had a look in `/boot/home/haikuports/repository/`?
This directory contains a bunch of files. If a recipe, e.g. `slack++`, is available for e.g. x86_64 primary arch and x86 secindary arch, and if you happen to be running x86_64 hybrid and have `x86` in your `haikuports.conf` then you will have one file for `slack++` and another for `slack++_x86`, The first one is welcome. The second one is totally unneeded.
Why would we give up making a change that only adds 2 lines but saves us an unneeded file in the `repository` directory?

I'm not making this change on existing recipes if I don't have another real reason to edit them.
But on **new** recipes which do not provide any library, not adding the condition for the secondary arch is IMO a bad practice.

net-im/slack++/slack++-0.recipe
fbrosson: Here's a variant for the first part of the patch in case you don't like the first part of the previous suggestion.
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+fi
+
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandBinDir=$prefix/bin
+fi
```

net-im/slack++/slack++-0.recipe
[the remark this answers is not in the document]
fbrosson: Shall we vote?
Comments welcome.

net-im/slack++/slack++-0.recipe
fbrosson:
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+fi
+
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandBinDir=$prefix/bin
+fi
 
 PROVIDES="
 	slack++$secondaryArchSuffix = $portVersion
-	cmd:slack++$secondaryArchSuffix = $portVersion
+	cmd:slack++ = $portVersion
```

net-im/slack++/slack++-0.recipe
fbrosson:
```diff
-	mkdir -p $binDir
+	mkdir -p "$commandBinDir"
-	install -t $binDir -s slack++
+	install -t "$commandBinDir" -s slack++
```

net-im/slack++/slack++-0.recipe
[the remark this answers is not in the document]
cb88: A recipe should just encode the minimum information that it needs to here.

Just consider how ugly this could get if more Haiku architectures are added.... and there are binary translators on those arches that allow running other arches packages etc... on ARM for instance people may want to use qemu-user-static-i386 to run old BeOS apps and they could just install the packages as anohter secondary arch.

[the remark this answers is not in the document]
fbrosson: > A recipe should just encode the minimum information that it needs to here.

The changes I'm requesting will:
1) allow x86_gcc2h users to type `slack++` instead of `slack++-x86`.
2) contribute to keep the `repository` directory free of unneeded files.

If you want point 1 then you need the test, otherwise the recipe will not be 100% valid because, on x86_84 hybrid, `cmd:slack++` will be provided by both `slack++` and `slack++_x86`. To avoid this conflict there are two solutions: Adding a conditional `CONFLICTS`, or adding the condition for the secondary arch.

> Just consider how ugly this could get if more Haiku architectures are added.... 

It is very unlikely that a new arch would be supported as a secondary arch on any of the primary arches were `slack++` builds, i.e. x86 and x86_64, so the test won't need to be changed.

Now, once the ARM port is ready, if `slack++` builds on it, then there will be no need to support x86 as a secondary arch for ARM, hence no need to change the test, which will still be:
```bash
 SECONDARY_ARCHITECTURES="x86"
fi
```
Otherwise, if the ARM arch gets support for using Qemu to run x86 code from packages built for x86 secondary arch then all we would need to do would be to change the test to:
```bash
if [ "$targetArchitecture" = x86_gcc2 ] || [ "$targetArchitecture" = arm ]; then
SECONDARY_ARCHITECTURES="x86"
fi
```
Of course, we could use this instead:
```bash
case "$targetArchitecture" in
x86_gcc2|arm) SECONDARY_ARCHITECTURES="x86";;
esac
```

My conclusion is that x86_64 hybrid **_will_** happen, and it will be much sooner than x86 secondary arch for ARM, if this ever happens.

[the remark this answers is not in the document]
fbrosson: Oops, if we ever need to add x86 secondary arch for ARM+Qemu then this would probably be better:
```diff
-ARCHITECTURES="!x86_gcc2 x86 x86_64"
+ARCHITECTURES="!x86_gcc2 x86 x86_64 !arm"
-SECONDARY_ARCHITECTURES="x86"
+case "$targetArchitecture" in
+x86_gcc2|arm) SECONDARY_ARCHITECTURES="x86";;
+esac
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
 
 PROVIDES="
 	slack++$secondaryArchSuffix = $portVersion
-	cmd:slack++$secondaryArchSuffix = $portVersion
+	cmd:slack++$commandSuffix = $portVersion
```
But all this would be totally unneeded if `slack++` builds on ARM.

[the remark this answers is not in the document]
KapiX: Still, you're talking about adding this for every app recipe. Let's say there is 500 of them. 2 lines x 500 = 1000. Or you can write 10 lines of Python and do that in haikuporter. Bonus: if the condition ever changes you have one place to change, not 500.
You can't have bugs in lines you didn't write.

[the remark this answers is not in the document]
fbrosson: First; there are **only 119 recipes** which meet all the conditions for them to have x86 secondary arch hidden on all arches other than x86_gcc2.
Second, I don't intend to change them in one commit. I will change them when there is a valid reason for changing a recipe.
I'll only make a batch edit when official images of x86_64 hybrid become available, and I'll do that only with those recipes that will not have been "fixed" in the meanwhile.

Moreover, handling these kind of special cases is not a job for haikuporter.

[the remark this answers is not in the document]
fbrosson: >  You can't have bugs in lines you didn't write.

I'm sorry, I don't understand what you are suggesting.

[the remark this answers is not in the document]
cb88: He is suggesting that HaikuPorter *is* the correct place to do this ... These are all information that can be gathered from the provides and arch definitions yes??? Then there is no reason *at all* to statically define it....even if doing it in HaikuPorter only covered part of them it would still be the right thing to do....in any case can we just move on from this and move this conversation to a ticket instead of my PR?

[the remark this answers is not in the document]
cb88: I have zero interest in impleting this hack so I'm closing it we can revisit it once something has generally been decided on this.

PR #3165, opened 2018-09-30
iASL: bump to 20181003.

fbrosson:


fbrosson:
The tests for the previous release, `20180810`, are OK and do not crash.
The tests for this release, `20180927`, do crash. 2 of the 5 crashes segmentation violations.

This PR is therefore a reminder that we should **not** bump iASL before someone investigates these crashes.

**Anyone finding the fix is welcome to open another PR.**
Of course, if you prefer, I can add the fix to this PR and a `Co-authored-by:` tag in the commit.
Hmm, I can even put your authorship, alone, in the commit.

PR #3174, opened 2018-10-01
afl, bump version, dirty hack to get it to build

Begasus:
This has been in the dark for a while, not tested on x86_64 (yet)


korli:
Really dirty!! ;)

Begasus:
Well, it could be closed if to dirty ;)

Begasus:
I'll close this, and leave the branch open, in case someone would like to look into it :)

korli:
Well it's a better to have something that builds, is up to date, and with a patch in the tree. 
But does it really work without shared memory?

Begasus:
It's been a while since I tested it, but as I said the branch is still around so if someone wants to have a go at it they can :)

PR #3180, opened 2018-10-02
Change the style of ADDITIONAL_FILES in 4 recipes.

fbrosson:
Also hide x86 secondary arch on x86_64 because the recipe does not provide any library.


fbrosson:
Recipes | Count
--- | ---:|
which define the `ADDITIONAL_FILES` variable: | 248
with only one additional-file: | 217
with two or more additional-files: | 31

Among these 31 recipes, this is what we had before this PR:

Recipes with | Count
--- | ---:|
two or more additional-files: | 31
and a style spreading on N+2 lines: | 27
and a style spreading on N or N+1 lines: | 4

PR #3183, opened 2018-10-04
Luckybackup, new recipe

mazbrili:
icon is not good.. i'm not artist


app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,96 @@
...
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://sourceforge.net/projects/luckybackup/files/0.4.9/source/luckybackup-0.4.9.tar.gz"
+CHECKSUM_SHA256="c161e98a11de5ad0a32a8e060e3fd45695a93ea3adf0581764a6b5d32e8da23b"
+SOURCE_DIR="luckybackup"
+SOURCE_FILENAME="luckybackup-$portVersion.tar.gz"
+SOURCE_DIR="luckybackup-$portVersion"
+ADDITIONAL_FILES="luckybackup.rdef.in"
+
+
```
Begasus: Remove empty line, 1 empty line before ARCHITECTURES is enough :)

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,96 @@
...
+ARCHITECTURES="x86_gcc2 x86_64"
+
+
```
Begasus: Same here, only 1 empty line

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,96 @@
...
+PROVIDES="
+	luckybackup$secondaryArchSuffix = $portVersion
+	app:Luckybackup = $portVersion
```
Begasus: Suggestion (we have a thing with Capitals) :)
```diff
-	app:Luckybackup = $portVersion
+	app:LuckyBackup = $portVersion
```
That would also mean renaiming the cp'd binary the resources and deskbarlink lines

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,96 @@
...
+
+BUILD()
+{
+	qmake
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/Luckybackup
```
Begasus: Suggeestion, you could combine the creation of the folders with one line:
```diff
-	mkdir -p $appsDir/Luckybackup
+	mkdir -p $appsDir/Luckybackup/{manual,translations} $manDir
```

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://sourceforge.net/projects/luckybackup/files/0.4.9/source/luckybackup-0.4.9.tar.gz"
+CHECKSUM_SHA256="c161e98a11de5ad0a32a8e060e3fd45695a93ea3adf0581764a6b5d32e8da23b"
+SOURCE_DIR="luckybackup"
+SOURCE_FILENAME="luckybackup-$portVersion.tar.gz"
+SOURCE_DIR="luckybackup-$portVersion"
+ADDITIONAL_FILES="luckybackup.rdef.in"
+
+ARCHITECTURES="x86_gcc2 x86_64"
```
Begasus: x86_gcc2 won't work with the libQT* deppendencies, you should add x86 and SECONDARY_ARCHITECTURES in here

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+PROVIDES="
+	luckybackup$secondaryArchSuffix = $portVersion
+	app:LuckyBackup = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
```
Begasus: In strict mode haikuporter gives a policy error for this one
```diff
+	lib:libGL$secondaryArchSuffix
```

mazbrili: @Begasus ..hm.. what should i do? maybe try to disableit and try rerun receipt?

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: ouch.. sorry .. i'l revise it

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+	lib:libQt5Core$secondaryArchSuffix
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Network$secondaryArchSuffix
+	lib:libQt5WebKit$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:librsync
+	lib:libssl
```
Begasus: Add $secondaryArchSuffix
```diff
-	lib:librsync
-	lib:libssl
+	lib:librsync$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
```

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+	devel:libGL$secondaryArchSuffix
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Network$secondaryArchSuffix
+	devel:libQt5WebKit$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:librsync
+	devel:libssl
```
Begasus: Add $secondaryArchSuffix
```diff
-	devel:librsync$secondaryArchSuffix
-	devel:libssl$secondaryArchSuffix
+	devel:librsync$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
```

app-backup/luckybackup/luckybackup-0.4.9.recipe
Begasus: The package is requiring it, haven't checked if it actually uses GL, try disabling the devel package and rerun the recipe

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,93 @@
+SUMMARY="...a powerful, fast and reliable backup & sync tool"
+DESCRIPTION="an application that backs-up and/or synchronizes any directories with the power of rsync.\
+It is simple to use, fast (transfers over only changes made and not all data),\
+safe (keeps your data safe by checking all declared directories before proceeding in any data manipulation ),\
+reliable and fully customizable."
```
Begasus: Wrap the DESCRIPTION into 80 characters per line

mazbrili: ok done

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+COPYRIGHT="2008-2017	Loukas Avgeriou	"
+LICENSE="GNU GPL v3"
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://sourceforge.net/projects/luckybackup/files/0.4.9/source/luckybackup-0.4.9.tar.gz"
+CHECKSUM_SHA256="c161e98a11de5ad0a32a8e060e3fd45695a93ea3adf0581764a6b5d32e8da23b"
+SOURCE_DIR="luckybackup"
```
Begasus: Only 1 SOURCE_DIR should be defined :)
```diff
-SOURCE_DIR="luckybackup"
```

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
+SUMMARY="...a powerful, fast and reliable backup & sync tool"
+DESCRIPTION="an application that backs-up and/or synchronizes any directories with the power of rsync.\
+It is simple to use, fast (transfers over only changes made and not all data),\
+safe (keeps your data safe by checking all declared directories before proceeding in any data manipulation ),\
+reliable and fully customizable."
+COPYRIGHT="2008-2017	Loukas Avgeriou	"
+LICENSE="GNU GPL v3"
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://sourceforge.net/projects/luckybackup/files/0.4.9/source/luckybackup-0.4.9.tar.gz"
```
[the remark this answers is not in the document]
mazbrili: ok

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+	devel:librsync
+	devel:libssl
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:qmake$secondaryArchSuffix >= 5
+	cmd:tar
```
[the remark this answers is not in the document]
mazbrili: ..just for tar xzfs thing..  can i remove it?

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: ok..

Begasus:
Nice work @mazbrili only 2 issues when I run the program (without checking if backup works) :+1: 
The icon in the About window isn't showing, and I can't change the language (there is nothing showing up in the language droplist)

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+reliable and fully customizable."
+COPYRIGHT="2008-2017	Loukas Avgeriou	"
+LICENSE="GNU GPL v3"
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://sourceforge.net/projects/luckybackup/files/0.4.9/source/luckybackup-0.4.9.tar.gz"
+CHECKSUM_SHA256="c161e98a11de5ad0a32a8e060e3fd45695a93ea3adf0581764a6b5d32e8da23b"
+SOURCE_DIR="luckybackup"
+SOURCE_FILENAME="luckybackup-$portVersion.tar.gz"
+SOURCE_DIR="luckybackup-$portVersion"
```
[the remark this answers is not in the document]
mazbrili: oh i see.. ok done

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,92 @@
+SUMMARY="...a powerful, fast and reliable backup & sync tool"
```
alaviss:
```diff
-SUMMARY="...a powerful, fast and reliable backup & sync tool"
+SUMMARY="A powerful, fast and reliable backup & sync tool"
```

`SUMMARY` should start with a capital letter, not ellipsis.

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,92 @@
...
+DESCRIPTION="an application that backs-up and/or synchronizes any directories with the power\
+of rsync.It is simple to use, fast (transfers over only changes made and not all data),safe \
+(keeps your data safe by checking all declared directories before proceeding in any data \
+manipulation ),reliable and fully customizable."
```
alaviss:
```diff
-DESCRIPTION="an application that backs-up and/or synchronizes any directories with the power\
-of rsync.It is simple to use, fast (transfers over only changes made and not all data),safe \
-(keeps your data safe by checking all declared directories before proceeding in any data \
-manipulation ),reliable and fully customizable."
+DESCRIPTION="An application that backs-up and/or synchronizes any directories with the power \
+of rsync. It is simple to use, fast (transfers over only changes made and not all data),safe \
+(keeps your data safe by checking all declared directories before proceeding in any data \
+manipulation), reliable and fully customizable."
```

Some styling fixes.

app-backup/luckybackup/luckybackup-0.4.9.recipe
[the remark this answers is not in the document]
alaviss: If it builds without `cmd:tar` then I don't see why you need it here.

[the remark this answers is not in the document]
mazbrili: i see.. done edit

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+INSTALL()
+{
+	#create directories
+	mkdir -p $appsDir/LuckyBackup/{manual,translations}
+	#install
+	install -T ./luckybackup "$appsDir"/LuckyBackup/LuckyBackup
+
+	# copy manpages
+	mkdir -p $manDir/man8
```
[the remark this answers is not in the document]
mazbrili: thanks.. done edit

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: ok done

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,91 @@
+SUMMARY="A powerful, fast and reliable backup & sync tool"
+DESCRIPTION="An application that backs-up and/or synchronizes any directories with the power \
+of rsync. It is simple to use, fast (transfers over only changes made and not all data),safe \
```
alaviss:
```diff
-of rsync. It is simple to use, fast (transfers over only changes made and not all data),safe \
+of rsync. It is simple to use, fast (transfers over only changes made and not all data), safe \
```

mazbrili: ok done

mazbrili:
> Nice work @mazbrili only 2 issues when I run the program (without checking if backup works) 👍
> The icon in the About window isn't showing, and I can't change the language (there is nothing showing up in the language droplist)

i make post on our forum https://discuss.haiku-os.org/t/help-problem-porting-luckybackup/7570
please check

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,91 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://downloads.sourceforge.net/luckybackup/luckybackup-$portVersion.tar.gz"
+CHECKSUM_SHA256="c161e98a11de5ad0a32a8e060e3fd45695a93ea3adf0581764a6b5d32e8da23b"
+ADDITIONAL_FILES="luckybackup.rdef.in"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
```
fbrosson: Suggestion:
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+fi
```

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,91 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libQt5Core$secondaryArchSuffix
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Network$secondaryArchSuffix
+	lib:libQt5WebKit$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
```
fbrosson:
```diff
 	lib:libQt5Widgets$secondaryArchSuffix
-	lib:libz$secondaryArchSuffix
 	lib:librsync$secondaryArchSuffix
 	lib:libssl$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
 	"
```
(sorting :)

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,91 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libGL$secondaryArchSuffix
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Network$secondaryArchSuffix
+	devel:libQt5WebKit$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
```
fbrosson:
```diff
 	devel:libQt5Widgets$secondaryArchSuffix
-	devel:libz$secondaryArchSuffix
 	devel:librsync$secondaryArchSuffix
 	devel:libssl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
 	"
```
(more sorting :)

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: ok done

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,91 @@
+SUMMARY="A powerful, fast and reliable backup & sync tool"
+DESCRIPTION="An application that backs-up and/or synchronizes any directories with the power \
+of rsync. It is simple to use, fast (transfers over only changes made and not all data), safe\
+ (keeps your data safe by checking all declared directories before proceeding in any data \
+manipulation), reliable and fully customizable."
```
alaviss: Sorry, I should have checked with a text editor. This description needs to be wraped so that it won't breach 80-collumn limit. Below is a wrapped version for your convenience. 

```sh
DESCRIPTION="An application that backs-up and/or synchronizes any directories \
with the power of rsync. It is simple to use, fast (transfers over only \
changes made and not all data), safe (keeps your data safe by checking all \
declared directories before proceeding in any data manipulation), reliable and \
fully customizable."
```

app-backup/luckybackup/luckybackup-0.4.9.recipe
fbrosson: The purpose of the above suggestion is to be friendly to users of x86_64 hybrid, even if builds of that platform are not yet available.
What "friendly" means is that the repository cache won't be cluttered with recipes caches which are not needed: Indeed, on x86_64 hybrid, we don't need `luckybackup_x86` because we will have `luckybackup`.

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+HOMEPAGE="http://luckybackup.sourceforge.net"
+SOURCE_URI="https://downloads.sourceforge.net/luckybackup/luckybackup-$portVersion.tar.gz"
+CHECKSUM_SHA256="c161e98a11de5ad0a32a8e060e3fd45695a93ea3adf0581764a6b5d32e8da23b"
+ADDITIONAL_FILES="luckybackup.rdef.in"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+ SECONDARY_ARCHITECTURES="x86"
```
fbrosson: Oops, there's a white space before `SECONDARY_ARCHITECTURES`.
(I guess it came from a copy/paste.)

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,93 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:qmake$secondaryArchSuffix >= 5
+	"
+
+BUILD()
+{
+	qmake
```
fbrosson: Thanks for the update!

I saw that there is a `luckybackup.pro` file. I guess it could be passed to `qmake`:
```diff
- 	qmake
+	qmake luckybackup.pro
```

BTW, last time I built the app I noticed that the License tab of the **_About_** box displays an error saying it cannot find the license. It would be nice to fix this :)

Maybe patching `luckybackup.pro` with `sed` in `PATCH` would allow to fix the "License not found" issue described above.
(If you need some hints for this I can guide you.)

Other possible improvements could be to move some files to more standard directories.
The documentation, for example, would be better in `"$docDir"` instead of having them in a subdirectory of `"$appsDir"/LuckyBackup/`.
OK, this is polishing, and we can do this later if you prefer.

mazbrili: ok done.. 
for about box. . i can edit that. .but i afraid that .. if other platform show ok but not on haiku.. maybe its related to other bug.. qt issue maybe.. 
what do you think?

mazbrili: sorry.. i misunderstand your meaning.. ok i'll check cannot found license issue..

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: ok.. updated

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: done

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: i've been update recipe.. it is ok when execute from terminal on path installed (/system/apps/LuckyBackup/LuckyBackup)
license show... but not when executed via menu haiku

app-backup/luckybackup/luckybackup-0.4.9.recipe
mazbrili: @alaviss  as you suggested have been change.. please review again..

app-backup/luckybackup/luckybackup-0.4.9.recipe
fbrosson: Oops, I'm still getting the error: `Could not locate the license file!!`
I believe this can be fixed by telling qmake where the license should be, using `make install` to install it and leave it there. Installing the license file with cp is probably what prevents the app from finding it because it is probably looking in `/usr/share/doc/luckybackup/license/`.
Maybe something like:
```bash
PATCH()
{
	sed -i \
		-e "/^license.path =/ s|=.*|= $docDir|;" \
		luckybackup.pro
}
```
could help, even if turns out that we cannot use `qmake install` to install it.
So maybe patching `luckybackup.pro` and modifying the target directory of the `cp` in `INSTALL` can do the trick.

Begasus:
LGTM, as reported the issues comming up are mostly gone when launched from Terminal in $appsDir, Maybe you could open an issue for the remaining ones?
1 When launched from the Deskbar Menu it doesn't show the license, you can't switch languages, manual isn't launching (those work fine when launched in Terminal
2 The missing Icon in the About box
What do you think @fbrosson?
PS, checked to see if I could localy create a backup from a folder to the Desktop, halted with an error "Empty line passed to function"

mazbrili:
image

i'm found path is hardcoded on source .. need assist to fix it

fbrosson:
@mazbrili, in https://github.com/haikuports/haikuports/pull/3183#discussion_r222870218 I was suggesting to patch the file `luckybackup.pro` with sed in `PATCH` to fix the issue about the license not being found.
There are many other hard-coded paths in that file, and I think digging in this direction might help.

Have you noticed that haikuporter generates a `wrapper-script` file in `/boot/home/haikuports/app-backup/luckybackup/work-0.4.9` (or `/boot/home/haikuports/app-backup/luckybackup/work-x86-0.4.9` if you're building for x86 secondary arch on x86_gcc2 hybrid) when you kick a build? That file defines lots of variables. There will certainly be some in relation with the translations directory. If you wish I can make a few suggestions of sed commands for patching `luckybackup.pro`, but I'll let you investigate first ;)

mazbrili:
> @mazbrili, in [#3183 (comment)](https://github.com/haikuports/haikuports/pull/3183#discussion_r222870218) I was suggesting to patch the file `luckybackup.pro` with sed in `PATCH` to fix the issue about the license not being found.
> There are many other hard-coded paths in that file, and I think digging in this direction might help.
> 
> Have you noticed that haikuporter generates a `wrapper-script` file in `/boot/home/haikuports/app-backup/luckybackup/work-0.4.9` (or `/boot/home/haikuports/app-backup/luckybackup/work-x86-0.4.9` if you're building for x86 secondary arch on x86_gcc2 hybrid) when you kick a build? That file defines lots of variables. There will certainly be some in relation with the translations directory. If you wish I can make a few suggestions of sed commands for patching `luckybackup.pro`, but I'll let you investigate first ;)

..thank you for your suggestion... i'm sure it can be done.. but need time to do that.. 
i'll try to fix it when time available..

mazbrili:
i've been make patch.. tested.. ok 
please retest again

fbrosson:
@mazbrili, I just pushed a tiny commit to remove a few unneeded lines in the patchset.
I also moved `PATCHES` before `ADDITIONAL_FILES`, as recommended by the [HaikuPorter-Guidelines#ordering](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering).

Regarding the icon not being shown, this is a minor issue which can be fixed later. That said, there is no urgency, so if you would like to try to fix this before the merge, you are welcome. I think we'll probably merge before next tuesday in order to create a task for the Google Code-in contest it if the issue is not fixed yet.

Regarding the `~/.luckybackup` file, if you can find it in the sources and hardcoded it to `~/config/settings/luckybackup` then that would allow us to mark the recipe as tested. Otherwise we need to mark it as untested.
(I did not dive into the source code yet, but I guess @Begasus already did :) So maybe he can point you to the interesting files.)

Finally, regarding the installation directories, I'd like to suggest a few neutral changes. But I'll do so in a separate comment.

Begasus:
I did some checking, but haven't found a sollution yet also (ps, not familiar with the wrapper.script yet, so I need some investigation on that part) :)

fbrosson:
Here are the changes I have in mind:
* `apps/LuckyBackup/LuckyBackup` :arrow_right: `apps/luckybackup`
* `apps/LuckyBackup/license/`:arrow_right: `data/luckybackup/`
* `apps/LuckyBackup/manual/`:arrow_right: `data/luckybackup/manual/`
* `apps/LuckyBackup/translations/`:arrow_right: `data/luckybackup/translations/`

I believe these changes can be obtained by patching the `luckybackup.pro` file.
And at the end of `INSTALL` we would change
```diff
-	addResourcesToBinaries luckybackup.rdef "$appsDir"/LuckyBackup/LuckyBackup
+	addResourcesToBinaries luckybackup.rdef "$appsDir"/luckybackup
-	addAppDeskbarSymlink "$appsDir"/LuckyBackup/LuckyBackup
+	addAppDeskbarSymlink "$appsDir"/luckybackup LuckyBackup
```
The advantages of this layout are that:
1) The only file we create in `$appsDir` (i.e. `/system/apps/`) will be the app itself, and we won't need to create any other directory in that tree because we'll be creating them in `$dataDir/luckybackup/` (i.e. `/system/data/luckybackup/`)
2) Typing `luckybackup` from the command line will launch the app because `/boot/system/apps` _is_ in the `PATH`, while with the current layout, typing `LuckyBackup` won't work because `/boot/system/LuckyBackup` is _not_ in the `PATH`. To launch the app from the command line with the current layout one needs to type the full path: `/system/apps/LuckyBackup/LuckyBackup`.

That's why I think we should try to use the installation directories described above.

diversys:
Please use LuckyBackup as a binary name.

fbrosson:
I don't think we should rename it to LuckyBackup. The author chose to call his runtime luckybackup so I think we should respect his choice. Moreover, the users of this app already know that it can be invoked with `luckybackup`, not `LuckyBackup`.

I believe we should only use `LuckyBackup` in the call to `addAppDeskbarSymlink`. This is enough to have the app shown as LuckyBackup in the Applications menu.

That said, if there is a consensus for renaming the runtime to `LuckyBackup` then I'm ok with that.

mazbrili:
@fbrosson your explanation is very good.. i'm not yet familiar with haiku system hierarchy..  that change is most suitable i think.. but right now i'm still learning to achieve like that..

app-backup/luckybackup/luckybackup-0.4.9.recipe
```diff
@@ -0,0 +1,97 @@
...
+	# copy manpages
+	mkdir -p "$manDir"/man8
+	cp -r ./manpage/*.8 "$manDir"/man8/
+
+	# copy html documentation
+	cp -r ./manual/* "$appsDir"/LuckyBackup/manual/
+	# copy translation
+	cp -rd ./translations/*.qm "$appsDir"/LuckyBackup/translations/
+	# copy license
+	cp -rd ./license/* "$appsDir"/LuckyBackup/license/
```
fbrosson: I was if we could use `make install` and it turns out that we can do that if we patch the `luckybackup.pro`. I tried something and it allows to simplify `INSTALL` and switch to the layout I was suggesting... but the bad news is that the issue about the license not being displayed (`Could not locate the license file!`) is back...
And the "Help" > "LuckyBackup Handbook" menu displays an error:
```
ERROR
Could not locate the file: manual/index.html
```
(I did not check if it was OK without my changes.)
Anyway, this would require more digging...
So I don't know if this is worth the trouble.
Maybe we should just investigate about the settings file.
```diff
 	cmd:qmake$secondaryArchSuffix >= 5
 	"
 
+PATCH()
+{
+	sed -i \
+		-e "s|^\(documentation\.path =\).*|\1 $docDir|" \
+		-e "s|^\(translations\.path =\).*|\1 $dataDir/luckybackup|" \
+		-e "s|^\(manpage\.path =\).*|\1 $manDir/man8|" \
+		-e "s|^\(license\.path =\).*|\1 $docDir|" \
+		luckybackup.pro
+}
+
 BUILD()
 {
 	qmake luckybackup.pro
 	make $jobArgs
 }
 
 INSTALL()
 {
-	#create directories
-	mkdir -p "$appsDir"/LuckyBackup/{manual,translations,license}
-	#install
-	install -T ./luckybackup "$appsDir"/LuckyBackup/LuckyBackup
-
-	# copy manpages
-	mkdir -p "$manDir"/man8
-	cp -r ./manpage/*.8 "$manDir"/man8/
-
-	# copy html documentation
-	cp -r ./manual/* "$appsDir"/LuckyBackup/manual/
-	# copy translation
-	cp -rd ./translations/*.qm "$appsDir"/LuckyBackup/translations/
-	# copy license
-	cp -rd ./license/* "$appsDir"/LuckyBackup/license/
+	make install
+	install -d "$appsDir"
+	install -T luckybackup "$appsDir"/LuckyBackup
+	gunzip "$manDir"/man8/luckybackup.8.gz
 
 	local APP_SIGNATURE="application/x-vnd.LuckyBackup"
 	local MAJOR="`echo "$portVersion" | cut -d. -f1`"
 	local MIDDLE="`echo "$portVersion" | cut -d. -f2`"
 	local MINOR="`echo "$portVersion" | cut -d. -f3`"
 	local LONG_INFO="$SUMMARY"
 	sed \
 		-e "s|@APP_SIGNATURE@|$APP_SIGNATURE|" \
 		-e "s|@MAJOR@|$MAJOR|" \
 		-e "s|@MIDDLE@|$MIDDLE|" \
 		-e "s|@MINOR@|$MINOR|" \
 		-e "s|@LONG_INFO@|$LONG_INFO|" \
 		"$portDir"/additional-files/luckybackup.rdef.in > luckybackup.rdef
 
-	addResourcesToBinaries luckybackup.rdef "$appsDir"/LuckyBackup/LuckyBackup
-	addAppDeskbarSymlink "$appsDir"/LuckyBackup/LuckyBackup
+	addResourcesToBinaries luckybackup.rdef "$appsDir"/LuckyBackup
+	addAppDeskbarSymlink "$appsDir"/LuckyBackup
 }
```

Begasus: Maybe change:
* `apps/LuckyBackup/license/`:arrow_right: `data/luckybackup/`
into ?
* `apps/LuckyBackup/license/`:arrow_right: `data/luckybackup/license/`
Also with sed you used $docDir?

fbrosson:
The only remaining issue is the icon which is not being displayed.
For the default config, I've pushed a dirty "fix" with hard-coded paths.

app-backup/luckybackup/luckybackup-0.4.9.recipe
fbrosson: `src/global.cpp` had everything we needed to fix most issues. :smile:

app-backup/luckybackup/patches/luckybackup-0.4.9.patchset
```diff
@@ -0,0 +1,84 @@
...
+-QString relativeLicense = "license/gpl.html";
+-QString systemLicense = "/usr/share/doc/luckybackup/license/gpl.html";
++QString relativeLicense = "../data/luckybackup/license/gpl.html";
++QString systemLicense = "/boot/system/data/luckybackup/license/gpl.html";
+ QString suseLicense = "/usr/share/doc/packages/luckybackup/license/gpl.html";
+ QString debianLicense = "/usr/share/common-licenses/GPL-3";
+-QString relativeManual = "manual/index.html";
+-QString systemManual = "/usr/share/doc/luckybackup/manual/index.html";
++QString relativeManual = "../data/luckybackup/manual/index.html";
++QString systemManual = "/boot/system/data/luckybackup/manual/index.html";
```
alaviss: @fbrosson 

Are there any reasons why these can't be changed into `@Template@` which then you can `sed` in the recipe?

I'm a bit concerned about the "relative" paths, what exactly are they relative to? As seen last time, they might be relative to the current working dir, which may causes unintended content injection. I'd like to get rid of these relative paths instead.

app-backup/luckybackup/patches/luckybackup-0.4.9.patchset
```diff
@@ -0,0 +1,84 @@
...
+--- a/src/global.cpp
++++ b/src/global.cpp
+@@ -52,8 +52,8 @@ QString appVersionString = "0.4.9";
+ double const validProfileVersion = 0.21;
+ double const validScheduleVersion = 0.34;
+ double const validSettingsVersion = 0.3;
+-QString luckyBackupDir = myHome + "/."+appName+"/";
+-QString luckyBackupDefaultDir = myHome + "/."+appName+"/";
++QString luckyBackupDir = myHome + "/config/settings/LuckyBackup/";
++QString luckyBackupDefaultDir = myHome + "/config/settings/LuckyBackup/";
```
alaviss: [`QStandardPaths`](https://doc.qt.io/qt-5/qstandardpaths.html), or more specifically `QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)` should be used.

The usage is simple, just

    luckyBackupDir = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation)

simpler than our `find_directory` in fact... And it's even cross-platform.

fbrosson: Cool! This can become an excellent GCI task!
(I'll leave this to another mentor :)

mazbrili:
thanks all @alaviss  @fbrosson @Begasus @scottmc @diversys

app-backup/luckybackup/patches/luckybackup-0.4.9.patchset
waddlesplash: All apps will have to tolerate relative paths eventually, as they can get installed in HOME as well as /system/.

alaviss: Not really, `/package/name-version/.self/` stay the same regardless of installation location. Also, there's `find_path` if you want the fancy `/boot/{home,system}` paths.

PR #3193, opened 2018-10-05
media-libs/libsdl: fix of black box in top-left corner while win lost mouse focus

gl00my:
this pr fixes black box artefact (cursor background) while mouse left windows


diversys:
Thanks!

PR #3198, opened 2018-10-07
smooth: New recipe for smooth 0.8.74.0-pre4.

enzo1982:
The smooth class library is the system and UI library used by fre:ac.

### DIFF
--- a/dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
+++ b/dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
@@ -1,0 +1,107 @@
+SUMMARY="A cross-platform C++ class library"
+DESCRIPTION="smooth is an object oriented C++ class library for Windows, \
+macOS, Linux and most Unix-like operating systems. It provides basic \
+functionality and platform support for applications and libraries.
+
+Features provided by smooth include:
+
+- user interface API with various widgets
+- simple to use multithreading API
+- file and network IO interface
+- completely transparent Unicode and software internationalization support
+- a libxml2 based XML parser"
+HOMEPAGE="http://www.smooth-project.org/"
+COPYRIGHT="1998-2018 Robert Kausch"
+LICENSE="Artistic v2.0"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/smooth/smooth-${portVersion/\~/-}.tar.gz"
+CHECKSUM_SHA256="815f0329b5546c78feeb9ad1813855dfcb7bc062e28de44d0b4aadd7c6625d80"
+SOURCE_DIR="smooth-${portVersion/\~/-}"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libVersion="${portVersion%~*}"
+libVersionCompat="${libVersion##*.} compat >= 0"
+
+PROVIDES="
+	smooth$secondaryArchSuffix = $portVersion
+	lib:libsmooth_${libVersion%.*}$secondaryArchSuffix = $libVersionCompat
+	"
+if [ -z "$commandSuffix" ]; then
+PROVIDES+="
+	cmd:smooth_translator = $portVersion
+	"
+fi
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libbz2$secondaryArchSuffix
+	lib:libcurl$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	smooth${secondaryArchSuffix}_devel = $portVersion
+	devel:libsmooth_${libVersion%.*}$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	smooth$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libbz2$secondaryArchSuffix
+	devel:libcurl$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libjpeg$secondaryArchSuffix
+	devel:libpng16$secondaryArchSuffix
+	devel:libxml2$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:g++$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+PATCH()
+{
+	if [ -n "$secondaryArchSuffix" ]; then
+		sed -i -r \
+			-e "s|/system(/[^/]+)?/lib|/system\1/$relativeLibDir|" \
+			classes/system/dynamicloader.cpp
+	fi
+}
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make prefix="$prefix" bindir="$commandBinDir" libdir="$libDir" \
+		includedir="$includeDir" install
+
+	if [ -n "$commandSuffix" ]; then
+		rm -rf "$commandBinDir"
+	fi
+
+	prepareInstalledDevelLib libsmooth-${libVersion%.*}
+
+	packageEntries devel \
+		"$developDir"
+}


Begasus:
When trying to build smooth_x86 (on gcc2h) I'm getting:
Warning: POLICY ERROR: "/packaging/smooth_x86/bin/smooth-translator" needs library "libsmooth-0.8.74.so.0", but the package doesn't seem to declare that as a requirement

fbrosson:
I also got this. This is because haikuporter only checks for the libraries if they are in `$libDir` which is `$prefix/lib/x86` on x86 secondary arch. Here the recipe is installing them in `$prefix/lib` and this is not OK. It's only OK for primary arch builds, of course.
The build succeeds, but we should pass some parameters to have the libs (and the headers) installed in the correct locations.
I've already tried passing `LIBDIR="$libDir"` in `BUILD` and `LIBDIR="$libDir" INCDIR="$includeDir"` in `INSTALL` but this did not help.
I'll let you dig. If you're stuck then I'll see if I can find something.

Begasus:
Looking into the Makefile atm, no luck so far

enzo1982:
Passing libdir=$libDir to make should fix this. I'll post an update soon.

Begasus:
Latest suggestions work for _x86 build (picked up ok by BoCA too), not checked on x86_64 @fbrosson could you check?

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
```diff
@@ -0,0 +1,95 @@
...
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+fi
+
+libVersion="${portVersion%~*}"
+libVersionCompat="$libVersion compat >= ${libVersion%.*}"
+
+PROVIDES="
+	smooth$secondaryArchSuffix = $portVersion
+	lib:libsmooth$secondaryArchSuffix = $libVersionCompat
```
fbrosson: I'm not getting any `libsmooth.so` symlink nor a `libsmooth.so.0` file. Well, I'm getting them with different names:
```
lib
  libsmooth-0.8.74.so -> libsmooth-0.8.74.so.0
  libsmooth-0.8.74.so.0
```
Traditionally we would get:
```
lib
  libsmooth.so.0 -> libsmooth.so.0.8.74
  libsmooth.so.0.8.74
```
I don't know what's the best way to handle your layout. Maybe:
```diff
-libVersion="${portVersion%~*}"
+libRealVersion="${portVersion%~*}"
-libVersionCompat="$libVersion compat >= ${libVersion%.*}"
+libVersionCompat="${libRealVersion%%.*}"
 
 PROVIDES="
 	smooth$secondaryArchSuffix = $portVersion
-	lib:libsmooth$secondaryArchSuffix = $libVersionCompat
- 	lib:libsmooth_${libVersion%.*}$secondaryArchSuffix = $libVersionCompat
+ 	lib:libsmooth_${libRealVersion%.*}$secondaryArchSuffix = $libVersionCompat
```
I think it would work, but I'm not satisfied with this.

I suspect that any package that would require the lib from the smooth package would have to be rebuilt everytime the version of smooth is bumped.

Wouldn't it be better to have a `libsmoth.so.0` symlink point to `libsmooth.so.0.8.74` ?

BTW, the devel package does not provide and symlinks for the lib. We would normally do this with:
```
	prepareInstalledDevelLib libsmooth
	fixPkgconfig
```
but here it would have to be:
```
	prepareInstalledDevelLib libsmooth-${libRealVersion%.*}
```

This is not an easy question. Let's see what other contributors think about this.

enzo1982: The smooth API and ABI still change quite often, that's why I would like to keep it as `libsmooth-0.8.74.so`. A program linked against smooth 0.8.74.0 will be compatible with 0.8.74.3, but might not be compatible with smooth 0.8.75.0. I plan to keep the ABI stable when going to a 0.9 release.

I'll add the prepareInstalledDevelLib call though.

enzo1982:
> ```diff
> -	make prefix=$prefix bindir=$binDir libdir=$libDir includedir=$includeDir install
> +	make prefix=$prefix bindir=$commandBinDir libdir=$libDir includedir=$includeDir install
> ```

What's the point of defining a $commandBinDir over using $binDir directly?

enzo1982:
The latest revision removes the unversioned PROVIDES declarations and adds the call to prepareInstalledDevelLib.

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
```diff
@@ -0,0 +1,95 @@
...
+	"
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make prefix=$prefix bindir=$binDir libdir=$libDir includedir=$includeDir install
```
fbrosson: > > ```diff
> > -	make prefix=$prefix bindir=$binDir libdir=$libDir includedir=$includeDir install
> > +	make prefix=$prefix bindir=$commandBinDir libdir=$libDir includedir=$includeDir install
> > ```
>
> What's the point of defining a $commandBinDir over using $binDir directly?

Good catch. What we need here is:
```diff
-	make prefix=$prefix bindir=$binDir libdir=$libDir includedir=$includeDir install
+	make prefix=$prefix bindir=$prefix/bin libdir=$libDir includedir=$includeDir install
```
We used to define `commandBinDir` as `$binDir` on all arches, except on x86 secondary arch, where we define it as `$prefix/bin`.
(On x86 secondary arch `$binDir` is `$prefix/bin/x86` but we don't want to use it because we want the runtime to be in `$prefix/bin` in all cases, that is, even if we're build for x86 secondary arch.)

Now, given that:
1) smooth is not available on x86_gcc2 primary arch
2) smooth_x86 has its `$prefix/bin` removed on all arches except when `$commandSuffix` empty,

we don't need to define `commandBinDir` anymore. We can use `$prefix/bin` in all cases.

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
```diff
@@ -0,0 +1,95 @@
...
+	devel:libpng16$secondaryArchSuffix
+	devel:libxml2$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:gcc$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:pkg_config$secondaryArchSuffix
```
Begasus: Could you sort these? A-Z

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
Begasus: @fbrosson would it be prefered to use the qoutes here too? eg "$prefix" etc

fbrosson: I know some contributors argue that putting quotes around these variables is not needed, but I personally prefer to use them because they **_could_** be useful in some non common configurations: if a user decides to insert a space in the path that is used as the `TREE_PATH` in `haikuports.conf` and then runs `haikuporter --nochroot --test` for a recipe which uses e.g. `$sourceDir` in `TEST`, or chooses to cross-build a `"any"`-arch package which uses e.g. `$prefix` or `$dataDir`, then the build will fail if we don't have the quotes.

So, instead of having the quotes only on those recipes which could require them, I'd rather use them in all new recipes and in all recipes we update for some other reason.

That's why I recommend using them in e.g. `"$prefix"`, `"$dataDir"`, `"$binDir"`,`"$commandBinDir"`,`"$libDir"`, `"$includeDir"`, etc...

Moreover, even if `--nochroot` is currently only usable with `--test`, nobody can claim it won't ever be usable for building recipes.

N.B. I don't use quotes for `relativeBinDir`and other similar `relative*Dir`variables because these are relative to `$prefix` and will never contain any spaces.
That said, I don't object using quotes for them too. It's just that I think its easier to have the quotes adopted for the non-relative variables rather than for the non-relative variables and the relative variables.

enzo1982:
New commit has the following changes:

- remove secondary architecture
- fix library version numbers
- add quotes around path variables
- order build prerequires alphabetically

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
fbrosson: The `cmd:` lines in `BUILD_PREREQUIRES` are now sorted.

Begasus:
> New commit has the following changes:
> 
>     * remove secondary architecture
> 
>     * fix library version numbers
> 
>     * add quotes around path variables
> 
>     * order build prerequires alphabetically

Why remove secondary architecture? Without this and boca freac can't be build on secondary architecture (see my latest comment at freac for that)

enzo1982:
> Why remove secondary architecture? Without this boca and freac can't be build on secondary architecture (see my latest comment at freac for that)

Well, we can leave secondary architecture support in place here. I think there are no issues with it for smooth. For BoCA and fre:ac I would like to remove it now and re-add it later though, for the reasons explained in the fre:ac thread.

fbrosson:
Shall we keep x86 secondary arch, but mark it as untested for now?
Well, we're not supposed to add an arch, even marking it as untested, if it fails to build, but since we know that it can be made to build on x86 secondary arch (with some changes which we did not find yet), I believe we can make an exception here.
```diff
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="?x86"
```
@enzo1982, @Begasus, shall we do that?
(Or would you prefer to keep this PR open for now?)

pulkomandy:
If it doesn't build, mark it as broken. Untested is nonsense in this case.

Begasus:
Both BoCA and smooth build and install ok for secondary architecture, so they "aren't" broken

enzo1982:
I'll re-add the secondary architecture for smooth. This package is not broken.

BoCA, however, will build, install and run, but will not find it's components on a secondary architecture which renders it pretty useless. But I think we are on a good path to a solution there with @fbrosson's ideas about patching the relevant source files when building for a secondary arch.

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
```diff
@@ -0,0 +1,104 @@
...
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+PATCH()
+{
+	if [ -n "$secondaryArchSuffix" ]; then
+		sed -i -r\
```
fbrosson: Suggestion:
```diff
-		sed -i -r\
+		sed -i -r \
```
(It looks nicer with a space :)

Everything else is OK!

dev-libs/smooth/smooth-0.8.74.0~pre4.recipe
```diff
@@ -0,0 +1,104 @@
...
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make prefix="$prefix" bindir="$prefix"/bin libdir="$libDir" includedir="$includeDir" install
+
+	if [ -n "$commandSuffix" ]; then
+		rm -rf "$prefix"/bin
```
fbrosson: The first line in `INSTALL` is > 80 chars. We should fold it. While at it, shall we use the following variant?
```diff
 INSTALL()
 {
-	make prefix="$prefix" bindir="$prefix"/bin libdir="$libDir" includedir="$includeDir" install
+	make prefix="$prefix" bindir="$commandBinDir" includedir="$includeDir" \
+		libdir="$libDir" install
 
 	if [ -n "$commandSuffix" ]; then
-		rm -rf "$prefix"/bin
+		rm -rf "$commandBinDir"
 	fi
```
Of course, if we choose to use `"$commandBinDir"` (instead of `"$prefix"/bin`) then we need to define it :P
```diff
 SECONDARY_ARCHITECTURES="x86"
 
 commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
 if [ "$targetArchitecture" = x86_gcc2 ]; then
 	commandSuffix=
+	commandBinDir=$prefix/bin
 fi
```

enzo1982: Thanks! Just pushed a new update with these changes.

PR #3200, opened 2018-10-07
patch: bump to 2.7.6, switch to tar.xz, add gcc2 patch.

fbrosson:


fbrosson:
If you run the tests for patch `2.7.5` and `2.7.6` on all possible architectures, you can see that on x86_gcc2 there are several crashes, whereas on all other arches there are none.
So this is the motivation for making `patch_x86` available for x86_gcc2h.

Now, why didn't I use a single commit in this PR, instead of two, since I knew I was going to do a version bump?
I guess you already know, but here's the answer for those who might not:
With a single commit I would have had an unfriendly diff for the recipe: git would have shown a `deleted file` and a `new file`, while with two commits I get a nicer diff, showing a `similarity index` > 50% with a `rename from ...` and a `rename to ...`:
```diff
iff --git a/sys-devel/patch/patch-2.7.5.recipe b/sys-devel/patch/patch-2.7.6.recipe
similarity index 54%
rename from sys-devel/patch/patch-2.7.5.recipe
rename to sys-devel/patch/patch-2.7.6.recipe
index 13eb1a1f4..a46abbf51 100644
--- a/sys-devel/patch/patch-2.7.5.recipe
+++ b/sys-devel/patch/patch-2.7.6.recipe
```
Reminder: if the similarity index is smaller than 50% then git does not display the diff this way. It displays it with a first block, with all the lines from the old recipe being deleted, and a second block with all the lines of the new recipe being added, which means that if you want to understand what has changed you need to checkout the old recipe and compare it manually against the new one.

This PR will allow all users of x86_gcc2 to choose which package they wish to use, between `patch` for x86_gcc2 primary arch, and `patch_x86`. This way if we ever notice that patch is crashing _in real life_, I mean, not only when given corner case input like those from the test suite, then the fix will be already available: installing `patch_x86`.

Why did I not drop x86_gcc2 primary arch in `2.7.6` and keep the recipe for `2.7.5`? Well, if I had done so then x86_gcc2 non hybrid users would have been stuck with `2.7.5`, which passes less tests than `2.7.6`.
When x86_gcc2 gets deprecated it will be easy to remove the gcc2 patch because it is in a separate patchset.

Regarding the crashes, `2.7.5` (for x86_gcc2 primary arch) already crashed. So there are no regressions here.

I hope this answers all the possible questions :wink:

pulkomandy:
I would prefer that we remove the gcc2 version and use a REPLACES entry so that the old package is removed, if possible. Keeping both versions will be confusing for new users.

fbrosson:
Ok. I'll add a `REPLACES` (and bump `REVISION`, of course.)
Shall I get rid of the x86_gcc2 build right now? (Scenario A)
Or do we keep it until the next release of patch, since it is already in the repo? (Scenario B)

I had not dared to remove x86_gcc2 primary arch. I guess I should have marked it as untested and stick to the gzip tarball.
Hmm, shall I switch back to `tar.gz` and mark x86_gcc2 as untested? (Scenario C)
This would allow those users who like to play with x86_gcc2 non hybrid to build patch `2.7.6` themselves.
Then, when the next version of patch becomes available I'll drop the gcc2 patchset if the one we have right now does not apply nicely.

Ah, there's something else we can do: drop the third patch of the patchset and add `cmd:bison` if we're building for x86_gcc2 primary arch. (Scenario D)

Shall we go for scenario B? Or do you prefer A?

pulkomandy:
I prefer we stop spending any efforts getting things to run with gcc2. If it doesn't work, mark it as broken and that's it.

PR #3208, opened 2018-10-11
gnu_efi_kernel: bump version to 3.0.9

return:


fbrosson:
Thanks @return for your PR!
gnu-efi-3.0.9 was released 205 minutes ago.
Have you had the time to make some checks like those I did for #2425 ?
It would be nice to check that with this release installed we can build the EFI loader with:
```
jam -q -sHAIKU_BOOT_PLATFORM=efi haiku_loader.efi
```
and that it is able to boot Haiku x86_64. Thanks!

return:
I've tried building the haiku_loader using gnu-efi-3.0.9 and can confirm that it boots fine on real hardware. Tested R1 beta1 hrev52295+100 on x86_64.

PR #3211, opened 2018-10-13
sword, bump version

Begasus:


Begasus:
Not checked on x86_64 (yet)

app-text/sword/sword-1.8.1.recipe
```diff
@@ -6,18 +6,21 @@ societies to write new Bible software more quickly and easily. We also create \
...
 the Bible, and have a growing collection of over 200 texts in over 50 languages."
 HOMEPAGE="http://www.crosswire.org/sword"
-COPYRIGHT="1998-2014 CrossWire Bible Society"
+COPYRIGHT="1998-2018 CrossWire Bible Society"
 LICENSE="GNU GPL v2"
-REVISION="3"
-SOURCE_URI="http://www.crosswire.org/ftpmirror/pub/sword/source/v1.7/sword-1.7.3.tar.gz"
-CHECKSUM_SHA256="5a3d87ff155d5ecb0cfec052ba333b2b74d9273e2cc66fb1ca75747dfd8ea9ea"
+REVISION="1"
+SOURCE_URI="http://www.crosswire.org/ftpmirror/pub/sword/source/v1.8/sword-$portVersion.tar.gz"
```
fbrosson: Suggestion:
```diff
-SOURCE_URI="http://www.crosswire.org/ftpmirror/pub/sword/source/v1.8/sword-$portVersion.tar.gz"
+SOURCE_URI="http://www.crosswire.org/ftpmirror/pub/sword/source/v${portVersion%.*}/sword-$portVersion.tar.gz"
```
(This is safe because all versions seem to follow the `x.y.z` pattern.)

app-text/sword/sword-1.8.1.recipe
```diff
@@ -6,18 +6,21 @@ societies to write new Bible software more quickly and easily. We also create \
...
-CHECKSUM_SHA256="5a3d87ff155d5ecb0cfec052ba333b2b74d9273e2cc66fb1ca75747dfd8ea9ea"
+REVISION="1"
+SOURCE_URI="http://www.crosswire.org/ftpmirror/pub/sword/source/v${portVersion%.*}/sword-$portVersion.tar.gz"
+CHECKSUM_SHA256="ce9aa8f721a737f406115d35ff438bd07c829fce1605f0d6dcdabc4318bc5e93"
 
-ARCHITECTURES="!x86_gcc2 x86 ?x86_64"
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+GLOBAL_WRITABLE_FILES="settings/sword.conf keep-old"
```
fbrosson: I don't know if this is desirable for this particular recipe, but maybe we could try this:
```diff
-GLOBAL_WRITABLE_FILES="settings/sword.conf keep-old"
+GLOBAL_WRITABLE_FILES="settings/sword.conf auto-merge"
```
(I also don't know if `auto-merge` is already implemented, but if it's not then it will behave like `keep-old` if I'm not mistaken.)

BTW, I've never seen this file, but if it looks like an INI file then I guess it will be fine to use `auto-merge`.

Begasus: I'll update to it, this is the content
```
[Install]
 DataPath=/packages/sword_x86-1.8.1-1/.self/data/sword
```

app-text/sword/sword-1.8.1.recipe
```diff
@@ -54,33 +63,43 @@ REQUIRES_devel="
...
+	runConfigure ./configure
 	make $jobArgs
 }
 
 INSTALL()
 {
-	make $jobArgs install
+	make install
+
+	rm "$libDir"/libsword.la
```
fbrosson: Suggestion:
```diff
-	rm "$libDir"/libsword.la
+	rm -f "$libDir"/libsword.la
```
(This way, if the build tools are taught, some day, to stop creating the libtool files then this recipe will not break suddenly, with all the other recipes which still don't use `-f`.)

app-text/sword/sword-1.8.1.recipe
fbrosson: What you can do is insert a
```
false
```
near the end of `INSTALL`, then kick a build (that will fail, leaving the installed files in the packaging subdirectory) and then `cd work-*/packaging/sword/.settings/` to test the sed command in there (or in a copy somewhere else.)

fbrosson: Oops, I forgot to mention that you'll need to write a tiny script that will define dataDir with the value you have in `work-*/wrapper-script` before the actual calling the sed command (with `/tmp/sword.conf`instead of `"$settingsDir"/sword.conf`.)

Begasus: With everything I tried so far I ended up with a patched file in the "packaging" folder, but somehow it still reverts to the default one

fbrosson: Did you check the contents of the `sword.conf` in your package by extracting it to a temporary directory?
In case you would like to copy/paste:
```bash
mkdir tmp && cd tmp
package extract /boot/home/haikuports/packages/sword_x86-1.8.1-1-x86_gcc2.hpkg
```

If the `sword.conf` in your package is OK then you might need to:
1) uninstall sword
2) `rm /system/settings/sword.conf`
3) `rm -rf /system/packages/administrative/writable-files/sword-*`
4) reinstall a fresh hpkg of sword.

Begasus: With some change in the Makefiles I'm getting the next now:
```
[Install]
DataPath=/boot/system/data
```
I'm not a sed guru, so if this is ok I'll create a patchset for it :)

Begasus:
I guess this looks ok?
[Install]
DataPath=/packages/sword_x86-1.8.1-1/.self/data/

Begasus:
Bumped, should be ok now, could use someone to check it on x86_64 :)

app-text/sword/sword-1.8.1.recipe
Begasus: Seems it wasn't so hard after all :) Could you check?

Begasus:
Checked with scriptureguide, builds ok also with this version (ScriptureGuide crashes however with this and previous version from sword)

PR #3212, opened 2018-10-13
speedtest-cli: new recipe

Crestwave:


net-analyzer/speedtest_cli/speedtest_cli-2.0.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+the application being written in Python, \
+differing performance between Python versions, \
+and CPU and Memory capacity and speed."
+HOMEPAGE="https://github.com/sivel/speedtest-cli"
+COPYRIGHT="2012-2018 Matt Martz"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="$HOMEPAGE/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="5e22f2dcce1c9020f33faf559b22727483f022008a2912b85d573e578374b6a0"
+SOURCE_FILENAME="v$portVersion.tar.gz"
```
fbrosson: Thanks @Crestwave for your PR!

Recommendation:
```diff
-SOURCE_FILENAME="v$portVersion.tar.gz"
+SOURCE_FILENAME="speedtest-cli-$portVersion.tar.gz"
```
([HaikuPorter-Guidelines#archive-name](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#archive-name))

fbrosson: Could you please move the recipe from `net-analyzer/speedtest_cli` to `net-analyzer/speedtest-cli`, but keeping the name of the recipe unchanged? (`speedtest_cli-2.0.2.recipe`)
Thanks!

Crestwave: Whoops; I did read that but I thought that SOURCE_FILENAME was for specifying it if it didn’t follow the standard for some reason. Is the dash for the directories a new custom? I looked through the tree first and they had underscores instead. Thanks for the corrections!

fbrosson: You're welcome! That's right. We only need `SOURCE_FILENAME` if the original name does not contain a meaningful basename.
For some recipes, e.g. sg3_utils, we do have an underscore in both the directory and the name of the recipe, but that's because the name of that project really has an underscore.
That said, we do have quite some recipes in directories having an underscore in their names while they could be renamed to use a `-`. I guess we did not move them because doing so makes it harder to track the full git log for a port.
(For new recipes we like to use the best name.)

net-analyzer/speedtest-cli/speedtest_cli-2.0.2.recipe
```diff
@@ -0,0 +1,84 @@
...
+PROVIDES_python="$PROVIDES_python
+	cmd:speedtest_cli_python
+	"
+REPLACES_python="
+	python_speedtest_cli
+	"
+PROVIDES_python3="$PROVIDES_python3
+	cmd:speedtest_cli_python3
+	"
+REPLACES_python3="
```
fbrosson:
```diff
 PROVIDES_python="$PROVIDES_python
 	cmd:speedtest_cli_python
 	"
-REPLACES_python="
-	python_speedtest_cli
-	"
 PROVIDES_python3="$PROVIDES_python3
 	cmd:speedtest_cli_python3
 	"
-REPLACES_python3="
-	python3_speedtest_cli
-	"
```
I think we don't need any `REPLACES*` here, but maybe I'm missing something.

BTW, I tested the program and it works like a charm!

net-analyzer/speedtest_cli/speedtest_cli-2.0.2.recipe
Crestwave: No, I originally thought that you would use it to specify the downloaded filename in that situation; doesn’t make sense, I know. And thanks for the explanation!

Begasus:
Checked and works great, only one question, if I build I get 3 packages, 1 for python and one for python3, but the third one "speedtest_cli-2.0.2-1-any.hpkg" is empty?

net-analyzer/speedtest-cli/speedtest_cli-2.0.2.recipe
Crestwave: Oh, I didn’t understand that at first and all the other Python packages I saw had it so I just added it just in case. It just occurred to me now that the ones listed there are probably an old naming scheme...

fbrosson:
Yeah, we can't get rid of the empty hpkg because it's the base package.

Hmm, would it be desirable to add this block?
```diff
 PROVIDES="
 	$portName = $portVersion
 	"
 REQUIRES="
 	haiku
 	"
+SUPPLEMENTS="
+	speedtest_cli_python
+	speedtest_cli_python3
+	"
+
 BUILD_REQUIRES="
 	haiku_devel
 	"
```
Hmm, if we add this then it might be better to it added inside the "for i" loop.

fbrosson:
haikuporter will complain if we don't have a `PROVIDES` with `speedtest_cli` in it, so this means we can't avoid having the empty package.

PR #3221, opened 2018-10-15
OpenSceneGraph: Enable GL3

kenmays:


games-engines/openscenegraph/openscenegraph-3.6.3.recipe
```diff
@@ -180,6 +173,10 @@ INSTALL()
 	mv $prefix/lib-temp/* $libDir
 	rm -rf $prefix/lib-temp
 
+	# Move plugins to system addons directory
+	mkdir -p $addOnsDir
+	mv $libDir/osgPlugins-$portVersion $addOnsDir
```
extrowerk: Will OSG find them?

games-engines/openscenegraph/openscenegraph-3.6.3.recipe
```diff
@@ -48,7 +48,6 @@ PROVIDES="
 REQUIRES="
 	haiku$secondaryArchSuffix
 	mesa$secondaryArchSuffix
-#	lib:libavcodec$secondaryArchSuffix
```
extrowerk: Please, do not remove them, i plan to enable them later.

games-engines/openscenegraph/openscenegraph-3.6.3.recipe
```diff
@@ -158,7 +150,8 @@ BUILD()
 	mkdir -p build && cd build
 	cmake .. \
 		-DCMAKE_INSTALL_PREFIX=$prefix \
-		-DBUILD_OSG_APPLICATIONS=OFF \
+		-DOSG_GL3_AVAILABLE:BOOL=ON \
+		-DBUILD_OSG_APPLICATIONS:BOOL=OFF \
```
extrowerk: ```:BOOL``` isn't necessary.

games-engines/openscenegraph/openscenegraph-3.6.3.recipe
```diff
@@ -10,7 +10,7 @@ virtual reality industries."
 HOMEPAGE="http://www.openscenegraph.org/"
 COPYRIGHT="2002-2018 Robert Osfield and others"
 LICENSE="OSGPL"
-REVISION="1"
+REVISION="2"
```
fbrosson: @kenmays, you should probably update your master branch and rebase your PR against 3e709df1958efa83eecc90bfe5044b4baff709e3 or any newer commit, and then:
```diff
-REVISION="2"
+REVISION="3"
```

Also, I think it might be nice to revert the minor cleanups, i.e. keep the dependencies which are commented out, if @threedeyes added them on purpose. I guess there might be plans to uncomment these dependencies in the future.

fbrosson:
@kenmays, your PR was based on a commit that was older than 3e709df1958efa83eecc90bfe5044b4baff709e3. I've just rebased it against master.
I also added a commit to revert the cleanups which were removing some dependencies that had been commented out, but I can undo my "revert commit" if you prefer.

games-engines/openscenegraph/openscenegraph-3.6.3.recipe
fbrosson: Right, but it doesn't hurt, and it seems this is a newer syntax which is probably encouraged.
It's like `-DCMAKE_INSTALL_PREFIX:PATH="$prefix"` which we already have in a few other recipes.

That said, I don't have any strong opinion about keeping vs dropping the `:BOOL` "qualifier" :)

kenmays:
@fbrosson - Ok, Fix as you see fit.

PR #3227, opened 2018-10-17
net-libs/libssh: update to 0.8.4

alaviss:
This is a rather hillarious bug:
https://www.libssh.org/security/advisories/CVE-2018-10933.txt

Luckily none of our stuff in HaikuPorts actually uses this.
Thanks @waddlesplash for notifying

Changes to recipe:
- Cleaned up cmake call
- Added ugly hack for `-pie` and `-fstack-protector`, probably a cmake
  bug, but since linux distros have all gone with default-pie and
  default-ssp, it probably got ignored
- Added a _doc subpackage

Torture tests requires: https://review.haiku-os.org/c/haiku/+/632
Some of the torture tests are broken, but I didn't log them...

Tests are passed on both nightlies. I would need some help testing this on R1Beta1.


net-libs/libssh/libssh-0.8.4.recipe
```diff
@@ -57,43 +58,58 @@ BUILD_PREREQUIRES="
 	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
-	cmd:libtoolize
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+if [ -z "$secondaryArchSuffix" ]; then
+	ARCHITECTURES_doc="any"
```
fbrosson: Do we really want to move the doc to a subpackage ?
x86_gcc2(h) users won't be able to build it, nor to install it from a repo. The only way to install that doc subpackage would be to copy it from a x86_64 system and install it manually on the x86_gcc2(h) system.

alaviss: It's an `any` package isn't it? Shouldn't our builders be able to reuse the result from x86_64?

And I moved it since it's 2MB in size, while the `_devel` is only 224KB

fbrosson: As you can image, it would be possible, technically speaking, and with "some" work, to teach the build bots running on x86_gcc2h to detect a recipe which is broken on x86_gcc2 primary arch but has one or more _"any"_-arch subpackages and have them fetched from the x86_64 repo. But I believe this would be too much work for a very small benefit. So until x86_gcc2 gets deprecated we'll probably have to keep docs in the base package for those recipes which do not build on x86_gcc2 primary arch.
That said, we can still make exceptions and decide that a given doc subpackage won't be available for x86_gcc2(h).
I don't have a strong opinion about this.
BTW, "any"-arch packages or subpackages created on x86_gcc2 and those created on x86_64 live in separate pools, so creating a doc subpackage here wouldn't even save any disk space.
BTW, a few "any"-arch packages (such as ExifTool if I'm not mistaken) are not perfectly indentical (because they have subdirectoris with names that refer to the architecture where they got built) so if we wanted to save some disk space we would first have to check that they are interchangeable, i.e. that e.g. an ExifTool package built on x86_64 can be installed on x86_gcc2 and work.

Related (closed) issue: https://github.com/haikuports/haikuporter/issues/84

alaviss: > BTW, "any"-arch packages or subpackages created on x86_gcc2 and those created on x86_64 live in separate pools, so creating a doc subpackage here wouldn't even save any disk space.

@pulkomandy, @waddlesplash can we have them shared, please? We can have an `any` pool like most distros out there.

> BTW, a few "any"-arch packages (such as ExifTool if I'm not mistaken) are not perfectly indentical (because they have subdirectoris with names that refer to the architecture where they got built) so if we wanted to save some disk space we would first have to check that they are interchangeable, i.e. that e.g. an ExifTool package built on x86_64 can be installed on x86_gcc2 and work.

That's a bug in the recipe. An `any`-arch package means that it's independent of the underlying architecture, if the contents depends on the arch that's used to build it, it's not `any`-arch.

net-libs/libssh/libssh-0.8.4.recipe
```diff
@@ -108,6 +124,12 @@ INSTALL()
 	packageEntries devel \
 		"$developDir" \
 		"$libDir"/cmake
+
+	if [ -z "$secondaryArchSuffix" ]; then
+		install -d -m 755 "$developDocDir"
+		mv -T doc/html "$developDocDir"
```
fbrosson: I'd replace  `mv -T doc/html "$developDocDir"` by `cp -r doc/html/* "$developDocDir"` because using `mv` makes the **second** and following invocations of `hp libssh` fail until `/boot/home/haikuports/net-libs/libssh/work-0.8.4/flag.build` is deleted or `hp -c libssh` is called:
```
...
-- Installing: /packages/libssh-0.8.4-1/.self/lib/libssh.so
/packages/libssh_devel-0.8.4-1/.self
mv: cannot stat 'doc/html': No such file or directory
```

BTW, I don't know if this would be an improvement, but if we wanted to have the html files installed in `develop/documentation/libssh/html/` instead of `develop/documentation/libssh/` then, of course, we should use `cp -r doc/html "$developDocDir"` instead of `cp -r doc/html/* "$developDocDir"`.

fbrosson:
* Test results for libssh 0.8.4 on x86_64: `100% tests passed, 0 tests failed out of 20`
* Test results for libssh 0.8.**4** on x86 (second arch): `45% tests passed, 11 tests failed out of 20`
```
The following tests FAILED:
	  2 - torture_callbacks (Failed)
	  6 - torture_misc (Failed)
	  7 - torture_config (Failed)
	  8 - torture_options (Failed)
	 13 - torture_packet (Failed)
	 14 - torture_keyfiles (Failed)
	 15 - torture_pki (Failed)
	 16 - torture_pki_rsa (Failed)
	 17 - torture_pki_ed25519 (Failed)
	 18 - torture_pki_dsa (Failed)
	 19 - torture_pki_ecdsa (Failed)
Errors while running CTest
```
* Test results for libssh 0.8.**3** on x86 (second arch): `95% tests passed, 1 tests failed out of 19`
```
The following tests FAILED:
	  6 - torture_misc (Failed)
Errors while running CTest
```
(I did not investigate.)

fbrosson:
`libssh-0.8.3` was added (3a101a88872550a39814d3077f1cbffd28836411) on September 22 and then got its `REVISION` bumped (27234a78fa008a6a53b8ebecd436037a2067df1d, 8af0eb1184afe2803705db56e0e43f61ac897bf5) on September 26, before the release of R1B1.

alaviss:
What I mean was, was those test results from R1B1?

fbrosson:
Not exactly. The tests were run on hrev52397, which is newer than R1B1.

alaviss:
I don't feel adventurous enough for that... maybe I'll disable it for `_x86` and use a patched `0.8.3` instead. No idea if the buildbots would build the `0.8.3` version in that case.

fbrosson:
Yep, in such cases we can keep the recipe for the version (`0.8.3`) which works fine on all arches, and add a parallel recipe (with the exact same basename) for the new version (`0.8.4`), marking that one as broken on x86 and x86 second arch.
(I'd say that this is the easiest scenario.)
But maybe we should have a look at the tests before merging, in case adding a few `cmd:*` entries in `TEST_REQUIRES` helps.

alaviss:
The latest commit should somehow magically fix gcc2, but broke x86_64 for me. Can someone double check this on x86_64 since my installation might be faulty

net-libs/libssh/libssh-0.8.4.recipe
```diff
@@ -57,43 +58,63 @@ BUILD_PREREQUIRES="
...
+
+	# Doesn't work in chroot, probably due to the lack of the passwd database
+	sed -i 's/cmocka_unit_test(torture_path_expand_tilde_unix),//' \
+		tests/unittests/torture_misc.c
 }
 
 BUILD()
 {
+	_maybe_pie=()
+	if [ "$targetArchitecture" != x86_gcc2 ]; then
```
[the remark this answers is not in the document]
alaviss: Check my 3 last commits :P

Bash is a much more horrible beast than I expected. I've tested this on gcc2 actually, and seems to work fine.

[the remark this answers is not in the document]
fbrosson: My previous suggestion to check `[ $effectiveTargetArchitecture" != x86_gcc2 ]` was wrong because the recipe does not build for x86_gcc2 primary arch.
So the right test is neither `[ "$targetArchitecture" != x86_gcc2 ]` nor `[ "$effectiveTargetArchitecture" != x86_gcc2 ]` but  `[ "$effectiveTargetArchitecture" = x86_64 ]`.
(OK, `[ "$targetArchitecture" != x86_gcc2 ]` works for x86_gcc2 hybrid and x86_64, but it would not work for x86 primary arch, so what we need is really `[ "$effectiveTargetArchitecture" = x86_64 ]`.)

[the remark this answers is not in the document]
fbrosson: `100% tests passed, 0 tests failed out of 20` on x86 second arch and x86_64.
I'll test it asap on x86 primary arch to see if we need this change:
```diff
  	_maybe_pie=()
-	if [ "$targetArchitecture" != x86_gcc2 ]; then
+	if [ "$effectiveTargetArchitecture" = x86_64 ]; then
 		_maybe_pie=("-DCMAKE_EXE_LINKER_FLAGS=-pie -fstack-protector")
 	fi
```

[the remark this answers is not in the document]
fbrosson: Hmm, I don't have gcc7 on my x86 primary arch, and libssh is failing to build with gcc 5.4.0.
Anyway, the above can't be wrong. I'll push it for you, then I'll let you squash and merge with your GPG signature :)

fbrosson: @alaviss, I tried to push this tiny change, but I got a `(permission denied)` (probably because you did not check "Allow edits from maintainers.") That's OK. I also always uncheck it on my PRs :)
However, could you please make the following change before merging?
```diff
  	_maybe_pie=()
-	if [ "$targetArchitecture" != x86_gcc2 ]; then
+	if [ "$effectiveTargetArchitecture" = x86_64 ]; then
 		_maybe_pie=("-DCMAKE_EXE_LINKER_FLAGS=-pie -fstack-protector")
 	fi
```
This would allow the recipe to be 100% correct for those users who might choose to build a x86 primary arch system.
Also, when we test the CPU for a particular arch to decide which compiler options to use we should always use "$effectiveTargetArchitecture".
Thanks!

net-libs/libssh/libssh-0.8.4.recipe
alaviss: Let's just omit the docs on gcc2 for now. I think some changes should be made to the Haiku infrastructure.

PR #3228, opened 2018-10-17
easyzma, new recipe

Begasus:


korli:
Is this recipe really necessary? It's fast inexistant: https://repology.org/metapackage/easylzma/versions
IMO it's not needed to build Calibre, but please prove me wrong.

Begasus:
Only checked added it as it was requested for Calibre, if not this could be closed @extrowerk can you check?

PR #3229, opened 2018-10-18
llvm6/7: enable float128 support in clang.

return:
This patch is required to build a dependency of swift-lang when built with LLVM 6 and higher, and enables support for float128 in the compiler. Tested on Haiku x86_64, untested on x86.

### DIFF
--- a/sys-devel/llvm/patches/clang-6.0.1.patchset
+++ b/sys-devel/llvm/patches/clang-6.0.1.patchset
@@ -1,4 +1,4 @@
-From 264717b79e4c1013603826f56fdad9a687b57e67 Mon Sep 17 00:00:00 2001
+From 2224160e6b72457a3b386f412831f48c21710bf3 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?J=C3=A9r=C3=B4me=20Duval?= <jerome.duval@gmail.com>
 Date: Mon, 18 Jul 2016 14:13:19 +0200
 Subject: support for secondary arch.
@@ -104,10 +104,10 @@ index 8c6face..4c04668 100644
      break;
    case llvm::Triple::RTEMS:
 -- 
-2.16.4
+2.19.1
 
 
-From 62248207b775044c44feaf1cce6a4e053898b225 Mon Sep 17 00:00:00 2001
+From 01c867dbd8530f5531532cba7d07f1e3e47f38c5 Mon Sep 17 00:00:00 2001
 From: Jerome Duval <jerome.duval@gmail.com>
 Date: Thu, 7 Apr 2016 18:30:52 +0000
 Subject: add a test for haiku driver.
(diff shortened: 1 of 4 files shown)


sys-devel/llvm/patches/clang-6.0.1.patchset
```diff
@@ -170,5 +170,37 @@ index 8b5b48e..ccd851f 100644
...
+
+diff --git a/lib/Basic/Targets/OSTargets.h b/lib/Basic/Targets/OSTargets.h
+index 790732b..3ff1fbc 100644
+--- a/lib/Basic/Targets/OSTargets.h
++++ b/lib/Basic/Targets/OSTargets.h
+@@ -251,6 +251,9 @@ protected:
+     Builder.defineMacro("__HAIKU__");
+     Builder.defineMacro("__ELF__");
+     DefineStd(Builder, "unix", Opts);
++    if (this->HasFloat128)
```
waddlesplash: Missing braces?

waddlesplash:
Might be worth upstreaming?

sys-devel/llvm/patches/clang-6.0.1.patchset
return: That was done intentionally to be consistent with LLVM's code style guidelines. A similar [patch](https://reviews.llvm.org/D41240#change-onZRG3.R04Cv) for many targets like OpenBSD, Solaris, etc leave out the braces in OSTargets.h. So I think this should be acceptable even for upstreaming.

return:
@waddlesplash As soon as I add the testcase for float128 support for Haiku into this PR, then I'll upstream it later.

sys-devel/llvm/patches/clang-6.0.1.patchset
waddlesplash: Except you have 2 lines indented as if they are part of the if, but only 1 is actually encapsulated by the if.

return:
@waddlesplash Tests for __float128 have now been added. Can this PR be merged in?

KapiX:
You still didn't address the braces thing. Why are two lines indented if only one has to be part of the if?

sys-devel/llvm/patches/clang-6.0.1.patchset
return: The patch has now been updated.

return:
@waddlesplash @KapiX I've included the braces back in, please check if these patches are ok to merge in or not.

korli:
Is float128 supposed to be working on x86?

sys-devel/llvm/patches/clang-6.0.1.patchset
```diff
@@ -170,5 +170,53 @@ index 8b5b48e..ccd851f 100644
...
++      Builder.defineMacro("_GLIBCXX_USE_FLOAT128");
++    }
+   }
+ 
+ public:
+@@ -260,6 +264,7 @@ public:
+     this->IntPtrType = TargetInfo::SignedLong;
+     this->PtrDiffType = TargetInfo::SignedLong;
+     this->ProcessIDType = TargetInfo::SignedLong;
++    this->HasFloat128 = true;
```
korli: I would use the same code as Solaris:
https://github.com/llvm-mirror/clang/commit/41ef6cd00386db07cc02e3b618873e87527170da#diff-ae1b30e40847dc1832f14faa7dc8d174R563

return: Patchset updated.

Issue #139, opened 2015-06-30
LLVM 3.5.0 unittest failures


soakbot:
(Original issue 139 created by kallisti5 on 2014-12-15T03:44:53.330452+00:00)

This bug is to track the unittest failures for llvm.

Instructions:
- Download LLVM 3.5.0
- export REQUIRES_RTTI=1
- ./configure --enable-optimized
- make
- cd tests
- make

soakbot:
(Original comment by kallisti5 on 2014-12-15T03:45:48.943739+00:00)

Haiku shredder 1 hrev48500 Dec 14 2014 10:13:47 x86_64 x86_64 Haiku

42 unexpected failures.

```
#!text

/Builds/llvm-3.5.0.src/test> make
Making LLVM 'lit.site.cfg' file...
Making LLVM unittest 'lit.site.cfg' file...
( ulimit -t 600 ; ulimit -d 512000 ; ulimit -m 512000 ; ulimit -s 8192 ; \
  /bin/python /Builds/llvm-3.5.0.src/utils/lit/lit.py -s -v . )
/bin/sh: line 0: ulimit: cpu time: cannot get limit: Invalid Argument
/bin/sh: line 0: ulimit: data seg size: cannot get limit: Invalid Argument
/bin/sh: line 0: ulimit: -m: invalid option
ulimit: usage: ulimit [-SHabcdefilmnpqrstuvxT] [limit]
/bin/sh: line 0: ulimit: stack size: cannot modify limit: Invalid Argument
FAIL: LLVM :: CodeGen/X86/fast-isel-args-fail2.ll (4826 of 11242)
******************** TEST 'LLVM :: CodeGen/X86/fast-isel-args-fail2.ll' FAILED ********************
Script:
--
not --crash /Builds/llvm-3.5.0.src/Release+Asserts/bin/llc < /Builds/llvm-3.5.0.src/test/CodeGen/X86/fast-isel-args-fail2.ll -fast-isel -fast-isel-abort-args -mtriple=x86_64-apple-darwin10
--
Exit Code: 1

Command Output (stderr):
--
FastISel didn't lower all arguments
UNREACHABLE executed at SelectionDAGISel.cpp:1103!
Abort
Stack dump:
0.	Program arguments: /Builds/llvm-3.5.0.src/Release+Asserts/bin/llc -fast-isel -fast-isel-abort-args -mtriple=x86_64-apple-darwin10 
[...]

PR #3234, opened 2018-10-20
LLVM: Remove older patches/recipes < LLVM v5.0.0

kenmays:
Per discussion of #139:
Removing all LLVM recipes/patches older than LLVM 5.0.0.


fbrosson:
@kenmays, may I update your PR to squash all its commits **except the last one,** _"Update LLVM license to current"_, in order to have only 2 commits, the first one for the removal of all the old recipes and patches, and the second one for the _"Update LLVM license to current"_, and then use the "Rebase and merge" method?
(Using "Squash and merge" here would **not** be a appropriate for this PR because the last commit would be mixed with all the file deletions.)

Is everyone OK with this PR? @korli ?
(I think @return had written a comment somewhere about the removal of the old recipes of llvm but I don't remember where it is.)

PR #3251, opened 2018-10-22
ripgrep: new recipe

Crestwave:


Begasus:
It's using rust as base to build? (cargo) don't think maby people already have that installed, should be checked :)

app-text/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+fi
+
+PROVIDES="
+	ripgrep$commandSuffix = $portVersion
+	cmd:rg
+	cmd:rg$commandSuffix
```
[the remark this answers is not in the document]
Begasus: Not sure how cargo builds work, and if it's possible to use $commandSuffix here, changed that to $secondaryArchSuffix and it's working ok (it seems)

app-text/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+	"
+
+BUILD()
+{
+	cargo build --release
+}
+
+INSTALL()
+{
+	install -Dm555 target/release/rg $binDir/rg
```
Begasus: I using $commandSuffix will work, you should use $prefix/bin here instead of $binDir, and then no ln -s would be required

app-text/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+fi
+
+PROVIDES="
+	ripgrep$commandSuffix = $portVersion
+	cmd:rg
+	cmd:rg$commandSuffix
+	"
+REQUIRES="
+	haiku$commandSuffix
+	gcc${commandSuffix}_syslibs
```
Begasus: No need for gcc requirement here

app-text/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${commandSuffix}_devel
+	lib:libxslt$commandSuffix
```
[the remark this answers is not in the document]
Begasus: See my comment about it :) https://github.com/haikuports/haikuports/pull/3251/files/54564cfd153f7062be4e7bae914534950392779d#r226991549

Crestwave:
> It's using rust as base to build? (cargo) don't think many people already have that installed, should be checked :)

Oh, I thought that `cmd:` would install the package providing the command if it wasn't available; there wasn't any explanation of it in the manual, although I suppose that I should've figured it out from the context.

Also, I failed to mention this earlier due to GitHub's issues, but unfortunately, it fails to read `stdin`. Here's the test result with the backtrace:

```failures:

---- multiline::stdin stdout ----
thread 'multiline::stdin' panicked at '

==========
command failed but expected success!

Did your search end up with no results?

command: "/boot/home/haikuports/app-text/ripgrep/ripgrep-0.10.0/target/debug/deps/../rg" "--path-separator" "/" "-n" "-U" "of this world\\p{Any}+?detective work"
cwd: /tmp/ripgrep-tests/stdin/138

status: exit code: 1

stdout: 

stderr: 

==========
', tests/util.rs:403:13
stack backtrace:
   0: _init
   1: _init
   2: _init
   3: _init
   4: _init
   5: _init
   6: _init
   7: _init
   8: _init
   9: _init
  10: _init
  11: _init
  12: _init
  13: _init
  14: _init
  15: _init
  16: _init
  17: _init
  18: _init
  19: _init
  20: pthread_exit
  21: <unknown>


failures:
    multiline::stdin

test result: FAILED. 186 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

Finally, am I supposed to specify/include Unlicense anywhere or is using Public Domain instead fine?

app-text/ripgrep/ripgrep-0.10.0.recipe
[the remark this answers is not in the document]
Crestwave: As for using `devel:` instead, it is only needed for `asciidoc`’s support for `xsltproc` (should it be added to `asciidoc`’s `REQUIRES`?), so its developmental files aren’t necessary.

Begasus:
From the COPYRIGHT file "Unlicense and MIT, also the "cargo test --all" ends up in KDL, so all arch's should be marked as broken "!" for now.

app-text/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
+SUMMARY="Combines the usability of ag with the raw speed of grep"
+DESCRIPTION="ripgrep is a line-oriented search tool that recursively searches \
+your current directory for a regex pattern while respecting your gitignore \
+rules. It is similar to other popular search tools like The Silver Searcher, \
+ack and grep."
+HOMEPAGE="https://github.com/BurntSushi/ripgrep"
+COPYRIGHT="2015 Andrew Gallant"
+LICENSE="MIT
+	Public Domain"
```
alaviss:
```suggestion
	Unlicense"
```

Our "Public Domain" license is the CC0 (Creative Commons Zero) license, and not the Unlicense. Please add a file named `Unlicense` containing said license text to the `licenses` folder relative to this recipe.

app-text/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	lib:libxslt$secondaryArchSuffix
+	rust_bin$secondaryArchSuffix
+	asciidoc
```
alaviss: Please modify this to be `cmd:<command used>` instead. For example:

```diff
-	rust_bin$secondaryArchSuffix
+	cmd:cargo$secondaryArchSuffix
```

And if they're `cmd`s, please move them to `BUILD_PREREQUIRES`

Crestwave:
Where should I place the fetching command?

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,64 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	lib:libxslt$secondaryArchSuffix
+	rust_bin$secondaryArchSuffix
+	asciidoc
```
Begasus: Suggestion:
```diff
-	rust_bin$secondaryArchSuffix
-	asciidoc
```
rust_bin will be fetched with the cmd:cargo$secondaryArchSuffix
put asciidoc also in BUILD_PREREQUIRES

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	lib:libxslt$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:cargo$secondaryArchSuffix
+	cmd:a2x
```
Begasus: Sort A->Z

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+fi
+
+PROVIDES="
+	ripgrep$secondaryArchSuffix = $portVersion
+	cmd:rg$commandSuffix
```
fbrosson: This recipe does not provide any library so we can get rid of the `commandSuffix` with these changes:
```diff
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+fi
 
-commandSuffix=$secondaryArchSuffix
-if [ "$targetArchitecture" = x86_gcc2 ]; then
-	commandSuffix=
-fi
 
 PROVIDES="
 	ripgrep$secondaryArchSuffix = $portVersion
-	cmd:rg$commandSuffix
+	cmd:rg
 	"
```
(If you copy/paste the if then fi block, make sure you get rid of the white space before `SECONDARY_ARCHITECTURES="x86"` :)

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+{
+	install -Dm555 target/release/rg $prefix/bin/rg
+
+	install -Dm444 target/release/build/ripgrep-*/out/rg.bash $dataDir/bash-completion/completions/rg
+	install -Dm444 complete/_rg $dataDir/zsh/site-functions/_rg
+
+	install -Dm444 target/release/build/ripgrep-*/out/rg.1 $manDir/man1/rg.1
+	for doc in *.md; do
+		install -Dm444 $doc $docDir/$doc
+	done
```
fbrosson: Suggestion:
```diff
-	for doc in *.md; do
-		install -Dm444 $doc $docDir/$doc
-	done
+	install -Dm444 -t $docDir *.md
```

alaviss:
> Where should I place the fetching command?

The required files must be have their respective `SOURCE_URI`s. AFAIK Gentoo has a way to do this, but I haven't investigated yet.

sys-apps/ripgrep/ripgrep-0.10.0.recipe
fbrosson: Good catch! The guide does need a small update... I'll fix it soon :)
What changed is that with the conditional secondary arch the commandSuffix var is now always empty, which means we can drop it.

fbrosson: Wiki updated :tada:
https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines/_compare/bbcb32a%5E...bbcb32a

nielx:
@Begasus The Rust way of doing thing is to do static linking to all dependencies, except for the system libraries. Depending on the dependencies of ripgrep it could result in a large binary. Also I don't know how much debug info is included per default.

Crestwave:
Debug symbols *are* included in the binary, but it can be stripped if it’s preferred.

@nielx Any idea what’s causing the error in the test above?

> > Where should I place the fetching command?
> 
> The required files must be have their respective `SOURCE_URI`s. AFAIK Gentoo has a way to do this, but I haven't investigated yet.

Gentoo has a function which converts a list of `crates` and their versions to their respective URLs, then the `ebuild` sets `SRC_URI` to the list of URLs. HaikuPorter doesn’t seem to support multiple `SOURCE_URI`‘s nor multiple URLs in one, though.

Crestwave:
> > Where should I place the fetching command?
> 
> The required files must be have their respective `SOURCE_URI`s. AFAIK Gentoo has a way to do this, but I haven't investigated yet.

Gentoo has a function which converts a list of `crates` and their versions to their respective URLs, then the `ebuild` sets `SRC_URI` to the list of URLs. HaikuPorter doesn’t seem to support multiple `SOURCE_URI`‘s nor multiple URLs in one, though.

fbrosson:
@Crestwave, in case you are/were looking for a method to provide more than one tarball, it is possible to do that.
Git, for instance, has 3 tarballs: one with the sources, and two others for the docs (manpages & htmldocs).
https://github.com/haikuports/haikuports/blob/50978de12f6c9444e3790dffbe7f86141fb1b394/dev-vcs/git/git-2.19.1.recipe#L18

nielx:
Re download comment @alaviss 

The cargo build system will download the packages. There are ways to manually download all dependencies, but an application like ripgrep easily depends on 30+ dependencies. I would consider this request untenable.

I would make the suggestion to bundle the package with a Cargo.lock file. By default cargo quite aggressively tries to fetch the most recent package versions (following the semver conventions). This will inevitably break the package, because not all developers are very careful in their version numbering. 

I don't know how it would be best to add the Cargo.lock file. If there is a way to inject files into the source dir from the haikuports tree, that would be preferred. Otherwise a patch could be supplied.

alaviss:
@Crestwave 

You could use [`defineDebugInfoPackage`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#definedebuginfopackage) to strip the debuginfo

@nielx 

I understand, but if we were to inject a `Cargo.lock`, we would still need a list of them, wouldn't we? And if we have a list of them, we could just pull the cargos before the building process starts.

Crestwave:
I’ve added all the dependencies’ `SOURCE_URI`’s but `cargo` uses crates for its cache, so it continues to download them as HaikuPorter unpacks them. Is there any way to prevent HaikuPorter from unpacking sources or have `cargo` use the source directories instead of crates?

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,320 @@
...
+	cmd:gcc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage ripgrep$secondaryArchSuffix \
+	$binDir/rg
+
+BUILD()
+{
+	mkdir -p ~/.cargo/registry/src/github.com-1ecc6299db9ec823
+	cp -r /sources-*/* ~/.cargo/registry/src/github.com-1ecc6299db9ec823
```
fbrosson: Suggestion:
```diff
+	for i in {2..87}; do
+		eval temp=\$sourceDir$i
-	cp -r /sources-*/* ~/.cargo/registry/src/github.com-1ecc6299db9ec823
+		cp -r "$temp"/* ~/.cargo/registry/src/github.com-1ecc6299db9ec823
+	done
```
Well, this will not fix the issue. It's just a variant that avoids using the hard-coded `/sources-*/`.
BTW, I did not test it.

alaviss:
@Crestwave 

Append `#noarchive` to the `SOURCE_URI`s. That should prevent HaikuPorter from unpacking them.

sys-apps/ripgrep/ripgrep-0.10.0.recipe
Crestwave: Works great; thanks!

Crestwave:
@alaviss 

Thanks; it works now. So should I mark the architectures as broken for it to be merged or will this just be left here until the issue with `stdin` is resolved?

alaviss:
Please mark as broken then open an issue for this problem. We could make it a GCI task

nielx:
I understand that the current recipe is a preferred solution when we emphasize the wish to have haikuporter download all the dependencies, however I find the solution (generate a list based on a script) a hack and I also find the recipe nearing unreadability. Is this principle worth the trade-offs? 

Anyway, this PR might not be a good place to have that discussion, but for the record.

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,323 @@
...
+
+SOURCE_URI_86="https://static.crates.io/crates/winapi-x86_64-pc-windows-gnu/winapi-x86_64-pc-windows-gnu-0.4.0.crate#noarchive"
+CHECKSUM_SHA256_86="712e227841d057c1ee1cd2fb22fa7e5a5461ae8e48fa2ca79ec42cfc1931183f"
+
+SOURCE_URI_87="https://static.crates.io/crates/wincolor/wincolor-1.0.1.crate#noarchive"
+CHECKSUM_SHA256_87="561ed901ae465d6185fa7864d63fbd5720d0ef718366c9a4dc83cf6170d7e9ba"
+
+ARCHITECTURES="!x86_gcc2 !x86 !x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+SECONDARY_ARCHITECTURES="x86"
```
alaviss: Please disable this also, thanks!

sys-apps/ripgrep/ripgrep-0.10.0.recipe
```diff
@@ -0,0 +1,323 @@
...
+SOURCE_URI_83="https://static.crates.io/crates/winapi/winapi-0.3.5.crate#noarchive"
+CHECKSUM_SHA256_83="773ef9dcc5f24b7d850d0ff101e542ff24c3b090a9768e03ff889fdef41f00fd"
+
+SOURCE_URI_84="https://static.crates.io/crates/winapi-i686-pc-windows-gnu/winapi-i686-pc-windows-gnu-0.4.0.crate#noarchive"
+CHECKSUM_SHA256_84="ac3b87c63620426dd9b991e5ce0329eff545bccbbb34f3be09ff6fb6ab51b7b6"
+
+SOURCE_URI_85="https://static.crates.io/crates/winapi-util/winapi-util-0.1.1.crate#noarchive"
+CHECKSUM_SHA256_85="afc5508759c5bf4285e61feb862b6083c8480aec864fa17a81fdec6f69b461ab"
+
+SOURCE_URI_86="https://static.crates.io/crates/winapi-x86_64-pc-windows-gnu/winapi-x86_64-pc-windows-gnu-0.4.0.crate#noarchive"
```
alaviss: Are these actually used/needed?

Crestwave: No; thanks for pointing it out! Seems that you can't list situational dependencies in `Cargo.lock`, hence them being automatically pulled in, but I've included a patch to remove them from it now.

nielx: Are you sure it builds like this? I have seen these dependencies being pulled in and compiled with cargo as well. Are you sure they won't be pulled in through other means if you don't download them in this phase?

PR #3253, opened 2018-10-22
OpenCOLLADA: bump

extrowerk:


extrowerk:
No breaking change, tested with dependant recipes.

PR #3255, opened 2018-10-22
MSGPACK: bump

extrowerk:


extrowerk:
Nothing uses it yet, so it is safe to bump.

PR #3269, opened 2018-10-25
New recipe for Freeciv 2.6.0

bach5000:
The recipe has been tested on x64 and works fine 😄

### DIFF
--- a/games-strategy/freeciv/freeciv-2.6.0.recipe
+++ b/games-strategy/freeciv/freeciv-2.6.0.recipe
@@ -58,6 +58,7 @@ REQUIRES="
 	lib:libsdl_mixer_1.2$secondaryArchSuffix
 	lib:libsdl_ttf_2.0$secondaryArchSuffix
 	lib:libssl$secondaryArchSuffix
+	lib:libsqlite3$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
 
@@ -76,6 +77,7 @@ BUILD_REQUIRES="
 	devel:libsdl_image$secondaryArchSuffix
 	devel:libsdl_mixer$secondaryArchSuffix
 	devel:libsdl_ttf$secondaryArchSuffix
+	devel:libsqlite3$secondaryArchSuffix
 	devel:libz$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
@@ -92,8 +94,9 @@ BUILD()
 {
 	autoreconf -vfi
 	CFLAGS="-O2 -pipe" runConfigure --omit-dirs binDir ./configure \
-		--bindir=$commandBinDir --enable-client=sdl --enable-fcmp=no \
+		--bindir=$commandBinDir --enable-client=sdl --enable-fcmp=sqlite3 \
 		--enable-sys-lua --enable-shared --disable-static
+	false
 	make $jobArgs
 }
 
--- a/games-strategy/freeciv/freeciv-2.6.0.recipe
+++ b/games-strategy/freeciv/freeciv-2.6.0.recipe
@@ -14,6 +14,7 @@ LICENSE="GNU GPL v2"
 REVISION="1"
 SOURCE_URI="https://netix.dl.sourceforge.net/project/freeciv/Freeciv%202.6/$portVersion/freeciv-$portVersion.tar.bz2"
 CHECKSUM_SHA256="7c20399198d6c7d846fed9a69b02e01134ae5340a3ae0f99d1e38063ade6c999"
+PATCHES="freeciv-$portVersion.patchset"
 ADDITIONAL_FILES="freeciv.rdef.in"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
--- a/games-strategy/freeciv/freeciv-2.6.0.recipe
+++ b/games-strategy/freeciv/freeciv-2.6.0.recipe
@@ -92,7 +92,7 @@ BUILD()
 {
 	autoreconf -vfi
 	CFLAGS="-O2 -pipe" runConfigure --omit-dirs binDir ./configure \
-		--bindir=$commandBinDir --enable-client=sdl --enable-fcmp=sqlite3 \
+		--bindir=$commandBinDir --enable-client=sdl --enable-fcmp=no \
 		--enable-sys-lua --enable-shared --disable-static
 	make $jobArgs
 }
--- a/games-strategy/freeciv/freeciv-2.6.0.recipe
+++ b/games-strategy/freeciv/freeciv-2.6.0.recipe
@@ -1,4 +1,3 @@
-
 SUMMARY="A free Civilization-like game"
 DESCRIPTION="Freeciv is a free Civilization-like game, primarily \
 for X under Unix. It has support for multiplayer games locally or \
@@ -123,6 +122,3 @@ INSTALL()
 	rmdir $prefix/share
 	rm -f $libDir/*.a $libDir/*.la
 }
-
-
-
--- a/games-strategy/freeciv/freeciv-2.6.0.recipe
+++ b/games-strategy/freeciv/freeciv-2.6.0.recipe
@@ -58,7 +58,6 @@ REQUIRES="
 	lib:libsdl_mixer_1.2$secondaryArchSuffix
 	lib:libsdl_ttf_2.0$secondaryArchSuffix
 	lib:libssl$secondaryArchSuffix
-	lib:libsqlite3$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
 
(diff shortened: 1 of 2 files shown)


games-strategy/freeciv/freeciv-2.6.0.recipe
```diff
@@ -0,0 +1,128 @@
...
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	autoreconf -vfi
+	CFLAGS="-O2 -pipe" runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir --enable-client=sdl --enable-fcmp=no \
+		--disable-sys-lua --enable-shared --disable-static
```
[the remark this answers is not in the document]
bach5000: Oops! Disabled lua but forgot to remove liblua. Will fix right away :)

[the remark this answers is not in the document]
extrowerk: But the earlier recipes sports lua support? What is your reason for disabling it?

extrowerk:
The earlier recipes have ppatchsets too. Are sou sure they arent required anymore?

games-strategy/freeciv/freeciv-2.6.0.recipe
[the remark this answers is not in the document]
bach5000: Freeciv 2.6 seems to have its own little lua5.3 in its source, so external lua doesn't seem to be needed.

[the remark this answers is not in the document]
extrowerk: It would be much better to use the system provided lua, and AFAIK HaikuPorts guideline prefers the usage of shared system provided libs, instead on the in-tree versions. Please, try to use it if possible.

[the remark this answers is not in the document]
bach5000: The patch from the previous version might have a fix. I might have to alter the patchset to get it to work with the new version (the build crashes whenever I put the whole patchset in, but I think some parts of the patchset fixes the lua problem). I'll experiment and see if I could get it to run, with lua enabled, and with a patchset 😄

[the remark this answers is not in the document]
bach5000: Strange. @extrowerk I can't get freeciv to work with lua from the system. It keeps telling me `configure: error: Use of included lua disabled, and no lua found from system` even though I made sure Lua 5.3 and its libraries were installed.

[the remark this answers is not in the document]
extrowerk: @bach5000 : could this be useful? https://github.com/haikuports/haikuports/blob/master/dev-lang/lua/lua-5.3.4.recipe

Begasus:
You can enable-sys-lua, add next lines in REQUIRES and BUILD_REQUIRES
´´´diff
+	lib:liblua$secondaryArchSuffix
+	devel:liblua$secondaryArchSuffix >= 5.3
```
Looking into config.log after configure is done I also see it checks for libxml2

File: games-strategy/freeciv/freeciv-2.6.0.recipe [hunk 2]
```diff
@@ -0,0 +1,123 @@
...
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	autoreconf -vfi
```
Begasus: Suggestion:
```diff
-	autoreconf -vfi
```
Seems it's not needed, with that you could enable sqlite3 and MagickWand (with a small patch in the configure file) :)

games-strategy/freeciv/freeciv-2.6.0.recipe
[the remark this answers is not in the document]
bach5000: @extrowerk 
Never mind then :) Didn't see it on pkgman.

games-strategy/freeciv/freeciv-2.6.0.recipe
bach5000: @Begasus @extrowerk I have Lua working :)

bach5000: @Begasus I think autoreconf is useful for getting missing dependencies. Without it, I kept getting `aclocal-1.14 is missing from your computer`.

Also, on the recipe, do I add ImageMagick as `lib:libimagemagick` and `devel:libimagemagick`, and sqlite3 as `lib:libsqlite` and `devel:libsqlite`?

bach5000: @Begasus:
We could use
* `--enable-fcdb=no/all/mysql/postgres/sqlite3`
* `--with-sqlite3-prefix=PFX`: Prefix where SQLite3 is installed
* `--with-magickwand=DIR`: Imagemagick installation directory.

I think we can use these instead of adding libs and devel. What do you think?

Begasus: first one is needed to set sqlite3 :) pkg-config makes sure the packages (lib + devel) are found, there is just a check for the header in configure were the path isn't correct as used by imagemagic

Begasus: My bad, configure is not included in the source, so you do need autoreconf :)

bach5000: I think I'm lost :o 
I did set `--enable-fcdb` to `sqlite3`, but I have no idea what to do from there.
I looked in `configure` but no success :(

Begasus: --enable-fcmp=sqlite3 is right, just need to add it to the requirements then too (lib + devel), at the end of configure you'll get a summarize where it shows you it is enabled

Begasus: Looks like there is some problem with libsqlite3 and the beta, will proceed without it

bach5000:
@Begasus The game refuses to Start a New Game, and the error message is `Failed to obtain the required acess level to take control of the server`.
It might be because of one of the patches. I'll experiment and see what works.

games-strategy/freeciv/freeciv-2.6.0.recipe
```diff
@@ -0,0 +1,125 @@
...
+commandBinDir=$binDir
+if [ -z "${ARCHITECTURES##*!$targetArchitecture*}" ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+GLOBAL_WRITABLE_FILES="
+	settings/freeciv/database.lua keep-old
+	"
+
```
alaviss: Please add the `USER_SETTINGS_FILES` variable containing files/directories that would be used by the program in the user settings folder. The format is simple:

```sh
USER_SETTINGS_FILES="
	settings/<directory> directory
	settings/<file>
"
```

If the program stores its data in `B_USER_SETTINGS_DIRECTORY/freeciv` (which _currently_ is `~/config/settings/freeciv`), then the corresponding line should be:

```
	settings/freeciv directory
```

Same goes for files, for example, if the program read its configuration from `~/config/settings/freeciv.conf`, then the corresponding entry should be:

```
	settings/freeciv.conf
```

If, the file has a template that's installed with it in `dataDir/freeciv/freeciv.example.cfg`, then the line should be:

```
	settings/freeciv.conf template $dataDir/freeciv/freeciv.example.cfg
```

games-strategy/freeciv/freeciv-2.6.0.recipe
```diff
@@ -0,0 +1,125 @@
...
+		-e "s|@MINOR@|$MINOR|" \
+		-e "s|@INTERNAL@|$INTERNAL|" \
+		$portDir/additional-files/freeciv.rdef.in > freeciv.rdef
+
+	addResourcesToBinaries freeciv.rdef \
+		$appsDir/Freeciv
+
+	addAppDeskbarSymlink $appsDir/Freeciv
+	mv $prefix/share/* $dataRootDir
+	rmdir $prefix/share
```
alaviss:
```diff
-	mv $prefix/share/* $dataRootDir
-	rmdir $prefix/share
+	mv -T $prefix/share $dataRootDir
```

games-strategy/freeciv/freeciv-2.6.0.recipe
bach5000: @alaviss @Begasus There currently is a problem with freeciv saving its prefs to `~/.freeciv` instead of `~/config/settings/freeciv`. Additional patches to the source code might be needed...

games-strategy/freeciv/freeciv-2.6.0.recipe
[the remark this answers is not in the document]
pulkomandy: Does freeciv actually use Lua 5.3? The C API for 5.1, 5.2 and 5.3 is different and some projects still require an older version.

korli:
merged manually in abb57971f20f9381e1ff0ba314a76d73d0ce8d43

PR #3275, opened 2018-10-26
OpenDUNE: sdl game, playable but with some issues

theclue:
Howdi,

after an hard work and tnx to the __huge__ support of our portmasters on #haiku i've found my way into haikuports

The game is playable but suffering some issues and I'm afraid I need a root help here:

- **No sounds and music** - the app could use some optional packages for ex. for mt32 emulation. None of them are available on haikuports. But they are supposed to be optional. Nonetheless, the game is totally mute: only few cracks and pops coming from the speakers (to stop them, closing the app is not enough; media_server must be restarted)
- **Keyboard Input is messed up** - found while trying to give a name to a savename. Keys input are totally scrambled. I don't think is a bug of the game, I bet it's something related to SDL.
- **Works only in x86_64** - Game requires gcc4, I'm going to search my way into secondary architecture building after the aforementioned issues are gone.

ad maiora!


Begasus:
Seems you have a trailing whitespace there :)

theclue:
> Seems you have a trailing whitespace there :)

yeah, it would be nice if the linter tell you even _where_ the stupid trailing space is 😄

waddlesplash:
How does SDL read the keyboard? It's possible it uses raw scancodes which bypass the system keymap entirely, which would explain that.

theclue:
> How does SDL read the keyboard? It's possible it uses raw scancodes which bypass the system keymap entirely, which would explain that.

This was the first idea that came in mind to me too. But it doesn't seem a keymap/locale issue, actually. These are _complitely_ messed: for ex. when one press 'n', '9' is sent...

theclue:
I've tested the build with SDL2 and the keyboard issue is gone _but_ the game is terribly slow, so I revert back to SDL for now.

pulkomandy:
About SDL and keymaps: SDL provides both a raw keycode and a character representation in key events. On all other systems, the raw keycode is usually the PS/2 keycode. But on Haiku, they are the Be key number.

This is not really an error on our side: the raw keycodes are not supposed to be used by the app to build their own keymap, and they are platform specific. It just happens common platforms all use the same ones, and some apps wrongly assume they can use it to build a keymap. Other affected apps are dosbox and dcmo5, for example.

Ideally the app would be fixed to use the character representation when doing text input, and the raw keycode only when they need to match with a physical position on the keyboard, by matching it with one of the SDLK_* constants. Assuming the keycodes are in a known order and ignoring the SDLK_* constants is a misuse of the SDL API, unfortunately only visible in the Haiku port.

scottmc:
Maybe report this issue upstream so they can fix their code.  Or better yet, fix their code and send them the patch.

Begasus:
Conflicting int types atm, should patch "includes/types.h" for this
PS, is there any separate download for the data?

Begasus:
Merged in https://github.com/haikuports/haikuports/pull/5385

PR #3277, opened 2018-10-27
aspell: build for gcc7, filters fix

gl00my:
This fixes compilation of aspell on current Haiku.

Also, the 	--enable-compile-in-filters was added. This fixes error with default ngroff filter.

See: http://aspell.net/man-html/Loadable-Filter-Notes.html

Now, flyspell-mode is working in emacs.

### DIFF
--- a/app-text/aspell/aspell-0.60.6.1.recipe
+++ b/app-text/aspell/aspell-0.60.6.1.recipe
@@ -9,11 +9,12 @@ dictionary. Aspell will also do its best to respect the current locale \
 setting. Other advantages over Ispell include support for using multiple \
 dictionaries at once and intelligently handling personal dictionaries when \
 more than one Aspell process is open at once."
-HOMEPAGE="http://aspell.net"
+HOMEPAGE="http://aspell.net/"
 COPYRIGHT="2000-2006 Kevin Atkinson"
 LICENSE="GNU LGPL v2"
 REVISION="4"
-SOURCE_URI="https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
+SOURCE_URI="https://ftpmirror.gnu.org/aspell/aspell-$portVersion.tar.gz
+	https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
 CHECKSUM_SHA256="f52583a83a63633701c5f71db3dc40aab87b7f76b29723aeb27941eff42df6e1"
 PATCHES="aspell-$portVersion.patchset"
 
--- a/app-text/aspell/aspell-0.60.6.1.recipe
+++ b/app-text/aspell/aspell-0.60.6.1.recipe
@@ -13,7 +13,7 @@ HOMEPAGE="http://aspell.net"
 COPYRIGHT="2000-2006 Kevin Atkinson"
 LICENSE="GNU LGPL v2"
 REVISION="4"
-SOURCE_URI="ftp://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
+SOURCE_URI="https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
 CHECKSUM_SHA256="f52583a83a63633701c5f71db3dc40aab87b7f76b29723aeb27941eff42df6e1"
 PATCHES="aspell-$portVersion.patchset"
 
--- a/app-text/aspell/aspell-0.60.6.1.recipe
+++ b/app-text/aspell/aspell-0.60.6.1.recipe
@@ -12,9 +12,10 @@ more than one Aspell process is open at once."
 HOMEPAGE="http://aspell.net"
 COPYRIGHT="2000-2006 Kevin Atkinson"
 LICENSE="GNU LGPL v2"
-REVISION="3"
-SOURCE_URI="ftp://ftp.gnu.org/gnu/aspell/aspell-0.60.6.1.tar.gz"
+REVISION="4"
+SOURCE_URI="ftp://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
 CHECKSUM_SHA256="f52583a83a63633701c5f71db3dc40aab87b7f76b29723aeb27941eff42df6e1"
+PATCHES="aspell-$portVersion.patchset"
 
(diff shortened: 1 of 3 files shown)


app-text/aspell/aspell-0.60.6.1.recipe
```diff
@@ -12,9 +12,10 @@ more than one Aspell process is open at once."
 HOMEPAGE="http://aspell.net"
 COPYRIGHT="2000-2006 Kevin Atkinson"
 LICENSE="GNU LGPL v2"
-REVISION="3"
-SOURCE_URI="ftp://ftp.gnu.org/gnu/aspell/aspell-0.60.6.1.tar.gz"
+REVISION="4"
+SOURCE_URI="ftp://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
```
extrowerk: Can we switch to github?
https://github.com/GNUAspell/aspell/archive/rel-0.60.6.1.tar.gz

gl00my: For what? 
Official page is here: http://aspell.net/
The link on official page leads to gnu.org.
So, i do not see any reason this is the case to NOT merge this pr.

gl00my: sry, found link on github on official page :)

gl00my: Ok, we CAN. But github page  has no packet signatures. The tarballs are generated from git tree (binded to tags). So, those tarballs not the same on the gnu ftp...
So, in my opinion, we should stay on ftp.gnu.org...

Anyway, technically, we can do that. So, what shall i do?

gl00my:
So, if the problem is that it is not good to use ftp, i have changed it to https.
Is it fine?

Begasus:
You probably could wrap up the autotools with one line (instead of using libtoolize, aclocal, autoconf ...) ->
autoreconf -vfi

app-text/aspell/aspell-0.60.6.1.recipe
```diff
@@ -12,9 +12,10 @@ more than one Aspell process is open at once."
 HOMEPAGE="http://aspell.net"
```
fbrosson: Could you please add a trailing slash for us at the end of [`HOMEPAGE`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#homepage)?
```diff
-HOMEPAGE="http://aspell.net"
+HOMEPAGE="http://aspell.net/"
```
Thanks!

app-text/aspell/aspell-0.60.6.1.recipe
```diff
@@ -12,9 +12,10 @@ more than one Aspell process is open at once."
...
 COPYRIGHT="2000-2006 Kevin Atkinson"
 LICENSE="GNU LGPL v2"
-REVISION="3"
-SOURCE_URI="ftp://ftp.gnu.org/gnu/aspell/aspell-0.60.6.1.tar.gz"
+REVISION="4"
+SOURCE_URI="https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
```
fbrosson: It would be nice to use the mirrors:
```diff
-SOURCE_URI="https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
+SOURCE_URI="https://ftpmirror.gnu.org/aspell/aspell-$portVersion.tar.gz
+	https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
```

fbrosson: What kind of errors did you get? If your recipe was correct then what you can get is a warning saying that the URI from where it got the tarball is different from what is in the recipe. That happens if you change the `SOURCE_URI` after having downloaded the tarball.
What you probably have right now in `download/aspell-0.60.6.1.tar.gz.uri` is the URI you had when haikuporter downloaded the archive the first time you kicked `hp aspell`.
To get rid of the warning you can delete `download/aspell-0.60.6.1.tar.gz.uri` but this will force haikuporter to download the tarball again.
Another option is to edit that file, `download/aspell-0.60.6.1.tar.gz.uri` in order to replace the `ftp.gnu.org/gnu/` by `ftpmirror.gnu.org/`. Then make sure you have the new two-line `SOURCE_URI` and run `hp -n aspell`. It should not complain.
(Believe me, we have many recipes with two-line `SOURCE_URI`. I can think of `sys-apps/sed/`, `sys-devel/patch/`, for example.)

But maybe you got a real error. In this case this might have been that you forgot to remove the quotes at the end of the first line of the `SOURCE_URI` when you added a second line.

gl00my: You say me to add in SOURCE_URI TWO urls. They are both quoted in one string:

SOURCE_URI="first second"

Haikuporter thinks it is ONE uri first%20second

gl00my:
May be i added \ at eol. Will check it later...

app-text/aspell/aspell-0.60.6.1.recipe
fbrosson: I'm copying it here, without the "\`\`\`diff" ... "\`\`\`" block (to get rid of the `-` and `+`):
```
SOURCE_URI="https://ftpmirror.gnu.org/aspell/aspell-$portVersion.tar.gz
	https://ftp.gnu.org/gnu/aspell/aspell-$portVersion.tar.gz"
```
Is this what you tried?
(BTW, the first character in the second line is a tab, and there are no spaces.)

gl00my: Yes, but i might add \ at the end of the first line. Will recheck later.

fbrosson:
You PR is OK on all architectures.
Shall I update your PR with the cosmetic edit to address my comments about the `HOMEPAGE` and the `SOURCE_URI` ?

fbrosson:
Sorry, I should have waited for your green lights.
Anyway, the PR is still in good shape.
Shall I merge it?

gl00my:
> Sorry, I should have waited for your green lights.

It's my bad :) I did not notice gits warnings :)

> Anyway, the PR is still in good shape.
> Shall I merge it?

I think so.

PR #3278, opened 2018-10-27
ZSH: bump

extrowerk:


waddlesplash:
@extrowerk https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/2259/builds/31221.log

PR #3286, opened 2018-10-27
Nano: bump

extrowerk:


fbrosson:
@extrowerk, if you write "bump" in the commit message, then it should be a bump :smile:
Here, as well as for ZSH, HSTR, MKSH and probably other older PRs, you're not bumping but adding a recipe for a newer release.

IMHO it would be better to really do a bump, i.e. remove the previous release in the same PR, once you have checked that there are no regressions.
The git diffs for real bumps are much nicer than those for "add recipe for version x.y.z" commits.

Now, if you want to keep the old recipe to be able to easily disable the new one in case an unexpected regression is found after the merge, that's another story. But for most recipes I believe it is better to do real bumps.
BTW, if you just add recipes, who's going to do the cleaning? :smile:

app-editors/nano/nano-3.1.recipe
```diff
@@ -27,12 +27,19 @@ PROVIDES="
 REQUIRES="
 	haiku
 	cmd:groff
+	lib:libintl
+	lib:libmagic
```
fbrosson: Oops, I did the same mistake (adding `lib:libmagic`) a few months ago in 4fc82ec8f9616944234701a3b45b414dcaebdded :smile:
Like you, I thought it was a good idea but it was not, as @waddlesplash explained in https://github.com/haikuports/haikuports/pull/2697#issuecomment-397898548. (He fixed my mistake just after.)
(I guess it's still not a good idea, but something might have changed. Well, probably not.)

Regarding the removal of the old recipes, I just realized we had `2.9.8` and 2 older recipes, `2.7.5` and `2.8.7`.
I should have suggested that you could, maybe, only remove one of the older recipes.
But that's ok. If nano `3.1` is working fine then we can drop all the old recipes.

waddlesplash: >  Like you, I thought it was a good idea but it was not, as @waddlesplash explained in #2697 (comment). (He fixed my mistake just after.)

The code in nano which uses libmagic is about 5 lines long IIRC; someone could just take the time to write a patch that fetches the mimetype using Haiku's native APIs instead.

PR #3296, opened 2018-10-28
GPXSee: bump to 6.3

sikmir:


sikmir:
Rebased.

fbrosson:
* Thanks @sikmir for the rebase (against HEAD as we can see by following the link **_1&nbsp;parent 3cb5e840563807818c1c8151f13051d8c6d97366_** that is visible in f1fcdbe04684c03d8f9088e9552155398ff47d68) and for signing your commit with a nice GPG signature!
* Sorry for not having noticed the update before your PR got merged...
(I would have done a manual merge using git on the command line and your GPG signature would have been preserved.)

Don't worry, @korli. You were faster than me and although I think we should try and preserve GPG signatures (when contributors use them) I'm fine with traditional merges too :smile:

korli:
@fbrosson I'm pretty sure @sikmir doesn't care :)

fbrosson:
I think he did care, at least a little, otherwise he would not have squashed his commit, 33acadf4cc16acae0b63e1793674cba72bee7a60, and the one I pushed a bit later,  9b3041042efa76d9f64ac95f60ca51fd38283816.

But I agree that this is not very important :)
So that's OK.
For his new PRs, however, I suggest that we keep his GPG signature if the commit is based on HEAD.

sikmir:
> I think he did care, at least a little, otherwise he would not have squashed his commit

Actually, I squashed it just to keep PR (and git history) clean. It's just a common practice to squash "fixup" commits into a single one.
So that's OK, it is not very important here.

PR #3308, opened 2018-10-31
nose: new recipe

Crestwave:


dev-python/nose/nose-1.3.7.recipe
```diff
@@ -0,0 +1,109 @@
+SUMMARY="Extends unittest to make testing easier"
+DESCRIPTION='nose extends the test loading and running features of unittest, \
+making it easier to write, find and run tests.
+
+By default, nose will run tests in files or directories under the current \
+working directory whose names include "test" or "Test" at a word boundary \
+(like "test_this" or "functional_test" or "TestClass" but not \
```
fbrosson: A key contributor prefers to use double quotes, so I think we should do so :)
```diff
-DESCRIPTION='nose extends the test loading and running features of unittest, \
+DESCRIPTION="nose extends the test loading and running features of unittest, \
 making it easier to write, find and run tests.
 By default, nose will run tests in files or directories under the current \
 working directory whose names include "test" or "Test" at a word boundary \
-(like "test_this" or "functional_test" or "TestClass" but not \
+(like \"test_this\" or \"functional_test\" or \"TestClass\" but not \
-"libtest"). Test output is similar to that of unittest, but also includes \
+\"libtest\"). Test output is similar to that of unittest, but also includes \
 captured stdout output from failing tests, for easy print-style debugging.
 These features, and many more, are customizable through the use of plugins. \
 Plugins included with nose provide support for doctest, code coverage, and \
 profiling, flexible attribute-based test selection, output capture and more. \
 More information about writing plugins may be found on in the nose API \
 documentation, here:
-http://readthedocs.io/docs/nose/'
+http://readthedocs.io/docs/nose/"
```
BTW, since the above URL is already in `HOMEPAGE` maybe we can drop the last 3 lines:
```diff
-profiling, flexible attribute-based test selection, output capture and more. \
+profiling, flexible attribute-based test selection, output capture and more."
-More information about writing plugins may be found on in the nose API \
-documentation, here:
-http://readthedocs.io/docs/nose/'
```

dev-python/nose/nose-1.3.7.recipe
```diff
@@ -0,0 +1,109 @@
...
+
+TEST()
+{
+	for i in "${!PYTHON_PACKAGES[@]}"; do
+		pythonPackage=${PYTHON_PACKAGES[i]}
+
+		cd "$sourceDir"-$pythonPackage
+
+		# Ironically, the test seems to fail everywhere with python3
+		python setup.py test
```
fbrosson: This should make the tests work with python3 :)
```diff
-		# Ironically, the test seems to fail everywhere with python3
+		python=python$pythonVersion
-		python setup.py test
+		$python test_tzinfo.py
```

Crestwave: The file doesn't seem to exist.

fbrosson: My mistake. It was an incorrect copy/paste from dev-python/pytz/...

fbrosson: Sorry, I forgot to add a line to define `pythonVersion`:
```diff
 		pythonPackage=${PYTHON_PACKAGES[i]}
+		pythonVersion=${PYTHON_VERSIONS[$i]}
 
 		cd "$sourceDir"-$pythonPackage/pytz/tests
 
 		python=python$pythonVersion
 		$python test_tzinfo.py
```

Crestwave: That doesn't really change anything, though.

alaviss: @fbrosson suggestion does work for me though, however, some tests do fail.

fbrosson:
@Crestwave, I pushed a few minor changes:
* `SUMMARY` (if you prefer yours I can revert my change.)
* rework the main `for i in "${!PYTHON_PACKAGES[@]}"; do` loop
* use `$relativeManDir` (instead of `documentation/man`) to patch setup.py
* copy the docs only once
* also start the tests with python3

The resulting packages are the same as with your recipe, except that the pyc files in the python3 subpackages have different checksums.

Shall I merge your PR?

Crestwave:
@fbrosson Sure! I omitted the python3 tests at first as I thought that it would be misleading since the tests themselves are broken, but it's fine if you guys prefer including it. Thanks for the improvements!

PR #3309, opened 2018-10-31
yamagi_quake2: fix find path

Nikolay0054:
Yamagi Quake 2 searches for libraries in the current directory


Nikolay0054:
it's printing "Couldn't determine executable path. Using ./ instead."

Nikolay0054:
> Wouldn't this patch breaks -datadir relative/path/to/cwd?

yes but absolute path will work

alaviss:
> it's printing "Couldn't determine executable path. Using ./ instead."

Yep, looks like it's [not implemented][0] for Haiku. Since this can be trivially implemented using [`find_path()`](https://www.haiku-os.org/docs/api/FindDirectory_8h.html#ad7aea8c7f59a75196467df38f9c5635a), would you like to do it?

[0]: https://github.com/yquake2/yquake2/blob/d96a46dc1f832c5d4ab94dee871fb27d0ee6a385/src/backends/generic/misc.c#L130

Nikolay0054:
I'll try when i get home

games-fps/yamagi-quake2/patches/yamagi_quake2-7.30.patchset
```diff
@@ -86,3 +86,45 @@ index a2e6f18..9397d97 100644
...
+ // longer paths anyway.. this might not be the maximum allowed length, but is
+@@ -119,6 +123,15 @@ static void SetExecutablePath(char* exePath)
+ 	// TODO: realpath() ?
+ 	// TODO: no idea what this is if the executable is in an app bundle
+ 
++#elif defined(__HAIKU__)
++
++	size_t bufSize = PATH_MAX;
++	
++	if(find_path(B_APP_IMAGE_SYMBOL, B_FIND_PATH_IMAGE_PATH, NULL, exePath, bufSize) != B_OK)
```
alaviss: `PATH_MAX` should be put directly in place of `bufSize`, since `find_path()` takes a value directly.

PR #3324, opened 2018-11-02
Finally found some time to push a full source fpc recipe.

oco27:
Mostly done at last year Haiku coding sprint.

### DIFF
--- a/dev-lang/fpc/fpc-3.0.4.recipe
+++ b/dev-lang/fpc/fpc-3.0.4.recipe
@@ -1,0 +1,144 @@
+SUMMARY="Professional Pascal compiler"
+DESCRIPTION="Free Pascal is a 32, 64 and 16 bit professional Pascal compiler. \
+It can target many processor architectures: Intel x86 (including 8086), AMD64\
+/x86-64, PowerPC, PowerPC64, SPARC, ARM, AArch64, MIPS and the JVM. \
+Supported operating systems include Linux, FreeBSD, Haiku, macOS/iOS/\
+iPhoneSimulator/Darwin, DOS (16 and 32 bit), Win32, Win64, WinCE, OS/2, \
+MorphOS, Nintendo GBA, Nintendo DS, Nintendo Wii, Android, AIX and AROS. \
+Additionally, support for the Motorola 68k architecture is available in the \
+development versions."
+HOMEPAGE="https://freepascal.org/"
+COPYRIGHT="1993-2017 Florian Klaempfl and others"
+LICENSE="GNU LGPL v3
+	GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/freepascal/fpcbuild-$portVersion.tar.gz"
+CHECKSUM_SHA256="f66514e6f2c2e4e1bccccb4d554c24b77682ed61c87811ae5dd210f421855e76"
+SOURCE_URI_2="https://sourceforge.net/projects/freepascal/files/Documentation/$portVersion/doc-pdf.tar.gz#noarchive"
+CHECKSUM_SHA256_2="0c70ac388a5a6a1c775308661ce4feb9cfa9d9c1084fd6939bfa101d5488817b"
+SOURCE_DIR="fpcbuild-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
+DISABLE_SOURCE_PACKAGE="yes"
+
+PROVIDES="
+	fpc$secondaryArchSuffix = $portVersion
+	cmd:fpc$secondaryArchSuffix = $portVersion
+	cmd:ppc386$secondaryArchSuffix = $portVersion
+	cmd:bin2obj$secondaryArchSuffix = $portVersion
+	cmd:chmcmd$secondaryArchSuffix = $portVersion
+	cmd:chmls$secondaryArchSuffix = $portVersion
+	cmd:data2inc$secondaryArchSuffix = $portVersion
+	cmd:delp$secondaryArchSuffix = $portVersion
+	cmd:fd2pascal$secondaryArchSuffix = $portVersion
+	cmd:fp$secondaryArchSuffix = $portVersion
+	cmd:fpcjres$secondaryArchSuffix = $portVersion
+	cmd:fpclasschart$secondaryArchSuffix = $portVersion
+	cmd:fpcmake$secondaryArchSuffix = $portVersion
+	cmd:fpcmkcfg$secondaryArchSuffix = $portVersion
+	cmd:fpcres$secondaryArchSuffix = $portVersion
+	cmd:fpcsubst$secondaryArchSuffix = $portVersion
+	cmd:fpdoc$secondaryArchSuffix = $portVersion
+	cmd:fppkg$secondaryArchSuffix = $portVersion
+	cmd:fprcp$secondaryArchSuffix = $portVersion
+	cmd:h2pas$secondaryArchSuffix = $portVersion
+	cmd:h2paspp$secondaryArchSuffix = $portVersion
+	cmd:instantfpc$secondaryArchSuffix = $portVersion
+	cmd:makeskel$secondaryArchSuffix = $portVersion
+	cmd:mkarmins$secondaryArchSuffix = $portVersion
+	cmd:mkinsadd$secondaryArchSuffix = $portVersion
+	cmd:mkx86ins$secondaryArchSuffix = $portVersion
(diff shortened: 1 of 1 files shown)


dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,147 @@
+SUMMARY="Professional Pascal compiler"
+DESCRIPTION="Free Pascal is a 32, 64 and 16 bit professional Pascal compiler. \
+It can target many processor architectures: Intel x86 (including 8086), AMD64\
+/x86-64, PowerPC, PowerPC64, SPARC, ARM, AArch64, MIPS and the JVM. \
+Supported operating systems include Linux, FreeBSD, Haiku, Mac OS X/iOS/\
```
fbrosson: Suggestion:
```diff
-Supported operating systems include Linux, FreeBSD, Haiku, Mac OS X/iOS/\
+Supported operating systems include Linux, FreeBSD, Haiku, macOS/iOS/\
```

dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,147 @@
...
+development versions."
+HOMEPAGE="https://freepascal.org/"
+COPYRIGHT="1993-2015 Florian Klaempfl and others"
+LICENSE="
+	GNU LGPL v3
+	GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://sourceforge.net/projects/freepascal/files/Source/3.0.4/fpcbuild-3.0.4.tar.gz"
+CHECKSUM_SHA256="f66514e6f2c2e4e1bccccb4d554c24b77682ed61c87811ae5dd210f421855e76"
+SOURCE_URI_2="https://sourceforge.net/projects/freepascal/files/Documentation/3.0.4/doc-pdf.tar.gz#noarchive"
```
fbrosson: Suggestion:
```diff
-SOURCE_URI="https://sourceforge.net/projects/freepascal/files/Source/3.0.4/fpcbuild-3.0.4.tar.gz"
+SOURCE_URI="https://downloads.sourceforge.net/freepascal/fpcbuild-$portVersion.tar.gz"
 CHECKSUM_SHA256="f66514e6f2c2e4e1bccccb4d554c24b77682ed61c87811ae5dd210f421855e76"
-SOURCE_URI_2="https://sourceforge.net/projects/freepascal/files/Documentation/3.0.4/doc-pdf.tar.gz#noarchive"
+SOURCE_URI_2="https://sourceforge.net/projects/freepascal/files/Documentation/$portVersion/doc-pdf.tar.gz#noarchive"
```
(We can use the redirector service for the main tarball, but we can't do that for the second tarball, `doc-pdf.tar.gz`, because its name does not include a `3.0.4` so when a new version gets released we would no longer get the pdf for this one.)

dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,147 @@
...
+CHECKSUM_SHA256_2="0c70ac388a5a6a1c775308661ce4feb9cfa9d9c1084fd6939bfa101d5488817b"
+SOURCE_DIR="fpcbuild-$portVersion"
+DISABLE_SOURCE_PACKAGE="YES"
```
fbrosson:
```diff
 SOURCE_DIR="fpcbuild-$portVersion"
-DISABLE_SOURCE_PACKAGE="YES"
 
 ARCHITECTURES="x86_gcc2 x86"
 SECONDARY_ARCHITECTURES="x86"
+DISABLE_SOURCE_PACKAGE="YES"
```
(I need to check this.)

dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,147 @@
...
+	cmd:find
+	fpc_bin == 3.0.2
+	"
+
+
+BUILD()
+{
+	echo `date`
+	echo cp ${sourceDir2}/doc-pdf.tar.gz $sourceDir
+	cp ${sourceDir2}/doc-pdf.tar.gz $sourceDir
```
fbrosson: Suggestion:
```diff
-	echo `date`
-	echo cp ${sourceDir2}/doc-pdf.tar.gz $sourceDir
-	cp ${sourceDir2}/doc-pdf.tar.gz $sourceDir
+	cp $sourceDir2/doc-pdf.tar.gz $sourceDir
```

dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,145 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel >= r1~alpha4_pm_hrev51411
+	"
+BUILD_PREREQUIRES="
+	binutils$secondaryArchSuffix
+	cmd:tar
+	cmd:make
+	cmd:diff
+	cmd:find
+	fpc_bin == 3.0.2
```
alaviss: I think this should be `cmd:fpc >= 3.0.2` instead.

dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,145 @@
...
+/x86-64, PowerPC, PowerPC64, SPARC, ARM, AArch64, MIPS and the JVM. \
+Supported operating systems include Linux, FreeBSD, Haiku, macOS/iOS/\
+iPhoneSimulator/Darwin, DOS (16 and 32 bit), Win32, Win64, WinCE, OS/2, \
+MorphOS, Nintendo GBA, Nintendo DS, Nintendo Wii, Android, AIX and AROS. \
+Additionally, support for the Motorola 68k architecture is available in the \
+development versions."
+HOMEPAGE="https://freepascal.org/"
+COPYRIGHT="1993-2015 Florian Klaempfl and others"
+LICENSE="
+	GNU LGPL v3
```
fbrosson: Hmm, I saw this earlier but forgot to leave a comment about  it.
There are two possible fixes here:
* without alphabetical sorting:
```diff
-LICENSE="
-	GNU LGPL v3
+LICENSE="GNU LGPL v3
	GNU GPL v3"
```
* with alphabetical sorting:
```diff
-LICENSE="
-	GNU LGPL v3
+LICENSE="GNU GPL v3
-	GNU GPL v3"
+	GNU LGPL v3"
```
If you think most of the sources are covered by the LGPL then it's probably OK to keep `LICENSE` unsorted.
Otherwise I'd say that sorting would be better.
(If you copy/paste, don't forget to get rid of the white spaces, as we should only have tabs here.)

dev-lang/fpc/fpc-3.0.4.recipe
oco27: No, it should really be 3.0.2. Freepascal is only guaranteed to compile with the previous official version at the time of release (see the top of https://freepascal.org/develop.var).

dev-lang/fpc/fpc-3.0.4.recipe
oco27: Only the compiler is GNU GPL. Every thing else (especially what you can link in an application) is LGPL (Runtime library, packages). So, this order make sense, even if it was not done on purpose ;-)
More details here : https://freepascal.org/faq.var#general-license

fbrosson: Okay, so let's keep it with LGPL first.
Can you still apply the style change to obtain this?
```
LICENSE="GNU LGPL v3
	GNU GPL v3"
```
Thanks!

scottmc:
What's the status on this one?

oco27:
Well, waiting for some love...  
New review or commit if finally ok.

dev-lang/fpc/fpc-3.0.4.recipe
```diff
@@ -0,0 +1,144 @@
...
+LICENSE="GNU LGPL v3
+	GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/freepascal/fpcbuild-$portVersion.tar.gz"
+CHECKSUM_SHA256="f66514e6f2c2e4e1bccccb4d554c24b77682ed61c87811ae5dd210f421855e76"
+SOURCE_URI_2="https://sourceforge.net/projects/freepascal/files/Documentation/$portVersion/doc-pdf.tar.gz#noarchive"
+CHECKSUM_SHA256_2="0c70ac388a5a6a1c775308661ce4feb9cfa9d9c1084fd6939bfa101d5488817b"
+SOURCE_DIR="fpcbuild-$portVersion"
+
+ARCHITECTURES="x86_gcc2 x86"
```
Begasus: Haven't checked, but if it builds ok for x86_64 maybe add it here?

PR #3327, opened 2018-11-03
zeal, bump version

Begasus:


app-doc/zeal/zeal-0.6.1.recipe
```diff
@@ -8,7 +8,7 @@ COPYRIGHT="2013-2018 Oleg Shparber"
 LICENSE="GNU GPL v3"
 REVISION="1"
 SOURCE_URI="https://github.com/zealdocs/zeal/archive/v$portVersion.tar.gz"
-CHECKSUM_SHA256="ef307d3ad4f03c76fa5fc249f8e7e04de879aa7574bec1ff8be548dbc2c02973"
+CHECKSUM_SHA256="095c08f9903071849d5c79878abd48237ce1615f16d324afff1873ab6b5f0026"
 PATCHES="zeal-$portVersion.patchset"
```
fbrosson:
```diff
 SOURCE_URI="https://github.com/zealdocs/zeal/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="095c08f9903071849d5c79878abd48237ce1615f16d324afff1873ab6b5f0026"
+SOURCE_FILENAME="zeal-$portVersion.tar.gz"
 PATCHES="zeal-$portVersion.patchset"
```

fbrosson: One more suggestion:
```diff
+if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+fi
```

Begasus: This is only needed in case for binaries/applications correct?

fbrosson: For now it's only a nice to have. But when Haiku x86_64/x86 hybrid will be available it will be required, otherwise zeal_x86 and zeal would overlap on x86_64/x86 hybrid.

And yes, it is only OK to use this condition if the package does not provide any lib. Because if it did provide a lib then we would have to provide x86 secondary arch for both x86_gcc2 and x86_64.

PR #3328, opened 2018-11-03
xerces-c, bump version to 3.2.2

Begasus:


dev-libs/xerces_c/xerces_c-3.2.2.recipe
```diff
@@ -8,7 +8,7 @@ COPYRIGHT="1999-2018 The Apache Software Foundation"
 LICENSE="Apache v2"
 REVISION="1"
 SOURCE_URI="http://apache.mirrors.tds.net/xerces/c/3/sources/xerces-c-$portVersion.tar.bz2"
```
fbrosson: Maybe:
```diff
-SOURCE_URI="http://apache.mirrors.tds.net/xerces/c/3/sources/xerces-c-$portVersion.tar.bz2" 
+SOURCE_URI="https://www.apache.org/dyn/closer.cgi?path=xerces/c/3/sources/xerces-c-$portVersion.tar.bz2"
```
Note: I did not check this. I only found it here:
https://github.com/Homebrew/homebrew-core/blob/master/Formula/xerces-c.rb

fbrosson: The above suggestion might require defining a `SOURCE_FILENAME`.

Begasus: https://www-eu.apache.org/dist//xerces/c/3/sources/xerces-c-3.2.2.tar.bz2 could be used too I guess? :) (from the xerces-c website)

Begasus: Or https://www-us.apache.org/dist//xerces/c/3/sources/xerces-c-3.2.2.tar.bz2?

fbrosson: Sure, any mirror can be used, but if you choose a particular mirror then it will only benefit to the users near that mirror.
While with the `SOURCE_URI` I suggested every user will be dynamically redirected to the best possible mirror, depending on the GeoIP info from the user. (Well, this is what we can expect, if we notice the `closer.cgi?` in the URI.)

Begasus: Ok, will adjust :)

fbrosson: Oops, I just tried with curl the URI I suggested, but I'm obtaining ah HTML document...

Begasus: xerces-c

Begasus: Had the same thing :) So best leave as is?

fbrosson: I found https://reference.apache.org/pmc/mirror_scripts but this did not help.
So I guess there is not redirector service.
(So you can stick to the URI you have right now, or use any other specific mirror.)
In case I find something better, i.e. a "magic" URI with a working redirector service,  then I'll leave a comment.

Begasus: The previous URL has been here before (didn't change it) :) So for now I'll leave it as is :)

PR #3306, opened 2018-10-31
gn: new recipe

return:
This is the standalone version of the GN build tool developed by the Chromium authors which is used to build V8 and Chromium. There are some problems with versioning and obtaining a reliable build source however.

Firstly, GN doesn't have any particular versioning scheme, so instead I used the last commit made from a release version of V8 instead. Secondly, the checksums of tarballs downloaded from gn.googlesource.com/gn are not deterministic due to this [issue](https://github.com/google/gitiles/issues/84) with Gerrit, which makes it useless to verify. To solve this, the recipe is using my fork of GN with the commits integrated from upstream.

Tested on both x86 and x86_64.

Issue #2557, opened 2018-05-15
Update V8 recipe to current


extrowerk:
[V8](https://github.com/v8/v8) required for [QtWebEngine](https://github.com/haikuports/haikuports/issues/2512), and [NodeJS](https://github.com/haikuports/haikuports/issues/107), eventually for WebPositive.

A complete re-port required.

extrowerk:
@return : maybe interested in https://github.com/haikuports/haikuports/pull/1849

return:
@extrowerk PR created for V8: #3330

return:
V8 has been merged into HaikuPorts. This issue can now be closed.

extrowerk:
Thanks!

PR #3330, opened 2018-11-03
v8: add recipe for version 7.4.288.28

return:
This version of V8 is from Chrome 74.0.3729.157 as found in the version information in https://omahaproxy.appspot.com.

V8 requires several dependencies defined in its [DEPS](https://chromium.googlesource.com/v8/v8/+/8a28642518a30b588dda7c417fd820a9f2831fee) file which are cloned in before it builds and this recipe uses these git hashes to build the port.

I previously tried using 'depot_tools' to build it, but it doesn't work reliably for several reasons.
See: https://github.com/haikuports/haikuports/issues/2557#issuecomment-389212503

~Needs [GN](https://github.com/haikuports/haikuports/pull/3306) merged first to build correctly.~

EDIT: GN has been merged into HaikuPorts. It is now possible for the buildbots to build V8.

### DIFF
--- a/dev-lang/v8/v8-7.4.288.28.recipe
+++ b/dev-lang/v8/v8-7.4.288.28.recipe
@@ -1,0 +1,128 @@
+SUMMARY="Google's open source JavaScript engine"
+DESCRIPTION="V8 is Google's open source JavaScript engine. V8 is written in \
+C++ and is used in Google Chrome, the open source browser from Google. V8 \
+implements ECMAScript as specified in ECMA-262, 5th edition, and runs on \
+Windows (XP or newer), Mac OS X (10.5 or newer), and Linux systems that use \
+IA-32, x64, or ARM processors. V8 can run standalone, or can be embedded into \
+any C++ application."
+HOMEPAGE="https://v8.dev/"
+COPYRIGHT="2006-2018 The V8 Project Authors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/v8/v8/archive/$portVersion.tar.gz"
+SOURCE_DIR="v8-$portVersion"
+CHECKSUM_SHA256="a9b1c74ff2fae728e1c94acf525321669679ef56da9cc4d781c427b26032bbf7"
+srcChromeVersion="74.0.3729.157"
+SOURCE_URI_2="https://gsdview.appspot.com/chromium-browser-official/chromium-$srcChromeVersion-lite.tar.xz"
+CHECKSUM_SHA256_2="d7406d4efba9916000f8a0f087e7eaf443523f264a59a51d2089c4b2e4c66be0"
+SOURCE_DIR_2="chromium-$srcChromeVersion"
+
+PATCHES="v8-$portVersion.patchset"
+PATCHES_2="build-$portVersion.patch"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	SECONDARY_ARCHITECTURES="x86"
+	ARCH=x86
+else
+	ARCH=x64
+fi
+
+# On x86_gcc2 we don't want to install the commands in bin/<arch>/, but in bin/.
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+    commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	v8$secondaryArchSuffix = $portVersion
+	cmd:d8$commandSuffix = $portVersion
+	lib:libv8$secondaryArchSuffix = $portVersion
+	lib:libv8_for_testing$secondaryArchSuffix = $portVersion
+	lib:libv8_libbase$secondaryArchSuffix = $portVersion
+	lib:libv8_libplatform$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libexecinfo$secondaryArchSuffix
+	lib:libssp$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	v8${secondaryArchSuffix}_devel = $portVersion
+	devel:libv8$secondaryArchSuffix
+	"
+REQUIRES_devel="
+	v8$secondaryArchSuffix == $portVersion
+	"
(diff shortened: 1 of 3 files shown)


dev-lang/v8/v8-7.0.276.28.recipe
```diff
@@ -0,0 +1,139 @@
...
+	gcc${secondaryArchSuffix}_syslibs
+	gcc${secondaryArchSuffix}_syslibs_devel
+	llvm7$secondaryArchSuffix
+	llvm7${secondaryArchSuffix}_libs
+	devel:libgcc$secondaryArchSuffix
+	devel:libexecinfo$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:clang
+	cmd:clang++
```
waddlesplash: It really needs clang?

dev-lang/v8/v8-7.0.276.28.recipe
```diff
@@ -0,0 +1,139 @@
...
+	cp -af out.gn/$ARCH.release/*.so ${libDir}
+
+	prepareInstalledDevelLibs libv8
+	packageEntries devel $developDir
+}
+
+TEST()
+{
+	echo "Test"
+	# Disable this for now, KDLs on some tests
```
waddlesplash: Please report the KDLs and include reduced methods to reproduce them if possible. But at least report them.

dev-lang/v8/v8-7.0.276.28.recipe
return: @waddlesplash It's quite possible to build V8 using gcc, (AIX being the only one that does so). But I quickly ran into lots of nasty macro errors, incompatible flag errors, etc whilst using it. For now, I've had better success with using clang.

dev-lang/v8/v8-7.1.302.31.recipe
```diff
@@ -0,0 +1,138 @@
...
+any C++ application."
+HOMEPAGE="https://v8.dev/"
+COPYRIGHT="2006-2018 The V8 Project Authors"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/v8/v8/archive/$portVersion.tar.gz"
+SOURCE_DIR="v8-$portVersion"
+CHECKSUM_SHA256="45770770079b30c3229f6f58b1de4d6c953107aa89e72742a6cd224d2a93f1d0"
+srcGitRev2="211b3ed9d0481b4caddbee1322321b86a483ca1f"
+SOURCE_URI_2="git+https://chromium.googlesource.com/chromium/src/base/trace_event/common.git#$srcGitRev2"
```
waddlesplash: The buildbot is configured to not use `git+` sources. Any GitHub mirrors of these we could use?

If not, we might have to make archives ourselves and put them on the ports-mirror.

waddlesplash:
What else needs to be done here?

dev-lang/v8/v8-7.1.302.31.recipe
return: It should be possible to use a chromium mirror now that all the V8 dependencies have been integrated in the repository due to this commit [1].

However, the catch here is the size. The chromium mirrors on GitHub are 1GB in size and 4GB+ when extracted. This mirror [2] hosts a smaller chromium source size of 500 MB with the same extracted size and much higher download speed (Hosted on Google App Engine). So apart from the size of the source, this mirror can be used to replace the +git sources.   

[1] https://github.com/chromium/chromium/commit/e7001969a4527bc416ef354a78195b942160e79c
[2] https://gsdview.appspot.com/chromium-browser-official/

waddlesplash: It appears Gentoo does the same: https://data.gpo.zugaina.org/didactic-duck/dev-lang/v8/v8-3.25.9.ebuild

So that sounds fine.

return:
Apart from removing the git+ sources to use the tarballs or any other recipe style edits, nothing else needs to be done. I did do a build of V8 (7.4.138) using a canary chromium source (74.0.3701.0) on x86_64 and it builds without any issues.

Unless a stable version of the chromium source with the buildtools folder is available, we might have to temporarily archive the buildtools folder.

dev-lang/v8/v8-7.0.276.28.recipe
return: I haven't been able to solve the macro build issues when compiling with gcc 7. Instead I'm using clang to build V8 as a way around this for now.

dev-lang/v8/v8-7.0.276.28.recipe
return: I've re-enabled tests and omitted the ones that cause KDLs. I'll see if I can find a way to reproduce this KDL.

return:
As of Chromium 74.0.3729.61, it now includes the buildtools folder and the +git links in this recipe have been replaced with a link to the Chromium source instead.

Are there any other issues with this recipe?

waddlesplash:
Yes, it causes KDLs. If the tests do, it probably will also do so under normal operating conditions. That needs to be resolved, or at least reported, so that the cause is understood before this is merged.

dev-lang/v8/v8-7.4.288.28.recipe
```diff
@@ -0,0 +1,132 @@
...
+	gcc${secondaryArchSuffix}_syslibs
+	gcc${secondaryArchSuffix}_syslibs_devel
+	llvm7$secondaryArchSuffix
+	llvm7${secondaryArchSuffix}_libs
+	devel:libgcc$secondaryArchSuffix
+	devel:libexecinfo$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:clang
+	cmd:clang++
```
waddlesplash: It appears V8 can be built with GCC >= 6.3.0. Please remove all the Clang stuff and let's use that by default.

waddlesplash:
As discussed on IRC: The KDLs are apparently the same as https://dev.haiku-os.org/ticket/14017, and they have gone away with the new version. So this can probably be merged once it uses GCC instead of Clang.

dev-lang/v8/v8-7.4.288.28.recipe
return: V8 recipe now uses GCC instead of Clang.

PR #3331, opened 2018-11-03
OpenCV: enable contrib

extrowerk:


media-libs/opencv/patches/opencv-3.4.3-source2.patchset
```diff
@@ -0,0 +1,24 @@
...
+diff --git a/opencv_contrib-3.4.3/modules/sfm/src/libmv_light/libmv/numeric/numeric.cc b/opencv_contrib-3.4.3/modules/sfm/src/libmv_light/libmv/numeric/numeric.cc
+index 9007663..cd9bd86 100644
+--- a/opencv_contrib-3.4.3/modules/sfm/src/libmv_light/libmv/numeric/numeric.cc
++++ b/opencv_contrib-3.4.3/modules/sfm/src/libmv_light/libmv/numeric/numeric.cc
+@@ -21,6 +21,10 @@
+ 
+ #include "libmv/numeric/numeric.h"
+ 
++#ifdef __HAIKU__
++void		sincos(double, double *, double *);
```
alaviss: Do we even have an implementation for this function? Adding a prototype here may fix the build, but it'll render the module useless due of the missing symbol.

extrowerk: I think we dont have implementatin, you are right. Let me research it deeper.

extrowerk: I planned to tranplant it from openbsd, but i had to see that libroot already have sincos symbol, but no header file mentions it. What should we do now?

pulkomandy: Just add the declaration to headers/gnu/math.h, under a _GNU_SOURCE guard.

PR #3332, opened 2018-11-03
weechat, bump version to 2.3

Begasus:


fbrosson:
@Begasus, I pushed a tiny commit because I didn't want to bother you with a review.

But then I found a small issue: on x86 secondary arch we're shipping the `weechat.pc` in `lib/pkgconfig/` instead of `lib/x86/pkgconfig/`, and it is in the base package instead of the devel subpackage.

So we need to pass `$libDir` to cmake instead of letting it use the default `$prefix/lib`.
We also need to add some steps to have the pc file moved to `$developLibDir/pkgconfig/`.

fbrosson:
If nobody objects having `weechat.png` in `/system/data/icons/hicolor/32x32/apps/` then the PR is ready for a merge.
```bash
$ package list -p weechat-2.3-1-x86_64.hpkg | grep icon
data/icons
data/icons/hicolor
data/icons/hicolor/32x32
data/icons/hicolor/32x32/apps
data/icons/hicolor/32x32/apps/weechat.png
```

alaviss:
That icon should instead be converted and embedded into the executable.


-------- Original Message --------

fbrosson:
> That icon should instead be converted and embedded into the executable.

:+1:

But since other recipes (such as kde-base/marble and app-editors/retext) already ship icons in `/system/data/icons/hicolor/`, I'd say we can merge this PR now, with the icon, and have it embedded in the executable later.

@Begasus, what do you think? Shall we merge? Or would you like to address this issue before the merge?

alaviss:
The thing is, on Linux, those are used to display the icon, but on Haiku, they serve no purpose, at all. We should just remove them then open a new issue to convert the icon to an embeddable format.

pulkomandy:
On Sun, Nov 04, 2018 at 06:34:50AM -0800, fbrosson wrote:
> But since other recipes (such as kde-base/marble and app-editors/retext) already ship icons in `/system/data/icons/hicolor/`, I'd say we can merge this PR now, with the icon, and have it embedded in the executable later.

Yuck. Please remove these for application icons. Ideally we should
remove them for other icons too, but the app itself may want to display
its own icon. I would avoid putting them in a shared directory if
possible...

PR #3333, opened 2018-11-04
diskus: new recipe

Crestwave:


korli:
We don't package binaries sorry.

alaviss:
@korli What do you mean? This package builds from source.

korli:
@alaviss the recipe downloads binary crates to build a package. It could be anything untrusted, and we can't check the licensing terms.

Crestwave:
> @alaviss the recipe downloads binary crates to build a package. It could be anything untrusted, and we can't check the licensing terms.

The crates are archives that are unpacked and compiled during the build phase; there are no binaries that I am aware of.

korli:
Looks like a maintenance nightmare. I thought rustup would be the followed path.

How do we assert the licensing terms of the depending packages then?

Crestwave:
> How do we assert the licensing terms of the depending packages then?

On Linux, they usually just use the license(s) of the application itself; everything is licensed with MIT, anyway. Some are just dual licensed. If you want to include all the dependencies' licenses, the only other one used is Unlicense, so it won't be too much trouble to add. Adding their copyright notices would be more difficult, though.

sys-apps/diskus/diskus-0.3.1.recipe
```diff
@@ -0,0 +1,210 @@
...
+	cmd:cargo$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage diskus$secondaryArchSuffix \
+	$binDir/diskus
+
+BUILD()
+{
+	mkdir -p ~/.cargo/registry/cache/github.com-1ecc6299db9ec823
```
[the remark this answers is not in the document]
Crestwave: I honestly don't know exactly what it corresponds to (maybe @nielx knows?), but it's probably a commit, and is not specific to the application. If it somehow changes, cargo should automatically fetch the dependencies itself to the new directory; all you'll have to do is update the directory name to it.

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,190 @@
...
+	install -Dm444 -t $docDir README.md
+}
+
+TEST()
+{
+	mkdir -p ~/.cargo/registry/cache/github.com-1ecc6299db9ec823
+	for i in {2..43}; do
+		eval temp=\$sourceDir$i
+		cp "$temp"/* ~/.cargo/registry/cache/github.com-1ecc6299db9ec823
+	done
```
alaviss: Since you'd have to build first before haikuporter let you run `--test`, maybe this is not needed?

Crestwave: `/boot` is one of the temporary directories `haikuporter` makes for a proper environment, so it will be cleaned up/recreated before the testing.

alaviss: I believe `CARGO_HOME` can be set to elsewhere to prevent this.

Crestwave: Done; thanks!

sys-apps/diskus/diskus-0.3.1.recipe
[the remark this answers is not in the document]
nielx: The hash most likely corresponds with a git revision of the github.com repository.

nielx:
I really don't understand why we go through all these lengths to package rust binaries. The recipes are unreadable this way and the scalability for other tools (and future versions) is near nil. 

I understand that one of the fundamental points is that we want to have predictable builds. I understand that the most logical translation of that desire is to have all the sources listed and downloaded before build. But with this approach it is almost impossible to create portable, reusable recipes. Especially if we need to take manual steps to fix the dependency tree.

My proposal therefore is for packages that are build with cargo, to allow cargo to fetch the dependencies during the build phase. In order to create a repeatable build, it should be mandatory that there is a Cargo.lock file in there.

alaviss:
@nielx 
> My proposal therefore is for packages that are build with cargo, to allow cargo to fetch the dependencies during the build phase.

That's exactly what we want to avoid, see https://github.com/haikuports/haikuporter/issues/174. Predictable build is not an issue in this context, given the solution you showed us.

There are reasons why we don't want to download during build:

- The result can't be cached
- Environment variables are ignored, meaning that you can't use any kind of proxy

> The recipes are unreadable this way and the scalability for other tools (and future versions) is near nil.

IMO it's not, you can just skip through the `SOURCE_URI`s. They're still magnitude better than our python-based recipes.

What better would be to have a tool that can generate these kind of recipes automatically for us, which would easily solve the reusability problem.

@Crestwave I believe you have a tool for automated rust recipe generation, would you mind contributing it to [haikuporter/tools](https://github.com/haikuports/haikuporter/tree/master/tools)?

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,188 @@
...
+	cmd:cargo$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage diskus$secondaryArchSuffix \
+	$binDir/diskus
+
+BUILD()
+{
+	export CARGO_HOME=/.cargo
```
alaviss: How about this location?

```suggestion
	export CARGO_HOME=$sourceDir/../cargo
```

The `.` is unneeded since there aren't any good reason to hide the folder. The cargo folder is placed near `sourceDir` to support non-chrooted build (if that's ever fixed in the future).

sys-apps/diskus/diskus-0.3.1.recipe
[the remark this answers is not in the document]
Crestwave: It hasn't changed in quite a while, though, and it doesn't seem to match anything I can find. I was able to use a different directory with Gentoo's method, but it seems to need to find OS-specific dependencies, even when it knows not to use them.

So either we don't change it, clone the index ourselves (both of which take quite a while) and point it to it, try to find another way, or include the dependencies specific to other operating systems as sources, which should be the fastest unless theres a way to prevent it in the first place and is what Gentoo is doing.

Crestwave:
> @Crestwave I believe you have a tool for automated rust recipe generation, would you mind contributing it to [haikuporter/tools](https://github.com/haikuports/haikuporter/tree/master/tools)?

Well, it currently just outputs the crates.io dependencies, formatted properly, but that's probably better for the reusability problem, anyway. It actually reads the recipe to do so, but all it really needs is the `SOURCE_URI`, so it shouldn't be too hard to make it generate a recipe template if you want it to.

https://github.com/haikuports/haikuporter/pull/176

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,240 @@
...
+SOURCE_URI_31="https://static.crates.io/crates/rand_core/rand_core-0.3.0.crate"
+CHECKSUM_SHA256_31="0905b6b7079ec73b314d4c748701f6931eb79fd97c668caa3f1899b22b32c6db"
+
+SOURCE_URI_32="https://static.crates.io/crates/rayon/rayon-1.0.3.crate"
+CHECKSUM_SHA256_32="373814f27745b2686b350dd261bfd24576a6fb0e2c5919b3a2b6005f820b0473"
+
+SOURCE_URI_33="https://static.crates.io/crates/rayon-core/rayon-core-1.4.1.crate"
+CHECKSUM_SHA256_33="b055d1e92aba6877574d8fe604a63c8b5df60f60e5982bf7ccbb1338ea527356"
+
+SOURCE_URI_34="https://static.crates.io/crates/redox_syscall/redox_syscall-0.1.40.crate"
```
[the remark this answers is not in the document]
Crestwave: They're automatically fetched when in `Cargo.lock`, but they can be easily patched out of that (and they were, originally; I removed the patch in the latest commit) since it's only included in the main application, not the dependencies, and `Cargo.toml` has a way to specify them for their respected operating systems.

The thing is that cargo seems to need to *find* them when defined there, even when they're not pulled in; I discussed this [here](
https://github.com/haikuports/haikuports/pull/3333#discussion_r234051126). Patching it out of the `Cargo.toml` of those that require it is also another option, but is potentially tedious unless its scripted, too. Looking at it again, the format isn't as tricky as I thought it would be, so I'll see if it can be done. Hopefully there's a solution I'm missing, though, but it seems unlikely as Gentoo includes them as dependencies.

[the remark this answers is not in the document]
Crestwave: I've now scripted a good amount of the process and was able to patch them all out, although some manual actions were needed with the script's current state.

alaviss:
I was going to tell you that you don't have to patch them out if it's hard to do. But it's nice that you've taken care of it :smile:.

I thought you said only patching the `Cargo.lock` was needed? I saw patches to other cargos as well.

Crestwave:
> I thought you said only patching the `Cargo.lock` was needed? I saw patches to other cargos as well.

They're only automatically fetched when in `Cargo.lock`, but cargo seems to need to find them; it's usually does so through a registry update, but I've disabled that since it downloads during the build phase and takes *quite* a while, and before the patching, it was done by leaving the packages in the directory for it to find.

alaviss:
> > I thought you said only patching the `Cargo.lock` was needed? I saw patches to other cargos as well.
> 
> They're only automatically fetched when in `Cargo.lock`, but cargo seems to need to find them even when only specified in `Cargo.toml`; it's usually does so through a registry update, but I've disabled that since it downloads during the build phase and takes _quite_ a while, and before the patching, it was done by leaving the packages in the directory for it to find.

Then I believe we can just leave out the patching in the future, as it seems quite invasive. Hopefully those additional crates are light enough.

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,256 @@
...
+INSTALL()
+{
+	install -Dm555 -t $binDir target/release/diskus
+	install -Dm444 -t $docDir README.md
+}
+
+TEST()
+{
+	export CARGO_HOME=$sourceDir/../cargo
+	cargo test --all
```
alaviss: running 0 tests

maybe we could just remove `TEST()`

alaviss:
I've tested this on my haiku VM (x86_64)

`du -hs` shows 6.3G
`diskus` shows 5.68G

This is tested on my haikuports tree. So where has 0.62G gone?

alaviss:
`du -sh --apparent-size`: 4s real time, 5.3G
`diskus`: 3m7s real time, 5.68G

Still a mismatch here

Crestwave:
> `du -sh --apparent-size`: 4s real time, 5.3G
> `diskus`: 3m7s real time, 5.68G
> 
> Still a mismatch here

Try `-sbh`; `-b` is equivalent to `--apparent-size --block-size=1`.

alaviss:
`--block-size` is just a way to change the formatting... I tried `-BGB` for the same `GB` suffix, but got 7GB instead.

alaviss:
> > `du -sh --apparent-size`: 4s real time, 5.3G
> > `diskus`: 3m7s real time, 5.68G
> > Still a mismatch here
> 
> Try `-sbh`; `-b` is equivalent to `--apparent-size --block-size=1`.

Oops, my fault. Turns out `--block-size=1` does do something. It's all fine now.

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,256 @@
...
+SOURCE_DIR_43="version_check-0.1.5"
+
+SOURCE_URI_44="https://static.crates.io/crates/void/void-1.0.2.crate"
+CHECKSUM_SHA256_44="6a02e4885ed3bc0f2de90ea6dd45ebcbb66dacffe03547fadbb0eeae2770887d"
+SOURCE_DIR_44="void-1.0.2"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+SECONDARY_ARCHITECTURES="x86"
+fi
```
alaviss:
```diff
 ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
+commandBinDir=$binDir
 if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
+commandBinDir=$prefix/bin
 fi
```

To install without the `-x86` suffix on gcc2 secondary arch, you'd have to define `commandBinDir`.

@fbrosson, can you help with the aesthetic?

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,256 @@
...
+	[source.crates-io]
+	replace-with = "haiku"
+	EOF
+
+	cargo build --release
+}
+
+INSTALL()
+{
+	install -Dm555 -t $binDir target/release/diskus
```
alaviss:
```suggestion
	install -Dm555 -t $commandBinDir target/release/diskus
```

sys-apps/diskus/diskus-0.4.0.recipe
Crestwave: It does look pretty weird without the indent, but I applied it for now since @fbrosson hasn't responded yet.

sys-apps/diskus/diskus-0.4.0.recipe
```diff
@@ -0,0 +1,237 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cargo$secondaryArchSuffix
+	cmd:gcc$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage diskus$secondaryArchSuffix \
+	$binDir/diskus
```
alaviss:
```suggestion
	$commandBinDir/diskus
```

Crestwave: That's what I get for manually replacing them... thanks!

Crestwave:
>@Crestwave I found that diskus is terribly slow on Haiku, is it a bug? My test case was a populated haikuports tree on VBox.

I did various tests, and it seems to be caused the by the disk used; the author likely tested it directly against an SSD. I don't have one and thus wasn't able to verify it, but its slowness in my test cases were not specific to Haiku.

Also, this is off-topic, but since you mentioned VirtualBox: is there a way to use your host's screen resolution on a Haiku virtual machine? I run it on `qemu`, and the best I was able to get is much smaller than my screen, and unfortunately, I can't run it directly for daily use as Haiku does not support brightness control on NVIDIA yet, nor does it have one that works on the userland or a bluelight filter.

alaviss:
> I did various tests, and it seems to be caused the by the disk used; the author likely tested it directly against an SSD. I don't have one and thus wasn't able to verify it, but its slowness in my test cases were not specific to Haiku.

Probably, parallel access on HDD is pretty much insane IMO.

> Also, this is off-topic, but since you mentioned VirtualBox: is there a way to use your host's screen resolution on a Haiku virtual machine?

Sorry, I've never tried, as I'm perfectly happy with this small box on my screen :)

Anyhow, it's an excellent work that you've done here. Merging.

humdingerb:
> but since you mentioned VirtualBox: is there a way to use your host's screen resolution on a Haiku virtual machine?
@Crestwave: see https://www.haiku-os.org/guides/virtualizing/virtualbox#part_customVESA

Crestwave:
> but since you mentioned VirtualBox: is there a way to use your host's screen resolution on a Haiku virtual machine?
>> @Crestwave: see https://www.haiku-os.org/guides/virtualizing/virtualbox#part_customVESA

Worked perfectly; thanks! Too bad `qemu` doesn't support setting custom VESA resolutions without recompiling, though. You should add a black line between a quote and your comment, by the way; otherwise it thinks it's part of it.

> Sorry, I've never tried, as I'm perfectly happy with this small box on my screen :)

Well, it's true that you don't really need much for a terminal and such; it's mostly the immersion that bothers me. You should check out the link since you already use VirtualBox, though; you just run a command and set it in Screen.

PR #3359, opened 2018-11-10
ZSH: install fix

extrowerk:
Help generation requires cmd:col.
Interestingly the current util_linux package (2.33) not available in the depot, only an earlier version (2.32) which doesn't provides cmd:col. The 2.33 builds just fine, this i'm a bit puzzled.

### DIFF
--- a/app-shells/zsh/zsh-5.6.2.recipe
+++ b/app-shells/zsh/zsh-5.6.2.recipe
@@ -64,6 +64,7 @@ BUILD_REQUIRES="
 BUILD_PREREQUIRES="
 	cmd:autoreconf
 	cmd:awk
+	cmd:col
 	cmd:find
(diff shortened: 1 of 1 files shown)


fbrosson:
That's cool!
When I added `cmd:col` in 89d478886757309decd952460b4f5d15eb8093ec `util-linux: bump to 2.33, build cmd:col{,rm,umn}.` (#3345) I had not even thought it would be useful so quickly :)

I guess `2.33` is not yet available in the depot because I bumped util-linux only 2 days ago (Thursday, November 8) and it looks like the HaikuPorts repo hasn't been updated since Tuesday, November 6.

I guess you can still merge this PR.
(Don't forget to bump `REVISION`.)

extrowerk:
Should be working now.

PR #3369, opened 2018-11-13
VC: new recipe

extrowerk:
Useful for calligra and krita

### DIFF
--- a/dev-libs/vc/vc-1.4.0.recipe
+++ b/dev-libs/vc/vc-1.4.0.recipe
@@ -1,0 +1,64 @@
+SUMMARY="SIMD Vector Class Library for C++"
+DESCRIPTION="A library to ease explicit vectorization of C++ code."
+HOMEPAGE="https://github.com/VcDevel/Vc"
+COPYRIGHT="2006-2016 Arseny Kapoulkine"
+LICENSE="GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/VcDevel/Vc/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="449ce0a806ebb53cc99699ef223351ec770c76e819960b5849d162f9e8506d15"
+SOURCE_DIR="Vc-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="15.0.1"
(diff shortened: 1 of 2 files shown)


dev-libs/vc/vc-1.4.0.recipe
```diff
@@ -0,0 +1,64 @@
+SUMMARY="SIMD Vector Class Library for C++"
+DESCRIPTION="A library to ease explicit vectorization of C++ code."
+HOMEPAGE="https://github.com/VcDevel/Vc"
+COPYRIGHT="2006-2016 Arseny Kapoulkine"
+LICENSE="GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/VcDevel/Vc/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="449ce0a806ebb53cc99699ef223351ec770c76e819960b5849d162f9e8506d15"
+SOURCE_DIR="Vc-$portVersion"
```
alaviss:
```suggestion
SOURCE_FILENAME="vc-$portVersion.tar.gz"
SOURCE_DIR="Vc-$portVersion"
```

dev-libs/vc/vc-1.4.0.recipe
```diff
@@ -0,0 +1,64 @@
...
+	devel:libvc$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	"
+BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
```
alaviss: Should be moved to `BUILD_REQUIRES`

dev-libs/vc/vc-1.4.0.recipe
extrowerk: It provides only a static lib. Do i have to split it still into a devel package?

fbrosson: It depends. If, after moving to a devel subpackage everything that could be moved, you have nothing left in the base package then you can probably keep everything in the base package.
Otherwise a split desirable.

Would it make sense to install the base package only and leave the devel subpackage in `$PACKAGES_PATH` ? If yes then I'd say the split is desirable.

extrowerk: Nothing left in the base

Issue #3319, opened 2018-11-01
sys-devel/gcc: shared objects are always generated


alaviss:
`gcc` will always pass `-shared` to the linker, telling it to generate a shared object for all circumstances:

https://github.com/haikuports/haikuports/blob/0337d6953dee78b0a8890d4632311fe368c0b437/sys-devel/gcc/patches/gcc-7.3.0_2018_05_01.patchset#L840-L841

This was done for technical reasons within Haiku, namely:
- Required for replicants (as told on IRC)
- Haiku's native TLS (thread-local storage) support for ELF requires "executables" to be shared objects https://github.com/haiku/haiku/commit/44c0c4d3d0d68c3c3db23858b1089d3cf51bdf16

This, however, comes with its own downsides:
- *All* symbols in a binary is exported, which causes them to override any symbols from `dlopen()`-ed library and can lead to bugs like the one in [Ticket #14531][0]. On other OS, this must be requested explicitly via `-rdynamic` to the compiler or `--export-dynamic` to the linker.
- `-static` on Haiku is a lie, `-shared` have precedence over that flag, so building anything with `-static` will not result in a static binary, but rather a shared one (with no error/warning message). We currently don't support static binaries as our syscall numbers aren't locked, so this should've generated an error message instead.
- Might slow down program start up, since every binaries expose all of their symbols, which, in theory, could cause `runtime_loader` to spend more time resolving them (should be negligible).

Since PIE has pretty much [the same schematic][1] as a shared object, but doesn't export all symbols and does generate error if you try `-static`, I've looked in to it as a replacement for the current implicit `-shared`. However, this causes a problem: most (if not all) programs that make use of TLS will randomly segfault.

I've also tried combining `-pie` with `--export-dynamic` and `--no-runtime-linker` to (hopefully) emulate `-shared` (which should fix TLS), but that generates unusable binaries instead (`cannot relocate: Bad data`).

Any help on this is appreciated!

[0]: https://dev.haiku-os.org/ticket/14531
[1]: http://www.openbsd.org/papers/nycbsdcon08-pie/mgp00007.html

waddlesplash:
This really sounds like multiple bugs in runtime_loader, not in GCC, so why is this an issue here instead of at Haiku?

pulkomandy:
On Thu, Nov 01, 2018 at 06:51:58PM +0000, alaviss wrote:
[...]

PR #3376, opened 2018-11-16
sys-devel/gcc: add -Bsymbolic to link spec

alaviss:
This fixes an issue shown on IRC by @waddlesplash. This is highly experimental and completely untested.

Also, I've noticed that [gcc2's linkspec][0] is vastly different than that of gcc7, is there any reason in particular? I could try to sync up the linkspecs, but as mentioned in #3319, `-pie` by default will break TLS.

[0]: https://github.com/haiku/buildtools/blob/0c77ecec1fb2c5679db724046edea2c0a7bf62e2/legacy/gcc/gcc/config/i386/haiku.h#L188

### DIFF
--- a/sys-devel/gcc/gcc-7.3.0_2018_05_01.recipe
+++ b/sys-devel/gcc/gcc-7.3.0_2018_05_01.recipe
@@ -5,7 +5,7 @@ HOMEPAGE="https://gcc.gnu.org/"
 COPYRIGHT="1988-2018 Free Software Foundation, Inc."
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
-REVISION="4"
+REVISION="5"
 gccVersion="${portVersion%%_*}"
 SOURCE_URI="https://ftpmirror.gnu.org/gcc/gcc-$gccVersion/gcc-$gccVersion.tar.xz
 	https://ftp.gnu.org/gnu/gcc/gcc-$gccVersion/gcc-$gccVersion.tar.xz"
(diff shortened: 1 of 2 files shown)


pulkomandy:
Why do we have patches for gcc2 here? This should be upstreamed as people can also build the buildtools directly using ./configure when building Haiku, and a mismatch between the two would be a bad idea.

Ingo did some work on BSymbolic a few years ago but I don't remember what it was. Can you check the git logs of the buildtools repo to investigate why it was removed?

alaviss:
> Why do we have patches for gcc2 here?

What do you mean?

> Can you check the git logs of the buildtools repo to investigate why it was removed?

It was removed at https://github.com/haiku/buildtools/commit/8116a5d9ffdc704a2fe1095bd9bdac3a36ad56bb, reasons unknown.

pulkomandy:
I mean we have already forked gcc for Haiku (both versions 2 and 7, in fact) in our buildtools repo. Why are the patches not there instead of here?

Reasons not unknown, now I remember it. -Bsymbolic was interferring with weak symbols, causing problems with some GNU software (I don't remember exactly which one) which defined some C library functions as weak symbols, expecting that if the C library provided a non-weak implementation, it would override its own version.

alaviss:
> Reasons not unknown, now I remember it. -Bsymbolic was interferring with weak symbols, causing problems with some GNU software (I don't remember exactly which one) which defined some C library functions as weak symbols, expecting that if the C library provided a non-weak implementation, it would override its own version.

Then that would be a problem. I guess we'll have to look for solutions other than this PR. AFAIK OpenBSD do this kind of defining everything as weak, and not passing `-Bsymbolic` to the linker. I don't know if @waddlesplash bug could be reproduced on OpenBSD however.

PR #3382, opened 2018-11-16
cowsay: bump to 3.04

Crestwave:


games-misc/cowsay/cowsay-3.04.recipe
```diff
@@ -5,8 +5,8 @@ HOMEPAGE="https://github.com/tnalpgge/rank-amateur-cowsay"
 COPYRIGHT="1999-2000 Tony Monroe"
 LICENSE="GNU GPL v3"
 REVISION="1"
-SOURCE_URI="$HOMEPAGE/archive/$portName-$portVersion.tar.gz"
-CHECKSUM_SHA256="9a32cba0b561719e1302398857124fdf4dd21a7623c017eaadbf31277f9274af"
+SOURCE_URI="$HOMEPAGE/archive/cowsay-$portVersion.tar.gz"
```
alaviss: It's not strictly required in this case, since the architecture was `any`.

Crestwave: Oh, I just changed it since there's no alternative to the value, and `$portName` isn't shorter than `cowsay`.

fbrosson:
Thanks @Crestwave for your PR!

@alaviss:
> I think `cowsay` should be patched to look for cows in `B_FIND_PATH_DATA_DIRECTORY`, since with the current setup, it can only seek for cows that came with it. That's not required for this PR however.

```bash
$ grep COWPATH /bin/cowsay 
$cowpath = $ENV{'COWPATH'} || '/packages/cowsay-3.04-1/.self/data/cows';
```
@alaviss: Anyone willing to have other data files can set `COWPATH` to e.g. `/boot/home/data/cows:/boot/system/data/cows`, so I also think we can merge this PR and have a separate PR for the patch you are suggesting, in case someone would like to implement it.

So, shall we merge as is?

PR #3387, opened 2018-11-18
BeNetTris [recipe]

bach5000:
There's this problem with `Makefile having a modification time 19648s in the future` :(


haiku-games/benettris/benettris-0.2.recipe
```diff
@@ -0,0 +1,47 @@
...
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	cmd:find
+	"
+
+BUILD()
+{
+	find . -exec touch {} \;
```
alaviss: This should be moved to `PATCH()` instead.

haiku-games/benettris/benettris-0.2.recipe
```diff
@@ -0,0 +1,47 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	cmd:find
```
alaviss: Please order the `cmd:` alphabetically

haiku-games/benettris/benettris-0.2.recipe
alaviss: A clock skew is rather scare, have you checked if the clock on your machine was correct?

fbrosson: > There's this problem with `Makefile having a modification time 19648s in the future` :(

What happened here is that your system is not using the right timezone and/or time, so when you grab a tarball which contains files recently modified, they have a timestamp which appears to be in the future.
I guess your timezone is UTC instead of UTC+x (with x being probably something between 9 and 11).

BTW, once you have fixed your clock settings you will probably see that you don't need to touch the files.

fbrosson: @bach5000: Okay. I found what your real timezone is. It is correctly set in one (or some) of your systems, but not in all. One (or maybe more) of your systems is using UTC as its timezone. That's Okay to use this, but if you do that then you have to set your system's clock accordingly, i.e. to the real UTC time, not the localtime.

haiku-games/benettris/benettris-0.2.recipe
```diff
@@ -0,0 +1,46 @@
...
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:find
+	cmd:g++
+	cmd:make
+	"
+
+BUILD()
+{
+	make OBJ_DIR=objects
```
alaviss:
```suggestion
	make $jobArgs OBJ_DIR=objects
```

haiku-games/benettris/benettris-0.2.recipe
```diff
@@ -0,0 +1,46 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="5d04121340b7c730160d5eab6a97c79b97168464"
+SOURCE_URI="https://github.com/HaikuArchives/BeNetTris/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="bd3096b84601a257cd22818ba0396be600d4b99e2f9d02f000d029502af9eddc"
+SOURCE_FILENAME="BeNetTris-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="BeNetTris-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 x86_64"
+
```
alaviss: Please add [`USER_SETTINGS_FILES`](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#user_settings_files) containing configuration files/folders that the program puts in the user home directory.

bach5000:
@alaviss Added USER_SETTINGS_FILES. Don't know if I did it right, though...
The app saves its configs to /boot/home/config/settings/BeNetTrisSettings.

haiku-games/benettris/benettris-0.2.recipe
```diff
@@ -0,0 +1,50 @@
...
+USER_SETTINGS_FILES="
+	settings/BeNetTrisSettings
```
alaviss:
```suggestion
	settings/BeNetTrisSettings directory
```

Because it's a directory ;)

bach5000:
@alaviss Is this better?

PR #3389, opened 2018-11-19
edgar, bump version to 1.29

Begasus:


Begasus:
If someone could check this on x86_64 it would be nice :) (my setup isn't up to it for this one) :)

Begasus:
Checked and builds/runs fine on x86_64 too :)

PR #3404, opened 2018-11-23
ProjectConceptor [fix build on x86_64]

bach5000:
Fixed build on x86_64.

### DIFF
--- a/haiku-apps/projectconceptor/patches/ProjectConceptor-0.1.0.patchset
+++ b/haiku-apps/projectconceptor/patches/ProjectConceptor-0.1.0.patchset
@@ -1,23 +1,9 @@
-From f1b8bddf3595fe3a45735365fe041f3a0995400c Mon Sep 17 00:00:00 2001
+From 03c29710636482cca365bdeb471da1e769475bf2 Mon Sep 17 00:00:00 2001
 From: Bach Nguyen <bach5000@gmail.com>
-Date: Mon, 19 Nov 2018 21:30:03 +0000
-Subject: Fix build
+Date: Fri, 23 Nov 2018 12:14:25 +0000
+Subject: Fix build on x86_64
 
 
-diff --git a/src/app/About/AboutGitInfo.h b/src/app/About/AboutGitInfo.h
-new file mode 100644
-index 0000000..ecdc597
---- /dev/null
-+++ b/src/app/About/AboutGitInfo.h
-@@ -0,0 +1,8 @@
-+
-+#ifndef GIT_INFO_H
-+#define GIT_INFO_H
-+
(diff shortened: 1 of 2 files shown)


alaviss:
I'd be appreciated if you could upstream this patchset instead :)

bach5000:
@alaviss To the owner of ProjectConceptor? Sure :)

Begasus:
Compare your changes to the latest changes at the resource (maybe start from the most recent srcGitRev checkout?)

bach5000:
@Begasus Seems like the author only changed the directories and updated the Makefile since this revision.
I'll make a new patchset.

Begasus:
Even then @bach5000 starting out from the most recent commit should be used (when it doesn't brake the build) :)

Begasus:
Still builds ok on x86_gcc2 (primary architecture) :+1:

bach5000:
@Begasus Could you check my GCI task? :)

Begasus:
Just checked for primary arch in x86_gcc2, stilll need to check on x86_64 :)

bach5000:
Thanks, @Begasus! :D

Begasus:
Build is ok on x86_64, but upon closing the app it crashes (will try to upload a report later)

bach5000:
@Begasus This crash seems to be caused by a bug in ProjectConceptor... I might have to make a new issue on the upstream repo.
I'll also make a PR on the upstream repo :)

korli:
This patch leaves me puzzled and should be declined. Casting pointers to int32 will eventually lead to problems.
For instance the map<int32,BMessage*> is used with pointers as keys.  Why not just switch to map<void*,BMessage*>?

bach5000:
@korli I found out that I could use "-fpermissive" to get the build to run... I don't know if this would be a valid patch though...

korli:
@bach5000 no,  "-fpermissive" just hides the mess, while the problems pointed by the compiler are true problems for x86_64.

alaviss:
Can you update the version to `0.1.0_<date of commit in yyyymmdd>`. Also, it would be nice to upstream the patchset.

Begasus:
New version release, might be good to investigate if the patch still applies :)

Paradoxianer:
Thank you for alle the work done here, i will look into the issue with the x86_64 it as soon as i have a 64 Haiku running. There are definitly some serios problemes with my code here... not shure if i can fix it fast... since i didnt looked at the code for nearly a year and did no serios devolpment for serveral years... thank you @korli for the tipp with the map

Begasus:
Closing this one as wip is on the way in the new version https://github.com/haikuports/haikuports/pull/4405

PR #3406, opened 2018-11-23
Add SUMMARY + DESCRIPTION to 2 ports

bach5000:
Noticed that these had no SUMMARY and DESCRIPTION... so I added them.


scottmc:
The recipes also need to work.  These will not.  Usually when the Summary and/or Description is missing, it's on old recipe.  These need to be updated to work with the current recipe format.  Good news is that most of these used to work, so they shouldn't be too hard to get working.

bach5000:
@scottmc @alaviss @fbrosson:
The PyXML recipe should run now... still need some help with `LICENSE`: there's a license file with multiple licenses, but I don't know what to put in the `LICENSE` part of the recipe :(

Working on rdflib.

alaviss:
Your recipe would need to generate package for both python version (2.7, 3.0). You should take a look at `dev-python/tzlocal` as an example for how to write such recipe.

bach5000:
@alaviss Updated the recipes :)

dev-python/pyxml/pyxml-0.8.4.recipe
```diff
@@ -1,20 +1,94 @@
-DESCRIPTION="Python XML package"
+SUMMARY="Python2 XML package"
```
alaviss:
```suggestion
SUMMARY="Python XML package"
```

dev-python/pyxml/pyxml-0.8.4.recipe
```diff
@@ -1,20 +1,94 @@
...
+itself), and a C helper module that can speed up xmllib.py by a factor of 5."
 HOMEPAGE="http://pyxml.sourceforge.net/topics/index.html"
-SOURCE_URI="http://sourceforge.net/projects/pyxml/files/latest/download?source=files"
-CHECKSUM_MD5="1f7655050cebbb664db976405fdba209"
+COPYRIGHT="2000 Fourthought Inc."
+LICENSE="Python"
 REVISION="1"
-STATUS_HAIKU="untested"
-DEPEND="dev-lang/python >= 2.6.4"
+SOURCE_URI="https://downloads.sourceforge.net/project/pyxml/pyxml/$portVersion/PyXML-$portVersion.tar.gz"
```
alaviss:
```suggestion
SOURCE_URI="https://downloads.sourceforge.net/pyxml/PyXML-$portVersion.tar.gz"
```

SourceForge magic

dev-python/pyxml/pyxml-0.8.4.recipe
```diff
@@ -1,20 +1,94 @@
-DESCRIPTION="Python XML package"
+SUMMARY="Python2 XML package"
+DESCRIPTION="Python XML comes with a validating XML parser, an implementation of the SAX \
+and DOM programming interfaces, an interface to the Expat parser (and the Expat parser \
+itself), and a C helper module that can speed up xmllib.py by a factor of 5."
 HOMEPAGE="http://pyxml.sourceforge.net/topics/index.html"
-SOURCE_URI="http://sourceforge.net/projects/pyxml/files/latest/download?source=files"
-CHECKSUM_MD5="1f7655050cebbb664db976405fdba209"
+COPYRIGHT="2000 Fourthought Inc."
+LICENSE="Python"
```
alaviss: There are many more. Check the `LICENSE` file in the archive.

dev-python/pyxml/pyxml-0.8.4.recipe
```diff
@@ -1,20 +1,94 @@
...
 REVISION="1"
-STATUS_HAIKU="untested"
-DEPEND="dev-lang/python >= 2.6.4"
+SOURCE_URI="https://downloads.sourceforge.net/project/pyxml/pyxml/$portVersion/PyXML-$portVersion.tar.gz"
+CHECKSUM_SHA256="9fab66f9584fb8e67aebd8745a5c97bf1c5a2e2e461adf68862bcec64e448c13"
+SOURCE_DIR="PyXML-$portVersion"
+
+ARCHITECTURES="any"
```
alaviss: Since `gcc` is used, this package contains architecture-specific bits, please expand this to multiple architectures.

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,97 @@
+SUMMARY="Python library for working with RDF"
+DESCRIPTION="RDFLib is a Python library for working with RDF, a simple \
+yet powerful language for representing information as graphs."
+HOMEPAGE="https://github.com/RDFLib/rdflib"
+COPYRIGHT="2002-2009 Daniel Krech"
+LICENSE="Public Domain"
```
alaviss: I don't think [this](https://github.com/RDFLib/rdflib/blob/1503dae6049f0d9b14d9d7f884d8de4cb38b39a3/LICENSE) is "Public Domain"

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,97 @@
...
+REVISION="1"
+SOURCE_URI="http://github.com/RDFLib/rdflib/archive/$portVersion.zip"
+CHECKSUM_SHA256="fe7fb695e840d812062c2fdc81b7ddb4e2b551e07a6635b7b4e5ac64a8fdabdd"
+SOURCE_DIR="rdflib-$portVersion"
```
alaviss:
```suggestion
```

Since it's the recipe name, it can be omitted.

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,97 @@
+SUMMARY="Python library for working with RDF"
+DESCRIPTION="RDFLib is a Python library for working with RDF, a simple \
+yet powerful language for representing information as graphs."
+HOMEPAGE="https://github.com/RDFLib/rdflib"
+COPYRIGHT="2002-2009 Daniel Krech"
+LICENSE="Public Domain"
+REVISION="1"
+SOURCE_URI="http://github.com/RDFLib/rdflib/archive/$portVersion.zip"
```
alaviss:
```suggestion
SOURCE_URI="http://github.com/RDFLib/rdflib/archive/$portVersion.tar.gz"
```

Because it's smaller

bach5000:
@alaviss Done!

dev-python/pyxml/pyxml-0.8.4.recipe
```diff
@@ -1,20 +1,100 @@
-DESCRIPTION="Python XML package"
+SUMMARY="Python XML package"
+DESCRIPTION="Python XML comes with a validating XML parser, an implementation of the SAX \
+and DOM programming interfaces, an interface to the Expat parser (and the Expat parser \
+itself), and a C helper module that can speed up xmllib.py by a factor of 5."
 HOMEPAGE="http://pyxml.sourceforge.net/topics/index.html"
-SOURCE_URI="http://sourceforge.net/projects/pyxml/files/latest/download?source=files"
-CHECKSUM_MD5="1f7655050cebbb664db976405fdba209"
+COPYRIGHT="2000 Fourthought Inc
+	1991 - 1995, Stichting Mathematisch Centrum Amsterdam
```
alaviss:
```suggestion
	1991-1995, Stichting Mathematisch Centrum Amsterdam
```

dev-python/pyxml/pyxml-0.8.4.recipe
```diff
@@ -1,20 +1,100 @@
...
+	Greg Stein
```
alaviss:
```suggestion
```

Being in public domain means that the author has waived their copyright, so we shouldn't specify them here.

bach5000:
@alaviss Fixed some things :)

bach5000:
@alaviss ... and fixed them again.
If I don't use BUILD_PREREQUIRES, should I leave it as it is right now or do I just get rid of it?

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,95 @@
...
+	pythonPackage=${PYTHON_PACKAGES[i]}
+	pythonVersion=${PYTHON_VERSIONS[$i]}
+
+	eval "PROVIDES_$pythonPackage=\"
+		${portName}_$pythonPackage = $portVersion
+		cmd:csv2rdf
+		cmd:rdf2dot
+		cmd:rdfgraphisomorphism
+		cmd:rdfpipe
+		cmd:rdfs2dot
```
alaviss: @korli @fbrosson This package provides `cmd:`s for both `_python` and `_python3` package. How should we resolve the conflict when an user want to install both packages?

bach5000: @fbrosson I might need help...
The `if` statement on line 87 (the INSTALL function) is causing problems for me... I tried following the examples (tried both renaming and removing), but HaikuPorter keeps telling me that I can't remove or rename the directory `$binDir/rdflib` because it doesn't exist.

fbrosson: @bach5000: You need to find out what gets really installed in each subpackage (without any deletions or renames) before you can decide what you need to remove or rename.

bach5000: @fbrosson: oops. just figured out what I did wrong for my first problem :)
But what about the policy errors? I put the bins in PROVIDES for python... why can't it find them?

bach5000: @fbrosson Should I remove the binaries (like coverage) or should I rename them (like pbr)? What do you think?

bach5000:
@fbrosson @alaviss: Pushed the changes to this PR!

The new package contents seems to be correct for `rdflib_python3`... but I'm not sure if I fixed the right thing on the recipe.

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,108 @@
+SUMMARY="Python library for working with RDF"
+DESCRIPTION="RDFLib is a Python library for working with RDF, a simple \
+yet powerful language for representing information as graphs."
+HOMEPAGE="https://github.com/RDFLib/rdflib"
+COPYRIGHT="2002-2017 RDFLib Team"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://github.com/RDFLib/rdflib/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="a362926f77624a8abfb93042c28607f49445d003792d5271cd07f0610ab16033"
```
alaviss:
```suggestion
CHECKSUM_SHA256="a362926f77624a8abfb93042c28607f49445d003792d5271cd07f0610ab16033"
SOURCE_FILENAME="rdflib-$portVersion.tar.gz"
```

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,108 @@
+SUMMARY="Python library for working with RDF"
+DESCRIPTION="RDFLib is a Python library for working with RDF, a simple \
+yet powerful language for representing information as graphs."
+HOMEPAGE="https://github.com/RDFLib/rdflib"
+COPYRIGHT="2002-2017 RDFLib Team"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="http://github.com/RDFLib/rdflib/archive/$portVersion.tar.gz"
```
alaviss:
```suggestion
SOURCE_URI="https://github.com/RDFLib/rdflib/archive/$portVersion.tar.gz"
```

dev-python/rdflib/rdflib-4.2.2.recipe
fbrosson: It depends:
* on which version of python you want the "non-suffixed" entry point to use,
* and on what gets installed when there are no renames nor deletes.

Imagine you would like the "non-suffixed" entry point to use python3. If the python (i.e. python2) subpackage comes with "foobar" and "foobar2" (and if they are equivalent entry points) then you ~can~ _should_ delete "foobar" in the python subpackage.

If, for some reason, you prefer the "non-suffixed" entry point to use python 2 then you will need to:
* rename "foobar" to "foobar3" in the python3 subpackage (if it does not already install a "foobar3").
* delete "foobar" in the python3 subpackage (if it already comes with a "foobar3").

Last "rule": if the python module cannot be built with gcc2 then on x86_gcc2 there won't be any python (i.e. python 2) subpackage, so the "non-suffixed" entry point should probably use python3.
BTW, in that case you need to enable x86 secondary arch and "populate" `PYTHON_PACKAGES`  and `PYTHON_VERSIONS` like in dev-python/coverage. That's the only way to have a 100% correct recipe in all cases because on x86_gcc2 non hybrid the recipe would not even be buildable.

bach5000:
@fbrosson @leorize Does this look correct?
@fbrosson The package comes with some commands, but none of them seem to be specific to a python version... and I'd prefer them to use python3 instead.

But it seems to build on all arch... Was deleting the binaries for python2 package the right thing?

dev-python/pyxml/pyxml-0.8.4.recipe
alaviss: That's how most of our ports present ranged years, so please apply the change.

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,109 @@
...
+
+		python=python$pythonVersion
+		installLocation=$prefix/lib/$python/vendor-packages/
+		export PYTHONPATH=$installLocation
+		mkdir -p "$installLocation"
+		$python setup.py install \
+			--root=/ --prefix="$prefix"
+
+		if [ "$pythonPackage" = python ]; then
+			rm "$prefix"/bin/*
```
alaviss:
```suggestion
			rm -rf "$prefix"/bin
```

This will prevent empty `bin` directory in the `_python` package.

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,109 @@
...
+
+		python=python$pythonVersion
+		installLocation=$prefix/lib/$python/vendor-packages/
+		export PYTHONPATH=$installLocation
+		mkdir -p "$installLocation"
+		$python setup.py install \
+			--root=/ --prefix="$prefix"
+
+		if [ "$pythonPackage" = python ]; then
+			rm -rf "$prefix"/bin
```
alaviss:
```suggestion
			rm -rf "$binDir"
```

Since this is an `any` architecture package, `$binDir` and `$prefix/bin` is the same.

dev-python/rdflib/rdflib-4.2.2.recipe
```diff
@@ -0,0 +1,109 @@
...
+		fi
+
+		packageEntries $pythonPackage \
+			"$prefix"/lib/$python \
+			"$binDir"
```
alaviss: This should only be packaged if it actually exists. An approach mainly used here is `maybe_`  prefix, see [this] for an example.

[this]: https://github.com/haikuports/haikuports/blob/1e7e03ae1a02a3e070fb051b082bfdf35ed603cc/dev-libs/fribidi/fribidi-1.0.5.recipe#L79

PR #3423, opened 2018-12-01
jigit [recipe]

bach5000:
May need some help with `configure` again...

Says `configure: error: cannot guess build type; you must specify one`. 
Also mentions in the beginning of the build that `This script, last modified 2009-12-30, has failed to recognize the operating system you are using`.

So... I made a Haiku Makefile instead... is this a reasonable option?


extrowerk:
```libtoolize -fci``` should fix it.

**Comment by Begasus** [comment]
For libjte you can use "autoreconf -vfi" to recreate configure, you also need to "touch" a few files before you can actually run "configure" :)

File: app-misc/jigit/jigit-1.20.recipe [hunk 1]
```diff
@@ -0,0 +1,59 @@
...
+CHECKSUM_SHA256="459285bd743645381912301578316f2bfafb99ce1c10140a01af2b6cb011f193"
+PATCHES="jigit-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+
+PROVIDES="
+	jigit = $portVersion
+	cmd:jigit = $portVersion
+	cmd:jigdump = $portVersion
+	cmd:jigsum = $portVersion
```
Begasus: missing a few cmd's here

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
...
+	"
+
+BUILD()
+{
+	make
+}
+
+INSTALL()
+{
+	install -d $appsDir/jigit
```
Begasus: these are binaries and should go into $binDir

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
...
+	install -d $appsDir/jigit/lib
+	cp extract-data $appsDir/jigit
+	cp jigdump $appsDir/jigit
+	cp jigit $appsDir/jigit
+	cp jigit-mkimage $appsDir/jigit
+	cp jigsum $appsDir/jigit
+	cp mkjigsnap $appsDir/jigit
+	cp rsyncsum $appsDir/jigit
+	cd libjte/object*
+	cp libjte.so $appsDir/jigit/lib
```
Begasus: if you build the library first you can use "make install" in libjte to install correctly (library + headers), you should also provide a library in PROVIDES and create a devel package also

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
+SUMMARY="A tool to ease the distribution of very large files"
+DESCRIPTION="Jigsaw Download, or jigdo, is a tool designed to ease the \
+distribution of very large files over the internet such as CD or DVD \
+images. Its aim is to make downloading these images as easy as a click \
+on a direct download link in a browser, while avoiding all the problems \
+that server administrators could have with hosting large files."
+HOMEPAGE="http://www.einval.com/~steve/software/JTE/"
```
fbrosson: Suggestion: `http` :arrow_right: `https`

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
...
+COPYRIGHT="2014 Steve McIntyre"
```
fbrosson: Maybe:
```diff
-COPYRIGHT="2014 Steve McIntyre"
+COPYRIGHT="2004-2014 Steve McIntyre"
```
(I'm not sure about this. In particular, I don't know if the project was initiated by other authors in 2004.)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://www.einval.com/~steve/software/JTE/download/jigit_$portVersion.orig.tar.gz"
```
fbrosson: Suggestion: `http` :arrow_right: `https`

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
...
+	lib:libz
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libbz2
+	devel:libz
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
```
fbrosson: Probably:
```diff
-	makefile_engine
```
makefile_engine is only needed for programs written for BeOS or Haiku.

Begasus:
The patchset isn't needed if you first build the library (also no need for the extra Makefile)

Begasus:
The recipe doesn't build for gcc2 so you can set "!x86_gcc2" also (cc1: unknown C standard c99 (-std=c99 already included in the makefile))

app-misc/jigit/jigit-1.20.recipe
Begasus: EDIT: as the build doesn't build for gcc2, maybe use $commandSuffix for the binaries and $commandBinDir here (or $prefix/bin)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
+SUMMARY="A tool to ease the distribution of very large files"
+DESCRIPTION="Jigsaw Download, or jigdo, is a tool designed to ease the \
+distribution of very large files over the internet such as CD or DVD \
+images. Its aim is to make downloading these images as easy as a click \
+on a direct download link in a browser, while avoiding all the problems \
+that server administrators could have with hosting large files."
+HOMEPAGE="http://www.einval.com/~steve/software/JTE/"
+COPYRIGHT="2014 Steve McIntyre"
+LICENSE="GNU GPL v2"
```
Begasus: libjte comes with license "GNU LGPL v2.1"

app-misc/jigit/jigit-1.20.recipe
Begasus: "George Danchev" and "Thomas Schmitt" probably need to be included here too (for libjte)

app-misc/jigit/jigit-1.20.recipe
Begasus: Another Suggestion :)
```diff
- SOURCE_URI="http://www.einval.com/~steve/software/JTE/download/jigit_$portVersion.orig.tar.gz" 
+ SOURCE_URI="$HOMEPAGE/download/jigit_$portVersion.orig.tar.gz" 
```

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,59 @@
...
+REVISION="1"
+SOURCE_URI="http://www.einval.com/~steve/software/JTE/download/jigit_$portVersion.orig.tar.gz"
+CHECKSUM_SHA256="459285bd743645381912301578316f2bfafb99ce1c10140a01af2b6cb011f193"
+PATCHES="jigit-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
Begasus: Needed because libjte doesn't build for gcc2
```diff
+SECONDARY_ARCHITECTURES="x86"
```

bach5000:
Sorry for taking so long! Had to go somewhere for an entire day :)
Fixed everything! Will update the PR shortly.

bach5000:
@fbrosson @Begasus @extrowerk:
Thanks for all your help! I updated the PR, and the recipe should now run correctly :)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,94 @@
...
+	cmd:autoconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize
+	cmd:make
+	cmd:touch
+	"
+
+BUILD()
+{
+	libtoolize -fci
```
Begasus: Not needed here:
```diff
-	libtoolize -fci
```

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,94 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libbz2$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize
+	cmd:make
+	cmd:touch
```
Begasus: Missing cmd's for aclocal, awk, pkg-config

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libbz2$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:autoconf
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize
```
Begasus:
```diff
-	cmd:libtoolize
+	cmd:libtoolize$secondaryArchSuffix
```

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,94 @@
...
+PROVIDES="
+	jigit$secondaryArchSuffix = $portVersion
+	cmd:extract_data$commandSuffix = $portVersion
+	cmd:jigit$commandSuffix = $portVersion
+	cmd:jigdump$commandSuffix = $portVersion
+	cmd:jigsum$commandSuffix = $portVersion
+	cmd:jigit_mkimage$commandSuffix = $portVersion
+	cmd:mkjigsnap$commandSuffix = $portVersion
+	cmd:rsyncsum$commandSuffix = $portVersion
+	lib:libjte$secondaryArchSuffix = $portVersion
```
Begasus: The actual version is "1.0.0" for libjte so suggestion here and in devel with:
```diff
-	lib:libjte$secondaryArchSuffix = $portVersion
+	lib:libjte$secondaryArchSuffix = 1.0.0 compat >= 1
```
Or you could use libVersion(Compat) here :)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,94 @@
...
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	jigit$secondaryArchSuffix = $portVersion
+	cmd:extract_data$commandSuffix = $portVersion
+	cmd:jigit$commandSuffix = $portVersion
+	cmd:jigdump$commandSuffix = $portVersion
```
Begasus: jigdump comes before jigit

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,96 @@
...
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config
```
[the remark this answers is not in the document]
Begasus: Doesn't hurt if you add it if you are building for secondaryarch :)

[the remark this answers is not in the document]
bach5000: @Begasus @fbrosson So I guess I should add `$secondaryArch`?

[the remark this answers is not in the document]
fbrosson: I'd say yes, because we do that on all other recipes, but I'm ok without it, because it looks like it isn't needed here.

Yeah, adding it seems slightly better.

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,96 @@
...
+	touch NEWS README AUTHORS
+	autoreconf -vfi
+	runConfigure ./configure
+	make $jobArgs
+	make -C ../
+}
+
+INSTALL()
+{
+	install -d "$commandBinDir"
```
fbrosson: Suggestion:
```diff
 INSTALL()
 {
-	install -d "$commandBinDir"
+	install -m 755 -d "$commandBinDir" "$manDir"/man1 "$manDir"/man8
-	cp extract-data "$commandBinDir"
-	cp jigdump "$commandBinDir"
-	cp jigit "$commandBinDir"
-	cp jigit-mkimage "$commandBinDir"
-	cp jigsum "$commandBinDir"
-	cp mkjigsnap "$commandBinDir"
-	cp rsyncsum "$commandBinDir"
+	install -m 755 -t "$commandBinDir" \
+		extract-data \
+		jigdump jigit jigit-mkimage jigsum \
+		mkjigsnap \
+		rsyncsum
+	install -m 644 -t "$manDir"/man1 *.1
+	install -m 644 -t "$manDir"/man8 *.8
```
(Feel free to arrange/rework as you wish.)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,96 @@
+SUMMARY="A tool to ease the distribution of very large files"
+DESCRIPTION="Jigsaw Download, or jigdo, is a tool designed to ease the \
+distribution of very large files over the internet such as CD or DVD \
+images. Its aim is to make downloading these images as easy as a click \
+on a direct download link in a browser, while avoiding all the problems \
```
fbrosson: Oops, that's the description for [jigdo](https://github.com/haikuports/haikuports/blob/c2a214e6473117adfdbf84d0f792341b30e1bbd7/app-misc/jigdo/jigdo-0.7.3.recipe#L2..L6).

Keeping part of it is probably ok, but you should probably also mention that jigit is the name of the user-friendly wrapper around mkimage.

More info on https://www.einval.com/~steve/software/JTE/:

> **mkimage**, a simple and very fast tool to reconstruct image files from .jigdo and .template files. It doesn't have any logic to cope with downloading missing files, but will list the missing files that are needed. It is also much faster for people (like me!) who already have full local mirrors.
>
> **jigit** is a first attempt at a user-friendly wrapper for mkimage on a user's machine.

(I guess including these whole paragraphs in `DESCRIPTION` would not be ok, but I don't have have a strong opinion here.)

What about:
> jigit aims to be a user-friendly wrapper for mkimage.
> mkimage is a simple and very fast tool to reconstruct image files from .jigdo and .template files. It doesn't have any logic to cope with downloading missing files, but will list the missing files that are needed. It is also convenient in case the user has a partially outdated optical media or ISO image containing at least some files which are not outdated.

Other suggestions for `DESCRIPTION` (and `SUMMARY`) are welcome!.

bach5000:
Sorry for taking so long! I patched the recipe according to the suggestions above :)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,96 @@
+SUMMARY="A user-friendly tool to ease the distribution of very large files"
+DESCRIPTION="jigit aims to be a user-friendly wrapper for mkimage. mkimage is \
+a simple but fast tool to reconstruct large image files from .jigdo and .template \
+files. It doesn't have any logic to cope with downloading missing files, but will \
+list the missing files that are needed. It is also convenient in case the user has \
```
fbrosson: Hmm, I just realized that my suggestion for `DESCRIPTION` needs a little fix because the mkimage tool is called jigit-mkimage.
```diff
-DESCRIPTION="jigit aims to be a user-friendly wrapper for mkimage. mkimage is \
-a simple but fast tool to reconstruct large image files from .jigdo and .template \
-files. It doesn't have any logic to cope with downloading missing files, but will \
-list the missing files that are needed. It is also convenient in case the user has \
-a partially outdated optical media or ISO image containing at least some files \
-which are not outdated."
+DESCRIPTION="jigit aims to be a user-friendly wrapper for jigit-mkimage. \
+jigit-mkimage is a simple but fast tool to reconstruct large image files from \
+\".jigdo\" and \".template\" files. \
+It doesn't have any logic to cope with downloading missing files, but will \
+list the missing files that are needed. \
+It is also convenient in case the user has a partially outdated optical media \
+or ISO image containing at least some files which are not outdated."
```

fbrosson:
A few more remarks:
* `cmd:jigit` is declared as a bash script, but I'm quite sure it would work with `/bin/sh`.
* In case we don't replace the `#!/bin/bash` by `#!/bin/sh` (in `cmd:jigit`) then we should probably add `cmd:bash` to `REQUIRES`.
* `cmd:jigit` calls several tools provided by other packages: `wget`, `xargs`, `wc`, `awk` and `printf`, so it would probably be a good idea to mention the corresponding `cmd:` items in `REQUIRES`.
Yeah, `printf` is available as a bash builtin command and as a runtime (from coreutils), so if `cmd:bash` is added to `REQUIRES` then we don't need to mention `cmd:printf`.
* `cmd:jigit` might need some polishing (e.g. reading `jigdo.conf` from `/boot/system/settings/` and `/boot/home/config/settings/` instead of the current default locations) because the unmodified `cmd:jigdo`:
  * reads `/etc/jigit.conf` if that file exists,
  * reads `~/.jigit.conf` if that file exists,
  * creates `/tmp/jigit/` if `TMPDIR` is not defined.

Would you like to take care of all these little details? Thanks!
(There is no urgency at all, so you can leave that for after December 12 :)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,98 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libbz2$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:awk
+	cmd:bash
```
Begasus: I think @fbrosson meant that you put those in "REQUIRES" instead of "BUILD_PREREQUIRES" :)

Begasus:
I see they released 1.21 (released this year), maybe you could bump to the newer version? :)

bach5000:
@Begasus I will do so, but currently I have to apply to several universities with deadlines coming up and this is taking up quite a lot of time.
I'm really sorry that I won't be much of a help, at least for two weeks. I hope you understand :(

Begasus:
No problem, you need to set your priorities :+1:

korli:
@Begasus is this recipe ready (except for the cmd:bash in REQUIRES)? I would leave out the version bump for now.

Begasus:
@korli I'll do a final check on both arch's and let you know, but I think it was when we checked it back then :)

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,98 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libbz2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libjte${secondaryArchSuffix}_devel = $portVersion
+	devel:libjte$secondaryArchSuffix = $portVersion
```
Begasus: This should be:
```diff
-	devel:libjte$secondaryArchSuffix = $portVersion
	devel:libjte$secondaryArchSuffix = 1.0.0 compat >= 1
```

app-misc/jigit/jigit-1.20.recipe
```diff
@@ -0,0 +1,98 @@
...
+	"
+REQUIRES_devel="
+	libjte$secondaryArchSuffix == $portVersion base
```
Begasus: This should be:
```diff
-	libjte$secondaryArchSuffix == $portVersion base
+	jigit$secondaryArchSuffix == $portVersion base
```

Begasus:
Those last comments should do it @korli and @bach5000

korli:
@Begasus do you take over your changes?

Begasus:
> @Begasus do you take over your changes?

I'll take a look at it later today @korli  :)

Begasus:
> Any progress on this? Checking up on something that's using libjte also :)

If only I could remember what I was checking it for now :)

PR #3427, opened 2018-12-01
libyaml, new recipe

Begasus:


dev-libs/libyaml/libyaml-0.2.1.recipe
```diff
@@ -0,0 +1,76 @@
+SUMMARY="A C library for parsing and emitting YAML"
+DESCRIPTION="LibYAML covers presenting and parsing processes. Thus LibYAML \
+defines the following two processors:
+* Parser, which takes an input stream of bytes and produces a sequence of parsing events.
```
fbrosson: Suggestion:
```diff
-* Parser, which takes an input stream of bytes and produces a sequence of parsing events.
+* Parser, which takes an input stream of bytes and produces a sequence of \
+parsing events.
```
(Everything else is OK and I can confirm that the recipe builds fine and passes all tests on all arches.)

PR #3430, opened 2018-12-01
fre:ac: bump to 1.1~alpha20181201

enzo1982:
Update fre:ac to version 1.1 alpha 20181201.

Finally we can get rid of the patches and additional files for fre:ac as Haiku is now fully supported by fre:ac. :)

The smooth and BoCA PRs should be merged first.


enzo1982:
The new version depends on some additions in the new smooth pre-release. Looks like the build server did not update smooth to ~pre5 before trying to build fre:ac.

Lan72:
Crashes when launching. Here is the error report:
file:///boot/home/Desktop/freac-164612-debug-02-12-2018-01-33-32.report.txt

enzo1982:
@Lan72 I think this is caused by dependency issues too. As smooth 0.8.74 is still a pre-release it sometimes introduces API changes. This was not reflected in the BoCA and fre:ac recipes which caused incompatibilities here.

PRs #3431 and #3432 should fix this by declaring dependencies on specific package versions.

PR #3432, opened 2018-12-02
fre:ac fix dependencies and bump to 1.1~alpha20181201a

enzo1982:
This bumps the version to the 20181201a hotfix release and declares smooth 0.8.74.0~pre5 and BoCA 1.0~alpha20181201 as explicit dependencies to fix issues with the previous recipe.

The dependency issue here is that smooth 0.8.74 still is a pre-release and the ~pre5 revision introduced API changes compared to ~pre4. This caused issues with BocA and fre:ac.

Once fre:ac 1.1 goes beta it will depend on an API/ABI stable smooth release and BoCA will freeze its API too.


Lan72:
Thank you very much. Now fre:ac works well. I just now ripped two CDs.

PR #3437, opened 2018-12-02
GZDoom: bump

extrowerk:


diversys:
Does it still work tho? I specifically used https://github.com/coelckers/gzdoom/tree/legacy branch as it requires older OpenGL(?) whereas other version would complain about it and quit right away.

extrowerk:
Exporting the Mesa overridden env var makes it works, but it is definetely slow.
Thanks for the info, i know now ehy the legacy one was used. Will use that.

diversys:
gcc2 build is broken https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/2420/builds/31480.log

PR #3443, opened 2018-12-04
faad2: bump to 2.8.8

enzo1982:


korli:
Just a remark: when you reuse an older patch, it would be nice to keep the original author.

enzo1982:
> Just a remark: when you reuse an older patch, it would be nice to keep the original author.

Yeah, sorry! I wasn't sure whether it's better to add another patch commit or to consolidate into one.

Thanks for merging!

PR #3456, opened 2018-12-08
OpenBLAS: bump, patches upstreamed

extrowerk:
There were no ABI break.
https://abi-laboratory.pro/index.php?view=timeline&l=openblas


extrowerk:
Breaks on x86-64 but never did locally.
I think something wrong with gcc.

PR #3459, opened 2018-12-09
Recipe for SIDPlayer.

arnavbhatt288:
Recipe for SIDPlayer. Keep in mind that it doesn't uses any makefile (even BeOS makefile after updating it didn't work) so I have to use configure to make new makefile for it.
GCI Student: arnavbhatt288

### DIFF
--- a/haiku-apps/sidplayer/sidplayer-4.4.recipe
+++ b/haiku-apps/sidplayer/sidplayer-4.4.recipe
@@ -10,7 +10,6 @@ LICENSE="GNU GPL v1"
 REVISION="1"
 SOURCE_URI="http://sidplayer.cebix.net/downloads/SIDPlayer-4.4.tar.gz"
 CHECKSUM_SHA256="fd1f73047abeb8234f850975786c805262678e77013835a8e621dcfd237624d7"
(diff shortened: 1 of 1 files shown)


haiku-apps/sidplayer/sidplayer-4.4.recipe
```diff
@@ -0,0 +1,49 @@
...
+	cmd:gcc
+	cmd:make
+	cmd:autoreconf
+	cmd:aclocal
+	"
+
+BUILD()
+{
+	autoreconf -vfi
+	runConfigure --omit-dirs binDir ./configure --bindir=$binDir
```
korli: Isn't --bindir already the standard option used by runConfigure?

arnavbhatt288: @korli I don't know.

korli: Then just use: "runConfigure ./configure"

arnavbhatt288:
Hello, I will make more recipes for Haiku but since I need to contribute for other project (ReactOS), it will take some time.
Hopefully I will make new PR soon :)

PR #3460, opened 2018-12-09
SciPy: new recipe

extrowerk:
python3 version is broken, no idea, why.


extrowerk:
Compiling eith python3 the compiler complains about (afaik) redefinition of true and false. I dont really have the experience to tell if it is bad programming practice or a bug in the compiler (gcc, not python!), python3 incompatibility or just a header conflicts somewhere.
But with python2 it builds fine.

PR #3472, opened 2018-12-16
fff: new recipe

Crestwave:


app-misc/fff/fff-0.0~git.recipe
```diff
@@ -0,0 +1,42 @@
+SUMMARY=":rocket: fucking fast file-manager"
```
waddlesplash: I think we should ban explicit profanity from SUMMARY/DESCRIPTION. Haiku's package manager should be a mostly-professional environment language-wise, IMO.

Crestwave: I did think of those, but I couldn't find anything on them, so I just copied it from the package's info (`:rocket:` was in the man page for some reason). I've removed them from the recipe now.

app-misc/fff/fff-0.3.2.recipe
```diff
@@ -0,0 +1,39 @@
...
+	"
+
+PROVIDES="
+	fff = $portVersion
+	cmd:fff
+	"
+REQUIRES="
+	bash
+	cmd:file
+	haiku
```
alaviss: The `*REQUIRES` variable have an "unofficial" sorting convention: `haiku` is always the first entry, after that the package names, then `add_on:`, `app:`, `cmd:` and `lib:`.

app-misc/fff/fff-0.3.2.recipe
```diff
@@ -0,0 +1,39 @@
...
+REVISION="1"
+SOURCE_URI="$HOMEPAGE/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="f4d64abd0adacfdcc3d2e8f8adc80a878c41b982c662a97e8b4f1989f01aea61"
+SOURCE_FILENAME="$portVersionedName.tar.gz"
+PATCHES="fff-0.3.2.patch"
+
+ARCHITECTURES="any"
+
+GLOBAL_WRITABLE_FILES="
+	cache/fff/trash directory
```
alaviss: `GLOBAL_WRITABLE_FILES` only applies to global directories, such as the ones in `/system/var/cache`. As this program write into your user directory instead, this entry should be removed.

Crestwave:
Is the "see the README" part of the `DESCRPTION` inappropriate? If so, would a link be preferred or should I ust omit that line?

alaviss:
Yea, you should remove the `(see the README)` part.

Crestwave:
> Yea, you should remove the `(see the README)` part.

Just to confirm, by that you mean the line with it (`* Automatic CD on exit (see the README)`)?

alaviss:
I think it's OK either way :P

app-misc/fff/fff-1.0.recipe
```diff
@@ -0,0 +1,35 @@
...
+REVISION="1"
+SOURCE_URI="$HOMEPAGE/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="6e5eed18b50b3e1a2497092e618665ff00b71b27b4f7653f9693ee40517d1a8a"
+SOURCE_FILENAME="fff-$portVersion.tar.gz"
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	fff = $portVersion
+	cmd:fff
```
alaviss:
```suggestion
	cmd:fff = $portVersion
```

app-misc/fff/fff-1.0.recipe
```diff
@@ -0,0 +1,35 @@
...
+	bash
+	cmd:file
+	"
+
+INSTALL()
+{
+	install -d "$binDir" "$manDir"/man1 "$docDir"
+	install -m 755 fff "$binDir"
+	install -m 644 fff.1 "$manDir"/man1
+	install -m 644 {CONTRIBUTING,README}.md "$docDir"
```
alaviss:
```suggestion
	install -m 644 README.md "$docDir"
```

The author has removed the `CONTRIBUTING` document right after release, so we should also do the same?

app-misc/fff/fff-1.0.recipe
Crestwave: Should this be done for all `cmd:` lines?

alaviss: Yep, it's a good practice to version everything.

pulkomandy: I don't see much use for that, tbh.
It would be relevant if the command line options were used in scripts and they were renamed a lot, so there is a need to distinguish different versions of a command (eg. iperf 2 vs iperf 3). But in other cases, I think it's a waste of effort to specify versionning.

Crestwave:
> It'd be nice if this program can integrate with Haiku's trash instead of inventing it's own, but I don't think that could be trivially implemented.

Well, I could at least implement an unupstreamable patch to set the trash directory to `B_TRASH_DIRECTORY` and use the `trash` command to delete them if you want.

PR #3479, opened 2018-12-18
icecast, new recipe

Begasus:


Begasus:
builds ok on all architectures

korli:
Shouldn't the architectures be enabled?

PR #3484, opened 2018-12-19
scourge_data, add data files for scourge

Begasus:
Needed for scourge


Begasus:
Based on the original bep files from @michaelvoliveira

PR #3486, opened 2018-12-21
edgar, bump for correct libpng deppendencies

Begasus:


Begasus:
@humdingerb @alaviss could you check?

games-arcade/edgar/edgar-1.29.recipe
```diff
@@ -24,7 +24,7 @@ PROVIDES="
 REQUIRES="
 	haiku
 	lib:libintl
-	lib:libpng16
+	lib:libpng
```
korli: Why not libpng16?

alaviss: The `BUILD_REQURIES` already contains `devel:libpng16`. And since `libpng16` provides `libpng` with correct version constraints, using `lib:libpng` will allow us to just modify one occurance when we decide to bump `libpng` version

korli: IMO this is pointless to use tricks like this. It helps neither newcomers, nor consistency.

pulkomandy: The lib:libpng provide in the libpng package should be removed. The package indeed does not provide that library at all, only libpng16.

Begasus: If the general opinion is that it should be lib:libpng16 I'll bump it back again to use it like that

PR #3491, opened 2018-12-23
Mercurial version 4.8.1 recipe.

SpartanJ:


dev-vcs/mercurial/mercurial-4.8.1.recipe
```diff
@@ -0,0 +1,54 @@
+SUMMARY="Free, distributed source control management tool"
+DESCRIPTION="Mercurial (hg) is a free, distributed source control management \
+tool."
+HOMEPAGE="https://www.mercurial-scm.org/"
+COPYRIGHT="2005-2018 Matt Mackall et al."
+LICENSE="GNU GPL v2"
+REVISION="6"
```
korli: For new versions revision is to be reset to 1

SpartanJ:
Hi, i'm new here, so i'm not sure how i should procede. My question is:
It's a requirement that the package builds with gcc2? Or it's possible to at least flag the recipe only for 64bit haiku that comes with a newer gcc version?
Otherwise i'll take a look if i can fix the issues for gcc2.
PD: The current mercurial version on haiku doesn't work with bitbucket, that's why i made the recipe.
Thanks.

PR #3495, opened 2018-12-23
aspell_en, bump version

Begasus:
Not sure how this is used internally so maybe some checks should be used before merging?


waddlesplash:
I believe the dictionaries are compatible across versions.

PR #3503, opened 2018-12-24
PostgreSQL 11: new recipe

extrowerk:


extrowerk:
Test broken, possibly the LIBRARY_PATH is broken.
Server works was able to create a test db and tables in it.

Begasus:
Builds ok on gcc2h for _x86 (secondaryArch) on primary arch it builds ok without libreadline (missing symbol BC when configure is running a test)

PR #3506, opened 2018-12-25
Fftw: baseline cpu features

extrowerk:
Reference: https://github.com/haikuports/haikuports/commit/bf7f38150cd2131e87f2f07cf9fe1116b294780c


extrowerk:
According to the docs and other ports 32 bit targets Northwood, while 64 bit targets Prescott.

But Prescott is 32 bit processor and gcc complains that it cannot build 64 bit programs for that arch, thus i selected Nocona, which is improved Prescott with 64 bit support. So technically the oldest 64 bit x86_64 consumer arch which is supported by GCC.

Excerpt from the GCC docs:
```
pentium4, pentium4m
Intel Pentium4 CPU with MMX, SSE and SSE2 instruction set support. 
prescott
Improved version of Intel Pentium4 CPU with MMX, SSE, SSE2 and SSE3 instruction set support. 
nocona
Improved version of Intel Pentium4 CPU with 64-bit extensions, MMX, SSE, SSE2 and SSE3 instruction set support. 
```

extrowerk:
Summon @mmlr and @waddlesplash

waddlesplash:
Does this stop fftw from including its own -march? Or is there no way to do that, this just overrides?

sci-libs/fftw/fftw-3.3.8.recipe
```diff
@@ -116,22 +116,24 @@ BUILD()
 	# 32 bit targets NORTHWOOD
 	# 64 bit targets PRESCOTT
 
+	MARCH="pentium4m"
 	CONFIGURE="--enable-shared --enable-threads"
 if [ "$effectiveTargetArchitecture" = x86_64 ]; then
+	MARCH="nocona" # Nocona is Prescott with 64 bit support. Prescott is 32 bit.
```
alaviss:
```suggestion
	MARCH="x86-64"
```

sci-libs/fftw/fftw-3.3.8.recipe
```diff
@@ -116,22 +116,24 @@ BUILD()
 	# 32 bit targets NORTHWOOD
 	# 64 bit targets PRESCOTT
 
+	MARCH="pentium4m"
```
alaviss:
```suggestion
	MARCH="i586"
```

PR #3513, opened 2018-12-28
MPD: version bump, change to meson based build system

extrowerk:
No Haiku native audio output, but openAL works nicely.
All test passes.
Buildsystem needs more love, but it is ok for now.


korli:
Unless you tested x86 secondary architecture, I would leave the previous version enabled for it and disabled for the new version.

waddlesplash:
@Begasus, can you test on x86?

media-sound/mpd/mpd-0.21.3.recipe
```diff
@@ -139,34 +145,48 @@ BUILD_REQUIRES="
 	devel:libz$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
-	cmd:autoconf
+	setuptools_python3
```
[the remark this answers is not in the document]
waddlesplash: No, it's a host tool, it belongs in PREREQUIRES.

media-sound/mpd/mpd-0.21.3.recipe
```diff
@@ -139,34 +145,48 @@ BUILD_REQUIRES="
...
 {
-	make install
+	cd haiku_build
+	ninja install
+
+	# Should be fixed in the build-system
+	mkdir -p $binDir
+	cp mpd $binDir
+	rc -o mpd.rsrc ../src/haiku/mpd.rdef
+	xres -o $binDir/mpd -- mpd.rsrc
```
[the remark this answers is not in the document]
Begasus: This did the trick into building it for _x86 on gcc2h
```diff
-	mkdir -p $binDir
-	cp mpd $binDir
-	rc -o mpd.rsrc ../src/haiku/mpd.rdef
-	xres -o $binDir/mpd -- mpd.rsrc
+	cp mpd $commandBinDir
+	rc -o mpd.rsrc ../src/haiku/mpd.rdef
+	xres -o $commandBinDir/mpd -- mpd.rsrc
+	chmod +x $commandBinDir/mpd
```

media-sound/mpd/mpd-0.21.8.recipe
```diff
@@ -139,34 +144,49 @@ BUILD_REQUIRES="
 	devel:libz$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
-	cmd:autoconf
+	setuptools_python3
 #	cmd:avahi_client$secondaryArchSuffix
-	cmd:awk
-	cmd:find
+	cmd:cmake
```
korli: are you sure about cmake

extrowerk: It definetely searching for it. I was also surprised, but maybe it using cmake in pkg-config mode (it is possible, btw) to search for its deps.

extrowerk:
AFAIK the meson build system doesn't know how to build the haiku output module, so it isn't supported, only openal output usable.

media-sound/mpd/mpd-0.21.8.recipe
korli: or there are third party deps which could require cmake, dunno

korli:
well i see it added:
https://github.com/MusicPlayerDaemon/MPD/blob/master/src/output/plugins/meson.build#L29

Maybe haiku isn't correctly detected:
https://github.com/MusicPlayerDaemon/MPD/blob/master/meson.build#L102

extrowerk:
Afaik meson uses Python, and Haiku declared as "haiku1" in python. I will try to change that line, and we will see, what happens.

extrowerk:
Just changed the meson script to print the osname and it prints just "haiku", so the tests should be ok, but still doesn't builds the haiku plugins. Any idea, whats wrong?

PR #3514, opened 2018-12-29
mozc: Fix build break on x86_gcc2

hanya:


Begasus:
Confirmed ok build on x86_gcc2 for secondary architecture

PR #3522, opened 2018-12-31
tinyxml2, bump version, use cmake for the build

Begasus:
For some reason I haven't been able to figure out, it's using -fPIC to build the library but -fPIE to build xmltest.
on x86_gcc2 (for secondary architecture) this doesn't fail, for x86_64 it fails with -fPIE so that needs to be -fPIC there


waddlesplash:
This version appears to change the SONAME. Can you check what recipes depend on it? Those will all need to be rebuilt at the very least.

Begasus:
So far I can't find anything depending on it? https://github.com/haikuports/haikuports/search?q=lib%3Alibtinyxml2&unscoped_q=lib%3Alibtinyxml2

waddlesplash:
Indeed.

PR #3529, opened 2019-01-01
Algol: bump

extrowerk:
Enabled compiler too


extrowerk:
_x86 tested by @Begasus

PR #3533, opened 2019-01-03
gobject-introspection, new recipe

Begasus:
Second try (sorry for the noise) :)
This is still version 1.56.1 as it relies on glib2 (version 2.57.1 atm in haikudepot for secondary architecture), 1.56.1 is the last available version that is in line with glib2's version (there is no 1.57.1, next one is 1.57.2 which needs glib2 2.57.2). Disabled gcc2 as the available version is even older for glib2.

### DIFF
--- a/dev-libs/gobject-introspection/gobject_introspection-1.56.1.recipe
+++ b/dev-libs/gobject-introspection/gobject_introspection-1.56.1.recipe
@@ -1,0 +1,120 @@
+SUMMARY="Introspection system for GObject-based libraries"
+DESCRIPTION="GObject introspection is a middleware layer between C libraries \
+(using GObject) and language bindings. The C library can be scanned at \
+compile time and generate a metadata file, in addition to the actual native \
+C library. Then at runtime, language bindings can read this metadata and \
+automatically provide bindings to call into the C library."
+HOMEPAGE="https://wiki.gnome.org/Projects/GObjectIntrospection"
+COPYRIGHT="Colin Walters
+	Johan Dahlin
+	Matthias Clasen (girepository)
+	Jürg Billeter (scanner)"
+LICENSE="GNU LGPL v2
+	GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://ftp.gnome.org/pub/GNOME/sources/gobject-introspection/${portVersion%.*}/gobject-introspection-$portVersion.tar.xz"
+CHECKSUM_SHA256="5b2875ccff99ff7baab63a34b67f8c920def240e178ff50add809e267d9ea24b"
+SOURCE_DIR="gobject-introspection-$portVersion"
+PATCHES="gobject_introspection-$portVersion.patchset"
+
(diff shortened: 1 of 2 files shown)


dev-libs/gobject-introspection/gobject_introspection-1.56.1.recipe
```diff
@@ -0,0 +1,120 @@
...
+	cmd:aclocal
+	cmd:autoconf
+	cmd:bison
+	cmd:flex
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python3.6m
```
korli: why not cmd:python3?

dev-libs/gobject-introspection/gobject_introspection-1.56.1.recipe
```diff
@@ -0,0 +1,120 @@
...
+
+defineDebugInfoPackage gobject_introspection$secondaryArchSuffix \
+	"$libDir"/libgirepository-1.0.so.$libVersion
+
+BUILD()
+{
+	runConfigure --omit-dirs binDir ./autogen.sh \
+		--bindir=$commandBinDir --without-cairo \
+		CPPFLAGS=-I"$(finddir B_SYSTEM_HEADERS_DIRECTORY)/python3.6m/" \
+		PYTHON=python3.6m
```
korli: PYTHON=python3 should be enough.

dev-libs/gobject-introspection/gobject_introspection-1.56.1.recipe
Begasus: Made the switch :)

dev-libs/gobject-introspection/gobject_introspection-1.56.1.recipe
```diff
@@ -0,0 +1,120 @@
...
+	"
+
+defineDebugInfoPackage gobject_introspection$secondaryArchSuffix \
+	"$libDir"/libgirepository-1.0.so.$libVersion
+
+BUILD()
+{
+	runConfigure --omit-dirs binDir ./autogen.sh \
+		--bindir=$commandBinDir --without-cairo \
+		CPPFLAGS=-I"$(finddir B_SYSTEM_HEADERS_DIRECTORY)/python3.6m/" \
```
alaviss:
```suggestion
		CPPFLAGS=-I"$portPackageLinksDir/cmd~python3/$relativeIncludeDir/python3.6m/" \
```

Haven't tested, but should work :)

Begasus: Works :)

korli:
It seems to me that the build requires the python headers, but the runtime has no dependency on python. This is weird.

Begasus:
Anything I could do to check into this @korli ?

PR #3534, opened 2019-01-04
edgar, bump version

Begasus:


Begasus:
Been using Edgar without problems, so I'll push these changes

PR #3540, opened 2019-01-05
clazy, new recipe

Begasus:


Begasus:
Initial recipe, could use some looking into and testing :)

dev-util/clazy/clazy-1.4.recipe
```diff
@@ -0,0 +1,58 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libLLVM_5.0$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:clang
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:llvm_config
```
korli: the required version for cmd:llvm_config should be added, otherwise it could end being different from BUILD_REQUIRES one.

Begasus: checking, "cmd:llvm_config >= 5.0" should do it I guess?

korli:
Could you try with LLVM7?

Begasus:
Don't have it installed, but I could try yes :)

korli:
Tried "pkgman install devel:libllvm_7_x86" here and it finds the package.

Begasus:
that seems to work here also, but it should pick it up from the recipe? (like mesa's bump to llvm7)

korli:
just place the llvm7 package in yours haikuports packages directory. haikuporter will pick it up.

Begasus:
probably yes, I installed the needed changes and so far it's compiling ok with llvm7(_clang), as it seems you can't change the dependencies for llvm7 if you have 5.0.0 installed (in the recipe)

Begasus:
Build finished successful with llvm7 :)

Begasus:
Would this work if a user still has llvm 5.0.0 installed? (Or would it updated the existing install from llvm 5.0.0 to 7?)

korli:
llvm7_libs will be installed alongside llvm5_libs.

Begasus:
If it works it's ok, I got some conflicts regarding both versions so I had to manually install newer version from llvm which haikuporter didn't seem to handle

PR #3542, opened 2019-01-05
libaddressinput, Co-authored-by: Shiroko

Begasus:


Begasus:
Passes all tests, not checked on x86_64 (yet)

Begasus:
The download for grit changes the checksum with each download, needs to be looked into ...

alaviss:
I'm thinking if we could just drop this PR. The reasons are:

- This library appears to be designed for bundling. There aren't any `SONAME`.
- This library is not packaged by any linux distribution, which means most (if not all) of its user will be bundling the library (for example: chromium).

Begasus:
Ok, closing this one then :)

PR #1894, opened 2017-12-13
OpenJazz: Add a new recipe

przemub:
It is a recipe for a free re-implementation of the engine of Jazz Jackrabit - a platform game.
It's my first recipe, so I'd be thankful if someone took a careful look on it.

PR #3544, opened 2019-01-05
openjazz, new recipe, Co-Authered by Przemysław Buczkowski

Begasus:
Previous patch has been upstreamed, they've switched to use autotools which makes it easier to build/install (original work here: https://github.com/haikuports/haikuports/pull/1894)


games-action/openjazz/openjazz-20180522.recipe
```diff
@@ -0,0 +1,58 @@
...
+	cmd:autoreconf
+	cmd:awk
+	cmd:gcc
+	cmd:make
+	cmd:pkg_config
+	"
+
+BUILD()
+{
+	export LDFLAGS="-lnetwork -lbe"
```
carstene1ns: `$LDFLAGS` can be omitted when I merge the [buildsystem branch](https://github.com/AlisterT/openjazz/blob/buildsystem/configure.ac#L17) and do the next release.

games-action/openjazz/patches/openjazz-20180522.patchset
```diff
@@ -0,0 +1,24 @@
...
+index 7a888f7..76047cd 100644
+--- a/src/io/network.cpp
++++ b/src/io/network.cpp
+@@ -59,7 +59,9 @@
+ #elif defined USE_SDL_NET
+ 	#include <arpa/inet.h>
+ #endif
+-
++#ifdef __HAIKU__
++	#include <sys/select.h>
```
carstene1ns: Any idea why it worked before without this?
Can also upstream this patch.

Begasus: I don't know to be honest, just came across this when I was looking over some of the older PR's here, was thinking about upstreaming it but this would even be better, thanks! :+1:

Begasus:
PS as I don't have the data files I wasn't able to see if it actually works :) But it looks to be ok and the mentioning of it not finding the data in Terminal is correct then.

games-action/openjazz/patches/openjazz-20180522.patchset
Begasus: If you have upstreamed the patches could you give a notice @carstene1ns ? I'll do a git checkout to look into it before the next release with them included then :)

carstene1ns:
About data files: You can get the shareware episode and christmas episodes from a variety of sites. OJ will work with them.

Begasus:
Thanks! Just checked with one of the downloads there, works :)

carstene1ns:
Currently preparing a patch release to sort out the outstanding issues.

games-action/openjazz/patches/openjazz-20180522.patchset
alaviss: We do have [`<sys/select.h>`](http://xref.plausible.coop/source/xref/haiku/headers/posix/sys/select.h). What error did you get with that `#include`?

pulkomandy: Often other os will include sys/select.h from another .h file, where we don't. So in Haiku it needs to be included explicitly if you use it.

@alaviss: this is a #ifdef not #ifndef!

Begasus:
> Currently preparing a patch release to sort out the outstanding issues.
> 
> Edit:
> 
> ```
> 91341adcc4908db12aad6b82d2fb0125429a26585f65d7eb32d403656313eaab  openjazz-20190106.tar.xz
> ```

Looking into it, also trying to bump the icon to use the one from openjazz :)

PR #3555, opened 2019-01-11
simutrans: fix build for gcc2hybrid

taosxx:
I forgot to remove the "$secondaryArchSuffix" for the corresponding pakset in the REQUIRES section after switching to platform agnostic "any" paksets for simutrans release recipes. This prevented the buildbots from compiling packages for gcc2hybrid. Nightlies are not affected because they still need to compile platform-specific paksets from source code.


waddlesplash:
Thanks!

PR #3556, opened 2019-01-12
peaclock: new recipe

Crestwave:
I'm not sure if I got everything related to the secondary architecture correctly; how are you supposed to build it with HaikuPorter? Also, should we use British or American spellings (or just use the one that the package does) for the `SUMMARY` and `DESCRIPTION`?

### DIFF
--- a/app-misc/peaclock/peaclock-0.1.3.recipe
+++ b/app-misc/peaclock/peaclock-0.1.3.recipe
@@ -1,0 +1,58 @@
+SUMMARY="A colourful binary clock for the terminal"
+DESCRIPTION="Features:
+* digital clock
+* binary clock
+* 12/24 hour time
+* customize with hex colour codes
+* set a custom unicode character for the binary clock graphic
+* compact or expanded mode
+* toggle visibility of digital and binary clocks
+* command prompt with a readline-like interface"
+HOMEPAGE="https://github.com/octobanana/peaclock"
+COPYRIGHT="2019 Brett Robinson"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="$HOMEPAGE/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="0041de0e50429786dff4152c60f6dbd84249287d289ad6d7573e47c9f9b62afc"
+SOURCE_FILENAME="peaclock-$portVersion.tar.gz"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+SECONDARY_ARCHITECTURES="x86"
+commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	peaclock$secondaryArchSuffix = $portVersion
+	cmd:peaclock
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage peaclock$secondaryArchSuffix \
+	"$commandBinDir"/peaclock
+
+BUILD()
+{
+	mkdir -p build
+	cd build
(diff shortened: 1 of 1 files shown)


Begasus:
At first sight it looks ok, if you want to check the build for secondaryArch you just add _x86 to your name like this "peaclock_x86"

Crestwave:
@Begasus I already tried that, but I've got it now; I didn't uncomment `SECONDARY_TARGET_ARCHITECTURES`. Thanks for your help!

app-misc/peaclock/peaclock-0.1.3.recipe
```diff
@@ -0,0 +1,56 @@
...
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+commandBinDir=$binDir
+if [[ $targetArchitecture == x86_gcc2 ]]; then
+SECONDARY_ARCHITECTURES="x86"
+commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	$portName = $portVersion
+	cmd:$portName
```
Begasus: This didn't work for me (got a policy error for no matching provides for /bin/peaclock), changing it to "cmd:peaclock" worked for me

app-misc/peaclock/peaclock-0.1.3.recipe
```diff
@@ -0,0 +1,58 @@
...
+
+defineDebugInfoPackage "$portName" \
+	"$commandBinDir"/peaclock
+
+BUILD()
+{
+	mkdir -p build
+	cd build
+	cmake ../ -DCMAKE_BUILD_TYPE=RelWithDebInfo
+	make
```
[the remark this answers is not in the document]
Crestwave: Done; thanks for the review!

app-misc/peaclock/peaclock-0.1.3.recipe
```diff
@@ -0,0 +1,59 @@
...
+	cd build
+	cmake ../ -DCMAKE_BUILD_TYPE=RelWithDebInfo
+	make $jobArgs
+}
+
+INSTALL()
+{
+	install -d "$commandBinDir" "$docDir"
+	install -m 755 build/peaclock "$commandBinDir"
+	install -m 644 README.md "$docDir"
```
alaviss: This README seems to contain more information about how to build and install the software than using it, so I don't think we should include it.

Crestwave: It does include a nice explanation of how to read it, though. But I guess that people wouldn't install it without knowing how to read it...

app-misc/peaclock/peaclock-0.1.3.recipe
```diff
@@ -0,0 +1,58 @@
...
+COPYRIGHT="2019 Brett Robinson"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="$HOMEPAGE/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="0041de0e50429786dff4152c60f6dbd84249287d289ad6d7573e47c9f9b62afc"
+SOURCE_FILENAME="peaclock-$portVersion.tar.gz"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+commandBinDir=$binDir
+if [[ $targetArchitecture == x86_gcc2 ]]; then
```
alaviss:
```suggestion
if [ "$targetArchitecture" = x86_gcc2 ]; then
```

Currently most of our recipes are using the sh version of this condition, so I think we should stick to that.

Crestwave: Oh, I thought that it would be fine since HaikuPorter seems to require and use Bash. So should only POSIX `sh` features be used in recipes or is this just because they're nearly equivalent?

alaviss: Yes, we use bash, but @fbrosson tries to avoid bash-specific construct and established that as an unofficial convention across HaikuPorts sometimes ago. The reasoning is in some merged PRs long ago IIRC.

Crestwave: Where is the line drawn between when it is acceptable to use Bashisms and not? Is it when POSIX has no equivalent feature? Also, is it fine to go all out with Bashisms when a recipe already uses some?

PR #3560, opened 2019-01-13
capstone: bump version to 4.0.1 and add tests

return:
Tested on x86_64 hrev52725, and passes all normal tests and fuzz tests.


Begasus:
Why not switch to cmake? checked a build in Terminal with gcc7 (on gcc2h), builds out of the box and test run ok without patching

Begasus:
Missing "cmd:cstool", libversion = 4.0.1

dev-util/capstone/capstone-4.0.1.recipe
```diff
@@ -65,3 +71,10 @@ INSTALL()
 	fixPkgconfig
 	packageEntries devel $developDir
 }
+
+TEST()
+{
+	wget ${fuzz_test_url}/raw/master/corpus/${fuzz_test_file}
```
alaviss: Please don't download within the recipe. This will cause it to not be cached and would be redownloaded everytime `TEST()` is run.

If you want to download the code at a specific commit, use https://github.com/groundx/capstonefuzz/archive/<commit hash>.tar.gz

dev-util/capstone/patches/capstone-4.0.1.patchset
```diff
@@ -0,0 +1,80 @@
...
++	stat(dir->d_name, &s);
++	if(!S_ISREG(s.st_mode)){
++	   continue;
++	}
+         //printf("Running %s\n", dir->d_name);
+         fp = fopen(dir->d_name, "rb");
+         if (fp == NULL) {
+-- 
+2.19.1
+
```
Begasus: Maybe this looks a "bit" better and could be upstreamed?
```
diff --git a/suite/fuzz/driverbin.c b/suite/fuzz/driverbin.c
index 57eea2c..65a046c 100644
--- a/suite/fuzz/driverbin.c
+++ b/suite/fuzz/driverbin.c
@@ -3,6 +3,7 @@
 #include <stdio.h>
 #include <dirent.h>
 #include <unistd.h>
+#include <sys/stat.h>
 
 int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size);
 
@@ -13,6 +14,8 @@ int main(int argc, char** argv)
     size_t Size;
     DIR *d;
     struct dirent *dir;
+    struct stat s;
+
     int r = 0;
 
     if (argc != 2) {
@@ -32,9 +35,16 @@ int main(int argc, char** argv)
 
     while((dir = readdir(d)) != NULL) {
         //opens the file, get its size, and reads it into a buffer
+        #ifndef __HAIKU__
         if (dir->d_type != DT_REG) {
             continue;
         }
+        else
+        stat(dir->d_name, &s);
+        if(!S_ISREG(s.st_mode)){
+        continue;
+        }
+        #endif
         //printf("Running %s\n", dir->d_name);
         fp = fopen(dir->d_name, "rb");
         if (fp == NULL) {
-- 
2.19.1```

File: dev-util/capstone/capstone-4.0.1.recipe [hunk 3]
```diff
@@ -45,10 +45,16 @@ BUILD_REQUIRES="
...
 BUILD_PREREQUIRES="
 	cmd:awk
+	cmd:find
 	cmd:gcc$secondaryArchSuffix
 	cmd:make
+	cmd:unzip
+	cmd:wget
 	"
 
+fuzz_test_url="https://github.com/groundx/capstonefuzz"
```
Begasus: This can be removed, next line should be moved into TEST()

dev-util/capstone/capstone-4.0.1.recipe
```diff
@@ -14,11 +14,11 @@ Solaris confirmed).
 * High performance & suitable for malware analysis (capable of handling \
 various X86 malware tricks)."
 HOMEPAGE="http://www.capstone-engine.org"
-COPYRIGHT="2013-2015, COSEINC"
+COPYRIGHT="2013-2018, COSEINC"
 LICENSE="BSD (3-clause)"
 REVISION="1"
 SOURCE_URI="https://github.com/aquynh/capstone/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256="913dd695e7c5a2b972a6f427cb31f2e93677ec1c38f39dda37d18a91c70b6df1"
+CHECKSUM_SHA256="79bbea8dbe466bd7d051e037db5961fdb34f67c9fac5c3471dd105cfb1e05dc7"
```
Begasus: Some suggestions:
```diff
+srcGitRev="b01a30bf595c23081d0acfbcbe36fc8c6d829a71"
+SOURCE_URI_2="https://github.com/groundx/capstonefuzz/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256_2="837be0138cb44384bf1266e76d67947a5be30a1dbd8d8ea532491677c1609b78"
```

dev-util/capstone/capstone-4.0.1.recipe
```diff
@@ -65,3 +71,10 @@ INSTALL()
 	fixPkgconfig
 	packageEntries devel $developDir
 }
+
+TEST()
+{
+	wget ${fuzz_test_url}/raw/master/corpus/${fuzz_test_file}
+	unzip -o -q ${fuzz_test_file} -d suite/fuzz
```
Begasus:
```diff
-	wget ${fuzz_test_url}/raw/master/corpus/${fuzz_test_file}
-	unzip -o -q ${fuzz_test_file} -d suite/fuzz	
+	fuzz_test_file="corpus-libFuzzer-capstone_fuzz_disasmnext-latest.zip"
+	unzip -o -q $sourceDir2/capstonefuzz-$srcGitRev/corpus/${fuzz_test_file} -d suite/fuzz 
```

Begasus:
Add "export INCDIR=$includeDir" in BUILD() and INSTALL() saves the first part of the patchset

dev-util/capstone/capstone-4.0.1.recipe
```diff
@@ -45,10 +45,16 @@ BUILD_REQUIRES="
 	"
 BUILD_PREREQUIRES="
 	cmd:awk
+	cmd:find
```
Begasus: cmd:find and cmd:unzip should be moved to TEST_REQUIRES as they are not needed for the build

dev-util/capstone/capstone-4.0.1.recipe
Begasus: Without the zipfile the test fails with make

dev-util/capstone/patches/capstone-4.0.1.patchset
```diff
@@ -0,0 +1,80 @@
+From 203f796d921948ea60d9941f4a19317b41d66dbe Mon Sep 17 00:00:00 2001
+From: Calvin Hill <calvin@hakobaito.co.uk>
```
korli: The original author should be left as it was.

dev-util/capstone/capstone-4.0.1.recipe
Begasus: Is there a way we could use an "if" "fi" argument for these (so it doesn't unzip on every test or errors out on being already there)?
Or maybe move this into PATCH()?

dev-util/capstone/capstone-4.0.1.recipe
Begasus: Using srcGitRev="commit" and SOURCE_URI_2 will fix this, no need for wget

korli: SOURCE_URI_2 would mean this file is needed to build the project. it is not.

Begasus: Got me on that one, then we need another way to download this just when the test is invoked

return: As discussed on IRC, we have decided to go with downloading the fuzz tests file using SOURCE_URI_2 which is downloaded once on every clean build rather than every-time TEST() is run.

dev-util/capstone/patches/capstone-4.0.1.patchset
return: Applied suggested changes into patchset.

dev-util/capstone/capstone-4.0.1.recipe
return: Done.

dev-util/capstone/capstone-4.0.1.recipe
return: @Begasus I've added a check to see if the folder exists or not to prevent it unzipping every-time TEST() is run.

dev-util/capstone/capstone-4.0.1.recipe
return: Done.

dev-util/capstone/patches/capstone-4.0.1.patchset
return: @korli I've added the equivalent change of the first patch directly into the recipe. This makes the original patchset obselete.

dev-util/capstone/capstone-4.0.1.recipe
return: Done.

dev-util/capstone/patches/capstone-4.0.1.patchset
korli: thanks

dev-util/capstone/capstone-4.0.1.recipe
korli: I don't know whether this is correct: it will be packaged in the _source package (possibly from another license). Test data shouldn't be needed to build this recipe.

Begasus: @korli, would a separate recipe like "capstone_test" do?

Begasus:
So far I only see 3 options for the zip download for the tests?

1. We create a separate recipe for it (not advised I think because it would end up in HaikuDepot)
2. We disable the source package
3. We disable the tests all together

korli:
I don't see a problem when TEST() download a file it doesn't already have. Why would this be a problem?

Begasus:
> I don't see a problem when TEST() download a file it doesn't already have. Why would this be a problem?

Because it was mentioned here https://github.com/haikuports/haikuports/pull/3560#discussion_r247378366 (but should be solved with the check https://github.com/haikuports/haikuports/pull/3560#discussion_r351528534
added by @return ) now?

dev-util/capstone/capstone-4.0.1.recipe
```diff
@@ -14,27 +14,31 @@ Solaris confirmed).
 * High performance & suitable for malware analysis (capable of handling \
 various X86 malware tricks)."
 HOMEPAGE="http://www.capstone-engine.org"
-COPYRIGHT="2013-2015, COSEINC"
+COPYRIGHT="2013-2019, COSEINC"
 LICENSE="BSD (3-clause)"
 REVISION="1"
 SOURCE_URI="https://github.com/aquynh/capstone/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256="913dd695e7c5a2b972a6f427cb31f2e93677ec1c38f39dda37d18a91c70b6df1"
+CHECKSUM_SHA256="79bbea8dbe466bd7d051e037db5961fdb34f67c9fac5c3471dd105cfb1e05dc7"
```
Begasus: Rename source archive:
```
+SOURCE_FILENAME="capstone-$portVersion.tar.gz"
```

return: Done.

korli:
> Because it was mentioned here [#3560 (comment)](https://github.com/haikuports/haikuports/pull/3560#discussion_r247378366) (but should be solved with the check [#3560 (comment)](https://github.com/haikuports/haikuports/pull/3560#discussion_r351528534)
> added by @return ) now?

The check to avoid to download again should be enough indeed.

Begasus:
@return how's the stats on this?

PR #3561, opened 2019-01-13
asciinema: bump to 2.0.2.

Crestwave:


app-misc/asciinema/asciinema-2.0.2.recipe
```diff
@@ -46,11 +45,9 @@ BUILD()
 
 INSTALL()
 {
-	sed -i \
-		-e "s|documentation/man|$relativeManDir|" \
-		-e "s|documentation/packages/asciinema|$relativeDocDir|" \
-		setup.py
-	python3 setup.py install --prefix=$prefix --root=/ --skip-build
+	sed -i "s|share/doc/asciinema|$relativeDocDir|
+		s|share/man|$relativeManDir|" setup.py
```
alaviss: I think this should go into `PATCH()` instead.

app-misc/asciinema/asciinema-2.0.2.recipe
```diff
@@ -46,11 +45,9 @@ BUILD()
...
+	python3 setup.py install --prefix="$prefix" --root=/ --skip-build
 }
 
 TEST()
```
alaviss: Can you remove `TEST()`? I've tried it and got:

    Ran 0 tests in 0.001s

app-misc/asciinema/asciinema-2.0.2.recipe
Crestwave: I placed it there because of [this conversation](https://github.com/haikuports/haikuports/pull/3219#discussion_r225128937), which *seems* to apply here. Relevant quote:

> Move the sed commands from PATCH to INSTALL. That's because, imho, the changes are not a real patch but a requirement for the installation.

He might have meant that because they were *meant* to be replaced, though, but he also let this pass when he reviewed it the previous version of this recipe, as well as in another I submitted. Maybe we should ask him to be sure?

app-misc/asciinema/asciinema-2.0.2.recipe
```diff
@@ -46,14 +45,11 @@ BUILD()
...
-	python3 setup.py install --prefix=$prefix --root=/ --skip-build
-}
+	sed -i "s|share/doc/asciinema|$relativeDocDir|
+		s|share/man|$relativeManDir|" setup.py
+	python3 setup.py install --prefix="$prefix" --root=/ --skip-build
 
-TEST()
-{
-	python3 setup.py test
+	packageEntries $pythonPackage \
```
alaviss: But there aren't any `$pythonPackage` in sight? I don't think this is needed.

app-misc/asciinema/asciinema-2.0.2.recipe
alaviss: This `sed` is applied directly into the source, meaning that it will work once and only once, so it shouldn't be in `INSTALL()`. Also, `PATCH()` is un-upstreamable by design as factors like cross-platform support can't usually be expressed there.

We can ask @fbrosson of course, but he's MIA atm. If you have a good idea on why something shouldn't be like that, you can always convince him :)

app-misc/asciinema/asciinema-2.0.2.recipe
Crestwave: Ah, I forgot why I didn't include it in the first place and assumed that I forgot to. I'll fix it as soon as I get back to my computer.

app-misc/asciinema/asciinema-2.0.2.recipe
Crestwave: My bad; just disregard what I said earlier, please. I was confused.

app-misc/asciinema/asciinema-2.0.2.recipe
```diff
@@ -38,6 +37,12 @@ TEST_REQUIRES="
 	nose_python3
```
alaviss: As `TEST()` is removed, this should be removed as well.

app-misc/asciinema/asciinema-2.0.2.recipe
```diff
@@ -7,12 +7,15 @@ COPYRIGHT="2011-2018 Marcin Kulik"
...
 SOURCE_URI="https://github.com/asciinema/asciinema/archive/v$portVersion.tar.gz"
-CHECKSUM_SHA256="7087b247dae36d04821197bc14ebd4248049592b299c9878d8953c025ac802e4"
+CHECKSUM_SHA256="2578a1b5611e5375771ef6582a6533ef8d40cdbed1ba1c87786fd23af625ab68"
 SOURCE_FILENAME="asciinema-$portVersion.tar.gz"
-PATCHES="asciinema-$portVersion.patchset"
 
 ARCHITECTURES="any"
 
+USER_SETTINGS_FILES="
+	settings/asciinema
```
alaviss:
```suggestion
	settings/asciinema directory
	settings/asciinema/config
```

PR #3562, opened 2019-01-13
SystemInfo, bump after upstream patch merge

Begasus:


Begasus:
https://github.com/haikuports/haikuports/pull/3562 :)

Begasus:
Builds ok on gcc2h (gcc2) and x86_64, only thing I got is the OS being in the way with the indication bar if you run it on R1/Beta1

PR #3572, opened 2019-01-16
uriparser, switch to cmake, bump version

Begasus:


Begasus:
Not tested on x86_64 yet, new git checkout, no soname/api changes

Begasus:
Would it be ok to drop the support for gcc2? (If I check haikuports I only see boca using it "secondaryarchSuffix"), if it's not really needed I could drop support for it, if things need to be simplified I could drop the use for the test and documentation to create a less complicated recipe.

dev-libs/uriparser/uriparser-0.9.1~20190116.recipe
```diff
@@ -42,25 +49,53 @@ REQUIRES_devel="
...
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
 BUILD()
 {
-	autoreconf
-	runConfigure ./configure --disable-test --disable-doc
+	cmake -DCMAKE_INSTALL_PREFIX=$prefix \
+		-DCMAKE_INSTALL_BINDIR:PATH=$binDir \
+		-DCMAKE_INSTALL_INCLUDEDIR:PATH=$includeDir \
+		-DCMAKE_INSTALL_LIBDIR:PATH=$libDir \
+		-DCMAKE_INSTALL_DOCDIR:PATH=$docDir
```
waddlesplash: Try using `$cmakeDirArgs` instead of all this, that's what it's for.

Begasus: I'll be sure to try it, not a big fan for cmake and seen it passing by, but never had much luck in using it :)

Begasus:
> This can be postponed until the official release arrives.

As we discussed on IRC :)

dev-libs/uriparser/uriparser-0.9.1~20190116.recipe
Begasus: Working, thanks! :)

dev-libs/uriparser/uriparser-0.9.1~20190120.recipe
```diff
@@ -42,28 +49,54 @@ REQUIRES_devel="
...
 	"
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+BUILD_PREREQUIRES="$BUILD_PREREQUIRES
+	cmd:dot
+	cmd:doxygen
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:qhelpgenerator$secondaryArchSuffix
+	"
+fi
+
```
alaviss:
```suggestion
```

dev-libs/uriparser/uriparser-0.9.1~20190120.recipe
```diff
@@ -42,28 +49,54 @@ REQUIRES_devel="
...
 
 defineDebugInfoPackage uriparser$secondaryArchSuffix \
 	"$libDir"/liburiparser.so.$libVersion
 
+if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
```
alaviss: Instead of doing like so, we can use a simpler variant:

```sh
cmakeArgs=( $cmakeDirArgs )
  cmakeArgs+=( -DURIPARSER_BUILD_TESTS=OFF -DURIPARSER_BUILD_DOCS=OFF )
fi

cmake .. "${cmakeArgs[@]}"

Begasus: Maybe @hartwork could take a look here, perhaps a check for the compiler could be added to switch the option to OFF in the CMakeLists.txt file? but yes looks ok (not tested yet)

hartwork: GCC 2.x is EOL outside of Haiku so I welcome that you applied GCC 2.x specifics on Haiku package level now

pulkomandy: Or not at all... Is there an existing gcc2 app that depends on it and cannot be easily migrated to gcc7?

Begasus: Searching haikuports I see only boca using it: https://github.com/haikuports/haikuports/search?q=lib%3Aliburiparser&unscoped_q=lib%3Aliburiparser
Not sure if Haiku itself is using it?

Begasus: If not we could remove the gcc2  existence :)

Begasus:
@hartwork thanks for importing the patch :) I think this one should be ok, disabled gcc2 build (nothing relies on it)

Begasus:
@enzo1982 could you check this? A local build for boca_x86 seems to be ok here (with this one)

enzo1982:
All looks good!
Built uriparser_x86, tests passed. Built boca_x86 with it without any issues.

PR #3577, opened 2019-01-17
Python 2.7.15 x86 arch support.

SpartanJ:
I'll need this to build another package. Hope it's OK.
I'm not sure if i need to bump the revision number, since nothing change for other archs. Should i bump it to force the rebuild?


SpartanJ:
Mmh sorry. I'll close it since it breaks the other python installation. I'll take a look and try to fix it.

PR #3579, opened 2019-01-19
gsoap, new recipe

Begasus:


Begasus:
Set ?68 secondaryarchitecture as untested as one file is blocking the build, soapcpp2_yacc.h gets generated at build time and Haiku can't seem to find it on first run, on second run things work out ok

PR #3580, opened 2019-01-20
openoriginpackage: update to v1.1.0

humdingerb:
Very strange:  
"make" creates the binary "`Open Origin Package`", but in the package, the binary is named "`Open\ Origin\ Package`". The build process hasn't changed since v1.0.0 and the package in HaikuDepot shows the correct name.

@waddlesplash: Can this change of the makefile_engine be the issue?
https://git.haiku-os.org/haiku/commit/?id=f84c53e4acb4c8a9e4dadd4ce165ca3a126114b1


humdingerb:
After reverting the change to the makefile_engine the naming issue disappeared. Let's merge.
This wouldn't have been an issue probably anyway, because the buidbots usually are not that up-to-date and wouldn't have used the erroneous makefile_engine. At least I hope so, we'll see which makefile_engine they are currently using soon...

PR #3589, opened 2019-01-27
Netsurf: fix packaging, enable some features

extrowerk:


extrowerk:
We always had the res folder, but till this changes it was in /boot/system/apps/boot/apps/netsurf/res and i hated to see a boot folder in the apps folder, so i corrected it. No idea if the resources are actually required or not.

pulkomandy:
I removed the need for the folder some time ago (probably netsurf 3.6 or so). And it was installed at the wrong place so the app wouldn't find it anyway. Just remove it, if NetSurf still manages to show readable text in the menu and a homepage (both are stored in resources), you're good to go.

PR #3592, opened 2019-01-29
fff: bump to 2.0.

Crestwave:


app-misc/fff/patches/fff-2.0.patch
```diff
@@ -0,0 +1,25 @@
+--- a/fff	2019-01-28 21:36:41.029622272 +0000
++++ b/fff	2019-01-29 11:10:31.277348352 +0000
+@@ -14,6 +14,10 @@
+ 
+         haiku)
+             opener="open"
++            [[ -z $FFF_TRASH_CMD ]] && {
++                FFF_TRASH_CMD="trash"
++                FFF_TRASH="$(finddir B_TRASH_DIRECTORY)"
```
alaviss: Note that unlike `fff`'s trash, Haiku's trash is per-volume, so you'd need to pass the `-v` parameter to `finddir` for the correct trash folder.

Crestwave: I haven't tested it as I think I only have one volume, but I've implemented it.

PR #3601, opened 2019-02-03
aquaskk: new recipe

hanya:


waddlesplash:
Looks good to me.

So there will now be 3 Japanese input methods in HaikuPorts: Canna, Mozc, and this one. I think we currently ship Canna with release builds of Haiku (or maybe not? looks like we forgot to do that for R1/beta1). But I guess that one is pretty old and hasn't seen much attention lately.

Do you have any suggestions for which one we should bundle with Haiku by default?

app-i18n/aquaskk/aquaskk-4.7.0.0.recipe
```diff
@@ -0,0 +1,111 @@
...
+	$portPackageLinksDir/cmd~python2/bin/python2 build.py gyp \
+		--platform=Haiku --arch=$targetArchitecture
+
+	$portPackageLinksDir/cmd~python2/bin/python2 build.py build \
+		-c Release \
+		method/method.gyp:aquaskk \
+		preference/preference.gyp:preference
+}
+
+INSTALL()
```
extrowerk: Maybe you should provide a makefile as additional file instead this.

app-i18n/aquaskk/aquaskk-4.7.0.0.recipe
```diff
@@ -0,0 +1,111 @@
...
+
+	cp $AOutDir/aquaskk $InputMethodDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	cp $AOutDir/aquaskk_task $DataDir
+fi
+	cp $AOutDir/aquaskk_preference $DataDir
+
+	cp $APlatDir/src/preference/dict_downloader.py $DataDir
+
+	cp $AConfigDir/act.conf $DataDir
```
extrowerk: You can write

cp $AConfigDir/{act.conf,act.rule} $DataDir

So you can copy more than one file from the same source folder to the same target folder.

app-i18n/aquaskk/aquaskk-4.7.0.0.recipe
waddlesplash: Huh? Gyp works fine, there's no need for that.

hanya:
Last fix contains crash while uninstall package.

hanya:
> Do you have any suggestions for which one we should bundle with Haiku by default?

SKK way to input is not normal. AquaSKK should not be default one.

Issue #3594, opened 2019-01-31
x86_gcc2 build failures


waddlesplash:
I started an "everything" build on x86_gcc2 and there are quite a lot of build failures: https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=2597&viewMode=expanded

So far (list incomplete):
 - [x] advancemame_x86
 - [x] aobook_x86
 - [x] aplayer
 - [x] atftp
 - [x] autoconf264
 - [x] backup
 - [x] berdp
 - [x] besol_x86
 - [x] bladeenc
 - [x] bmake
 - [ ] bs2b (unsafe source)
 - [ ] bs2b_x86 (unsafe source)
 - [x] bsnow
 - [x] flatzebra_x86
 - [x] cal3d
 - [x] cantata_x86
 - [x] im_x86
 - [x] cdplayer
 - [x] cdplayer_x86
 - [x] cloog_x86 PR created, probably obsolete, so removing cloog if no one steps up
 - [x] corsix_th_x86
 - [ ] coveredcalc (unsafe source)
 - [x] cppunit
 - [x] cronie
 - [x] csound_x86
 - [x] cyberdogs_sdl
 - [x] cython_python3_x86
 - [x] distcc
 - [x] dragonmemory
 - [x] drive_encryption (patch send upstream)
 - [x] e2fsprogs_x86
 - [x] eduke32_x86
 - [x] visualboyadvance (unsafe source)

hanya:
bladeenc: tar archive contains file named 'bladeenc-0.94.2/other_makefiles/BladeEnc.\xb6'
It gives 'utf8' codec can't decode byte 0xb6 in position 41: invalid start byte, error while unpacking.
Not possible to build with current haikuporter.

Begasus:
bladeenc was removed/rejected a few days back of being obsolete ... old sources (but put them on my github account without the offending file (in a commit), also with the latest patches available with credits to the people behind it, it's there for historical reason, no need to hunt if further as it is now gone from haikuports

Begasus:
cantata_x86 probably fixed with the latest release (checked the build and it went further then the log viewer indicates, system froze during build (current laptop isn't up for the build)

kenmays:
Receipts that **didn't build successfully on x86_64 R1B1** buildbot platform yet:
- bmake-20180919
- cloog-0.19.0
- coveredcalc-1.10.0
- drive_encryption-1.0
[...]

PR #3614, opened 2019-02-05
distcc: fix for x86_gcc2

hanya:
Part of #3594.


pulkomandy:
Thanks for restoring my patches, however don't hesitate to disable gcc2 support if that's the simpler solution, especially for non-libraries packages.

PR #3619, opened 2019-02-08
NetHack: bump

extrowerk:


games-roguelike/nethack/nethack-3.6.1.recipe
```diff
@@ -6,8 +6,8 @@ HOMEPAGE="http://www.nethack.org/"
 COPYRIGHT="1985-2015 Stichting Mathematisch Centrum and M. Stephenson"
 LICENSE="nethack"
 REVISION="1"
-SOURCE_URI="http://sourceforge.net/projects/nethack/files/nethack/3.6.0/nethack-360-src.tgz"
-CHECKSUM_SHA256="1ade698d8458b8d87a4721444cb73f178c74ed1b6fde537c12000f8edf2cb18a"
+SOURCE_URI="https://www.nethack.org/download/3.6.1/nethack-361-src.tgz"
```
Begasus: I think it would be possible to use $portVersion instead of "3.6.1" here, and some magic to use "361"?

PR #3631, opened 2019-02-11
Readline, Bash: bump, Dependant recipes: bump

extrowerk:
Not all of the dependant package checked. Postgresql known fails.


diversys:
Have you checked if aliases defined in bash profile still work? I tried a month ago to bump bash to 5.0 and my aliases stopped working. I didn't investigate any further though.

extrowerk:
@diversys: eeehhh..

PR #3635, opened 2019-02-13
mozc: bump version

hanya:


hanya:
Contains fix crash while uninstall the package.

PR #3641, opened 2019-02-17
distcc, bump version, disable x86_gcc2

Begasus:


diversys:
Ouch! Didn't notice wip-no-merge label. Should it be reverted?

Begasus:
I guess so, the test aren't running (have been looking into the declaration of the line in TEST() but haven't found a solution yet)

Issue #3647, opened 2019-02-18
DragonMemory doesn't exit, after a few levels squeres are out of the window ...


Begasus:
![dragon-memory](https://user-images.githubusercontent.com/16057090/52972317-cc896200-33ba-11e9-9604-c1dfe730db22.png)

PR #3648, opened 2019-02-18
dragonmemory, fix recipe, disable build

Begasus:


Begasus:
Related to https://github.com/haikuports/haikuports/issues/3647 and https://github.com/haikuports/haikuports/issues/3594

Issue #3421, opened 2018-11-29
Update Python to 3.7


victordomingos:
Python 3.7 was releases almost half an year ago (we are already at Python 3.7.1 right now). Among other useful features, it has gone through a considerable optimisation work, which allows to get much better performance on most platforms.

Somebody should update the Python recipe(s) to the latest version and, if possible, make sure they're available on both 32 and 64 bits.

korli:
You should probably volunteer :)

victordomingos:
I would do it if I was knowledgeable about it, of course. Sorry.

Issue #3076, opened 2018-09-18
python3: Interactive mode creates ~/.python_history


alaviss:
This file should be moved elsewhere, maybe `B_USER_VAR_DIRECTORY/python/history`

Issue #3075, opened 2018-09-18
python: User base directory is ~/.local


alaviss:
```
$ python -c 'import site; site.getuserbase()'
/boot/home/.local
$ python3 -c 'import site; site.getuserbase()'
/boot/home/.local
```

This should be moved to `B_USER_NONPACKAGED_DIRECTORY`

PR #3664, opened 2019-02-24
python3: add recipe for 3.7.2

hanya:


hanya:
* The setup.py from Python 3.7 can find OpenSSL 1.0.2 (or 1.1 is required) without a hack.
* libffi is no longer shipped in the source package, no need to touch.
* lzma module should work.

hanya:
Related to #3421.

korli:
I would also rename the 3.6 recipe python36-3.6.7 and adjust the provides.

hanya:
Added patch for:
* #3076: ~/.python_history file for interactive console place in B_USER_VAR_DIRECTORY/python/history
* #3075: change site.USER_BASE from ~/.local to B_USER_NONPACKAGED_DIRECTORY

hanya:
Current struct spwd has sp_lstchg member in shadow.h, so we do not omit it.
Skipping sp_lstchg without hacking field list broke return value of spwd.getspnam function.

Still test_spwd fails because spwd.getspnam with invalid user name gives FileNotFoundError. It should be KeyError. It looks error mapping between Haiku error and Python is not match or incomatibility of the Haiku internal.

extrowerk:
Nice, this will be required for the upcoming Blender release.

PR #3671, opened 2019-03-03
fltk, cleanup, cosmetic changes, add defineDebugInfoPackage

Begasus:


Begasus:
Not sure the extra libs are required in REQUIRES_devel, fltk-config --cflags/--libs don't mention them

x11-libs/fltk/fltk-1.3~git.recipe
```diff
@@ -101,19 +104,16 @@ BUILD()
...
-	mv -f $libDir/*.a $developLibDir/
+		libfltk libfltk_forms libfltk_images
+	fixPkgconfig
 
 	# TODO: fix configure.ac
-	mkdir -p $developDocDir
-	mv -f $dataDir/doc/fltk/* $developDocDir/
+	mkdir -p $docDir
+	mv -f $dataDir/doc/fltk/* $docDir/
+	rm -rf $dataDir/doc
```
Begasus: Still left with an empty $docDir/packages in the base package here (that has been moved to the devel package)

mmuman:
The libs are required by fluid itself so we must keep them.

x11-libs/fltk/fltk-1.3~git.recipe
mmuman: Yeah I tried to clean those up but…

Begasus:
> The libs are required by fluid itself so we must keep them.

Hence why they are declared in REQUIRES="", when checking and leaving them out of REQUIRES_devel fluid runs as expected

mmuman:
Well if we keep fluid in the devel package we should have its deps as well.

Begasus:
> The libs are required by fluid itself so we must keep them.

Hence why they are declared in REQUIRES="", when checking and leaving them out of REQUIRES_devel fluid runs as expected

> Well if we keep fluid in the devel package we should have its deps as well.

Probably right there (didn't check), but why keep the binary in the devel package anyway? imho it shouldn't be there, the flt-config file, headers and needed libs should go there yes

khallebal:
I think it makes perfectly sense to put fluid in the devel package as it is a development tool, a non developer user has no use for it in the base package.

Begasus:
Ok, I can live with that, I only think the binary (application) is something that should be provided in the base package, otherwise the base package only supplies the library (needed by?)
Most development packages (Paladin/BeBuilder - and probably others) are supplied in the base package, the devel packages should provide libraries/headers that could be used by other packages - if they require them) ... just my 2 cents, if not this could be closed without merging

pulkomandy:
The goal is to keep the non-devel package as small as possible and limit its dependencies, so that users can have a light system, even if developers have to install more stuff.

Begasus:
Thanks for clearing this up @pulkomandy I guess this could be closed then :)

Begasus:
Well I could alter the PR and leave the binary in the development package, could still strip some debug info of the libraries (the binary doesn't can't be stripped if it's inside the devel package), I think some changes would still be valid enough to import ...

mmuman:
Yeah, it's ok to leave fluid unstripped if it's in the devel package I think.

Begasus:
Moved the binary back to the devel package, also included a small patch for the installation path of the examples, think this should do it

mmuman:
Thanks I'll have a look tonight.

x11-libs/fltk/fltk-1.3~git.recipe
```diff
@@ -13,43 +13,41 @@ HOMEPAGE="https://www.fltk.org/"
 COPYRIGHT="1998-2019 by Bill Spitzak and others."
 LICENSE="FLTK
 	GNU LGPL v2"
-REVISION="1"
-#SOURCE_URI="git+https://github.com/mmuman/fltk.git#haiku-1.3-final"
+REVISION="2"
 SOURCE_URI="https://github.com/mmuman/fltk/archive/haiku-1.3-final.zip"
```
alaviss:
```suggestion
SOURCE_URI="https://github.com/mmuman/fltk/archive/haiku-1.3-final.tar.gz"
```

x11-libs/fltk/fltk-1.3~git.recipe
```diff
@@ -127,6 +122,7 @@ INSTALL()
 	packageEntries devel \
 		$developDir \
 		$binDir \
+		$docDir \
```
alaviss: I have a feeling that these should better be in `developDocDir`

Begasus: Moved them to $developDocDir, guess this makes the patch obsolete?

Begasus:
@mmuman & @alaviss could you check this? If it's ok it could be merged, if not it could be closed?

Begasus:
In case of the binary, maybe move that to a separate tools package (would also reduce the devel package)?

mmuman:
The goal of the devel package is to provide everything needed for development. While not everyone will want to use the IDE (I suppose most people will just need it to build existing apps), making a separate package just for it is overkill IMO. Plus we know devs have bigger drives anyway :D

Begasus:
Fair enough :)

Begasus:
Think this should do it :) reviewers welcome

x11-libs/fltk/fltk-1.3~git.recipe
Begasus: fixed :)

x11-libs/fltk/fltk-1.3~git.recipe
Begasus: Patch still needed, but it's ok now :)

x11-libs/fltk/fltk-1.3~git.recipe
```diff
@@ -13,45 +13,47 @@ HOMEPAGE="https://www.fltk.org/"
...
+
+libVersion="1.3"
+
 PROVIDES="
-	fltk$secondaryArchSuffix = $portVersion compat >= 1.3
-	lib:libfltk$secondaryArchSuffix = $portVersion compat >= 1.3
-	lib:libfltk_forms$secondaryArchSuffix = $portVersion compat >= 1.3
-	lib:libfltk_images$secondaryArchSuffix = $portVersion compat >= 1.3
+	fltk$secondaryArchSuffix = $portVersion
+	lib:libfltk$secondaryArchSuffix = $libVersion
```
Begasus: Not sure here, there is only **libfltk.so.1.3** (**libfltk.so** links to that), so I presume we don't need compat >= ?

Begasus:
Me pokes @mmuman :)

PR #3677, opened 2019-03-07
Quazip : new recipe

extrowerk:


korli:
The build failed: https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/2846/builds/13018.log

Maybe a patch was needed?

PR #3679, opened 2019-03-08
vala, bump version

Begasus:


Begasus:
@korli could you check? If it's ok I'll remove the previous version to reduce the noice on github :)

Begasus:
Not sure if the issue on haikuporter is still in the way (https://github.com/haikuports/haikuporter/issues/183) but at this moment I think the recipe is ok

dev-lang/vala/vala-0.44.1.recipe
```diff
@@ -0,0 +1,123 @@
...
+	haiku$secondaryArchSuffix
+	lib:libcdt$secondaryArchSuffix
+	lib:libcgraph$secondaryArchSuffix
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libgmodule_2.0$secondaryArchSuffix
+	lib:libgobject_2.0$secondaryArchSuffix
+	lib:libgvc$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	# seems like vala needs a vala version to be installed
+	lib:libvalaccodegen$secondaryArchSuffix >= 0.42
```
korli: This makes no sense. When vala needs vala to build, then it's in BUILD_REQUIRES, and probably devel:libvalaccodegen$secondaryArchSuffix

dev-lang/vala/vala-0.42.3.recipe
```diff
@@ -40,6 +40,7 @@ PROVIDES="
 	cmd:vapigen_$portVers$commandSuffix
 	lib:libvala_$portVers$secondaryArchSuffix = $libVersionCompat
 	lib:libvaladoc_$portVers$secondaryArchSuffix = $libVersionCompat
+	lib:libvalacodegen$secondaryArchSuffix = $portVersion
```
korli: What is this? Haikuporter has built this recipe, and hasn't found this lib as provided:
https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/2486/builds/12519.log

dev-lang/vala/vala-0.44.1.recipe
Begasus: Will look further into it, this comes from their homepage in the requirements: Vala 0.16.1 or later (for the initial build)

dev-lang/vala/vala-0.42.3.recipe
Begasus: There should be at least something providing libvalacodegen, this is working atm but maybe should be changed into devel: instead of lib:?

Begasus:
Reverted the changes, at this moment I'm still left with "Warning: POLICY ERROR: can't find used library "libvalaccodegen.so"" when running haikuporter in strict mode

dev-lang/vala/vala-0.44.1.recipe
Begasus: Removed ...

Begasus:
Bumped to 0.47.1, packages build OK if you don't run it in strict mode with haikuporter, issue for that is created at haikuporter: https://github.com/haikuports/haikuporter/issues/183

korli:
haikuporter normally follows the RPATH configured to check dependent libs.

Begasus:
Don't know what else I can do here, merging it would probably end up in a broken build if the buildbot runs haikuporter in strict mode

korli:
The buildbot doesn't run haikuporter in strict mode.

Begasus:
Then this should be ok as is :)

> The buildbot doesn't run haikuporter in strict mode.

PR #3694, opened 2019-03-17
NumPy: rework, bump, cleanup, extra libs

extrowerk:
Refer: https://discuss.haiku-os.org/t/package-python3-numpy-not-working-on-nightly-build-hrev52992/8380

Import works now, but both version (python2 and 3) fails with missing symbols:
* Python2: cblas_call_sub
* Python3: cblas_sgemm

Both symbol defined in `libopenblas-r0.3.5.so`:
```
/s/lib » objdump -x libopenblasp-r0.3.5.so| ag cblas_sgemm
00000000000b07e0 g     F .text  00000000000006d1              cblas_sgemm
/s/lib » objdump -x libopenblasp-r0.3.5.so| ag cblas_cdotc_sub
00000000000c13b0 g     F .text  00000000000000b2              cblas_cdotc_sub
```
Disabling blas results Python crash at import.


extrowerk:
@Begasus can you pls check blis on x86?

hanya:
How did you obtain the missing symbol error? I tried to run ```numpy.test()``` but no such error happened.
But I needed to patch to avoid invalid opcode (xgetbv, GCC 7.3.0 problem) while import the module.

extrowerk:
I'll try to add your patch and retest it.

extrowerk:
Now it works ok, at least no problem at importing it.
I would say, it would be better to merge mine, as it is uniformized, one recipe for both python version, but i'm open for other ideas.

extrowerk:
User reports: working.
https://discuss.haiku-os.org/t/package-python3-numpy-not-working-on-nightly-build-hrev52992/8380/6?u=extrowerk

dev-python/numpy/numpy-1.16.2.recipe
```diff
@@ -0,0 +1,133 @@
...
+	"
+
+PYTHON_PACKAGES=(python python3)
+PYTHON_VERSIONS=(2.7 3.6)
+for i in "${!PYTHON_PACKAGES[@]}"; do
+pythonPackage=${PYTHON_PACKAGES[i]}
+pythonVersion=${PYTHON_VERSIONS[$i]}
+eval "PROVIDES_${pythonPackage}=\"\
+	${portName}_$pythonPackage = $portVersion\n\
+	cmd:f2py\n\
```
alaviss: Wouldn't this mean that the two python versions will conflict?

dev-python/numpy/numpy-1.16.2.recipe
```diff
@@ -0,0 +1,133 @@
...
+	python_numpy
+	"
+
+INSTALL()
+{
+# Check at every update in site.cfg.example!
+cat > site.cfg << EOF
+[ALL]
+libraries = execinfo,lapack,openblas
+library_dirs = /system/`echo ${relativeDevelopLibDir}`
```
alaviss:
```suggestion
library_dirs = /system/$relativeDevelopLibDir
```

It'd be shorter to just use the variables directly :P

I'm not a fan of hardcoding, but it seems to be rather complex to patch this up using `findpaths`, so it's fine for this case.

dev-python/numpy/numpy-1.16.2.recipe
extrowerk: True

dev-python/numpy/numpy-1.16.2.recipe
alaviss: Can you do the same with all other variables?

dev-python/numpy/numpy-1.16.2.recipe
```diff
@@ -0,0 +1,133 @@
...
+HOMEPAGE="http://numpy.scipy.org/"
+COPYRIGHT="2005-2019 Travis E. Oliphant et al."
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/numpy/numpy/releases/download/v$portVersion/numpy-$portVersion.tar.gz"
+CHECKSUM_SHA256="8088221e6e27da8d5907729f0bfe798f526836f22cc59ae83a0f867e67416a3e"
+SOURCE_DIR="numpy-$portVersion"
+PATCHES="numpy-$portVersion.patch"
+
+ARCHITECTURES="any"
```
korli: This shouldn't be any. $secondaryArchSuffix will only work with a secondary architecture defined anyway.

dev-python/numpy/numpy-1.16.2.recipe
```diff
@@ -0,0 +1,133 @@
...
+
+PROVIDES="
+	$portName = $portVersion
```
Begasus: Why use $portName here?

dev-python/numpy/numpy-1.16.2.recipe
Begasus: blis can't be build for x86_gcc2, so agreed, any shouldn't be used here

dev-python/numpy/numpy-1.16.2.recipe
extrowerk: The recipe template contains it.

dev-python/numpy/numpy-1.16.2.recipe
```diff
@@ -0,0 +1,133 @@
...
+	lib:libblis$secondaryArchSuffix
+	lib:libexecinfo$secondaryArchSuffix
+#	lib:libfftw3$secondaryArchSuffix
+	lib:liblapack$secondaryArchSuffix
+	lib:libopenblas$secondaryArchSuffix
+#	lib:libumfpack$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku_devel$secondaryArchSuffix
```
Begasus: Some mixup? :)
```diff
-	haiku_devel$secondaryArchSuffix
+	haiku${secondaryArchSuffix}_devel
```

extrowerk:
Seems to be ok, at least it works, unlike the earlier, so lets merge it.

Issue #3692, opened 2019-03-16
quazip build fails on x86_gcc2


kenmays:
Ref: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/2709/builds/31988.log

CMake Error: CMAKE_C_COMPILER not set, after EnableLanguage
CMake Error: CMAKE_CXX_COMPILER not set, after EnableLanguage

kenmays:
This recipe is still a work-in-progress. Closing ticket (not an bug/issue).

PR #3695, opened 2019-03-17
QuaZIP: build/install fix

extrowerk:


dev-libs/quazip/quazip-0.7.6.recipe
```diff
@@ -56,7 +57,7 @@ BUILD()
 
 	cmake .. \
 		$cmakeDirArgs \
-		-DCMAKE_ROOT=$libDir/cmake
+		-DINSTALL_PATH=$libDir/cmake
```
Begasus: This works for quazip_x86:
```diff
-		-DINSTALL_PATH=$libDir/cmake
+		-DINSTALL_PATH=$prefix/lib/cmake
```

Begasus: Ìn INSTALL() you should add the next lines to move the library to the correct place:
```
	# move the libraries to the correct location
		mv $prefix/lib $prefix/lib2
		mkdir -p $libDir
		mv $prefix/lib2/* $libDir/
		rmdir $prefix/lib2
	fi
```

Begasus: Checked and working ok with the suggested changes :) (also for x86_64)

extrowerk:
closes https://github.com/haikuports/haikuports/issues/3692

dev-libs/quazip/quazip-0.7.6.recipe
```diff
@@ -56,7 +57,7 @@ BUILD()
 
 	cmake .. \
 		$cmakeDirArgs \
-		-DCMAKE_ROOT=$libDir/cmake
+		-DINSTALL_PATH=$prefix/lib/cmake
```
alaviss: I've take a look at the [cmake module][0]. It seems that this module utilizes `pkg-config` to do it's work, so I'd say it's architecture independent and can be put in `$dataDir/cmake`.

[0]: https://github.com/stachenov/quazip/blob/d5eaa8be998da2430ffd43e0eefa617cf1550ca3/FindQuaZip.cmake

dev-libs/quazip/quazip-0.7.6.recipe
```diff
@@ -70,6 +71,14 @@ INSTALL()
 	mv $prefix/include/* $includeDir
 	rm -rf $prefix/include
 
+	# move the libraries to the correct location
+	if [ -n "$secondaryArchSuffix" ]; then
+		mv $prefix/lib $prefix/lib2
+		mkdir -p $libDir
+		mv $prefix/lib2/* $libDir/
+		rmdir $prefix/lib2
+	fi
```
alaviss: I think instead of adding hacks like this we should patch the `CMakeLists.txt` to use [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) then upstream it.

Begasus: If you could point to some examples maybe I could look into it :)

Begasus: Ok, with a small patch the libs are ok, couldn't find a place where to change the destination for the headers yet

alaviss: [Here](https://github.com/stachenov/quazip/blob/d5eaa8be998da2430ffd43e0eefa617cf1550ca3/quazip/CMakeLists.txt#L31) you go :) It should be set to `${CMAKE_INSTALL_INCLUDEDIR}/quazip`.

Begasus:
This should do it (please check on x86_64)

diff --git a/CMakeLists.txt b/CMakeLists.txt
index 9f92b8f..70b6151 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,7 @@ endif(UNIX OR MINGW)
 set(LIBRARY_OUTPUT_PATH ${CMAKE_BINARY_DIR})
 
 set(LIB_SUFFIX "" CACHE STRING "Define suffix of directory name (32/64)")
-set(LIB_DESTINATION "${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX}" CACHE STRING "Library directory name" FORCE)
+set(LIB_DESTINATION "${CMAKE_INSTALL_LIBDIR}" CACHE STRING "Library directory name" FORCE)
 set(QUAZIP_LIB_TARGET_NAME quazip${QUAZIP_LIB_VERSION_SUFFIX} CACHE
     INTERNAL "Target name of libquazip" FORCE)
 
diff --git a/quazip/CMakeLists.txt b/quazip/CMakeLists.txt
index abcdcf7..3355b61 100644
--- a/quazip/CMakeLists.txt
+++ b/quazip/CMakeLists.txt
@@ -28,5 +28,5 @@ set_target_properties(${QUAZIP_LIB_TARGET_NAME} quazip_static PROPERTIES VERSION
 target_link_libraries(${QUAZIP_LIB_TARGET_NAME} ${QT_QTMAIN_LIBRARY} ${QTCORE_LIBRARIES} ${ZLIB_LIBRARIES})
 target_link_libraries(quazip_static ${QT_QTMAIN_LIBRARY} ${QTCORE_LIBRARIES} ${ZLIB_LIBRARIES})
 
-install(FILES ${PUBLIC_HEADERS} DESTINATION include/quazip${QUAZIP_LIB_VERSION_SUFFIX})
+install(FILES ${PUBLIC_HEADERS} DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/quazip${QUAZIP_LIB_VERSION_SUFFIX})
 install(TARGETS ${QUAZIP_LIB_TARGET_NAME} quazip_static LIBRARY DESTINATION ${LIB_DESTINATION} ARCHIVE DESTINATION ${LIB_DESTINATION} RUNTIME DESTINATION ${LIB_DESTINATION})
--

Begasus:
@extrowerk any progress on this?

extrowerk:
Cleaned a bit up. Bega could you please retest it?

extrowerk:
Bega: i just incorporated all your fixes, but had to add a "5" to the .pc file, because otherwise it contains only -lquazip
Seems to be working ok.

Begasus:
As I told you on IRC, there was only 1 thing left to do, that was it ;)

> @Begasus: i just incorporated all your fixes, but had to add a "5" to the .pc file, because otherwise it contains only -lquazip
> Seems to be working ok.

dev-libs/quazip/patches/quazip-0.8.1.patchset
```diff
@@ -44,3 +44,73 @@ index abcdcf7..3355b61 100644
...
++prefix=@CMAKE_INSTALL_PREFIX@
++exec_prefix=${prefix}
++libdir=@CMAKE_INSTALL_LIBDIR@
++includedir=@CMAKE_INSTALL_INCLUDEDIR@/quazip@QUAZIP_LIB_VERSION_SUFFIX@
++
++
++Name: Quazip
++Description: Quazip Library
++Version: @QUAZIP_LIB_VERSION@
++Libs: -L${libdir} -lquazip5
```
Begasus: Need to check this in a bit, but think you can use @QUAZIP_LIB_VERSION_SUFFIX@ also for the "5" :)

Issue #3584, opened 2019-01-25
Update some fonts to newer versions


twse:
Please update the following fonts to newer versions:
- Fira 4.202
- Fira Code 1.206
- Lato 3.001 dev
- Roboto 2.138
- Source Sans Pro 2.040 upright and 1.090 italic
- Source Serif Pro 2.010 upright and 1.010 italic

hanya:
Lato 3.001 dev has not released as font file.
Current Roboto 2.136 uses roboto-hinted.zip which contains hinting. But Roboto 2.138 does not have hinted one.

waddlesplash:
Is the main Roboto now hinted, they just removed the unhinted version?

twse:
Roboto 2.138 does have unhinted version.

waddlesplash:
I see. Yeah, there's not much we can do then; we certainly need the hinted versions.

twse:
Overpass should be bumped to 3.0.3 – the font also have LGPL 2.1 license as well.

twse:
The Nanum font should be updated to v2.5.

twse:
IBM Plex is now in version 1.3.0 – this release adds the Arabic font so please do that.

twse:
The Liberation fonts just got an update: it is now version 2.00.5 and is released under the SIL Open Font License v1.1. Please do that.

PR #3703, opened 2019-03-20
ibm_plex: bump version

hanya:


hanya:
Requested in #3584.

Changed from TrueType to OpenType.
Could we split fonts into some packages such as Arabic, Devanagari, Thai and others? Maybe later.

PR #3707, opened 2019-03-23
RocksnDiamonds, bump version, switch to SDL2

Begasus:
SDL 1.2 isn't supported anymore


Begasus:
The issue of saving files in /boot/home/. still remains, have been checking again to find out where to change it in the source but nothing seems to work (maybe I need to write an email to the developers) :)

Begasus:
Got an email back from the developers, patch will be upstreamed and unneeded for the next release then, he would also investigate why the saving path isn't working as it should when declared

PR #3710, opened 2019-03-24
ack, bump version

Begasus:
If it's ok I could remove the previous version also?


extrowerk:
> If it's ok I could remove the previous version also?
Yep.

PR #3715, opened 2019-03-25
atk, bump version

Begasus:


Begasus:
Builds ok for _x86, test run ok, needs to be tested on x86_64
Comments welcome :)

korli:
So one test fails on x86?
gdbus-codegen, glib-genmarshal, glib-mkenums, gtester-report need python. I see cmd:python in BUILD_PREREQUIRES.

Begasus:
> So one test fails on x86?
> gdbus-codegen, glib-genmarshal, glib-mkenums, gtester-report need python. I see cmd:python in BUILD_PREREQUIRES.

No, python3 doesn't provide cmd:python as expected by glib-genmarshal, as @alaviss  menstioned on IRC probably our glib port needs to be fixed for that?

korli:
You say that haikuporter doesn't bring python2 in the chroot?

Begasus:
At the moment there is setuptools_python3 in BUILD_REQUIRES (python3?) and cmd:python (python2), as I asked about it it wouldn't be a good thing to mix the both together? @alaviss could you explain this a bit further? :)

Begasus:
Builds and test are ok for _x86, if someone could check this on x86_64 I think it should be ok ;)

PR #3722, opened 2019-03-27
boehm_gc: bump to 8.0.4

ivmai:
Update from v7.6.12 to v8.0.4
Remove --enable-parallel-mark which is default now

### DIFF
--- a/dev-libs/boehm_gc/boehm_gc-8.0.4.recipe
+++ b/dev-libs/boehm_gc/boehm_gc-8.0.4.recipe
@@ -20,14 +20,14 @@ COPYRIGHT="1988, 1989 Hans-J. Boehm, Alan J. Demers
 LICENSE="Boehm"
 REVISION="1"
 SOURCE_URI="https://github.com/ivmai/bdwgc/releases/download/v$portVersion/gc-$portVersion.tar.gz"
-CHECKSUM_SHA256="6cafac0d9365c2f8604f930aabd471145ac46ab6f771e835e57995964e845082"
+CHECKSUM_SHA256="436a0ddc67b1ac0b0405b61a9675bca9e075c8156f4debd1d06f3a56c7cd289d"
 SOURCE_DIR="gc-$portVersion"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
-libcordVersion=1.3.2
-libgcVersion=1.3.6
+libcordVersion=1.4.0
+libgcVersion=1.4.3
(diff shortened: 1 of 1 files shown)


alaviss:
Can you keep the old recipe as `boehm_gc76`? Since this update changes the `SONAME`, we would need to keep the old version for software that are still using the old one.

ivmai:
I've changed the commit to keep v7.6.x recipe. (The SONAME is changed because new API is added but the new version is fully backward compatible with v7.6.x.)

ivmai:
Do I need to change anything else?

extrowerk:
I see no Problem here. Could we merge this?

korli:
This old recipe is still named boehm_gc instead of boehm_gc76.

ivmai:
> This old recipe is still named boehm_gc instead of boehm_gc76.

How to do it?

korli:
is there really a SONAME change? I see 1 for both libs for both recipes.

ivmai:
SONAMEs have been changed:
boehm_gc-7.6.12.recipe:
libcordVersion=1.3.2
libgcVersion=1.3.6
->
boehm_gc-8.0.4.recipe:
libcordVersion=1.4.0
libgcVersion=1.4.3

REVISION="1" is in both receipts. Do I need to increment it? Which one?

extrowerk:
@ivmai : for every new version the REVISION should start from 1. If you change the recipe (adding new confoguration flags or patch, etc), but not the version, you have to bump te REV .
If you change the version, the REV should be switched back to 1.

The soname changed, that's right, but the compatibility flag is still the same, for example:
libcordVersion=1.4.0
libcordVersionCompat="$libcordVersion compat >= ${libcordVersion%%.*}"

So while the soname changed, the ${libcordVersion%%.*} still evaluates to 1.
Afaik HaikuPorter creates symlinks to your lib, so it will create in this case a libcord.so (for _devel), a libcord.so.1 and a libcord.so.1.4 to the real lib called libcord.so.1.4.0

Now while i write it i'm thinking about which one will be written then into the binaries.... Hope somebody could clear us booth up. @alaviss maybe?

extrowerk:
In case of Guile, the main lib (libguile) depends on ```libgc.so.1```, so according to this we don't need the old recipe, as both recipe provides this, and both recipe declares ```compatible >=1```.
So if the ABI is still compatible, you can safely delete the old recipe and keep only the new one, the dependant programs will work with the new version.

ivmai:
Yes, the ABI is backward compatible. I have deleted old recipe.

korli:
SONAME is only libgc.so.1. See https://en.wikipedia.org/wiki/Soname for explanation.

PR #3724, opened 2019-03-28
Pyqt : disable OpenGL

extrowerk:


extrowerk:
Seems to be ok, at least doesnt breaks ports.

PR #3733, opened 2019-03-31
bmake, fix recipe

Begasus:
Not checked on x86_64 yet


Begasus:
Related to https://github.com/haikuports/haikuports/issues/3594

PR #3741, opened 2019-04-05
isl, bump version, remove obsolete isl13

Begasus:

PR #3742, opened 2019-04-05
cloog, bump version

Begasus:


Begasus:
Not tested on x86_64 (yet)

Begasus:
Don't merge before latest PR for isl got merged :) https://github.com/haikuports/haikuports/pull/3741

korli:
we could also remove the package.

Begasus:
> we could also remove the package.

As in, remove isl and cloog? (searching on haikuports I don't see any use of it so I wouldn't have a problem with it)

korli:
gcc can directly use isl, cloog could be removed.

Begasus:
If this is ok with the others I will remove it then, I'll wait for a few days before doing so.

PR #3747, opened 2019-04-07
LibJPEG-Turbo: bump, switch to cmake, enable x86_64 and x86

extrowerk:


Begasus:
Missing "cd haiku_build" in TEST(), builds ok and tests are ok for _x86
Build and tests are ok also for x86_gcc2 but fails to collect the files due to a problem with defineDebugInfoPackage?

PR #3748, opened 2019-04-07
GCC: enable objc without objc lib

extrowerk:


korli:
Please also apply on the 8.3 recipe.

extrowerk:
GCC8.3.0 built without any problem on 64 bit.

PR #3761, opened 2019-04-11
OpenMPI: bump

extrowerk:
all test passes


extrowerk:
Nothing using it yet, so it should not harm to merge.

PR #3772, opened 2019-04-20
wizznic, bump version

Begasus:


Begasus:
Switched to lib:libpng16, works ok on _x86 and x86_64

PR #3774, opened 2019-04-21
gn: bump version and add supporting code for Qt

KapiX:
These 3 commits from Qt are needed to build QtWebEngine.
@return please review. I didn't check if it breaks V8, but it shouldn't; these commits only add functionalities and don't remove or change anything.


dev-util/gn/gn-2c72317.recipe
```diff
@@ -1,14 +1,14 @@
...
-COPYRIGHT="2008-2018 The Chromium Authors."
+COPYRIGHT="2008-2019 The Chromium Authors."
 LICENSE="BSD (3-clause)"
 REVISION="1"
-srcGitRev="091169beda92aff21ded3e9dfb392ea12b10c0e0"
-SOURCE_URI="https://github.com/return/gn/archive/$srcGitRev.tar.gz"
-CHECKSUM_SHA256="e17c58a90e26212c74cd527b1a10d315ede15a64c9cf0755b4ef78418c379b76"
+srcGitRev="2c72317e3a1f579c6a527c067ed575c0b92e544e"
+SOURCE_URI="https://gn.googlesource.com/gn/+archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="c180838996a6f0e77cbece31ec0fc03f75a43fdb2d0fa3d9233610ca98157aaa"
```
return: The tar archives from gn.googlesource.com are not deterministic and the checksums will be different on every download. This issue is on gerrits side see: https://github.com/google/gitiles/issues/84

A solution on our side would be to create an archive of this ourselves and host it in the ports-mirrror Similar to this suggestion https://github.com/haikuports/haikuports/pull/3330#discussion_r252917404

return:
@KapiX I have tested GN with your changes and it doesn't break V8. However, I ran the GN unittests and there was a failure with one of the tests:

[354/494] QMakeLinkWriter.WriteLinkPri
*** FAILURE /sources/tools/gn/qmake_link_writer_unittest.cc:140: expected2 == out_str2

[494/494] XmlElementWriter.TestXmlEscape
FAILED

I'm not sure about whether if this affects the QtWebEngine port that much, but it would be nice if this was fixed.

korli:
We should use something better than the git commit for the recipe version. arch uses: 0.{number of commits}.{8 first characters of the git commit}

PR #3776, opened 2019-04-21
bpg, bump version

Begasus:


Begasus:
x265 disabled because it's using internal source to build it (only for x86_64 which fails because of "nice")

media-libs/bpg/bpg-0.9.8.recipe
```diff
@@ -22,7 +22,9 @@ PROVIDES="
 REQUIRES="
 	haiku$secondaryArchSuffix
 	lib:libjpeg$secondaryArchSuffix
-	lib:libpng$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
+	lib:libSDL$secondaryArchSuffix
```
waddlesplash: What part of this requires SDL? Can we break that out into a subpackage?

korli:
I would rename this package to libbpg:
https://repology.org/project/bpg/versions (only 1)
https://repology.org/project/libbpg/versions (a lot)

media-libs/bpg/bpg-0.9.8.recipe
Begasus: The image viewer bpgview (which isn't installed by default), I could check to see if I can disable that one ...

Begasus: Changed the recipe a bit so it doesn't build the viewer

korli:
You can add
REPLACES="bpg$secondaryArchSuffix"
and
REPLACES_devel="bpg${secondaryArchSuffix}_devel"

media-libs/libbpg/libbpg-0.9.8.recipe
```diff
@@ -3,30 +3,30 @@ DESCRIPTION="BPG is an image format based on the HEVC standard that aims to \
...
+COPYRIGHT="2014-2018 Fabrice Bellard"
 LICENSE="MIT
 	GNU LGPL v2.1"
-REVISION="4"
-SOURCE_URI="http://bellard.org/bpg/libbpg-0.9.4.tar.gz"
-CHECKSUM_SHA256="9986d66408f36fb678014bfcbb693c0f8358767719bc05797e4af54fa4237c2a"
+REVISION="1"
+SOURCE_URI="http://bellard.org/bpg/libbpg-$portVersion.tar.gz"
+CHECKSUM_SHA256="c0788e23bdf1a7d36cb4424ccb2fae4c7789ac94949563c4ad0e2569d3bf0095"
 SOURCE_DIR="libbpg-$portVersion"
```
korli: SOURCE_DIR is now superfluous.

korli:
Is there no cmd:bpgview? see https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libbpg#n29

Begasus:
> Is there no cmd:bpgview? see https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=libbpg#n29

There is, but the default "make install" doesn't install it, I could add it however, it would need SDL then, also I see that they install 2 headers, make install doesn't do that also, so I have to manually add them

Begasus:
> You can add
> REPLACES="bpg$secondaryArchSuffix"
> and
> REPLACES_devel="bpg${secondaryArchSuffix}_devel"

Removing/renaming the recipe doesn't do that automatically?

media-libs/bpg/bpg-0.9.8.recipe
Begasus: Checked with installing the binary bpgview, can't split the cmd's into separate packages?

PR #3777, opened 2019-04-21
lcdproc, enable secondary arch

Begasus:


Begasus:
x86_gcc2 broken due to parse error(s)

PR #3781, opened 2019-04-22
MPlayer: bump to release

extrowerk:


media-video/mplayer/mplayer-1.4.0.recipe
```diff
@@ -193,6 +191,7 @@ BUILD()
 		--enable-haiku \
 		--enable-menu \
 		--enable-runtime-cpudetection \
+		--enable-openssl-nondistributable \
```
waddlesplash: "nondistributable"?

pulkomandy: https://people.gnome.org/~markmc/openssl-and-the-gpl.html

Can mplayer use another SSL library instead?

PR #3782, opened 2019-04-22
visualboyadvance, mark broken due to unsafe source

Begasus:


waddlesplash:
If you update the mGBA recipe, you can probably remove this one entirely.

Begasus:
> If you update the mGBA recipe, you can probably remove this one entirely.

Tried to pick up with the latest version around on GitHub, but it depends on wxWidgets/GTK : https://github.com/visualboyadvance-m/visualboyadvance-m (could disable wxWidgets with cmake but then it only builds two static libs and only installs an *.ini file)

Begasus:
@waddlesplash bumped mGBA at https://github.com/haikuports/haikuports/pull/3788
If that is ok I'll remove visualboyadvance from the tree

Begasus:
Looks like there might be a problem with Perl when checking in Terminal ... (need to check with a working recipe): https://github.com/haikuports/haikuports/issues/3789

waddlesplash:
mGBA succeeded, so VBA can be deleted. :)

PR #3785, opened 2019-04-22
FFMPEG: Bump, switch to openmpt

extrowerk:
Haiku is now able to play module music files!


extrowerk:
Tested with many different media files, looks ok to me.

diversys:
So, do we need to replace libmodplug with libopenmpt in all of the recipes?

KapiX:
Do you mean recipes which depend on FFMPEG and libmodplug? I'd say they should not have a dependency on libmodplug in the first place. FFMPEG is supposed to be an abstraction layer and what library it uses should not be important. Just let FFMPEG pull in whatever it wants.

extrowerk:
> So, do we need to replace libmodplug with libopenmpt in all of the recipes?

There are some recipes which depends directly on libmodplug, but openmtp provides the same lib for compatibility reasons, so no change required for now afaik.

The other ones which indirectly (through FFMPEG) depend on it should not notice the change because the abstraction.

PR #3790, opened 2019-04-23
Ruby: bump

extrowerk:
Ruby bump.

Problems:
- [x] [rubygems.rb is read only](https://bugs.ruby-lang.org/issues/15787)
- [x] [crash in pthread_attr_get_np](https://github.com/haikuports/haikuports/issues/3791)


extrowerk:
I think it could be ok for merge, but i haven't tested the existing depending recipes.
2.2.x isalready deprecated, so we should make this one as default, thus no ruby26 recipe.

korli:
at least I found no users of lib:libruby. There are a few recipes which depend on devel:libruby though.

PR #3792, opened 2019-04-24
libzmf, fix libpng16 reference

Begasus:


Begasus:
This lib is used by libreoffice (shouldn't affect it as there are no structural changes) @diversys could you check?

PR #3793, opened 2019-04-24
golang: Update patchset to enable bootstrapping.

return:
This patches the go compiler for Haiku to fix process
waiting when executing Wait() on a process. Should also fix
the net/url hang [0] when attempting to bootstrap Go 1.5 [1] and
above.

Also changes the GOROOT_FINAL location to a globally writeable
location in the non-packaged folder to allow golang to install
third-party libraries.

[0] For context on this issue, see https://github.com/haikuports/haikuports/pull/3119#issuecomment-424038086
[1] See: https://gist.github.com/return/6e1d24dc67da616b5f828a3b84b12704

### DIFF
--- a/dev-lang/golang/golang-1.4.3.recipe
+++ b/dev-lang/golang/golang-1.4.3.recipe
@@ -12,7 +12,7 @@ interpreted language."
 HOMEPAGE="https://golang.org/go"
 COPYRIGHT="2009-2018 The Go Authors"
 LICENSE="BSD (3-clause)"
-REVISION="1"
+REVISION="2"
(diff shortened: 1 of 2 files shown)


kallisti5:
Looks good to me.

The end-goal of this package is to allow us to bootstrap a later version... thus i'm 100% with us focusing on getting it functional vs being perfect.   We can focus on perfection when this stuff is up-streamed.  We are going to try to upstream some of this stuff on golang master... right? :-D

return:
Yup, for now 64 bit will be upstreamed first and then 32 bit support will follow.

PR #3797, opened 2019-04-26
libharu, fix _x86 build

Begasus:


Begasus:
@korli could you check? This is working ok for me (no broken links in $libDir) for secondary architecture (strange thing is that you can use CMAKE_INSTALL_LIBDIR but the build doesn't seem to use that at INSTALL)

PR #3798, opened 2019-04-26
love, bump version

Begasus:


Begasus:
Checked out latest version, builds OK but needs some investigations (should be OK), also if someone could check this for x86_64 (don't think my system is up for it - heavy load)

PR #3799, opened 2019-04-27
Mediainfo: bump, shared, use system libs

extrowerk:


korli:
Error: No match found for license BSD 2-Clause

Issue #3800, opened 2019-04-27
meson broken?


Begasus:
After trying to build something with meson (libmpdclient, libepoxy) I'm getting the next errors:
From haikuporter:
```
Traceback (most recent call last):
    File "/bin/meson", line 6, in <module>
      from pkg_resources import load_entry_point
ModuleNotFoundError: No module named 'pkg_resources'
```
From Terminal:
```
[x86] ~/WIP/libepoxy-1.5.3/build> meson --buildtype=release --prefix=/boot/home/destdir/
  Traceback (most recent call last):
    File "/bin/meson", line 6, in <module>
      from pkg_resources import load_entry_point
ModuleNotFoundError: No module named 'pkg_resources'
```

Begasus:
Seems like this is linked to the new python version (3.7), setuptools should be rebuild against the new version for pymhon, then the latest meson can be build against the correct version of setuptools (which requires 3.6 atm)

PR #3801, opened 2019-04-27
setuptools, use python3.7 now

Begasus:


Begasus:
Related to https://github.com/haikuports/haikuports/issues/3800

Begasus:
This should be checked before merging as quite a lot of recipes make use of setuptools_python3

Begasus:
Installed this one on gcc2h and build latest version for meson with it, then after installing the new meson version checked with a build for libepoxy (which uses meson) and things are ok so far (installing this didn't remove any other packages relying on setuptools_python3 ...

PR #3802, opened 2019-04-27
libcdio-paranoia: bump version to 10.2+2.0.0

enzo1982:


enzo1982:
The first version of this PR had `devel:libcdio >= 18` listed under `BUILD_REQUIRES`, but the new libcdio 2.1.0 release bumps the .so version to 19 so building libcdio-paranoia failed even though it's completely fine with libcdio 2.1.0.

So what's the right way to deal with this? To me it feels wrong to list an .so version under `BUILD_REQUIRES` and I'm more comfortable with listing a minimum release version there - hence the change in the new version of this PR. But maybe I'm wrong and should say `devel:libcdio >= 19` instead?

korli:
devel:libcdio >= 18 was correct. that libcdio 19 (or 20, etc) is compatible with 18 can't be foreseen.

PR #3805, opened 2019-04-27
libcdio: bump version to 2.1.0

enzo1982:


korli:
SONAME change.

korli:
Please consider renaming libcdio-2.0.0.recipe for compatibility. That way you don't have to worry right away about rebuilds, and it keeps third party repositories happy.

PR #3811, opened 2019-04-28
freac: bump version to 1.1~alpha20190423

enzo1982:


korli:
Some advice: adding version constraints in REQUIRES shouldn't be needed, haikuporter will take what it finds at build time to add runtime constraints anyway.

Some remarks: The versioning scheme in dependencies of freac seems too complex: if the API isn't stable, maybe it would be better to package everything in the freac package. The recipe can download the four tarballs, build each one, then package everything. I don't see any consumer for the libraries yet, so this shouldn't be a problem.

enzo1982:
Thank you for merging!

I'd like to keep three separate packages. I plan to stabilize the API with the next release which is scheduled to be 1.1 beta 1. Dependencies will be much simpler from then on as I don't plan to package any more development release after that point - only beta and stable.

PR #3813, opened 2019-04-29
libclaw, fix libpng16 reference

Begasus:


Begasus:
Not sure here, but shouldn't **cmd:claw_config** go into the devel package? @korli what do you think?

Begasus:
claw-config revealed the --cflags and --libs, so imho it should go into the devel package

PR #3818, opened 2019-04-30
fontforge, bump

Begasus:


Begasus:
Needs someone to check this for x86_64 :)

Begasus:
@diversys could you check this for libreoffice?

Begasus:
Still needs to be tested, my system freezes over this build from time to time (not related to Haiku)

Begasus:
Not sure on enabling cairo, I'm probably going to comment the needed libs for that out for now, recipe should be ok after that (system keeps freezing during the build atm due to the heavy load while building - should have a descent dedicated system to build these heavier parts) :)

media-gfx/fontforge/fontforge-20190501~git.recipe
```diff
@@ -112,9 +112,7 @@ BUILD()
 	export LDFLAGS="-lnetwork $LDFLAGS"
 	./bootstrap
 	runConfigure --omit-dirs binDir ./configure \
-		--prefix=$commandBinDir \
 		--enable-python-scripting=3
-		false
```
Begasus: That was still in the phase :) you also removed $commandBinDir?

korli:
tested on x86_64

media-gfx/fontforge/fontforge-20190501~git.recipe
korli: Yeah, it would install in $prefix/bin instead of $prefix

Begasus: That was the intention https://github.com/haikuports/haikuports/blob/7d74d6d048fe65a06c1fe8b6709da832bc031899/media-gfx/fontforge/fontforge-20190501%7Egit.recipe#L23

Begasus: ah jikes! should be --bindir=$commandBinDir :)

korli: I added --bindir now

korli:
can we remove WIP?

Begasus:
@korli thanks for checking out x86_64! :+1:

Begasus:
@korli & @diversys this is used by libreoffice, so should be checked before merging in :)

korli:
for libreoffice it seems to be a build time requirement, so no stress.

PR #3819, opened 2019-05-01
libuninameslist, new recipe

Begasus:


korli:
is it used by fontforge?

Begasus:
> is it used by fontforge?

Yes, haven't bumped the PR for fontforge yet to include this and the one for libspiro, fontforge is pretty heavy to my build system so it tends to freeze my laptop (not related to Haiku)

Begasus:
Not really sure about the license for both this and libspiro, so maybe someone could take a closer look at this?

korli:
Maybe we could skip this one. ArchLinux doesn't seem to use this library for fontforge.

Begasus:
No, problem, this could be closed then without merging

PR #3826, opened 2019-05-02
MPD: use libcdio_paranoia

extrowerk:


korli:
btw it's outdated, last is 0.21.8

extrowerk:
@korli: true, here is the newest one. https://github.com/haikuports/haikuports/pull/3513

PR #3829, opened 2019-05-03
psi_plus: use native DNS resolver from Qt

tehnick:
It was recently improved in Haiku port of Qt:
https://github.com/haikuports/haikuports/blob/53f29b56/dev-qt/qt5/patches/qt5-5.12.3.patchset#L343


Begasus:
There is a tab too much at the end of line 30 :) (trailing whitespace)

PR #3830, opened 2019-05-03
graphicsmagick, bump version, fix libpng16 reference

Begasus:


Begasus:
Could use some checking in builds (pretty heavy on my system) before merging

Begasus:
Recipe should be ok, system freezes here when building (not related to the builds) :)

Begasus:
Finished the build for _x86 :) Ran the testsuite (partial) and things look ok from my side :)

PR #3844, opened 2019-05-08
pixman: bump to 0.38.4

kenmays:
Built and tested on hrev53102.


waddlesplash:
ABI-laboratory says no breakage.

PR #3847, opened 2019-05-09
libdvdread: bump to 6.0.1

kenmays:
libdvdnav is a library for developers of multimedia applications. It allows easy use of sophisticated DVD navigation features such as DVD menus, multiangle playback and even interactive DVD games.


waddlesplash:
Did you build this?

waddlesplash:
Ah, I see this is just a point bump.

PR #3861, opened 2019-05-17
psi_plus: build program without webkit

tehnick:
After recent changes in memory allocator in Haiku kernel
QtWebKit regulary segfaults at program startup.

As a temporary workaround we may disable webkit usage in
Psi+. Once QtWebKit will be fixed this commit should be
reverted.


waddlesplash:
IMO this should not be merged. Beta1 did not get the changes of course, and most users are probably on that. We shouldn't downgrade their experience because of a temporary bug in QtWebKit.

Also the changes were in the libc, not in the kernel.

tehnick:
Ok.

PR #3863, opened 2019-05-18
armagetronad, fix libpng16 reference

Begasus:


Begasus:
No functional changes, armagetronad crashes on launch here, not sure if it already did that :)

diversys:
It did, that't why it was disabled in the first place.

PR #3865, opened 2019-05-18
freegish, fix reference for libpng16, cleanup

Begasus:


Begasus:
Still not working, so left the builds disabled :)

PR #3871, opened 2019-05-20
armagetronad, fix reference to lib:libSDL*_1.2

Begasus:


Begasus:
Looks like quite a few packages are involved in libSDL*_1.2 issue

PR #3879, opened 2019-05-24
nodejs: add recipe for v12.3.1

return:
This is a very early and functioning port of Node.js 12.3.1 to Haiku, based on V8 7.4.288.28.

This recipe currently uses the bundled dependencies from libuv, openssl, V8 and brotli. This might change to use shared libraries in the future when the correct dependencies are available in HaikuPorts.

What currently works:

* Node REPL.
* Networking.
* npm (Node modules (global) needs to be in non-packaged folder to download packages)
* Remote Debugging (via Chrome Debugger and inspector)

What does not work:
* CPU activity is missing.
* File Watching (fsevents unimplemented).

Please note that this recipe is very early and is only intended to be used by testers until it is ready to be merged into HaikuPorts.

The patches in this port were adapted and imported from the following ports:

Node.js (experimental): @pahefu 
https://github.com/pahefu/haikuNodeJs

libuv: @CodeForEvolution and @alaviss
https://github.com/alaviss/libuv
https://github.com/CodeforEvolution/libuv 

V8: @return
HaikuPorts PR: https://github.com/haikuports/haikuports/pull/3330

And finally: https://imgur.com/a/FEFOAvu

### DIFF
--- a/net-libs/nodejs/nodejs-12.3.1.recipe
+++ b/net-libs/nodejs/nodejs-12.3.1.recipe
@@ -1,0 +1,137 @@
+SUMMARY="A JavaScript runtime built on Chrome's V8 Engine"
+DESCRIPTION="Node.js is a JavaScript runtime built on Chrome's V8 JavaScript engine. \
+As an asynchronous event driven JavaScript runtime, Node is designed to build scalable \
+network applications. In the following \"hello world\" example, many connections can be \
+handled concurrently. Upon each connection the callback is fired, but if there is no work \
+to be done, Node will sleep."
+HOMEPAGE="https://nodejs.org/"
+COPYRIGHT="2006-2019 The Node.js Foundation"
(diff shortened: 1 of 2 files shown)


waddlesplash:
Libuv has been merged; so all the statically bundled dependencies should be pared down now.

Additionally, we can use GCC instead of Clang at this point too.

return:
@waddlesplash The recipe is now compiles using GCC. It also uses a shared V8, libuv, c_ares and libz. The last one to be made shared is openssl11 which fails to compile.

With the required recipes (V8 and libuv) being available in HaikuPorts I'm opening this recipe for further review.

waddlesplash:
Looks good to me!

return:
@waddlesplash PR #3897 needs to be merged or else npm won't work properly.

waddlesplash:
Done.

return:
Thanks!

PR #3880, opened 2019-05-25
fife, fix references to libpng16 and libSDL*

Begasus:


Begasus:
Can't check this as it's not enabled for _x86, no structural changes

PR #3881, opened 2019-05-26
openclonk, bump git revision, fix reference to libpng16/libSDL_*

Begasus:


Begasus:
Haven't been able to build to the end here, so if any one could check thanks :)

Begasus:
Just a small fix in the recipe comming up, but it seems to be ok :)
OpenClonk

Begasus:
@threedeyes could you check? As you wrote the original recipe :)

Begasus:
Related to https://github.com/haikuports/haikuports/pull/3881

PR #3891, opened 2019-06-01
crypto++: version bump to 8.2

kikadf:


dev-libs/crypto++/crypto++-8.2.0.recipe
```diff
@@ -86,8 +84,7 @@ Version: $portVersion
 Libs: -lcryptopp
 EOF
 
-	packageEntries devel \
-		$relativeDevelopDir
+	packageEntries devel $relativeDevelopDir
```
Begasus: Is $reletiveDevelopDir needed here? Or could you just use $developDir?

kikadf: Well, I don't know what is the differences between $relativeDevelopDir
and $developDir.
I tried build without this line and I got error, so just leave there.
Can you help me, where can I found this descriptions? Should I check the haikuporter source?

kikadf: Ok, I found in the haikuporter's wiki. Absolut path vs relative path.
So:
developDir=/boot/system/develop
relativeDevelopDir=/develop ?

Begasus: From the wiki "$relativeDevelopDir: $developDir as a relative directory (to prefix)." https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes, haven't checked if it works with $developDir :)

waddlesplash: Most invocations of `packageEntries` use $developDir, so try that.

kikadf: Done.

Begasus:
Haven't checked but LGTM

kikadf:
Please wait for merge, I found the patch isn't complete. I am fixing that, then upgrade the PR.

kikadf:
The patchset is fixed, build and works fine on x86_64. Only found one package yet, what has cryptopp dependency: codecrypt (Tested and build fine on x86_64).

waddlesplash:
Very nice, thanks! Sorry for the delay!

Issue #3890, opened 2019-05-31
libxml2 broken?


Begasus:
I've been seeing this a few times already trying to build something where libxml2 is involved:
```
/boot/system/develop/headers/libxml2/libxml/dict.h:23: multiple storage classes in declaration of `xmlDict'
/boot/system/develop/headers/libxml2/libxml/dict.h:24: multiple storage classes in declaration of `xmlDictPtr'
/sources/AbiWord-217fbb622ee64efc181cd67d2742d6a60a086e2d/abi/src/../src/config/abi_rules.mk:243: recipe for target '/sources/AbiWord-217fbb622ee64efc181cd67d2742d6a60a086e2d/abi/src/../src/Haiku_1_x86_OBJ/obj/ut_xml_libxml2.o' failed
```

Begasus:
Not all packages are infected by this, for instance the new PR for fontconfig works ok with libxml2

Begasus:
xmoto for x86_gcc2 is infected by this also, building xmoto_x86 is ok

korli:
seems a header isn't compatible with gcc2.

korli:
fixed in cd615d7aea9d95155ee41d4f3aff400470d128d6

Begasus:
Thanks! Will check later once it's available :)

Issue #3878, opened 2019-05-24
libSDL* recipes needs cleanup for lib:libSDL*_*


Begasus:
There are a lot of recipes using lib:libSDL/lib:libSDL2 in their recipes, libSDL/libSDL2 aren't provided in the base package (they however are provided in the devel package)
These recipes should be changed to use **lib:libSDL_1.2** and **devel:libSDL_1.2** instead (just like the issue for libpng16 (https://github.com/haikuports/haikuports/issues/3779).
This involves all the subpackages like sdl_mixer, sdl_image etc (should be **lib:libSDL_image_1.2** etc ...

pulkomandy:
The subpackages can be different I think (for example sdl_ttf is version 2 even for SDL 1.2).
How to check:

package list libsdl_image-*

If there is a libsdl_image-1.2.so* (version number BEFORE the .so), we need lib:libsdl_image_1.2
But if there is a libsdl_image.so.1.2 or similar, we need lib:libsdl_image

package list libsdl_image_devel-*

If there is a libsdl_image.so in the develop/lib dir, then we need devel:libsdl_image

Begasus:
Didn't check the libSDL_ttf one :) The others I've checked so far provide lib:libSDL_*_1.2, but yes, things need to be checked before changing the actual recipes

Begasus:
Packages ok (with a quick look):
´´´
lib:libSDL_gfx
```
Packages affected:
´´´
lib:libSDL_1.2
lib:libSDL_image_1.2
lib:libSDL_mixer_1.2
lib:libSDL_net_1.2
lib:libSDL_rtf_0.1
lib:libSDL_sound_1.0
lib:libSDL_ttf_2.0
```

lonemadmax:
I may try and update some recipes so that they at least compile and comply with this, but there's surely something I don't understand. As far as I can see, (from libsdl and libsdl2, I've not checked the rest yet) only lib:libSDL2 is not provided. The rest: devel:libSDL2, lib:libSDL and devel:libSDL do exist. Is it that the objective is to remove those from the packages and only provide the numbered version, and so the four references should always be changed to x_n.m?

pulkomandy:
Yes. The sdl recipe currently provides some lib: and devel: entries that do not match the actual library name. These should be removed, but it would break a lot of things that rely on it. So first we have to fix all other packages.

Begasus:
libSDL fixed in: https://github.com/haikuports/haikuports/pull/8960

[...]

PR #3894, opened 2019-06-02
xmoto, fix build on x86

Begasus:


Begasus:
x86_gcc2 fails to build due to the libxml2 issue (https://github.com/haikuports/haikuports/issues/3890), _x86 builds but fails to start (fatal exeption: Unable to get xmDb version:)

games-sports/xmoto/xmoto-0.5.11.recipe
```diff
@@ -9,77 +9,83 @@ COPYRIGHT="2006-2014 Nicolas Adenis-Lamarre
...
-	lib:libpng$secondaryArchSuffix
-	lib:libsdl$secondaryArchSuffix
-	lib:libsdl_mixer$secondaryArchSuffix
-	lib:libsdl_net$secondaryArchSuffix
-	lib:libsdl_ttf$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
+	lib:libSDL_1.2$secondaryArchSuffix
+	lib:libSDL_mixer_1.2$secondaryArchSuffix
+	lib:libSDL_net_1.2$secondaryArchSuffix
+	lib:libSDL_ttf_2.0$secondaryArchSuffix
```
alaviss:
```suggestion
	lib:libSDL_ttf_2.0$secondaryArchSuffix
```

Shouldn't this be 1.2 or removed? We can't mix different version of sdl like that...

pulkomandy: SDL ttf up to 2.0.11 goes with SDL 1.2 (https://www.libsdl.org/projects/SDL_ttf/release-1.2.html)
Not sure what the so versions are...

alaviss: The other lib would be libSDL2_ttf it seems, so this isn't much of a problem.

Begasus: Issue for this is at https://github.com/haikuports/haikuports/issues/3878

korli:
maybe switch back to gcc2?

Begasus:
I'll check in later once the libxml2 is in the depot :)

Begasus:
@korli problem with libxml2 is fixed, thanks for that! But disabled the build for gcc2 due to too many parse errors, build for _x86 is ok but still can't launch it (see comment above), marked broken and disabled for now

korli:
Ok thanks. I thought the game was old and originally with gcc2.

Begasus:
It is old, no new changes since first merge for this, that's why it's strange it doesn't build anymore, maybe something changed in our gcc2 that broke it?

Begasus:
Got past the parse errors but at the end ran into:
```
boot/home/WIP/xmoto-0.5.11/src/xmscene/ChipmunkWorld.cpp:178: Internal compiler error, output_operand_lossage `invalid expression as operand'
```
Guess that's the end?

korli:
you could try https://github.com/xmoto/xmoto and switch to x86.

Begasus:
Not really a success with our libode (I think)
xmoto-github

Begasus:
@korli bumped to the github latest commit there, it builds ok in Terminal, not tested with haikuporter yet, but wanted to share the work so far already

korli:
What's defined in ode/precision.h?

Begasus:
@korli check pkg-config-x86 ode --cflags it gives you -dSINGLE

Begasus:
Latest push needs "cd build" in INSTALL before "make install", (build is ok as it seems) but that still fails on 
```
file INSTALL cannot find
"sources/xmoto-$srcGitRev/build/bin/xmoto.bin"
```
That was the problem before also when trying to build from Terminal setting the prefix ...

Begasus:
Latest push needs "cd build" in INSTALL before "make install", (build is ok as it seems) but that still fails on 
```
file INSTALL cannot find
"sources/xmoto-$srcGitRev/build/bin/xmoto.bin"
```
That was the problem before also when trying to build from Terminal setting the prefix ...

> What's defined in ode/precision.h?

Seems this is created at configure time, the next comes from ode 0.16
```
#ifndef _ODE_PRECISION_H_
#define _ODE_PRECISION_H_

/* Define dSINGLE for single precision, dDOUBLE for double precision,
* but never both!
*/

#if defined(dIDESINGLE)
#define dSINGLE
#elif defined(dIDEDOUBLE)
#define dDOUBLE
#else
#define dSINGLE
#endif

#endif
```

Begasus:
If no one objects, I'm closing this one, even with the patches I can't get xmoto to start

PR #3897, opened 2019-06-02
libuv: Implement various missing functions.

return:
This patch syncs the changes of @alaviss's libuv port from upstream into HaikuPorts with the following changes:

* Implements uv__fs_utime() using utimensat()
* Implements uv__fs_futime() using futimens()
* Stub implementation of uv_get_constrained_memory()

Note: Whilst testing uv_fs_utime() and uv_fs_futime() that specific test will still fail due to this issue [0], but it is just enough to get npm (Node Package Manager) fully functioning in Haiku.

[0] https://dev.haiku-os.org/ticket/15102


waddlesplash:
@return, @alaviss: nodejs build failed due to libuv trying to use setpriority, which does not exist: https://build.haiku-os.org/buildmaster/master/x86_64/?buildrunDir=&viewMode=expanded

How did the libuv build succeed? Aren't we using --no-undefined?

waddlesplash:
It may just be time to drop R1/beta1 "support" from the repos. @korli, @pulkomandy, any thoughts? Beta2 is theoretically only 3 months away, fwiw.

pulkomandy:
I would prefer that there is continuous support for stable releases. Maybe we should release beta2 earlier?

return:
@pulkomandy @waddlesplash Would it be possible to add the get/setpriority() commit into the beta1 branch and then we update the builders to fix libuv? Unless there is a reason that this cannot be done?

pulkomandy:
Because no one has been maintaining beta1 much. Yes it's possible, someone has to backport the changes to beta1 and submit a change request to Gerrit.

waddlesplash:
There is also the problem that beta1 branch cannot be built because the buildbot thinks it's master and uploads it to the wrong place, which @kallisti5 still has not fixed.

kwri-avongluck:
so.. last I checked the beta1 builds were working.   @waddlesplash, could you provide some context around what you saw? (errors, links to relevant builts, etc?)

PR #3899, opened 2019-06-03
tesseract, fix reference to libpng16, cleanup

Begasus:


Begasus:
No structural changes, after checking it seems it at least needs the data from eng so added that, also imported libVersion(Compat) and defineDebugInfoPackage (needs to be checked on x86_64 before merging)

Begasus:
Could someone check this on x86_64?

Begasus:
Will look into it later :) (not today)

Begasus:
> Can you please change the `HOMEPAGE`? Google Code has already been shutdown :P
> 
> And we should probably update this package to 3.05.02.

Also see that there is a 4.0.0 version out, should we check the 3.* version then also?

Begasus:
Updated to the latest release, this builds with the current leptonica package but maybe wait until the updated recipe (https://github.com/haikuports/haikuports/pull/3903) for that is merged?

PR #3906, opened 2019-06-06
povray: Fix Boost x86 issue

kenmays:


extrowerk:
Fix the current in-tree version instead of re-adding the old one.

extrowerk:
Oh, and btw, your PR have conflicts.

PR #3914, opened 2019-06-09
povray, fix build for x86_gcc2

Begasus:


Begasus:
@extrowerk could you check this on x86_64? If it's ok the other issue can be closed :)
PS, after install and a first check it seems that povray is looking for the config file in ~/.povray/3.7/*

extrowerk:
Will do

PR #3927, opened 2019-06-16
argon2, bump revision, don't use $(MACHINE_NAME)

Begasus:


Begasus:
x86_64 doesn't list BePC with uname -m, so don't use it, this fixes build on all arch's

PR #3940, opened 2019-06-18
dumb2, update to latest version

Begasus:


Begasus:
Not sure here if it's best to use dumb2 or just dumb?

PR #3942, opened 2019-06-19
freeimage: add libfreeimage.so.3 to $libDir

kikadf:
The freeimage package only contain libfreeimage-3.18.0.so, but some package (one of these porting in progress :P) need libfreeimage.so.3, and runtime_loader don't find in $developDir.

### DIFF
--- a/media-libs/freeimage/freeimage-3.18.0.recipe
+++ b/media-libs/freeimage/freeimage-3.18.0.recipe
@@ -70,8 +70,10 @@ INSTALL()
 	prepareInstalledDevelLibs libfreeimage libfreeimage-$portVersion \
 		libfreeimageplus libfreeimageplus-$portVersion
 
-	ln -s $libDir/libfreeimage-"$portVersion".so $libDir/libfreeimage.so.${portVersion%.*.*}
-	ln -s $libDir/libfreeimageplus-"$portVersion".so $libDir/libfreeimageplus.so.${portVersion%.*.*}
+	ln -s $libDir/libfreeimage-"$portVersion".so \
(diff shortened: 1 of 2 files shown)


pulkomandy:
This sounds wrong. prepareInstalledDevelLibs scans the libraries for the soname and creates the appropriate files. No manual copy should be needed here.

korli:
in Arch Linux, it's already weird:
> usr/lib/libfreeimage-3.18.0.so
> usr/lib/libfreeimage.so
> usr/lib/libfreeimage.so.3

So it's possible one has to produce a symlink manually.

kikadf:
no libfreeimage.so in $libDir, just libfreeimage-3.18.0.so
The  prepareInstalledDevelLibs create the correct symlinks to the $developLibDir, but need a libfreeimage.so.3 symlink to the $libDir as well, but yes I failed with cp, use ln instead.

Begasus:
While you are at it, maybe include freeimageplus also? ;)
https://ibb.co/mzy1TCH

kikadf:
Begasus, done :)

media-libs/freeimage/freeimage-3.18.0.recipe
```diff
@@ -66,7 +83,7 @@ includedir=${includeDir}
 Name: freeimage
 Description: FreeImage
 Version: ${portVersion}
-Libs: -L${developLibDir} -lfreeimage
+Libs: -L${developLibDir} -lfreeimage -lfreeimageplus
```
korli: freeimageplus shouldn't be by default included here.

PR #3949, opened 2019-06-21
gcc-8.3: add --enable-libstdcxx-filesystem-ts to configure

kikadf:


korli:
Please check the 7.3 patchset and apply/rebase on 8.3.

````
From 2d840327dc8f00d11355cc7502eb098fc9f88d6d Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?Zolt=C3=A1n=20Mizsei?= <zmizsei@extrowerk.com>
Date: Wed, 2 May 2018 08:37:20 +0200
Subject: Enable libstdcxx_filesystem_ts for Haiku

diff --git a/libstdc++-v3/acinclude.m4 b/libstdc++-v3/acinclude.m4
index 805148d..e94dc11 100644
--- a/libstdc++-v3/acinclude.m4
+++ b/libstdc++-v3/acinclude.m4
@@ -4297,6 +4297,9 @@ AC_DEFUN([GLIBCXX_ENABLE_FILESYSTEM_TS], [
       solaris*)
         enable_libstdcxx_filesystem_ts=yes
         ;;
+      haiku*)
+        enable_libstdcxx_filesystem_ts=yes
+        ;;
       *)
         enable_libstdcxx_filesystem_ts=no
         ;;
-- 
2.16.4

**Comment by kikadf** [comment]
Fixed patchset, work with patching libstdc++-v3/configure instead of libstdc++-v3/acinclude.m4.
Removed --enable-libstdcxx-filesystem-ts conf opt from recipe.

However, I got error in strict mode:
`Warning: POLICY ERROR: no matching provides "devel:libgcc_s" for "develop/lib/libgcc_s.so"
Warning: POLICY ERROR: no matching provides "devel:libgcc_s" for "develop/lib/libgcc_s.so.1"
Warning: Error: packaging policy violation(s) in strict mode`

Fixed this as well.

I tested, works fine with using -lstdc++fs.

File: sys-devel/gcc/gcc-8.3.0_2019_05_24.recipe [hunk 1]
```diff
@@ -326,7 +327,8 @@ INSTALL()
 		$gccLibDir/libstdc++.so.$libstdcxxSoVersion \
 		$gccLibDir/libstdc++.so.$libstdcxxLibVersion \
 		$libDir/
-	rm $gccLibDir/libstdc++*.a
+	rm $gccLibDir/libstdc++.a
+	cp $gccLibDir/libstdc++fs.a $developLibDir
```
waddlesplash: Is there no .so variant of this library?

kikadf: I didn't find, but I admit didn't check the Makefile.

kikadf: I checked the Makefile, as I see make only static stdc++fs.

korli: from https://gitlab.kitware.com/cmake/cmake/issues/17834
````
> Is libstdc++fs going to be a permanent feature, required to be linked for
> C++17 std:filesystem support, or will it be folded into the main libstdc++
> at some point?  If so, is it known when that point will be?

It will be part of libstdc++.so at some point, but we don't want to
export new symbols from the DSO until our C++17 support is complete
and stable, because once we export a symbol from the DSO we have to
keep it forever.

I expect it will be done for GCC 9.1

korli:
configure scripts are generated, they shouldn't be patched manually.

kikadf:
The originally libstdcxx-filesystem-ts changes in libstdc++-v3/acinclude.m4 doesn't work.
I didn't find a way to enable libstdcxx-filesystem-ts in libstdc++-v3/configure.ac, this is why I added first time simply the --enable-libstdcxx-filesystem-ts configure option in the recipe.

However, in the orignal patchset already patched:
configure
gcc/configure
libatomic/configure
libgomp/configure
libstdc++-v3/configure (as well)

korli:
I edited this PR and merged manually in b5df226545c6212143185949d998a052ff81a542

PR #3951, opened 2019-06-22
fifechan, fix references to libSDL*, bump version

Begasus:


Begasus:
This fixes refereces to libSDL* see issue https://github.com/haikuports/haikuports/issues/3878
Marked ! for x86_64 (new version) as they switched to libSDL2, fife is relying on this lib so fife's new version needs to be checked if it still builds here (my system isn't up for that job memory uses, 1% in the cmake build for over an hour before the system frooze here)
@mmlr I didn't apply the sdl patches you originally wrote for version 0.1.2, as 0.1.5 is using libSDL2, so not sure if they still apply?

extrowerk:
Builds packages fine on 64 bit.

Begasus:
Moved fifechan from dev-games to games-engines according to Gentoo layout

Begasus:
Disabled new version with "?", should be OK (fife doesn't complain when installing it, but rebuilding with new fifechan will fail due to changed functions)

PR #3962, opened 2019-07-03
advancemame, bump, fix build for _x86

Begasus:


Begasus:
Not checked if this works for x86_64, maybe the patch for _x86 should be in a separate patchset?
(and maybe the variables used are not the correct ones) :) waiting for reply's :)

Begasus:
Related to https://github.com/haikuports/haikuports/issues/3594?

extrowerk:
Pls remove old versions.

extrowerk:
On 64 bit:
```
/boot/system/develop/headers/os/support/SupportDefs.h:25:26: error: conflicting types for 'uint32'
typedef __haiku_uint32   uint32;
```

extrowerk:
Without the x86 patch it builds fine on 64 bit. Please move the patch into a separate patchfile and guard it with targetarch == x86 or something so it won't be applied on 64 bit.

Begasus:
@extrowerk Could you check? Moved the fix to a seperate patchset

Begasus:
Should be good to go now, if someone could check this on x86_64?

PR #3968, opened 2019-07-06
devilutionx: Bump to version 0.4.0

stenstorp:
Built and ran fine on x86_64 nightly VM.


extrowerk:
Does it works, can we merge?

stenstorp:
Worked fine on my machine, should be safe to merge.

extrowerk:
I am ok with the changes, but this is still not a bump but a downgrade as the current recipe is younger. Maybe we should just wait for the next release....

stenstorp:
This is an update from a git revision (0.3.1 with a few added commits) from the May 10th to 0.4.0, a release version released on May 20th. I'm fine waiting for a new release but what about it is "younger" than the current recipe?

extrowerk:
You have right, my bad.

PR #3970, opened 2019-07-08
maelstrom, fix reference to libSDL*

Begasus:


Begasus:
Needs to be checked on x86_64

PR #3981, opened 2019-07-13
f1spirit, fix references to libSDL_*

Begasus:


Begasus:
Maybe remove the $secondaryArchSuffix here? (quick check revealed it's not building for gcc8)

Begasus:
Needs to be checked on x86_64

PR #3989, opened 2019-07-14
fish, don't use  secondrayArchSuffix when using binDir for defineDebugInfoPackage

Begasus:


Begasus:
@extrowerk can you check on x86_64?

PR #3996, opened 2019-07-15
kbuild: update to latest revision

simonsouth:
These changes replace the existing [kBuild](https://trac.netlabs.org/kbuild) recipe with one that builds the latest revision in Subversion, 3296. (The preexisting revision, 2997, was too old to be used for building VirtualBox.)

The new recipe improves on the old one by

- Using a currently available source for the source distribution ([anonscm.debian.org is no more](https://anonscm.debian.org/)),
- Patching the distribution to bring it up-to-date with modern Haiku releases,
- Providing an original and more concise description,
- Building on x86 (which I've tested),
- Providing a debuginfo package, and
- Setting the `KBUILD_PATH` and `KBUILD_BIN_PATH` environment variables on installation so the kBuild tools can be used immediately.


waddlesplash:
Nice! Thanks, and welcome to HaikuPorts!

PR #4007, opened 2019-07-17
chromaprint: Build with fpcalc utility

phw:
This builds chromaprint with the `fpcalc` command line utility.

`fpcalc` can be used to generate fingerprints for audio files on the command line. Among others this tool is needed by MusicBrainz Picard (for which I'm going to provide a recipe later) for audio fingerprinting.

This change introduces ffmpeg as a dependency since `fpcalc` cannot be build without it. However, ffmpeg is also the preferred choice for a FFT library as described in https://github.com/acoustid/chromaprint, so the dependency on libfftw3 is no longer needed.

### DIFF
--- a/media-libs/chromaprint/chromaprint-1.4.3.recipe
+++ b/media-libs/chromaprint/chromaprint-1.4.3.recipe
@@ -59,6 +59,7 @@ BUILD()
 		-DCMAKE_INSTALL_INCLUDEDIR=$includeDir \
 		-DCMAKE_INSTALL_BINDIR=$binDir \
 		-DBUILD_TOOLS=ON \
+		-DFFT_LIB=avfft \
 		-DCMAKE_BUILD_TYPE=Release .
 
 	make $jobArgs
--- a/media-libs/chromaprint/chromaprint-1.4.3.recipe
+++ b/media-libs/chromaprint/chromaprint-1.4.3.recipe
@@ -4,7 +4,7 @@ algorithm for extracting fingerprints from any audio source."
 HOMEPAGE="https://acoustid.org/chromaprint"
 COPYRIGHT="2010-2012, 2015 Lukas Lalinsky"
 LICENSE="GNU LGPL v2.1"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/acoustid/chromaprint/releases/download/v$portVersion/chromaprint-$portVersion.tar.gz"
 CHECKSUM_SHA256="ea18608b76fb88e0203b7d3e1833fb125ce9bb61efe22c6e169a50c52c457f82"
 SOURCE_DIR="chromaprint-v$portVersion"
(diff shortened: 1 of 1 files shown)


media-libs/chromaprint/chromaprint-1.4.3.recipe
```diff
@@ -19,10 +19,13 @@ libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 PROVIDES="
 	chromaprint$secondaryArchSuffix = $portVersion
 	lib:libchromaprint$secondaryArchSuffix = $libVersionCompat
+	cmd:fpcalc
```
diversys: Shouldn't it read cmd:fpcalc$secondaryArchSuffix?

media-libs/chromaprint/chromaprint-1.4.3.recipe
```diff
@@ -52,6 +57,7 @@ BUILD()
 	cmake -DCMAKE_INSTALL_PREFIX=$prefix \
 		-DCMAKE_INSTALL_LIBDIR=$libDir \
 		-DCMAKE_INSTALL_INCLUDEDIR=$includeDir \
+		-DBUILD_TOOLS=ON \
```
Begasus: Check if it can use -DCMAKE_INSTALL_BINDIR also and maybe move the binary to a seperate _tools package?

media-libs/chromaprint/chromaprint-1.4.3.recipe
phw: Thanks, changed to `cmd:fpcalc$secondaryArchSuffix = $portVersion`

media-libs/chromaprint/chromaprint-1.4.3.recipe
phw: I added `-DCMAKE_INSTALL_BINDIR`, that is supported.

But I don't think a single 35K binary warrants the trouble for having a separate package. I would really prefer to keep this together.

Begasus: -DCMAKE_INSTALL_BINDIR isn't picked up by the project (gcc2h build), you probably would have to add a line in the CmakeList.txt file or use cmd:fpcalc without secondaryArchSuffix (as it's installed in $prefix/bin now)

phw: I only partially get what you mean here :) But I see how `-DCMAKE_INSTALL_BINDIR` gets ignored, I think it is even the same with `-DCMAKE_INSTALL_LIBDIR`. But the installed package seems to work, at least I can run `fpcalc` and it ends up in `/boot/system/bin/fpcalc`, and inside the package it is in `./bin/fpcalc` . Where should it be placed instead?

Begasus: If you build this for secondary architecture it "should" end up in $binDir
On gcc2h that would be /boot/system/bin/x86, but as you said it's now in /boot/system/bin (on primary x86_64 that is ok and you don't need the secondaryArchSuffix
PS for the lib and headers it's ok

Begasus: Created a PR to set this right :)

Issue #3822, opened 2019-05-01
Migration to Python 3.8


pulkomandy:
Hi,
I would like some clarification on how migration to python3.7 is supposed to happen.
Currently all python libs explicitly depend on version 3.6, but are suffixed with _python3 and depend on setuptools_python3.

This means we have to be careful updating them. I guess it should be something like:

PYTHON_PACKAGES=(python python3 python37)
PYTHON_VERSIONS=(2.7 3.6 3.7)

Does that sound right? And later we drop support for the "python3" suffix completely as we move on to later versions to avoid confusion?

I already hit a problem with youcompleteme. It runs "python3" as the python interpreter and depends on _python3 packages, so with the current setup, it does not work, and complains that the dependencies are not found (indeed, it is running a python 3.7 interpreter and the dependencies are in the directory for version 3.6).

waddlesplash:
I think Python 3.x breaks internal ABI across minor versions, so we can't link against it and not do rebuilds, yes?

pulkomandy:
Probably, in any case the files get installed to different directories anyway so they can't be used.
This is why "python3" does not seem to make much sense as a package suffix, and we should switch to "python37"?

korli:
waddlesplash merged the 3.7 python with no care. Please don't assign this to me.

extrowerk:
I personally doesn't like the idea to have 3 different python installed parallel, so I would decomission 3.6 alltogether, so 3.7 would become the python3.
Pro: Less changes required in the existing recipes, but they have to be revbumped to get all rebuilt against 3.7
Contra: Possible incompatibility with some existing python-module recipes, in that case if there is already a newer, 3.7 compatible version they should be version bumped.

This is a should teach us a good lesson what we can reuse if 3.8 gets released.

extrowerk:
Beause Calibre i worked much with python modules lately, so i could try to do it as i wrote in my earlier comment. Let me know if you guys are ok with that.

pulkomandy:
Sorry, I tried to find people involved but I may have misread the git log.

pulkomandy:
@extrowerk well, Python 3.6 and 3.7 are incompatible, so this is an ABI break. Just like libpng16 vs libpng15, for example. I agree we should eventually remove Python 3.6, but it seems unreasonable to do this atomically (one super-mega-huge commit changing all recipes involving Python all at once, including cooperation with non-haikuports repos).

[...]

PR #4008, opened 2019-07-17
mutagen: add recipe for version 1.42.0

phw:
This adds a recipe for python-mutagen.

Adding as a requirement for MusicBrainz Picard, for which I'll submit a recipe later.

### DIFF
--- a/dev-python/mutagen/mutagen-1.42.0.recipe
+++ b/dev-python/mutagen/mutagen-1.42.0.recipe
@@ -1,0 +1,84 @@
+SUMMARY="Python library to read and write audio tags for many formats"
+DESCRIPTION="Mutagen is a Python module to handle audio metadata. It supports \
+ASF, FLAC, MP4, Monkey's Audio, MP3, Musepack, Ogg Opus, Ogg FLAC, Ogg Speex, \
+Ogg Theora, Ogg Vorbis, True Audio, WavPack, OptimFROG, and AIFF audio files. \
+All versions of ID3v2 are supported, and all standard ID3v2.4 frames are parsed. \
+It can read Xing headers to accurately calculate the bitrate and length of MP3s. \
+ID3 and APEv2 tags can be edited regardless of audio format. It can also manipulate \
+Ogg streams on an individual packet/page level."
+HOMEPAGE="https://github.com/quodlibet/mutagen"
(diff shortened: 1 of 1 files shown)


dev-python/mutagen/mutagen-1.42.0.recipe
```diff
@@ -0,0 +1,83 @@
...
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+PYTHON_PACKAGES=(python python36)
+PYTHON_VERSIONS=(2.7 3.6)
```
extrowerk: 3.7 should be added too.

phw: Ok, I was already wondering about this. What is the actual longer term plan here? When playing with this I had some trouble since some things (e.g. pip3) use python 3.7, which is in a package called python3. Will there be only a single python3 package in the future, or will we get python36, python37, python38 etc.?

pulkomandy: We don't know, yet, see https://github.com/haikuports/haikuports/issues/3822
If you know more about Python and the possible compatibility between versions, help is welcome

phw: Good question, not sure. A new Python minor version in this case definitely requires a rebuild of all packages, but that probably is less of a mess then having packages being build against 5 different versions of Python 3.

Anyway, for now I did the changes to this package. Does this look correct:

```
PYTHON_PACKAGES=(python python36 python3)
PYTHON_VERSIONS=(2.7 3.6 3.7)
```

dev-python/mutagen/mutagen-1.42.0.recipe
```diff
@@ -0,0 +1,84 @@
+SUMMARY="Python library to read and write audio tags for many formats"
+DESCRIPTION="Mutagen is a Python module to handle audio metadata. It supports \
+	ASF, FLAC, MP4, Monkey's Audio, MP3, Musepack, Ogg Opus, Ogg FLAC, Ogg Speex, \
```
diversys: No need to ident.

phw: Fixed, also squashed commits

PR #4009, opened 2019-07-17
discid: add recipe for version 1.2.0

phw:
This adds a recipe for python-discid.

Adding as a requirement for MusicBrainz Picard, for which I'll submit a recipe later.

PR #4010, opened 2019-07-17
picard: new recipe for version 2.1.3

phw:
This is a port of MusicBrainz Picard for Haiku. Minor patches where needed to make this run on Haiku.

This requires the following merge requests to be merged:

- #4007: Adding fpcalc to chromaprint port
- #4008: python-mutagen
- #4009: python-discid

With all of the above I was able to use the full functionality of Picard, including CD lookup and audio fingerprinting.


media-sound/picard/picard-2.1.3.recipe
```diff
@@ -0,0 +1,71 @@
...
+SOURCE_URI="ftp://ftp.eu.metabrainz.org/pub/musicbrainz/picard/picard-$portVersion.tar.gz"
+CHECKSUM_SHA256="8e044fe68c44d345c19f98952c3c7f20c72da74dbac9ce7c7b0621e2d69885a7"
+SOURCE_DIR="picard-release-$portVersion"
+PATCHES="picard-$portVersion.patchset"
+ADDITIONAL_FILES="picard.rdef.in"
+
+ARCHITECTURES="x86 x86_64"
+
+PROVIDES="
+		picard = $portVersion
```
diversys: Two tabs instead of one here and below.

media-sound/picard/picard-2.1.3.recipe
```diff
@@ -0,0 +1,71 @@
+SUMMARY="MusicBrainz Picard music tagger"
+DESCRIPTION="Tag your music with the next generation MusicBrainz tagger"
+HOMEPAGE="https://picard.musicbrainz.org/"
+COPYRIGHT="2004-2019 Robert Kaye, Lukas Lalinsky, Laurent Monin, Sambhav Kothari, Philipp Wolfer and others"
```
diversys: Wrap at 80th character

media-sound/picard/picard-2.1.3.recipe
```diff
@@ -0,0 +1,71 @@
...
+	local MINOR="`echo "$portVersion" | cut -d. -f3 | cut -d~ -f1`"
+
+	sed \
+		-e "s|@MAJOR@|$MAJOR|" \
+		-e "s|@MIDDLE@|$MIDDLE|" \
+		-e "s|@MINOR@|$MINOR|" \
+		$portDir/additional-files/picard.rdef.in > $sourceDir/build/picard.rdef
+
+	rc $sourceDir/build/picard.rdef
+	resattr -o $appsDir/Picard $sourceDir/build/picard.rsrc
```
diversys: These 2 lines can be replaced with `addResourcesToBinaries  $sourceDir/build/picard.rdef`

phw: Thanks, much simpler. Done.

media-sound/picard/picard-2.1.3.recipe
phw: Done

media-sound/picard/picard-2.1.3.recipe
phw: Done

media-sound/picard/picard-2.1.3.recipe
phw: I reverted this in caf4b343e5c7b3cc794fef15e7121448fae26660 . I don't think addResourcesToBinaries works here, the executable is a Python script. addResourcesToBinaries destroyed it and turned it into binary garbage.

I took the `rc` + `resattr` combo from another recipe which also had a script based launcher. I don't fully understand the background yet how this is handled on Haiku, but it seems using `resattr` the details are stored as file system attributes instead of embedding them into the binary file. If there is a better way to handle that, let me know.

diversys: Ah, yes. addResourcesToBinaries on works with, well, binaries. So `rc` + `resattr` combo is good enough.

phw:
@diversys Could you look at what I did in 3f609d01bb0872212dd042afc174761bf46ba43d . What this fixes is the deskbar showing the proper application icon and name. Without this change the menu would show the icon, but once started the running application would be shown with the generic cubes icon and the name "python3.6m". 

grafik

With `settype -t application/x-vnd.Be-elfexecutable $appsDir/Picard` the application is shown with proper icon and name.

grafik

I stole this change from the [OpenShot recipe](https://github.com/haikuports/haikuports/blob/9a10c5e06c44211c55fba8b65b7f01086f8a9912/media-video/openshot/openshot-2.4.4.recipe#L78), and while it works it seems rather hackish to me. What other implications could it have to set a Python script to `application/x-vnd.Be-elfexecutable`? Clearly this is no ELF executable :D

diversys:
This is a limitation of Deskbar API I think, so the only way around for now is to use hacks like this.

phw:
Ok, thanks for the quick response. I tested this today and actually used the application and I from my side this is good to go once #4008 has been merged.

PR #4018, opened 2019-07-19
picard: add patch to use Haiku Qt style

phw:
As suggested in https://discuss.haiku-os.org/t/musicbrainz-picard-for-haiku/8759

### DIFF
--- a/media-sound/picard/patches/picard-2.1.3.patchset
+++ b/media-sound/picard/patches/picard-2.1.3.patchset
@@ -73,3 +73,28 @@ index 7d003ccf..bacd8a13 100644
          webbrowser.open(url)
 --
 2.21.0
+
+From b9e96ea8fd0907c797b2424986023941ee8e1474 Mon Sep 17 00:00:00 2001
+From: Philipp Wolfer <ph.wolfer@gmail.com>
+Date: Fri, 19 Jul 2019 10:37:17 +0000
+Subject: [PATCH 4/4] Haiku: use Haiku Qt style
+
+---
+ picard/tagger.py | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/picard/tagger.py b/picard/tagger.py
+index f3f0ff85..daf76a41 100644
+--- a/picard/tagger.py
++++ b/picard/tagger.py
+@@ -130,7 +130,7 @@ class Tagger(QtWidgets.QApplication):
+
+         # Use the new fusion style from PyQt5 for a modern and consistent look
+         # across all OSes.
+-        if sys.platform != "darwin":
++        if sys.platform not in ("darwin", "haiku1"):
+             self.setStyle('Fusion')
+
+         # Set the WM_CLASS to 'MusicBrainz-Picard' so desktop environments
+--
+2.21.0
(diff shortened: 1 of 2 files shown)


diversys:
Thanks! However, it looks like there is a missing dependency.

```
~> /system/apps/Picard 
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/picard/disc.py", line 24, in <module>
    from libdiscid.compat import discid
ModuleNotFoundError: No module named 'libdiscid'

```

phw:
@diversys This is strange. Actually I think this might be a Picard bug (ate least insufficient test).

There are two modules Picard can use for using libdiscid: python-libdiscid and python-discid. I packaged python-discid, but your system is failing on importing python-libdiscid.

Picard should actually catch this, but it only checks for `ImportError`. I wonder under which condition a `ModuleNotFoundError` is raised. Can you check whether you have `python-libdiscid` installed? Maybe separately using pip?

phw:
> So the recipe is missing it.

No it isn't, since it shouldn't be required. python-discid is provided for that, I also don't have python-libdiscid installed. For me the question here is why it doesn't catch the `ModuleNotFoundError` on your installation. I checked, and `ModuleNotFoundError` is derived from `ImportError` which gets catched.

Could you check the output of:

    python3.6 -c "print('%r, %r' % (ModuleNotFoundError.__bases__, isinstance(ModuleNotFoundError(), ImportError)))"

Expected would be `(<class 'ImportError'>,), True`.

diversys:
Yep, this is what I get as well.

diversys:
Here is the full output just in case:

```
/system/apps/Picard 
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/picard/disc.py", line 24, in <module>
    from libdiscid.compat import discid
ModuleNotFoundError: No module named 'libdiscid'

During handling of the above exception, another exception occurred:

  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/discid/libdiscid.py", line 99, in _open_library
    return ctypes.cdll.LoadLibrary(lib_name)
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/ctypes/__init__.py", line 426, in LoadLibrary
    return self._dlltype(name)
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/ctypes/__init__.py", line 348, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: No such file or directory

During handling of the above exception, another exception occurred:

  File "/system/apps/Picard", line 2, in <module>
    from picard.tagger import main; main('/packages/picard-2.1.3-3/.self/data/share/locale', False)
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/picard/tagger.py", line 66, in <module>
    from picard.disc import Disc
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/picard/disc.py", line 28, in <module>
    import discid
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/discid/__init__.py", line 30, in <module>
    from discid.disc import read, put, Disc, DiscError, TOCError
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/discid/disc.py", line 24, in <module>
    from discid.libdiscid import _LIB, FEATURES
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/discid/libdiscid.py", line 108, in <module>
    _LIB = _open_library(_LIB_NAME)
  File "/packages/python36-3.6.8-2/.self/lib/python3.6/vendor-packages/discid/libdiscid.py", line 103, in _open_library
    raise OSError("could not find libdiscid: %s" % lib_name)
OSError: could not find libdiscid: libdiscid.so.0
```

phw:
Thanks, this makes more sense. `libdiscid` (the C library) itself was missing, it's a problem with the python-discid package. Proposed fix at https://github.com/haikuports/haikuports/pull/4022

Actually Picard isn't supposed to crash in this case either, but we already had a bugfix for this sepcific case recently.

PR #4030, opened 2019-07-21
mog, fix references to libSDL*, enable x86_64

Begasus:


Begasus:
builds ok with gcc8 on x86_gcc2, needs to be tested on x86_64 but should be ok?

PR #4036, opened 2019-07-24
swars, move to libpng16, fix reference to libSDL

Begasus:


Begasus:
Checked the build, now builds with libpng16 (iirc swars was the only recipe still using libpng15)

Begasus:
x86_64 should be marked broken (!) as it will only build for i386 systems?

Begasus:
If this turns out to be ok (looks ok here) do we still need libpng15 in the tree?

PR #4078, opened 2019-08-09
drive_encryption, fix build

Begasus:
Disabled build because it KDL'd after latest update when installed ...


Begasus:
Related to https://github.com/haikuports/haikuports/issues/3594

PR #4082, opened 2019-08-10
freesci, remove obselete patch, freesci is imported in scummvm now

Begasus:


Begasus:
Well it has been for some years :)

PR #4089, opened 2019-08-14
ghostwriter, bump version

Begasus:


Begasus:
Now that hunspell is merged, this can be too :) (latest version 1.8.0 requires qtwebengine which I don't think is available yet (atleast not found when trying to build from Terminal))

PR #4096, opened 2019-08-14
enchant, bump version, disable for failing test

Begasus:


Begasus:
Merging as there is no real need for it atm, could always be picked up in case :)

PR #4099, opened 2019-08-15
goldendict, revbump after hunspell bump, fix path to the myspell dict…

Begasus:
…ionaries


Begasus:
Not tested on x86_64 here

PR #4100, opened 2019-08-15
psi_plus, revbump after hunspell bump

Begasus:


Begasus:
Not checked on x86_64 here

PR #4124, opened 2019-08-23
PRapidJSON: pkgconfig fix

extrowerk:
No idea, why it is happening.

Fixes https://github.com/haikuports/haikuports/issues/4123


extrowerk:
Meh, this is not ok.

extrowerk:
Seems to be a working fix

PR #4125, opened 2019-08-23
Restore libutf8proc 1.x

extrowerk:
Mame depends on it


waddlesplash:
The utf8proc documentation specifically states that 2.x is source-compatible, only ABI-incompatible. What happens with mame?

extrowerk:
The 2.x's pkgconfig file is broken because fixpkgconfig so mame can't find the headers. Made an issue about it at haikuporter repo: https://github.com/haikuports/haikuporter/issues/189

korli:
I've something local to switch libutf8proc2 to the other upstream https://juliastrings.github.io/utf8proc/
and fixes for netsurf to use it. I'll close this PR afterwards.

korli:
superseded by 52c38393a34416285456bdd3e67c66f5e0dc0355

PR #4126, opened 2019-08-23
Update to Roman-Numbers

cnelmortimer:
Moving the download target of the RomanNumers package to a newer version with a minor bug-fix.


haiku-apps/romannumbers/romannumbers-1.0~git.recipe
```diff
@@ -10,9 +10,9 @@ HOMEPAGE="https://github.com/cnelmortimer/RomanNumbers"
 COPYRIGHT="2019 N.C. Cruz"
 LICENSE="MIT"
 REVISION="1"
```
waddlesplash: This number needs to be bumped.

cnelmortimer: Fixed... but I finally forgot adding "--amend" when committing so the PR might be outdated. Do I need to fix that (how?) or the PR automatically goes to the latest commit?

waddlesplash:
The PR automatically goes to the latest commit, so it looks fine here.

PR #4129, opened 2019-08-24
libOpenMPT: bump

extrowerk:
Patch upstreamed.
Test passes, plays music. Even rebuilt ffmpeg with it (not needed, just to make sure)


extrowerk:
https://github.com/OpenMPT/openmpt/pull/4

PR #4130, opened 2019-08-24
picard: build 32bit package

phw:
This is an attempt to make picard available as a 32 bit package.

@Begasus Can you take a look? To be honest I cannot fully wrap my head around this secondary architecture thing and I cannot find any documentation about it. I just copied this from other recipes.


media-sound/picard/picard-2.1.3.recipe
```diff
@@ -4,35 +4,36 @@ HOMEPAGE="https://picard.musicbrainz.org/"
...
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
-	picard = $portVersion
+	picard${secondaryArchSuffix} = $portVersion
 	cmd:picard = $portVersion
 	"
 REQUIRES="
-	haiku
+	haiku${secondaryArchSuffix}
```
Begasus: Remove the brackets:
```diff
-	haiku${secondaryArchSuffix}
+	haiku$secondaryArchSuffix
```

media-sound/picard/picard-2.1.3.recipe
```diff
@@ -4,35 +4,36 @@ HOMEPAGE="https://picard.musicbrainz.org/"
...
 	cmd:python3.6
-	pyqt_python3
+	pyqt_python3${secondaryArchSuffix}
```
Begasus: just pyqt_python3 will do (no need for secondaryArchSuffix)
PS we only use the brackets for a few things :)

media-sound/picard/picard-2.1.3.recipe
```diff
@@ -4,35 +4,36 @@ HOMEPAGE="https://picard.musicbrainz.org/"
...
 
 BUILD_REQUIRES="
-	haiku_devel
+	haiku${secondaryArchSuffix}_devel
 	cmd:gcc
 	cmd:msgfmt
 	cmd:python3.6
 	setuptools_python36
-	pyqt_python3
+	pyqt_python3${secondaryArchSuffix}
```
Begasus: Same here, without the secondaryArchSuffix

Begasus:
FWIW I pushed my changes to build it on gcc2h to https://github.com/Begasus/haikuports/commit/70a8099f9561fe94fe584281912498823a3080f5

Begasus:
Also left a comment on the chromaprint recipe https://github.com/haikuports/haikuports/pull/4007/commits/1c1a4d93b17b849f40029a1accd6727d1f6cf599#r317354235

phw:
@Begasus Thanks for this, I replaced my commit with yours for this PR. Thanks for your clarification on the architectures also. What do the prefixes `!` and `?` in `!x86_gcc2 ?x86` do? Is `!` "don't build" or priority?

phw:
Also tested on x86_64, still works ;)

media-sound/picard/picard-2.1.3.recipe
```diff
@@ -4,35 +4,38 @@ HOMEPAGE="https://picard.musicbrainz.org/"
...
-	haiku_devel
-	cmd:gcc
-	cmd:msgfmt
-	cmd:python3.6
+	haiku${secondaryArchSuffix}_devel
 	setuptools_python36
-	pyqt_python3
+	"
+BUILD_PREREQUIRES="
+	cmd:fpcalc$secondaryArchSuffix
```
phw: @Begasus I don't think we need this for build, or was there a specific reason you added it?

Begasus: It was already in there (but as a left the comment on chromaprint, cmd:fpcalc isn't provided in the recipe, but the binary is in $prefix/bin so cmd:fpcalc is correct), for now I used this to bypass the problem, haven't checked if it's really needed

Begasus:
! tells haikuporter the build is broken for that architecture ? tells haikuported the build hasn't been tested (for ?x86 it's ok, we don't have a release for x86 haiku version (gcc8 only))

phw:
> ! tells haikuporter the build is broken for that architecture ?

See how this is confusing me to no end :D So we get a build for gcc2h by telling the system the build for x86_gcc2 is broken? Why?

Begasus:
> > ! tells haikuporter the build is broken for that architecture ?
> 
> See how this is confusing me to no end :D So we get a build for gcc2h by telling the system the build for x86_gcc2 is broken? Why?

Because x86_gcc2 is the primary target (using gcc2) which can't build Picard :) hence it's broken "!"

media-sound/picard/picard-2.1.3.recipe
Begasus: @phw correct not needed so can be removed in the recipe

Begasus:
LGTM only remove the requirement for fpcalc :) (after you tested on x86_64) :)

phw:
> LGTM only remove the requirement for fpcalc :) (after you tested on x86_64) :)

fpcalc removed from `BUILD_PREREQUIRES` and build tested on x86_64. Ready for merging then.

PR #4132, opened 2019-08-24
GCC size reduction

extrowerk:


extrowerk:
Should work but had no time yet to rebuild it.

PR #4133, opened 2019-08-25
amoebax, fix references to libSDL*, add icon, fix config path

Begasus:


Begasus:
Needs to be checked on x86_64

Begasus:
Checked and working ok on x86_64 :)

PR #4134, opened 2019-08-25
FFMPEG: switch to OpenSSL

extrowerk:
Be warned, OpenSSL needs the --enable-nonfree configure switch!

### DIFF
--- a/media-video/ffmpeg/ffmpeg-4.2.recipe
+++ b/media-video/ffmpeg/ffmpeg-4.2.recipe
@@ -7,7 +7,7 @@ COPYRIGHT="2000-2003 Fabrice Bellard
 	2003-2019 the FFmpeg developers"
 LICENSE="GNU LGPL v3
 	GNU GPL v3"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://ffmpeg.org/releases/ffmpeg-$portVersion.tar.xz"
 CHECKSUM_SHA256="023f10831a97ad93d798f53a3640e55cd564abfeba807ecbe8524dac4fedecd5"
 PATCHES="ffmpeg-$portVersion.patchset"
--- a/media-video/ffmpeg/ffmpeg-4.2.recipe
+++ b/media-video/ffmpeg/ffmpeg-4.2.recipe
@@ -63,8 +63,8 @@ REQUIRES="
 	"
 if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
 	REQUIRES+="
-		lib:libssl$secondaryArchSuffix
 		lib:libopenmpt$secondaryArchSuffix
+		lib:libssl$secondaryArchSuffix
 		lib:libvpx$secondaryArchSuffix
 		"
 fi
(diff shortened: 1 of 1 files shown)


extrowerk:
Without nonfree configure fails with:
```
Building ...
openssl is incompatible with the gpl and --enable-nonfree is not specified.

If you think configure made a mistake, make sure you are using the latest
version from Git.  If the latest version fails, report the problem to the
ffmpeg-user@ffmpeg.org mailing list or IRC #ffmpeg on irc.freenode.net.
Include the log file "ffbuild/config.log" produced by configure as this will help
solve the problem.
Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 1
Error: Build has failed - stopping.```

waddlesplash:
We definitely cannot --enable-nonfree.

What parts of FFmpeg are GPL and will be disabled by this? I'm not so sure we should have FFmpeg GPL enabled...

extrowerk:
I cant answer your question.
Do we need ssl/tls in ffmpeg at all?

pulkomandy:
I had enabled GPL code to experiment with their rescaling/colorspace conversion code, which I think has since then been replaced by something with a different license?

As for SSL, it was added so qmplay2 can play youtube videos directly. For the use in Haiku MEdia Kit we don't care, but that's not the only place ffmpeg is needed.

extrowerk:
according to ffmpeg guys openssl switched to apache license in dec 2018, so if our openssl newer than that (and it is, 1.0.2s release in may. 2019) then we should not need --enable-nonfree
the configure script doesn't got updated yet
https://www.openssl.org/source/license.html
according to openssl, only 3.0.0 release is apache v2
 so 1.x is still not compatible with GPL
yeah, just checked the commits that went into the latest 1.0.2 and 1.1.0 and 1.1.1 branches, and none of them had the new license"
 "openssl3.0.0 hasn't been released yet afaik - they're referring to current git master"

korli:
Not sure where the problem really is: if there are errors in some libraries, just report upstream. To me "we should not have multiple TLS libraries if we can avoid it" makes no sense.

pulkomandy:
It would be nice if the core Haiku package and dependencies (that includes ffmpeg and webkit, for example) would bring in only one SSL implementation.
That does not means the other libraries should be removed from Haikuports, just that the base system should stay small and with as few dependencies as possible.

Unfortunately, WebKit also depends on GnuTLS, so we have no choice unless we add an OpenSSL backend for Web Crypto there.

This also doesn't mean GnuTLS shouldn't be fixed if it's broken, it's of course needed by other things in HaikuPorts.

waddlesplash:
WebKit only needs libgcrypt, not gnutls, which is a different matter. It appears FFmpeg has actually not much GPL code: https://www.ffmpeg.org/doxygen/4.0/md_LICENSE.html

So, I'll try using only LGPLv3 and see what happens.

extrowerk:
OpenSSL needs the nonfree switch, but it should not enable other functionality (according to the FFMPEG IRC).
I can patch the check out, but i dont think this is what you meant. Care to explain?

-- Z.

2019.09.02. dátummal, 17:08 időpontban waddlesplash <notifications@github.com> írta:

waddlesplash:
When GPL is not enabled, OpenSSL works without the nonfree switch. We likely should not ship GPL FFmpeg anyway, so that commercial software can link against it.

It appears mplayer and qtav currently depend on libpostproc, though...

waddlesplash:
https://github.com/haikuports/haikuports/commit/8b7fe74322f708090d57dae4f7a26f3bacf51d80

PR #4159, opened 2019-09-03
qutebrowser: bump version

kenmays:
UPDATE: QuteBrowser 1.8.1, Tested on hrev53511.


kenmays:
Updated. Good to merge.

extrowerk:
Builds, packages fine on 64 bit

Begasus:
ninja_python3 can't be installed due to missing package for markupsafe_python3 for x86_gcc2

kenmays:
qutebrowser-1 7 0-haiku_cocobeaan
Qutebrowser 1.7.0 on x86. Python 3.6 downloads MarkupSafe 1.1.1 w/Jinja.

Begasus:
Looks good, just how did you get MarkupSafe 1.1.1 installed? Couldn't find a recipe for it in Haikuports

kenmays:
Src imports from the Net if not in the repo, if a specific version is defined.

Begasus:
Maybe if you build from Terminal? haikuporter will pull the package from the depot, anyway, created PR's to bump both jinja and markupsafe to bring them back on par :)
Mind you jinja3 will point to python3.7, if you need python3.6 use jinja36 then (same for markupsafe36)

Begasus:
markupsafe and jinja have been merged, with a small change to the recipe (as mentioned in my earlier comment) it builds and installs fine, only thing is that I have to enable QTWebkit backend to start it up and got this warning:
QuteBrowser
Other then that it starts up without the warning (even could watch youtube movie with it) :+1:

www-client/qutebrowser/qutebrowser-1.7.0.recipe
```diff
@@ -27,7 +27,7 @@ REQUIRES="
 	pypeg2_python3
 	pyqt_python3
 	pyyaml_python3
-	setuptools_python3
+	setuptools_python36
```
Begasus: This is also needed for jinja_python3 vs jinja_python36 :)

Begasus:
With my build the binaries go into $prefix/bin instead of $binDir, so a few changes needed:
1 missing: cmd:pygmentize (in PROVIDES)
2 change $binDir to $prefix/bin (doesn't change anything on x86_64): mv $prefix/bin/qutebrowser $appsDir/QuteBrowser (in INSTALL)

www-client/qutebrowser/qutebrowser-1.7.0.recipe
kenmays: Resolved

www-client/qutebrowser/qutebrowser-1.8.1.recipe
```diff
@@ -1,13 +1,13 @@
 SUMMARY="A VIM-like web browser"
 DESCRIPTION="QuteBrowser is a keyboard-focused browser with a minimal GUI. \
-It’s based on Python and PyQt5 and free software, licensed under the GPL. \
+Itâ€™s based on Python and PyQt5 and free software, licensed under the GPL. \
```
diversys: Typo?

kenmays: fixed

PR #4172, opened 2019-09-10
picard: add back dependency on cmd:fpcalc

phw:
Add back the dependency for `cmd:fpcalc` to the recipe. This got lost in the last update, but actually is required for the fingerprinting to work.

### DIFF
--- a/media-sound/picard/picard-2.1.3.recipe
+++ b/media-sound/picard/picard-2.1.3.recipe
@@ -28,12 +28,12 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libdiscid$secondaryArchSuffix
-	cmd:python3.6
-	pyqt_python3
-	mutagen_python36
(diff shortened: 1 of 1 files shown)


media-sound/picard/picard-2.1.3.recipe
```diff
@@ -33,6 +33,7 @@ REQUIRES="
 	pyqt_python3
 	mutagen_python36
 	discid_python36
+	cmd:fpcalc
```
diversys: Please sort this section alphabetically.

phw: Ok, did this, but kept the haiku dependency at the top because this is what I usually see in other recipes. Is this how it is expected? Not sure if I like this the best currently. What about some rule like:

1. First list haiku[_devel]
2. Then list dependencies with prefixes (`cmd:`, `lib:` and probably others?) alphabetically
3. List all other dependencies alphabetically

PR #4175, opened 2019-09-14
picard: Final 2.2.0 release

phw:
Picard 2.2 has just been released: https://github.com/metabrainz/picard/releases/tag/release-2.2.0

This adds the recipe for this version.

Also notable: Haiku is among the first to receive the new package! Official release announcement will be made tomorrow (we had some build delays today), also the website will be updated then and will feature Haiku on the download page.


diversys:
You probably want to rename old recipe to new one. Does the new version need patches as well?

phw:
@diversys I'm not familiar with the policies of how old recipes are treated. So I should just remove the old recipe then?



> Does the new version need patches as well?

Yes, they are supplied in https://github.com/haikuports/haikuports/pull/4175/files#diff-bf70196a3abbe742facd90481b76b5aa

diversys:
> @diversys I'm not familiar with the policies of how old recipes are treated. So I should just remove the old recipe then?

Yep.

phw:
Here we go.

waddlesplash:
Looks good! 🎊

PR #4176, opened 2019-09-15
markupsafe, bump version

Begasus:


dev-python/markupsafe/markupsafe-1.1.1.recipe
```diff
@@ -2,14 +2,15 @@ SUMMARY="XML/HTML/XHTML Markup safe string for Python"
...
-SOURCE_URI="https://pypi.io/packages/source/M/MarkupSafe/MarkupSafe-$portVersion.tar.gz"
-CHECKSUM_SHA256="a6be69091dac236ea9c6bc7d012beab42010fa914c459791d627dad4910eb665"
-SOURCE_DIR="MarkupSafe-$portVersion"
+SOURCE_URI="https://github.com/pallets/markupsafe/archive/1.1.1.tar.gz"
+CHECKSUM_SHA256="222a10e3237d92a9cd45ed5ea882626bc72bc5e0264d3ed0f2c9129fa69fc167"
+SOURCE_FILENAME="markupsafe-$portVersion.tag.gz"
+SOURCE_DIR="markupsafe-$portVersion"
 
-ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="x86_gcc2 ?x86 x86_64"
```
Begasus: Maybe this could be set to "any"?

korli: I think the package includes some kind of binary module, arch specific.

Begasus: Ok, then it can stay as it is, I'm not touching this for a secondaryArch (doubt it would be needed?) :)
This should be merged before  jinja :)

PR #4199, opened 2019-09-23
python: patched webbrowser module for Haiku

phw:
This patches the Python webbrowser module to open the default browser on Haiku.

With this patch code like this actually works and opens the browser:

```python
import webbrowser
webbrowser.open('https://haiku-os.org')  # Will return True and open browser
```

Patch for all currently supported Python versions included (2, 3.6, 3.7). The code was previously already used in the patches for https://github.com/haikuports/haikuports/tree/08258e72b706303a4c61c72b370d670fdf8cd297/media-sound/picard/patches


phw:
Tested manually with all Python versions with locally built packages..

PR #4206, opened 2019-09-28
pfetch: new recipe

Crestwave:


app-misc/pfetch/pfetch-0.1.recipe
```diff
@@ -0,0 +1,29 @@
...
+	"
+REQUIRES="
+	haiku
+	"
+
+INSTALL()
+{
+	install -d "$binDir" "$docDir"
+	install -m 755 pfetch "$binDir"
+	install -m 644 README.md "$docDir"
```
Begasus: Maybe install in $docDir/pfetch?

Crestwave: Doesn't `$docDir` already include the port name in the path?

Begasus: My bad, didn't check the recipe as I think it looks ok, if the subfolder is created it's ok as it is :)

Issue #1261, opened 2017-03-29
Missing Provide for lib:libgomp


CodeforEvolution:
There are recipes that depend on lib:libgomp, but nothing provides it. Seemingly, gcc should provide it as it seems gcc supposedly should provide libgomp.

waddlesplash:
Yeah, HaikuPorter's handling of gcc syslibs is kinda bogus and hacky, which is my fault. I should look into fixing that one of these days.

korli:
It's actually provided by the gcc_syslibs package:
https://github.com/haikuports/haikuports/blob/master/sys-devel/gcc/gcc-5.4.0_2016_06_04.recipe#L393

CodeforEvolution:
Oops, didn't catch that. Though on the x86 version of Haiku, no package provides devel:libgomp.

korli:
Right, the gcc packages weren't updated since then. Any recipe needs devel:libgomp? They usually need lib:libgomp only.

Begasus:
Looks like there is no gcc_syslibs (and therefore no libgomp) for x86_gcc2

Begasus:
> Right, the gcc packages weren't updated since then. Any recipe needs devel:libgomp? They usually need lib:libgomp only.

fluidsynth needs one for x86_gcc2 it seems? https://github.com/haikuports/haikuports/blob/0975f29d4efcb1130fd8873bb417c8b63a368413/media-sound/fluidsynth/fluidsynth2-2.0.7.recipe#L39

PR #4215, opened 2019-10-01
fluidsynth, there is no gcc_syslibs package for x86_gcc2, hence no li…

Begasus:
…bgomp


Begasus:
Related to https://github.com/haikuports/haikuports/issues/1261

Begasus:
This is also needed as there is no fluidsynth x86_gcc2 package atm for sdl2_mixer (which also need a revbump after this)

PR #4218, opened 2019-10-01
Libjpeg_turbo:bump

extrowerk:


extrowerk:
All test passed on 64 bit

PR #4221, opened 2019-10-02
Setuptools: bump

extrowerk:
Maybe this could create regressions with older ports.


waddlesplash:
Possible, but setuptools is pretty well tested. So hopefully they would have already worked out regressions. We can bump revisions as needed.

PR #4223, opened 2019-10-02
aiodns: bump

extrowerk:
introduce versions


extrowerk:
Probably broken, as it requires non-exxisting module, but it was broken before the bump too.

korli:
I merged manually.

PR #4224, opened 2019-10-02
Alembic: bump, rename, pyversions

extrowerk:


extrowerk:
Probably broken, as it requires non-exxisting module, but it was broken before the bump too.

korli:
should cmd:alembic be provided in main package?

Begasus:
> should cmd:alembic be provided in main package?

If not it could be moved to a _tools package?

korli:
> If not it could be moved to a _tools package?

I mean, it probably doesn't exist.

Begasus:
Just checked on the locale packages, I have a binary for alembic{2.7,3.6,3.7}

Begasus:
Running alembic3.7 -h
```
~> alembic3.7 -h
usage: alembic3.7 [-h] [-c CONFIG] [-n NAME] [-x X] [--raiseerr]
                  {branches,current,downgrade,edit,heads,history,init,list_templates,merge,revision,show,stamp,upgrade}
                  ...

positional arguments:
  {branches,current,downgrade,edit,heads,history,init,list_templates,merge,revision,show,stamp,upgrade}
    branches            Show current branch points.
    current             Display the current revision for a database.
    downgrade           Revert to a previous version.
    edit                Edit revision script(s) using $EDITOR.
    heads               Show current available heads in the script directory.
    history             List changeset scripts in chronological order.
    init                Initialize a new scripts directory.
    list_templates      List available templates.
    merge               Merge two revisions together. Creates a new migration
                        file.
    revision            Create a new revision file.
    show                Show the revision(s) denoted by the given symbol.
    stamp               'stamp' the revision table with the given revision;
                        don't run any migrations.
    upgrade             Upgrade to a later version.

optional arguments:
  -h, --help            show this help message and exit
  -c CONFIG, --config CONFIG
                        Alternate config file
  -n NAME, --name NAME  Name of section in .ini file to use for Alembic config
  -x X                  Additional arguments consumed by custom env.py
                        scripts, e.g. -x setting1=somesetting -x
                        setting2=somesetting
  --raiseerr            Raise a full stack trace on error
```

korli:
so there is no cmd:alembic?

Begasus:
I guess you are right in some point ... but why doesn't "haikuporter -S" give an error on this? (should be something like "cmd:alembic$PYTHON_VERSION")?

korli:
haikuporter gives warnings for missing requires, but not missing binaries :)

Begasus:
Still need to check, but looking closely in the recipe the cmd should be covered in line: https://github.com/haikuports/haikuports/blob/52e89c0ada43ed1dcdd6313121044e2fa6d0d5ae/dev-python/alembic/alembic_py-1.2.1.recipe#L40

Begasus:
Ok, this is working as you suggested @korli no need for cmd:alembic in the main package, on the other hand (seen this also with other python packages), the main package is empty and only the ones for python{2.7,3.6,3.7) contain the actual content, isn't there a way to not create the main package for these?

korli:
should be added as dependencies:
dateutil
editor
mako
sqlalchemy

Begasus:
Had to install/build them locally also, haven't checked if other dependencies are needed though ...

PR #4232, opened 2019-10-02
SIP: bump, pyversion

extrowerk:
WAIT, this will break all pyqt based stuff.
Let me finish the pyqt recipes too.

PR #4233, opened 2019-10-02
Pyqt 5.13.1

extrowerk:
Bump, switched to python 3.7
No PyQt for Python 3.6, one version should be enough.

This recipe depends on https://github.com/haikuports/haikuports/pull/4232


extrowerk:
Tested, seems to be working with Python2 and 3

PR #4252, opened 2019-10-02
GitDB: switch to p3.7

extrowerk:
Needs smmap


korli:
this will impact gertty right?
https://github.com/haikuports/haikuports/blob/master/dev-util/gertty/gertty-1.5.0.recipe

korli:
bumped to 2.0.6 in c4f6888efe3e38df506893ddc3cdcd6ef2eba52a
gertty is probably broken.

extrowerk:
Sorry, korli, i didn't had too much time lately. Thanks for the update and for cleaning up the python recipes.

PR #4268, opened 2019-10-04
six, bump, add pyversions

Begasus:


dev-python/six/six-1.12.0.recipe
```diff
@@ -41,6 +41,10 @@ BUILD_PREREQUIRES="$BUILD_PREREQUIRES
 	cmd:python$pythonVersion"
 done
 
+REPLACES_python="
+	python_six
+	"
```
korli: Any reason to add a REPLACES_python? If none, please remove as always.

Begasus: Haven't been checking up on these, I used the same base as for the other recipes, as far as I can see they build ok for all python versions?

korli: Did python_six ever exist?

Begasus:
Was on my way but you beat me to it :) :+1:

PR #4270, opened 2019-10-04
editor, new recipe

Begasus:


Begasus:
This should fix the missing module for Alembic https://github.com/haikuports/haikuports/pull/4224

PR #4283, opened 2019-10-09
picard: release 2.2.2

phw:
New upstream release 2.2.2, drop files not needed for Haiku from package.

https://blog.metabrainz.org/2019/10/08/picard-2-2-2-released/


media-sound/picard/picard-2.2.2.recipe
```diff
@@ -81,5 +81,6 @@ INSTALL()
 	resattr -o $appsDir/Picard $sourceDir/build/picard.rsrc
 	addAppDeskbarSymlink $appsDir/Picard
 
-	rm -rf $dataDir/applications
+	rm -rf $dataDir/share/applications
+	rm -rf $dataDir/share/icons
```
Begasus: Can't check atm, but is $dataDir/"share" used by anything else? Maybe the path should be changed to $dataDir/"picard"?
On another note, you could combine them in one line:
rm -rf $dataDir/share/{applications,icons}

phw: Maybe I misunderstand, but I thouhjt `$dataDir` points to the datadir which contains the build results and which is ultimately included in the package. What else should there be other than what got build?

The alternative would be to list all files individually.

Begasus: Don't get me wrong, but "share" isn't a directory commonly used in Haiku, it's not a problem sitting in $dataDir though

phw: Ok, now I get it. I first thought you were concerned the rm commands would remove something unrelated to the package.

Yes, `$dataDir/share` also contains the locales (in `locale` directory). We can relocate them, where would they usually get placed on Haiku?

Begasus: locales can go to $localeDir :)

phw: Much cleaner, I changed that. I did not have success with $localeDir though when building locally, that seemed to be empty. I set `$dataDir/locale` now directly, is this ok?

Begasus: I'll try to see if I can around to it tomorrow (getting late here), thanks for looking into this!

phw: Thanks a lot of checking and the feedback.

As a side note, testing the language selection on Haiku I discovered and fixed a general issue with our default locale detection on POSIX systems. Once again this shows that there is value in running on different systems :)

PR #4297, opened 2019-10-15
libretro: a handful of core slight changes

kwyxz:
I am in the process of automating the build process of the libretro cores I maintain. In order to do this, I need to harmonize the versions of the currently existing cores. These three were created before we reached a consensus in versioning format. Renaming them to allow the automated build process in a near future.


kwyxz:
Dammit this caught too many changes. Closing and will recreate.

PR #4317, opened 2019-10-28
wxqt: Bump version to 3.1.3

return:
This updates wxqt to version 3.1.3.

Compiles fine on x86_64 but is a WIP to make sure that existing applications don't break with this update.

Please mark this as a WIP until apps/libraries using wxqt are tested to function correctly.


diversys:
The only users of wxqt are far and audacity both of which are disabled. So it's ok to merge as is.

korli:
well the old version is to be removed.

PR #4320, opened 2019-10-30
lua, bump version

Begasus:


Begasus:
Recipe should be OK, checked on gcc2h with x86_gcc2 and _x86
I see other systems using libreadline to link to also, maybe it can be added? (works also)

Begasus:
Maybe we could rename this to lua54 for testing purpose?

dev-lang/lua/lua-5.4.0.recipe
```diff
@@ -0,0 +1,103 @@
...
+	PATCHES+="
+		lua-$portVersion.gcc2.patchset
+	"
+fi
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+PROVIDES="
+	lua$secondaryArchSuffix = $portVersion compat >= 5.4
```
Begasus: Do we need a compat here?

dev-lang/lua/lua-5.4.0.recipe
```diff
@@ -0,0 +1,103 @@
...
+	cmd:lua$secondaryArchSuffix = $portVersion compat >= 5.4
+	cmd:lua5.4$secondaryArchSuffix = $portVersion compat >= 5.4
+	cmd:luac$secondaryArchSuffix = $portVersion compat >= 5.4
+	cmd:luac5.4$secondaryArchSuffix = $portVersion compat >= 5.4
+	lib:liblua$secondaryArchSuffix = $portVersion compat >= 5.4
```
Begasus: We could start using libVersionCompat here, but would it be worth the effort?

Begasus:
OK, moved the recipe name to lua54-5.4.0~beta.recipe (and the patches, still have the previous ones around) and added variables that should make bumping 5.4.* version easier (IMHO)
Some variables will need changes once stable is released, at least this wont interfere with our default 5.3.5 version ...
Reviews welcome :)

dev-lang/lua/lua54-5.4.0~beta.recipe
```diff
@@ -0,0 +1,110 @@
...
+REQUIRES_devel="
+	lua54$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:gcc$secondaryArchSuffix
```
korli: cmd:make comes after cmd:gcc

Begasus: Fixed (this should be done for lua 5.3.5 also) :) (leftover from there)

dev-lang/lua/lua54-5.4.0~beta.recipe
```diff
@@ -0,0 +1,110 @@
...
+	"
+
+PROVIDES_devel="
+	lua54${secondaryArchSuffix}_devel = $portVersion
+	devel:liblua$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	lua54$secondaryArchSuffix == $portVersion base
+	"
+
```
Begasus: Shouldn't we add a CONFLICTS here (like in lua52)?

korli: Probably, if the link is named the same.

Begasus: How do I check? lua.pc is called the same in all versions?

Begasus: Well, lua$libVersion.pc is version related ... (but writes over the lua.pc file if you check in Terminal)

Begasus: added CONFLICTS

Begasus:
I think this should be OK to merge, comments welcome :)

Begasus:
@korli any ideas on this one? Or wait until the official release?

PR #4328, opened 2019-11-01
sdllopan, fix reference to libSDL, cleanup

Begasus:


games-puzzle/sdllopan/sdllopan-10.recipe
```diff
@@ -26,21 +25,20 @@ ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
 	sdllopan = $portVersion
-	app:lopan = $portVersion
+	app:Lopan = $portVersion
```
Begasus: Is there any way to set this to "SDL Lopan"?

pulkomandy: Why? The executable and deskbar entry are called just Lopan and that's how it should be. The end-user doesn't need to know sdl is involved here.

Begasus: I guess a picture tells more then words (check window title) :)
SDL-Lopan

Begasus:
As a sidenote ... this also doesn't launch from Deskbar ...

games-puzzle/sdllopan/sdllopan-10.recipe
korli: app:SDL\ Lopan doesn't work as expected?

Begasus: Need to check :)

Begasus: Looks like it doesn't, maybe I'll just use SDL_Lopan:
```
Warning: failure creating the build package: Parse error in .PackageInfo(23:12) -> invalid character in resolvable name
```

Begasus: I guess we should poke @scottmc for creating the window tittle? ;)

Begasus:
@scottmc since you changed the window tittle on this one, should I go for the window tittle for the app or just leave it as "Lopan"?

scottmc:
> 
> 
> @scottmc since you changed the window tittle on this one, should I go for the window tittle for the app or just leave it as "Lopan"?

Fix it to be just Lopan as Pulkomandy suggests above.

Begasus:
Ok, will do :)

PR #4333, opened 2019-11-02
afl, remove broken recipe, depends on shm that we dont have

Begasus:
Also checked with the latest 2.52b release, still depends on shm ...


korli:
Well this is a tool one might need. Any chance to use BeOS shared memory instead?

Begasus:
> Well this is a tool one might need. Any chance to use BeOS shared memory instead?

Any place where that could be found? It would probably mean a rewrite of the shm functions then?

korli:
Well PostgreSQL also depends on shm, replacing with BeOS areas is probably a hack but can work.

Begasus:
This can be closed without merging then until someone steps up to fix the shm issue :)

PR #4337, opened 2019-11-03
libshout, bump version

Begasus:


korli:
Remove old recipe?

Begasus:
> Remove old recipe?

Previous recipe still builds for x86_gcc2, if that can be left out I'm willing to remove the old recipe

Begasus:
Removed the previous recipe/patches, nothing is using it atm and only the _x86 probably will see some use once :)

PR #4339, opened 2019-11-05
kwave, fix reference to libpng16

Begasus:


Begasus:
Not checked locally but should be ok, no structural changes @threedeyes could you check?

PR #4340, opened 2019-11-05
texlive, fix reference to libpng16, re-order lines

Begasus:
No structural changes


Begasus:
Somehow it fails to build with these changes, closing this.

return:
@Begasus Do you have a link to the build failure with this package when using libpng16?

TeXLive would be a great package to be made available in HaikuDepot.

Begasus:
@return It builds out of the box with the current recipe in haikuports, it builds fine with libpng16

PR #4359, opened 2019-11-08
nss: bump version

kenmays:


diversys:
It's better to use "git mv" to bump recipes. This way it is easy for the reviewer to see what was changed.

pulkomandy:
"git mv" does not do anything special. In git, wether a file is a copy of move of another is detected when displaying the diff, only, there is no metadata tracking this.
So if you do a "git mv" and then lots of changes to the recipe, it will still not show as a move.

PR #4383, opened 2019-11-13
nspr: bump version

kenmays:
NSPR 4.23, built and tested on hrev53578 x86.
- Bumped from NSPR 4.19 to 4.23.
- Removed old recipes.


dev-libs/nspr/nspr-4.23.recipe
```diff
@@ -45,8 +45,8 @@ BUILD_REQUIRES="
 	"
 BUILD_PREREQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	cmd:aclocal
-	cmd:autoconf
+	cmd:autoreconf
```
return: This should be sorted in alphabetical order, including the letters in between:
```suggestion
	cmd:automake
	cmd:autoreconf
```

return:
Thanks @kenmays! You might need to squash your commits and mention that you have bumped the version and that you have removed the old recipes in your commit message.

kenmays:
Pushed to #4387

PR #3786, opened 2019-04-22
GnuCOBOL, new recipe

Knezzen:

PR #4386, opened 2019-11-14
Gnucobol

Begasus:
Based on the initial work from @Knezzen version 3.0~rc1 https://github.com/haikuports/haikuports/pull/3786 at (this is stable version 2.2 released in 2017), out of the tests 3 fail on both 32bit and 64bit, tests for cobol85 went through without errors, logs have been sent upstream.

### DIFF
--- a/dev-lang/gnucobol/gnucobol-2.2.recipe
+++ b/dev-lang/gnucobol/gnucobol-2.2.recipe
@@ -1,0 +1,123 @@
+SUMMARY="A free COBOL compiler"
+DESCRIPTION="GnuCOBOL (formerly OpenCOBOL) is a free COBOL compiler."
+HOMEPAGE="https://www.gnu.org/software/gnucobol/"
+COPYRIGHT="2001-2017 Free Software Foundation, Inc
+	Keisuke Nishida
+	Roger While
+	Ron Norman
+	Simon Sobisch
+	Edward Hart
+	Brian Tiffin
+	Sergey Kashyrin
+	Dave Pitts"
+LICENSE="GNU GPL v3
+	GNU LGPL v3
+	GNU FDL v1.3"
+REVISION="1"
+SOURCE_URI="https://sourceforge.net/projects/open-cobol/files/gnu-cobol/$portVersion/gnucobol-$portVersion.tar.gz"
+CHECKSUM_SHA256="925838decd65864b2aa3a4bf1385ce4bc708b942e05e8406945a730d7aab32cb"
+
+ARCHITECTURES="!x86_gcc2 ?x86_64"
+SECONDARY_ARCHITECTURES="?x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libVersion="4.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	gnucobol$secondaryArchSuffix = $portVersion
+	cmd:cobc$commandSuffix = $portVersion
+	cmd:cobcrun$commandSuffix = $portVersion
+	lib:libcob$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libdb$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	gnucobol${secondaryArchSuffix}_devel = $portVersion
+	cmd:cob_config$commandSuffix
+	devel:libcob$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	gnucobol$secondaryArchSuffix == $portVersion base
+	devel:libdb$secondaryArchSuffix
+	devel:libgmp$secondaryArchSuffix
+	devel:libintl$secondaryArchSuffix
+	devel:libltdl$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libdb$secondaryArchSuffix
+	devel:libgmp$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	devel:libintl$secondaryArchSuffix
+	devel:libltdl$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:awk
+	cmd:bison
+	cmd:find
+	cmd:flex
+	cmd:gcc$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+TEST_REQUIRES="
+	cmd:wget
+	cmd:perl
+	"
+
+BUILD()
+{
+	autoreconf -vfi
+	runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm $libDir/*.la
+
+	prepareInstalledDevelLib libcob
+
+	# devel package
+	packageEntries devel \
+		$developDir \
+		$commandBinDir/cob-config
+}
+
+TEST()
+{
+	export LIBRARY_PATH=$sourceDir/libcob/.libs:$LIBRARY_PATH
+	make check
+	# ERROR: 761 tests were run,
+	# 9 failed (6 expected failures).
+	# 5 tests were skipped.
+
+	make test
+	# this will download a testsuite for cobol85 (about 4MB)
+	# all tests for cobol85 passed
+}
--- a/dev-lang/gnucobol/licenses/GNU FDL v1.3
+++ b/dev-lang/gnucobol/licenses/GNU FDL v1.3
@@ -1,0 +1,450 @@
+                GNU Free Documentation License
+                 Version 1.3, 3 November 2008
+
+
+ Copyright (C) 2000, 2001, 2002, 2007, 2008 Free Software Foundation, Inc.
+     <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+0. PREAMBLE
+
+The purpose of this License is to make a manual, textbook, or other
+functional and useful document "free" in the sense of freedom: to
+assure everyone the effective freedom to copy and redistribute it,
+with or without modifying it, either commercially or noncommercially.
+Secondarily, this License preserves for the author and publisher a way
+to get credit for their work, while not being considered responsible
+for modifications made by others.
+
+This License is a kind of "copyleft", which means that derivative
+works of the document must themselves be free in the same sense.  It
+complements the GNU General Public License, which is a copyleft
+license designed for free software.
+
+We have designed this License in order to use it for manuals for free
+software, because free software needs free documentation: a free
+program should come with manuals providing the same freedoms that the
+software does.  But this License is not limited to software manuals;
+it can be used for any textual work, regardless of subject matter or
+whether it is published as a printed book.  We recommend this License
+principally for works whose purpose is instruction or reference.
+
+
+1. APPLICABILITY AND DEFINITIONS
+
+This License applies to any manual or other work, in any medium, that
+contains a notice placed by the copyright holder saying it can be
+distributed under the terms of this License.  Such a notice grants a
+world-wide, royalty-free license, unlimited in duration, to use that
+work under the conditions stated herein.  The "Document", below,
+refers to any such manual or work.  Any member of the public is a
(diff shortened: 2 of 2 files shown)


Begasus:
Including the test suite log and log from the tests for cobol85
[summary.log](https://github.com/haikuports/haikuports/files/3853996/summary.log)
[testsuite.log](https://github.com/haikuports/haikuports/files/3853997/testsuite.log)

Knezzen:
Late reply. Great that you guys fixed this! Sorry for not replying earlier. Onwards with OpenCobolIDE for me :)

Begasus:
You made the start @Knezzen :+1:

dev-lang/gnucobol/gnucobol-2.2.recipe
```diff
@@ -0,0 +1,123 @@
...
+COPYRIGHT="2001-2017 Free Software Foundation, Inc
+	Keisuke Nishida
+	Roger While
+	Ron Norman
+	Simon Sobisch
+	Edward Hart
+	Brian Tiffin
+	Sergey Kashyrin
+	Dave Pitts"
+LICENSE="GNU GPL v3
```
GitMensch, line 13: Late note: that's actually wrong, it is GPLv3+ (same for LGPL3+)

dev-lang/gnucobol/gnucobol-2.2.recipe
```diff
@@ -0,0 +1,123 @@
...
+PROVIDES="
+	gnucobol$secondaryArchSuffix = $portVersion
+	cmd:cobc$commandSuffix = $portVersion
+	cmd:cobcrun$commandSuffix = $portVersion
+	lib:libcob$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libdb$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
```
GitMensch, line 42: Is there an optional / variant definition for haikoports? If yes then it would be reasonable to have the libintl, curses and bdb parts marked as those).
Note: for 3.1 you'll likely add cjson (or put it into "libcob" to get it included in libcob) and libxml2, too.

dev-lang/gnucobol/gnucobol-2.2.recipe
```diff
@@ -0,0 +1,123 @@
...
+	devel:libiconv$secondaryArchSuffix
+	devel:libintl$secondaryArchSuffix
+	devel:libltdl$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:aclocal
+	cmd:autoconf
+	cmd:awk
+	cmd:bison
```
GitMensch, line 76: bison, flex, aclocal, autoconf are only required when building from VCS, they are *not* necessary when building from a tarball which was done here (when building from VCS you'd also need sed, grep, help2man and texinfo).
Which parts need `awk` and `find`?

dev-lang/gnucobol/gnucobol-2.2.recipe
```diff
@@ -0,0 +1,123 @@
...
+	"
+
+TEST_REQUIRES="
+	cmd:wget
+	cmd:perl
+	"
+
+BUILD()
+{
+	autoreconf -vfi
```
GitMensch, line 92: If autoreconf is not required in haikoports it shouldn't be done at all.
For a VCS build or "force to use system variant": use autogen.sh, for a tarball: just use the already included versions of the scripts and makefiles.

Begasus:
Thanks for taking a look at this @GitMensch ! Don't have the time to spend on it today, but we'll take all in consideration and inform you on progress! :+1:

dev-lang/gnucobol/gnucobol-2.2.recipe
Begasus: Looking at COPYING (says GNU GPL v3) and COPYING.LESSER (says GNU LGPLv3) is there any change noticeable to see this would be a 3+ (can't seem to find it)

GitMensch: Thanks for asking and taking care.

Those two are the license files shipped with the source, the license that the software is under is written in the specific files. All below libcob [have][1]:

>    The GnuCOBOL runtime library is free software: you can redistribute it
>   and/or modify it under the terms of the GNU Lesser General Public License
>   as published by the Free Software Foundation, either version 3 of the
>   License, or (at your option) any later version.

and the files found [in cobc have][2]:

>    The GnuCOBOL compiler is free software: you can redistribute it
>   and/or modify it under the terms of the GNU General Public License
>   as published by the Free Software Foundation, either version 3 of the
>   License, or (at your option) any later version.

... and a good summary is found in the [README][3] (which may should include a note about the documentation license, which is just referred in the files that use it ):

> All programs except those in lib and libcob are distributed under
the GNU General Public License.  See COPYING for details.
> 
> Programs in lib and libcob are distributed under the GNU Lesser
General Public License.  See COPYING.LESSER for details.
> 
> For any copyright year range specified as YYYY-ZZZZ in this package,
that the range specifies every single year in that closed interval.

I don't know if the recipe format has a specific way to express license, many formats (like [npm / package.json][4]/nuspec use the [SPDX identifiers][5] (in this case), some use textual names (in this case often LGPLv3+) and again others like [PKGBUILD][6] use the name of an actual folder name that is shipped, for LGPv3+this would be LGPL3 (and the only has an extra one with "only" included), some use textual names (in this case often LGPLv3+), the [.spec Format][7].

... ok, found it, [haikuports recipes + License][8] says:

> Use the names from http://spdx.org/licenses/ if possible.

So that would be (for OpenCOBOL 2 and GnuCOBOL 2.2+):
~~~
GPL-3.0-or-later
LGPL-3.0-or-later
GFDL-1.3-or-later
~~

Thanks for the chance to update my license identifier knowledge ;-)

[1]:https://sourceforge.net/p/open-cobol/code/HEAD/tree/trunk/libcob/common.c
[2]:https://sourceforge.net/p/open-cobol/code/HEAD/tree/trunk/cobc/cobc.c
[3]:https://sourceforge.net/p/open-cobol/code/HEAD/tree/trunk/README
[4]:https://docs.npmjs.com/files/package.json#license
[5]:https://spdx.org/licenses/
[6]:https://wiki.archlinux.org/index.php/PKGBUILD#license
[7:]https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Good_Licenses
[8]:https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#license

korli: GPL v3+ is not a license, LICENSE content has to be a license.

dev-lang/gnucobol/gnucobol-2.2.recipe
korli: Well that's a generic information, does it apply to Haiku? Have you tried to build the recipe with your suggestions?

dev-lang/gnucobol/gnucobol-2.2.recipe
korli:
```suggestion
	autoreconf -vfi
```
Some architectures are not known in the generated scripts, hence they require regenerated scripts.

dev-lang/gnucobol/gnucobol-2.2.recipe
korli: I don't see such requirements for ArchLinux: https://www.archlinux.org/packages/community/x86_64/gnucobol/
so we'd rather keep things simple.

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: Please specify which ones are "not known" before but after which are relevant to the haiku port.
Thank you. Note: while this *may* be relevant to version 2.2 it shouldn't be relevant for 3.1 which uses a newer toolset, is it?

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: I have never build any haiku recipe, I came here from a mailing list entry to the gnucobol-bug list only and wanted to give some general advice concerning GnuCOBOL as I'm... a bit experienced to build it with different systems.
As said below: they are not needed at all if you build from a tarball, if you build from VCS then more tools will be missing; if you do want to use autogen you'll definitely need aclocal, and autoconf (if those use awk then it should be their dependency), I see no reason for bison/flex, if there aren't any broken timestamps then those two should never be triggered.

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: I'd say that the archlinux package would be ideally also use the [`optdepends` array][https://wiki.archlinux.org/index.php/PKGBUILD#optdepends] ;-)

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: as noted above the license field, should be, according to the [haikuports recipes + License](https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#license) contain the SPDX identifiers, one per line.

I guess I either did not understand your post (as a non-native speaker) or mis-read the docs?

dev-lang/gnucobol/gnucobol-2.2.recipe
Begasus: Would need to re-check what configure is looking for, if libintl, ncurses are being looked for we could work on this, but in this version is didn't seem to be a requirement (eg can build without it)

dev-lang/gnucobol/gnucobol-2.2.recipe
Begasus: We need to re-check this, this was done on work someone did prior to the merge, if autoreconf isn't needed we'll sort it out :) (PS, downloaded the link you gave me over email, just need some time to look into it) 
Thanks for keeping track on this @GitMensch

dev-lang/gnucobol/gnucobol-2.2.recipe
korli: Here is the list: https://git.haiku-os.org/haiku/tree/data/system/data/licenses

dev-lang/gnucobol/gnucobol-2.2.recipe
korli: well we're experienced at building autotools software too, thanks anyway.

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: I see, so the old entry does not to seem wrong - but how should one understand the haikoports documentation linked above in this case? Full quote:

> The licenses should either be picked from the list in /system/data/licenses, or provided next to the recipe in a licenses folder as text file for inclusion inside the package.
Use the names from http://spdx.org/licenses/ if possible.

I understand it as "if possible do not use those in system/data/licenses but the spdx-identifier"; the big difference is that system/data/licenses has the license text while the spdx-identifier tell the actual license (which in this case is "or newer", something that can by definition be not part of the "not yet newer" license).

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: ncursesw/ncurses + libintl are looked for, if not available they are not included; bdb is always looked for and if not available configure stops. All of those can be forced to be enabled/disabled with --enable-stuff/--disable-stuff (see `configure --help`).

dev-lang/gnucobol/gnucobol-2.2.recipe
pulkomandy: The backstory here is that the system/data/licenses files are very old, and did not use spdx naming convention, and now we can't really change them because so many package rely on the current names.

So I think what is meant is "if you have to add your license in the package because it's not part of the standard set, use SPDX identifier as the filename". But indeed the wording isn't clear at all.

As for the "or later" clause, it is not technically part of the license itself. Maybe we can handle it like we do for GPL linking exceptions, as a separate file containing only the differences from the inital license. So in this case it would be something like "This software can, at your option, be used and redistributed under the terms of any later version of the GPL"?

dev-lang/gnucobol/gnucobol-2.2.recipe
Begasus: Aside from cjson all things should be ok to build this ...

Begasus: Builds OK, checking the tests atm and will send you the results over email @GitMensch

Begasus:
> The FDL license is part of the project's sources (both VCS and tarball) as COPYING.DOC, it likely is a good idea to reference this instead of including a copy (which may differ, now or in the future).

As you can see in the link there isn't a default FDL in our licenses, so we have to include this as a seperate file (https://git.haiku-os.org/haiku/tree/data/system/data/licenses)

korli:
There is already a gnu fdl specified in this recipe, is it not correct?

Begasus:
> There is already a gnu fdl specified in this recipe, is it not correct?

That's correct, and added as a separate file because there is no default for it in our licenses

korli:
> That's correct, and added as a separate file because there is no default for it in our licenses

OK. Sorry i understood you wanted to include this as a separate file... again.

dev-lang/gnucobol/gnucobol-2.2.recipe
GitMensch: @pulkomandy

> The backstory here is that the system/data/licenses files are very old, and did not use spdx naming convention, and now we can't really change them because so many package rely on the current names.

Why not change the "old" ones to the SPDX identifier as the filename and then add a link from the "old" names pointing to the new ones? This should solve every conflict, shouldn't it?

GitMensch: > The backstory here is that the system/data/licenses files are very old

What about updating the folder in general to include __common__ newer licenses, too?

PR #4410, opened 2019-11-30
mxml, bump version, rename to mxml

Begasus:


Begasus:
According to Gentoo it should be mxml instead of mini_xml: http://gpo.zugaina.org/dev-libs/mxml
On repology there are 3 for mini_xml (2 from haikuports) and 99 for mxml: https://repology.org/project/mxml/versions

Begasus:
Patch has been sent upstream

Begasus:
Actual test is run during build, so there is no need to call (obsolete) TEST():
```
Linking testmxml...
Testing library...
Stdio file test passed!
String test passed!
File descriptor test passed!
```

PR #4432, opened 2019-12-08
apache: bump version

alwayslivid:


alwayslivid:
Relevant screenshot of HaikuDepot.

image

pulkomandy:
While we are at it, I see this recipe has no TEST() section. Does the apache buildsystem have some tests to run? Can we add them there and run them? That would be better than just checking that it compiles and can be installed.

alwayslivid:
> While we are at it, I see this recipe has no TEST() section. Does the apache buildsystem have some tests to run? Can we add them there and run them? That would be better than just checking that it compiles and can be installed.

I tried to look into it and I managed to find [this page](https://httpd.apache.org/test/), but it doesn't seem to contain any substantial information. 

I also just used the officially available, upstream [Arch Linux package as a point of reference](https://git.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/apache) in order to check whether that project has implemented any testing techniques, but there doesn't seem to be anything out there regarding the type of testing you're talking about either.

PR #4437, opened 2019-12-09
added scons-3.1.1 receipt

LordOfDragons:


extrowerk:
```
Warning: POLICY ERROR: no matching provides "cmd:scons_3.1.1.bat" for "bin/scons-3.1.1.bat"
Warning: POLICY ERROR: no matching provides "cmd:scons.bat" for "bin/scons.bat"
Warning: Error: packaging policy violation(s) in strict mode
```

extrowerk:
REVISION should be 1 as it is a new version.

extrowerk:
Here is my version, it should work ok: http://0x0.st/zUme.recipe
I cleaned up the -INSTALL section too, no idea, why was it so complicated.

I don't like the idea to make it depend on python3.6, the current python3 version is 3.7.5, and i think we should use that instead.

LordOfDragons:
> Here is my version, it should work ok: http://0x0.st/zUme.recipe
> I cleaned up the -INSTALL section too, no idea, why was it so complicated.
> 
> I don't like the idea to make it depend on python3.6, the current python3 version is 3.7.5, and i think we should use that instead.

Can the recipe point to "python3" or is it required to point to a specific version like "python3.7.5"? SCons 3.1.1 is compatible with python 3.5+. Can I specify something like a minimum version?

mmuman:
Some grep shows:
`sci-mathematics/ginac/ginac-1.6.7.recipe:       cmd:python >= 2.7`

extrowerk:
Check the Python recipes, you will see, Python 3.7.5 provides "cmd:
python3" and "cmd:python3.7" while Python 3.6 provides "cmd:python3.6" 
only.

It means Python 3.7 is the default one from the 3.x series.
According this i would state cmd:python3 in the SCons recipe as 
REQUIRES.

But that's my opinion, maybe somebody thinks differently.

Plüss Roland <notifications@github.com>:
> > Here is my version, it should work ok: http://0x0.st/zUme.recipe
> > I cleaned up the -INSTALL section too, no idea, why was it so 
> > complicated.
> > 
> > I don't like the idea to make it depend on python3.6, the current 
> > python3 version is 3.7.5, and i think we should use that instead.
> 
> Can the recipe point to "python3" or is it required to point to a 
> specific version like "python3.7.5"? SCons 3.1.1 is compatible with 
python 3.5+. Can I specify something like a minimum version?

LordOfDragons:
Commited the changes. Is it better now?

dev-util/scons/scons-3.1.1.recipe
```diff
@@ -0,0 +1,53 @@
+SUMMARY="A software construction tool"
+DESCRIPTION="SCons is an open-source software construction tool - that is, a \
+next-generation build tool. Think of SCons as an improved, cross-platform \
+substitute for the classic Make utility with integrated functionality similar \
+to autoconf/automake and compiler caches such as ccache. In short, SCons is \
+an easier, more reliable and faster way to build software."
+HOMEPAGE="http://www.scons.org/"
```
korli: https://www.scons.org/

PR #4438, opened 2019-12-09
Ghostscript: bump, JBIG2DIC: fix

extrowerk:
Ghostscript is currently broken, this fixes it.


Begasus:
Maybe create a separate PR for jbig2dec? If the buildbots would pick up ghostscript before it it will probably fail again?

media-libs/jbig2dec/jbig2dec-0.17.recipe
```diff
@@ -59,6 +59,8 @@ INSTALL()
 {
 	make install
 
+	fixPkgconfig
```
Begasus: Shouldn't this come after prepareInstallDevelLib? (PS it's only one lib so you can remove the extra "s" there :)

Begasus:
Builds ok for secondary architecture :)

PR #4448, opened 2019-12-14
dbus: fixed 32-bit support

alwayslivid:
dbus, which is necessary for using certain pieces of ported software, seems to work just fine from my point of view in a 32-bit Haiku VM.


Begasus:
Can confirm that it worked (build) here on my side also

PR #4462, opened 2019-12-18
fossil: update to version 2.10, circa October 2019.

agmsmith:
It compiles and runs OK on 32 bit x86 (not on GCC2) and 64 bit Haiku.  Didn't have to change the recipe, other than the checksum and new version number.  Though when running you see a lot of log messages about failing to find libroot-addon-icu.so, which is actually there in system/libs.  And some probably harmless warnings when compiling on 32 bit about printing the long int GetPID() with a %d.

### DIFF
--- a/dev-vcs/fossil/fossil-2.10.recipe
+++ b/dev-vcs/fossil/fossil-2.10.recipe
@@ -1,17 +1,17 @@
 SUMMARY="A tool for distributed software configuration management"
 DESCRIPTION="Fossil is a simple, high-reliability, distributed version \
 control system like Git and Mercurial, but Fossil also supports distributed \
-bug tracking, distributed wiki, and a distributed blog mechanism all in a \
-single integrated package.
+bug tracking, distributed wiki, distributed forums, and a distributed blog \
+mechanism all in a single integrated package.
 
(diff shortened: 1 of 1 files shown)


agmsmith:
Actually the previous 2.6 version had the same libroot-addon-icu.so bug.

Begasus:
fuel-1.0.1.recipe is (afaict) the only recipe using fossil atm, could you check if it still builds with this version?
If that is OK you could drop the older version for fussil (keeps git noise down) :)

Begasus:
Checked out a build for fuel, that's ok, so you can remove at least version 2.6 with this PR

agmsmith:
Thanks for checking, I didn't think to look for other programs using fossil (I use it stand-alone).  Will update my pull request momentarily.

agmsmith:
I've updated the commit - removed fossil 2.6 recipe, also updated the copyright date in 2.10 recipe.

Begasus:
Thanks for the changes! :+1:

agmsmith:
That was fast, thanks Begasus.  Now I have to look into moving a database from fossil 1.xx to 2.xx...

PR #4465, opened 2019-12-20
Blender: fix regression by using python 3.6, small fixes

### LINKED ISSUE #4459: [blender] empty menus and toolbars
Component: media-gfx/blender
Platform: Haiku hrev53634 x86_64
Build on: Haiku hrev53634 x86_64
Compiled by: haikuporter --get-dependencies --no-source-packages blender

blender
[blender-log.txt](https://github.com/haikuports/haikuports/files/3977054/blender-log.txt)

X547:
Fixes #4459.


X547:
> Libdir also always available, why does libz needs hardcoding?

Without that, blender links with wrong version of libz causing `libpng error: bad parameters to zlib` error and bitmaps in Blender interface are not displayed.

> And git, that should be definetely unneeded.

cmake says that git is needed for version information.

extrowerk:
> cmake says that git is needed for version information.
We dont have to provide everything, what cmake wants.
But if you want to do so, you still have to add cmd:git as it is not declared at all in the recipe, so your hardcoded path points a non-existing file.
We still dont want recipes depending on git/svn/etc.

extrowerk:
> Without that, blender links with wrong version of libz causing libpng error: bad parameters to zlib error and bitmaps in Blender interface are not displayed.

I see. Thats explains some things.

X547:
> We dont have to provide everything, what cmake wants.
> ...
> We still dont want recipes depending on git/svn/etc.

As you wish.

PR #4477, opened 2019-12-23
tmux: Fix crash loading config

ambroff:
The tmux code depends on some BSD functionality, namely fparseln,
when parsing configs. This is called in load_cfg during startup after
finding the appropriate config file to load.

On platforms where fparseln isn't available, tmux uses a polyfill
library called compat that it ships with, which includes a fork of the
FreeBSD fparseln function.

The autoconf script for tmux searches the libutil header for fparseln
in order to determine whether to use this compat library or the
fparseln from the the system.

Before this patch, on my x86_64 system at least, tmux would crash on
startup trying to load $HOME/.tmux.conf. It turns out that this is
because on Haiku, the FreeBSD compatibility in libutil code is
disabled in /system/develop/headers/bsd/features.h unless either
_BSD_SOURCE is defined.

    #if defined(_BSD_SOURCE) \
            || (!defined(__STRICT_ANSI__) && !defined(_POSIX_C_SOURCE))
            #define _DEFAULT_SOURCE
    #endif

Since _BSD_SOURCE is not defined, and _POSIX_C_SOURCE *is* defined
(because the compiler is given -std=gnu99), then _DEFAULT_SOURCE is
never defined.

The end result is that fparseln is visible to autoconf, but is not
visible to the compiler when compiling cfg.c.

This leads fparseln to be implicitly defined, which changes the return
type from char* to int, and that conversion can cause a SEGFAULT.

The effect of this patch is to just define _BSD_SOURCE so that the
FreeBSD compat code is visible to the compiler, which produces a tmux
binary that works as expected.


extrowerk:
Great explanation!

korli:
FYI Haiku hasn't yet switched to _DEFAULT_SOURCE on stable releases.

PR #4479, opened 2019-12-23
icecast, fix running as root, move PATCH and libnetwork check to configure.in,…

Begasus:
… remove secondary architecture


Begasus:
Bugger, forgot to revbump ...

stoltenhoff:
Thank you for the fast reply/work!

PR #4501, opened 2019-12-28
Add noteshrink recipe

Ra-V2:


return:
Also since this is a python recipe, its probably worth waiting for the other maintainers to have a second look at this whether if it installs properly or not.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+to compact PDFs.
+"
+HOMEPAGE="https://github.com/mzucker/noteshrink"
+COPYRIGHT="2016 Matt Zucker"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/mzucker/noteshrink/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="2a29c09768973e688b693b61337c6c49384e8123cf88824dba335cc8c4ed2ca8"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```
korli: Should be "any"

Ra-V2: Are you sure? Some of the required are untested or can't be build on other architectures (e.g. scipy).

aarroyoc: But the recipe itself doesn't contain any compiled code, so should be any. If a dependency cannot be built in the architecture it's already reflected and the package won't be installed because the dependency doesn't work. But the package itself could still work if someone makes a version of the dependency that works

Ra-V2:
Good to merge now?

Begasus:
Does this actually work? Checking on gcc2h fails (running make in the sources attemps to run the cmd on the exmaples, but fail due to missing deppendencies), haven't checked on x86_64 (yet)

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="Convert scans of handwritten notes to beautiful, compact PDFs"
+DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes \
+to compact PDFs.
+"
```
Begasus: Move double quote to the end of previous line, no new line here

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,46 @@
...
+	"
+
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:python3
+	"
+
+BUILD()
+{
+	mkdir -p $binDir
```
Begasus: mkdir not needed in BUILD():
```diff
-	mkdir -p $binDir
+	true
```

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,46 @@
...
+}
+
+INSTALL()
+{
+	mv noteshrink.py noteshrink
+	cp noteshrink $binDir
```
Begasus: Move mkdir here and maybe:
```diff
-	mv noteshrink.py noteshrink
-	cp noteshrink $binDir
+	mkdir -p $binDir
+	cp noteshrink.py $binDir/noteshrink
```

Begasus:
Running this in Terminal on x86_64 seems to be ok (with the given command in the makefile)

media-gfx/noteshrink/noteshrink-0.1.1.recipe
Begasus: I would at least add a comment somewhere that it doesn't work on gcc2hybrid due to missing dependencies, maybe create an issue for those not available (if not done already)
Or even try to fix the missing dependencies :)

Ra-V2: I added comment to the recipe. It seems on gcc2hybrid we can't install numpy_python, pillow_python and scipy_python which are needed for the program to work properly.

Begasus: For pillow and numpy I have already a local recipe, but not for scipy (yet), still fighting with python over this :), but yes, for now a comment for that would be good

smccrearyx:
I'll defer to @Begasus on this one.  Are all issues addressed now?

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:python3
+	"
+
+BUILD()
```
korli: the BUILD() can be removed.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,45 @@
...
+	setuptools_python
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+BUILD_PREREQUIRES="
+	cmd:make
+	cmd:python3
```
korli: the BUILD_REQUIRES and BUILD_PREREQUIRES sections can be removed.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -0,0 +1,31 @@
...
+PROVIDES="
+	noteshrink = $portVersion
+	cmd:noteshrink
+	"
+REQUIRES="
+	haiku
+	imagemagick
+	numpy_python >= 1.10
+	pillow_python
+	python_imaging
```
korli: python_imaging is not required. please remove.

PR #4504, opened 2019-12-29
Fixed cylindrix

alwayslivid:
Running this recipe in its current form returns an error that reads:
```python
'NoneType' object is not iterable
```
A full log of that error can be found [here](https://paste.ubuntu.com/p/dsxhjkpcSH/).

My proposed change is a small fix in the **COPYRIGHT** section of the file. cylindrix now compiles successfully on the x86_64 architecture.

image

The recipe itself has not been tested on other architectures, hence the reason why I marked them as "untested". My assumption is that the recipe was solely broken due to that error.


pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

pulkomandy:
Can we fix haikuporter to not crash in this case?

games-action/cylindrix/cylindrix-1.0
```diff
@@ -1,15 +1,16 @@
...
+	2001 Anthony Thibault"
 LICENSE="GNU GPL v2"
 REVISION="1"
 SOURCE_URI="https://d1cxvcw9gjxu2x.cloudfront.net/projects/1235#cylindrix-1.0.tar.gz"
 CHECKSUM_SHA256="33bfc264d906a1977090e47a38f544551d2939d4c0ea01e1792f63ab4fa06f20"
 SOURCE_DIR="cylindrix"
 
-ARCHITECTURES="!x86_gcc2 !x86 !x86_64"
-SECONDARY_ARCHITECTURES="!x86"
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
```
Begasus: Don't set the architecture as working before it actually runs the game
```diff
-ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="?x86_gcc2 ?x86 ?x86_64"
```

alwayslivid: image

Will change, it does launch the game, but it does get stuck in the splash screen and it also seems to be buggy.

Begasus: Already been at that point in the past, hence it was disabled :)

alwayslivid: > Already been at that point in the past, hence it was disabled :)

So fixing the COPYRIGHT section seems to be enough-- I'll invert the changes.

games-action/cylindrix/cylindrix-1.0
```diff
@@ -1,7 +1,8 @@
 SUMMARY="A combat game"
 DESCRIPTION="Cylindrix is a 3 degrees of freedom combat game"
 HOMEPAGE="https://www.allegro.cc/depot/Cylindrix"
-COPYRIGHT="2001 John R. McCawley III, Anthony Thibault"
+COPYRIGHT="2001 John R. McCawley III
+	2001 Anthony Thibault"
```
Begasus: Should check on x86_64, but build is ok here on gcc2h secondary architecture without this change

Begasus:
No real fix here

PR #4509, opened 2019-12-31
pybind11, new recipe

Begasus:


dev-libs/pybind11/pybind11-2.4.3.recipe
```diff
@@ -0,0 +1,70 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+
+BUILD_PREREQUIRES="$BUILD_PREREQUIRES
```
korli: please remove $BUILD_PREREQUIRES

korli:
otherwise looks good.

dev-libs/pybind11/pybind11-2.4.3.recipe
Begasus: Still working on it, should also be "any" as there is no compiler involved? (at least it's working that   way, wip) :)

Begasus:
This is a wip recipe to get support for scipy multi python version

dev-libs/pybind11/pybind11-2.4.3.recipe
Begasus: Updated ...

korli:
> This is a wip recipe to get support for scipy multi python version

are you sure?

Begasus:
I was looking into the missing dependencies for noteshrink and saw that scipy_python was missing the headers for this one at one point

Begasus:
atm I'm looking into scipy_python 2.4.1 and you can find pylibs11 in their requirements now: https://github.com/scipy/scipy/blob/345eb722978868788236a0f6ee12ccd0bd5c1553/INSTALL.rst.txt#prerequisites

korli:
right, I would stick to scipy_python3-1.3.1.recipe, as that's what uses ArchLinux ATM. I think they dropped python2 support also, so you need two recipes.

Begasus:
> right, I would stick to scipy_python3-1.3.1.recipe, as that's what uses ArchLinux ATM. I think they dropped python2 support also, so you need two recipes.

OK thanks, I'll look into 1.3.1 then, wanted to see if I could get both python versions (3.6 and 3.7) in one recipe, but getting some errors, worth investigating :) (or maybe create a task for it?)

Begasus:
Looks OK from my side

PR #4522, opened 2020-01-02
box2d, fix installation paths

Begasus:


Begasus:
This was a fix to see if caveexpress would find Box2D files, from looking into the files they seem to be OK here, but caveexpress still doesn't seem to pick it up ...

Begasus:
@korli could you check, looking at the cmake config files they seem to point to the correct paths for the headers and library

dev-games/box2d/box2d-2.3.1.recipe
```diff
@@ -35,6 +36,7 @@ BUILD_REQUIRES="
 	devel:libGLU$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
+	cmd:ccache
```
korli: Really?

korli:
Box2D is used by a few recipes. Do they still build?

dev-games/box2d/box2d-2.3.1.recipe
Begasus: Needs to go, but in the process of "rebuilding" it's nice to have around :)

Begasus:
> Box2D is used by a few recipes. Do they still build?

Need to check a few, but so far being used doesn't mean being "detected" ...

dev-games/box2d/box2d-2.3.1.recipe
Begasus: removed :)

Begasus:
Tried out with Terrarium (0.1.9), with the previous version it doesn't show the version for box2d, with this one it shows version as >= 2.3.0
box2d

PR #4523, opened 2020-01-02
Bump: various libretro cores

kwyxz:
Various libretro cores bumped to their newer versions


Begasus:
Could you check https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=&viewMode=expanded
Several recipes need a revbump there

PR #4524, opened 2020-01-03
terrarium, bump, disable (not really usefull)

Begasus:


Begasus:
Game is playable'sh :) For those wanted to give it a go they could build it with haikuporter, not suited to end up in the depot ...

PR #4549, opened 2020-01-12
Cleaned up libSDL* dependencies

alwayslivid:
Relevant issue: #3878

I updated the recipes of the following recipes: `t4k_common`, `xmoto`, `xut`, `openxcom`, `flatzebra`, `crimson` and `tuxmath`.

I also made a small correction in the `tuxmath` recipe that prevented it from being compiled. I checked all recipes on the x86_64 architecture, however, some recipes (`xmoto`, `xut`) have not been checked properly due to the architecture and `openxcom` seems to be a bit broken either way, but has been presumptively updated, as well as the two other aforementioned recipes.

If there's a problem with that, I made my changes in separate commits so that they can be easily reverted.


Begasus:
Please make separate PR's for all these, for xmoto for instance I still have an outstanding PR myself, check haikuports for these

PR #4551, opened 2020-01-12
Suitesparse 5.6.0

extrowerk:
Bump, consolidation


korli:
note: ArchLinux provides metis as a separate package. 
It might be good to add a REPLACES section for old subpackages.

PR #4555, opened 2020-01-12
tuxmath: cleaned libSDL* dependencies

alwayslivid:
Relevant pull request: #4549


Begasus:
TuxMath and TuxTyper both use t4k_common, if that needs changes it goes first, also maybe I missed it but what is the "additional change"?

Begasus:
While looking at it, maybe see if there are newer versions released atm (not needed for this atm)

alwayslivid:
> TuxMath and TuxTyper both use t4k_common, if that needs changes it goes first, also maybe I missed it but what is the "additional change"?

Oops, that was another recipe, my bad.

korli:
Bump revision after changes

PR #4560, opened 2020-01-12
zopfli: bump to 1.0.3

Ra-V2:


korli:
Could you switch to cmake? I see mentions of Haiku in CMakeLists.txt, it could compile out of the box, and provide zopflipng

app-arch/zopfli/zopfli-1.0.3.recipe
```diff
@@ -31,17 +32,23 @@ BUILD_REQUIRES="
...
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_INSTALL_PREFIX="$commandBinDir"
 	make $jobArgs
 }
 
 INSTALL()
 {
 	install -d "$commandBinDir"
 	install -t "$commandBinDir" zopfli
+	install -t "$commandBinDir" zopflipng
```
Begasus: Didn't check, but isn't there a "make install"?

Ra-V2: @Begasus When I try to do "make install" it says there is no rule to make target "install".

Begasus: You probably want to change directory for that first :)
```diff
+	cd build
+	make install
```

Ra-V2: Ok, nevermind. Now should be good I think.

korli:
$prefix/include should be moved to $includeDir

PR #4573, opened 2020-01-12
Lapack: bump

extrowerk:
SONAME unchanged


korli:
It seems this patch is needed:
https://github.com/Reference-LAPACK/lapack/commit/87536aa3c8bb0af00f66088fb6ac05d87509e011

extrowerk:
Will add the patch.

PR #4579, opened 2020-01-13
zopfli: move $prefix/include to $includeDir

Ra-V2:
https://github.com/haikuports/haikuports/pull/4560

### DIFF
--- a/app-arch/zopfli/zopfli-1.0.3.recipe
+++ b/app-arch/zopfli/zopfli-1.0.3.recipe
@@ -4,7 +4,7 @@ compatible compressor programmed in C"
 HOMEPAGE="https://github.com/google/zopfli"
 COPYRIGHT="2011-2018 Google Inc."
 LICENSE="Apache v2"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="$HOMEPAGE/archive/zopfli-$portVersion.tar.gz"
 CHECKSUM_SHA256="e955a7739f71af37ef3349c4fa141c648e8775bceb2195be07e86f8e638814bd"
 SOURCE_DIR="zopfli-zopfli-$portVersion"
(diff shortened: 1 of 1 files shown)


Begasus:
You can use cmake options to install the libraries and headers without the need to move them, also, you can create shared libs if you turn that option on
You'll need to provide the libraries and add a devel package then (also use the libVersion variant)

Ra-V2:
@Begasus I did something like this. But I have no idea where are these options to move include in cmake.

korli:
please bump the revision
where is the actual move of $prefix/include to $includeDir ?

Ra-V2:
@korli I bump this revision and added again this move to recipe. No policy errors are currently popping up.

app-arch/zopfli/zopfli-1.0.3.recipe
```diff
@@ -19,35 +19,53 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
...
 	"
 
 BUILD()
 {
 	mkdir -p build && cd build
 	cmake .. \
 		-DCMAKE_BUILD_TYPE=Release \
-		-DCMAKE_INSTALL_PREFIX=$prefix
+		-DCMAKE_INSTALL_PREFIX=$prefix \
+		-DZOPFLI_BUILD_SHARED=ON
```
Begasus: Check CMakeCache.txt for the variables you can use for LIBDIR and INCLUDEDIR

app-arch/zopfli/zopfli-1.0.3.recipe
```diff
@@ -19,35 +19,53 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
...
 	make $jobArgs
 }
 
 INSTALL()
 {
 	cd build
 	make install
+	mkdir -p $(dirname $includeDir)
+	mv $prefix/include $includeDir
```
Begasus: No need for them:
```diff
-	mkdir -p $(dirname $includeDir)
-	mv $prefix/include $includeDir
```

Ra-V2:
@Begasus Ok, I got this. I was checking these settings only with -L and I don't see all of them :)

Begasus:
@Ra-V2 the option -L doesn't always reveal the needed options, but it's where you start, thanks for your contribution!

app-arch/zopfli/zopfli-1.0.3.recipe
```diff
@@ -19,30 +19,48 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
...
+	"
+
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	"
 BUILD_PREREQUIRES="
 	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
 	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
```
korli: This is probably not needed.

PR #4525, opened 2020-01-04
flite: bump to 2.1

alwayslivid:
Additionally, I also recreated the patchset that was originally made by @TimothyGu.

PR #4580, opened 2020-01-13
flite: bump to 2.1

alwayslivid:
This is a continuation of #4525, as the branch that contained my modifications was accidentally deleted.


app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,142 @@
...
+	2001 Cepstral, LLC
+	2001 Sun Microsystems, Inc.
+	2003-2004 Nagoya Institute of Technology
+	2003-2004 Nara Institute of Science and Technology
+	2008-2012 University of Illinois at Urbana-Champaign"
+# Although individually many files have different licenses, as a whole this
+# license covers all the files.
+LICENSE="Flite"
+REVISION="1"
+SOURCE_URI="http://www.festvox.org/${portName}/packed/${portName}-${portVersion}/${portName}-${portVersion}-release.tar.bz2"
```
Begasus: Using $portName here wont work for secondaryArchSuffix, it will change the name to "flite_x86"

app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,142 @@
...
+		libflite_cmu_indic_lex \
+		libflite_cmu_time_awb \
+		libflite_cmu_us_awb \
+		libflite_cmu_us_kal \
+		libflite_cmu_us_kal16 \
+		libflite_cmu_us_rms \
+		libflite_cmu_us_slt \
+		libflite_cmulex \
+		libflite_usenglish
+	packageEntries devel $developDir
```
Begasus: I'm in favour of removing the binaries from secondary architecture here, they will probably collide as they are both installed in $prefix/bin for gcc2h ...

app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,142 @@
...
+	$libDir/libflite_cmu_us_awb.so.$libVersion \
+	$libDir/libflite_cmu_us_kal16.so.$libVersion \
+	$libDir/libflite_cmu_us_slt.so.$libVersion \
+	$libDir/libflite_cmulex.so.$libVersion \
+	$libDir/libflite_usenglish.so.$libVersion
+
+BUILD()
+{
+	autoconf
+	runConfigure ./configure --enable-shared
```
Begasus: Whoops, you forgot to use --omit-dirs here

app-accessibility/flite/flite-2.1.recipe
alwayslivid: Should I use a case that detects whether the x86_gcc2 architecture is being used and exclude everything from line 127 and below if that's the case?

Begasus: It's a bit more complex now, with the debuginfopackage

Begasus:
As this is only used by Qt5, maybe disable x86_gcc2 with "?", that way there shouldn't be a problem for the colliding binaries

alwayslivid:
> As this is only used by Qt5, maybe disable x86_gcc2 with "?", that way there shouldn't be a problem for the colliding binaries

Oh yeah, I remember this conversation regarding Qt5 and x86_gcc2 in the IRC. Will handle it right now.

Begasus:
Didn't test in it's current state, but LGTM

app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,143 @@
...
+	$libDir/libflite_cmu_us_slt.so.$libVersion \
+	$libDir/libflite_cmulex.so.$libVersion \
+	$libDir/libflite_usenglish.so.$libVersion
+
+BUILD()
+{
+	autoconf
+	runConfigure --omit-dirs "binDir" ./configure \
+		--bindir="$commandBinDir"
+	make $jobArgs
```
Begasus: $jobArgs was disable in previous version (make # Parallel builds not working), does this work ok now?

app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,143 @@
...
+}
+
+INSTALL()
+{
+	make $jobArgs install
```
Begasus: Suggestion:
```diff
- 	make $jobArgs install
+ 	make install
```

app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,143 @@
...
+SECONDARY_ARCHITECTURES="?x86_gcc2 x86"
+
+commandBinDir=$binDir
+commandSuffix=$secondaryArchSuffix
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libVersion="2.1"
```
Begasus: Maybe? (saves changes for future bump):
```diff
- libVersion="2.1"
+ libVersion="$portVersion"
```

alwayslivid: I'm going to change this, although I can't be too sure about it due to the fact that this may actually mess up version bumps. (example: minor change bumps are much less likely to have their entire library versions bumped as well)

Begasus: Looking at the previous recipe it used libVersion = portVersion also (2.0.0), so I guess we can assume they will stick to that

Begasus:
atm doesn't seem to link to -lnetwork ...

app-accessibility/flite/flite-2.1.recipe
Begasus: Seems to be OK, so can stay as is

app-accessibility/flite/flite-2.1.recipe
```diff
@@ -0,0 +1,143 @@
...
+	$libDir/libflite_cmu_us_kal16.so.$libVersion \
+	$libDir/libflite_cmu_us_slt.so.$libVersion \
+	$libDir/libflite_cmulex.so.$libVersion \
+	$libDir/libflite_usenglish.so.$libVersion
+
+BUILD()
+{
+	autoconf
+	runConfigure --omit-dirs "binDir" ./configure \
+		--bindir="$commandBinDir"
```
Begasus: Somewhere along the line you removed the --enable-shared option, without it doesn't link to -lnetwork
```diff
- 		--bindir="$commandBinDir"
+ 		--bindir="$commandBinDir" \
+		--enable-shared
```

PR #4592, opened 2020-01-16
abuse: change save localization to settings

Ra-V2:
There is a problem with header FindDirectory.h. When I try to compile it with this header, I get a lot of compilation errors:
[errors.txt](https://github.com/haikuports/haikuports/files/4071135/errors.txt)
For now I only changed this directory to settings without find_directory because this doesn't work.

### DIFF
--- a/games-action/abuse/patches/abuse-0.8.patchset
+++ b/games-action/abuse/patches/abuse-0.8.patchset
@@ -1,4 +1,4 @@
-From 70602ecce40c1502303d83500a01fff697cfcd60 Mon Sep 17 00:00:00 2001
+From 702f1407e6990dd1ed714815d8fa3a0cfcd181e5 Mon Sep 17 00:00:00 2001
 From: Yourself <user@shredder.home>
 Date: Tue, 25 Sep 2018 16:05:19 +0000
 Subject: removed reference to unused sys/shm.h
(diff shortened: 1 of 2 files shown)


Begasus:
> There is a problem with header FindDirectory.h. When I try to compile it with this header, I get a lot of compilation errors:
> [errors.txt](https://github.com/haikuports/haikuports/files/4071135/errors.txt)
> For now I only changed this directory to settings without find_directory because this doesn't work.

Looks like some changes to our gcc2 compiler are creating these parse errors, this isn't due to your patch, could you try if you can build it for secondary architecture?

korli:
No, FindDirectory.h tries to import the system image.h but the local image.h gets used instead.
You have to fix the header search path obviously.

korli:
for gcc 2.95 specifically in src/sdlport/Makefile.am the INCLUDES line should something like:
INCLUDES =  $(SDL_CFLAGS) -I- -I$(srcdir)/../ -I$(srcdir)/../imlib

Ra-V2:
@korli After change still getting these errors I when include FindDirectory header.

korli:
> @korli After change still getting these errors when I include FindDirectory header.

Well, you need to add "autoreconf -fi" before runConfigure in the recipe. You'll need additional dependencies (autoconf, etc)
It seems I got this backward. Something like this should help:
INCLUDES = -I$(srcdir)/../ -I$(srcdir)/../imlib -I- $(SDL_CFLAGS) 

This will not work because of a missing include path for lisp. Hope you'll find out yourself.

Ra-V2:
Thanks @korli. Now everything works well. Files are saved in settings directory.

games-action/abuse/patches/abuse-0.8.patchset
```diff
@@ -39,5 +39,75 @@ index 8df9cbf..85fc0ec 100644
...
++    char buffer[B_PATH_NAME_LENGTH+B_FILE_NAME_LENGTH];
++    if(find_directory(B_USER_SETTINGS_DIRECTORY, volume, true, buffer, sizeof(buffer)) == B_OK)
++    {
++    	strlcat(buffer, "/abuse/", sizeof(buffer));
++    	mkdir(buffer, 0700);
++    	set_save_filename_prefix(buffer);
++    }
++    else
++    {
++        printf( "WARNING: Unable to get $HOME environment variable.\n" );
```
korli: This warning isn't correct.

PR #4595, opened 2020-01-16
irrlamb: added recipe

AlwaysLivid:
This recipe is a WIP, due to the fact that I'm not really sure about how I should deal with this error:

image

### DIFF
--- a/games-arcade/starfighter/patches/starfighter-2.0.0.3.patch
+++ b/games-arcade/starfighter/patches/starfighter-2.0.0.3.patch
@@ -1,0 +1,34 @@
+--- a/src/engine.c
++++ b/src/engine.c
+@@ -22,9 +22,13 @@
+ #include <sys/stat.h>
+ #include <unistd.h>
+ 
++#ifdef __HAIKU__
++#include <FindDirectory.h>
++#else
+ #ifndef _WIN32
+ #include <pwd.h>
+ #endif
++#endif
+ 
+ #include "SDL.h"
+ 
+@@ -212,6 +216,17 @@
+ 		engine_showError(2, dir);
+ 
+ 	snprintf(engine.configDirectory, PATH_MAX, "%s/", dir);
++#elif __HAIKU__
++	subdir = "starfighter";
++
++	char path[PATH_MAX];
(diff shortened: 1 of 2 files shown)


games-simulation/irrlamb/irrlamb-1.0.1r613.recipe
```diff
@@ -0,0 +1,64 @@
...
+	irrlamb$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libgl$secondaryArchSuffix
+	lib:libglu$secondaryArchSuffix
+	lib:libopenal$secondaryArchSuffix
+	lib:libvorbis$secondaryArchSuffix
+	lib:libx11$secondaryArchSuffix
```
extrowerk: The X11 port cannot draw. You can either port the game to use Haiku gui or abandon this.

extrowerk:
Please do not use X11 libs in your ports, they are not useful as is, they are usable only if one want to build programs which draws on a remote X11 server running on other OS.

alwayslivid:
> Please do not use X11 libs in your ports, they are not useful as is, they are usable only if one want to build programs which draws on a remote X11 server running on other OS.

Will keep that in mind, thanks for the remark. I'm not planning to port the game at this time.

extrowerk:
Ok, i made a mistake: the project needs X11 just because it uses an in-tree irrlicht, which uses X11 on unixes, but we have an irrlicht port which uses SDL for output.

I just eliminated every references for irrlicht and added support to use the system one. The project builds fine but crashes at start.
So technically it is possible to build irrlamb on Haiku, but it is also not trivial.

extrowerk:
Debug log: http://0x0.st/zhIi.report

extrowerk:
AFAIK @threedeyes made the Minetest port, where i had similar problem as i tried to port it. Maybe he can give some hint why it crashing.

alwayslivid:
> AFAIK @threedeyes made the Minetest port, where i had similar problem as i tried to port it. Maybe he can give some hint why it crashing.

I'm reopening the pull request.

PR #4599, opened 2020-01-16
freac: bump to 1.1~beta2

enzo1982:


media-sound/freac/freac-1.1~beta2.recipe
```diff
@@ -54,7 +54,7 @@ REQUIRES="
 	lib:libfaac$secondaryArchSuffix
 	lib:libfaad$secondaryArchSuffix
 	lib:libFLAC$secondaryArchSuffix
-	lib:libMAC$secondaryArchSuffix >= 5
+	lib:libMAC$secondaryArchSuffix >= 6
```
korli: I'm concerned that no recipe in the repo seems to use devel:libMAC, and yet the freac recipe is depending on lib:libMAC.
Could you clarify this?

korli:
Why not bump "boca${secondaryArchSuffix}_devel >= 1.0~beta1" to beta2?

enzo1982:
> Why not bump "boca${secondaryArchSuffix}_devel >= 1.0~beta1" to beta2?

It's not necessary. The API is stable now and fre:ac 1.1 beta 2 can use BoCA 1.0 beta 1 or later. So I thought it was more correct to keep it a the minimum required version.

media-sound/freac/freac-1.1~beta2.recipe
enzo1982: The Monkey's Audio library is loaded dynamically at runtime (using dlopen) which is why it is not required during build. fre:ac will run without it, but would be missing the Monkey's Audio codec then.

This would probably be a RECOMMENDS dependency if Haiku recipes supported that.

PR #4600, opened 2020-01-17
pcsc_lite: added recipe

alwayslivid:
This recipe is a WIP and not currently usable in its current state. Some of the known issues include:

* ~Undefined devel libraries~
* ~The recipe does not compile on x86_64 due to an undefined reference error during compilation, Makefile most likely needs patching.~
* ~Identation issues~
* ~Missing $secondaryArchSuffixes~
* ~Potentially a leftover `commandBinDir`-related case.~

~This recipe is a draft and the current goal is to get a proper `.hpkg` build with defined devel libraries.~

**UPDATE:** As of the time of this edit, an `.hpkg` package can be compiled and used with this recipe.

The problem is that the executable won't run in its current state.


alwayslivid:
I'm aware that the commas are wrong for the time being, I'll correct it once I have access to my development environment again.

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,56 @@
...
+to smartcard reader communication."
+COPYRIGHT="1999-2002 David Corcoran,
+	2001-2014 Ludovic Rousseau,
+	2003-2004 Damien Sauveron
+	"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://pcsclite.apdu.fr/files/pcsc-lite-$portVersion.tar.bz2"
+CHECKSUM_SHA256="d76d79edc31cf76e782b9f697420d3defbcc91778c3c650658086a1b748e8792"
+SOURCE_DIR="pcsc-lite-1.8.25"
```
korli:
```suggestion
SOURCE_DIR="pcsc-lite-$portVersion"
```

alwayslivid:
A library that's required for using the CLI tool is also bundled with the program, while my recipe does not seem to provide a library as of right now.

In [Arch Linux](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=pcsclite-git), there's no separate recipe for the library itself, nor any specific conditions in the corresponding port of the program. It does install the drivers manually though. I will look into the build log and share it here (as well as the compilation error for the sake of proper documentation and future reference) a bit later.

alwayslivid:
I got a build, will push in a bit.

sys-apps/pcsc_lite/patches/pcsc_lite-1.8.25.patchset
```diff
@@ -0,0 +1,23 @@
...
+diff --git a/configure.ac b/configure.ac
+index 0fa5168..a6f8660 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -147,6 +147,9 @@ PCSCLITE_FEATURES="${PCSCLITE_FEATURES} $PCSC_ARCH $host"
+ # See if socket() is found from libsocket
+ AC_CHECK_LIB(socket, socket, [LIBS="$LIBS -lsocket"])
+ 
++# Check for the network library.
++AC_CHECK_LIB(socket, socket, [LIBS="$LIBS -lnetwork"])
```
Begasus: Try with AC_SEARCH_LIBS instead

alwayslivid: Does that replace any functionality within the recipe?

Begasus: The idea is to pass the lib to configure with autoreconf

alwayslivid: Would this look alright?

image

Begasus: Look at other recipes how they use it, it bit overkill here :)

Begasus: Actually this works with a small change:
```diff
-+AC_CHECK_LIB(socket, socket, [LIBS="$LIBS -lnetwork"])
++AC_CHECK_LIB(network, socket, [LIBS="$LIBS -lnetwork"])
```

alwayslivid: Seems surprising enough to me. I pushed the changes, but I'll test it out in an hour.

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,92 @@
...
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	pcsc_lite$secondaryArchSuffix = $portVersion
+	lib:libpcsclite$secondaryArchSuffix = $libVersionCompat
+	lib:libpcscspy$secondaryArchSuffix = $libVersionCompat
+ 	cmd:pcsc_spy$commandSuffix = $portVersion
+	cmd:pcscd$commandSuffix = $portVersion
```
Begasus: cmd's before libs, also, there is an outstanding policy error if you build this for secondary architecture on gcc2h

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,92 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libusb_1.0$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	pcsc_lite${secondaryArchSuffix}_devel = $portVersion
+	devel:libpcsclite$secondaryArchSuffix = $libVersionCompat
```
Begasus: This libVersionCompat isn't correct :) (same for _devel)

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,95 @@
...
+	devel:libpcsclite$secondaryArchSuffix = $libpcscliteVersionCompat
+	devel:libpcscspy$secondaryArchSuffix = $libpcscspyVersionCompat
+	"
+REQUIRES_devel="
+	pcsc_lite$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:libusb_config$secondaryArchSuffix
```
Begasus: Check https://github.com/haikuports/haikuports/commit/5968411e28d563bbe1301e60cf96a7defc587eff

alwayslivid:
> Are you sure about these 2 lines?

I added them in a presumptive fashion because I wasn't sure as to whether anything in the headers could get borked by the lack of them, but in retrospect, I feel like this was a wrong move. I'll correct it as soon as I check `cmd:libusb_config`.

korli:
This should only  depend on libusb-1.0, not libusb-0.1

alwayslivid:
> This should only depend on libusb-1.0, not libusb-0.1

How about [this change](https://github.com/haikuports/haikuports/commit/5968411e28d563bbe1301e60cf96a7defc587eff) that was brought up by @Begasus a bit ago?

Begasus:
> This should only depend on libusb-1.0, not libusb-0.1

Even if it requires libusb_config?

alwayslivid:
> > This should only depend on libusb-1.0, not libusb-0.1
> 
> Even if it requires libusb_config?

It works without libusb_config that way, by the way.

Begasus:
> > > This should only depend on libusb-1.0, not libusb-0.1
> > 
> > 
> > Even if it requires libusb_config?
> 
> It works without libusb_config that way, by the way.

Ah! Then indeed probably only needs libusb_1.0

alwayslivid:
> > > > This should only depend on libusb-1.0, not libusb-0.1
> > > 
> > > 
> > > Even if it requires libusb_config?
> > 
> > 
> > It works without libusb_config that way, by the way.
> 
> Ah! Then indeed probably only needs libusb_1.0

No, I mean, it works *without* `cmd:libusb_config` and *with `lib:libusb_0.1`.

korli:
pcsc_lite seems to use libusb.h, see https://github.com/LudovicRousseau/PCSC/blob/master/src/hotplug_libusb.c#L54
which is only provided by libusb-1.0

Begasus:
Not quite there yet it seems:
```
~> pcscd -f
00000000 [24679488] pcscdaemon.c:623:main() cannot create /var/run/pcscd: No such file or directory
```

alwayslivid:
> Not quite there yet it seems:
> 
> ```
> ~> pcscd -f
> 00000000 [24679488] pcscdaemon.c:623:main() cannot create /var/run/pcscd: No such file or directory
> ```

I spent a lot of time looking into this, and I also added a few extra variables that would be (maybe) a good idea to remove later down the line in the meantime.

The variable that's needed to fix this is `--enable-ipcdir=DIR", however, I'm really unsure as to which directory should be used over `/var/run/pcscd`. Most server binaries are either in the `/boot/system/servers` folder (read-only directory). On the other hand, some daemon applications (e.g. `/boot/system/bin/sshd -D`) are in the `/boot/system/bin` folder.

The structure of this program seems very `systemd`-centric, in the sense that it seems like it's best used with it. However, the only other recipe that I can come up with that shows similar behavior is [Apache's recipe](https://github.com/haikuports/haikuports/blob/master/www-servers/apache/apache-2.4.41.recipe) and I'm not sure that I can tell apart anything useful.

korli:
/var/run/pcscd is meant to store a pcscd.pid file. I see the nginx recipe use $localStateDir for this usecase.

alwayslivid:
> /var/run/pcscd is meant to store a pcscd.pid file. I see the nginx recipe use $localStateDir for this usecase.

I think that this one's handled by `--enable-runpid`, I'll test it out either way.

alwayslivid:
Nope, seems like it's a read-only directory. (ipcdir points to `/packages/pcsc_lite-1.8.25-1/.self/var`, so there's that)

alwayslivid:
I attempted to change up things a little bit (using `$localStateDir/run` seems to have helped a bit). The behavior is a bit different-- it creates a config file unsuccessfully and prints out every single character individually (and defining it as a GLOBAL_WRITABLE_FILE doesn't quite do the trick-- it doesn't exist). The script then moves on with deleting the file located at the /var/run directory of the package, but it doesn't exist.

Here's some `tail -f` output of the `syslog`:

image

korli:
actually /var/run/pcscd seems correct, it just needs to be created by a post-install script (see www-servers/lighttpd/additional-files/create-lighttpd-directories.sh for instance)

alwayslivid:
> actually /var/run/pcscd seems correct, it just needs to be created by a post-install script (see www-servers/lighttpd/additional-files/create-lighttpd-directories.sh for instance)

I looked at the recipe, experimented and pushed the changes. Just creating a /var/run folder seems to work just fine, and there doesn't seem to be a need for the inclusion of an entire `.sh` file over a simple `mkdir` command. Should be fine now. (?)

Begasus:
> > actually /var/run/pcscd seems correct, it just needs to be created by a post-install script (see www-servers/lighttpd/additional-files/create-lighttpd-directories.sh for instance)
> 
> I looked at the recipe, experimented and pushed the changes. Just creating a /var/run folder seems to work just fine, and there doesn't seem to be a need for the inclusion of an entire `.sh` file over a simple `mkdir` command. Should be fine now. (?)

@korli any take on this? Creating the script shouldn't be too hard, but using mkdir in the recipe seems to do the same thing (or not?)

korli:
> > > actually /var/run/pcscd seems correct, it just needs to be created by a post-install script (see www-servers/lighttpd/additional-files/create-lighttpd-directories.sh for instance)
> > 
> > 
> > I looked at the recipe, experimented and pushed the changes. Just creating a /var/run folder seems to work just fine, and there doesn't seem to be a need for the inclusion of an entire `.sh` file over a simple `mkdir` command. Should be fine now. (?)
> 
> @korli any take on this? Creating the script shouldn't be too hard, but using mkdir in the recipe seems to do the same thing (or not?)

/var/run can't be part of the package. For this reason a post-install script is needed (see lighttpd for instance)

Begasus:
Thanks for the reply, so @alwayslivid could you take care of that?

alwayslivid:
> 
> 
> Thanks for the reply, so @alwayslivid could you take care of that?

I can only do it later today, unfortunately.

alwayslivid:
> 
> 
> Thanks for the reply, so @alwayslivid could you take care of that?

Actually @Begasus, I think that I managed to create a script on the go by using the `lighttpd` recipe and documentation as a point of reference, but I'm away from my development environment right now, so I'm not in the position to test it out right now.

If you could help with that, that would be appreciated, but I can still do it myself in a few hours (as well as fix any issues that may occur).

sys-apps/pcsc_lite/additional-files/create-pcscd-directory.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+mkdir -p var/run
```
korli:
```suggestion
mkdir -p /var/run
```

korli:
Looks good! To be tested as @alwayslivid suggested :)

alwayslivid:
> 
> 
> Looks good! To be tested as @alwayslivid suggested :)

I was about to commit the change in the directory path and also about to ask why the `lighttpd` uses var/run rather than /var/run, but you deleted in just in time.

But while we're at it, why does it actually do that? Just asking out of pure curiosity.

korli:
> But while we're at it, why does it actually do that? Just asking out of pure curiosity.

I'm surprised too. I suppose that it depends where the package is installed, user or system.
This directory will actually be created: /system/var/run (/var is a symlink to /system/var).

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,100 @@
...
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libpcscliteVersion="1.0.0"
+libpcscspyVersion="0.0.0"
+libpcscspyVersionCompat="$libpcscspyVersion compat >= ${libpcscspyVersion%%.*}"
+libpcscliteVersionCompat="$libpcscliteVersion compat >= ${libpcscliteVersion%%.*}"
+
+POST_INSTALL_SCRIPTS="$relativePostInstallDir/create-pcscd-directories.sh"
```
Begasus:
```diff
- POST_INSTALL_SCRIPTS="$relativePostInstallDir/create-pcscd-directories.sh"
+ POST_INSTALL_SCRIPTS="$relativePostInstallDir/create-pcscd-directory.sh"
```

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,100 @@
...
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make $jobArgs install
+
+	mkdir -p $postInstallDir
+	cp -f $portDir/additional-files/create-pcscd-directories.sh $postInstallDir
```
Begasus:
```diff
- 	cp -f $portDir/additional-files/create-pcscd-directories.sh $postInstallDir
+ 	cp -f $portDir/additional-files/create-pcscd-directory.sh $postInstallDir
```

Begasus: OK, seems you need to set the script to be executable :)

alwayslivid: Hopefully this commit that I just pushed will do it. I'm not marking the issue as "Resolved" for the time being.

sys-apps/pcsc_lite/pcsc_lite-1.8.25.recipe
```diff
@@ -0,0 +1,101 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make $jobArgs install
+
+	mkdir -p $postInstallDir
+	cp -f $portDir/additional-files/create-pcscd-directory.sh $postInstallDir
+	chmod +X $postInstallDir/create-pcscd-directory.sh
```
Begasus: This probably will work OK, but maybe you can set the executable at the script prior and push that one?
@korli if a script is set as executable, will it stay that way when you push to github?
(don't see the script for lighttpd being modified in the recipe)

alwayslivid: File permissions get retained on Git as far as I know.

I'll have to deal with this the way you're instructing me to in approximately 2 hours.

korli: > File permissions get retained on Git as far as I know.
 This is correct.

PR #4605, opened 2020-01-17
openxcom: clean up libSDL dependencies

Ra-V2:


Begasus:
Could you provide a buildlog?

Ra-V2:
@Begasus  It seems we have a lot of problems here:
[openxcom.txt](https://github.com/haikuports/haikuports/files/4080664/openxcom.txt)

PR #4607, opened 2020-01-17
mupen64: clean up libSDL dependencies

Ra-V2:


korli:
This is not needed.

Begasus:
Style change can be done after a version bump, but thanks for looking into it!

PR #4593, opened 2020-01-16
recipe for Shotcut video editor

TURX:
This is for #4028.

PR #4611, opened 2020-01-17
recipe for qtx11extras

TURX:
This is for #4593.


TURX:
I do not know if the build produces any file. The log is at https://paste.ubuntu.com/p/pb4YrbD5cY/.

extrowerk:
“ mv: cannot stat '/packages/qtx11extras-5.9.9-1/.self/lib/qml': No such file or directory”
This is obviously wrong.

Begasus:
Maybe it would be good to test a recipe locally before pushing to a PR?

Begasus:
For now, not needed

PR #4612, opened 2020-01-17
Updated recipe for elixir

trungnt2910:
I've updated the recipe for elixir from v1.4.2 to v1.9.4 (latest).
Terminal window running elixir v1.9.4
Elixir v1 9 4


extrowerk:
Please do not link packages here, we are sure you dont want to steal our password, data, etc with a malicious crafted package, but the devil never sleeps.
So please remove that link. Thank you.

extrowerk:
I forgot that i can also remove it.

trungnt2910:
> Could you please remove the old recipe? Thank you!
I've removed it, like you said.
However, I've seen that there are many apps in haikuports that keep recipes of their older versions.

extrowerk:
But now we can see what did you changed: https://github.com/haikuports/haikuports/pull/4612/files

Begasus:
LGTM, ready to merge @extrowerk ?

trungnt2910:
In the last commit, I've updated the build requirements for Elixir. The documentation says that Elixir 1.9.4 requires Erlang 20.0 or later, so I fixed the recipe to check for this.

Begasus:
Not on x86_64 atm so can't check, could you provide a buildlog?

trungnt2910:
You mean when running haikuporter?

Begasus:
> You mean when running haikuporter?

Yes (output in Terminal)

trungnt2910:
Here is the terminal output:

[elixir_build.log](https://github.com/haikuports/haikuports/files/4080515/elixir_build.log)

Begasus:
Ah :) just saw that there isn't actually anything build :)
Would be nice to see if this would work on the gcc2h, but that's another thing (erlang needs to be checked if it builds there)

trungnt2910:
Thanks.

PR #4613, opened 2020-01-17
simutrans: fix references to libSDL*

AlwaysLivid:


Begasus:
@taosxx as you are keeping track on this, this is an older recipe, maybe we can just drop it?

alwayslivid:
@korli I tried to compile it, but there seems to be a definition conflict with the integers that was also mentioned in the Haiku IRC a few days ago.

Should I patch the definitions out?

korli:
> @korli I tried to compile it, but there seems to be a definition conflict with the integers that was also mentioned in the Haiku IRC a few days ago.
> 
> Should I patch the definitions out?

Try to build on x86_64 first. We can disable the build on 32-bit if required.

alwayslivid:
> > @korli I tried to compile it, but there seems to be a definition conflict with the integers that was also mentioned in the Haiku IRC a few days ago.
> > Should I patch the definitions out?
> 
> Try to build on x86_64 first. We can disable the build on 32-bit if required.

Done, the recipe builds, but it is going to need further updates due to unmet dependencies. (namely, svn.)

https://paste.ubuntu.com/p/cQCJHtxbDX

Begasus:
There is already an updated recipe for this, should we still look into this?

Begasus:
Closing as the newer versions already have the fix

PR #4603, opened 2020-01-17
xut: clean up libSDL dependencies

Ra-V2:

PR #4616, opened 2020-01-18
xut: disable recipe

Ra-V2:
For now, the sources cannot be downloaded from the URI, so I disable this recipe.
https://github.com/haikuports/haikuports/pull/4603


Begasus:
Could you mark secondary architecture as broken also?

Begasus:
While you are at it, could you file an issue about the missing download link?

Issue #233, opened 2015-06-30
JACK


soakbot:
(Original issue 233 created by danboid on 2015-05-13T22:37:47.602634+00:00)

I have been successful in getting JACK 0.124.1 to build under Haiku but with the following gotchas:

* It only builds with GCC4.
* There is no support in JACK for Haiku's native sound system, mediaserver, and I've been unable to get the OSS backend to pass the required JACK configure checks so that it gets enabled and built. This means I have not been able to properly test JACK under Haiku yet.
* There is currently an issue with the libvorbis devel package in HaikuDepot that prevents JACK compiling if you have libsndfile devel package installed and enabled when configuring. Rather than re-compile the packages in question I decided just to uninstall libsndfile devel, without which you don't get the jackrec example client.

I have also had success getting ecasound to build under Haiku. In the case of ecasound I have managed to get it to build with OSS support so there's no reason we can't do the same with JACK. I'd like to get JACK working and packaged first so then we can build and package ecasound with JACK support. I've not tried it yet but apparently supercollider works under Haiku so that would be another Haiku app that could take advantage of JACK. Personally I'd love to see the Muse DAW get ported to Haiku, as has already been done with LMMS. Ardour needs GTK and a bunch of other libraries not ported to Haiku whilst qtractor is heavily dependent upon ALSA for its MIDI support so that won't port easily either. Muse uses Qt and JACK for both audio and MIDI.

If you'd like to help me get JACK working properly under Haiku, read on!

All the dependencies required to build JACK are currently packaged in depot except for Berkeley DB, which is available via Haikuporter. You will need to install the x86 (GCC4) version of db and its devel package which you get but running:

```
#!bash

hp db_x86
```

That's assuming you have downloaded and configured haikuporter, of course, which I'm not going to cover here.

Download and uncompress http://jackaudio.org/downloads/jack-audio-connection-kit-0.124.1.tar.gz

First we need to fix a prob in 0.124.1 that affects all platforms that has been fixed in git now. JACK tries to build with a file called uuid/uuid.h that it doesn't need anymore so remove or comment out the line

```
#!C

#include <uuid/uuid.h>
```

From these files:

```
#!bash

jack/types.h
jackd/clientengine.c
libjack/port.c:
```

At around line 168 in configure.ac we need to add a new case statement for Haiku to disable mlock so add these three lines after the darwin case but before the esac:

```
#!bash

 Haiku*)
    JACK_DO_NOT_MLOCK=1
    ;;
```

[...]

PR #4623, opened 2020-01-20
recipe for jack1

TURX:
This is for #4593 as a dependency.

### DIFF
--- a/media-sound/jack1/jack1-0.125.0.recipe
+++ b/media-sound/jack1/jack1-0.125.0.recipe
@@ -32,25 +32,22 @@ PROVIDES="
 	cmd:jack_alias
 	cmd:jack_bufsize
 	cmd:jack_connect
-	cmd:jack_control
-	cmd:jack_cpu
-	cmd:jack_cpu_load
 	cmd:jack_disconnect
 	cmd:jack_evmon
 	cmd:jack_freewheel
+	cmd:jack_impulse_grabber
 	cmd:jack_iodelay
 	cmd:jack_latent_client
 	cmd:jack_load
+	cmd:jack_load_test
 	cmd:jack_lsp
 	cmd:jack_metro
 	cmd:jack_midi_dump
-	cmd:jack_midi_latency_test
 	cmd:jack_midiseq
 	cmd:jack_midisine
 	cmd:jack_monitor_client
-	cmd:jack_multiple_metro
-	cmd:jack_net_master
-	cmd:jack_net_slave
+	cmd:jack_netsource
+	cmd:jack_property
 	cmd:jack_rec
 	cmd:jack_samplerate
 	cmd:jack_server_control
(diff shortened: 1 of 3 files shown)


TURX:
There is still a lack of some functions: https://paste.ubuntu.com/p/y3NPGz3kwh/

Begasus:
Check issue: https://github.com/haikuports/haikuports/issues/233

TURX:
The build has gone successful, but a warning prevents this to be a package.

https://paste.ubuntu.com/p/Sp8Scf5VJ3/

```
Warning: 'NoneType' object has no attribute 'split'
```

media-sound/jack1/jack1-0.125.0.recipe
```diff
@@ -0,0 +1,129 @@
...
+	lib:libreadline$secondaryArchSuffix
+	lib:libsamplerate$secondaryArchSuffix
+	lib:libsndfile$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	jack1${secondaryArchSuffix}_devel
+	devel:libjack = 0.1.0 compat >= 0
+	devel:libjacknet = 0.1.0 compat >= 0
+	devel:libjackserver = 0.1.0 compat >= 0
```
Begasus: Missing $secondaryArchSuffix

media-sound/jack1/jack1-0.125.0.recipe
```diff
@@ -0,0 +1,129 @@
...
+	"
+REQUIRES_devel="
+	jack1$secondaryArchSuffix == base
```
Begasus: Missing $portVersion

TURX: Where to add the $portVersion?

TURX:
After building and installing jack1, Shotcut crashed.

Terminal log: https://paste.ubuntu.com/p/QHYMYvRpV4/
Error report: https://paste.ubuntu.com/p/gJvSSFrBrP/

How could I disable munlock()?

korli:
Shotcut doesn't have a direct dependency on jack; mlt should then be built with jack. Maybe it's better first to disable the warning in shotcut, and investigate whether we can start with shotcut+mlt+sdl.

Begasus:
mlt has been rebuild with extra modules that should take care of this, no need for jack ... (crash for shotcut doesn't mention jack)

alaviss:
Why are we adding JACK to Haiku? Don't we already have Media Kit?

Also, this recipe doesn't seem to patch in a media backend for Haiku, so it wouldn't be able to do JACK things (pluggable outputs), making it rather useless.

korli:
> Why are we adding JACK to Haiku? Don't we already have Media Kit?
> 
> Also, this recipe doesn't seem to patch in a media backend for Haiku, so it wouldn't be able to do JACK things (pluggable outputs, or even audio), making it rather useless.

One could use the network backend :)

Begasus:
There is already a jack2 recipe (marked broken), closing this one too

PR #4624, opened 2020-01-20
fix amd64 build of jack2

TURX:
This is for #4593 as a dependency.


Begasus:
This is not to be used as explained in the issue ...

PR #4625, opened 2020-01-20
Added recipe for QGit

trungnt2910:
I've added the recipe for [QGit](https://github.com/tibirna/qgit).
I've packaged with haikuporter and tested the package on a clean installation of Haiku (in order to check the package's dependencies).
QGit works nicely without patches.
QGit

### DIFF
--- a/dev-vcs/qgit/qgit-2.9.recipe
+++ b/dev-vcs/qgit/qgit-2.9.recipe
@@ -1,0 +1,56 @@
+SUMMARY="A Qt-based git GUI viewer"
+DESCRIPTION="With qgit you will be able to browse revisions history,\
+view patch content and changed files, graphically following different \
+development branches."
+HOMEPAGE="https://github.com/tibirna/qgit"
+COPYRIGHT="2005-2008 Marco Costalba
+	2011-2019 Cristian Tibirna"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/tibirna/qgit/archive/qgit-$portVersion.tar.gz"
+CHECKSUM_SHA256="69788efda7d434d1bc094bb414bd92c269dc7894326320634500b05d63c008e8"
+SOURCE_DIR="qgit-qgit-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qgit$secondaryArchSuffix = $portVersion
+	app:QGit
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	cmd:git
+	lib:libGL$secondaryArchSuffix
+	lib:libQt5Core$secondaryArchSuffix
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:qmake$secondaryArchSuffix >= 5
+	"
+
+defineDebugInfoPackage qgit$secondaryArchSuffix \
+	$appsDir/QGit
+
+BUILD()
+{
+	qmake qgit.pro
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	cp bin/qgit $appsDir/QGit
+	addAppDeskbarSymlink $appsDir/QGit
+}


dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+	app:QGit
+	"
+REQUIRES="
+	haiku
+	cmd:git
+	lib:libGL$secondaryArchSuffix
+	lib:libQt5Core$secondaryArchSuffix
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	qthaikuplugins$secondaryArchSuffix
```
korli: Should not depend on qthaikuplugins.
```suggestion
```

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	mkdir -p $binDir
+	cp bin/qgit $appsDir
+	mv $appsDir/qgit $appsDir/QGit
+	cp $appsDir/QGit $binDir/qgit
```
korli: Should be a symlink instead of a copy.

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+BUILD()
+{
+	qmake qgit.pro
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir
+	mkdir -p $binDir
```
korli: Please combine the two lines in one.

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: I've tried ln -s $appsDir/QGit $binDir/qgit but it says that the target does not exist.
How should I fix this?

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: The older version of my recipe does not include `qthaikuplugins`. However, when installing the resulting `.hpkg` file, QGit says that `qthaikuplugins` are missing and opens a dialog allowing the user to install it.
Adding this line solves this problem.

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: I've fixed this.

trungnt2910:
I've seen your comments, will check about these later, as well as adding screenshots of errors I've met.

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: Haikux64_clean-2020-01-21-12-52-53

This is what running the QGit executable while having `qthaikuplugins` uninstalled give.

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,59 @@
...
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qgit$secondaryArchSuffix = $portVersion
+	cmd:qgit
+	app:QGit
+	"
+REQUIRES="
+	haiku
```
Begasus: Needs $secondaryArchSuffix
```diff
- 	haiku
+ 	haiku$secondaryArchSuffix
```

dev-vcs/qgit/qgit-2.9.recipe
Begasus: Any take on this, on bare install these plugins aren't installed?

korli: Every Qt application will show this popup. The dependency is not "hard". On a blank system installing QGit will install the required and recommended packages.

dev-vcs/qgit/qgit-2.9.recipe
korli: You mean the target already exists?

dev-vcs/qgit/qgit-2.9.recipe
Begasus: Clear enough :)

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,59 @@
...
+SOURCE_URI="https://github.com/tibirna/qgit/archive/qgit-$portVersion.tar.gz"
+CHECKSUM_SHA256="69788efda7d434d1bc094bb414bd92c269dc7894326320634500b05d63c008e8"
+SOURCE_DIR="qgit-qgit-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qgit$secondaryArchSuffix = $portVersion
+	cmd:qgit
```
Begasus: Needs $secondaryArchSuffix
```diff
- 	cmd:qgit
+ 	cmd:qgit$secondaryArchSuffix
```

trungnt2910: So all of the requirements needs $secondaryArchSuffix

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: No. The target **does not** exist.

trungnt2910: @korli This time, I cleaned the whole work directory, and built it from scratch, and the symlink worked.

dev-vcs/qgit/qgit-2.9.recipe
Begasus: @trungnt2910 can you take care of this?

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,59 @@
...
+	qmake qgit.pro
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p {$appsDir,$binDir}
+	cp bin/qgit $appsDir
+	mv $appsDir/qgit $appsDir/QGit
+	ln -s $appsDir/QGit $binDir/qgit
```
Begasus: Any specific reason you want this in $binDir?

trungnt2910: Well, this app is a GUI app but can also work in the terminal as a command-line app, so I put it in `$binDir` where all the command-line apps belong.

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,59 @@
+SUMMARY="A Qt-based git GUI viewer"
+DESCRIPTION="With qgit you will be able to browse revisions history,\
+view patch content and changed files, graphically following different \
+development branches."
+HOMEPAGE="https://github.com/tibirna/qgit"
+COPYRIGHT="2011-2019 Cristian Tibirna"
```
Begasus: About window shows more:
```diff
- COPYRIGHT="2011-2019 Cristian Tibirna"
+ COPYRIGHT="2005-2008 Marco Costalba
+	2011-2019 Cristian Tibirna"
```

trungnt2910: Only now do I know that it's possible to include more than one line in `COPYRIGHT`.

dev-vcs/qgit/qgit-2.9.recipe
Begasus: You can use QGit from Terminal if it's installed in $appsDir :)

Begasus: Just checked, no options when launching from Terminal, so no use of it linking to $binDir

dev-vcs/qgit/qgit-2.9.recipe
Begasus: It's explained in the wiki guidelines :)

Begasus: Updated comment
Seems there were more people involved between 2005-2008

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+SOURCE_URI="https://github.com/tibirna/qgit/archive/qgit-$portVersion.tar.gz"
+CHECKSUM_SHA256="69788efda7d434d1bc094bb414bd92c269dc7894326320634500b05d63c008e8"
+SOURCE_DIR="qgit-qgit-$portVersion"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	qgit$secondaryArchSuffix = $portVersion
+	cmd:qgit$secondaryArchSuffix
```
Begasus: Not needed:
```diff
- 	cmd:qgit$secondaryArchSuffix
```

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: This is from QGit's [official page](https://github.com/tibirna/qgit)
> 
> Run qgit from a git working directory, command line arguments are filtered by git log. Some examples:
> qgit --no-merges
> qgit v2.6.18.. include/scsi drivers/scsi
> qgit --since="2 weeks ago" -- kernel/
> qgit -r --name-status release..test
> If qgit is launched without arguments or if you change archive with open menu, a dialog for range select is shown. You can select top and bottom rev tags from the list or paste a specific revision. Values are passed to git log to narrow data loading to chosen revisions.
> As a special command-line argument, --view-file foo or --view-file=foo is not passed to git log, but opens the file foo in the QGit file viewer, showing the history and the annotated (blame) view for the file.

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+	app:QGit
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	cmd:git
+	lib:libGL$secondaryArchSuffix
+	lib:libQt5Core$secondaryArchSuffix
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	qthaikuplugins$secondaryArchSuffix
```
Begasus: Please remove (as @korli already mentioned):
```diff
- 	qthaikuplugins$secondaryArchSuffix
```

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: I think it is some command line options? If it is not, then I will remove the links and also `cmd:qgit` from the PROVIDES section.

dev-vcs/qgit/qgit-2.9.recipe
```diff
@@ -0,0 +1,60 @@
...
+	qmake qgit.pro
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p {$appsDir,$binDir}
+	cp bin/qgit $appsDir
+	mv $appsDir/qgit $appsDir/QGit
+	ln -s $appsDir/QGit $binDir/qgit
```
Begasus: Suggestion:
```diff
-	mkdir -p {$appsDir,$binDir}
-	cp bin/qgit $appsDir
-	mv $appsDir/qgit $appsDir/QGit
-	ln -s $appsDir/QGit $binDir/qgit
+	mkdir -p $appsDir
+	cp bin/qgit $appsDir/QGit
```

trungnt2910: Thanks.

dev-vcs/qgit/qgit-2.9.recipe
trungnt2910: Yes, I've read his comment. Thanks.

dev-vcs/qgit/qgit-2.9.recipe
Begasus: I think the options (haven't checked those) can be used with QGit the same as with qgit, need to check

trungnt2910:
Thanks. I've made the edits for the recipe.

dev-vcs/qgit/qgit-2.9.recipe
Begasus: Even when linking it to $prefix/bin it launches the GUI here:
QGit

trungnt2910: I see it. So there's no point in including cmd:qgit.

PR #4619, opened 2020-01-19
Updated the MPD recipe to latest released version

trungnt2910:
I've updated MPD to the latest release version [0.21.19](http://www.musicpd.org/download/mpd/0.21/mpd-0.21.19.tar.xz).
The updated recipe builds and packages properly, also, the package can install.

[mpd_build.log](https://github.com/haikuports/haikuports/files/4082880/mpd_build.log)

PR #4628, opened 2020-01-21
Fixed boost dependencies for mpd

trungnt2910:
Fixed dependencies for boost in recipe for mpd-0.21.19 (#4619)
This should fix gcc2h broken package.


Begasus:
Thanks, will check in a bit :+1:

Begasus:
Please, also bump REVISION

trungnt2910:
Ok. I've done that.

Begasus:
Build is almost finished here, so the requirements seems to be OK :)

trungnt2910:
Thanks.

korli:
Why not >= 1.66.0?

Begasus:
> Why not >= 1.66.0?

The lowest version required seemed to be >= 1.58
Checked a local build with this, worked ok, and the buildbot seems to be ok with it too

Begasus:
For the record, I have 1.69.0 here

korli:
> For the record, I have 1.69.0 here

One day, we want to drop everything older than 1.69. So we should tell the builders to use a current version. Otherwise users will have to install multiple versions.

PR #4632, opened 2020-01-21
edgar: correct user settings directory

TURX:


Begasus:
Please also bump the REVISION so we can have the buildbot pick this up for a rebuild

TURX:
It seems to be working for me :)

Begasus:
Doesn't seem to work on x86_gcc2 (checked with gcc2 and gcc8)

TURX:
"path" is undefined at l83, and there is an error at l82 (the definition of "path"): parse error before `char'.
Plus, what is before "char" is "#ifdef __HAIKU__".
What is the problem here?

Begasus:
LGTM, builds ok on x86_gcc2, @korli could you give the patchset a closer look? :)

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
++#endif
++
+ static char gameSavePath[MAX_PATH_LENGTH], tempFile[MAX_PATH_LENGTH], saveFileIndex[MAX_PATH_LENGTH], continueFile[MAX_PATH_LENGTH];
+ static int temporaryDataExists;
+ 
+@@ -57,6 +62,7 @@ extern Game game;
+ 	{
+ 		char *userHome;
+ 		char dir[MAX_PATH_LENGTH];
++		char path[MAXPATHLEN];
```
alaviss:
```suggestion
+		char path[MAX_PATH_LENGTH];
```

`dir` is no bigger than this, and for consistency reasons.

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
+ 		struct passwd *pass;
+ 
+ 		pass = getpwuid(geteuid());
+@@ -73,8 +79,14 @@ extern Game game;
  		#if MACOS == 1
  		snprintf(dir, sizeof(dir), "%s/Library/Application Support", userHome);
  		#else
--		snprintf(dir, sizeof(dir), "%s/.parallelrealities", userHome);
-+		snprintf(dir, sizeof(dir), "%s/config/settings/parallelrealities", userHome);
++		#ifdef __HAIKU__
```
alaviss:
```suggestion
+		#elif defined(__HAIKU__)
```

Use `#elif` to keep the code short :)

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
  		if ((mkdir(dir, S_IRWXU|S_IRWXG|S_IROTH|S_IXOTH) != 0) && (errno != EEXIST))
-@@ -86,7 +86,7 @@ extern Game game;
+ 		{
+@@ -86,8 +98,14 @@ extern Game game;
  		#if MACOS == 1
  		snprintf(dir, sizeof(dir), "%s/Library/Application Support/Edgar", userHome);
  		#else
--		snprintf(dir, sizeof(dir), "%s/.parallelrealities/edgar", userHome);
-+		snprintf(dir, sizeof(dir), "%s/config/settings/parallelrealities/Edgar", userHome);
++		#ifdef __HAIKU__
```
alaviss:
```suggestion
+		#elif defined(__HAIKU__)
```

Same as above.

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
  		if ((mkdir(dir, S_IRWXU|S_IRWXG|S_IROTH|S_IXOTH) != 0) && (errno != EEXIST))
-@@ -99,7 +99,7 @@ extern Game game;
+ 		{
+@@ -99,8 +117,14 @@ extern Game game;
  		#if MACOS == 1
  		snprintf(gameSavePath, sizeof(gameSavePath), "%s/Library/Application Support/Edgar/", userHome);
  		#else
--		snprintf(gameSavePath, sizeof(gameSavePath), "%s/.parallelrealities/edgar/", userHome);
-+		snprintf(gameSavePath, sizeof(gameSavePath), "%s/config/settings/parallelrealities/Edgar/", userHome);
++		#ifdef __HAIKU__
```
alaviss:
```suggestion
+		#elif defined(__HAIKU__)
```

Same as above.

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
 --- a/src/system/load_save.c
 +++ b/src/system/load_save.c
-@@ -73,7 +73,7 @@ extern Game game;
+@@ -41,6 +41,11 @@ Foundation, 51 Franklin Street, Suite 500, Boston, MA 02110-1335, USA.
+ #include "load_save.h"
+ #include "resources.h"
+ 
++#ifdef __HAIKU__
++#include <FindDirectory.h>
++#include <string.h>
```
alaviss:
```suggestion
```

Should not be required?

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
+ 		pass = getpwuid(geteuid());
+@@ -73,8 +79,14 @@ extern Game game;
  		#if MACOS == 1
  		snprintf(dir, sizeof(dir), "%s/Library/Application Support", userHome);
  		#else
--		snprintf(dir, sizeof(dir), "%s/.parallelrealities", userHome);
-+		snprintf(dir, sizeof(dir), "%s/config/settings/parallelrealities", userHome);
++		#ifdef __HAIKU__
++		path[MAXPATHLEN] = 0;
++		find_directory(B_USER_SETTINGS_DIRECTORY, 0, false, path, MAXPATHLEN);
```
alaviss:
```suggestion
+		find_directory(B_USER_SETTINGS_DIRECTORY, 0, true, path, sizeof(path));
```

The application is going to create a path here, so we can give them some help.

Using `sizeof` allows us to change the path size as needed.

alaviss: We should also check for errors here, but since the application doesn't even try to do so for its other paths... maybe we can just pass.

games-arcade/edgar/patches/edgar-1.30.patchset
```diff
@@ -1,40 +1,81 @@
...
+ 
+ 		pass = getpwuid(geteuid());
+@@ -73,8 +79,14 @@ extern Game game;
  		#if MACOS == 1
  		snprintf(dir, sizeof(dir), "%s/Library/Application Support", userHome);
  		#else
--		snprintf(dir, sizeof(dir), "%s/.parallelrealities", userHome);
-+		snprintf(dir, sizeof(dir), "%s/config/settings/parallelrealities", userHome);
++		#ifdef __HAIKU__
++		path[MAXPATHLEN] = 0;
```
alaviss:
```suggestion
```

There is no need for this, `find_directory` will terminate the path with a `\0`, and it won't exceed the size you provided. Also the code is wrong since you're overstepping the bounds by a byte (max index of array is length - 1)

TURX:
> The recipe should define `USER_SETTINGS_FILES` as the program saves data into user settings directory.
> 
> Something like this should do:
> 
> ```shell
> USER_SETTINGS_FILES="settings/parallelrealities/Edgar"
> ```
> 
> Consult the Guidelines for where to insert this variable.

I can find no USER_SETTINGS_FILES in the source code.

games-arcade/edgar/edgar-1.30.recipe
```diff
@@ -18,6 +18,8 @@ ADDITIONAL_FILES="edgar.rdef.in"
 
 ARCHITECTURES="x86_gcc2 x86 x86_64"
 
+USER_SETTINGS_FILES="settings/parallelrealities/Edgar directory"
```
Begasus: Suggestion:
```diff
- USER_SETTINGS_FILES="settings/parallelrealities/Edgar directory"
+ USER_SETTINGS_FILES="
+	settings/parallelrealities/Edgar directory
+	"
```

Begasus:
Checked with saving and loading, works OK, thanks!

PR #4442, opened 2019-12-12
Added dbuspython recipe.

alwayslivid:
Had some trouble with a trailing space and a few other string and directory related issues.

Feel free to merge in one commit.

PR #4654, opened 2020-01-26
DBUS_python: new recipe

extrowerk:


alwayslivid:
I did work on a [dbus-python recipe](https://github.com/haikuports/haikuports/pull/4442) the other day, although I can see that the name, as well as some metadata, is different and perhaps (I don't think that I'm the person to judge) even better than the metadata on the recipe that I worked on in some ways.

However, I remember that there was a specific reason that we opted for Python 3.6 and 3.7 and make, rather than using setuptools. We also did drop 2.7 consciously. However, I can't remember any specific details.

CC: @Begasus and @alaviss?

extrowerk:
Havent seen your recipe. Will drop this then.
Calibre uses python2.x.

Begasus:
Think there was a problem with python 2.7 on the gcc2hybrid: https://github.com/haikuports/haikuports/pull/4442#discussion_r358357903

alwayslivid:
> Think there was a problem with python 2.7 on the gcc2hybrid: [#4442 (comment)](https://github.com/haikuports/haikuports/pull/4442#discussion_r358357903)

Since getting Calibre to work would be pretty nice, would it be a good idea to consider using setuptools with Python 2.7 under specific circumstances (e.g. !x86_gcc2)?

extrowerk:
Whatever, it is an optional dep for Calibre.

alwayslivid:
> Whatever, it is an optional dep for Calibre.

Understood.

PR #4657, opened 2020-01-26
Calibre: new recipe

extrowerk:
Current state:
* this is the most recent version which doesn't requires qtwebengine
* Calibre starts and works **if started from Terminal**
* Crashing if started from Tracker/Deskbar
* Device management not working, you won't be able to synchronize your books automatically

Except these, it is pretty stable, the viewer works pretty good.

Starting with Haiku UI style is possible:
```
export CALIBRE_USE_SYSTEM_THEME=1
```


extrowerk:
From my side it is ok to merge, because  i definetely can't fix the listed bugs.

extrowerk:
It seems starting from Tracker/Deskbar works after running it from Terminal first.

theclue:
I've found an error while launching from Terminal which prevents the app starting:

`DBusExport: Failed to import dbus, with error: no module named dbus`

This was in my first launch with no device connected.

hope this helps!

diversys:
Does installing dbus help?

extrowerk:
> Does installing dbus help?

No idea to which listed problem are @theclue referring, but probably to the device sync one, but that doesn't makes any difference, as installing dbus won't help: the device discovery code needs to be rewritten for Haiku.

theclue:
> Does installing dbus help?

yes, installing dbus makes calibre stars. Ofc I'm not expecting the device connection to work (i've not tried actually). While trying to open an epub it goes in out of memory error (how much memory does this app require? omg)

extrowerk:
> > Does installing dbus help?
> 
> yes, installing dbus makes calibre starts. Ofc I'm not expecting the device connection to work (i've not tried actually). While trying to open an epub it goes in out of memory error (how much memory does this app require? omg)

That's new, i tought i disabled every dbus stuff in it. Thanks for the feedback,we can add it to the requiments.

>out of memory
Calibre a bunch of different technologies bolted on eachother, the reason for oom can be lurk everywhere.
Calibre works ok for me on 64 bit Haiku tho.

PR #4662, opened 2020-01-27
ripgrep: enable PCRE2 suppport in build

Crestwave:


extrowerk:
Build failed on 32 bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/3862/builds/34412.log

Crestwave:
Sorry for the late reply; I didn't see the notification. Why is it using Rust 1.15.1? That seems to be the issue and not anything with this particular change; `ripgrep` only started passing all the tests at Rust 1.40.0.

Crestwave:
Oh, it's not rust_bin and apparently the secondary architecture wasn't marked as untested before so could it be possible that it actually managed to build Rust or am I missing something?

Crestwave:
@nielx Are you aware of this (`rust_x86` 1.15.1 building successfully)? Also, I'm going to specify a minimum `cmd:cargo` version to fix this, and I noticed that the version in the recipe differs from `cargo --version`. Is there a reason for this? Should I require `cmd:rustc` instead for consistency even when I only directly execute `cargo`?

nielx:
Rust used to build on Haiku. Around 1.15 sound about right when it became impossible to build. Bumping the minimum versions sounds smart

Crestwave:
I see. Strange that it only started choosing rust_x86 instead of rust_bin_x86 just now, though. Any comment on the cargo vs rustc versions?

nielx:
Are you talking about the rust_bin 1.41 package? It is indeed very odd. The build output definitely puts it at [version 0.42.0](http://dl.rust-on-haiku.com/dist/1.41.0/), but cargo --version currently outputs 1.41.0. 

I don't quite know the logic, but the [Cargo.toml ](https://github.com/rust-lang/cargo/blob/rust-1.41.0/Cargo.toml) also specifies version 0.42.0. The explanation for this can be found in [the source](https://github.com/rust-lang/cargo/blob/674453308d776a31a877f925e9c6aeace1e53575/src/cargo/lib.rs#L200).

For the current way things are packaged, I don't see any harm in the way the way we version Cargo. I think you should depend on a minimum version of rustc, and not on cargo.

Crestwave:
Ah, I see. Got it, thanks.

PR #4663, opened 2020-01-27
libu2f_host: removed unnecessary dependencies

alwayslivid:
Successfully compiles on x86_64.

`cmd:automake` seems unnecessary and most other recipes use `cmd:autoreconf`, rather than `cmd:autoconf`.


korli:
This is OK, provided the packages are not referenced by name.

PR #4668, opened 2020-01-30
partio: fix and enable x86 32 bit build

X547:


Begasus:
Any thought on pushing this to the latest release? (haven't checked but saw there are new releases for it)

media-libs/partio/patches/partio-1.1.0.patchset
```diff
@@ -18,5 +21,32 @@ index e331c9c..9c592ca 100644
...
+ src/lib/io/MC.cpp | 4 ++--
+ 1 file changed, 2 insertions(+), 2 deletions(-)
+
+diff --git a/src/lib/io/MC.cpp b/src/lib/io/MC.cpp
+index 7125954..7460c0b 100644
+--- a/src/lib/io/MC.cpp
++++ b/src/lib/io/MC.cpp
+@@ -105,8 +105,8 @@ bool ReadAttrHeader(std::istream& input, Attribute_Header& attribute){
+ 
+ int CharArrayLen(char** charArray){
```
Begasus: Upstream has https://github.com/wdas/partio/blob/032a1ec24246b28ad26dbc7f163417c1b921de42/src/lib/io/MC.cpp#L106
Could this work (doesn't bail or errors out here)?

Begasus:
Doing a check with version 1.10.1 (for this) for OSL, so far it picks up partio

X547:
Latest version has problems with python. Python library path is wrong.

Begasus:
> Latest version has problems with python. Python library path is wrong.

OK, strike that then :+1:

Begasus:
Build current recipe on you changes I saw that it only builds a static lib, maybe you can import an upstream patch to also include the shared lib?
https://github.com/wdas/partio/commit/56ddeea28d1d31ee6f74224c9b590bef1118717e#diff-af3b638bc2a3e6c650974192a53c7291

PR #4669, opened 2020-01-30
osl: fix and enable build on 32 bit x86 (not yet working)

X547:


X547:
This patch is not working yet. Build fails in `OpenShadingLanguage-Release-1.10.7/src/liboslexec/pointcloud.cpp`:
```cpp
    static_assert (sizeof(size_t) == sizeof(Partio::ParticleIndex),
            "Only will work if Partio ParticleIndex is the size of a size_t");
    // FIXME -- if anybody cares about an architecture in which that is not
    // the case, we can easily allocate local space to retrieve the indices,
    // then copy them back to the caller's indices.
```

Begasus:
There is release 1.10.8 out, maybe check that one?

Begasus:
https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-openshadinglanguage/0013-disable-assert-partio.patch
Haven't checked though ...

Begasus:
The patch seems to be working, on another note, is partio really required here? (Looking at some other OS recipes it doesn't seem obligated?) (also haven't checked)

kenmays:
Tested with osl-1.10.13 on x86 R1B2:
```
mimesetting files for package osl_x86-1.10.13-1-x86_gcc2.hpkg ...
creating package osl_x86-1.10.13-1-x86_gcc2.hpkg ...

heap size:                  2710435
TOC size:                      4149
package attributes size:       1542
total size:                 2710515
```
Latest: 1.11.10.0

Begasus:
Closing in favour of https://github.com/haikuports/haikuports/pull/5641

PR #4678, opened 2020-02-03
dukpy: mark as untested for 32 bit

extrowerk:
No idea why it fails, probably easy to solve but i have no 32 bit install.


Begasus:
It fails on python 2.7 (as mentioned in the first PR for the recipe), think I mentioned something about adding a check for targetarchtecture so it's enabled for x86_64 but disabled for _x86

PR #4692, opened 2020-02-04
Vim: build fix

extrowerk:


extrowerk:
Tested on 64 bit with gcc8.3, ok.

PR #4707, opened 2020-02-09
DevilutionX: disable debug

### LINKED ISSUE #4703: DevilutionX is a debug build
https://github.com/haikuports/haikuports/blob/11f919c776c4ec6c393880d1b08c7631ebfce809/games-rpg/devilutionx/devilutionx-1.0.0.recipe#L55

This disables all debug features but still produces a debug build, it should either be updated to use `-DBINARY_RELEASE=ON` or add `-DCMAKE_BUILD_TYPE=Release -DLTO=ON`.

Note that `DBINARY_RELEASE` may well fail for Haiku until 1.0.1

extrowerk:
closes #4703


humdingerb:
Thanks. The binary is about an MiB smaller. Maybe also more efficient when running, who knows...

AJenbo:
Defiantly should be much faster, this basically just enables a bunch of optimizations (O3 and LTO) :)

PR #4715, opened 2020-02-11
scriptureguide, fix install path

Begasus:


Begasus:
This was checked on x86_64 and x86_gcc2 (nightly), builds are OK, fails to build on the buildbot :-1:

PR #4723, opened 2020-02-15
libebml, bump version, re-apply patchset

Begasus:


Begasus:
libmatroska needs a revbump after this, no soname changes but fails on unresolved symbol if not rebuild

Issue #1978, opened 2017-12-27
mplayer always crash: Could not map image: Out of memory


Giova84:
I installed, on Haiku gcc2h (hrev51714) the new mplayer (mplayer_x86-1.3.0-5-x86_gcc2.hpkg) package. Whenever I attempt to play a media file, I always get: `runtime_loader: /boot/system/bin/mplayer: Could not map image: Out of memory`.
On this  Desktop PC I have enough memory: 6 GiB of DDR3 RAM.

I attach the debog output:
https://pastebin.com/M7aGa6Qs

Giova84:
For the sake of completeness: here I have `mplayer_x86-1.1.4.8.3-5-x86_gcc2.hpkg` which I built by myself a long time ago, which still works without any issue.

extrowerk:
Does it still happens?

kenmays:
Yes, the issue is still revelant. Tested on hrev52080 x86_gcc2 w/ mplayer_x86-1.3.0-8.

humdingerb:
I see the same issue with mplayer_x86-1.4.0-9-x86_gcc2.hpkg on hrev53872.
It's showing `runtime_loader: /boot/system/bin/mplayer: Could not map image: Out of memory` even when just calling "mplayer --help"...

Does it work on 64bit? Should we mark the 32bit version as broken in the recipe?

extrowerk:
64 bit works for me.

kenmays:
@humdingerb - Yes, verified as broken on x86_gcc2 (mplayer_x86-1.4.0-8) using hrev53874 x86_gcc2.. Crash report included.

[mplayer-747-debug-16-02-2020-16-49-13.txt](https://github.com/haikuports/haikuports/files/4210295/mplayer-747-debug-16-02-2020-16-49-13.txt)

PR #4730, opened 2020-02-17
mplayer: mark as unsupported on 32bit

humdingerb:
* On 32bit running mplayer fails with:
  'runtime_loader: /boot/system/bin/mplayer: Could not map image: Out of memory'
  See #1978.

* Remove "x86", it's not an official target for Haiku anymore.

### DIFF
--- a/media-video/mplayer/mplayer-1.4.0.recipe
+++ b/media-video/mplayer/mplayer-1.4.0.recipe
@@ -22,15 +22,15 @@ DVD subtitles (SPU streams, VOBsub and Closed Captions) are supported as well."
 HOMEPAGE="http://www.mplayerhq.hu/"
 COPYRIGHT="2001-2019 The MPlayer Team"
 LICENSE="GNU LGPL v2.1"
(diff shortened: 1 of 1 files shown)


waddlesplash:
It would be nice to investigate why instead of just disabling it...

humdingerb:
Of course. That's why it says "See #1978" not "Fixes #1978". :)
If nobody is able to investigate/fix it soon though, we should at least remove the package from HaikuDepot.

Begasus:
Could this perhaps help?
```
-cache <kbytes>
    This option specifies how much memory (in kbytes) to use when precaching a file/:URL. Especially useful on slow media (default is -nocache). 
```
From the manual page: http://tivo-mplayer.sourceforge.net/docs/mplayer-man.html

humdingerb:
The cache option does help in a way for very small files - I tested with a 200kB wav. With it mplayer crashes with the debug report linked by kenmays at #1978... :)
Larger files still fail with the runtime_loader error.

PR #4732, opened 2020-02-19
ripgrep: fix x86 secondary arch build

Crestwave:


Begasus:
Good to merge?

Crestwave:
From as much as I can determine, yeah. This just sets a minimum version so it doesn't pull the outdated `rust_x86` instead of `rust_bin_x86`.

Issue #4722, opened 2020-02-15
Mediatomb: upnp error -117


unarix:
Seems to be a problem in the port of mediatomb in the architectur x64. When start the aplication an error message shows:

2020-02-11 01:02:11 INFO: Configuration check succeeded.
2020-02-11 01:02:11 ERROR: main: upnp error -117
2020-02-11 01:02:11 ERROR: upnp_cleanup: UpnpUnRegisterRootDevice failed

Begasus:
Can confirm, checked and working fine on x86_gcc2, error on x86_64

kenmays:
Verified mediatomb on hrev53874 x86_gcc2 laptop from remote Windows client. Works fine on that platform. Snapshot included.
![mediatomb_haiku_server_test](https://user-images.githubusercontent.com/5168719/74609470-0a3ccb80-509f-11ea-8efd-0aac146b1274.jpg)

On R1B1 x86_64, seems mediatomb config is good. Seems network UDP port setup is 'blocked' versus setup on x86_gcc2. Error reflects a network config is not correct for proper setup. Needs review.

Begasus:
Same error when trying to launch gerbera (on 32bit and 64bit)
```
~> gerbera 
2020-02-22 18:42:35    INFO: Gerbera UPnP Server version 1.4.0 - http://gerbera.io/
2020-02-22 18:42:36    INFO: ===============================================================================
2020-02-22 18:42:36    INFO: Gerbera is free software, covered by the GNU General Public License version 2
2020-02-22 18:42:36    INFO: Copyright 2016-2019 Gerbera Contributors.
2020-02-22 18:42:36    INFO: Gerbera is based on MediaTomb: Copyright 2005-2010 Gena Batsyan, Sergey Bostandzhyan, Leonhard Wimmer.
2020-02-22 18:42:36    INFO: ===============================================================================
2020-02-22 18:42:36    INFO: Loading configuration from: /boot/home/config/settings/gerbera/config.xml
2020-02-22 18:42:36    INFO: Checking configuration...
2020-02-22 18:42:36    INFO: Setting filesystem import charset to UTF-8
2020-02-22 18:42:36    INFO: Setting metadata import charset to UTF-8
2020-02-22 18:42:36    INFO: Setting playlist charset to UTF-8
2020-02-22 18:42:36    INFO: Configuration check succeeded.
2020-02-22 18:42:36    INFO: About to migrate metadata from mt_cds_object to mt_metadata
[...]

PR #4734, opened 2020-02-19
gerbera, new recipe

Begasus:
Needs someone to fix the path to search for the config.xml file (or a way to find to cp the config.xml file to ~/config/settings/gerbera)


extrowerk:
Postinstallscript?

Begasus:
> Postinstallscript?

Had been thinking about it, need to give it a closer look :+1:

Begasus:
I think it has something to do with it here?
https://github.com/gerbera/gerbera/blob/0f1083e1dd4f390a548c3e4fdc5dc29dc159b4f6/src/main.cc#L223

Begasus:
Almost there, need to parch CMakeList.txt to check for "network" instead of "socket" (same for "nsl")

Begasus:
Think I got as far as I can here, reported on the issue for mediatomb: https://github.com/haikuports/haikuports/issues/4722

PR #4735, opened 2020-02-20
supertux: bump to 0.6.1 and fix build

### LINKED ISSUE #4456: SuperTux: moves files everything from $HOME to ~/config/settings/supertux2
Possibly it happens somewhere here https://github.com/SuperTux/supertux/blob/master/src/supertux/main.cpp#L267

Crestwave:
Resolves #4456 

Apparently it tries to move an old config directory, constructed like `~/.$PACKAGE_NAME`, to the new one, and `PACKAGE_NAME` wasn't defined for some reason... resolving in `~/.`. Yikes! I just disabled the entire thing because I don't think we ever had the old config directory, anyway.


diversys:
Cool, thanks!

PR #4736, opened 2020-02-20
jumpnbump: new recipe

Crestwave:
Some notes:

- I couldn't figure out what copyright statement to use, so I just put a generic one with the dates.
- I wasn't sure whether to package it with the menu or not, so I just decided not to because it's the default.
- ~Networking support doesn't seem to work, not even open a port, but I'm not sure if that's because I'm using it in a VM or something. Would be great if someone could test this on real hardware.~

EDIT: The commands I used didn't work on Linux, either. I guess I'm doing something wrong

EDIT 2: Okay, I tried a different set of commands it works perfectly.

### DIFF
--- a/games-arcade/jumpnbump/jumpnbump-1.61.recipe
+++ b/games-arcade/jumpnbump/jumpnbump-1.61.recipe
@@ -1,0 +1,69 @@
+SUMMARY="Play cute bunnies jumping on each other's heads"
+DESCRIPTION="This is a game for the whole family. You play cute fluffy little \
+bunnies and hop on each other's heads.
+
+At the beginning you are in the menu, where you have to let each active \
+player juimp over the tree trunk to enter the play area, and then walk to the \
+right. You will then enter the arena. The aim is to jump on the other bunnies' \
+heads...
+
+Jump 'n Bump was originally a DOS game by Brainchild Design, which was open \
+sourced under the GPL license and ported to SDL, and then SDL 2."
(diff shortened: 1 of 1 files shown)


extrowerk:
Didnt we had recipe for this already?

games-arcade/jumpnbump/jumpnbump-1.61.recipe
```diff
@@ -0,0 +1,69 @@
...
+{
+	make $jobArgs \
+		PREFIX="$prefix" \
+		BINDIR="$commandBinDir" \
+		DATADIR="$dataDir"
+}
+
+INSTALL()
+{
+	make PREFIX="$prefix" BINDIR="$commandBinDir" DATADIR="$dataDir" install
```
Begasus: Maybe?
```diff
+	rm -rf $dataDir/{applications,icons}
```

Begasus:
Seems to work OK, although I didn't fully investigate, launches, playable :)

games-arcade/jumpnbump/jumpnbump-1.61.recipe
Crestwave: Yeah, I wasn't sure about that. Done (didn't use brace expansion as it's a Bashism and fbrosson wanted me to avoid them when they were here).

Begasus: Was the first solution that came to mind ;)

Begasus:
Checked on 64bit?

PR #4737, opened 2020-02-20
dopewars: new recipe

Crestwave:
Networking support only works on x86_gcc2. The UI looks kinda weird for some reason, but I'm not sure if that's just because of the terminal.

### DIFF
--- a/games-strategy/dopewars/dopewars-1.5.12.recipe
+++ b/games-strategy/dopewars/dopewars-1.5.12.recipe
@@ -1,0 +1,58 @@
+SUMMARY="Make a fortune dealing drugs on the streets of New York"
+DESCRIPTION="dopewars is a free Unix/Win32 rewrite of a game originally based \
+on \"Drug Wars\" by John E. Dell. The idea of dopewars is to deal in drugs on \
+the streets of New York, amassing a huge fortune and paying off the loan \
+shark, while avoiding the ever-annoying police. The Unix/Win32 rewrite, as \
+featuring a so-called \"antique\" mode which closely follows the original, \
+introduces new features such as the ability to take part in multi-player \
+games. dopewars aims to be highly configurable, and what you can't change in \
+the configuration files you can change by poking around in the source, which \
+is freely available under the terms of the GNU General Public License.
+
+ADVISORY: dopewars is not recommended for children below 16 without parental \
+supervision. We also do not condone real-life drug dealing."
+HOMEPAGE="https://dopewars.sourceforge.io/"
+COPYRIGHT="1999-2005 dopewars developers"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/dopewars/dopewars-$portVersion.tar.gz"
+CHECKSUM_SHA256="23059dcdea96c6072b148ee21d76237ef3535e5be90b3b2d8239d150feee0c19"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+GLOBAL_WRITABLE_FILES="
+	var/dopewars.sco keep-old
+	"
+
+PROVIDES="
+	dopewars$secondaryArchSuffix = $portVersion
+	cmd:dopewars$secondaryArchSuffix
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libglib$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libglib$secondaryArchSuffix
+	devel:libncurses$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86_64
+	sed -i '/dopewars_DEPENDENCIES/ s/-lintl -liconv//g' src/Makefile
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+}
--- a/games-strategy/dopewars/dopewars-1.5.12.recipe
+++ b/games-strategy/dopewars/dopewars-1.5.12.recipe
@@ -47,7 +47,7 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86_64
+	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86
 	make $jobArgs
 }
 
--- a/games-strategy/dopewars/dopewars-1.5.12.recipe
+++ b/games-strategy/dopewars/dopewars-1.5.12.recipe
@@ -59,8 +59,8 @@ PATCH()
 
 BUILD()
 {
+	export LIBS="-lintl -liconv"
 	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86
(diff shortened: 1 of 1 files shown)


Begasus:
Network checks for -lsocket (afaikt), maybe patch configure for this (haven't looked into it)
Also see it can use libSDL/libSDL_mixer, did you try that?

Begasus:
After enabling libSDL/libSDL_mixer/libintl/libiconv cmd's awk/bison/pkg-config (exporting lib -lintl and -lnetwork before runConfigure) I'm getting this:
dopewars
Trying to connect to a server doesn't seem to work (either with -lsocket nor -lnetwork)

Crestwave:
That seems to be the same result as my recipe. I can force enable networking, but it fails to compile, except on x86_gcc2, where it automatically enables it anyway. My understanding with SDL is that it's for the GUI, and that the GUI also requires GTK for some reason, but I could be wrong.

Begasus:
One thing I noticed is that I got a sound when quitting the game (with SDL_mixer) :)
But still looking into it, I'm guessing there are no servers online (anymore?), so maybe try to start your own server and see if you can get a notice that's it been found?

games-strategy/dopewars/dopewars-1.5.12.recipe
```diff
@@ -0,0 +1,57 @@
...
+supervision. We also do not condone real-life drug dealing."
+HOMEPAGE="https://dopewars.sourceforge.io/"
+COPYRIGHT="1999-2005 dopewars developers"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/dopewars/dopewars-$portVersion.tar.gz"
+CHECKSUM_SHA256="23059dcdea96c6072b148ee21d76237ef3535e5be90b3b2d8239d150feee0c19"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
Begasus: As this builds ok for gcc2 mayb remove secondary architecture support?

Begasus:
Didn't fully investigate, but you could patch configure to check for "network" instead of "socket" for "gethostbyname" it should fix detection for network connection on 64bit also (didn't check that yet on 32bit)

Crestwave:
> One thing I noticed is that I got a sound when quitting the game (with SDL_mixer) :)

Oh, I see. Sorry; my desktop speakers are broken. I tried it on my laptop and... that's quite a sound. I thought it would just be some clicking sound or something.

> But still looking into it, I'm guessing there are no servers online (anymore?), so maybe try to start your own server and see if you can get a notice that's it been found?

Oh, I thought you didn't make it to the connection page either. Yeah, that's what I did for gcc2.

> Didn't fully investigate, but you could patch configure to check for "network" instead of "socket" for "gethostbyname" it should fix detection for network connection on 64bit also (didn't check that yet on 32bit)

Would a patch file be preferred over `sed` because it likely won't be updated, anyway?

games-strategy/dopewars/dopewars-1.5.12.recipe
Crestwave: Ah, I thought we built for as much architecture as possible for some reason; probably becasue barely anything works on x86_gcc2 so I never noticed.

Begasus:
> Would a patch file be preferred over `sed` because it likely won't be updated, anyway?

I'm guess using sed would be enough here as you mention :)

games-strategy/dopewars/dopewars-1.5.12.recipe
Begasus: There has already been an issue created for things like this :)
https://github.com/haikuports/haikuports/pull/4737#issuecomment-589887632

games-strategy/dopewars/dopewars-1.5.12.recipe
```diff
@@ -0,0 +1,65 @@
...
+PROVIDES="
+	dopewars = $portVersion
+	cmd:dopewars
+	"
+REQUIRES="
+	haiku
+	lib:libglib
+	lib:libncurses
+	lib:libSDL
+	lib:libSDL_mixer
```
Begasus: Could you update these with the correct naming? See https://github.com/haikuports/haikuports/issues/3878

Begasus:
There are some locale files included, maybe add libiconv and libintl to the recipe?

games-strategy/dopewars/dopewars-1.5.12.recipe
```diff
@@ -0,0 +1,65 @@
...
+
+BUILD()
+{
+	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
```
Begasus: Some dirs that could be removed after install $dataDir/{pixmaps,gnome} and $libDir (contains charset.alias)

Crestwave: Oh, I didn't notice those. I'm guessing I should move `$dataDir/doc/dopewars-$portVersion` to `$docDir` as well. Should I remove the `LICENCE` file and such from there or just leave it alone?

games-strategy/dopewars/dopewars-1.5.12.recipe
Crestwave: Oh, I *was* wondering what to name it since I appended `_2.0` for the SDL2 libs. Thanks!

Crestwave:
> There are some locale files included, maybe add libiconv and libintl to the recipe?

Ah, I see. I was unaware that those were for locale stuff and disregarded them, as I mentioned in one of the commits. My Makefile patch makes a return!

games-strategy/dopewars/dopewars-1.5.12.recipe
```diff
@@ -0,0 +1,73 @@
...
+
+PATCH()
+{
+	sed -i '/ac_lib/ s/socket/network/g' configure
+}
+
+BUILD()
+{
+	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86
+	sed -i '/dopewars_DEPENDENCIES/ s/-lintl -liconv//' src/Makefile
```
Begasus: The sed command doesn't really add anything here (no changes to the Makefile that shows when linking dopewars)
using export * pulls in -lintl and -liconv at linking time
```diff
+	export LIBS="-lintl -liconv"
	runConfigure --omit-dirs "dataRootDir docDir" ./configure --build=x86
-	sed -i '/dopewars_DEPENDENCIES/ s/-lintl -liconv//' src/Makefile
```

games-strategy/dopewars/dopewars-1.5.12.recipe
Begasus: Nice catch, missed that one :)

games-strategy/dopewars/dopewars-1.5.12.recipe
Crestwave: I mean, it's to prevent it from failing because it lists `-lintl` and `-liconv` as file dependencies that need to be made. But for some reason it doesn't happen when they're exported so I guess that's cleaner. Or maybe I misunderstood something?

Begasus: running configure without the options also fill in the needed dependencies, but I've seen it before that INTLLIBS=-lintl doesn't get pulled in at linking time, hence using LIBS=* works

Crestwave: To clarify, the `sed` command was because `-lintl` and `-liconv` were listed in `dopewars_DEPENDENCIES`, which is for files that need to be made. So obviously it fails, saying "No rule to make target '-lintl', needed by 'dopewars'". But it doesn't get listed there when they're in `LIBS`, so it's not needed with that, but it wasn't useless earlier! :)

PR #4742, opened 2020-02-23
puckman: new recipe

Crestwave:
This is a random game I found on FreeBSD ports, but I find it surprisingly high-quality. There's no sound, but my desktop's speakers are broken, anyway. ;) The secondary arch build works fine except that you stop when you reach the top four-way intersection; I'll look into it tomorrow as it's getting late here.


Begasus:
Looks good to me, I'll let you investigate what you encountered before we merge :)

Crestwave:
I don't see any obvious indicators of what's wrong, so maybe we should just merge it with the secondary arch build disabled.

Begasus:
Didn't really encounter a problem for the _x86 build, but if it's a nogo for you we could merge it as is, maybe create a ticket about the problem you encountered with some explanation and/or maybe a screenshot showing the problem?

Crestwave:
Were you able to complete a level?

Begasus:
> Were you able to complete a level?

Didn't get that far ;) (but didn't stick to check) :)

Crestwave:
Well, the problem I experienced was getting stuck in the four-way intersections on the top. If you were able to pass through there, let me know.

Begasus:
At the first level (heading up/left/right)? If so, I'll check again later

Crestwave:
Yes, at the first level. Just go straight then head up when you hit the wall.

PR #4747, opened 2020-02-24
openmortal: new recipe

Crestwave:
Notes:

- I didn't know anything good I could set the host to so I left it blank.
- I patched the Perl scripts to use relative paths for `require`; is there a better way to handle this?
- It fails to build in strict mode on x86 secondary architecture because it complains that it `can't find used library "libperl.so"`, but it seems to work fine otherwise. I'm guessing this is because it wants a secondary arch library but Perl doesn't have a secondary arch build enabled for some reason; any ideas on what do to here?
- I didn't split the package data as I felt that the difference isn't too much nowadays, but let me know if you feel otherwise.
- How much space should debug info take to warrant a debuginfo package? The debuginfo package takes 1.2 MB on my system, which I'm guessing is enough.

### DIFF
--- a/games-arcade/openmortal/openmortal-0.7.recipe
+++ b/games-arcade/openmortal/openmortal-0.7.recipe
@@ -1,0 +1,69 @@
+SUMMARY="A parody of the once popular coin-up fighting game, Mortal Kombat"
+DESCRIPTION="Mortál Szombat is a parody of the popular coin-up game, Mortal \
+Kombat. It is currently playable (maybe even enjoyable), although it is still \
+under development. Only two-player game is supported, single-player games \
+against computer opponent is not planned yet. There are currently 15 available \
+characters, and several more in the making!"
+HOMEPAGE="https://openmortal.sourceforge.net/"
+COPYRIGHT="2003 the OpenMortal Team"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/openmortal/openmortal-$portVersion.tar.bz2"
+CHECKSUM_SHA256="9c302560d71ddcd3473ba2a1022fd118cbce87a4cd4b3080cb46288a410bb710"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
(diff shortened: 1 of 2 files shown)


games-arcade/openmortal/openmortal-0.7.recipe
```diff
@@ -0,0 +1,73 @@
...
+	devel:libSDL_gfx$secondaryArchSuffix
+	devel:libSDL_image$secondaryArchSuffix
+	devel:libSDL_mixer$secondaryArchSuffix
+	devel:libSDL_net$secondaryArchSuffix
+	devel:libSDL_ttf$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:perl
```
korli: please add cmd:sed

Crestwave: done

korli:
about libperl.so is there a rpath on the binary?

Crestwave:
Yeah

games-arcade/openmortal/openmortal-0.7.recipe
```diff
@@ -0,0 +1,74 @@
...
+{
+	sed -i 's/CChallengeMenu::CChallengeMenu/CChallengeMenu/g' src/OnlineChat.cpp
+	sed -i "s|require '\([A-Z].*\)'|require './\1'|g" data/script/*
+	sed -i "s|require 'DataHelper.pl'|require '../script/DataHelper.pl'|g" data/characters/*.pl
+}
+
+BUILD()
+{
+	runConfigure --omit-dirs "docDir dataRootDir" \
+		./configure --build=x86 --host=
```
korli: is --build=x86 valid for x86_64?

Crestwave: Yep. --build=x86_64 is also valid for x86 but I felt that was more appropriate. Doesn't seem to do much, as with the host, which works fine with it empty.

PR #4752, opened 2020-02-26
starfighter: new recipe

Crestwave:
Should the `x86` architecture always be marked as untested?

EDIT: Also, I wasn't sure whether to put this in games-action or games-arcade


korli:
it would be nice to convert the patch to a git patch. It could eventually upstreamed with your author name.

Crestwave:
I'll submit a PR upstream when this gets merged.

korli:
games-arcade is correct.

Crestwave:
Upstreamed: https://github.com/pr-starfighter/starfighter/pull/4

PR #4753, opened 2020-02-26
yaml_cpp: fix relative include and lib directory

Crestwave:
This is used in the installed cmake files, making it error from trying to use `/boot/system/include` when using it in cmake without this patch.


Crestwave:
Whoops, didn't read the rest of the recipe. Same thing happens for the library on secondary arch.

Begasus:
Could you check if you could patch CMakeLists.txt to use CMAKE_INSTALL_INCLUDEDIR and upstream it (instead of using a hardcoded "include" directory)?

Crestwave:
So should I still try to patch it when there's no need to upstream it?

PR #4755, opened 2020-02-27
supermariowar: new recipe

Crestwave:
Notes:
- Depends on #4753 
- I was going to put this in `games-arcade`, but Gentoo had it in `games-action`
- I'm not sure if the unzipping should be done in another function
- Am I supposed to put the apps in a directory in `$appsDir`?
- I wasn't sure what to name the version; would it be preferred to make it a `~git` package?
- I create shell scripts to run them because it's designed to have the data directory path passed as an argument, but because of that it uses the icon for text files and has a different name on startup. Does this mean I have to patch the program?
- Should I enable `USE_PNG_SAVE`?
- I was also able to run the original version (this is an unofficial fork); would a recipe for it be accepted if I feel like making one in the future?

### DIFF
--- a/games-action/supermariowar/supermariowar-beta1.1.recipe
+++ b/games-action/supermariowar/supermariowar-beta1.1.recipe
@@ -55,10 +55,10 @@ BUILD_PREREQUIRES="
 	"
 
 defineDebugInfoPackage supermariowar$secondaryArchSuffix \
-	"$appsDir/smw" \
-	"$appsDir/smw-leveledit" \
+	"$appsDir/SMW" \
+	"$appsDir/SMW-LevelEdit" \
 	"$commandBinDir/smw-server" \
-	"$appsDir/smw-worldedit"
+	"$appsDir/SMW-WorldEdit"
 
 PATCH()
 {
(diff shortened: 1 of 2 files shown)


games-action/supermariowar/supermariowar-beta1.1.recipe
```diff
@@ -0,0 +1,112 @@
...
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	install -m 755 -d "$appsDir" "$commandBinDir"
+	install -m 755 Binaries/Release/smw-server "$commandBinDir"
+	set smw SMW "Super Mario War" \
```
extrowerk: What does this do?

games-action/supermariowar/supermariowar-beta1.1.recipe
```diff
@@ -0,0 +1,112 @@
...
+		smw-leveledit SMW-LevelEdit "Super Mario War Level Editor" \
+		smw-worldedit SMW-WorldEdit "Super Mario War World Editor"
+
+	while [ $# != 0 ]; do
```
extrowerk: And this one?

games-action/supermariowar/supermariowar-beta1.1.recipe
Crestwave: The `set` command sets its arguments to positional parameters. So the first argument would be `$1`, the second `$2`, etc.

games-action/supermariowar/supermariowar-beta1.1.recipe
Crestwave: `$#` is the number of positional parameters. Basically, every loop it uses a batch of three parameters, then it `shift`s it away to the next batch until there are none left.

Crestwave:
I went ahead and patched the source instead of using the shell script launchers, simplifying the recipe a lot, but I still use the looping you commented on.

games-action/supermariowar/supermariowar-beta1.1.recipe
extrowerk: Aha! Thanks for the explanation!

PR #3445, opened 2018-12-05
New fork to add the wormux game recipe to the repo

dominikstrak2001:
This is recipe to build the Wormux game on 64-bit Haiku.

PR #4764, opened 2020-02-28
warmux: new recipe

Crestwave:


Begasus:
I think it's called "wormux" :)
PS, did you look into an earlier attempt on porting this? https://github.com/haikuports/haikuports/pull/3445

Crestwave:
> I think it's called "wormux" :)
It was renamed to Warmux to avoid trademark issues with Worms. 
> PS, did you look into an earlier attempt on porting this? #3445 
No; I actually tried to find any other attempts or comments on it, but didn't check on the closed pull requests.

Crestwave:
Regarding the reason for closing the last pull request, we bundle lots of stuff that are no longer maintained, so I don't think that's really much of an issue. One of the reasons I wanted to port these to Haiku is so they could be preserved; did I just waste my time? It's not like Hedgewars is a fork or anything AFAIK, and GHC also doesn't run on x86_64 yet as far as I'm aware of.

At least I got my answer to where I should put the fonts, which I asked in IRC earlier.

Begasus:
You didn't waste your time, and as explained on the closed (other PR) this one isn't maintained any more, but hedgewars doesn't build (last time I tried), so having this is about as good as we get for now :)
Didn't check the build here (maybe I'll get around to it tomorrow), but if there are things from the other (closed) PR you could use, maybe it's good to investigate on that before we merge

games-strategy/warmux/warmux-11.0.4~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+PROVIDES="
+	warmux$secondaryArchSuffix = $portVersion
+	app:Warmux = $portVersion
+	cmd:warmux_list_games
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libfribidi$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
```
Begasus:
```diff
- 	lib:libpng$secondaryArchSuffix
+ 	lib:libpng16$secondaryArchSuffix
```

games-strategy/warmux/warmux-11.0.4~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libintl$secondaryArchSuffix
+	devel:libfribidi$secondaryArchSuffix
+	devel:libSDL$secondaryArchSuffix
+	devel:libSDL_gfx$secondaryArchSuffix
+	devel:libSDL_image$secondaryArchSuffix
+	devel:libSDL_net$secondaryArchSuffix
+	devel:libSDL_mixer$secondaryArchSuffix
+	devel:libSDL_ttf$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
```
Begasus: libpng16 also here, also change the references for libSDL*
Sort alphabetically

games-strategy/warmux/warmux-11.0.4~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+	"
+
+defineDebugInfoPackage warmux$secondaryArchSuffix \
+	"$appsDir/Warmux" \
+	"$commandBinDir/warmux-list-games"
+
+BUILD()
+{
+	export LIBS="-lnetwork"
+	runConfigure ./configure \
```
Begasus:
```diff
- 	runConfigure ./configure \
+ 	runConfigure --omit-dirs binDir ./configure \
```

games-strategy/warmux/warmux-11.0.4~git.recipe
Crestwave: So I guess it's final, then? I was holding out on applying it to my current PRs because of you asking korli and pulkomandy. Done

Begasus: For libpng there is also a libpng.so file in the devel package, IIRC there were some comments to use devel:libpng16 also, so we can apply this for libSDL* in devel:* too imho

pulkomandy: There is a libpng in the package so you can link with -lpng instead of -lpng16. But it is only a symlink to libpng16, and the soname is libpng16, which is all that matters here. Eventually devel:libpng will be removed everywhere and then the libpng package can stop "providing" it altogether (the lib will still be there but there will be no matching PROVIDE entry in the package metadata).

Begasus:
LGTM

PR #4765, opened 2020-02-29
mgba: bump to 0.8.1 and fix config directory

Crestwave:
I'm guessing that this was overlooked because it tries to use `XDG_CONFIG_HOME` by default, which is set in the terminal, but not when launched from Deskbar.

Note that I removed `libswresample` as it seems that it didn't seem to link against it. It optionally uses libavresample, but it isn't available in the current version of ffmpeg. Let me know if libavresample works on the build bots or if libswresample does something.


waddlesplash:
Please submit the patch upstream, I've submitted Haiku patches before and they were accepted.

Crestwave:
Yes, I always submit patches like this upstream after the recipe PR is merged.

Crestwave:
Merged into upstream: https://github.com/mgba-emu/mgba/pull/1682

PR #4768, opened 2020-03-03
blobby: new recipe

Crestwave:
There's actually a [GitHub repo](https://github.com/danielknobe/blobbyvolley2) (which I very conveniently found right before making this pull request...) with recent commits, but no release; not sure if I should make it a `~git` package.

### DIFF
--- a/games-arcade/blobby/blobby-1.0.recipe
+++ b/games-arcade/blobby/blobby-1.0.recipe
@@ -1,0 +1,74 @@
+SUMMARY="The head-to-head multiplayer ball game"
+DESCRIPTION="Official continuation of the famous Blobby Volley 1.x arcade \
+game.
+
+The \"Dedicated Server\" runs with a Gamespeed of 100%, which means 75 FPS
+The Port for the Server is 1234."
+HOMEPAGE="http://blobby.sourceforge.net/"
+COPYRIGHT="2006 Daniel Knobe
+	2006 Jonathan Sieber"
+LICENSE="GNU GPL v2"
+REVISION="1"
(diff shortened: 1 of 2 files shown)


Begasus:
As they seem to have moved to github maybe try to start from there and rename the recipe to "blobby2-1.1~git.recipe"?
It seems to be active and maybe you could upstream some of the patches?

Crestwave:
Yeah, I was just a bit hesitant to switch to git as I literally found the repo right before clicking open a pull request and my patches don't apply there. :P

Begasus:
> Yeah, I was just a bit hesitant to switch to git as I literally found the repo right before clicking open a pull request and my patches don't apply there. :P

Maybe not all (some may be fixed by now), but I think some Haiku specific changes could be upstreamed? ;)

Begasus:
PS, thanks for your contributions! I think some could learn some things from the patches you are making here! :+1:

Crestwave:
> > Yeah, I was just a bit hesitant to switch to git as I literally found the repo right before clicking open a pull request and my patches don't apply there. :P
> 
> Maybe not all (some may be fixed by now), but I think some Haiku specific changes could be upstreamed? ;)

Yeah, don't worry, I always upstream patches in these situations. :)

> PS, thanks for your contributions! I think some could learn some things from the patches you are making here! 👍

Np, I also learned some things making these contributions! `FindDirectory.h` is more convenient than I thought.

Crestwave:
Decided to go with the GitHub name (blobbyvolley2) and opened a new PR: https://github.com/haikuports/haikuports/pull/4769. I'm going to bed soon so I haven't tested it too thoroughly, but it seems to work fine. Only thing I noticed is that there's a translation error message but I didn't see any effects of it and it *seems* like the problem is from upstream.

PR #4780, opened 2020-03-08
neofetch: bump to 7.0.0

Crestwave:


app-misc/neofetch/neofetch-7.0.0.recipe
```diff
@@ -34,5 +33,5 @@ BUILD_PREREQUIRES="
 
 INSTALL()
 {
-	make install PREFIX=$prefix MANDIR=$manDir
+	make install PREFIX="$prefix" MANDIR="$manDir"
```
Begasus: Any specific reason for the quotes?

Crestwave: It prevents the expansion from being subject to word splitting and globbing. It's not necessary right now, as the paths don't have any whitespace or patterns, but it's good practice to quote them if you don't intend for it to expand.

The author of Neofetch takes shell safety very seriously, so I'm sure that he'll like that the recipe for it uses proper quoting. Part of the reason he created his own independent Linux distribution is because all the others use "unsafe" shell scripts!

Crestwave: Also, while unsafe things like this usually won't cause any problems on Linux distributions, Haiku has always had a philosophy to not hardcode paths so that they can be changed without breaking anything. If the path is ever changed to have a space, pretty much all recipes with unquoted paths will break.

PR #4784, opened 2020-03-09
Enable Turkish localisation

bitigchi:
This proposal adds Turkish language localisation to LO Haiku port.


diversys:
Please adjust commit message to read "LibreOffice: enable Turkish localisation"
Also bump revision so new version of package could be built.

PR #4795, opened 2020-03-15
Match SDL BUILD_REQUIRES to REQUIRES

Crestwave:
Also sneaked in a fix to `jumpnbump`'s `SOURCE_URI`; I forgot to replace the placeholder one I used.


Begasus:
Maybe bump REVISION for them? (keeps track of the history then)

Crestwave:
Oh, I didn't think this would change the output package. Done

Begasus:
> Oh, I didn't think this would change the output package. Done

I doesn't, but it could be helpful to others referring to the same issue

Begasus:
Ready? :)

Crestwave:
To merge? Sure, I guess, if it's finalized that this is better

Begasus:
Thanks! :+1:

PR #4796, opened 2020-03-15
ant_core: bump version

panos:


AlwaysLivid:
I added the `(WIP)` prefix to the PR due to the following error, brought to my attention by @Begasus.

```
waiting for build package ant_core-1.10.7-1 to be activated
Building ...
Warning: JAVA_HOME environment variable not set.
  If build fails because sun.* classes could not be found
  you will need to set the JAVA_HOME environment variable
  to the installation directory of java.
Error: JAVA_HOME is not defined correctly.
  We cannot execute java
Bootstrap FAILED
```

dev-java/ant_core/ant_core-1.10.7.recipe
```diff
@@ -20,9 +20,9 @@ The Apache Ant project is part of the Apache Software Foundation.
 HOMEPAGE="http://ant.apache.org/"
 COPYRIGHT="1999-2018 The Apache Software Foundation"
 LICENSE="Apache v2"
-REVISION="1"
+REVISION="2"
```
Begasus: No need to bump revision here (it's not merged yet) :)

AlwaysLivid: Yeah, I just fixed it.

Begasus:
LGTM

Begasus:
Checking in Terminal still fails ...
```
~/haikuports_games> ant -version
Error: JAVA_HOME is not defined correctly.
  We cannot execute java
```

dev-java/ant_core/ant_core-1.10.7.recipe
```diff
@@ -54,7 +54,7 @@ BUILD()
 {
 	source /system/data/profile.d/openjdk8.sh
 	export PATH=$JDK8_HOME/bin:$PATH
-	ANT_HOME= ./build.sh
+	ANT_HOME=./build.sh
```
Crestwave: This line seems nonsensical. It basically makes the `BUILD` phase do nothing; the compilation then happens on the `INSTALL` phase.

Crestwave: This line seems nonsensical. It basically makes the `BUILD` phase do nothing; the compilation then happens on the `INSTALL` phase. Was this a typo?

Crestwave:
COPYRIGHT should also be updated to 1999-2019. Beyond that, the recipe itself could be improved in a number of ways, but I'm not sure if that's within the scope of this PR. For example, the quotes in DESCRIPTION should be escaped, HOMEPAGE and SOURCE_URI can use https, SOURCE_FILENAME could be defined, `$relativeSettingsDir` could be used in GLOBAL_WRITABLE_FILES, BUILD_REQUIRES could be removed (not sure about this one), a non-hardcoded path could be sourced, and a Java >= 8 could be possibly used for building.

dev-java/ant_core/ant_core-1.10.7.recipe
korli: With this change, build.sh isn't executed anymore.

Crestwave: In the `BUILD` phase at least; what I want to know is why? This also isn't a particularly clean way to disabling it if that was intended.

AlwaysLivid: > With this change, build.sh isn't executed anymore.

Oh, that could explain why it worked, it was confusing for me as well. For some reason, the recipe was broken, due to JAVA_HOME not being defined. I will remove that line for the time being.

Crestwave: Yes, but I'm fairly sure that we have a policy to keep building in BUILD. You could just define JAVA_HOME in BUILD as well.

AlwaysLivid:
> COPYRIGHT should also be updated to 1999-2019. Beyond that, the recipe itself could be improved in a number of ways, but I'm not sure if that's within the scope of this PR. For example, the quotes in DESCRIPTION should be escaped, HOMEPAGE and SOURCE_URI can use https, SOURCE_FILENAME could be defined, `$relativeSettingsDir` could be used in GLOBAL_WRITABLE_FILES, BUILD_REQUIRES could be removed (not sure about this one), a non-hardcoded path could be sourced, and a Java >= 8 could be possibly used for building.

I mean, a little makeover wouldn't really hurt. The only problem here is with `Java >= 8`. Should a version higher than 8 be used, then the variable `$JDK8_HOME` would not be valid in all contexts, thus breaking the recipe in all scenarios where `Java != 8`.

dev-java/ant_core/ant_core-1.10.7.recipe
AlwaysLivid: Hm, the problem is that the `./build.sh` line also seems to be the culprit for installing the program itself.

I will move the line there for the time being and make the appropriate changes later.

Crestwave:
> I mean, a little makeover wouldn't really hurt. The only problem here is with `Java >= 8`. Should a version higher than 8 be used, then the variable `$JDK8_HOME` would not be valid in all contexts, thus breaking the recipe in all scenarios where `Java != 8`.

Yes, that's the problem I found as well. There is an `openjdk.sh` provided by the _default packages that defines JAVA_HOME, but for some reason requiring something like `cmd:java >= 8` doesn't work. Perhaps someone with more experience in how it resolves dependencies would have a solution to that. You could hack in another solution, but it would probably be better to try to use this.

AlwaysLivid:
> > I mean, a little makeover wouldn't really hurt. The only problem here is with `Java >= 8`. Should a version higher than 8 be used, then the variable `$JDK8_HOME` would not be valid in all contexts, thus breaking the recipe in all scenarios where `Java != 8`.
> 
> Yes, that's the problem I found as well. There is an `openjdk.sh` provided by the _default packages that defines JAVA_HOME, but for some reason requiring something like `cmd:java >= 8` doesn't work. Perhaps someone with more experience in how it resolves dependencies would have a solution to that. You could hack in another solution, but it would probably be better to try to use this.

I can give write access to my repository if needed as well, let me push some changes first.

AlwaysLivid:
> There is an `openjdk.sh` provided by the _default packages that defines JAVA_HOME, but for some reason requiring something like `cmd:java >= 8` doesn't work. Perhaps someone with more experience in how it resolves dependencies would have a solution to that. You could hack in another solution, but it would probably be better to try to use this.

A possible hack would be running something like:

```
source /system/data/profile.d/openjdk8.sh
source /system/data/profile.d/openjdk9.sh
source /system/data/profile.d/openjdk10.sh
source /system/data/profile.d/openjdk11.sh

if $JDK11_HOME not empty, then...
  else if $JDK10_HOME not empty, then...
```

Maybe checking whether the shell files exist in the first place in reverse order and then saving the version number to a variable accordingly would also be a good workaround for the time being.

Crestwave:
If you want to go down the hacky route, you could probably do:

```
set /packages/openjdk*/.self/lib/openjdk*
export JAVA_HOME=$1
```

But it might be better to keep it requiring 8 if we can't use the clean solution.

dev-java/ant_core/ant_core-1.10.7.recipe
Crestwave: Hm? I haven't looked much into it, but doesn't seem to install it if you don't pass install-lite and/or ANT_HOME is empty.

AlwaysLivid:
> If you want to go down the hacky route, you could probably do:
> 
> ```
> set /packages/openjdk*/.self/lib/openjdk*
> export JAVA_HOME=$1
> ```
> 
> But it might be better to keep it requiring 8 if we can't use the clean solution.

That would be a good idea, which is what @korli originally did here.

> If you want to go down the hacky route, you could probably do:
> 
> ```
> set /packages/openjdk*/.self/lib/openjdk*
> export JAVA_HOME=$1
> ```
> 
> But it might be better to keep it requiring 8 if we can't use the clean solution.

What would happen if multiple instances of OpenJDK are found?

Crestwave:
> What would happen if multiple instances of OpenJDK are found?

In the hacky solution, it would get the lowest version. However, multiple versions shouldn't be found since we only require once, no?

AlwaysLivid:
> > What would happen if multiple instances of OpenJDK are found?
> 
> In the hacky solution, it would get the lowest version. However, multiple versions shouldn't be found since we only require once, no?

Yeah, you're right about that.

Begasus:
Any progress on this @AlwaysLivid ?

AlwaysLivid:
Haven't done much, not sure how to take care of that problem. @Begasus

Crestwave:
If I remember correctly, this is a problem with haikuporter itself handling `cmd:java >= 8`; I also remember a similar problem with Qt. So it would be great if one of the haikuporter developers took a look at it

korli:
superseded by a868d85792caa0b1ba3a6518331beb6dfd35f07a

PR #4802, opened 2020-03-16
melonds_libretro: bump to 0.8.3

Crestwave:
Also, I'm not sure if `portname-$portVersion-$srcGitRev.fileext` should be used or `portname-$srcGitRev.fileext` for `SOURCE_FILENAME`. Maybe it's appropriate to use the former here since it's an actual version and not a dummy made-up one?

### DIFF
--- a/games-emulation/melonds_libretro/melonds_libretro-0.8.3_20190912.recipe
+++ b/games-emulation/melonds_libretro/melonds_libretro-0.8.3_20190912.recipe
@@ -36,7 +36,7 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	sed -e "s/@DISPLAY_VERSION@/v$portVersion/" \
+	sed -e "s/@DISPLAY_VERSION@/v${portVersion%%_*}-${portVersion##*_}/" \
 		"$portDir"/additional-files/melonds_libretro.info.in \
 		> melonds_libretro.info
 	make $jobArgs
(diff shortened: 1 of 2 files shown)


Begasus:
Why not use the official release? https://github.com/Arisotura/melonDS/releases

Crestwave:
That isn't a libretro core, no?

Crestwave:
If you mean why not package the official release too or something, it uses GTK. I haven't looked too much into it beyond that, but it seems that it also has some Linux-specific code and the latest release fails to build even in OpenBSD/macOS.

Begasus:
> That isn't a libretro core, no?

Didn't check that far ;)

Begasus:
It's probably best to stick to the checkout then :)

kwyxz:
I am super ok with upgrading the version number, even though most libretro cores don't see their version upgraded much (they usually reflect the version when the code was forked into a core, but then don't keep up). If you don't mind, could you keep the Git Revision even if bumping the actual version number? I have scripts that build all the libretro cores for Haiku which use this to find out which ones were updated. Thanks!

Crestwave:
Sorry for the late reply; my GitHub notifications were going into my spam folder. What do you mean by that? The `20180309` part?

kwyxz:
Yep, that. It's what my scripts use to find out if there's been a new version and pull it then generate the new recipe if so.

Begasus:
@kwyxz , since you have been doing the most of the work on the libretro recipes, anything needs to be done here (eg keep the recipe format as it was?)

kwyxz:
@Begasus #4939 should address what the current PR was created for. Once #4939 is merged you can close #4802 thanks!

Crestwave:
I thought I fixed the version with the latest commit?

kwyxz:
Doesn't look like it : it's important that the git commit number is included as a lot (if not the majority) of libretro cores do not follow the versioning of the emulator / game they are forked from. Moreover, having the git commit number allows me to use the (rudimentary but working) build system I wrote, which pulls all the libretro cores I maintain, updates them and commits them all at once.

Crestwave:
I'm confused. I thought that the problem was that the date wasn't included in the filename; I've now renamed it and it's `melonds_libretro-0.8.3_20190912.recipe` and yours is `melonds_libretro-0.8.3_20190911.recipe`. It's off by one probably because of a timezone difference, but that shouldn't matter, right?

kwyxz:
Sorry, I misread : yes the date in the recipe file name is the important one. Regarding srcGitRev used in the source file name, it's no big deal but it makes more sense as we're pulling git revisions, not actual versionned releases.

PR #4804, opened 2020-03-16
bunch of LibreOffice translations are enabled to pack

iam-git:
It is still not everything (about 70 % of all LO translations - as a compromise taken only languages I've at least heard about)


extrowerk:
How much bigger the generated hpkg is? If the difference significant would it be possible to split it up to subpackages?

iam-git:
That is already done by diversys. Every langpack is in separate package. So, recipe now produce 76 packages instead of 17.

diversys:
Could you resolve the conflict?

iam-git:
something like this?

PR #4811, opened 2020-03-21
Package for Guile 1.8.8

mgubi:
I've renamed the package guile1 to avoid conflicts with the already existing package guile which is for version 2.2.4. The two versions are incompatible and some programs still require guile 1.8 to run.

### DIFF
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -116,7 +116,7 @@ INSTALL()
 	make install
 
 	rm -f "$libDir"/charset.alias
-#	rm -f "$libDir"/lib*.la
+	rm -f "$libDir"/lib*.la
 	rm -f "$libDir"/guile/$portVers/extensions/*.la
 	rm -f "$libDir"/guile/$portVers/extensions/*.a
 
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -54,6 +54,9 @@ REQUIRES="
 	lib:libreadline$secondaryArchSuffix
 	lib:libunistring$secondaryArchSuffix
 	"
+CONFLICTS="
+	guile$secondaryArchSuffix
+	"
 
 PROVIDES_devel="
 	guile${secondaryArchSuffix}_devel = $portVersionCompat
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -128,19 +128,19 @@ INSTALL()
 	rm -f "$libDir"/guile/$portVers/extensions/*.la
 	rm -f "$libDir"/guile/$portVers/extensions/*.a
 
-	prepareInstalledDevelLib libguile
-	prepareInstalledDevelLib libguilereadline-v-17
-	prepareInstalledDevelLib libguile-srfi-srfi-1-v-3
-	prepareInstalledDevelLib libguile-srfi-srfi-13-14-v-3
-	prepareInstalledDevelLib libguile-srfi-srfi-4-v-3
-	prepareInstalledDevelLib libguile-srfi-srfi-60-v-2
+	prepareInstalledDevelLibs libguile \
+		libguilereadline-v-17 \
+		libguile-srfi-srfi-1-v-3 \
+		libguile-srfi-srfi-13-14-v-3 \
+		libguile-srfi-srfi-4-v-3 \
+		libguile-srfi-srfi-60-v-2
 
 	fixPkgconfig
 
 	packageEntries tools \
-		"$commandBinDir"/guile\
-		"$commandBinDir"/guile-snarf\
-		"$commandBinDir"/guile-tools\
+		"$commandBinDir"/guile \
+		"$commandBinDir"/guile-snarf \
+		"$commandBinDir"/guile-tools
 
 	packageEntries devel \
 		"$commandBinDir"/guile-config \
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -118,9 +118,9 @@ INSTALL()
 	rm -f "$libDir"/guile/$portVers/extensions/*.a
 
 	prepareInstalledDevelLib libguile
-	prepareInstalledDevelLib libguilereadline-v-17 
-	prepareInstalledDevelLib libguile-srfi-srfi-1-v-3 
-	prepareInstalledDevelLib libguile-srfi-srfi-13-14-v-3 
+	prepareInstalledDevelLib libguilereadline-v-17
+	prepareInstalledDevelLib libguile-srfi-srfi-1-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-13-14-v-3
 	prepareInstalledDevelLib libguile-srfi-srfi-4-v-3
 	prepareInstalledDevelLib libguile-srfi-srfi-60-v-2
 
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -98,7 +98,7 @@ BUILD_PREREQUIRES="
 	cmd:gperf
 	cmd:grep
 	cmd:make
-	cmd:ld
+	cmd:ld$secondaryArchSuffix
 	cmd:pkg_config$secondaryArchSuffix
 	cmd:sed
 	"
@@ -148,7 +148,6 @@ INSTALL()
 		"$commandBinDir"/guile-config \
 		"$dataDir"/aclocal \
 		"$developDir"
-
 }
 
 TEST()
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -130,7 +130,7 @@ INSTALL()
 	fixPkgconfig
 
 	packageEntries devel \
-		"$binDir"/guile-config \
+		"$commandBinDir"/guile-config \
 		"$dataDir"/aclocal \
 		"$developDir"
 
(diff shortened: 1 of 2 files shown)


dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,142 @@
...
+	prepareInstalledDevelLib libguilereadline-v-17
+	prepareInstalledDevelLib libguile-srfi-srfi-1-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-13-14-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-4-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-60-v-2
+
+	fixPkgconfig
+
+	packageEntries devel \
+		"$binDir"/guile-config \
```
Begasus: For secondary architecture you need:
```diff
- 		"$binDir"/guile-config \
+ 		"$commandBinDir"/guile-config \
```

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,142 @@
...
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f "$libDir"/charset.alias
+#	rm -f "$libDir"/lib*.la
```
Begasus: I think you can enable this one :) No need for *.la files in $libDir

mgubi: done.

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,146 @@
...
+	commandBinDir=$prefix/bin
+fi
+
+libguileVersion="17.4.0"
+libguileVers="${libguileVersion%.*}"
+libguileVersionCompat="$libguileVersion compat >= ${libguileVersion%%.*}"
+portVersionCompat="$portVersion compat >= ${portVersion%%.*}"
+portVers="${portVersion%.*}"
+
+PROVIDES="
```
extrowerk: It would be better to split up the main packe. It should only contain the runtime libs, nothing else, and the commands should go into guile_tools or something, so the 2 different runtime could be installed parallel.

extrowerk: You should change the 2.x recipe too to accomplish this.

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,146 @@
...
+	sed --in-place '/HAVE_SYS_MMAN_H/d' config.h
+
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+
+# rename to avoid conflicts with newer guile verions (allow parallel install)
```
extrowerk: Split the binaries into a subpackage instead, please. Change the 2.x recipe accordingly.

dev-scheme/guile/guile1-1.8.8.recipe
mgubi: It seems a viable strategy, better than my hack. Maybe the libraries should go into guile1_lib and the executable into a guile1 package. What do you think? Where I can read about splitting a package?

mgubi: OK. I found the docs. I'm implementing the change with libraries in guile1 and tools in guile1_tools as you suggest (seems the most obvious way, given the docs)

mgubi: > You should change the 2.x recipe too to accomplish this.

There is no need now to split also the guile 2.x recipe. There is no runtime conflict anymore between libraries and is natural that the more recent package provide both libraries and executable. It will be a unnecessary complexification to split all the packages without reason. I would tend to leave the guile 2.x recipe as it is now. Everything works at present.

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,155 @@
...
+	rm -f "$libDir"/lib*.la
+	rm -f "$libDir"/guile/$portVers/extensions/*.la
+	rm -f "$libDir"/guile/$portVers/extensions/*.a
+
+	prepareInstalledDevelLib libguile
+	prepareInstalledDevelLib libguilereadline-v-17
+	prepareInstalledDevelLib libguile-srfi-srfi-1-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-13-14-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-4-v-3
+	prepareInstalledDevelLib libguile-srfi-srfi-60-v-2
```
Begasus: These should be combined with "prepareInstalledDevelLibs"

mgubi: Ok. I'm fixing and checking right now. The splitting of the package into base and tools  as suggested by @extrowerk  solves the conflicts with guile 2.2. So I think I'm done, unless you see some other packaging problems. I've also updated accordingly the PR #4812 for TeXmacs.

Begasus: Hope I'll get some time to check the things, but so far it looks OK

mgubi: No hurry. I just wanted to be sure was on the radar. I would like to announce soon that TeXmacs is officially in Haiku. :)

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,155 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	rm -f "$libDir"/charset.alias
+	rm -f "$libDir"/lib*.la
+	rm -f "$libDir"/guile/$portVers/extensions/*.la
+	rm -f "$libDir"/guile/$portVers/extensions/*.a
```
Begasus: @korli any thoughts on this? I know we don't like the *.la in the packages, but I the *.a files should be OK?

Begasus: Haven't checked ... but maybe the option "--disable-static" could resolve this?

mgubi: > @korli any thoughts on this? I know we don't like the *.la in the packages, but I the *.a files should be OK?

I copied this part from the recipe of Guile 2.2 . Where can I read about Haiku policies for libraries?  
Anyway I'm right now checking if --disable-static works (seem indeed that it does).

mgubi: Even with --disable-static added, the build process produces .la files. As far as I understand .la files are needed for libtool. Are they removed in Haiku?

Begasus: They are removed in Haiku yes, some recipes still have a mention "remove libtool file(s)"

Begasus: https://github.com/haikuports/haikuports/search?q=remove+libtool+file&unscoped_q=remove+libtool+file

korli: Static libraries aren't usually used on Haiku. But for ABI unstable projects they can still make sense.

Begasus: I think you can leave --disable-static out and use the remove of the *.la files, if there is a specific reason you want to remove the *.a files also I think we are good here :)

Begasus: Running a final build, install (with guile 2.2.4 installed) and check TeXmacs then

mgubi: Great! Thanks.

mgubi: Btw, I appreciate very much the carefulness you all put in this. Impressive.

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,157 @@
...
+	lib:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libreadline$secondaryArchSuffix
+	lib:libunistring$secondaryArchSuffix
+	"
+
+PROVIDES_tools="
+	guile1${secondaryArchSuffix}_tools$secondaryArchSuffix = $portVersionCompat
```
Begasus: One to many ... 
```diff
- 	guile1${secondaryArchSuffix}_tools$secondaryArchSuffix = $portVersionCompat
+ 	guile1${secondaryArchSuffix}_tools = $portVersionCompat
```

mgubi: too fast...

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,157 @@
...
+	packageEntries tools \
+		"$commandBinDir"/guile \
+		"$commandBinDir"/guile-snarf \
+		"$commandBinDir"/guile-tools
+
+	packageEntries devel \
+		"$commandBinDir"/guile-config \
+		"$dataDir"/aclocal \
+		"$developDir"
+
```
Begasus: Please remove empty line here

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -0,0 +1,157 @@
...
+	devel:libreadline$secondaryArchSuffix
+	devel:libunistring$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:awk
+	cmd:gcc$secondaryArchSuffix
+	cmd:gperf
+	cmd:grep
+	cmd:make
+	cmd:ld
```
Begasus: Maybe:
```diff
- 	cmd:ld
+ 	cmd:ld$secondaryArchSuffix
```

Begasus:
LGTM, let's merge :) Thanks for your patience :)

mgubi:
Great. Thanks to you for the detailed handling. :)

PR #4812, opened 2020-03-21
Texmacs 1.99.12

mgubi:
This PR provides a package for the GNU TeXmacs scientific editor. (It depends on a previous PR for guile1 needed during build)


Begasus:
I see you included byteswap.h in the patchset, did you check haikuports for it's use (or not being used)?
https://github.com/haikuports/haikuports/search?q=byteswap.h&unscoped_q=byteswap.h
Also I see that there is an official (?) github repository at: https://github.com/texmacs/texmacs (maintained by sadhen, you could use that as a base and try to create a PR to upstream Haiku specific patches?)

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,111 @@
...
+DESCRIPTION="TeXmacs provides a unified and user friendly framework for editing \
+structured documents with different types of content: text, mathematics, graphics, \
+interactive content, slides, etc.\
+Its rendering engine uses high-quality typesetting algorithms so as to produce \
+professionally looking documents, which can either be printed out or presented from \
+a laptop."
+HOMEPAGE="http://www.texmacs.org"
+COPYRIGHT="Free Software Foundation"
+LICENSE="GNU GPL v3"
+
```
extrowerk: Please remove this empty line.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,111 @@
...
+SOURCE_URI="https://github.com/mgubi/texmacs/archive/texmacs-$portVersion-haiku-qt5.tar.gz"
+CHECKSUM_SHA256="607fa4db4676b78780f7af086feec2e2332dc18a371d79fa50d640de6db10b43"
+SOURCE_DIR="texmacs-texmacs-$portVersion-haiku-qt5/src"
+PATCHES="texmacs-$portVersion.patchset"
+
+REVISION="1"
```
extrowerk: Reorder according to the WIKI, please.
https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,111 @@
...
+a laptop."
+HOMEPAGE="http://www.texmacs.org"
+COPYRIGHT="Free Software Foundation"
+LICENSE="GNU GPL v3"
+
+SOURCE_URI="https://github.com/mgubi/texmacs/archive/texmacs-$portVersion-haiku-qt5.tar.gz"
+CHECKSUM_SHA256="607fa4db4676b78780f7af086feec2e2332dc18a371d79fa50d640de6db10b43"
+SOURCE_DIR="texmacs-texmacs-$portVersion-haiku-qt5/src"
+PATCHES="texmacs-$portVersion.patchset"
+
```
extrowerk: No need for empty line here.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,111 @@
...
+REVISION="1"
+
+ADDITIONAL_FILES="texmacs.rdef.in"
+DISABLE_SOURCE_PACKAGE="yes"
```
extrowerk: It is GPL, AFAIK we should generate source package for it.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,111 @@
...
+INSTALL()
+{
+	make install
+
+	rm $binDir/texmacs
+
+	mkdir $appsDir
+	mkdir $appsDir/lib
+	mv $libDir/TeXmacs/bin/texmacs.bin $appsDir/TeXmacs
+	cp /system/lib/libguile.so* $appsDir/lib/.
```
extrowerk: Why would you do that?

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,111 @@
...
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	texmacs$secondaryArchSuffix = $portVersion
+	app:TeXmacs
+	cmd:fig2ps$secondaryArchSuffix
+	"
+
+REQUIRES="
```
extrowerk: Why libguile not listed here and in BUILD_REQs?

app-office/texmacs/texmacs-1.99.12.recipe
extrowerk: You should add ```lib:libguile >= versionnumber``` instead to the REQs.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I was not sure what was the proper way. I will add guile1 (which is my package for guile 1.8, I submitted another PR for that). TeXmacs is not yet compatible with Guile 2.2.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I think mine is the right solution to the following problem: Guile 1.8 and Guile 2.2 cannot be installed at the same time (they both provide cmd:guile and lib:libguile ). So I require guile 1.8 only while building and copy the library with the program. So that the end used do not have to have guile 1.8 installed on the system and can use more modern guile while keeping TeXmacs functioning. It works on my machine.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: Ok, sorry, I will remove this.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: Ok I will change this. But, I followed one of the examples in the wiki, see the advanced example here https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: ok

mgubi: Again: I was following the advanced example here: https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes

mgubi:
> I see you included byteswap.h in the patchset, did you check haikuports for it's use (or not being used)?
> https://github.com/haikuports/haikuports/search?q=byteswap.h&unscoped_q=byteswap.h

Thanks! I will conform with these mode of operation.

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Not sure what the libversion for 2.2 is, but in your PR for guile 1.8 there is a libVersionCompat available: https://github.com/haikuports/haikuports/pull/4811/files#diff-c48bf806855e70b29ab17493d383c123R38

mgubi: This point is not clear to me (and maybe my PR for guile1 is not completely correct): I think the libraries are compatible but the point is that there are differences in the interpretation of the Scheme programs which breaks some programs. This is why I need the older version. Since is not a common problem and likely the end user would never use Guile 1.8 nowadays I was thinking that have a runtime dependence on the older version is not the optimal solution. Also both packages guile and guile1 install cmd:guile, so they have anyway conflicts.

Begasus: Looking at the package for guile 2.2.4 it provides guile-2.2.so, so this shouldn't interfere with the 1.8 version?
1.8 version provides:
```
lib:libguile$secondaryArchSuffix
```
2.2.4 provides:
```
lib:libguile_$portVers$secondaryArchSuffix
```

Begasus: If you would try to link against 2.2.4 you'd have to use "lib:libguild_2.2$secondaryArchSuffix" then ...

Begasus: PS, checking local build (32bit) atm for guile 1.8 and will check this one later :)

mgubi: Still, there is the problem of cmd:guile and cmd:guile-config : both guile1 and guile packages provide it and therefore they will conflict, if I understand this correctly. So the two packages cannot be installed at the same time (and haikuporter tries always to put my guile1 package in obsolete..., which I do not really understand why)

Begasus: I don't think you should have them as a requirement (not tested) if you require the right lib:* and devel:* packages (cmd:guild-config is provided in the devel:* package), will check in a bit ...

mgubi: I entered the IRC channel now if you want to discuss the problem.

extrowerk: Copyiing the lib into this package is definetely not an option. Please use the correct req declaration.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	texmacs$secondaryArchSuffix = $portVersion
+	app:TeXmacs
+	"
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	guile1$secondaryArchSuffix
```
extrowerk: Please do not declare packages, just libs here (except haiku, ofc)

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+	lib:libiconv$secondaryArchSuffix
+	lib:libQt5Concurrent$secondaryArchSuffix
+	lib:libQt5Core$secondaryArchSuffix
+	lib:libQt5Gui$secondaryArchSuffix
+	lib:libQt5Svg$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libguile$secondaryArchSuffix
```
extrowerk: Version number missing.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+	lib:libgmp$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	guile1${secondaryArchSuffix}_devel
```
extrowerk: Please remove this.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+	devel:libiconv$secondaryArchSuffix
+	devel:libQt5Concurrent$secondaryArchSuffix
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libguile$secondaryArchSuffix
```
extrowerk: Version constraint missing.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -25,15 +25,14 @@ PROVIDES="
...
-	guile1$secondaryArchSuffix
 	lib:libiconv$secondaryArchSuffix
 	lib:libQt5Concurrent$secondaryArchSuffix
 	lib:libQt5Core$secondaryArchSuffix
 	lib:libQt5Gui$secondaryArchSuffix
 	lib:libQt5Svg$secondaryArchSuffix
 	lib:libQt5Widgets$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
-	lib:libguile$secondaryArchSuffix
+	lib:libguile$secondaryArchSuffix <= 19
```
extrowerk: Guile 1.8.x recipe says:
```libguileVersion="17.4.0"```, which becomes 17 as compat, so why are you constraining it to 19?

File: app-office/texmacs/texmacs-1.99.12.recipe [hunk 12]
```diff
@@ -0,0 +1,102 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libiconv$secondaryArchSuffix
+	devel:libQt5Concurrent$secondaryArchSuffix
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libguile$secondaryArchSuffix <= 19
```
extrowerk: Should be 17

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+	cmd:autoconf
+	cmd:automake
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:gettext$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	cmd:guile_config$secondaryArchSuffix <= 1.9
```
extrowerk: It should be 1.8.8, as guile recipe says:
```cmd:guile_config$commandSuffix = $portVersion```

File: app-office/texmacs/texmacs-1.99.12.recipe [hunk 14]
```diff
@@ -0,0 +1,102 @@
...
+	cmd:guile$secondaryArchSuffix <= 1.9
```
extrowerk: Also ```cmd:guile$commandSuffix = $portVersion```

extrowerk: Should be 1.8.8, or  you should introduce compat string for the cmd's in the guile recipe to be able to use 1.8 here.

mgubi: As far as I understand compat is for "backward" compatibility. Here I just want to express that TeXmacs can work with any guile < 2.0. For example one user could install guile 1.6 and this should be fine.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I understand but (see below) I want to leave the user the freedom to choose different but compatible versions of guile. No need to overcontraint. TeXmacs will work just fine with any guile < 2.0.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: (see below) Guile 2.0 has libraries with version 20. This is the first version which does not work with TeXmacs, all the lower versioned libraries are ok.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: (see below)

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Also not needed ...

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+ADDITIONAL_FILES="texmacs.rdef.in"
+
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	texmacs$secondaryArchSuffix = $portVersion
+	app:TeXmacs
+	"
+
```
Begasus: No empty line here

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Shouldn't be needed, Guile 1.8 provides "lib:libguile$secondaryArchSuffix", Guile 2.2.4 provides "lib:libguile_2.2$secondaryArchSuffix" so only using lib/devel:libguile$secondaryArchSuffix should be enough ...

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: On my machine the above constraint seems to work fine.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I'm sorry. I'm so bad at newlines.

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Not needed, devel:libguile$secondaryArchsuffix provides cmd:guile_config$commandSuffix

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Again not needed (see above) ...

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: not anymore, I've splitted the guile1 package by putting cmd:guile in guile1_tools so that there are no conflicts anymore.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: It will not harm anyway and it would protect from some other package providing libguile with higher version. I would tend to keep. There is no advantage in removing it.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I was thinking that the idea is to have all the requirements explicit. I need here both the libraries and the guile_config script which are two independent resolvable. My understanding of the recommendations in the docs is that I need to explicitly put all the requirement, whether or not they are currently packaged together.

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Then I tend to agree with @extrowerk here, keep the version provided by Guile 1.8 (17), Guile doesn't provide libguile (but libguile-2.2)

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: cmd:guile_config is used to find the libraries headers needed to build TeXmacs, don't think they are required to run TeXmacs? (same for cmd:guile)?

mgubi: TeXmacs configuration script requires guile to be present in the path, otherwise abort with error. it also requires guile_config for the same reason. that is why I put them in build requirements but not in the runtime requirements.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: constraint is needed since if there is another guile package available on the machine (e.g. guile-2.x), TeXmacs configuration process could take up the wrong ones.

mgubi: I think is natural to requite cmd:guile_config, in the same way one requires cmd:pkg_config to be available. and cmd:guile as I said is necessary otherwise the configuration process of TeXmacs aborts.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: Ok. I will change.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I've changed to == 17.4.0

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,101 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	mkdir $appsDir
+	rm $binDir/texmacs
+	mv $binDir/fig2ps $libDir/TeXmacs/bin/fig2ps
+	mv $libDir/TeXmacs/bin/texmacs.bin $appsDir/TeXmacs
```
Begasus: Looks a bit odd to me, for one $libDir/TeXmacs/bin doesn't exist?
Is the location correct? (normally a binary shouldn't go into $libDir?)
Building atm to check

Begasus:
Output when building on gcc2h (secondary architecture)
https://paste.ubuntu.com/p/WyXJJV7M9f/

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: In other Unix systems these auxiliary binaries (which are needed internally to convert or to interface to external programs) are put in places like /usr/local/lib/TeXmacs/bin, so I kept the same style. The only difference is that usually fig2ps is installed in /usr/local/bin/fig2ps but I didn't wanted to pollute this place and moved it with the others. On my system the packaging works fine but I do not have a secondary architecture. I'm looking into it, probably there is something wrong in my use of $libDir .

mgubi: suggestions?

mgubi: what is the content of wrapper-script in your machine?

mgubi: I think I made a mistake here, and needed to use $libExecDir which according to https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes is the place where one can put the internal executables of the port.  I changed the recipe. Can you try again? My vm is 64 bit and I do not have a setup with a secondary architecture.

Begasus: Cleaning the build and will check again

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,101 @@
...
+	cmd:automake
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:gettext$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	cmd:guile_config$secondaryArchSuffix == 1.8.8
+	cmd:guile$secondaryArchSuffix == 1.8.8
```
Begasus: So far $secondaryArchSuffix isn't needed here:
```diff
-	cmd:guile_config$secondaryArchSuffix == 1.8.8
-	cmd:guile$secondaryArchSuffix == 1.8.8
+	cmd:guile_config == 1.8.8
+	cmd:guile == 1.8.8
```
(both use $prefix/bin so not needed)
On a sidenote, does TeXmacs need cmd:guild at runtime or buildtime?
cmd:guile_config should be provided by guile1_devel package (checking here now) so probably not needed at all here ...

Begasus: So far "cmd:guile_config" isn't needed here

mgubi: cmd:guild  is not needed. cmd:guile and cmd:guile_config are needed only at buildtime

mgubi: I put there  cmd:guile_config since I read that all dependencies have to be made explicit. However is ok for me to remove it. Is indeed provided by guile1_devel

mgubi: In the guile1 recipe they are declared as `cmd:guile_config$commandSuffix`and `cmd:guile$commandSuffix`

mgubi: I will wait to see if your build complete. Is not clear to me what to do here.

Begasus: > I will wait to see if your build complete. Is not clear to me what to do here.

Maybe move "cmd:guile  == 1.8.8" to BUILD_PREREQUIRES? (didn't check that one yet)

Begasus: > I put there cmd:guile_config since I read that all dependencies have to be made explicit. However is ok for me to remove it. Is indeed provided by guile1_devel

For packages relying on libSDL we don't need "cmd:sdl_config" also as it's provided by the _devel package (as long as pkgconfig file is supplied and finds it, which in this case it does)

Begasus:
Still needs some finetuning (build is still running atm) will check for updates tomorrow

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: This is not clear to me, many packages seems to declare the depency explicitly: https://github.com/haikuports/haikuports/search?q=cmd%3Asdl_config&unscoped_q=cmd%3Asdl_config

mgubi: Also, what about the ARCHITECTURES variable? I put "?" on archs I cannot test myself, but if it works on yours then I can modify this variable accordingly?

Begasus: > Also, what about the ARCHITECTURES variable? I put "?" on archs I cannot test myself, but if it works on yours then I can modify this variable accordingly?

15 recipes using cmd:sdl_config, some don't use cmd:pkg_config some do, depending if they pick up sdl-config from the sdl_devel package I presume, but shouldn't be needed in most cases (a lot more recipes use the SDL libraries without the dependency) :)

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,101 @@
...
+COPYRIGHT="Free Software Foundation"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/mgubi/texmacs/archive/texmacs-$portVersion-haiku-qt5.tar.gz"
+CHECKSUM_SHA256="607fa4db4676b78780f7af086feec2e2332dc18a371d79fa50d640de6db10b43"
+SOURCE_DIR="texmacs-texmacs-$portVersion-haiku-qt5/src"
+PATCHES="texmacs-$portVersion.patchset"
+ADDITIONAL_FILES="texmacs.rdef.in"
+
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
```
Begasus: Should be:
```diff
-ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+ARCHITECTURES="!x86_gcc2 x86_64"
```

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,101 @@
...
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libiconv$secondaryArchSuffix
+	devel:libQt5Concurrent$secondaryArchSuffix
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libguile$secondaryArchSuffix == 17.4.0
```
Begasus: Versioning not needed here (and for lib:libguile) as they are only provided by guile1:
```diff
- 	devel:libguile$secondaryArchSuffix == 17.4.0
+ 	devel:libguile$secondaryArchSuffix
```

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: ok

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,100 @@
+SUMMARY="A scientific editing platform designed to create beautiful technical documents"
+DESCRIPTION="TeXmacs provides a unified and user friendly framework for editing \
+structured documents with different types of content: text, mathematics, graphics, \
+interactive content, slides, etc.\
+Its rendering engine uses high-quality typesetting algorithms so as to produce \
+professionally looking documents, which can either be printed out or presented from \
+a laptop."
```
Begasus: Didn't focus on this yet :)
Maybe for the SUMMARY: "Wysiwyg text processor with high-quality maths" (from gentoo http://gpo.zugaina.org/app-office/texmacs) (we like summary to be short) :)
For DESCRIPTION, some lines cross the 80 char limit see: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#general-tips

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,100 @@
+SUMMARY="A scientific editing platform designed to create beautiful technical documents"
+DESCRIPTION="TeXmacs provides a unified and user friendly framework for editing \
+structured documents with different types of content: text, mathematics, graphics, \
+interactive content, slides, etc.\
```
Begasus: Remove the backslash here, that way the new paragraph starts on a new line in HaikuDepot

mgubi: done.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I'm not very keen of that description. I shortened ours.

Begasus: What's wrong with using yours, if so you can change that here, keeping in mind the 80 char limit

Begasus: Maybe use the description from https://github.com/texmacs/texmacs ?

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,100 @@
...
+}
+
+INSTALL()
+{
+	make install
+
+	mkdir $appsDir
+	rm $binDir/texmacs
+	mv $binDir/fig2ps $libExecDir/TeXmacs/bin/fig2ps
+	mv $libExecDir/TeXmacs/bin/texmacs.bin $appsDir/TeXmacs
```
Begasus: Some suggestions, atm there is an empty $binDir, $libExecDir/TeXmacs/lib and we don't use pixmaps ($dataDir/pixmaps)
```diff
-	rm $binDir/texmacs
	mv $binDir/fig2ps $libExecDir/TeXmacs/bin/fig2ps
	mv $libExecDir/TeXmacs/bin/texmacs.bin $appsDir/TeXmacs
+	rm -rf $binDir
+	rm -rf $dataDir/pixmaps
+	rm -rf $libExecDir/TeXmacs/lib
```

Maybe we need to review $libExecDir also here, atm it points to $prefix/lib instead of $libDir: https://github.com/haikuports/haikuporter/blob/8d5c14b6e9d4d0b0477e798afee40281f194c923/HaikuPorter/Port.py#L1173
Search reveals a --omit-dirs for this also: https://github.com/haikuports/haikuports/search?q=%24libExecDir&unscoped_q=%24libExecDir

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I've rephrased taking into account the description in wikipedia. The important thing is the "structured" feature, and technically is not more WYSIWYG then WYSIWYM. I think I should change that readme :)

app-office/texmacs/texmacs-1.99.12.recipe
Begasus: Strike the suggestion for now, not doing what I had in mind :)

mgubi: - We remove the pixmaps directory: it contains the images for all the icons in TeXmacs, both in xpm and png format. They are needed for the program to function in any platform.
- I think now the libExecDir stuff is fine. If we try to change it from the configure line I will have to  patch the Makefile. I do not see any problem with the current recipe. TeXmacs install a certain number of binary files into $libexecdir/TeXmacs/bin and I just moved there fig2ps in order not to pollute #binDir unnecessarily.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,102 @@
...
+
+INSTALL()
+{
+	make install
+
+	mkdir $appsDir
+	mv $binDir/fig2ps $libExecDir/TeXmacs/bin/fig2ps
+	mv $libExecDir/TeXmacs/bin/texmacs.bin $appsDir/TeXmacs
+	rm -rf $binDir
+	rm -rf $libExecDir/TeXmacs/lib
```
Begasus: For some reason this deletes the whole $libExecDir, so maybe we can just leave it out (checked with the content in HaikuDepot)?
```diff
- 	rm -rf $libExecDir/TeXmacs/lib
```
PS, you can check a rebuild with "hp texmacs -F" (this skips the build and jumps to "INSTALL")

mgubi: It works fine on my machine, lib/TeXmacs/bin is in the package the recipe produces. I do not see any problem.

mgubi: I checked with package list

mgubi:
@Begasus , as per your suggestion I removed the $dataDir/pixmaps directory which indeed is not used.

app-office/texmacs/texmacs-1.99.12.recipe
mgubi: I've also checked in HaikuDepot and the package seems ok (lib/TeXmacs/bin and its content are present)

Begasus: For some reason this is failing on gcc2h then? 
https://ibb.co/Tw1jF9c

Begasus: There hasn't been any change to the patchset aside from the recipe yes?

mgubi: No change. There are no local changes in my git repository.

mgubi: what is the value of $libExecDir in your machine? I do not anyway understand why lib/TeXmacs/bin disappears.

Begasus: I think it's due to "rm -rf $binDir", reverted that to what you initially had "rm $binDir/texmacs" and that seems to work, but leaves an empty $binDir behind

mgubi: What about `rm -rf $binDir/texmacs` and `rmdir $binDir$ ? I've pushed a change right now.

mgubi:
> From what I know this wouldn't work as $binDir is a directory not a file, changing that to -r instead of -f takes us back to a removed $libExecDir, maybe just leave it out, it doesn't harm and probably should be fixed in haikuporter itself

What do you mean? I've changed the recipe so that it removes the file `texmacs` in `$binDir`and then removes the directory itself with `rmdir`. It works on my machine.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,103 @@
...
+	lib:libQt5Svg$secondaryArchSuffix
+	lib:libQt5Widgets$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libguile$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libgmp$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
+	lib:libltdl$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
```
Begasus: Sort alphabetically (also in BUILD_REQUIRES)

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,103 @@
...
+	cmd:autoconf
+	cmd:automake
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:gettext$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:sed
+	cmd:guile == 1.8.8
```
Begasus: Some of these (autoconf/automake ... ) are probably obsolete as no autoreconf is used, needs to be checked

mgubi: I removed the unused dependencies and checked it works on my machine. I reodered also the dependencies in alphabetical order.

app-office/texmacs/texmacs-1.99.12.recipe
```diff
@@ -0,0 +1,100 @@
...
+{
+	make install
+
+	mkdir $appsDir
+	mv $binDir/fig2ps $libExecDir/TeXmacs/bin/fig2ps
+	mv $libExecDir/TeXmacs/bin/texmacs.bin $appsDir/TeXmacs
+	rm -f $binDir/texmacs
+	rm -rf $dataDir/pixmaps
+	rm -rf $libExecDir/TeXmacs/lib
+
```
Begasus: One empty line to much here

Begasus: If you could take care of that I'll merge, plenty of time for others if they had any objections :)

Begasus:
Thanks for sticking with this one, it's not been an easy ride, if buildbot is fine with it there shouldn't be any other changes needed for now :+1:

PR #4820, opened 2020-03-23
boehm_gc:  fix for "mmap(PROT_NONE) failed" in building GNU Guile 2.2.7 

mgubi:
The build of Guile 2.2.7 aborts with the following error
```
mmap(PROT_NONE) failed
```
which is due to the GC. This new patch solves the problem. I've also submitted an issue upstream.  The patch follows what happens in other architectures: see
(https://github.com/ivmai/bdwgc/blob/master/os_dep.c) line 2578 in the upcoming 1.0 version of the GC.

### DIFF
--- a/dev-libs/boehm_gc/boehm_gc-8.0.4.recipe
+++ b/dev-libs/boehm_gc/boehm_gc-8.0.4.recipe
@@ -22,6 +22,7 @@ REVISION="1"
 SOURCE_URI="https://github.com/ivmai/bdwgc/releases/download/v$portVersion/gc-$portVersion.tar.gz"
 CHECKSUM_SHA256="436a0ddc67b1ac0b0405b61a9675bca9e075c8156f4debd1d06f3a56c7cd289d"
 SOURCE_DIR="gc-$portVersion"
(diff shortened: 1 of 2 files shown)


dev-libs/boehm_gc/boehm_gc-8.0.4.recipe
```diff
@@ -22,6 +22,7 @@ REVISION="1"
 SOURCE_URI="https://github.com/ivmai/bdwgc/releases/download/v$portVersion/gc-$portVersion.tar.gz"
 CHECKSUM_SHA256="436a0ddc67b1ac0b0405b61a9675bca9e075c8156f4debd1d06f3a56c7cd289d"
 SOURCE_DIR="gc-$portVersion"
+PATCHES="boehm_gc-8.0.4.patchset"
```
diversys: Please use $portVersion instead of 8.0.4

diversys:
Please bump revision so the package can be rebuilt.

diversys:
Almost there! Please add more meaningful commit message to your fix in the patchset and rename this PR to "boehm_gc: fix crash". Thanks!

PR #4830, opened 2020-03-27
qtermwidget Add recipe

khallebal:
The recipe still doesn't work, i still don't see what's wrong with it? please take a closer look at it, i'm really clueless right now.

### DIFF
--- a/dev-qt/qtermwidget/patches/qtermwidget-0.14.1.patchset
+++ b/dev-qt/qtermwidget/patches/qtermwidget-0.14.1.patchset
@@ -1,0 +1,104 @@
+From bf2600d18ee7697fb03cf23a9d3a31bf73b0b07d Mon Sep 17 00:00:00 2001
+From: Khaled <khallbal@gmail.com>
+Date: Thu, 26 Mar 2020 14:32:27 +0100
+Subject: Disable utmp.h/utmpx.h stuff, missing in haiku
+
+
+diff --git a/lib/kpty.cpp b/lib/kpty.cpp
+index 542bde3..e66d437 100644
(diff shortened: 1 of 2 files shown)


extrowerk:
My buildlog :
http://0x0.st/iBv0.txt

extrowerk:
Adding the missing #endif to the ```lib/qtermwidget.cpp``` (line 273) build suceeded.

extrowerk:
It seems there is no tests, the example program builds fine and can run it from Terminal, but it complains about many things. Have you tested with something else? Did it work?

khallebal:
> Have you tested with something else? Did it work?

No not yet it's just a dependcy of the qterminal app, as you can see i got stuck with the recipe, i fixed the missing #endif and pushed it, my local recipe is still not working, i don't know what's wrong with my setup, can you please merge it so i can try to build it from the official repo?

extrowerk:
Remove the work folder and try again.
Nope, i won't merge it, as we need to know first if it actually works or not.

khallebal:
The terminal example starts up the menu works but it deosn't have a prompt and you can't type anything in it, so abviously the utmp/utmpx code is needed for this to work but haiku has no support for it yet, so i'm gonna have to give up on it as i don't know how to make it work without it.

extrowerk:
Utmpx is posix (but dont know which posix version). You can create an enhancement ticket for it at the Haiku bugtracker.

khallebal:
yeah, it's posix and of the latest i think? and there is already a 10 year old ticket about it but nobody implemented it partly because haiku is mono user and the kernel is missing some stuff as well if i'm not mistaking.

PR #4834, opened 2020-03-29
freac: bump to 1.1.0

enzo1982:


diversys:
`addAppDeskbarSymlink "$appsDir"/freac fre:ac
`
This link doesn't work

enzo1982:
> `addAppDeskbarSymlink "$appsDir"/freac fre:ac `
> This link doesn't work

Looks like this is related to how the application ID passed to BApplication is constructed in my class library. The ID does not match the one stored in the application resources unless the link is called freac without the colon.

This still worked with Beta 1, but fails with hrev54017. But I think it's something that should be fixed on my end.

Thank you for bringing this up! I'll send another PR changing the link for now.

diversys:
Removing ":" from the link name seems to work around it.

enzo1982:
Yes, #4835 should fix it.

I'll re-introduce the colon once I have the root-cause fixed in my class library. Probably with the next update in a few weeks.

PR #4844, opened 2020-04-04
LibreOffice: add several fonts families to requires

### LINKED ISSUE #4842: Include LibreOffice bundled fonts
LibreOffice comes with a number of fonts bundled [with it](https://github.com/LibreOffice/core/tree/master/external/more_fonts), however on Haiku these fonts are not available. It is imperative to have these at least with LibreOffice, since they include metric-compatible MS Fonts substitutes to avoid formatting problems.

bitigchi:
Closes #4842.


waddlesplash:
We should probably package the fonts separately IMO?

bitigchi:
Packaging the fonts separately only works if all the bundled fonts are available in HaikuDepot. LO provides fonts like Carlito and Caladea, which are required for metric compatibility with fonts like Calibri and Cambria.

diversys:
I think you can add them all at the top of requires.

app-office/libreoffice/libreoffice-6.4.3.1.recipe
```diff
@@ -48,6 +48,20 @@ PROVIDES="
...
+	liberation_fonts
+	libertine
+	noto
+	noto_emoji
+	noto_chroscore
+	noto_condensed
+	noto_regional
+	noto_sans_cjk
+	noto_serif_cjk
+	source_pro
```
waddlesplash, line 64: I don't think there's a reason to add the Fira or Noto fonts, is there?

bitigchi: I added Noto Fonts as per @diversys's suggestion in case there is a system font change. I also agree with his suggestion, plus we need to make sure Asian languages are well supported.

Fira is a great font with great language support, it is a substitute for the other fonts I couldn't add because it's not present in HP (also aesthetics).

augustin: Uh, if the system font changes, then so should LibreOffice's. We can also rebuild the package if needed when that happens. But we shouldn't tie people to Noto inside LibreOffice.

IMHO we should not have LibreOffice require any more fonts than it actually needs so users can choose. Requiring the Liberation fonts and others for MS metric compatibility sounds good, but we should not require more than that.

waddlesplash: Uh, if the system font changes, then so should LibreOffice's. We can also rebuild the package if needed when that happens. But we shouldn't tie people to Noto inside LibreOffice.IMHO we should not have LibreOffice require any more fonts than it actually needs so users can choose. Requiring the Liberation fonts and others for MS metric compatibility sounds good, but we should not require more than that.

bitigchi: Do you think we should not include any additional Noto font at all? What about Asian languages?

waddlesplash: Noto CJK is already installed in Haiku by default. You can put it in SUGGESTS (IIRC), I guess, but not REQUIRES.

KapiX: There is no such thing as SUGGESTS. There is only SUPPLEMENTS which works in the other direction.

bitigchi: @waddlesplash, AFAICS only `noto_sans_cjk_jp` is installed. There are two issues with this:

1. `noto_sans_cjk_jp` defaults to Japanese versions of the symbols. As far as I know, his is not ideal when viewing Traditional Chinese text, which would display the Japanese variant. Maybe a native speaker can confirm. In any case, `noto_sans_cjk` should be the package that is installed.
2. `noto_serif_cjk` is still missing.

waddlesplash: 1. The other CJK characters are contained within the font; and as long as LibreOffice knows how to use them (which it does), this should be just fine. So there is no need to require `noto_sans_cjk` itself, then.

2. Fair enough.

diversys: > Uh, if the system font changes, then so should LibreOffice's. 
Well at least on macOS LO still bundles Noto fonts while the OS is using San Francisco font by default.

waddlesplash: OK, fair enough.

korli: Tip: add a libreoffice_fonts subpackage which supplements the libreoffice package. The libreoffrice_fonts subpackage can require then a list of fonts.

diversys:
Also don't forget to bump revision.

Begasus:
@bitigchi any progress on this? (conflicting files by now)

PR #4845, opened 2020-04-05
dev-lang/nim: update to 1.2.0

alaviss:
Tested on x86_gcc2 :)

Some tests will fail due to known Haiku bugs and/or because we just don't have the libraries required (ie. SFML).


alaviss:
Too many tests are failing on x86_64. Don't merge yet.

korli:
@alaviss I see you pushed some commits directly upstream. The recipe was updated to 1.4.6 in 41d6ef10691ab8d4d4fd54adf3f378f875fcc4f7  (while some unit tests still fail). I take that this version is in better shape as 0.19.6.
Is this PR still relevant?

alaviss:
> Is this PR still relevant?

No, many versions has been released since and I haven't got the time to update this PR. All of the patches here are upstreamed.

The only issue that I haven't managed to solve is that a lot of tests using thread-local storage just straight up crashes, but I haven't tested again for a very long time so it might no longer be an issue.

Feel free to update this to new versions as needed and if you need help solving any test errors, just ping me and I'll have a look and help you push any fixes upstream.

korli:
Ok. I only tested on beta2, so maybe the situation is better on nightly. Closing.

PR #4847, opened 2020-04-06
CFITSIO - Added symlinks in build: fix

theclue:
I've just checked that some apps need the shared library having a different name. Darwin makefile has some symlinks in place to do that.

I''ve patched configure.in to support Haiku and to create those symlinks as well.


sci-libs/cfitsio/cfitsio-3.47.recipe
```diff
@@ -0,0 +1,87 @@
...
+	"
+
+BUILD()
+{
+	if [ $effectiveTargetArchitecture == x86_64 ]; then
+		GCC5FLAGS="--enable-x86-64"
+	fi
+
+	autoconf
+	runConfigure ./configure --enable-reentrant --enable-ssse3 --with-bzip2
```
extrowerk, line 61: Target Northwood on x86 and Prescott on 64 bit.
Ie.: Northwood got no SSE3

korli:
this can't be merged, as this conflicts with the current sci-libs/cfitsio/cfitsio-3.47.recipe in master branch.

sci-libs/cfitsio/cfitsio-3.47.recipe
theclue: done!

theclue:
I'm still investigating why suddenly haikuporters has started to say that some libcurl dependencies are directly needed (as it was statically linked). Cfr: https://discuss.haiku-os.org/t/haikuporter-waiting-for-build-package-to-be-activated-how-to-i-get-out-of-these/9364/5

As a workaround I directly added in the recipe, it is not a blocking issue I guess but I'm still investigating

korli:
obviously there are too many dependencies in BUILD_REQUIRES. Please don't copy paste between sections like this.

theclue:
> obviously there are too many dependencies in BUILD_REQUIRES. Please don't copy paste between sections like this.

I didn't copypaste; as I said above, with no apparent reason (recipe was working before) I've started to get all those POLICY ERROR for those which apparently seems dependencies of libcurl and not libcfitsio direcly.

@pulkomandy suggested to add these libs are requirements as a workaround (see the forum link).

I'm trying to understand why this happended so suddenly but i've no clue actually (there're no static libraries in the build)

sci-libs/cfitsio/cfitsio-3.47.recipe
```diff
@@ -51,11 +63,14 @@ TEST_REQUIRES="
 
 BUILD()
 {
+	WHITSSE=""
```
korli, line 66:
```suggestion
	WITHSSE=""
```

theclue:
About these dependencies, these are for sure libcurl dependencies. In the `Makefile` file i can see them clearly referenced:

`LIBS_CURL =	-L/packages/curl-7.69.1-1/.self/develop/lib -lcurl -lnghttp2 -lssl -lcrypto -lssl -lcrypto -lz -lnetwork`

and later in the build step:

`${SHLIB_LD} ${LDFLAGS} -o ${CFITSIO_SHLIB} ${OBJECTS} -lm ${LIBS_CURL} ${LIBS}`

I don't know how to fix this, but I will continue investigating....

waddlesplash:
cc @alaviss, we should probably remove that from curl?

theclue:
@waddlesplash I don't think the issue is coming from curl. I think it's from cfitsio itself.

I can clearly see that in `configure`. It calls `curl-config` and outputs all these libs in `LIBS_CURL`.

alaviss:
> cc @alaviss, we should probably remove that from curl?

We need `--as-needed` as a default `LDFLAGS` in HaikuPorter.

Begasus:
@theclue check this: https://github.com/haikuports/haikuports/search?q=shlib_ld&unscoped_q=shlib_ld
Using that enabled to build the shared library "libcfitsio.so.8.3.47"
Changed it locally with the patch you already used ... (PS, test still fails)

Begasus:
Test looks OK now :) https://ibb.co/DzvxjqQ

theclue:
I've tried moving the utils binaries in a `cfitsio_utils` subpackages, but for some reasons they won't work this way for a runtime error for `libcfitsio.so.8 not found`. No errors if the binaries are in the same (main) package.

I don't like that, I'm worried that other packages using the library will fail as well.

Begasus:
Haven't tried splitting the package into a subpackage for the cmd's, but I do have a versioned library now and the test runs OK, will try to see if I can get around tomorrow on splitting the package ... (PS, not a requirement at this stage) :)

Begasus:
Is there any package depending on this one that I could check @theclue ?

Begasus:
OK, created a seperate package for the utils, build is OK here and also the test seems to be OK :)

theclue:
> Is there any package depending on this one that I could check @theclue ?

Have one in draft, not submitted yet actually

theclue:
0k, is it ready to merge then?

Begasus:
> 0k, is it ready to merge then?

Not at this state I think, do you have a versioned library? Does the test work OK with haikuporter for you? ... (will check back tomorrow)

theclue:
> > 0k, is it ready to merge then?
> 
> Not at this state I think, do you have a versioned library? Does the test work OK with haikuporter for you? ... (will check back tomorrow)

yep, tests are ok in hp.

This is a quick recipe to test if the library eventually works: https://github.com/theclue/haikuports/tree/fitsToPNG/sci-astronomy/fitspng (I'll complete it tomorrow)

It works to me, but I've so many unesplicable behaviours in this port (symlinks sometimes working, sometimes not for ex.) that I cannot be 100% sure actually.

Begasus:
I'll try to link the changes I have here and checkout fitpng when I get around to it :+1:

Begasus:
Here are my changes, could you check them on x86_64? 
https://github.com/Begasus/haikuports/commit/9fb3502f0eaaa486d8d627c1aabf3a5c7b10d0b5

Begasus:
OK, checked a local build on x86_64, build is OK and the tests results seem to be OK also

theclue:
You anticipated me :) ok, glad it works. But I'm afraid i've lost something in all this mess. What're the next steps now?

Begasus:
> What're the next steps now?
You could try to include my changes (or some parts), update your PR here and check if it's working for you :)

theclue:
> You could try to include my changes (or some parts), update your PR here and check if it's working for you :)

Done and updated the PR with your changes. They are almost clear but...

In your recipe, `cfitsio_utils `package doesn't install `cfitsio` lib as a dependency. When manually installed the lib, the utils work perfectly

In my one i renamed the utils sibling package in `cfitsio${secondaryArchSuffix}_utils` (you left the base name in your). This way it WILL install the mainlib as a dependency but the utils won't work.

```
~/haikuports/sci-libs/cfitsio> fpack
runtime_loader: Cannot open file libcfitsio.so.8: No such file or directory
```
any idea?

Begasus:
Give me some time until I can check (or someone else) this out :)

PR #4870, opened 2020-04-12
pearpc, remove old bep file

Begasus:


Begasus:
Can't check if it actually works, so would be nice if someone could :)

CodeforEvolution:
I can confirm that this application builds on x86_gcc2 using the x86 secondary architecture. The program runs and is even able to open up a GUI window. However, the window stops redrawing as soon as you drag the window around, and keyboard input is not handled correctly. (F12 did not enable/disable the mouse as mentioned on the SDL interface)

CodeforEvolution:
Well, I was able to fix the BeOS interface build and it works a lot better than the SDL interface! :D
While it isn't perfect (the close button doesn't work and the F12 key still does not disable mouse capturing) it fixes keyboard input and redrawing problems. You'll be able to remove sdl dependencies from this recipe and your old patch.

Here's the quick fix of what you'll need to do to fix the build:

1. Go to the src/system/osapi/beos/syscdrom.cc file
2. Look for the IO_IDE_TRACE line at the top of the file
3. Replace IO_IDE_TRACE("[CDROM/BEOS] "msg) with IO_IDE_TRACE("[CDROM/BEOS] ", msg)
4. Look for the IO_IDE_WARN line at the top of the file
5. Replace IO_IDE_WARN("[CDROM/BEOS] "msg) with IO_IDE_WARN("[CDROM/BEOS] ", msg) 
6. That's all!

I'll try to upstream this small fix too.

Begasus:
@CodeforEvolution checked their github page, latest commit there seems to be from 2015, so I hardly think a PR will be merged any time soon there?
Could you show what's needed to have a working version with the changes you made? I'll gladly take my changes here down if need be :)

pulkomandy:
Can't we just drop the recipe? Who is going to need this?

Begasus:
> Can't we just drop the recipe? Who is going to need this?

We can, but in the process also remove the old bep file for it?

pulkomandy:
Yes, I think there is not much use for pearpc anymore these days, so I'd remove everything.

X547:
Why remove? Some people may want to run old Mac OS versions. Is there alternative for PearPC?

pulkomandy:
qemu seems to work: https://wiki.qemu.org/Documentation/Platforms/PowerPC#PowerMac_G4_and_G3_emulation

CodeforEvolution:
@Begasus, the only changes that would be necessary would be to apply the small changes I mentioned above, remove SDL dependencies from the recipe, remove the SDL options from the configure script invoke, and make sure to apply the rdef file in the beos directory within the source code. I guess PearPC might be showing its age, is there anything that PearPC provides that Qemu's emulation doesn't provide?

Begasus:
> @Begasus, the only changes that would be necessary would be to apply the small changes I mentioned above, remove SDL dependencies from the recipe, remove the SDL options from the configure script invoke, and make sure to apply the rdef file in the beos directory within the source code. I guess PearPC might be showing its age, is there anything that PearPC provides that Qemu's emulation doesn't provide?

@CodeforEvolution I haven't checked pearpc from back in the days with BeOS R5, so I have no idea there, maybe someone else could answer this that is still using it?

PR #4879, opened 2020-04-17
neon, bump version, remove old recipes, packages depending on neon should be rebuild due compat change

Begasus:


Begasus:
From their website: API and ABI backwards-compatible with 0.27.x and later

The provides mention compat >= 29 (should be OK) but maybe it should be changed to 27?

pulkomandy:
The libversioncompat is fine, I think that's the only thing that matters.

Begasus:
> The libversioncompat is fine, I think that's the only thing that matters.

Hence why I didn't change the other :)

Begasus:
Checked with installing LibreOffice, works OK

korli:
Please tell in the commit message, that the compat version changed. Packages depending on neon should be rebuilt.

Begasus:
Packages using neon (from what I can find with a search at haikuports):
- [ ] cadaver
- [ ] musicbrainz
- [ ] libcoverart
- [ ] audacious_plugins
- [ ] subversion
- [ ] libreoffice

Begasus:
@korli Would you prefer a bulk revbump after the merge, or seperate revbumps? (Some patches need an update also (versioning = 31 now, issues with -Werror))

korli:
Bulk revbump when possible. If patches are needed individual revbump plus patch.

PR #4883, opened 2020-04-18
Bescreencapture v2.5.1

jackburton79:
Bugfix release for BeScreenCapture
I fixed a bug in BeScreenCapture where the GIF export would only export a few frames.
Here's the PR.


diversys:
It's better to rename existing recipe to the new one.

jackburton79:
Ok, I thought you wanted to keep also older versions.

Il sab 18 apr 2020, 11:33 Sergei Reznikov <notifications@github.com> ha
scritto:

Begasus:
> Ok, I thought you wanted to keep also older versions.

If the version bump is needed and the older version isn't really needed it's best to rename it to the latest version, no need to keep an older version around (only if an architecture brakes the build and previous version still worked on that)

PR #4884, opened 2020-04-18
libcoverart, fix build

Begasus:


Begasus:
This needs to wait before merging until neon is merged: https://github.com/haikuports/haikuports/pull/4879

Begasus:
Builds and installs OK on _x86, test run OK, didn't test x86_64 (yet)

korli:
Looks good. Theorically doesn't need a revbump if it wasn't enabled til today.

Begasus:
Will revert the revbump for this one later :)

Begasus:
@phw any chance this could be used for Picard? If it's not very useful we can just delete the old bep file instead of trying to get this to work :)

phw:
No, Picard does not need this, it has it's own implementation in Python. I think just going the easy way to fix this is fine :)

Begasus:
As this isn't used earlier, it's ok to have it in for now so people could check if there is a need for it ...

PR #4908, opened 2020-04-25
add yab library in preporation for updating the yab version.

bbjimmy:
Since we added some yab commands. the next version needs a new library, but we need the old library for existing programs.

### DIFF
--- a/dev-lang/yab/libyab1-1.7.8.recipe
+++ b/dev-lang/yab/libyab1-1.7.8.recipe
@@ -31,7 +31,7 @@ REQUIRES="
 	"
 
 BUILD_REQUIRES="
-	haiku_devel 
+	haiku_devel
 	"
 BUILD_PREREQUIRES="
 	cmd:bison
@@ -51,7 +51,7 @@ BUILD()
 
 INSTALL()
 {
-	mkdir -p $libDir 
+	mkdir -p $libDir
 	cp tmp/libyab1.so $libDir/
 	cp ../../sources-2/$targetArchitecture/libyab.so  $libDir/
 	prepareInstalledDevelLibs libyab1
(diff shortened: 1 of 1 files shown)


dev-lang/yab/libyab1-1.7.8.recipe
```diff
@@ -0,0 +1,58 @@
...
+CHECKSUM_SHA256_2="ec4ef57214e7fe010c3646543dc1fdbc02e9303db30e8113e518a299483a790f"
+
+
+ARCHITECTURES="x86_gcc2 !x86 x86_64"
+
+PROVIDES="
+	devel:libyab = $portVersion
+	lib:libyab = $portVersion
+	devel:libyab1 = $portVersion
+	lib:libyab1 = $portVersion
```
Begasus, line 25: Should need "$portName = $portVersion" as a first line (haikuporter takes $portName as a string to start the build)
Why not split the lib:* and devel:* into seperate packages as most library packages do?

lorglas: hello, i miss the reference to the changes made by me, furthermore libyab1 does not correspond to the usual logic of the libs. I wrote a recipe that keeps the old libyab.so and creates the new libyab-1.7.8. Furthermore, I have already written bbjimmy that I will adjust the package. the current package only contains yab and the respective libyab. This has the advantage that the packages become cleaner and can be processed separately.

lorglas:
hello, i miss the reference to the changes made by me, furthermore libyab1 does not correspond to the usual logic of the libs. I wrote a recipe that keeps the old libyab.so and creates the new libyab-1.7.8. Furthermore, I have already written bbjimmy that I will adjust the package. the current package only contains yab and the respective libyab. This has the advantage that the packages become cleaner and can be processed separately.

Begasus:
> hello, i miss the reference to the changes made by me, furthermore libyab1 does not correspond to the usual logic of the libs. I wrote a recipe that keeps the old libyab.so and creates the new libyab-1.7.8. Furthermore, I have already written bbjimmy that I will adjust the package. the current package only contains yab and the respective libyab. This has the advantage that the packages become cleaner and can be processed separately.

Should we wait for a PR coming from you then @lorglas ?

lorglas:
libyab1 isn't a good solution, as example, after a half year ago, nobody knows with libversion is it. Use recipe from #4909

Begasus:
Obselte as for https://github.com/haikuports/haikuports/pull/4909

PR #4923, opened 2020-04-30
revbump packages after merge from updated neon

Begasus:


Begasus:
neon needs to be merged before this revbump: https://github.com/haikuports/haikuports/pull/4879

Issue #4907, opened 2020-04-25
Allow Vim to install without reliance on HP


bitigchi:
When building Vim from source with [these](https://github.com/vim/vim/blob/master/runtime/doc/os_haiku.txt) instructions, even though `make` finishes okay, during `make install`, the recipe fails here:

> rc: Error! -:8 syntax error, unexpected $undefined
> Makefile:3386: recipe for target 'objects/os_haiku.rsrc' failed

What happens here is, `make install` does not understand what to do with `os_haiku.rdef` file, therefore when gvim is launched from the source directory, icons are not visible, there are black boxes instead. Ideal way is, Vim should be able to handle the assets itself.

Instead of solely relying on Haikuports for Vim builds, it should be possible to build Vim on its own. Vim has infinite configurability regarding how its features are handled, users should be able to compile and install Vim as they please.

Begasus:
Haven't checked this (build in Terminal), I usually use a prefix pointing to a seperate destdir for those ...

bitigchi:
@Begasus, the related Makefile section is [here](https://github.com/vim/vim/blob/49b2fb36ca94be14b98caf86420863d9bbe81a24/src/Makefile#L3655).

Begasus:
Maybe look at https://github.com/vim/vim/blob/49b2fb36ca94be14b98caf86420863d9bbe81a24/src/Makefile#L3385?

Begasus:
If so, you could try to fix it and upstream a patch for it :)

bitigchi:
@Begasus, I'm afraid this needs someone more experienced.

extrowerk:
> rc: Error! -:8 syntax error, unexpected $undefined
> Makefile:3386: recipe for target 'objects/os_haiku.rsrc' failed

I think a malformed rdef is the reason for this error. I had the exact same error as this line had a comment in it. Make sure your rdef file have exactly the same line for the icon section like this:
https://github.com/haikuports/haikuports/blob/61f96eb4b45dcb10bb160d434bc61edf1e11eb1e/media-gfx/yacreader/additional-files/yacreader.rdef.in#L28

extrowerk:
Ummm... I just checked an old vim recipe and the rdef file looks ok. But i see the rdef gets sed'ed in the recipe. Probably this is what missing if one does a manual build. No idea how to fix it for both environment (hp and vanilla) tho.

pulkomandy:
[...]

Issue #4895, opened 2020-04-23
Vim: :term not available on gVim


bitigchi:
If enabled during compile time, in gVim, `:term` opens an empty window, and that's it. gVim should also have `:term` support.

PR #4925, opened 2020-04-30
Vim: bump version & improve recipe

### LINKED ISSUE #4894: Vim: It's not possible to 'exit' terminal
Steps:

1. Open vim
2. Type `:term`
3. Type `exit` on terminal window

It should exit, but it just echoes the command, and hangs.

bitigchi:
This PR updates Vim to 8.2.0671. This version includes support for installing Vim without dependence on a HaikuPorts recipe via passing commands about assets handling to the Vim Makefile. Several build warnings have been tackled, and some duplicate GUI code has been eliminated. Documentation has been improved as well.

In addition to this, I believe it's not necessary to compile Vim twice in this recipe. Separating GUI compile from the CLI compile prevents Vim from using the `+clipboard` function, therefore it is not possible to use the system clipboard on the console version. By compiling Vim with GUI support only, this feature is available on both ends.

I have been testing this for the last couple of days, and haven't encountered any issues.

Fixes #4894, #4895, #4907


pulkomandy:
On Thu, Apr 30, 2020 at 02:24:59PM -0700, Emir Sarı wrote:
> In addition to this, I believe it&#39;s not necessary to compile Vim twice in this recipe. Separating GUI compile from the CLI compile prevents Vim from using the `+clipboard` function, therefore it is not possible to use the system clipboard on the console version. By compiling Vim with GUI support only, this feature is available on both ends.

This will make it impossible to run Vim from users without an app_server
connection (for example if you created a new user account and let it
connect to your machine using ssh). Also it shows an entry in DeskBar
(at least we should set the B_BACKGROUND_APP flag on the app when
running in console, I guess).

bitigchi:
Are there any other disadvantages to this except the ssh issue?

theclue:
I use vi mainly from ssh so pls don't ;)

bitigchi:
> I use vi mainly from ssh so pls don't ;)

Who needs vi when you have the ed(1) standard text editor? :P

app-editors/vim/vim-8.2.0742.recipe
```diff
@@ -101,18 +101,17 @@ BUILD()
 # CLI -------------------------------------------
 
 	# Clean objects from possible previous run to avoid mixup of GUI/CLI objects
-	rm objects/*.o || true
+        rm objects/*.o || true
```
diversys, line 104: Please use tabs.

bitigchi: Fixed.

diversys:
app-editors/vim/.swp shouldn't be there.

bitigchi:
Fixed that, slipped through my eyes.

PR #4940, opened 2020-05-03
argon2: bump version

AlwaysLivid:
Compiles on x86_64.


Begasus:
This is used by keepassxc, should be checked if that build is ok with this update

AlwaysLivid:
> This is used by keepassxc, should be checked if that build is ok with this update

Will check shortly.

AlwaysLivid:
Successfully compiles with the patches on x86_64.

https://paste.ubuntu.com/p/6wf77v3vRq/

korli:
But in the log one sees that keepassxc is built with :
       /boot/system/packages/argon2-20190614-2-x86_64.hpkg
	/boot/system/packages/argon2_devel-20190614-2-x86_64.hpkg
This means it was built with the old version of argon2.

AlwaysLivid:
> This means it was built with the old version of argon2.

Ugh, I thought I changed the branch properly. Huge thanks for catching this one for me. Will try once more.

AlwaysLivid:
Haven't replaced the checksum yet, do not merge.

AlwaysLivid:
Should be good to go. @Begasus

PR #4941, opened 2020-05-03
aspell: bump version

AlwaysLivid:
Compiles on x86_64.


korli:
there is also an an armadillo change in this PR. Please remove it.

PR #4945, opened 2020-05-03
openssh: Work around first boot user creation bug.

agmsmith:
Here's a quick fix for https://dev.haiku-os.org/ticket/14382 where we create the user and group in the post-install script too.  Needed for beta 2.  BTW, couldn't download from https when building in Haiku GCC2, changed URL so it works.

### DIFF
--- a/net-misc/openssh/additional-files/fix_openssh_post_install_usergroup.sh
+++ b/net-misc/openssh/additional-files/fix_openssh_post_install_usergroup.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+# Create the users and groups needed by sshd.  They are part of the package
+# metadata, but that currently (Beta1, Beta2) doesn't work for packages in a
+# newly installed system (the installation actions aren't run since the files
+# are just copied to the new system).  So until running installation stuff
+# on first boot is fixed (Beta3?), redundantly create the user and group in
+# the post-install script, which is run on the first boot by launch_daemon.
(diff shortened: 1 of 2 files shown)


waddlesplash:
The buildmaster will not build anything with a HTTP download url.

I thought we agreed in the meeting to put the server in a subpackage, and not install that by default, which will be cleaner here?

agmsmith:
There was a problem with Network Preferences still showing SSH daemon options even if the package wasn't installed (or at least when it was uninstalled).  I'll have a look at the http/s settings and check the notes from the meeting, tomorrow :-)

waddlesplash:
We should just fix Network Prefs then, it sounds relatively easy to do.

agmsmith:
There was a problem with Network Preferences still showing SSH daemon options even if the package wasn't installed (or at least when it was uninstalled).  I'll have a look at the http/s settings and check the notes from the meeting, tomorrow :-)

agmsmith:
Amended the commit to use HTTPS, which is working today in my Haiku build.  Odd.  Anyway, looking back at the Beta2 meeting log (now appended to the bug report), this quick fix to the post-install script seems to be the one to do.  Then a proper post-install reboot processing implementation, then remove this quick fix.

agmsmith:
One more thing, found out why the sshd user was in the wrong group, fix the recipe to create them in their new group rather than the default general user group.  Linux does it that way for the sshd user too.

PR #4947, opened 2020-05-05
Godot: add Godot 3.2.1 recipe

kenmays:
Built successfully on Haiku hrev54131 x86_64.


CodeforEvolution:
Thank you for the recipe @kenmays! However, please do not merge this just yet, I'm still working on fixing up the Haiku port of Godot. Even though it builds, it is lacking in feature support in terms of implemented functionality in the OS abstraction layer and is buggy when it comes to drawing.

kenmays:
Merge outdated. Created a new working patchset for Godot 3.2.3 package. Once patches are approved upstream we can
update at that time with Haiku R1B3+.

PR #4954, opened 2020-05-07
New Recipe: CCfits

theclue:
CCfits is an object oriented interface to the cfitsio library, which I ported some weeks ago and it was already merged in HP. Build, install and unit tests are ok

Only one glitch I was not able to fix (I'm not ever sure it's a bug in the recipe, actually) : the SONAME has a wrong version number scheme (0.0.0). Perhaps someone could help me spot the issue?

Tested in x86_64, but should works on other platforms as well.


extrowerk:
Tested on x86_64:
- builds fine
- packages fine
- test: i see some errors
- libversioncompat should be added
- compat flags should be added
- debuginfo could be added
- Fixpkgconfig should be used in INSTALL()

theclue:
Hi @extrowerk, tnx for the feedback!

I hope I've done fixing your requests

- error in tests: these are false positives coming from underlying cfitsio library. It just says that the generated FITS file was not valid (but the test itself is about generation, not the validity). Perhaps the unit test should deserve more love. Howevere, it is considered passed
- compat flags: added
- debugInfo: added. But I must admit that i've copied from another recipe as I have no idea of the purpose of this block.

extrowerk:
> * debugInfo: added. But I must admit that i've copied from another recipe as I have no idea of the purpose of this block.

Extracts the debug informations/symbols from the library making the runtime package smaller in size.

sci-libs/ccfits/ccfits-2.5.recipe
```diff
@@ -0,0 +1,114 @@
+SUMMARY="C++ Wrappers for the cfitsio library"
+DESCRIPTION="CCfits is an object oriented interface to the cfitsio library. \
+It is designed to make the capabilities of cfitsio available to programmers \
+working in C++. It is written in ANSI C++ and implemented \
+using the C++ Standard Library with namespaces, exception handling, \
+and member template functions."
+HOMEPAGE="https://heasarc.gsfc.nasa.gov/fitsio/CCfits"
+COPYRIGHT="2002-2019 Ben Dorman, Craig Gordon, Bryan Irby - HEASARC, NASA/GSFC"
```
Begasus, line 8: Maybe use the multi line copyright guideline here? https://github.com/haikuports/haikuports/wiki/HaikuPorter-BuildRecipes#a-simple-example-using-the-recommended-style-for-a-multi-line-copyright

sci-libs/ccfits/ccfits-2.5.recipe
```diff
@@ -0,0 +1,114 @@
...
+PATCHES="ccfits-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86_64 ?arm ?ppc ?sparc"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	ccfits$secondaryArchSuffix = $portVersion compat >= 2.5
```
Begasus, line 23: Why compat >= 2.5 ?

sci-libs/ccfits/ccfits-2.5.recipe
```diff
@@ -0,0 +1,114 @@
...
+	cmd:test
+	"
+
+defineDebugInfoPackage ccfits$secondaryArchSuffix \
+	"$libDir"/libCCfits.so.$libVersion
+
+PATCH()
+{
+	mv configure.in configure.ac || true
+}
```
Begasus, line 65: Is this needed? autoconf maybe gives a warning to use configure.ac over configure.in, but haven't encountered problems with them so far ...

sci-libs/ccfits/ccfits-2.5.recipe
theclue: But they are all from the same team/university...

sci-libs/ccfits/ccfits-2.5.recipe
theclue: They used to have a messed up versioning schema. Future versions SHOULD be compat with 2.5. But ok, I'll remove it

sci-libs/ccfits/ccfits-2.5.recipe
theclue: Not needed, but why not saving a warning if you can do that so easily (Gentoo port does, actually)?

sci-libs/ccfits/ccfits-2.5.recipe
theclue: I've seen in other ports. There is no a common way to do that, but I've seen here and there that multi-author lines looks ok if they belong to the same entity (for ex. zlib)

sci-libs/ccfits/ccfits-2.5.recipe
```diff
@@ -0,0 +1,123 @@
...
+HOMEPAGE="https://heasarc.gsfc.nasa.gov/fitsio/CCfits"
+COPYRIGHT="2002-2019 Ben Dorman, Craig Gordon, Bryan Irby - HEASARC, NASA/GSFC"
+LICENSE="CCFITS"
+REVISION="1"
+SOURCE_URI="https://heasarc.gsfc.nasa.gov/fitsio/CCfits/CCfits-$portVersion.tar.gz"
+CHECKSUM_SHA256="938ecd25239e65f519b8d2b50702416edc723de5f0a5387cceea8c4004a44740"
+SOURCE_DIR="CCfits"
+PATCHES="ccfits-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86_64 ?arm ?ppc ?sparc"
```
Begasus, line 16: This won't work for x86_gcc2 as there is no libcfitsio for primary architecture, so:
```diff
-ARCHITECTURES="x86_gcc2 x86_64 ?arm ?ppc ?sparc"
+ARCHITECTURES="!x86_gcc2 x86_64 ?arm ?ppc ?sparc"
```

sci-libs/ccfits/ccfits-2.5.recipe
```diff
@@ -0,0 +1,123 @@
...
+	"
+
+PROVIDES_devel="
+	ccfits${secondaryArchSuffix}_devel = $portVersion
+	devel:libCCfits$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	ccfits$secondaryArchSuffix == $portVersion base
+	haiku$secondaryArchSuffix
+	lib:libcfitsio$secondaryArchSuffix
```
Begasus, line 46: Don't think this is needed (checked with output from pkg-config):
```diff
- 	lib:libcfitsio$secondaryArchSuffix
```

sci-libs/ccfits/ccfits-2.5.recipe
```diff
@@ -0,0 +1,123 @@
...
+SOURCE_DIR="CCfits"
+PATCHES="ccfits-$portVersion.patchset"
+
+ARCHITECTURES="x86_gcc2 x86_64 ?arm ?ppc ?sparc"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+commandSuffix=$secondaryArchSuffix
```
Begasus, line 22: commandSuffix* before libVersion* ... https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

PR #4972, opened 2020-05-11
New Recipe: Wcslib

theclue:
WCSLIB is a C library, supplied with a full set of Fortran wrappers, that implements the World Coordinate System (WCS) convention in FITS. Since PGPLOT is not yet available on Haiku, \
this package doesn't include PGBOX (routines for drawing general curvilinear coordinate graticules) and related utilities.

Porting PGPLOT seems to be tricky at first look. I'll probably need much time to port it. That's why I decided to push this library without that (very optional) dependency in the meantime.

SIDE NOTE: I'm not an experienced Fortran dev so there're room for future improvements in this recipe:

- ~~Where exactly the fortran wrappers are supposed to be installed in the tree? Perhaps a sibling package?~~
- ~~I don't know which files must be copied and which files must be not (install doesn't seem to handle these, neither original Gentoo port seems to do)~~


Begasus:
LGTM

PR #4991, opened 2020-05-17
Bump pycharm-community and idea-community, add pycharm-professional and idea-ultimate

przemub:
All recipes are nearly identical to pycharm-community and idea-community currently in the repo, all that needed change to update them were the URLs and SHA sums.

This also adds commercial versions of the respective packages. All four work quite well and it's really awesome to have them!


return:
Maybe it could be worth checking if the rest of the commercial JetBrains IDEs (CLion, Rider, WebStorm, RubyMine, GoLand, and PhpStorm) would also work on Haiku.

przemub:
@waddlesplash 

Okay, is there a separate repo/place where these could be put?

Gentoo keeps it in the same place in its ports: https://packages.gentoo.org/packages/dev-util/pycharm-professional while Arch keeps it in AUR: https://aur.archlinux.org/packages/pycharm-professional/

@return It's very probable they would since it's all the same base. I would like to get these approved first though since the recipes would be probably nearly-identical, then I can have a look since I have academic licence for all their products.

waddlesplash:
No, I don't think there is a good place to put these, unfortunately.

It's worth noting that Gentoo does not actually distribute built packages, only the ebuilds, and then the users download/install these themselves. AUR is also basically "uncontrolled", it's not that difficult to get things into.

pulkomandy:
We can store the recipe and mark them broken so they don't land in the repository, right? And we need to implement the EULA agreement in HaikuDepot and pkgman if not done already.

przemub:
@pulkomandy That's what I was about to suggest - or maybe a separate tree for non-repository packages so we don't abuse broken?

In case of these, EULA is accepted or not at the first run so I suppose it would not be necessary in HaikuDepot itself.

waddlesplash:
I still think we should not include software in HaikuPorts that cannot be, at the very least, freely redistributed with no legal concerns.

przemub:
I believe this one can be distributed with no legal concerns since it concurrently functions as a trial version.

waddlesplash:
> (A) You may:
> (i) Use any version of the Product not covered by the Toolbox Subscription on any number of Clients and on any operating system supported by the Product; and
> (ii) Make one backup copy of the Product solely for archival/security backup purposes.
> (B) You may not:
> (i) Rent, lease, reproduce, modify, adapt, create derivative works of, distribute, sell, or transfer the Product;

The EULA you have included quite literally says otherwise. Did you not read it?

pulkomandy:
Yes, we would not be allowed to distribute this. So we can keep the recipes, but disabled, and people will have to download and build things themselves. Or you can ask JetBrains if they'd like to put the HPKG on their website for download.

przemub:
I based that on:
>10. TOOLBOX SUBSCRIPTION TRIAL
> 10.1. Subject to the terms of this Agreement, User is granted a one-time right to install and use a Product covered by the Toolbox Subscription for evaluation purposes without charge for a period of thirty (30) days from the date of such Product installation, unless otherwise specified (the "Evaluation Period"). User's use of a Product during the Evaluation Period shall be limited to the evaluation of the Product for the sole purpose of determining whether the Product meets User's requirements and whether User desires to continue using the Product.

But you're right it says nothing to allow redistribution.

waddlesplash:
This is a legal document, you cannot "assume" anything, it is either spelled out or it isn't. And if it isn't, you can't do it; but in this case, it is spelled out that it is completely prohibited.

I would prefer we not have recipes for this at all, then. You can keep them in a personal repository of yours, if you wish, of course.

przemub:
So I will, but please re-open it since it contains bumps for the open-source parts already included in HP. I'll remove the commercial parts.

przemub:
Removed.

przemub:
I set up an alternative tree for the commercial versions if any one needs these or you get more PRs like this one. https://github.com/przemub/haikuports-nonfree

dev-util/pycharm-community/pycharm_community_bin-2020.1.1.recipe
```diff
@@ -1,18 +1,19 @@
...
-CHECKSUM_SHA256="8eedae360aaf3e0480171d43a12db66f0623b5a7cc1cf239e98e8d9aed272aaa"
+CHECKSUM_SHA256="715a6a1fc1638078a36b5b45e36e3dbe1ce097aa974555023b4b3d957d00585c"
 ADDITIONAL_FILES="PyCharm_Logo.hvif"
 
 ARCHITECTURES="any"
 DISABLE_SOURCE_PACKAGE="true"
 
 PROVIDES="
 	pycharm_community_bin = $portVersion
+	pycharm = $portVersion
```
korli, line 16: Why add this provides?

dev-util/idea-community/idea_community_bin-2020.1.1.recipe
```diff
@@ -2,12 +2,12 @@ SUMMARY="Intellij IDEA Community Edition"
...
 HOMEPAGE="https://www.jetbrains.com/idea/"
-COPYRIGHT="2018 JetBrains s.r.o."
+COPYRIGHT="2020 JetBrains s.r.o."
 LICENSE="Apache v2"
 REVISION="1"
-buildVersion="183.4588.61"
-SOURCE_URI="https://download.jetbrains.com/idea/ideaIC-$portVersion-no-jdk.tar.gz#noarchive"
-CHECKSUM_SHA256="c6bd74e4280c3cf850fd6953a47e0e91ffb882b3b0ca25de12ac745f65a5d3e3"
+buildVersion="201.7223.91"
+SOURCE_URI="https://download.jetbrains.com/idea/ideaIC-$portVersion.tar.gz#noarchive"
```
korli, line 9: Please use the no-jbr.tar.gz

dev-util/pycharm-community/pycharm_community_bin-2020.1.1.recipe
przemub: I wanted to have a common provide for `pycharm_community_bin` and `pycharm_professional_bin`. If it's not a correct usage, I'll remove it.

korli: usually we add "app:pycharm"

volo-droid:
> Yes, we would not be allowed to distribute this. So we can keep the recipes, but disabled, and people will have to download and build things themselves. Or you can ask JetBrains if they'd like to put the HPKG on their website for download.

[JetBrains mentions](https://www.jetbrains.com/help/clion/quick-tutorial-on-configuring-clion-on-macos.html#toolbox-install) CLion installation via `brew cask` on their official site:  
> Alternatively, you can install CLion via the Homebrew package manager : `brew cask install clion`. However, this option is unofficial, as the CLion team is not involved in its maintenance or support.

Am I right that installation via `haikuporter` would be similar to what `brew cask` is doing? (i.e. downloading a binary from the JetBrains site, like defined in [CLion cask](https://github.com/Homebrew/homebrew-cask/blob/499aa2b2b2b7d8dcf120ab3a8ed8c83f44e27e4b/Casks/c/clion.rb) and [IntelliJ IDE Ultimate cask](https://github.com/Homebrew/homebrew-cask/blob/b2f8a601ccd1c7986956bd4ff71def401cedcd8a/Casks/i/intellij-idea.rb)).

If so, can we actually go with @pulkomandy's suggestion and add the ultimate/professional (also potentially CLion) recipes but disable them? That way we won't be distributing HPKG, and users will download binaries directly from the JetBrains site, which they are fine with.

Begasus:
One can stillg grab the Clion archive, extract and run, I guess if you have a license key for it it can be unlocked (as I don't have I can't check), afair it only works for a limited time then.

CLion-2024 1 2

volo-droid:
Yes, that's completely doable, but my idea was to have haikuports keeping track of the official url where to download from (and possibly handling the uninstallation too), in a similar manner to what `homebrew-cask` does.

FWIW core (meaning regular) Haiku and HaikuPorts project committers can get CLion and other licenses for free: 
https://www.jetbrains.com/community/opensource/

From their [Open Source Program FAQ](https://sales.jetbrains.com/hc/en-gb/categories/13706169183250-Free-Licenses-for-OSS-development): 
> If approved, your project will receive an All Products Pack license (or licenses), which grants access to all of our desktop products – IntelliJ IDEA Ultimate, Rider, PhpStorm, and other [JetBrains IDEs](https://www.jetbrains.com/all/).

Begasus:
So far the CE are on par (atleast last time I checked) for NetBeans and PyCharm, I had to search for the Clion archive too, but shouldn't be so hard to find.
There was no install involved here for Clion, so not sure why we would have some uninstalling needed?

Nice link for the licenses link, as I can't even code "Hello World!" I don't really have a need for it, but maybe others would have a use for it.

Anyway, I think this should be moved to an issue then with supplied links for the users. (my 2 cents)

PR #4992, opened 2020-05-18
Add sys-process/parallel.

przemub:
Adds a recipe for GNU Parallel, tool to run stuff in parallel :) https://www.gnu.org/software/parallel/

The patch disables failing feature to detect the parent shell of `parallel`. It relies on `ps` but since `parallel` gets launched in a different team I don't see a way to get the name of the process which launched `parallel` from its Haiku output. Therefore I set it to default to `/bin/bash` - user can change it using environment variables `PARALLEL_SHELL` and `SHELL`, as described in parallel's docs.

### DIFF
--- a/sys-process/parallel/parallel-2020.04.22.recipe
+++ b/sys-process/parallel/parallel-2020.04.22.recipe
@@ -19,24 +19,24 @@ ARCHITECTURES="x86_gcc2 x86 x86_64"
 PROVIDES="
 	parallel = $portVersion
 	cmd:parallel = $portVersion
-	cmd:env_parallel
-	cmd:env_parallel.ash
-	cmd:env_parallel.bash
-	cmd:env_parallel.csh
-	cmd:env_parallel.dash
-	cmd:env_parallel.fish
-	cmd:env_parallel.ksh
-	cmd:env_parallel.mksh
-	cmd:env_parallel.pdksh
-	cmd:env_parallel.sh
-	cmd:env_parallel.tcsh
-	cmd:env_parallel.zsh
-	cmd:niceload
-	cmd:parcat
-	cmd:parset
-	cmd:parsort
-	cmd:sem
-	cmd:sql
+	cmd:env_parallel = $portVersion
+	cmd:env_parallel.ash = $portVersion
+	cmd:env_parallel.bash = $portVersion
+	cmd:env_parallel.csh = $portVersion
+	cmd:env_parallel.dash = $portVersion
+	cmd:env_parallel.fish = $portVersion
+	cmd:env_parallel.ksh = $portVersion
+	cmd:env_parallel.mksh = $portVersion
+	cmd:env_parallel.pdksh = $portVersion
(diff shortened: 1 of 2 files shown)


sys-process/parallel/parallel-2020.04.22.recipe
```diff
@@ -0,0 +1,61 @@
+SUMMARY="A shell tool for executing jobs in parallel locally or on remote machines"
+DESCRIPTION="GNU parallel is a shell tool for executing jobs in parallel using one or more computers. A job can be a single command or a small script that has to be run for each of the lines in the input. The typical input is a list of files, a list of hosts, a list of users, a list of URLs, or a list of tables. A job can also be a command that reads from a pipe. GNU parallel can then split the input and pipe it into commands in parallel.
```
korli, line 2: Please break with \ lines longer than 80 columns

sys-process/parallel/parallel-2020.04.22.recipe
```diff
@@ -0,0 +1,61 @@
...
+If you use xargs and tee today you will find GNU parallel very easy to use as GNU parallel is written to have the same options as xargs. If you write loops in shell, you will find GNU parallel may be able to replace most of the loops and make them run faster by running several jobs in parallel.
+
+GNU parallel makes sure output from the commands is the same output as you would get had you run the commands sequentially. This makes it possible to use output from GNU parallel as input for other programs.
+
+For each line of input GNU parallel will execute command with the line as arguments. If no command is given, the line of input is executed. Several lines will be run in parallel. GNU parallel can often be used as a substitute for xargs or cat | bash."
+HOMEPAGE="https://www.gnu.org/software/parallel/"
+COPYRIGHT="2004-2020 Ole Tango"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="ftp://ftp.gnu.org/gnu/parallel/parallel-20200422.tar.bz2"
```
korli, line 12: Could you check with this suggestion?
```suggestion
SOURCE_URI="ftp://ftp.gnu.org/gnu/parallel/parallel-${portVersion//./}.tar.bz2"
```

sys-process/parallel/parallel-2020.04.22.recipe
```diff
@@ -0,0 +1,61 @@
...
+SOURCE_DIR="parallel-20200422"
```
korli, line 13:
```suggestion
SOURCE_DIR="parallel-${portVersion//./}"
```

sys-process/parallel/parallel-2020.04.22.recipe
przemub: Done.

przemub:
@korli All done. I spent a bit too much time trying to figure out why adding `cmd:awk` to `BUILD_PREREQUIRES` causes HaikuPorter to wait forever for the build package. The solution was to add dots in the port version. Maybe it would be helpful to add a warning to HaikuPorter? Or is it a bug?

PR #5005, opened 2020-05-22
New Recipe: SCOTCH shared library for graph reduction

theclue:
[SCOTCH](http://www.labri.fr/perso/pelegrin/scotch/) is a real classic library in network analysis and would be a great addition in HP for several applications, not necessarly strictly related to scientific research. For example, it's widely used in operating systems development for optimizations of scheduling operations and other topics.

It comes in two flavours:

* SCOTCH
* PT-SCOTCH - this is actually not available on this port due to missing MPI library on Haiky (I'm going to port this too)

**PROBLEMS:**

I and @Begasus have performed a lot of experiments, but we're stucked.

SCOTCH builds ok; also the included tools build ok. **But tests don't**, for two reasons:

* **Non-standard and very weird Makefile target for checks.** This prevents make working when launched from HP. It works when launched from command line, though. This shoud be easy to fix, but we were not able to find the bug.

* **Several tests hang anyway.** No errors, just hang. Bugs could be everywhere. It could be in the test suite, the upstream library source code, or even for some incompatibilities with Haiku (I see a lot of exotic stuff with POSIX threads in the source code). How to go further from here to a more deeper debugging experience?

Perhaps, It could be an option to merge it even without test block working and rely on Haiku users feedback while using the library? But who's going to use the library, in the near future if sdo?


extrowerk:
Will look into it afternoon.

extrowerk:
Adding
```
export LIBRARY_PATH=$LIBRARY_PATH:$sourceDir/lib
```
to the beginning of the BUILD() section helps a bit...

theclue:
> Adding
> 
> ```
> export LIBRARY_PATH=$LIBRARY_PATH:$sourceDir/lib
> ```
> 
> to the beginning of the BUILD() section helps a bit...

done, but I have to say that nothing seemed to change here...

extrowerk:
> > Adding
> > ```
> > export LIBRARY_PATH=$LIBRARY_PATH:$sourceDir/lib
> > ```
> > 
> > 
> > to the beginning of the BUILD() section helps a bit...
> 
> done, but I have to say that nothing seemed to change here...

My mistake, put it into TEST(), obviously.

Begasus:
Seems to be ok with LIBRARY_PATH in TEST?
(link will only stay alive for a month): https://paste.ubuntu.com/p/7J9Wyn2RN2/

sci-libs/scotch/scotch-6.0.9.recipe
```diff
@@ -0,0 +1,226 @@
...
+	# devel package
+	packageEntries devel $developDir
+
+	# tools package
+	packageEntries tools $commandBinDir $manDir
+}
+
+TEST()
+{
+
```
Begasus, line 210: With this in place, the tests run OK here for gcc2h (secondary architecture), see Terminal output in the comment for this PR:
```diff
+	export LIBRARY_PATH=$LIBRARY_PATH:$sourceDir/lib
+
```

theclue: x86_64 is ok too. I see no glitches, now.

FYI tests were hanging due to a bug in test suite (a path in tgt file used to perform the diffs). Incidentally, adding LIBRARY_PATH solves this as well.

I think it could be merged, if it's ok from your side.

Begasus: @extrowerk could you give it a second look also?

Begasus:
LGTM

Issue #2797, opened 2018-07-12
Add recipe for Code::Block


AwlsomeAlex:
Okay, so Code::Block will compile, however it is useless if I can't even get it to install. It might be something wrong to where I'm installing it, something with the File System, or something about the directories not existing. But I'm not that familiar with Haiku OS to know. Could someone assist me?
I executed `make install` and it failed at `/bin/install -c -m 644 SqPlusConst.h (rest of files..) '/usr/local/include/codeblocks/scripting/sqplus'`
![capture](https://user-images.githubusercontent.com/11639639/42611650-26e96178-8565-11e8-9c1a-13f1afec6286.PNG)

waddlesplash:
Yep, we have a completely different directory structure vs. Linux. At this point it may be better if you try to write a recipe, as we have some tricks to take care of the directory differences in HaikuPorter...

AwlsomeAlex:
Stinks Linux is all I know, I'm excited to expand my knowledge. I'll create the patch file and upload it and call it a day. Will probably attempt a recipe this weekend, because I hope to work on my Linux Distribution tomorrow. Might actually break the LSB and model it based off of Haiku in a way with packages, I like it and like being different.

scottmc:
For learning about how haikuporter works you can start at the wiki page "Gentle intro to haikuporter", once you have it up and running and are ready to write your recipe(s), be sure to also read up on our development model.  We take new recipes and patches as pull requests.  Once a pull request is submitted it will be reviewed and corrections/updates/tips will be make and you can further improve on it.  Once it is acceptable we will likely merge it.

https://github.com/haikuports/haikuports/wiki/A-Gentle-Introduction-to-HaikuPorter-%231
https://github.com/haikuports/haikuports/wiki/Development

If you have any questions, feel free to ask here or on IRC #haiku on freenode, there's usually someone helpful there.

extrowerk:
@AwlsomeAlex : it is a standard auto* package, so for testing you can reconfigure it like:
```configure --prefix=~/Desktop/codeblocks_test```
If you used extra configure switches you have to add them too.
Let it configure, rebuild it as you did before and use ```make install```. It will install the complete package into the ```~/Desktop/codeblocks_test``` folder.

From now you can start it. Be warned, if there is a ```lib``` folder in the resulting folder, you have to copy/symlink it to the folder, where the binary located (```./lib``` is always in LDPATH).

But yeah, a proper recipe would be nice.

[...]

PR #5009, opened 2020-05-23
codeblocks: new recipe

return:
This is an initial work in progress recipe for CodeBlocks and addresses #2797 and I have disabled this recipe by default (Needs some improvements to wxqt first.)

First it soft-crashes after the splash screen on a missing implementation in the wxqt port.

screenshot-23

If you skip and continue running CodeBlocks, the IDE will show up. But closing down the editor will hard crash the app (wxqt related). See: https://gist.github.com/return/e58b55c65300b88d8ee13709ca9857f4 for the report.

screenshot2

Mark this as WIP.

### DIFF
--- a/dev-util/codeblocks/codeblocks-20.03.recipe
+++ b/dev-util/codeblocks/codeblocks-20.03.recipe
@@ -15,8 +15,8 @@ CHECKSUM_SHA256="15eeb3e28aea054e1f38b0c7f4671b4d4d1116fd05f63c07aa95a91db89eaac
 SOURCE_DIR="codeblocks-$portVersion"
 PATCHES="codeblocks-$portVersion.patchset"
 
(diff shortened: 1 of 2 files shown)


Begasus:
Would be nice if you could fix the issue for wxqt :+1:

waddlesplash:
The report is strange; I'm not sure how that function gets to abort(): https://github.com/wxWidgets/wxWidgets/blob/0ca45d1a59a51e6798b99eaa0e7fdbbcc33c3992/src/unix/utilsunx.cpp#L369

Begasus:
Maybe could be looked into again now  we also have wxgtk?

Begasus:
No progress so far, merging as disabled, someone who can tackle this can continue work based on this one.

PR #5011, opened 2020-05-23
glm, bump version

Begasus:


Begasus:
From a build in Terminal I got this (with cmake and gcc2), so I disabled gcc2_x86 for now, comments welcome:
```
/Opslag/haikuports/media-libs/glm/work-x86-0.9.9.8/sources/glm-0.9.9.8/glm/../glm/detail/../simd/platform.h:222: #error "GLM requires GCC 4.6 or higher"
```

Begasus:
PS, this is needed to bump the version for Solarus ...

korli:
lib:libGL$secondaryArchSuffix should be removed.

korli:
did you try to build libetonyek on x86 secondary arch?

Begasus:
Not yet, busy atm with imagemagick build and checkup on RC for new TuxPaint, will do in a bit
There are a few candidates IIRC that uses glm${secondaryArchSuffix}_devel atm, need to investigate those too (although I don't think it would break existing builds as they are only used at build time?)

Begasus:
> did you try to build libetonyek on x86 secondary arch?

I build the latest solarus_x86 with this though :)

Begasus:
Will revbump the others with the changes in a bit :)

PR #5016, opened 2020-05-24
Cascadia Code: Bump revision

bitigchi:
Forgot to bump revision, sorry!


diversys:
You don't need to do that when you bump version.
https://build.haiku-os.org/buildmaster/master/x86_64/?buildrunDir=4451&viewMode=expanded

bitigchi:
Shouldn't I have to do it for the name change I made though? In the Depot, it's still `cascadia`.

PR #5022, opened 2020-05-24
LibB64: new recipe

extrowerk:


Begasus:
Builds ok on gcc2 and gcc8

dev-libs/libb64/patches/libb64-20170101.patchset
```diff
@@ -0,0 +1,22 @@
...
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index d51293f..3f80228 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -3,7 +3,7 @@ project(b64 C)
+ 
+ include_directories(include)
+ 
+-add_library(${PROJECT_NAME} STATIC
++add_library(${PROJECT_NAME} SHARED
```
waddlesplash, line 16: If they didn't intend this be shipped as a shared library, maybe it shouldn't be. What do other distros do?

extrowerk: I was unable to build Transmission with the in-tree libs, and 3.00 tarball doesn't contains these libs (i was later informed that one supposed to use the .xz release for that).
I had the idea maybe other ports can also profit if we build them as shared, because B64 and DHT used in other software too.
The libs are updated separately from transmission, with shared libs we can integrate the changes from upstream more easily.
Somebody have to decide.

waddlesplash: Well, again, what do other distros do? If they ship these as shared, then OK; but I'd prefer we not diverge too far here.

Begasus: https://repology.org/project/libutp/versions
https://repology.org/project/libb64/versions

Begasus: On a quick look atleast ARCH is building a shared lib for this one too? https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/libb64

extrowerk:
Not needed.

PR #5023, opened 2020-05-24
Transmission 3.00

extrowerk:


extrowerk:
Seems to be working ok on 64 bit.

Begasus:
Quick check reveals conflicting types for int also here

extrowerk:
> Quick check reveals conflicting types for int also here

Can you please post the complete error?

extrowerk:
This should be ok now.

PR #5024, opened 2020-05-25
glm, bump revisions on packages relying on glm after update

Begasus:


Begasus:
Didn't include mame in this one as the build failed on my 32bit install, maybe someone could investigate this on 64bit?
Error part on build (stays up for about a month): https://paste.ubuntu.com/p/7BSjc9Jvh9/

PR #5025, opened 2020-05-25
[New Recipe] Acme Cross-Assembler for 65**

theclue:
ACME is a free cross assembler that can produce code for the following processors: \
6502, 6510 (including illegal opcodes), 65c02 and 65816.

I've ported the assembler and the migration tool. I've added some tests using the provided code examples, too.

Source code is packaged very badly. I had no choice other than binding to a particular commit (r190)

tested on x86_64, but the code looks like to be completely portable.

### DIFF
--- a/dev-lang/acme/acme-0.96.5.recipe
+++ b/dev-lang/acme/acme-0.96.5.recipe
@@ -62,7 +62,7 @@ INSTALL()
 	# copy documentation
 	mkdir -p $docDir/acme
 	mkdir -p $docDir/toacme
-	
+
 	cp -rd docs/* $docDir/acme
 	cp -rd contrib/toacme/docs/* $docDir/toacme
 	cp -rd examples $docDir/
--- a/dev-lang/acme/acme-0.96.5.recipe
+++ b/dev-lang/acme/acme-0.96.5.recipe
@@ -38,6 +38,7 @@ BUILD_REQUIRES="
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
 	cmd:make
+	cmd:sed
 	"
 
 TEST_REQUIRES="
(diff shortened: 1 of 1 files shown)


dev-lang/acme/acme-0.96.5.recipe
```diff
@@ -0,0 +1,86 @@
...
+	cmd:make
+	"
+
+TEST_REQUIRES="
+	cmd:test
+	"
+
+BUILD()
+{
+	sed -i "s|/usr/local/bin|$commandBinDir|g" src/Makefile
```
pulkomandy, line 49: this should be done in the PATCH() block rather than in BUILD()

dev-lang/acme/acme-0.96.5.recipe
```diff
@@ -0,0 +1,86 @@
...
+	cp -rd examples $docDir/
+
+	# copy the ACME_Lib
+	mkdir -p $dataDir
+	cp -rd ACME_Lib $dataDir
+}
+
+TEST()
+{
+	sed -i "s| acme| $sourceDir/src/acme|g" examples/Makefile
```
pulkomandy, line 77: this should be done in the PATCH() block

dev-lang/acme/acme-0.96.5.recipe
theclue: Done

dev-lang/acme/acme-0.96.5.recipe
theclue: this won't work in the PATCH() block as it is a reference to $sourceDir I think.

pulkomandy: why would that be a problem?

theclue: well, i haven't investigated deeper yet, but when that line is moved to the PATCH() block, the acme file is not found anymore from the Makefile in examples/ (there should be a path issue or something like that).

PR #5027, opened 2020-05-26
new recipe: sidreloc

theclue:
Sidreloc is a tool for relocating SID tunes and it is widely used by the community of C64 developers and demo makers out there. Not that many of them use Haiku, but who knows...

Additions to the original source code:

- Autotools support (as a learning excercise)
- ~~Two rough equivalent functions to err and errx functions which don't exist in BeOS/Haiku~~ **UPDATE:** yeah, they are available on Haiku ofc


pulkomandy:
err and errx do exist in Haiku, you just have to link with -lbsd

theclue:
oh...damn -_-

ok, fixing...

theclue:
> err and errx do exist in Haiku, you just have to link with -lbsd

Err...i feel a bit blonde now. I assumed there were not available, since they weren't on BeOS, at times.

ok, patched to link and use libbsd when available.

Just a note: I still have these warnings:

```
 warning: implicit declaration of function 'err'
 warning: implicit declaration of function 'errx'
```

but the library was added to the linker and err.h was included. In facts, the program works nicely. But I cannot understand the reason of that errors...prototypes mismatching maybe?

pulkomandy:
You have requested the c99 standard in your compiler flag. This disables any BSD extensions by default.
You have two options:
- use gnu99 instead, this will enable extensions by default
- #define _BSD_SOURCE (or -D_BSD_SOURCE in compiler flags)

dev-util/sidreloc/sidreloc-1.0.recipe
```diff
@@ -0,0 +1,63 @@
...
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	sidreloc$secondaryArchSuffix = $portVersion
+	cmd:sidreloc$commandSuffix = $portVersion
```
pulkomandy, line 31: this should be cmd:sidreloc since you put it in bin even when building ofr secondary architecture

theclue: I'm not sure I undestood what you mean. I'm copying into $commandBinDir. Perhaps I did it in the wrong way?

`runConfigure --omit-dirs binDir ./configure --bindir="$commandBinDir"`

theclue:
> * #define _BSD_SOURCE (or -D_BSD_SOURCE in compiler flags)

Done in congigure.ac + Makefile.am

dev-util/sidreloc/sidreloc-1.0.recipe
pulkomandy: Well $commandSuffix will be empty in all cases. But I see that other recipes do it the same way, so...

PR #5033, opened 2020-05-30
64tass: new recipe

theclue:
64tass is cross assembler targeting the 65xx series of micro processors which features the familiar syntax of Omicron TASS and TASM, one of the most popular assembler on C64 during the 80s and 90s.

Port is working perfectly, but has two small defects:

- Licensing is a bit confused. Main package is LGPL, but an embedded component has a custom one (provided) which I'm not sure is compatible with LGPL itself. No real names are provided (but demo scener/crew nicknames)

- In the INSTALL() block I copy the syntax file to vim settings dir. A better version check should be made here to proof which version of vim is installed. Have I to use a post install script?

### DIFF
--- a/dev-lang/64tass/64tass-1.55.recipe
+++ b/dev-lang/64tass/64tass-1.55.recipe
@@ -1,0 +1,64 @@
+SUMMARY="Multi pass optimizing macro assembler for the 65xx series of processors"
+DESCRIPTION="64tass is cross assembler targeting the 65xx series of micro processors \
+which features the familiar syntax of Omicron TASS and TASM"
+HOMEPAGE="https://sourceforge.net/projects/tass64/"
+COPYRIGHT="2013-2020 Soci/Singular
+	1997-2001 Benjamin Sittler"
+LICENSE="GNU LGPL v2.1
+	my_getopt"
+REVISION="1"
+SOURCE_URI="https://github.com/irmen/64tass/archive/v$portVersion.2202.tar.gz"
+CHECKSUM_SHA256="c9222531a711cf7f23fe0294dd953c2a7cda37f84070fdaeaf8c83a7284cc582"
+SOURCE_DIR="64tass-$portVersion.2202"
+
+ARCHITECTURES="!x86_gcc2 x86_64 ?arm ?ppc ?sparc"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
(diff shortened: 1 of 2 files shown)


dev-lang/64tass/64tass-1.55.recipe
```diff
@@ -0,0 +1,68 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:gzip
+	cmd:make
+	cmd:svnversion
```
korli, line 40: svnversion shouldn't be needed.

theclue: If i understood correctly, I think it is instead. It is used to check for revision number inside the makefile. It has no effect actually (let alone I get the source from a git mirror) but will break the build if absent.

korli: The makefile has to be fixed then.

dev-lang/64tass/64tass-1.55.recipe
```diff
@@ -0,0 +1,67 @@
...
+}
+
+INSTALL()
+{
+	make install $jobArgs
+
+	# vim syntax
+	# TODO: better version checking, still ok until
+	# we are on the 8.2 branch or we repackage vim itself
+	mkdir -p $dataDir/vim/vim82/syntax
```
bitigchi, line 65: For user syntax files, user runtime configuration should be used. Vim data directory is not supposed to be user modified. In this case correct path is `~/config/settings/vim/syntax`.

pulkomandy: This is not "user", it's a system-packaged files. This is how they are handled in Haikuports and in other distros (for example youcompleteme installs itself this way).

theclue: The intense debate on the official forum on [this thread](https://discuss.haiku-os.org/t/where-to-put-vim-and-other-editors-syntax-files/9595) made clear that there is no a common way on how to handle this aside that repackaging vim itself (which is partly resolutive anyway).

Begasus:
Any news on this?

theclue:
> Any news on this?

At the end of the day, there's no common agreement on how to handle this. Thus, I think that unless someone is going to repackage vim (and there's no agreement on this, too), this way is the best way to package the syntax file (and it's the way other packages do). IMHO, it could be merged

Begasus:
OK? If there is something in need of change for this it will show up I guess :)

PR #5042, opened 2020-05-31
Bumped streamradio to version 0.0.5-1

jsteinaker:
Bumped StreamRadio version to 0.0.5-1. The name of the github repo was changed so the recipe reflects this. Also, catkeys are now added to the final binary, so we can have translations (only Spanish for the moment)


haiku-apps/streamradio/streamradio-0.0.5_1.recipe
```diff
@@ -5,10 +5,10 @@ HOMEPAGE="https://github.com/fishpond-haiku/Haiku-Radio"
 COPYRIGHT="2017 Kai Niessen"
 LICENSE="MIT"
 REVISION="4"
```
diversys, line 7: Please reset the revision.

jsteinaker: Done. Sorry, I thought that field was referred to the license revision. My bad.

PR #5059, opened 2020-06-03
yab-compress.c added and missing directory

lorglas:


haiku-apps/yab_buildfactory/additional-files/yab_buildfactory.sh
```diff
@@ -1,7 +1,8 @@
 #!/bin/sh
 mkdir -p /boot/home/yab_work/BuildFactory
-
+mkdir -p /boot/home/config/settings/yab
 cp -R /boot/system/apps/yab-buildfactory/* /boot/home/yab_work/BuildFactory
 chmod 777 /boot/home/yab_work/BuildFactory/BeSly_Buildfactory.yab
 chmod 777 /boot/home/yab_work/BuildFactory/BuildFactory.yab
 chmod -R 777 /boot/home/yab_work/BuildFactory/compress
+chmod 777 /boot/home/config/settings/yab
```
Begasus, line 8: Don't think it's needed here? ~/config/settings should be writable?

lorglas: yes, but i don't want the next commit. i can't write on this folder

Begasus: Even without "chmod"? Creating it should be enough I "think" :)

PR #5064, opened 2020-06-04
Aranym: bump to 1.1.0

theclue:
Works ok, although this recipe should deserve more love

- Enabled UAE FPU emulation. IEEE emulation doesn't work (@pulkomandy do u know something about that?), but UAE is fine.
- Some minor performance tweaks
- JIT not enabled. This would greatly speed up the emulation, but i don't know if  this works under Haiku actually

All the most relevant optional features (ethernet bridging, disk sharing, etc.) still disabled.

### DIFF
--- a/app-emulation/aranym/aranym-1.1.0.recipe
+++ b/app-emulation/aranym/aranym-1.1.0.recipe
@@ -11,7 +11,7 @@ SOURCE_FILENAME="anarym-$portVersion.tar.gz"
 SOURCE_DIR="aranym-ARANYM_${portVersion//./_}"
 PATCHES="aranym-$portVersion.patchset"
 
-ARCHITECTURES="x86_gcc2 x86_64"
+ARCHITECTURES="!x86_gcc2 x86_64"
 
 PROVIDES="
 	aranym = $portVersion
--- a/app-emulation/aranym/aranym-1.1.0.recipe
+++ b/app-emulation/aranym/aranym-1.1.0.recipe
@@ -46,7 +46,7 @@ BUILD()
 	
 	export CXXFLAGS="$CXXFLAGS -Wno-unused-function -Wno-unused-variable"
 	export CFLAGS="$CFLAGS -Wno-unused-function -Wno-unused-variable"
-	runConfigure ./configure --enable-fpe=uae $WITHDISDIP
+	runConfigure ./configure --enable-fpe=uae --enable-cxx-exceptions $WITHDISDIP
 	make $jobArgs
 }
 
@@ -56,6 +56,6 @@ INSTALL()
 
 	make install
 
-	mimeset /bin/aranym
+	mimeset $prefix/bin/aranym
 	addAppDeskbarSymlink /bin/aranym ARAnyM
 }
--- a/app-emulation/aranym/aranym-1.1.0.recipe
+++ b/app-emulation/aranym/aranym-1.1.0.recipe
@@ -55,9 +55,7 @@ INSTALL()
 	mkdir -p $docDir
 
 	make install
-	#mv $prefix/share/doc/aranym/* $docDir
-	#rm -r $prefix/share
 
-	mimeset $prefix/bin/aranym
+	mimeset /bin/aranym
 	addAppDeskbarSymlink /bin/aranym ARAnyM
 }
--- a/app-emulation/aranym/aranym-1.1.0.recipe
+++ b/app-emulation/aranym/aranym-1.1.0.recipe
@@ -5,12 +5,12 @@ HOMEPAGE="https://aranym.github.io/"
 COPYRIGHT="2001-2014 ARAnyM developer team"
 LICENSE="GNU GPL v2"
 REVISION="1"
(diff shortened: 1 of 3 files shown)


pulkomandy:
I don't see why JIT wouldn't work? There is nothing "magic" about it. We use the same CPU as others systems so we can run the same code.

I have no idea what IEEE emulation is and I ran Aranym maybe once a few years ago, so I can't really help there. @mmuman probably knows better as he uses Aranym for the NeXT port of Haiku (surprisingly) and a few other things.

theclue:
ok, good to have those details. UAE works fine, actually (I was just curious)

About JIT, enabling it triggers this error in configure, so I've kept it disabled for now

`configure: error: Sorry, extended segfault handler not supported on your platform
`

I don't know if this is something that could be fixed in configure.ac or an effective missing feature.

Begasus:
Looking at some of the issues there, disabling/not using JIT wouldn't be so bad for now?
For gentoo they also disable it? https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/aranym/aranym-1.1.0.ebuild#n64

theclue:
> Looking at some of the issues there, disabling/not using JIT wouldn't be so bad for now?
> For gentoo they also disable it? https://gitweb.gentoo.org/repo/gentoo.git/tree/app-emulation/aranym/aranym-1.1.0.ebuild#n64

It is optional...but...

...I agree on keeping jit disabled until further investigation. I decided to enable it essentially because it's kind of a challenge to me (but not enough to keep the recipe outdated)

app-emulation/aranym/aranym-1.1.0.recipe
```diff
@@ -5,12 +5,12 @@ HOMEPAGE="https://aranym.github.io/"
...
-SOURCE_URI="https://github.com/aranym/aranym/archive/ARANYM_1_0_2.tar.gz"
-CHECKSUM_SHA256="c0935aee14cca55c08e75a94b77dcd69f326ae563ae371b22c7b9d8efd82e24e"
+SOURCE_URI="https://github.com/aranym/aranym/archive/ARANYM_${portVersion//./_}.tar.gz"
+CHECKSUM_SHA256="1636cf9ec68bd2089f537e3b5e82d04fafec61a4f8fb5e216a263d357e935b63"
 SOURCE_FILENAME="anarym-$portVersion.tar.gz"
-SOURCE_DIR="aranym-ARANYM_1_0_2"
+SOURCE_DIR="aranym-ARANYM_${portVersion//./_}"
 
-ARCHITECTURES="x86_gcc2 ?x86 ?x86_64"
+ARCHITECTURES="x86_gcc2 x86 x86_64"
```
Begasus, line 13: As there is no x86 primary architecture we should see to remove it from the recipes:
```diff
- ARCHITECTURES="x86_gcc2 x86 x86_64"
+ ARCHITECTURES="x86_gcc2 x86_64"
```

theclue:
Just found that Aranym writes its config file in `.aranym`. Going to write a small patch to write it in ./config. Pls do not merge yet.

theclue:
Here we go, I think it could be merged now

theclue:
Pls remove the WIP label from the recipe ;) tnx!

Begasus:
Didn't have a chance to check the changes, will hold on until I could check (or someone else could merge if it's OK) :)

Begasus:
> ok, good to have those details. UAE works fine, actually (I was just curious)
> 
> About JIT, enabling it triggers this error in configure, so I've kept it disabled for now
> 
> `configure: error: Sorry, extended segfault handler not supported on your platform `
> 
> I don't know if this is something that could be fixed in configure.ac or an effective missing feature.

Not sure if it's related, but tried using libsigsegv?

Begasus:
Build fais on gcc2 with "cc1: Invalid option `-Wno-unused-function'" ...

app-emulation/aranym/aranym-1.1.0.recipe
```diff
@@ -37,8 +38,16 @@ BUILD_PREREQUIRES="
...
+	WITHDISDIP=""
+	if [ $effectiveTargetArchitecture == x86_gcc2 ]; then
+		WITHDISDIP="--enable-disdip"
+	fi
+	
 	NO_CONFIGURE=1 ./autogen.sh
-	runConfigure ./configure
+	
+	export CXXFLAGS="$CXXFLAGS -Wno-unused-function -Wno-unused-variable"
+	export CFLAGS="$CFLAGS -Wno-unused-function -Wno-unused-variable"
```
Begasus, line 49: I was looking inside the code where it got initialized just to find out it was here ;)
These are not valid for gcc2, without these the build still fails for gcc2

Begasus:
Still no luck on getting this build on 32 bit Haiku, conflicting types, that could be fixed, but run into other int types later in the build process, builds OK on 64 bit (checked with this release and a commit checkout) ...

korli:
@Begasus maybe this could be merged only with x86_64 enabled, and the old recipe resurrected?

Begasus:
> @Begasus maybe this could be merged only with x86_64 enabled, and the old recipe resurrected?

@korli I'm ok with it, didn't (don't) have time to fully investigate further

PR #5073, opened 2020-06-05
assimp: bump to 5.0.1

theclue:
Changelog:

- Bump to 5.0.1, builds fine
- Added a TEST() block with provided unit tests; all tests passed
- Added doxygen build to $developDocDir

### DIFF
--- a/media-libs/assimp/assimp-5.0.1.recipe
+++ b/media-libs/assimp/assimp-5.0.1.recipe
@@ -35,7 +35,7 @@ REQUIRES="
 PROVIDES_devel="
 	assimp${secondaryArchSuffix}_devel = $portVersion
 	devel:libassimp$secondaryArchSuffix = $libVersionCompat
-	devel:libirrxml$secondaryArchSuffix = $libVersionCompat
+	devel:libirrxml$secondaryArchSuffix = $libVersion
 	"
 REQUIRES_devel="
 	assimp$secondaryArchSuffix == $portVersion base
@@ -79,8 +79,6 @@ BUILD()
 
 INSTALL()
 {
-	mkdir -p $developDocDir
-	
 	make -C build install
 
 	prepareInstalledDevelLibs libassimp libIrrXML
--- a/media-libs/assimp/assimp-5.0.1.recipe
+++ b/media-libs/assimp/assimp-5.0.1.recipe
@@ -3,12 +3,13 @@ DESCRIPTION="Open Asset Import Library (short name: Assimp) is a portable \
 Open Source library to import various well-known 3D model formats in a \
 uniform manner. The most recent version also knows how to export 3d files \
 and is therefore suitable as a general-purpose 3D model converter."
-HOMEPAGE="https://github.com/assimp/assimp"
-COPYRIGHT="2006-2017, assimp team"
+HOMEPAGE="https://assimp.org"
+COPYRIGHT="2006-2020, assimp team"
 LICENSE="BSD (3-clause)"
 REVISION="1"
 SOURCE_URI="https://github.com/assimp/assimp/archive/v$portVersion.tar.gz"
-CHECKSUM_SHA256="b0110a91650d6bb4000e3d5c2185bf77b0ff0a2e7a284bc2c4af81b33988b63c"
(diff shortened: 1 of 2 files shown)


media-libs/assimp/assimp-5.0.1.recipe
```diff
@@ -59,19 +63,24 @@ defineDebugInfoPackage assimp$secondaryArchSuffix \
 BUILD()
 {
 	mkdir -p build
-	cd build
```
Begasus, line 62: This won't work if you use cmake .. after this without switching to build

theclue: make **-C build** $jobArgs won't work either? ?_?

Begasus: You are calling that after **cmake * ..** :)

theclue: oh s*it! -_- going to fix immediately

Begasus:
Haven't checked yet, but could you add a defineDebugInfoPackage here?

media-libs/assimp/assimp-5.0.1.recipe
```diff
@@ -60,18 +64,24 @@ BUILD()
...
+		-DCMAKE_INSTALL_DOCDIR=$developDocDir \
 		..
 	make $jobArgs
 }
 
 INSTALL()
 {
-	cd build
-	make install
+	mkdir -p $developDocDir
```
Begasus, line 82: Not needed:
```diff
- 	mkdir -p $developDocDir
```

Begasus:
devel:libirrxml$secondaryArchSuffix = $libVersionCompat
there is no shared library for this, only a *.a library, do we need $libVersionCompat for this? (or maybe a way to create a shared one?)

theclue:
> devel:libirrxml$secondaryArchSuffix = $libVersionCompat
> there is no shared library for this, only a *.a library, do we need $libVersionCompat for this? (or maybe a way to create a shared one?)

I don't know how to creare a shared library from a static one :(

Begasus:
> > devel:libirrxml$secondaryArchSuffix = $libVersionCompat
> > there is no shared library for this, only a *.a library, do we need $libVersionCompat for this? (or maybe a way to create a shared one?)
> 
> I don't know how to create a shared library from a static one :(

Checked cmake options for this build and couldn't find a way to set some variable for the static lib, so I guess it's ok to leave it as is?

Begasus:
Didn't check this, but maybe you could create something from here? :) https://www.ambiera.com/irrxml/

pulkomandy:
If upstream doesn't want to create a shared lib, I don't see why we would force them to...
Just leave things as they are.

theclue:
> If upstream doesn't want to create a shared lib, I don't see why we would force them to...
> Just leave things as they are.

This is something that **sometimes** Gentoo ports do _but_....

...it looks really overkill to me here anyway

Begasus:
Looking at repology, there aren't even a lot of OS's that support the separate irrxml library: https://repology.org/project/irrxml/versions
So guess it's ok as it is atm, had a look at some other sources for irrxml and managed to build a shared library, but as said, probably overkill, so we can leave it as is :)

theclue:
I'll make a recupe for irrxml.so then...as I said...I like the challenges :)

bu i think that for the moment, this recipe could be merged as-is

Begasus:
> > If upstream doesn't want to create a shared lib, I don't see why we would force them to...
> > Just leave things as they are.
> 
> This is something that **sometimes** Gentoo ports do _but_....
> 
> ...it looks really overkill to me here anyway.
> 
> Said that, the library is part of the irrlicht project, an open source 3g gaming engine written in c++ and similar to the Unreal Engine (ok, a heavily stripped down version of the Unreal Engine). This would be great to have in Haiku, instead. Although I don't think it's gonna be easy to port, though.

We already have irrlicht in our depo :) https://github.com/haikuports/haikuports/blob/f0399cc720dbae651845a82152b8e2a83a5aea56/dev-games/irrlicht/irrlicht-1.8.4.recipe

If you could check if this builds with the system lib for irrxml (and make changes in case it does) we could merge irrxml before we merge this one using the "system lib" (there is an option for it) :)

PR #5104, opened 2020-06-10
premake5: bump to alpha15

theclue:
Business as usual, here. All the needed Haiku stuff is already in upstream.

I've just cleaned the port versioning scheme a bit and tied to a git release instead of a commit sha.


Begasus:
Isn't there anything still using premake4?

theclue:
> Isn't there anything still using premake4?

you're right. I restored the 4.4 recipe file in the PR.

(side question: how do an user can choose to install an older version of a given recipe, if available?)

Begasus:
> > Isn't there anything still using premake4?
> 
> you're right. I restored the 4.4 recipe file in the PR.
> 
> (side question: how do an user can choose to install an older version of a given recipe, if available?)

You can choose to use **cmd:premake4$secondaryArchSuffix** or **cmd:premake5$secondaryArchSuffix** for the ones needing a specific version

PR #5127, opened 2020-06-19
php: update to 7.4.3

X547:
Composer and Laravel 7.16.1 is working.

### DIFF
--- a/dev-lang/php/php-7.4.3.recipe
+++ b/dev-lang/php/php-7.4.3.recipe
@@ -1,0 +1,120 @@
+SUMMARY="A popular general-purpose scripting language"
+DESCRIPTION="PHP is a popular general-purpose scripting language that is \
+especially suited to web development.
+Fast, flexible and pragmatic, PHP powers everything from your blog to the most \
+popular websites in the world."
+HOMEPAGE="https://php.net/"
+COPYRIGHT="2001-2020 PHP Group"
+LICENSE="PHP License v3.01"
+REVISION="1"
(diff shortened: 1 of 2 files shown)


dev-lang/php/php-7.4.3.recipe
```diff
@@ -0,0 +1,120 @@
...
+SOURCE_URI="https://php.net/get/php-$portVersion.tar.bz2/from/this/mirror"
+CHECKSUM_SHA256="c1517ba49578fb2dcc64c73a3edc76d4fc507c4a7ac639981584cc7d3b4c6d14"
+SOURCE_FILENAME="php-$portVersion.tar.bz2"
+PATCHES="php-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	php = $portVersion
```
waddlesplash, line 19: This should be php$secondaryArchSuffix. However, you can install these commands in $prefix/bin and then drop the $secondaryArchSuffix on them, so then `pkgman install cmd:php` will work.

kenmays:
Fails on hrev54373 x86_64:
sapi/cgi/php-cgi - 'Undefined referencing...'
Some of this seems ICU/Unicode related (ICU-57.2 used, needs a higher version (ICU-62?!?)).

waddlesplash:
https://github.com/haikuports/haikuports/commit/212b723d149054fc5905930814d5285a64241171

PR #5128, opened 2020-06-20
swig, bump version

Begasus:


Begasus:
Added python3 and perl in BUILD_PREREQUIRES, looking at configure I'm guessing some other languages could be added also (eg ocaml, lua, tlc ...)
Would it be worth to look into those @return ?

PR #5142, opened 2020-06-28
nano: update version

bitigchi:


app-editors/nano/nano-4.9.3.recipe
```diff
@@ -39,7 +41,6 @@ BUILD_REQUIRES="
 	devel:libz
 	"
 BUILD_PREREQUIRES="
-	cmd:autoreconf
```
Begasus, line 42: Haven't checked this, but if this works without autoreconf I guess you could leave out libtoolize also
msgfmt and msgmerge are provided by the libintl package so I guess they shouldn't be needed here also

Begasus:
No test suite anymore? Also, as this can't be build for gcc2, could you see if you could add SECONDARY_ARCHITECTURE for this so it would be available at 32bit Haiku also?

bitigchi:
@Begasus, this patch is not mine, I'm just conveying it for a nano developer whom doesn't use Github. See commit message and e-mail.

I've restored the recipe for the 2.x branch, granted I didn't check before deleting it. :/

app-editors/nano/nano-4.9.3.recipe
bitigchi: Fails to build without those three.

Begasus:
LGTM, would be nice if some of the other contributors could check this out also :)

PR #5150, opened 2020-07-02
spotify-qt, new recipe

Begasus:


Begasus:
wip, done as part of forum post: https://discuss.haiku-os.org/t/tip-how-to-get-spotify-qt-running/9746/11

media-sound/spotify_qt/spotify_qt-2.4.recipe
```diff
@@ -0,0 +1,75 @@
...
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:qdbuscpp2xml$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	mkdir -p build && cd build
+	cmake .. -DCMAKE_INSTALL_PREFIX=/appsDir
```
diversys, line 54: $appsDir probably?

Begasus: Actually even that would be incorrect as "make install" doesn't work, but probably should refer to $prefix/bin rather then $appsDir :) (but yes a typo) :)

Begasus:
Think this should do it, not checked on x86_64 so far, would be nice if someone could check a build there :)

diversys:
Builds fine on x86_64, but the binary should be moved to $appsDir.

PR #5183, opened 2020-07-10
mednafen_psx_libretro: bump to git version

kwyxz:
Had to deactivate LightRec for now, will spend more time trying to make it work for a nice performance boost


kwyxz:
Looks like LightRec build fails because the linker cannot find `librt`, if anybody knows the Haiku equivalent, that'd help immensely

pulkomandy:
I think most of librt is inside libroot in haiku? Any missing symbols if you omit this lib?

Begasus:
> Looks like LightRec build fails because the linker cannot find `librt`, if anybody knows the Haiku equivalent, that'd help immensely

Looking at the Makefile, it also checks for shm (which afaik we don't have?
https://github.com/libretro/beetle-psx-libretro/blob/master/Makefile#L108

kwyxz:
If symbols are in libroot then I can try just removing the -lrt flag and see what happens. If it works I'll submit a PR upstream. Let's see.

kwyxz:
It works. I'll be submitting the PR upstream and will create a new one here once my changes are merged. Thanks!

Screenshot_Haiku_2020-07-11_08:53:36

PR #5185, opened 2020-07-11
nspr, fix build

Begasus:


Begasus:
@kenmays could you check 64bit?

Issue #5190, opened 2020-07-12
Kdevelop missing libclang_x86 >=5 library when installing from HaikuDepot


Kitsune64:
Kdevelop missing libclang_x86 >=5.0.0 library when installing from HaikuDepôt on a x86 Haiku system.

PR #5191, opened 2020-07-12
kdevelop: trigger build

kenmays:
Verified build on hrev54415 x86_64.
Closes: https://github.com/haikuports/haikuports/issues/5190

### DIFF
--- a/dev-util/kdevelop/kdevelop-5.4.6.recipe
+++ b/dev-util/kdevelop/kdevelop-5.4.6.recipe
@@ -7,7 +7,7 @@ HOMEPAGE="https://www.kdevelop.org/"
 COPYRIGHT="2010-2019 KDE Organisation"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
-REVISION="2"
+REVISION="1"
 SOURCE_URI="https://github.com/KDE/kdevelop/archive/v$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


dev-util/kdevelop/kdevelop-5.4.6.recipe
```diff
@@ -7,7 +7,7 @@ HOMEPAGE="https://www.kdevelop.org/"
 COPYRIGHT="2010-2019 KDE Organisation"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
-REVISION="2"
+REVISION="1"
```
Begasus, line 10: Shouldn't this be revision 3?

extrowerk: Awww...

Kitsune64:
I don't understand why you closed my Issue. That was on x86 system (32bits) about a missing library when installing by Haiku Depôt. It don't work for now...There is a way to make it work?

kenmays:
@Kitsune64 0 Kdevelop was bumped to 5.4.6, so that is the current supported build., We are phasing out porting support for the older LLVM 5.x build.

Kitsune64:
> Kdevelop was bumped to 5.4.6, so that is the current supported build., We are phasing out porting support for the older LLVM 5.x build

Okay so here I have here the version available 5.4.1-1 show in the list of Haiku Depôt after a Haiku update: R1/beta2 hrev54154+116 x86_gcc2.

PR #5199, opened 2020-07-15
paladin: remove old recipes

kenmays:


kenmays:
NOTE: Working 2.7 recipe on x86 and x86_64. Keep recipes for now per Begasus.

PR #5206, opened 2020-07-17
libavif: new recipe for version 0.8.0

linkmauve:
Only decoding is supported for now, as there is no package for [rav1e](https://github.com/xiph/rav1e) yet, because cargo-c fails to build due to https://github.com/alexcrichton/socket2-rs/commit/43b75d3dd4a90e33b4b9ddb764f84f68914c9648 not being in a release yet.

Once rav1e is in haikuports, it will only be a matter of passing `-DAVIF_CODEC_RAV1E=ON` to cmake, and encoding will be possible.

### DIFF
--- a/media-libs/libavif/libavif-0.8.0.recipe
+++ b/media-libs/libavif/libavif-0.8.0.recipe
@@ -1,0 +1,87 @@
+SUMMARY="Library for encoding and decoding .avif files"
+DESCRIPTION="This library aims to be a friendly, portable C implementation of \
+the AV1 Image File Format, as described here: \
+https://aomediacodec.github.io/av1-avif/
+It is a work-in-progress, but can already encode and decode all AOM supported \
+YUV formats and bit depths (with alpha).
+For now, it is recommended that you checkout/use tagged releases instead of \
+just using the master branch. I will regularly create new versions as \
+bugfixes and features are added."
+HOMEPAGE="https://https://github.com/AOMediaCodec/libavif"
+COPYRIGHT="2019 Joe Drago"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/AOMediaCodec/libavif/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="c917f4a27add61b8ad15f328090942d519460b71219f31e60e5fe6aa8cf3dc4e"
+
+ARCHITECTURES="!x86_gcc2 ?x86 !x86_64 ?arm ?riscv64 ?sparc ?m68k"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="5.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	libavif$secondaryArchSuffix = $portVersion
+	lib:libavif$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libdav1d$secondaryArchSuffix
+	"
+
+PROVIDES_tools="
+	cmd:avifenc$secondaryArchSuffix
+	cmd:avifdec$secondaryArchSuffix
+	"
+REQUIRES_tools="
+	haiku$secondaryArchSuffix
+	lib:libavif$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libpng$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	libavif${secondaryArchSuffix}_devel = $portVersion
+	devel:libavif$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	libavif$secondaryArchSuffix == $portVersion base
+	devel:libdav1d$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libdav1d$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libpng$secondaryArchSuffix
+	devel:libjpeg$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+defineDebugInfoPackage libavif$secondaryArchSuffix \
+	"$libDir"/libavif.so.$libVersion
+
+BUILD()
+{
+	mkdir -p build && cd build
+	cmake .. \
+		-DAVIF_BUILD_APPS=ON \
+		-DAVIF_CODEC_DAV1D=ON \
+		$cmakeDirArgs
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make -C build install
+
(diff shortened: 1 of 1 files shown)


media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,96 @@
...
+just using the master branch. I will regularly create new versions as \
+bugfixes and features are added."
+HOMEPAGE="https://https://github.com/AOMediaCodec/libavif"
+COPYRIGHT="2019 Joe Drago"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/AOMediaCodec/libavif/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="c917f4a27add61b8ad15f328090942d519460b71219f31e60e5fe6aa8cf3dc4e"
+
+ARCHITECTURES="?x86_gcc2 ?x86 !x86_64 ?arm ?riscv64 ?sparc ?m68k"
```
Begasus, line 20: You can mark x86_gcc2 as broken as there is no dav1d version that can be build for that architecture:
```diff
- ARCHITECTURES="?x86_gcc2 ?x86 !x86_64 ?arm ?riscv64 ?sparc ?m68k"
+ ARCHITECTURES="!x86_gcc2 ?x86 !x86_64 ?arm ?riscv64 ?sparc ?m68k"
```

media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,96 @@
...
+	lib:libavif$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libdav1d$secondaryArchSuffix
+	"
+
+PROVIDES_tools="
+	cmd:avifenc
+	cmd:avifdec
```
Begasus, line 37: On gcc2h they are "cmd:avifen_x86" so:
```diff
-	cmd:avifenc
-	cmd:avifdec
+	cmd:avifenc$secondaryArchSuffix
+	cmd:avifdec$secondaryArchSuffix
```
Or maybe use -DCMAKE_INSTALL_BINDIR=$prefix/bin ? (not tested)

media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,96 @@
...
+	prepareInstalledDevelLib libavif
+	fixPkgconfig
+	packageEntries devel $developDir
+	packageEntries tools $binDir
+}
+
+TEST()
+{
+	cd build
+	make test
```
Begasus, line 95: No rule to make target test?

media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,96 @@
...
+the AV1 Image File Format, as described here: \
+ \
+https://aomediacodec.github.io/av1-avif/ \
+ \
+It is a work-in-progress, but can already encode and decode all AOM supported \
+YUV formats and bit depths (with alpha). \
+ \
+For now, it is recommended that you checkout/use tagged releases instead of \
+just using the master branch. I will regularly create new versions as \
+bugfixes and features are added."
```
Begasus, line 12: IMHO this looks better when opening with HaikuDepot:
```
DESCRIPTION="This library aims to be a friendly, portable C implementation of \
the AV1 Image File Format, as described here: \
https://aomediacodec.github.io/av1-avif/ 

It is a work-in-progress, but can already encode and decode all AOM supported \
YUV formats and bit depths (with alpha).

For now, it is recommended that you checkout/use tagged releases instead of \
just using the master branch. I will regularly create new versions as \
bugfixes and features are added."
```

media-libs/libavif/libavif-0.8.0.recipe
linkmauve: The target was actually named `avif_test_all`, but it takes way too long so I instead removed them altogether from this package.

media-libs/libavif/libavif-0.8.0.recipe
linkmauve: Thanks!

media-libs/libavif/libavif-0.8.0.recipe
linkmauve: I’m not sure I understand your second suggestion, I did add the `$secondaryArchSuffix` but from what I understand the `packageEntries` command already does the correct work even on `$binDir`.

Begasus: Maybe this will explain a bit better on what I'm trying to say here: https://github.com/haikuports/haikuports/blob/43a37f4c039afa53b47f4ab79279a876d476ce6a/media-libs/partio/partio-1.1.0.recipe#L25

Begasus: > I’m not sure I understand your second suggestion, I did add the `$secondaryArchSuffix` but from what I understand the `packageEntries` command already does the correct work even on `$binDir`.

On x86_gcc2 $bindir = /prefix/bin/x86 (for secondaryArchSuffix)
On x86_x64 $bindir = /prefix/bin

If a package can't be build for primary architecture "!x86_gcc2" then we want the binaries installed in $prefix/bin rather then $prefix/bin/x86 for secondary architecture "x86" (secondaryArchSuffix), hence we use $commandBinDir and $commandSuffix then

media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,87 @@
+SUMMARY="Library for encoding and decoding .avif files"
+DESCRIPTION="This library aims to be a friendly, portable C implementation of \
+the AV1 Image File Format, as described here: \
+https://aomediacodec.github.io/av1-avif/
+It is a work-in-progress, but can already encode and decode all AOM supported \
+YUV formats and bit depths (with alpha).
+For now, it is recommended that you checkout/use tagged releases instead of \
+just using the master branch. I will regularly create new versions as \
+bugfixes and features are added."
+HOMEPAGE="https://https://github.com/AOMediaCodec/libavif"
```
sikmir, line 10:
```suggestion
HOMEPAGE="https://github.com/AOMediaCodec/libavif"
```

media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,87 @@
...
+	"$libDir"/libavif.so.$libVersion
+
+BUILD()
+{
+	mkdir -p build && cd build
+	cmake .. \
+		-DAVIF_BUILD_APPS=ON \
+		-DAVIF_CODEC_DAV1D=ON \
+		$cmakeDirArgs
+	make $jobArgs
```
sikmir, line 76:
```suggestion
	cmake --build . -- $jobArgs
```

media-libs/libavif/libavif-0.8.0.recipe
```diff
@@ -0,0 +1,87 @@
...
+}
+
+INSTALL()
+{
+	make -C build install
```
sikmir, line 81:
```suggestion
	cmake --build build --target install
```

Begasus:
@linkmauve any progress here?

Begasus:
rav1e added in https://github.com/haikuports/haikuports/pull/9723

PR #5209, opened 2020-07-19
Vim: Enable Python3 interpreter

bitigchi:
Bumps version and enables Python3 interpreter. Untested on 32-bit.

In an unrelated note, can someone try enabling the Lua interpreter? On my system I can't seem to get it working, maybe on 32-bit works. There is a growing number of Lua Neovim plugins and Vim saw some nice fixes for interoperability with Neovim. Since we don't have Neovim, it would be nice to have it.


bitigchi:
I would appreciate if someone could test this out on 32-bit.

PR #5235, opened 2020-08-06
Update: Flatstyle

unarix:
This version improves the display of:
- Menues: Flat borders + correct background color
- Scrollbars: Correct background color
- Size borders: Contrasts
- ToolBar: Gradient contrast
- Menu Items: Gradient contrast


unarix:
Tested in 32 bits too:

x86

PR #5246, opened 2020-08-11
hteditor: avoid putting th in /bin/x86

humdingerb:
Since there's no gcc2 version, just the x86 built.

### DIFF
--- a/app-editors/hteditor/hteditor-2.1.0.recipe
+++ b/app-editors/hteditor/hteditor-2.1.0.recipe
@@ -8,7 +8,7 @@ HOMEPAGE="http://hte.sourceforge.net/"
 COPYRIGHT="1999-2002 Stefan Weyergraf
     1999-2002 Sebastian Biallas"
 LICENSE="GNU GPL v2"
-REVISION="2"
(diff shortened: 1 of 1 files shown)


app-editors/hteditor/hteditor-2.1.0.recipe
```diff
@@ -45,7 +52,7 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	runConfigure ./configure
+	runConfigure ./configure --bindir="$commandBinDir"
```
Begasus, line 55: Is "--omit-dirs" depreceated?

humdingerb: I don't know... I've just had a look at existing recipes, and the above worked, so... :)

Begasus: If build fails, please adjust :)

humdingerb: Built succeeded and "ht" appears directly in /bin. Complete and utter success! :)

PR #5262, opened 2020-08-26
spdlog, bump version

Begasus:


Begasus:
This add a shared library for spdlog also

PR #5268, opened 2020-08-31
Yab

lorglas:
New version 1.7.9 of yab
Bug fixes
- Bug fixes in YabInterfaces (some return Statement added (bbjimmy/lorglas)
Reactivating Feature
- Columnbox sorting activated
Enhancement 
- Sound command loop activted, Example: sound play "file.mp3", loop
- texturl allow to use https
- Change Cursor to Haiku Style


bbjimmy:
shouldn't this pr have only the yab recipe, not all the others?

Begasus:
Indeed, and if yab needs the first boost it would be wise to mention this also before merging the other ones ;)

PR #5277, opened 2020-09-13
Paladin: Added Haiku header requirement

panos:
After testing out one of Paladin's templates, a user on social media seemed unable to compile a template which included the lines;

```cpp
#include <Application.h>
#include <Window.h>
```

Consequently, they claim to have resolved that issue by installing `haiku_devel` in their system afterwards.

I believe that if the program itself assumes that the user has the Haiku headers installed on their system, does not check for it and bundles Haiku-specific code that's meant to be illustrating the system's capabilities to users experimenting with the system, then the recipe itself should cover this case scenario.

### DIFF
--- a/haiku-apps/paladin/paladin-2.8~git.recipe
+++ b/haiku-apps/paladin/paladin-2.8~git.recipe
@@ -45,6 +45,8 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku
+	# required by templates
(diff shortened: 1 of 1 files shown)


panos:
Quick sidenote; Would adding `binutils` be a good idea as well, considering that Paladin does include a built-in feature involving executing instructions in `Makefile` files?

PR #5283, opened 2020-09-14
Drag[en]gine 1.3 Release

LordOfDragons:


games-engines/dragengine/dragengine-1.3.recipe
```diff
@@ -15,9 +15,9 @@ COPYRIGHT="2020-2020 Dragon Dreams"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
 REVISION="1"
-srcGitRev="1.2-1-haiku"
+srcGitRev="1.3"
```
Begasus, line 18: In previous version  srcGitRev was different from portVersion, I'm guessing some fix made it through so it can use the default release? If so, srcGitRev should be obselete and in SOURCE_URI portVersion could be used?

LordOfDragons: The idea is to use the same tag as for regular releases unless I need to apply a haiku only fix between releases.

LordOfDragons: I've commited the changes. This way it goes into the 1.4 release since I copy-paste the recipe file.

PR #5299, opened 2020-09-25
libretro cores : bump to latest git revisions

kwyxz:
As title says, bumping various libretro cores to the latest revisions that build properly


diversys:
https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/4823/builds/17542.log

kwyxz:
Interesting. It should not have pushed a new revision if the build failed. Looking into this, thanks!

kwyxz:
Actually, I don't see a commit for bsnes, so I am not sure what happened :|

kwyxz:
I see it in this commit... https://github.com/kwyxz/haikuports/commit/4b88971530af119b1bf5be5c0e59803ff9cf10ea definitely not normal.

diversys:
Watch this place https://build.haiku-os.org/buildmaster/master/x86_64/

kwyxz:
Fixed in #5300
Now looking at `easyrpg`

PR #5303, opened 2020-09-27
flex, add symlink to lex

Begasus:


Begasus:
Tests are failing with parse errors on gcc2, worked on a recipe where it's using gcc8 (flex_x86), there the tests are all OK
@korli as you did the last version bump could you have a look? (not sure if building flex with gcc8 would break other recipes on 32 bit) ... https://github.com/Begasus/haikuports/commit/90ad57c9635a938c763e8289e1934831aaccaec2

Begasus:
Merged this one as there are no structural changes, will leave the PR for the gcc8 version open for a while (guessing it needs a "CONFLICT=" in there to not have both versions installed at the same time, or use cmd:flex$secondaryArchSuffix for the _x86 version

PR #5316, opened 2020-10-11
lua 5.4, bump version

Begasus:


Begasus:
Would love some feedback on this one :) It's not enabled (by default) to not interfere with previous version from lua

PR #5324, opened 2020-10-14
Matplotlib: new recipe

iam-git:


dev-python/matplotlib_python3/matplotlib_python3-3.3.2.recipe
```diff
@@ -0,0 +1,76 @@
...
+also a procedural \"pylab\" interface, designed to closely resemble that of \
+MATLAB, though its use is discouraged."
+HOMEPAGE="https://matplotlib.org/"
+COPYRIGHT="2002 - 2012 John Hunter, Darren Dale, Eric Firing, Michael \
+Droettboom and the Matplotlib development team \
+2012 - 2020 The Matplotlib development team"
+LICENSE="Python" # "PSF-based" actually
+REVISION="1"
+SOURCE_URI="https://github.com/matplotlib/matplotlib/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="a17664009de73849906de076c1de60d2854d26325d012cdd0b5e75edb8bd8edb"
```
Begasus, line 15: Maybe add SOURCE_FILENAME here?
```diff
+SOURCE_FILENAME="matplotlib-$portVersion.tar.gz"
```

dev-python/matplotlib_python3/matplotlib_python3-3.3.2.recipe
```diff
@@ -0,0 +1,76 @@
...
+	"
+
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libpng16$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	"
+
+pythonPackage="python3"
+pythonVersion="3.7"
```
Begasus, line 32: Add a new line here

dev-python/matplotlib_python3/matplotlib_python3-3.3.2.recipe
```diff
@@ -0,0 +1,76 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libpng16$secondaryArchSuffix
+	devel:libfreetype${secondaryArchSuffix}
+	pyqt${secondaryArchSuffix}_$pythonPackage
+	numpy${secondaryArchSuffix}_$pythonPackage
+	dateutil_$pythonPackage
+	pytz_$pythonPackage
+	six_$pythonPackage
+	pillow${secondaryArchSuffix}_$pythonPackage
+	cmd:python$pythonVersion
```
Begasus, line 43: Sort A-Z

Begasus (commented):
Haven't checked the build (I'm guessing it depends on the PR for dateutil?), but some cosmetic changes

iam-git:
Should I change recipe to generate empty _matplotlib_ package and populated _matplotlib_python3_ one?

extrowerk:
> Should I change recipe to generate empty _matplotlib_ package and populated _matplotlib_python3_ one?

Yep, thats how it supposed to be.

pulkomandy:
Is it? Maybe we should stop doing that now, and stop adding python2 version of things? Or do we still need it for future python3 versions?

iam-git:
In common, there are still python versions 3.6, 3.7, 3.8, 3.9 (this recipe is limited by pyqt and contains stuff only for 3.7)

extrowerk:
> Is it? Maybe we should stop doing that now, and stop adding python2 version of things? Or do we still need it for future python3 versions?

Is it a reply to me? If so: basically haikuporter generates a package for every recipe, but we put the data un named subpackages (_$pythonversion), now in most case the main package will be empty (except if there is data which can be shared between the different pythonversion subpackages. In this case as this recipe will be built only against 3.7 the main package will be empty. I think this is how it works, or am i misunderstood something?

pulkomandy:
Yes, that's what we do currently but this scheme was introduced so we can handle multiple python versions. I think now that Python2 is dead, we should not add support for it in new packages. So if there is only an empty main package and a single subpackage, why make a subpackage? A single package would be fine.

The reason this may not work is if we need to support multiple Python 3 versions (3.7, 3.8, 3.9) with different packages.

Begasus:
pytest is one of those recipes where the base package isn't empty, building it resolves in the base package and one for python 3.6 and 3.7, running pytest from Terminal shows the use of python 3.7, so I'm guessing it should be narrowed down also to one base package for only 3.7, will check on this and see if we can use it for pytest also then ;)

iam-git:
I probably would not compare matplotlib and pytest: nobody (I guess) runs matplotlib from Terminal. But, should I put back single package matplotlib_python3$secondaryArchSuffix for now?

Does it means in case of pytest it will fall if I try to test my code developed in 3.8 (with dependencies installed only there)? If so, it looks for me it should be separate recipe for every python version, same as python itself. Never used pytest, though.

Begasus:
> I probably would not compare matplotlib and pytest: nobody (I guess) runs matplotlib from Terminal. But, should I put back single package matplotlib_python3$secondaryArchSuffix for now?
> 
> Does it means in case of pytest it will fall if I try to test my code developed in 3.8 (with dependencies installed only there)? If so, it looks for me it should be separate recipe for every python version, same as python itself. Never used pytest, though.

Don't know if you also need to add 3.8 packages for the one you are developing, at this stage no sub packages for 3.8 exists

dev-python/matplotlib/matplotlib-3.3.2.recipe
```diff
@@ -0,0 +1,89 @@
...
+language and its numerical mathematics extension NumPy. It provides an \
+object-oriented API for embedding plots into applications using \
+general-purpose  GUI toolkits like Tkinter, wxPython, Qt, or GTK+. There is \
+also a procedural \"pylab\" interface, designed to closely resemble that of \
+MATLAB, though its use is discouraged."
+HOMEPAGE="https://matplotlib.org/"
+COPYRIGHT="2002 - 2012 John Hunter, Darren Dale, Eric Firing, Michael \
+Droettboom and the Matplotlib development team \
+2012 - 2020 The Matplotlib development team"
+LICENSE="Python" # "PSF-based" actually
```
Begasus, line 12: You need to add the license here too (as a seperate file), see for instance:
https://github.com/haikuports/haikuports/blob/c2a214e6473117adfdbf84d0f792341b30e1bbd7/dev-python/lxml/licenses/PSF-2
https://github.com/haikuports/haikuports/blob/c2a214e6473117adfdbf84d0f792341b30e1bbd7/dev-lang/python/licenses/Python

iam-git: Got it

Begasus:
Not an expert on licenses here ... pokes @alaviss ;)

Begasus:
Gentoo reports this as LICENSE="BitstreamVera BSD matplotlib MIT OFL-1.1":
https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/matplotlib/matplotlib-3.3.2.ebuild#n24

iam-git:
I see some more stuff there (and it's somewhat different): https://github.com/matplotlib/matplotlib/tree/master/LICENSE
However, QHull for example still isn't mentioned there.

alaviss:
> Not an expert on licenses here ... pokes @alaviss ;)

I usually stick to Debian's analysis: https://metadata.ftp-master.debian.org/changelogs//main/m/matplotlib/matplotlib_3.3.2-2_copyright

And that's a lot of licenses to add :P

iam-git:
Any such recipe for inspiration?

PR #5325, opened 2020-10-14
dateutil: __version__ fix

iam-git:


Begasus:
Checking the full recipe reveals a trailing whitespace at line 61? 
https://github.com/haikuports/haikuports/blob/6a0db88b92ffc75d69e05d1f323969e2aa12cbe5/dev-python/dateutil/dateutil-2.8.1.recipe#L61

iam-git:
New releases of matplotlib check version of dateutil and raise ImportError if see something like "0.1.dev1+g6104a1a".

korli:
Wouldn't it be better to use python-setuptools-scm?

iam-git:
As I understand, python-setuptools-scm generates version number from git. In our case it is that empty git repo, made by haikuporter. 

So now I just set SETUPTOOLS_SCM_PRETEND_VERSION environment variable to get finally the right value.

dev-python/dateutil/dateutil-2.8.1.recipe
```diff
@@ -45,6 +45,9 @@ REPLACES_python="
 	python_dateutil
 	"
 
+SETUPTOOLS_SCM_PRETEND_VERSION=$portVersion
+export SETUPTOOLS_SCM_PRETEND_VERSION
```
Begasus, line 49: Stumbled upon a simular problem when checking up on zipp(_python), could you check if bumping setuptools to version 42.0.0 fixes it without this fix?

Begasus:
@korli  @iam-git see here (zipp also had the version number wrong before I updated setuptools to 42.0.0)
https://github.com/jaraco/zipp/issues/53#issuecomment-627062475

Begasus:
Ok, this indeed fixes the version format, checked with setuptools_scm vs setuptools, doesn't change anything, so setuptools should be ok to use here, while you are in the process of a revbump, maybe you could remove the python 2.7 dependency also ;)

Begasus:
Looks like I need this fix also for iniconfig

iam-git:
While I am in the process of a revbump, I would like to add python38 and python39 subpackages, but corresponding dependencies are not met.

Begasus:
> While I am in the process of a revbump, I would like to add python38 and python39 subpackages, but corresponding dependencies are not met.

Most of the python recipes haven't been updated to use the newer python version, help welcome :)

iam-git:
> Most of the python recipes haven't been updated to use the newer python version, help welcome :)

I'll consider revbumps for some packages I use, but at least in case of setuptools I'll rather rely to more skilled guys ;)

Begasus:
Merged, seems like there are still some dependencies for python 2.7 that needs to be cleaned before we can remove it here

Issue #4431, opened 2019-12-08
Add a scipy_python3 recipe


iam-git:
And maybe put **scipy** to **haikuports/dev-python/**?

korli:
No it should stay in sci-libs:
https://packages.gentoo.org/packages/sci-libs/scipy

Begasus:
Also create one for scipy_python36 ...

While in the process bump version to 1.3.1 https://github.com/scipy/scipy/releases/tag/v1.3.1

korli:
@Begasus can this be closed?

Begasus:
The recipe is merged, so it should be ok :)

PR #5328, opened 2020-10-15
scipy: python3 recipe

iam-git:
solves #4431

Recipe has following limitations:
- ndimage module is disabled
- no x86 build
- stuck on 1.3.3 version, newer falls

BTW, existing recipe doesn't work for me at all, donno why.

### DIFF
--- a/sci-libs/scipy/scipy-1.3.3.recipe
+++ b/sci-libs/scipy/scipy-1.3.3.recipe
@@ -4,7 +4,7 @@ SciPy supplements the popular numpy module, gathering a variety of high level \
 science and engineering modules together as a single package. SciPy includes \
 modules for graphics and plotting, optimization, integration, special \
 functions, signal and image processing, genetic algorithms, ODE solvers, and \
-others. 
+others.
 
 This package does not contain ndimage module."
 HOMEPAGE="https://www.scipy.org/"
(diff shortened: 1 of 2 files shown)


sci-libs/scipy/scipy-1.3.3.recipe
```diff
@@ -0,0 +1,123 @@
...
+	lib:libarpack$secondaryArchSuffix
+	lib:libblis$secondaryArchSuffix
+	lib:liblapack$secondaryArchSuffix
+	lib:libopenblas$secondaryArchSuffix
+	lib:libumfpack$secondaryArchSuffix # not sure if actually used
+	numpy$secondaryArchSuffix
+	"
+
+PYTHON_PACKAGES=(python36 python3)
+PYTHON_VERSIONS=(3.6 3.7)
```
korli, line 41: Please remove 3.6.

extrowerk:
Build, packaging ok on 64 bit.

sci-libs/scipy/scipy-1.3.3.recipe
Begasus: @korli should we start using 3.7, 3.8 and 3.9 in the recipes now?
On that note, maybe if we want 3.7 to be default we could use:
```diff
-PYTHON_PACKAGES=(python36 python3)
-PYTHON_VERSIONS=(3.6 3.7)
+PYTHON_PACKAGES=(python python38 python39)
+PYTHON_VERSIONS=(3.7 3.8 3.9)
```

korli: There is an open issue on python 3.8 or 3.9, would better fix it first

PR #5339, opened 2020-10-18
tcl, revbump, remove sqlite3 dependency (solves build on x86_64)

Begasus:
This should fix the build for 64 bit Haiku, checked so far with expect and sqlcipher

### DIFF
--- a/dev-lang/tcl/tcl-8.6.10.recipe
+++ b/dev-lang/tcl/tcl-8.6.10.recipe
@@ -8,13 +8,13 @@ is truly cross platform, easily deployed and highly extensible."
 HOMEPAGE="http://www.tcl.tk"
 COPYRIGHT="Regents of the University of California, Sun Microsystems, Inc., Scriptics Corporation, and other parties"
 LICENSE="BSD (2-clause)"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="http://sourceforge.net/projects/tcl/files/Tcl/$portVersion/tcl$portVersion-src.tar.gz"
 CHECKSUM_SHA256="5196dbf6638e3df8d5c87b5815c8c2b758496eb6f0e41446596c9a4e638d87ed"
 SOURCE_DIR="tcl$portVersion"
 PATCHES="tcl-$portVersion.patchset"
 
-ARCHITECTURES="x86 x86_gcc2 x86_64"
+ARCHITECTURES="x86_gcc2 x86_64"
 
 PROVIDES="
 	tcl = $portVersion compat >= 8.6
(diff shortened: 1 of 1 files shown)


Begasus:
Didn't find a full list of all recipes depending on tcl atm, will update later:
- [x] expect
- [x] sqlcipher

waddlesplash:
IIRC tcl has a bundled SQLite so we should use the package instead of removing the dependency.

Begasus:
> IIRC tcl has a bundled SQLite so we should use the package instead of removing the dependency.

That's one of the reasons I removed it from the recipe (it fails to link on x86_64 with the system sqlite3 (undefined references ...)
tcl still uses and builds in bundled one, for some reason it works ok for x86_gcc2 but not for x86_64, build also checked without the dependency on x86_gcc2 doesn't show an error either

Begasus:
Without this change (on x86_64, even with --enable-64bit):
```
gcc -pipe -O2 -DNDEBUG -Wall -fPIC   -shared -o libsqlite3.30.1.2.so tclsqlite3.o -lpthread  -lsqlite3 -lnetwork -L/sources/tcl8.6.10/unix -ltclstub8.6
...
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/8.3.0/../../../../x86_64-unknown-haiku/bin/ld: tclsqlite3.c:(.text+0x64ee): undefined reference to `sqlite3_errmsg'
```

extrowerk:
<Habbie> [22:46:07] extrowerk, aha! strace -f is 'fast mode', not 'follow forks'!
<Habbie> [23:09:10] extrowerk, https://imgur.com/a/reOnyTg
<Habbie> [23:09:21] extrowerk, 58 is the failing one (with some extra -I but those don't matter there)
<Habbie> [23:09:23] extrowerk, 59 works
<Habbie> [23:09:28] extrowerk, i do not know why that is
<Habbie> [23:11:21] extrowerk, ok, it is because one of the -many- -D options passed in building the .o
<Habbie> [23:17:00] extrowerk, specifically, everything works if i remove -DSQLITE_API and -DSQLITE_EXTERN
<Habbie> [23:17:37] extrowerk, so, simplest case, now we have something we can patch; but it would be better to find out why this is happening. It could even be a TCL bug, it really shouldn't be setting 50 -DSQLITE_ things when using system sqlite

extrowerk:
Adding
```
	export CFLAGS="-USQLITE_API -USQLITE_EXTERN"
```
to the recipe solves the build problem on 64 bit Haiku. I can't seem to find why they defined in the first place. Many test fails.

Habbie:
>  I can't seem to find why they defined in the first place.

It looks to me like they forgot to make that huge set of defines conditional on `--with-system-sqlite3`. I could not find how the TCL tar ball is put together, and hence could not find where to report that upstream.

Begasus:
@extrowerk @Habbie could you check, works for both 32 bit and 64 bit here

extrowerk:
Builds fine on 64 bit.

PR #5348, opened 2020-10-24
pytest, remove python 2.7 dependency

Begasus:


Begasus:
Maybe we should use "another" default python version, with this it builds: pytest, pytest_python3 and pytest_python36

PR #5305, opened 2020-10-01
ffmpeg: bump recipe, enabled x86_64

panos:
Tested on x86_64.

PR #5350, opened 2020-10-24
FFMPEG: bump, enable for 64 bit

extrowerk:
I was unable to find out why it was disabled on 64 bit, but if you happen to know please let me know.


diversys:
See #5305.

PR #5357, opened 2020-10-25
GPXSee: bump to 7.34

sikmir:
[Changelog](https://build.opensuse.org/package/view_file/home:tumic:GPXSee/gpxsee/gpxsee.changes)


humdingerb (commented):
Going by the changelog, you could change "* Print/export to PDF" inline 18to say "* Print/export to PDF or PNG".

I couldn't test it though, because building fails for me with "`runtime_loader: Cannot open file libicudata.so.57: No such file or directory`"
That's probably just my system, though...

extrowerk:
Builds fine on 64 bit.

humdingerb:
Thanks!

PR #5359, opened 2020-10-26
Kristall: Add recipe

win8linux:
[Kristall](https://kristall.random-projects.net/) is a cross-platform visual browser with a strong emphasis on supporting alternative internet protocols and document types, however it also still supports regular HTML over HTTP/HTTPS.

This program actually has [Haiku build instructions](https://github.com/MasterQ32/kristall/blob/master/BUILDING.md), but hasn't been packaged up before.


www-client/kristall/kristall-0.7.3~git.recipe
```diff
@@ -0,0 +1,68 @@
+SUMMARY="A cross-platform, multi-protocol visual browser for the small-internet"
+DESCRIPTION="Kristall is a browser without support for css/js/wasm or graphical \
+websites. It can display user-styled documents in several formats, \
+including Gemini, HTML, and Markdown provided by a server via Gemini, \
+Gopher, HTTP/HTTPS, and Finger protocols."
+HOMEPAGE="https://kristall.random-projects.net/"
+COPYRIGHT="2020 Felix Queißner"
+LICENSE="GNU GPL v3"
+REVISION="1"
+srcGitRev="b684f94f1af9a19c1a6fd70d72097a13b75e1ca6"
```
Begasus, line 10: Any specific reason to use srcGitRev here? Latest official release is from June this year: https://github.com/MasterQ32/kristall/releases/tag/V0.3

www-client/kristall/kristall-0.7.3~git.recipe
```diff
@@ -0,0 +1,68 @@
...
+SOURCE_URI="https://github.com/MasterQ32/kristall/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="ad8dd1fa28702e94d8830fe4ba0908e44417396e7402ebfa974897bfcd8be7dd"
+SOURCE_FILENAME="kristall-$portVersion-tar.gz"
+SOURCE_DIR="kristall-$srcGitRev"
+
+ARCHITECTURES="x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
```
Begasus, line 17: If primary architecture x86_gcc2 is ok with a build, there wouldn't be a reason to enable "SECONDARY_ARCHITECTURES" (didn't check the build) if it's not building for gcc2 please disable the build there (also remove the build for primary x86):
```diff
- ARCHITECTURES="x86_gcc2 x86 x86_64"
+ ARCHITECTURES="!x86_gcc2 x86_64"
```

www-client/kristall/kristall-0.7.3~git.recipe
```diff
@@ -0,0 +1,68 @@
...
+	devel:libQt5Core$secondaryArchSuffix
+	devel:libQt5Gui$secondaryArchSuffix
+	devel:libQt5Network$secondaryArchSuffix
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:git
```
extrowerk, line 49: git should not be used in recipes.

www-client/kristall/kristall-0.7.3~git.recipe
```diff
@@ -0,0 +1,68 @@
...
+
+BUILD()
+{
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/kristall
+	cp ./kristall $appsDir/kristall
```
extrowerk, line 66: If it is just one binary then we shouldn't use a subfolder.

www-client/kristall/kristall-0.7.3~git.recipe
win8linux: I only added it since during build, there were frequent failed attempts at using it. Building fails without it, oddly.

www-client/kristall/kristall-0.7.3~git.recipe
win8linux: Mostly to just have the package include the latest changes between the current release and the next one. I could change it to be based on the latest tagged release if that's more preferable.

extrowerk:
Builds, packages and runs fine on 64 bit, but the main executable have no icon yet.

win8linux:
The program itself doesn't have a logo or anything to base an icon off of.

win8linux:
Couldn't really figure out how to make a Haiku-esque icon, so I just converted the SVG icon into a format suitable for packaging.

Begasus:
LGTM, haven't been able to finish the build (out of memory) but from what I see there are no errors in the build so far :)

win8linux:
> haven't been able to finish the build (out of memory)

Yeah, I'd recommend 4 GB for building this recipe.

win8linux:
Kristall running on Haiku:
[](https://postimg.cc/NyH0qg92)

Begasus:
Merged, keep track on the buildbot for eventual errors :) 
https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=&viewMode=expanded

diversys:
virtual memory exhausted: Out of memory
Perhaps disabling ASLR could help here?

`export DISABLE_ASLR=1`

Begasus:
Worth a try if you think it could help, build on x86_64 finished ok

win8linux:
> virtual memory exhausted: Out of memory
> Perhaps disabling ASLR could help here?
> 
> `export DISABLE_ASLR=1`

Just tried this and it doesn't seem to make a notable difference with regards to memory usage during building. As previously mentioned, 4 GB of RAM should be enough to build the recipe successfully. Perhaps this could be cross-compiled from a `x86_64` build machine instead?

Begasus:
Not sure how I can do here anymore, tried building with "make -j 1", added a swap partition for 5GB, still fails with the above, at least it's available on x86_64 :)

korli:
yeah it seems the project embeds binary files in resources, which means gcc has to handle the binaries through multiple stages. This would need to be fixed to load resource files in the code (in this case some fonts!!).

PR #5391, opened 2020-11-08
GMP: bump

extrowerk:


extrowerk:
All test passes

extrowerk:
Was unable to update binutils, so you probably don't want to morge this.

extrowerk:
@kolri: should i rename the old versio here too?

korli:
no it should stay compatible.

PR #5395, opened 2020-11-10
Angband: fix libSDL references

lonemadmax:
There should be no problems, but this is untested, I've only checked that it compiles in my x86_64 system.

Once I'm sure this is the kind of change expected from #3878, I'll try and continue sending PR with more than one recipe, if that's OK.


Begasus:
If libraries are involved in a bulk PR please check there dependencies also, thanks!

lonemadmax:
I only expected to find the SDL ones and was leaving them for the end when the rest of packages pointed to the correct names, but there are other libraries using libSDL, of course. What needs to be checked in dependencies if the only changes in a library recipe are in requires and not in provides?

PR #5399, opened 2020-11-11
Poppler_data: bump

extrowerk:


korli:
This is a bump version. Is it compatible with poppler 0.76 and poppler 0.66?

extrowerk:
According to the #poppler IRC channel it should be backwards compatible.

PR #5400, opened 2020-11-11
Poppler: bump

extrowerk:


app-text/poppler/poppler-20.11.0.recipe
```diff
@@ -273,15 +278,15 @@ REQUIRES="
 PROVIDES_devel="
 	poppler${secondaryArchSuffix}_devel = $portVersion
 	devel:libpoppler$secondaryArchSuffix = $libVersionCompat
-	devel:libpoppler${secondaryArchSuffix}_cpp = $libCppVersionCompat
-	devel:libpoppler${secondaryArchSuffix}_glib = $libGlibVersionCompat
-	devel:libpoppler${secondaryArchSuffix}_qt5 = $libQt5VersionCompat
+	devel:libpoppler_cpp$secondaryArchSuffix = $libCppVersionCompat
+	devel:libpoppler_glib$secondaryArchSuffix = $libGlibVersionCompat
+	devel:libpoppler_qt5$secondaryArchSuffix = $libQt5VersionCompat
```
korli, line 283: Why is it modified?

app-text/poppler/poppler-20.11.0.recipe
```diff
@@ -264,6 +267,8 @@ REQUIRES="
 	lib:libintl$secondaryArchSuffix
 	lib:libjpeg$secondaryArchSuffix
 	lib:liblcms2$secondaryArchSuffix
+	lib:libnspr4$secondaryArchSuffix
+	lib:libnss3$secondaryArchSuffix
```
korli, line 271: but devel:libnss3$secondaryArchSuffix is commented out

Begasus: It's mentioned in line 385 (BUILD_REQUIRES)

Begasus:
Builds ok on 32 bit

app-text/poppler/poppler-20.11.0.recipe
```diff
@@ -294,6 +299,7 @@ REQUIRES_devel="
 	devel:libjpeg$secondaryArchSuffix
 	devel:liblcms2$secondaryArchSuffix
 	devel:libopenjp2$secondaryArchSuffix
+#	devel:libnss3$secondaryArchSuffix
```
korli, line 302: Why add this Line?

Begasus: That's one for @extrowerk :)

Begasus:
Seems to be OK, just did a build for TexStudio with this one, if this is merged, TexStudio should get a revbump :) (first merge poppler_data)

app-text/poppler/poppler-20.11.0.recipe
extrowerk: AFAIR poppler could use libnss, the configure stage searching for it, but was unable to find it. Could be removed.

Begasus:
Looking at the packages I see that for instance: libpoppler-glib.so.8.19.0 (and .8) is provided in the base package as well as in the *_glib package, is that correct?
Didn't check, but shouldn't those be separated from the base package?

extrowerk:
>Looking at the packages I see that for instance: libpoppler-glib.so.8.19.0 (and .8) is provided in the base package as well as in the *_glib package, is that correct?
Thats doesn't seems to be ok.

Basically the base package provides the cli tools and the poppler (which is the C version) and the cpp version.
qt5 provides the qt5 lib, while glib the glib one.

I checked on my side again, the base package contains only the c-version and the cpp version lib. Please check again on your side, i think you made a mistake somewhere.

PR #5401, opened 2020-11-11
coin, revbump, fix for 32 bit build

Begasus:


Begasus:
@extrowerk could you check this? ;)

PR #5402, opened 2020-11-11
CMake: add gui subpackage

extrowerk:


extrowerk:
Probably we don't want to merge this. Maybe some of the Qt's dependency requires cmake to build with.

korli:
you're right, but we can also try and see how problematic it is, the cmake package won't have runtime dependency on Qt anyway.

extrowerk:
Thats true, but actually the gui is not that useful as i expected. IT lets generate a makefile, but that's all. I tought it will also build things, but thats probably not the scope of the gui. So i don't really see the point to have a gui, but other platforms providing it, so if it doesn't makes a headache us then we could do the same.

Begasus:
Can this be merged @extrowerk ?

extrowerk:
I'vw tested it only on 64 bit, but i am not against it.

Schrijvers Luc <notifications@github.com>:

Begasus:
Let's merge then ;)

PR #5413, opened 2020-11-15
fuse_utils, bump version

Begasus:


Begasus:
Switched to secondary architecture because libspectrum does not provide on for primary architecture

PR #5419, opened 2020-11-17
OpenDUNE: add support for x86_gcc2

lonemadmax:


korli:
Please push this patch upstream. If it's not accepted there, it's better to drop it.

Begasus:
Don't think it's worth the effort, but good practice anyway ;) We already supply a package for 32bit (secondary architecture)

lonemadmax:
What can I say, I'm new here :-)

I chose OpenDUNE because there was a half working recipe already that didn't seem to need too much work, and I had the game files from my DOS days. Then I fell down a rabbit hole. I would have even tested other arches had I have a working installation.

Anyway, closing this request. Surprisingly (at least for me) [upstream has accepted](https://github.com/OpenDUNE/OpenDUNE/pull/347) even the patches that touch the build system and I've also moved some recipe magic there, so if there's a version 1.0 it will be simpler. Only thing left is the dsp sound patch, which I suspect workarounds a bug in Haiku and has nothing to do with the game.

PR #5445, opened 2020-11-26
Armadillo: bump

extrowerk:
Cannibalized old recipe kept for compatibility


Begasus:
Maybe gdal and mlpack needs a revbump after this?

PR #5447, opened 2020-11-26
Blis: bump

extrowerk:


Begasus:
Checked with numpy and scipy?

extrowerk:
This should be ok.

PR #360, opened 2015-12-21
Reformatted 10 recipes

dash102:
https://codein.withgoogle.com/dashboard/task-instances/4796400818716672/?sp-page=1

PR #5454, opened 2020-11-28
otter_browser: bump

kenmays:
Otter Browser - now provides Opera 12.18/Safari 14.0/Chrome 86/Firefox 82 user agent support.


diversys:
Huh? Opera 12.18 compatibility?

kenmays:
Yes... there was a Opera 12.18 rel;ease... latest is Opera 74.0.

extrowerk:
Do you mean the Opera 12.18 was added to the user-selectable user-agent list or is it something else? The commit log in the otter repo doesn't mentions anything like that. Could you express yourself ***less*** cryptic, maybe even with complete sentences? 

Also the last release is 1.0.01, why do you name the recipe as a release if it isn't one? (~git missing from the recipe name)

kenmays:
1. Yes. the user agent mask is at Opera 12.18

2. "Otter Browser aims to ****recreate** the best aspects of the **classic Opera (12.x)**** UI using Qt5".
- This means 're-engineered' or 'based on' or 'copied from'  Opera 12.x (in which there was a Opera 12.18) release (07/2016).

3. There is a v1.0.81.333 snapshot.... so this follows the naming convention based on the **weekly development report** releases.
which is at **#360** (2020-11-23) report - not a official dev snapshot source/binary/release.. If you look at GIT, you don;t see the **1.0.81.333** tagging either - just the release at sourceforge on July 2020 (same as the **weekly report #333**)..

extrowerk:
Builds on 64 bit, but i see no worth in the Opera user agent as one can easily add *any* custome one in the current version. This is not a good reason to bump to a git version.

kenmays:
OK. Wait till next stable release or as needed. Closing.

PR #5459, opened 2020-11-29
GSL: bump

extrowerk:


Begasus:
Checked with algol68g, xaos, krita and calligra?

PR #5477, opened 2020-12-08
Readline: bump

extrowerk:


app-shells/bash/bash-5.1.recipe
```diff
@@ -0,0 +1,87 @@
...
+	lib:libhistory
+	lib:libreadline
+	lib:libncursesw
+	lib:libintl
+	lib:libiconv
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libhistory >= 8
```
korli, line 37: is 8.1 not needed?

app-shells/bash/bash-5.1.recipe
```diff
@@ -0,0 +1,87 @@
...
+PROVIDES="
+	bash = $portVersion
+	cmd:bash = $portVersion
+	cmd:bashbug = $portVersion
+	cmd:sh
+	"
+REQUIRES="
+	haiku
+	lib:libhistory
+	lib:libreadline
```
korli, line 29: Could you check that bash gets a dependency on 8.1? that's how bash is packaged, even if 8.1 is compatible with 8.0

extrowerk: ~> objdump -x /boot/system/bin/bash | grep NEEDED
  NEEDED               libreadline.so.8
  NEEDED               libhistory.so.8
  NEEDED               libncurses.so.6
  NEEDED               libintl.so.8

korli: technically right, in reality bash 5.1 is to be deployed with readline 8.1: this is what is tested by bash developers.

See example in Gentoo: 
https://gitweb.gentoo.org/repo/gentoo.git/tree/app-shells/bash/bash-5.1.ebuild#n37
https://gitweb.gentoo.org/repo/gentoo.git/tree/app-shells/bash/bash-5.1.ebuild#n56

PR #5491, opened 2020-12-13
mbedtls, bump version to 2.25.0

Begasus:


net-libs/mbedtls/mbedtls-2.25.0.recipe
```diff
@@ -0,0 +1,143 @@
+SUMMARY="Open source, portable, easy to use, readable and flexible SSL library"
+DESCRIPTION="mbed TLS (formerly known as PolarSSL) makes it trivially easy for \
+developers to include cryptographic and SSL/TLS capabilities in their (embedded) \
+products, facilitating this functionality with a minimal coding footprint."
+HOMEPAGE="https://tls.mbed.org/"
+COPYRIGHT="2006-2018 ARM Limited"
+LICENSE="Apache v2"
+REVISION="1"
+SOURCE_URI="https://github.com/ARMmbed/mbedtls/archive/mbedtls-$portVersion.tar.gz"
```
korli, line 9: with this URL SOURCE_DIR is necessary. with https://github.com/ARMmbed/mbedtls/archive/v2.25.0.tar.gz SOURCE_DIR ist not.

Begasus: Need to check, but I don't think I changed the SOURCE_URI from previous recipe, but will change, much better :+1:

net-libs/mbedtls/mbedtls-2.25.0.recipe
```diff
@@ -0,0 +1,143 @@
...
+libtlsVersion="12"
+libx509Version="0"
+
+PROVIDES="
+	mbedtls$secondaryArchSuffix = $portVersion
+	lib:libmbedcrypto$secondaryArchSuffix = $libcryptoVersion compat >= $libcryptoVersion
+	lib:libmbedtls$secondaryArchSuffix = $libtlsVersion compat >= $libtlsVersion
+	lib:libmbedx509$secondaryArchSuffix = $libx509Version compat >= $libx509Version
+	"
+if [ -z "$secondaryArchSuffix" ]; then
```
Begasus, line 27: Not at the laptop atm, but thinking of moving the binaries to a "tools" package, any thoughts? (if not I'll proceed) ;)

net-libs/mbedtls/mbedtls-2.25.0.recipe
Begasus: On that note, SOURCE_URI maybe isn't needed, but then we need to use SOURCE_FILENAME :) (but granted, the link on github shows v$portVersion)

korli:
the latest 2.16.x is 2.16.9.

korli:
About 2.25.0, it's not compatible with 2.16.x, please explicitly write this in the PR description. When this is merge, a few packages will need a rebuild.

Begasus:
Checked out 2.16.8 and 2.16.9 that cary parse errors so I don't think they should be updated for gcc2 (2.16.7 seems to have some vulnerability according to repoligy, so didn't check that one
2.25.0 breaks compatibility so will check that when things are in order (before pushing)

Begasus:
@korli so far I haven't seen much recipes making use of the library?

korli:
@Begasus  OK then you can drop gcc2 and 2.16.x

net-libs/mbedtls/mbedtls-2.25.0.recipe
Begasus: @korli removed the 2.16 with this PR now, should I change the SOURCE_URI still?

korli: Thanks. It's OK as is.

net-libs/mbedtls/mbedtls-2.25.0.recipe
```diff
@@ -0,0 +1,151 @@
...
+PATCHES="mbedtls-$portVersion.patchset"
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="$portVersion"
+
+PROVIDES="
+	mbedtls$secondaryArchSuffix = $portVersion
+	lib:libmbedcrypto$secondaryArchSuffix = $libVersion compat >= 6
```
korli, line 21: this resolves to "2.25.0 compat >= 6". It doesn't look correct to me.

Begasus: Wasn't sure about this, provides .so libs are 2.25.0 and 6

korli: Then "make" a correct version: "6.25.0 compat >= 6"
The point is to be able to make the differences when the major versions change (ie for 2.16.x it's 3, 12 and 0)
The alternative is to use the minor version: 2.25.0 compat >= 2.25, but one can be sure the compat won't break in 2.25.x

Begasus: Like in: libVersion="6.25.0" and set the compat to these lines?

net-libs/mbedtls/mbedtls-2.25.0.recipe
```diff
@@ -0,0 +1,160 @@
...
+
+ARCHITECTURES="!x86_gcc2 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersionmbedCrypto="6.25.0"
+libVersionembedTls="13.25.0"
+libVersionembedX509="1.25.0"
+libVersionembedCryptoCompat="$libVersionmbedCrypto compat >= ${libVersionmbedCrypto%%.*}"
+libVersionembedTlsCompat="$libVersionembedTls compat >= ${libVersionembedTls%%.*}"
+libVersionembedX509Compat="$libVersionembedX509 compat >= ${libVersionembedX509%%.*}"
```
Begasus, line 22: @korli is this what you had in mind?

korli: @Begasus yes

PR #5492, opened 2020-12-14
haikuports: Lint licensing cleanup

kenmays:
Fixes Unknown licensing reported by lint.
NOTE: Used MIT licensing as default for haiku-* applications.


waddlesplash:
You can't just "assign" licenses, that's not how this works.

PR #5509, opened 2020-12-23
Medo, revbump, builds on 32bit, but crashes on launch

Begasus:


X547:
Ticket about crash with debug report should be created in https://github.com/smallstepforman/Medo/issues.

Begasus:
Will do later :) busy atm

PR #5518, opened 2020-12-24
libxslt, fix build for gcc2

Begasus:


Begasus:
Thanks for the sed help from augiedoggie!

PR #5519, opened 2020-12-24
OpenCV: bump

extrowerk:


extrowerk:
I think this should be ok to merge.

PR #5530, opened 2020-12-27
VLC: enable libass, cleanup

extrowerk:
*UNTESTED, unable to test, because "waiting for build package". Removing the worl folder doesn't helps.
Somebody please fix haikuporter or package daemon, this is a really annoying bug.


extrowerk:
Unchanged recipe works doesn't exhibits this problem, idk what am i doing wrong.

PR #5533, opened 2020-12-28
libimagequant, new recipe

Begasus:


Begasus:
Needed for upcoming new TuxPaint

PR #5539, opened 2020-12-29
tmux: update to 3.1c

zdykstra:
```
$ tmux -V
tmux 3.1c
```

Built and tested on a recent x86_64 nightly. Are there additional steps I should be taking to test/validate that this builds everywhere?


Begasus:
You forgot to remove the previous patchset also, will check a build on x86_gcc2 here

PR #5544, opened 2020-12-30
tmux: use preferred Haiku config path

zdykstra:
Instead of looking for a non-existent config file in the packages directory and then falling back to dot files in ~, `tmux` now looks first for `tmux.conf` in `B_USER_SETTINGS_DIRECTORY`. 

Tested locally on hrev54812/x86_64.


korli:
Could you patch the current code at https://github.com/tmux/tmux/ ? The Haiku patch has been merged a bit differently, so it's better to base your work on the current upstream git.

zdykstra:
> Could you patch the current code at https://github.com/tmux/tmux/ ? The Haiku patch has been merged a bit differently, so it's better to base your work on the current upstream git.

I'll see if I can get a patch set for upstream that they'll accept, yup! It could be tough with the man page changes, there's probably no reasonable way to 'fix' that upstream. Is the policy then to carry the patch/patches in haikuports until upstream includes it in a release?

korli:
> I'll see if I can get a patch set for upstream that they'll accept, yup! It could be tough with the man page changes, there's probably no reasonable way to 'fix' that upstream. Is the policy then to carry the patch/patches in haikuports until upstream includes it in a release?

OK, forget it. The Haiku upstream patch is planned for 3.3. It won't get released before mid-2021, it seems.

app-misc/tmux/patches/tmux-3.1c.patchset
```diff
@@ -176,22 +188,70 @@ index 4d54353..34f1a23 100644
...
+diff --git a/Makefile.am b/Makefile.am
+index 082a467..a867aa2 100644
+--- a/Makefile.am
++++ b/Makefile.am
+@@ -13,7 +13,7 @@ dist_EXTRA_tmux_SOURCES = compat/*.[ch]
+ # Preprocessor flags.
+ AM_CPPFLAGS += @XOPEN_DEFINES@ \
+ 	-DTMUX_VERSION="\"@VERSION@\"" \
+-	-DTMUX_CONF="\"$(sysconfdir)/tmux.conf:~/.tmux.conf:~/.config/tmux/tmux.conf\""
++	-DTMUX_CONF="\"$(shell finddir B_USER_SETTINGS_DIRECTORY)/tmux.conf:~/.tmux.conf:~/.config/tmux/tmux.conf\""
```
korli, line 210: sysconfdir should still be first.

app-misc/tmux/patches/tmux-3.1c.patchset
```diff
@@ -176,22 +188,70 @@ index 4d54353..34f1a23 100644
...
- 		[
- 			#include <sys/types.h>
+ # Additional object files.
+ LDADD = $(LIBOBJS)
+@@ -198,7 +198,7 @@ endif
+ # Install tmux.1 in the right format.
+ install-exec-hook:
+ 	if test x@MANFORMAT@ = xmdoc; then \
+-		sed -e "s|@SYSCONFDIR@|$(sysconfdir)|g" $(srcdir)/tmux.1 \
++		sed -e "s|@SYSCONFDIR@|$(shell finddir B_USER_SETTINGS_DIRECTORY)|g" $(srcdir)/tmux.1 \
```
korli, line 219: Why is this needed?

korli:
@SYSCONFDIR@/tmux.conf should still be used. Only the Home settings file is to be replaced, or the user settings dir searched first.

zdykstra:
I'll re-work this patch based on the comments left here.

app-misc/tmux/patches/tmux-3.1c.patchset
zdykstra: It's needed so that the man page showed the correct paths based on the other changes included in the patch.

zdykstra:
Just to confirm, by default tmux will look for a configuration file at `/packages/tmux-3.1c-2/.settings/tmux.conf` if I leave `@SYSCONFDIR@/tmux.conf` as-is. Is that the preferred behavior, or should it look somewhere else first?  The next location in the search path is `/boot/home/config/settings/tmux.conf`.

korli:
> Just to confirm, by default tmux will look for a configuration file at `/packages/tmux-3.1c-2/.settings/tmux.conf` if I leave `@SYSCONFDIR@/tmux.conf` as-is. Is that the preferred behavior, or should it look somewhere else first? The next location in the search path is `/boot/home/config/settings/tmux.conf`.

This is correct (see https://github.com/haikuports/haikuporter/blob/2b7ba23e93e799bb9ceeab552266cc1e35b26c5f/HaikuPorter/Port.py#L1173)

zdykstra:
I've adjusted the patch to leave @SYSCONFIG@ as the first location, and to then use `$(shell finddir B_USER_SETTINGS_DIRECTORY)/tmux.conf` as the next. The remainder of the search locations are left as-is. The man page has also been updated to fix references to `~/.tmux.conf` and instead refer to them as `$(shell finddir B_USER_SETTINGS_DIRECTORY)/tmux.conf` .  

This work has been tested locally on a recent x86_64 nightly build.

korli:
Thanks!

PR #5569, opened 2021-01-04
astroid: disable package

tommyjanna:
Made a mistake when pushing this package yesterday (#5565).

It's missing dependencies which I ignored that aren't in the repo yet... Will work on fixing these. 

Sorry for the mistake!


extrowerk:
thx

PR #5571, opened 2021-01-04
lazy-object-proxy: new recipe

tommyjanna:


dev-python/lazy-object-proxy/lazy_object_proxy-1.5.2.recipe
```diff
@@ -0,0 +1,64 @@
...
+	eval "PROVIDES_$pythonPackage=\"
+		${portName}_$pythonPackage = $portVersion
+		\""
+	eval "REQUIRES_$pythonPackage=\"
+		haiku
+		cmd:python$pythonVersion\
+		\""
+	BUILD_REQUIRES="$BUILD_REQUIRES
+		setuptools_$pythonPackage
+		cmd:git"
```
korli, line 42: cmd:git shouldn't be required.

tommyjanna: Hmm, yeah I was thinking the same thing. I tried removing it again and read through the error message of the failed build a bit more carefully and found it was asking for `setuptools_scm*`. Is this fix correct? I've pushed this onto the branch. (Had to disable python2.7 though as it lacks a `setuptools_scm`).

Thanks!

PR #5573, opened 2021-01-05
wrapt: new recipe

tommyjanna:


dev-python/wrapt/wrapt-1.12.1.recipe
```diff
@@ -0,0 +1,62 @@
+SUMMARY="Python module for decorators, wrappers and monkey patching"
+DESCRIPTION="The aim of the wrapt module is to provide a transparent object \
+proxy for Python, which can be used as the basis for the construction of \
+function wrappers and decorator functions."
+HOMEPAGE="https://github.com/GrahamDumpleton/wrapt/archive/1.12.1.tar.gz"
```
extrowerk, line 5: You could use $portVersion instead of hardcoding the version number

extrowerk: Also this is for the homepage, not for the tarball.

dev-python/wrapt/wrapt-1.12.1.recipe
```diff
@@ -0,0 +1,62 @@
...
+COPYRIGHT="2013-2019, Graham Dumpleton"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/GrahamDumpleton/$portName/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="19f2043e05b76ce9b68971b387bfd8eed50c2ee1c9c0bd230edc422fec753917"
+SOURCE_DIR="$portVersionedName"
```
extrowerk, line 11: Shouldn't it be "wrapt-$portVersion"? @Begasus?

dev-python/wrapt/wrapt-1.12.1.recipe
```diff
@@ -0,0 +1,62 @@
...
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+PYTHON_PACKAGES=(python python36 python3)
+PYTHON_VERSIONS=(2.7 3.6 3.7)
```
extrowerk, line 27: I think you could safely remove 2.7 already, we are trying to decomission it.

extrowerk:
Builds fine and importing wrapt works ok on 64 bit and with python3, others untested.

tommyjanna:
Yup, you're right. I wasn't paying attention when I filled out that homepage.

Good to know you're phasing out Python2, I removed it.

> 
> Builds fine and importing wrapt works ok on 64 bit and with python3, others untested.
Tested myself on 32-bit. It's working :)

`$portVersionedName` should be correct, although you guys know the guidelines and what not... I can wait to see if the change is suggested.

dev-python/wrapt/wrapt-1.12.1.recipe
Begasus: As this is "any" architecture it should be ok (need to check), if this was meant for secondary architecture it wont work (renames the source dir to wrapt_x86 then

PR #5576, opened 2021-01-05
Blender: bump

extrowerk:
WIP, could be useful  to test MESA.


extrowerk:
Be warned: it doesn't works with the current MESA, not even with the  common MESA overrides:
"GPUTexture: texture create failed"

waddlesplash:
We now have Blender 3 in the tree.

Begasus:
> We now have Blender 3 in the tree.

Only for 64bit, build failed on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/2865/builds/78213.log

PR #5577, opened 2021-01-06
TBB: bump, castrated oldrecipe , switch to cmake

extrowerk:
Compiling tests fails, so this is completely untested.
TBB is an important recipe, so we should fix it, but it is over my skill level.


Begasus:
Build is ok on 32-bit also, can you share your thoughts on the failing tests? (at this moment no tests are performed and didn't investigate further atm)

Begasus:
For now builds OK on 64bit R1B4, testcase if prety good: 98% tests passed, 2 tests failed out of 129

One related to a version string?
```
121/129 Test #121: test_malloc_compliance ...................***Failed    0.00 sec
[doctest] doctest version is "2.3.5"
[doctest] run with "--help" for options
Can't set limits: errno -2147483643
```

And the other one:
```
118/129 Test #118: test_malloc_init_shutdown ................Subprocess aborted***Exception:   2.01 sec
terminate called after throwing an instance of 'doctest::detail::TestFailureException'
Abort
[doctest] doctest version is "2.3.5"
[doctest] run with "--help" for options
===============================================================================
/sources/oneTBB-2021.1.1/test/tbbmalloc/test_malloc_init_shutdown.cpp:174:
TEST CASE:  test key dtor

/sources/oneTBB-2021.1.1/test/tbbmalloc/test_malloc_init_shutdown.cpp:141: FATAL ERROR: REQUIRE( (!prevSmall || currSmall==prevSmall) ) is NOT correct!
  values: REQUIRE( false )
  logged: Possible memory leak

/sources/oneTBB-2021.1.1/test/tbbmalloc/test_malloc_init_shutdown.cpp:174: FATAL ERROR: test case CRASHED: SIGABRT - Abort (abnormal termination) signal

===============================================================================
[doctest] test cases:      3 |      2 passed |      1 failed |      0 skipped
[doctest] assertions:      8 |      7 passed |      1 failed |
[doctest] Status: FAILURE!
```
Checking some builds at 32bit first before cleaning up and merging.

extrowerk:
I suspect it is the "Can't set limits: errno -2147483643"
You can use the error command in the Terminal to find out what the errno means.
If i recall it correctly, Haiku doesn't lets the user to change the limits on the fly, but i can be wrong. No idea if it needs any handling from Haiku side or if it is safe to ignore this.

The malloc seems more problematic, it seems. I have no idea about that, but I always had issues with malloc in TBB. Maybe we should simply drop the TBB port, AFAIK I have disabeld it in every port...

--Z--


Schrijvers Luc írta 2024. jún.. 24, H-n 09:47 órakor:

Begasus:
From the limits error I get: `0x80000005: Invalid Argumen`
On 32bit with disabled tests it builds OK, but with enabled I see a lot of undefined references.
FWIW we could just merge this as disabled for references, for latest release other/more fixes are needed.

Begasus:
> The malloc seems more problematic, it seems. I have no idea about that, but I always had issues with malloc in TBB. Maybe we should simply drop the TBB port, AFAIK I have disabeld it in every port...

Still in use by some it seems: https://github.com/search?q=repo%3Ahaikuports%2Fhaikuports%20libtbb&type=code

Begasus:
OK, pushed another commit to disable the recipe untill someone can do actual tests on it. Thanks! (been around for a while now) :)

jmairboeck:
I could test it with opencascade, for example. (but not too soon, I'm quite occupied at the moment)

see: https://github.com/haikuports/haikuports/blob/c4309c370f32c01e4977f1ef05afd49632004b66/sci-libs/opencascade/opencascade-7.8.1.recipe#L372-L377

Begasus:
Seems to require at least a newer version there, fwiw, disable the tests from building, saves a lot of time for tbb. :)

```
CMake Error at adm/cmake/tbb.cmake:169 (find_package):
  Could not find a configuration file for package "TBB" that is compatible
  with requested version "2021.5".

  The following configuration files were considered but not accepted:

    /boot/system/lib/cmake/TBB/TBBConfig.cmake, version: 2021.1.1

Call Stack (most recent call first):
  CMakeLists.txt:34 (include)
  CMakeLists.txt:742 (OCCT_INCLUDE_CMAKE_FILE)
```

Begasus:
OK, got 2021.12.0 up and build, doing a checkbuild with opencascade (64bit) with some fixes and cmake options from Fedora.

jmairboeck:
> Bumped to 2021.12.0 and ran a check on opencascade, found (in the end). :)
> 
> ```
> Warning: POLICY WARNING: "/packaging/opencascade_tools/bin/DRAWEXE" needs library "libtbb.so.12", but the package doesn't seem to declare that as a requirement
> ```

This looks good, it means that DRAWEXE is linked to TBB :)

Begasus:
Base libraries, samples and tools all linked to TBB, now how can I be sure it's working as it should? ;)

jmairboeck:
I don't know exactly. I'm not even sure, if it is a good idea to use TBB for OpenCASCADE. I would have to look at its documentation whether there are any advantages of using it. For testing OpenCASCADE it is probably easiest to just run the tests. You don't have to run them fully though, the test suite is quite large ;)

Begasus:
OK, seems that tcm thing has been added after 2021.10.0, checking with 2021.7.0 gives the same result in Terminal as the current one.
```
~> DRAWEXE
 the CASROOT variable is mandatory to Run OpenCascade 
No default file
% % 
```

Begasus:
Samples up and running, only OCCTOverview crashes on launch

jmairboeck:
> Samples up and running, only OCCTOverview crashes on launch

This is probably because the OpenGL visualization is just a `nullptr` "stub". I couldn't get that to work correctly. It would probably need a custom implementation with BGLView.

Begasus:
OK, no dealbreaker for now then, moving up to 2021.10.0 now.

Begasus:
Let the tests run for opencascade for a bit, results:

`Total cases: 2 BAD, 2 FAILED, 430 OK, 976 SKIPPED`

Begasus:
@jmairboeck @extrowerk can continue discussing on PR: https://github.com/haikuports/haikuports/pull/10631

PR #5578, opened 2021-01-06
hwloc: bump

extrowerk:
All test passes, but only outside of haikuporter.
Kept the full old version.


extrowerk:
resolve symbol "hwloc__xml_export_topology" returned: -2147478780

extrowerk:
```
objdump -x /boot/system/lib/hwloc/hwloc_xml_libxml.so | grep hwloc__xml_export_topology
0000000000000000         *UND*  0000000000000000              hwloc__xml_export_topology
```

extrowerk:
Seems to be ok now

PR #5582, opened 2021-01-06
Coreutils: ls: fix hiding of files (nephele)

extrowerk:


pulkomandy:
To be precise, what this does is stop hiding dotfiles by default.

It matches Tracker behavior, but I'm undecided if it's a good idea. Could it break some software using ls and expecting it to ignore dotfiles by default?

zdykstra:
I recognize that I'm new here, but this doesn't seem like a good idea to implement. The behavior of `ls` from coreutils is fairly standardized across systems that use it. It seems imprudent to diverge the behavior of `ls` from what people have come to expect, simply because Tracker shows dot-files/directories by default. 

This also gets rid of the `-A` flag - why?

augiedoggie:
> To be precise, what this does is stop hiding dotfiles by default.

It doesn't just change the default though, it makes it impossible to get a normal listing that doesn't have dotfiles.  As far as I know there is no command line switch that reverses this.

ahesford:
This is a terrible idea when `ls -A` gets you the output you want.

pulkomandy:
This removes the -A flag because -A would be the same as the default (and in fact it doesn't remove it, it just makes it not have any effect)

It turns out POSIX in fact allows this (I thought it wouldn't): "Filenames beginning with a <period> ( '.' ) and any associated information shall not be written out unless explicitly referenced, the -A or -a option is supplied, or an implementation-defined condition causes them to be written." we are in the "implementation defined condition" case here (with the condition being "we just always show them").

So, I now have no arguments against it. Still somewhat unconvinced it's a good idea however.

ahesford:
Has nobody heard of shell aliases?

zdykstra:
This seems like an undue burden on maintainers when `alias ls='ls -A'` would accomplish the exact same thing at effectively no cost / be able to be reverted by people that don't want this.

waddlesplash:
In fact we already alias `ls` by default: https://github.com/haiku/haiku/blob/master/data/etc/profile#L29

So, indeed, I don't see a reason to do this. If we really want, we can modify the default alias of `ls` instead.

PR #5583, opened 2021-01-06
OpenMPI: bump

extrowerk:
All test passes.


extrowerk:
Only scalapack depends on it, but thats also disabled for every arch.

PR #5592, opened 2021-01-08
sane-backends: bump to 1.0.31

kenmays:


Vidrep:
Successfully built package on 64 bit nightly. Tested with 2 different USB scanners, OK.

PR #5566, opened 2021-01-04
Kristall: UI changes, view page source, new shortcuts, fixes for emojis and more

win8linux:
This updates Kristall to master as of commit https://github.com/MasterQ32/kristall/commit/ba7f39e9f0f329e78710c67e1267c1fbc79ac428. Some of the notable changes involve UI additions and tweaks, the addition of viewing page source, new shortcuts, and a fix for emojis that will hopefully be more permanent.

More detailed breakdown:

- [Add 'view source' to right-click context menu](https://github.com/MasterQ32/kristall/commit/237d3a8bc9ed6be7196115e0c7b640ca59349b1d)
- Shortcuts for [View page source](https://github.com/MasterQ32/kristall/commit/06b254d686a9fbed702e8d1777dc5c7adcba1fbe), [Settings](https://github.com/MasterQ32/kristall/commit/2ccc644f98019885562c9c5f46d27be39595ff06), [Quit](https://github.com/MasterQ32/kristall/commit/bf7e6a8a03f6a623b342d08c6e3202640aa40aba),  [Document Outline, Bookmark, and History docks](https://github.com/MasterQ32/kristall/commit/1ceae995f1984b2e9512752cf5e8d8b56d03d574)
- [Theme changes reflected instantly upon confirmation](https://github.com/MasterQ32/kristall/commit/9ee987f8216b412957ffcb46c01bbe268b282746)
- [Use system icon theme when using OS Default theme](https://github.com/MasterQ32/kristall/commit/ee1ad97ec70717f77e1e420fdfa2565ce3fa13c0)
- [URL bar styling](https://github.com/MasterQ32/kristall/commit/eabbbebed714f4d54f57fb9065ecbeb303f08f9b) and its respective [preference](https://github.com/MasterQ32/kristall/commit/40df924bc7289155a21a7c4ee901fc7c5e1673bd)
- [UI density option](https://github.com/MasterQ32/kristall/commit/19c8bf98885074617f9cc0a0b2f60ebf47ddf714)
- [Resolves the emoji problem, hopefully once and for all!](https://github.com/MasterQ32/kristall/commit/808288ce691a3c1520db513c7cff79fd22d3c53f)

PR #5607, opened 2021-01-17
Kristall: Bump to 3

win8linux:
Ok, second attempt with much less commits.

Forgot to bump the revision for https://github.com/haikuports/haikuports/pull/5566, so this PR will remedy that oversight.


Begasus:
After a merge it's best to clean (delete) the current branch (in this case branch kristall) both local and remote, and keep your master on par with upstream, for future revbumps just create a new branch then en push for a PR (check with git status that everything is clean before making changes in the new branch) :)

Begasus:
64 bit is OK, 32 bit failed (but is known with the "out of memory" issue), thanks!

win8linux:
Much appreciate the tips for future PRs! As for 32-bit, perhaps cross-compilation from a 64-bit host might be an option?

Begasus:
Don't think much users of 32 bit are busy with that, or even know if it's possible :)

PR #5614, opened 2021-01-17
avisynthplus-3.7.0: new recipe

qyot27:
Hopefully I did this right.  It completes successfully for me on 64-bit, but on 32-bit it kept failing when trying to populate dependencies because it didn't seem to be recognizing $secondaryArchSuffix at all (or in other cases, just erroring out with a 'is broken on this architecture'; I'm hoping that maybe that's just an issue with my local configuration and not with the recipe).

The patch section there is something we'll address upstream.

### DIFF
--- a/media-libs/avisynthplus/avisynthplus-3.7.0.recipe
+++ b/media-libs/avisynthplus/avisynthplus-3.7.0.recipe
@@ -1,0 +1,67 @@
+SUMMARY="A frameserver controlled by scripting"
+DESCRIPTION="A powerful nonlinear scripting language for media. Includes a set \
+of core filters for editing, pre-production and post-production, and the \
+ability to extend its capabilities through the use of external plugins."
+HOMEPAGE="https://avs-plus.net"
+COPYRIGHT="2000-2015 Ben Rudiak-Gould, et al.
+	2013-2021 AviSynth+ Project"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/AviSynth/AviSynthPlus/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="8906d9e46dc90a194413d69b710e3106c0aafddc0c5c62004885d0c3beb79862"
+SOURCE_FILENAME="avisynthplus-$portVersion.tar.gz"
+SOURCE_DIR="AviSynthPlus-$portVersion"
+PATCHES="0001-CMakeLists-fix-setting-local-plugindir.patch
+	0002-avisynth_conf.h.in-use-mounted-system-libdir-on-Haik.patch
+	"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= 8"
+
+PROVIDES="
+	avisynthplus$secondaryArchSuffix = $portVersion
+	lib:libavisynth$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libIL$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	avisynthplus${secondaryArchSuffix}_devel = $portVersion
+	devel:libavisynth$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	avisynthplus$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libIL${secondaryArchSuffix}
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	mkdir -p avisynth-build
+	cd avisynth-build
+	cmake ../ $cmakeDirArgs -DCMAKE_CXX_FLAGS="-DRELEASE_TARBALL"
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make -C avisynth-build install
+
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
+}
--- a/media-libs/avisynthplus/avisynthplus-3.7.0.recipe
+++ b/media-libs/avisynthplus/avisynthplus-3.7.0.recipe
@@ -18,7 +18,7 @@ PATCHES="0001-CMakeLists-fix-setting-local-plugindir.patch
 ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
 SECONDARY_ARCHITECTURES="x86"
 
-libVersion="$portVersion"
+libVersion="8"
 libVersionCompat="$libVersion compat >= 8"
 
 PROVIDES="
--- a/media-libs/avisynthplus/patches/0001-CMakeLists-fix-setting-local-plugindir.patch
+++ b/media-libs/avisynthplus/patches/0001-CMakeLists-fix-setting-local-plugindir.patch
@@ -1,0 +1,41 @@
+From 20377502a81eefa9356b04174a12508133ed733e Mon Sep 17 00:00:00 2001
+From: Stephen Hutchinson <qyot27@gmail.com>
+Date: Sun, 17 Jan 2021 14:11:56 -0500
+Subject: [PATCH 1/2] CMakeLists: fix setting local plugindir
+
+---
+ CMakeLists.txt                                                | 4 +---
+ .../docs/english/source/avisynthdoc/contributing/posix.rst    | 2 +-
+ 2 files changed, 2 insertions(+), 4 deletions(-)
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 9ea2c954..7da15fbb 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -66,9 +66,7 @@ if(NOT HEADERS_ONLY)
+ 
+   option(ENABLE_PLUGINS "Build set of default external plugins" ON)
+   option(ENABLE_INTEL_SIMD "Enable SIMD intrinsics for Intel processors" "${INTEL_SIMD}")
+-  option(LOCAL_PLUGINDIR "Override path for user-local plugins" "${USER_AVS_PLUGINDIR_LOCATION}")
+-
+-  set(USER_AVS_PLUGINDIR_LOCATION "$ENV{HOME}/.local/lib/avisynth")
++  set(USER_AVS_PLUGINDIR_LOCATION "$ENV{HOME}/.local/lib/avisynth" CACHE STRING "Override path for user-local plugins (default: $HOME/.local/lib/avisynth)")
+ 
+   if(CMAKE_CONFIGURATION_TYPES)
+     set(CMAKE_CONFIGURATION_TYPES Debug Release RelWithDebInfo)
+diff --git a/distrib/docs/english/source/avisynthdoc/contributing/posix.rst b/distrib/docs/english/source/avisynthdoc/contributing/posix.rst
+index 43a4e71c..c7e51303 100644
+--- a/distrib/docs/english/source/avisynthdoc/contributing/posix.rst
++++ b/distrib/docs/english/source/avisynthdoc/contributing/posix.rst
+@@ -373,7 +373,7 @@ Plugin autoloading
+ 
+ AviSynth+ will use several directories for autoloading:
+ the `avisynth/` subdirectory where libavisynth.so was installed,
+-`$HOME/.avisynth`, and the directory given to the LOCAL_PLUGINDIR
++`$HOME/.avisynth`, and the directory given to the USER_AVS_PLUGINDIR_LOCATION
+ configuration option (defaults to `$HOME/.local/lib/avisynth`).
+ The latter of which can hold plugins (and symlinks to plugins)
+ or AVSI files without needing root permissions.
+-- 
+2.30.0
+
--- a/media-libs/avisynthplus/patches/0002-avisynth_conf.h.in-use-mounted-system-libdir-on-Haik.patch
+++ b/media-libs/avisynthplus/patches/0002-avisynth_conf.h.in-use-mounted-system-libdir-on-Haik.patch
@@ -1,0 +1,33 @@
+From acd6cab76425fed45a300822cd41f30548948d44 Mon Sep 17 00:00:00 2001
+From: Stephen Hutchinson <qyot27@gmail.com>
+Date: Mon, 18 Jan 2021 04:41:52 -0500
+Subject: [PATCH 2/2] avisynth_conf.h.in: use mounted system libdir on Haiku
+
+---
+ avs_core/avisynth_conf.h.in | 10 +++++++++-
+ 1 file changed, 9 insertions(+), 1 deletion(-)
+
+diff --git a/avs_core/avisynth_conf.h.in b/avs_core/avisynth_conf.h.in
+index 57bc107f..cdf96f19 100644
+--- a/avs_core/avisynth_conf.h.in
++++ b/avs_core/avisynth_conf.h.in
+@@ -5,7 +5,15 @@
+ 
+ #ifdef AVS_POSIX
+ #define user_avs_plugindir_configurable "@USER_AVS_PLUGINDIR_LOCATION@"
+-#define system_avs_plugindir "@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_LIBDIR@/avisynth"
++    #ifdef AVS_HAIKU
++        #ifdef X86_32
++            #define system_avs_plugindir "/system/lib/x86/avisynth"
++        #else
++            #define system_avs_plugindir "/system/lib/avisynth"
++        #endif
++    #else
++        #define system_avs_plugindir "@CMAKE_INSTALL_PREFIX@/@CMAKE_INSTALL_LIBDIR@/avisynth"
++    #endif
+ #endif
+ 
+ #endif // _AVS_CONF_H_
+-- 
+2.30.0
+


extrowerk:
Probably you dud not enabled the secondery arch support in the haikuporter config file. Check it again.

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="!x86_gcc2 x86"
+
+PROVIDES="
+	avisynthplus$secondaryArchSuffix = $portVersion
+	lib:libavisynth = $portVersion
+	lib:libconvertstacked = $portVersion
+	lib:libtimestretch = $portVersion
+	lib:libimageseq = $portVersion
+	lib:libshibatch = $portVersion
```
Begasus, line 23: Need to check, but are you sure these don't need $secondaryArchSuffix?

extrowerk:
Also for building the package xyz on 64 bit you call it like “hp xyz”, but on 32 bit for secondary arch like “hp xyz_x86”

Begasus:
> The patch section there is something we'll address upstream.

Is it needed to complete the build?

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
qyot27: Upon checking, it seems that they do.  I wasn't sure if that was a requirement if the package name had that variable set.

Begasus: When using secondary architecture on 32-bit cmakedirargs "should" install library in $libDir, for 32-bit this is $prefix/lib on primary, $prefix/lib/x86 on secondary architecture, hence the need for $secondaryArchSuffix

Begasus, line 23: Those libraries are in a subfolder in $libDir, no need to add them here

Begasus:
Made some local changes and pushed them to https://github.com/Begasus/haikuports/commit/ba655debada329c1041d0f8f3a84112ed8dd5492
Not sure about using libVersion(Compat) there, but could be a good idea

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+{
+	sed -i '8i#ifdef AVS_HAIKU\n#define system_avs_plugindir "/system/lib/avisynth"\n#else' avs_core/avisynth_conf.h.in
+	sed -i '12i#endif' avs_core/avisynth_conf.h.in
+}
+
+BUILD()
+{
+	mkdir -p avisynth-build
+	cd avisynth-build
+	cmake ../ $cmakeDirArgs -DCMAKE_CXX_FLAGS="-DRELEASE_TARBALL"
```
Begasus, line 50: Any reason for -DCMAKE_CXX_FLAGS="-DRELEASE_TARBALL"? Haven't seen it passing by in other builds

qyot27:
>Probably you dud not enabled the secondery arch support in the haikuporter config file. Check it again.

I tried it both ways, but

>but on 32 bit for secondary arch like “hp xyz_x86”

That's the part I was missing.  As soon as I appended the suffix, it built (after fixing the missing suffixes on the lib: entries, anyway).

>Is it needed to complete the build?

No, but it's required for plugin autoloading to function the way it's supposed to (although on 32-bit it'll be wrong since it's pointing at the main libdir and not libdir/x86).

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
qyot27: The initialization of a git repository despite being downloaded from a tarball causes a build error because the internal versioning assumes you're either compiling from a git clone of the repository or compiling from a tarball with no .git directory.  Because the initialized directory doesn't have the necessary data in it for the git method, it fails unless tarball mode is forced.

Begasus:
> although on 32-bit it'll be wrong since it's pointing at the main libdir and not libdir/x86

Looking at the output from my local changes and pkg-config the paths look ok?
```
~> pkg-config-x86 avisynth --libs
-L/packages/avisynthplus_x86-3.7.0-1/.self/develop/lib/x86 -lavisynth
~> pkg-config-x86 avisynth --cflags
-I/packages/avisynthplus_x86-3.7.0-1/.self/develop/headers/x86
```

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
Begasus: Thanks for clarifying :+1:

qyot27:
That's not what it's doing.  External plugins being autoloaded by the core at runtime.  And when dealing with plugins not part of the core distribution, there would otherwise be no way for the core to find them if we only rely on the package libdir (a script could manually invoke AddAutoloadDir but that defeats the purpose of autoloading).

Begasus:
I've seen you have up-streamed, Haiku support there, so I'll take your word for it, not an expert in these ... only trying to help out
PS, managed to build a separate plugin with haikuporter, install paths look fine there ... (just my 2 cents), thanks for taking the time to work on this!
avisynthplus

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+	lib:libimageseq = $portVersion
+	lib:libshibatch = $portVersion
+	"
+REQUIRES="
+    haiku$secondaryArchSuffix
+    devil$secondaryArchSuffix
+    "
+
+BUILD_REQUIRES="
+    devil${secondaryArchSuffix}_devel
```
korli, line 31: it should be some devel: entries, for instance devel:libIL$secondaryArchSuffix

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+	avisynthplus$secondaryArchSuffix = $portVersion
+	lib:libavisynth = $portVersion
+	lib:libconvertstacked = $portVersion
+	lib:libtimestretch = $portVersion
+	lib:libimageseq = $portVersion
+	lib:libshibatch = $portVersion
+	"
+REQUIRES="
+    haiku$secondaryArchSuffix
+    devil$secondaryArchSuffix
```
korli, line 27: haikuporter in strict mode will try to tell which libs are really required. devil$secondaryArchSuffix isn't really required, but for instance lib:libIL$secondaryArchSuffix might be

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,59 @@
...
+BUILD_PREREQUIRES="
+    haiku${secondaryArchSuffix}_devel
+    cmd:gcc$secondaryArchSuffix
+    cmd:cmake
+    cmd:make
+    "
+
+PATCH()
+{
+	sed -i '8i#ifdef AVS_HAIKU\n#define system_avs_plugindir "/system/lib/avisynth"\n#else' avs_core/avisynth_conf.h.in
```
korli, line 42:
```suggestion
	sed -i '8i#ifdef AVS_HAIKU\n#define system_avs_plugindir "/system/lib/avisynth"\n#else' \
		avs_core/avisynth_conf.h.in
```

qyot27:
Cleaned up with several of the suggestions and force pushed.

Instead of hassling around with sed, I resorted to actually using git patches (one of which is actually just fetched from upstream).  The change for the autoload dir in avisynth_conf.h.in is what I'll end up doing upstream as well.  In that form, it does allow both x86_64 and x86 to work correctly.

@Begasus 
The devel provides are something I wasn't really sure about here, but if that makes it more correct then I'll merge that in as well.

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
Begasus: On secondary architecture it's not /system/lib but system/lib/x86 so maybe something should be changed here?

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,52 @@
...
+
+BUILD_REQUIRES="
+    devel:libIL${secondaryArchSuffix}
+    "
+BUILD_PREREQUIRES="
+    haiku${secondaryArchSuffix}_devel
+    cmd:gcc$secondaryArchSuffix
+    cmd:cmake
+    cmd:make
+    "
```
Begasus, line 37: Don't use spaces here, use a tab:
```diff
-    haiku$secondaryArchSuffix
-    lib:libIL$secondaryArchSuffix
-    "

BUILD_REQUIRES="
-    devel:libIL${secondaryArchSuffix}
-    "
BUILD_PREREQUIRES="
-    haiku${secondaryArchSuffix}_devel
-    cmd:gcc$secondaryArchSuffix
-    cmd:cmake
-    cmd:make
-    "
+	haiku$secondaryArchSuffix
+	lib:libIL$secondaryArchSuffix
+	"

BUILD_REQUIRES="
+	devel:libIL${secondaryArchSuffix}
+	"
BUILD_PREREQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	cmd:gcc$secondaryArchSuffix
+	cmd:cmake
+	cmd:make
+	"
```

Begasus:
I've seen you have up-streamed, Haiku support there, so I'll take your word for it, not an expert in these ... only trying to help out
PS, managed to build a separate plugin with haikuporter, install paths look fine there ... (just my 2 cents), thanks for taking the time to work on this!
avisynthplus


> The devel provides are something I wasn't really sure about here, but if that makes it more correct then I'll merge that in as well.

From what I've learned so far, libraries are needed at run time, develop files (headers/*.pc etc) are needed for builds, so shouldn't be needed for runtime

qyot27:
Rebased and force-pushed again, with the devel stuff included, ensuring just tabs are used, and fleshing out the description a bit.

prepareInstalledDevelLib is commented out because when testing it, it was preventing the libavisynth.so symlink from being included in the package, so avs2yuv failed to find the install.  I noticed other recipes (Lua) re-do an explicit symlinking operation in the install block; is that what I also need to do?

Begasus:
> it was preventing the libavisynth.so symlink from being included in the package

That one should install into the _devel package as others also do (at least it shows up in the _devel package here)?

pulkomandy:
prepareInstalledDevelLibs normally relies on the libraries having a soname/soversion defined.
The layout is the following:

.../lib: libraries used for runtime. These use the soname as a name, something like libfoo.so.2
.../develop/lib: used for the linker. This is where ld seach for libraries. Here the library is available without the versionning (for example: libfoo.so) so you can simply use the -lfoo linker flag and the library will be found

This setup allows us to cleanly separate the development files from the runtime ones, and making sure executables depend only on the versionned libraries. This makes it possible to have multiple versions of the libraries installed, for example for compatibility with older apps that are not able to use a newer version of the lib. But it forces developers to have only one version of the development package installed on their machine (hopefully the most recent one).

Since this setup is a bit unusual (normally development and runtime libraries are just mixed together in /lib), there isn't support for it in standard buildtools (autotools, cmake, ...). This is the reason we do this as a post-processing step in prepareInstalledDevelLib.

If possible, it would be better to make sure the libraries have a soname/soversion set, so that newer versions can be installed next to old ones later on.

Begasus:
> If possible, it would be better to make sure the libraries have a soname/soversion set, so that newer versions can be installed next to old ones later on.

Didn't go there yet, libVersion is the same as portVersion (3.7.0) but there is a linked *.so.8 version in there (together with the *.so file in _devel package), so wasn't sure if compat should be taken from libVersion or set to 8 here

qyot27:
The existing cross-platform client programs for AviSynth+ dynamically load libavisynth.so, they don't link to it.  So a standard package that lacks libavisynth.so would mean clients that can't load it unless you install the _devel package too*.

*except it didn't seem to work like this (as I said above, avs2yuv couldn't see it, even with the avisynthplus_devel package installed), is there an executable bit that gets unset by prepareInstalledDevelLib or something?

[We've set the SONAME/SOVERSION using CMake's internal PROJECT_VERSION mechanism since shortly after we started supporting Linux.](https://github.com/AviSynth/AviSynthPlus/blob/master/avs_core/CMakeLists.txt#L16).

3.7.0 = point release
8 = AVISYNTH_INTERFACE_VERSION, tells external libraries what API functionality they can expect.  Changes much less often (it's been at 8 since 3.6.0).

Begasus:
Not an expert here, maybe @korli could have a look?

pulkomandy:
In that case you don't need prepareInstalledDevelLib at all (the devel package does not need to contain any lib, since there is no need to link against something at build time).

And ideally you would move your lib to the add-ons/ directory instead of lib/, but that may create some trouble for ported applications.

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,66 @@
...
+COPYRIGHT="2000-2015 Ben Rudiak-Gould, et al., \
+	2013-2021 AviSynth+ Project"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/AviSynth/AviSynthPlus/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="8906d9e46dc90a194413d69b710e3106c0aafddc0c5c62004885d0c3beb79862"
+SOURCE_FILENAME="avisynthplus-$portVersion.tar.gz"
+SOURCE_DIR="AviSynthPlus-$portVersion"
+PATCHES="
+	0001-CMakeLists-fix-setting-local-plugindir.patch
```
Begasus, line 15: Some cosmetics (no need to start a new line here):
```diff
-PATCHES="
-	0001-CMakeLists-fix-setting-local-plugindir.patch
+PATCHES="0001-CMakeLists-fix-setting-local-plugindir.patch
```

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,66 @@
...
+	0002-avisynth_conf.h.in-use-mounted-system-libdir-on-Haik.patch
+	"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="x86"
+
```
Begasus, line 21: Maybe using libVersion in here (and further up in the recipe):
```diff
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= 8"
+
```

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,66 @@
...
+PROVIDES="
+	avisynthplus$secondaryArchSuffix = $portVersion
+	lib:libavisynth$secondaryArchSuffix = $portVersion
```
Begasus, line 24: Changes for libVersionCompat:
```diff
-	lib:libavisynth$secondaryArchSuffix = $portVersion
+	lib:libavisynth$secondaryArchSuffix = $libVersionCompat
```

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,66 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libIL$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	avisynthplus${secondaryArchSuffix}_devel = $portVersion
+	devel:libavisynth$secondaryArchSuffix = $portVersion
```
Begasus, line 33: Ditto:
```diff
- 	devel:libavisynth$secondaryArchSuffix = $portVersion
+ 	devel:libavisynth$secondaryArchSuffix = $libVersionCompat
```

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,66 @@
...
+	cd avisynth-build
+	cmake ../ $cmakeDirArgs -DCMAKE_CXX_FLAGS="-DRELEASE_TARBALL"
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make -C avisynth-build install
+
+#	prepareInstalledDevelLib libavisynth
```
Begasus, line 61: As the library isn't moved into the devel package you can remove this line:
```diff
- #	prepareInstalledDevelLib libavisynth
```

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,66 @@
+SUMMARY="A frameserver controlled by scripting"
+DESCRIPTION="A powerful nonlinear scripting language for media. Includes a set \
+of core filters for editing, pre-production and post-production, and the \
+ability to extend its capabilities through the use of external plugins."
+HOMEPAGE="https://avs-plus.net"
+COPYRIGHT="2000-2015 Ben Rudiak-Gould, et al., \
+	2013-2021 AviSynth+ Project"
```
Begasus, line 7: No need for the backslash here, you can use multiple lines for COPYRIGHT:
```diff
-COPYRIGHT="2000-2015 Ben Rudiak-Gould, et al., \
	2013-2021 AviSynth+ Project"
+COPYRIGHT="2000-2015 Ben Rudiak-Gould, et al.
	2013-2021 AviSynth+ Project"
```

Begasus (commented):
Some cosmetic changes (see some tips at https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines)
Maybe use libVersion(Compat) here also

qyot27:
Force-pushed again with the latest cosmetics suggestions.

>And ideally you would move your lib to the add-ons/ directory instead of lib/, but that may create some trouble for ported applications.

Despite trying to Google for it, I can't seem to find a description of what add-ons/ should actually be used for.  It's not mentioned at all on [filesystem layout](https://www.haiku-os.org/docs/userguide/en/filesystem-layout.html) or [applications](https://www.haiku-os.org/docs/userguide/en/applications.html) pages, and the [installing applications](https://www.haiku-os.org/guides/daily-tasks/install-applications) page doesn't really explain it beyond saying that some of the app's filters would be going in there to get picked up by the system.  I'm not sure how that could map to what AviSynth is doing (and because it doesn't look like add-ons/ is on $LIBRARY_PATH, clients attempting to dlopen it would likely fail).

I would think that, say, a Media Kit interface/bridge talking to libavisynth would be the thing that would go into add-ons/, not the library itself.  Or perhaps the external plugins (which would also require changing the avisynth_conf.h.in path again), assuming they aren't regular libraries of their own that just happen to include an AviSynth plugin interface.

Begasus:
From my side .. LGTM ;) Thanks for your patience @qyot27 !

pulkomandy:
It's not in LIBRARY_PATH but it's in ADDON_PATH. Normally that would be used with load_add_on (the BeAPI equivalent to dlopen) and things can be searched there using find_directory.

But, yes, for the simplicity of portability, it's probably simpler to keep this with libraries.

media-libs/avisynthplus/avisynthplus-3.7.0.recipe
```diff
@@ -0,0 +1,67 @@
...
+SOURCE_FILENAME="avisynthplus-$portVersion.tar.gz"
+SOURCE_DIR="AviSynthPlus-$portVersion"
+PATCHES="0001-CMakeLists-fix-setting-local-plugindir.patch
+	0002-avisynth_conf.h.in-use-mounted-system-libdir-on-Haik.patch
+	"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64 ?arm ?ppc"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="$portVersion"
```
korli, line 21: the port version isn't helpful here, just use 8

PR #5616, opened 2021-01-19
ytdl-gui: Add recipe

win8linux:
[YouTubeDL-GUI](https://github.com/JaGoLi/ytdl-gui) is a Qt-based, graphical interface for `youtube-dl` that aims to be simple-to-use yet still provide various options such as file formats and quality levels.

image
image


win8linux:
WTF

PR #5617, opened 2021-01-19
youtubedl-GUI: Add recipe

win8linux:
[YouTubeDL-GUI](https://github.com/JaGoLi/ytdl-gui) is a Qt-based, graphical interface for `youtube-dl` that aims to be simple-to-use yet still provide various options such as file formats and quality levels.

image
image

It does have an icon, but perhaps it could be replaced later on for something more appropriate for Haiku:
image


humdingerb:
I created an icon for it: Icon

You can export the rdef from the IOM file: http://0x0.st/-i2o.iom
I suggest to add the .IOM file to the additional-files as well, in case people would like to improve it.

humdingerb:
I know you just did the port, but the GUI would be so much nicer when it either used pop-menus for the options, or at least arrange the radio buttons vertically. Oh well... :)

win8linux:
Those seem like good suggestions for upstream. ;)

humdingerb:
Done: https://github.com/JaGoLi/ytdl-gui/issues/10

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,79 @@
+SUMMARY="A simple-to-use, cross-platform graphical interface for youtube-dl"
+DESCRIPTION="YouTubeDL-GUI is a cross-platform graphical interface for \
+youtube-dl, a downloader for online media sharing services like YouTube,\
+ Vimeo, SoundCloud, Bandcamp, and more. It is based on Qt and aims to be\
```
humdingerb, line 4: To be nice and consistent, the spaces should be moved from the start of the line to the end of the previous line. As it's done already for the above lines.

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,79 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+srcGitTag="1.0"
+SOURCE_URI="https://github.com/JaGoLi/ytdl-gui/archive/$srcGitTag.tar.gz"
+CHECKSUM_SHA256="20d262cd3e5ce77b075e0ffee6b9d0480560856f687e1c63097fc146825166c6"
+SOURCE_FILENAME="ytdl-gui-$portVersion-tar.gz"
+SOURCE_DIR="ytdl-gui-$srcGitTag"
+ADDITIONAL_FILES="ytdl_gui.rdef.in"
+
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
```
humdingerb, line 18: "x86" isn't officially supported anymore. It can be removed.

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,79 @@
...
+
+BUILD()
+{
+	make build
+}
+
+INSTALL()
+{
+	mkdir $appsDir
+	mv ./src/ytdl_gui $appsDir/ytdl_gui
```
humdingerb, line 67: Since the app appers in the Deskbar as "YouTubeDL-GUI" I think the executable should be "YouTubeDL-GUI" as well.
Maybe even just simply "YouTubeDL" ?

humdingerb (commented):
If you put the Icon-O-Matic file in the additional-files folder (without mentioning it in the recipe, ofc), people have the icon with nice meta data (named shapes/styles/paths) to edit.

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,79 @@
...
+	"
+
+BUILD()
+{
+	make build
+}
+
+INSTALL()
+{
+	mkdir $appsDir
```
humdingerb, line 66:
```suggestion
	mkdir -p $appsDir
```

humdingerb:
Does it work for you? I always just get a "Failed! Recheck input for errors." after a second. In Terminal I get: "grep: Perl matching not supported in a --disable-perl-regexp build".

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,79 @@
...
+ARCHITECTURES="!x86_gcc2 x86 x86_64"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	ytdl_gui$secondaryArchSuffix = $portVersion
+	app:YouTubeDL_GUI$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	cmd:youtube_dl$secondaryArchSuffix
```
Begasus, line 27: youtube_dl is primary architecture so:
```diff
- 	cmd:youtube_dl$secondaryArchSuffix
+ 	cmd:youtube_dl
```

Begasus:
youtube-dl works for me in Terminal, I'm getting another report when trying to download with this one (translated from Dutch):
```
grep: support for the option '-P' in not included in this program version
```

Begasus:
Seems to be a grep thing? 
```
-#: src/search.c:606
-msgid "The -P option is not supported"
-msgstr "Ondersteuning voor de optie -P is in deze versie niet opgenomen"
-
 #: src/search.c:619
+msgid "Support for the -P option is not compiled into this --disable-perl-regexp binary"
+msgstr "Ondersteuning voor de optie '-P' is in deze programmaversie niet opgenomen"
```

Begasus:
Removing the "P" from the line solved it for me, download went well, this line needs a fix for this to work:
https://github.com/JaGoLi/ytdl-gui/blob/927f129a4613d27f3dc7f3cc98fd71e3e3199798/src/mainactions.cpp#L122
```diff
-    std::string parse_output = R"(stdbuf -o0 grep -oP '^\[download\].*?\K([0-9]+)' | tee /tmp/ytdl_prg)";
+    std::string parse_output = R"(stdbuf -o0 grep -o '^\[download\].*?\K([0-9]+)' | tee /tmp/ytdl_prg)";
```

Begasus:
Progress bar is not showing the progress for the download

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
win8linux: It would be more preferable to have the executable name be the same as either the package name or the upstream project's shorthand (`ytdl-gui`). I have thought about changing the Deskbar link to "*YouTube Downloader*" to match the window title.  However, that seems to be quite generic as there are other such YouTube downloaders that label themselves as such. If any of the non-web ones get ported to Haiku, then some confusion may ensue.

If others feel that it is ok though, then I would be willing to change the Deskbar link name to match the window title.

Begasus: Tab shows "Youtube Downloader" maybe that's more in line?

humdingerb: On whatever name you settle, I suggest to have it be the same for executable and Deskbar link. E.g.  a launcher like QuickLaunch won't find the app when entering "youtube", though the app appears as "YouTubeDL-GUI" in the Application menu of the Deskbar. When running, it appears in the Deskbar as the executable's name, currently "ytdl_gui".

The window title of the app "Youtube Downloader" can differ from the app name IMO. But the above mess should be avoided. :)

Begasus:
After making some local changes to the recent grep recipe (doesn't build for gcc2) and some alternate option to configure this is working ok now :)
youtubedl-gui

win8linux:
Should I just switch to using Git revisions instead of tags? Based on the conversation in the closed upstream issue, it seems like the better idea.

Begasus:
> Should I just switch to using Git revisions instead of tags? Based on the conversation in the closed upstream issue, it seems like the better idea.

I don't think it would solve the grep problem, need to check it out first, for now I would just stick to this version

win8linux:
Not so much due to the `grep` problem, but rather due to what @JaGoLi said about the rapid pace of development.

Begasus:
From his comment there:
```
There you can also get a preview of the changes so you can test the app before pushing it out to the haiku userbase.
```

win8linux:
On a related note, should I add `ffmpeg` as a dep? Not sure whether it already comes with the base system, which would render it redundant.

Begasus:
So far it's not a deppendency, so lets wait on how it progresses there, I think ffmpeg is only needed for playback

Begasus:
Still some issues left for it, but I got to build and save audio files from a playlist earlier on the beta branch :)
youtube-dl-gui

Begasus:
Still some issues left for it, but I got to build and save audio files from a playlist earlier on the beta branch :)



> On a related note, should I add `ffmpeg` as a dep? Not sure whether it already comes with the base system, which would render it redundant.

At least "cmd:ffmpeg$secondaryArchSuffix" is needed in REQUIRES

Begasus:
And another milestone :)
youtube-dl-gui

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,80 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	ytdl_gui$secondaryArchSuffix = $portVersion
+	app:YouTubeDL_GUI$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	cmd:youtube_dl
+	cmd:ffmpeg$secondaryArchSuffix
```
Begasus, line 28: Still looking into this, maybe we should move this dependency to youtube_dl

Begasus: You can drop this, added it to the dependency for youtube-dl :)
```diff
- 	cmd:ffmpeg$secondaryArchSuffix
```

net-misc/ytdl-gui/ytdl_gui-1.0.recipe
```diff
@@ -0,0 +1,80 @@
...
+	lib:libcrypto$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
```
Begasus, line 31: Add this for supporting mp4 files:
```
+	lib:libmp4v2$secondaryArchSuffix
```

Begasus:
Thanks! Let's merge this one for now, maybe there will be a new release shortly :) (PS, you can switch to cmake build then) :)

PR #5620, opened 2021-01-21
grep, revbump for rebuild with pcre

Begasus:


sys-apps/grep/grep-3.6.recipe
```diff
@@ -5,50 +5,61 @@ matching lines."
...
 	"
 REQUIRES="
-	haiku
-	lib:libiconv
-	lib:libintl
+	haiku$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	lib:libpcre$secondaryArchSuffix
 	"
```
korli, line 34: REPLACES is missing for x86_gcc2

Begasus: Need to see how to do that part right :) disabled x86_gcc2 in the previous versions, shouldn't that take care of it?

korli: No, grep_x86 should replace grep when it's already installed on x86_gcc2

Begasus: I think the below section should cover that now (at least it worked here locally)
Should I revert the changes to the prior version (eg enable x86_gcc2 again)?

korli: See the tar recipe for example:
https://github.com/haikuports/haikuports/commit/4b890dcedfcb01de02a18ec8ffa257595ade56ad#diff-14128caa66153dbe9da8ad6b8d6d46569e369c273edc7075a2a483a30e17a1a8R38

Begasus: Didn't think of that ;)

korli (approved):

PR #5641, opened 2021-01-26
osl: bump to v1.10.13

kenmays:
Bump Haiku build to Open Shading Language v1.10.13.
Ref: #4669


extrowerk:
The pointcloud.cpp patch seems to be unnecessary, at least on 64 bit it builds fine.
Every test fails tho:
```
OSError: [Errno -2147459070] File or Directory already exists
```

Begasus:
> The pointcloud.cpp patch seems to be unnecessary, at least on 64 bit it builds fine.

That was the starting point on trying to enable it for 32-bit @extrowerk

Begasus:
Build failed on 32-bit, please check
https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/1186/builds/2071.log

PR #5644, opened 2021-01-30
mc, move to secondary architecture

Begasus:


Begasus:
Moving to gcc8 as gcc2 bails out with parse errors: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/1100/builds/1938.log

PR #5646, opened 2021-01-30
libftdi, bump version, fix installation paths for the headers

Begasus:


Begasus:
Changes to the installation paths for the headers are needed to build avrdude (this needs a revbump after libusb-compat is merged)

Begasus:
Packages depending on this:
```
- [ ] avrdude (in progress)
- [ ] puseview
- [ ] libsigrok
- [ ] sigrok_cli
```

Begasus:
This works for avrdude, but fails for libsigrok :/

Begasus:
Looking at: https://github.com/haikuports/haikuports/blob/95f8b9103b168146cefb19b189248a94572764cd/dev-embedded/libftdi/patches/libftdi-1.0.patch#L36
This should be the way to go imho

Begasus:
Looking at the generated file vs the installed file for pkg-config it seems that fixPkgconfig removes the ${PROJECT_NAME}  in there?

dev-embedded/libftdi/libftdi-1.5.recipe
```diff
@@ -5,36 +5,40 @@ library is linked with your program in userspace, no kernel driver required.\
...
-CHECKSUM_SHA256="ec36fb49080f834690c24008328a5ef42d3cf584ef4060f3a35aa4681cb31b74"
+CHECKSUM_SHA256="7c7091e9c86196148bd41177b4590dccb1510bfe6cea5bf7407ff194482eb049"
 SOURCE_DIR="libftdi1-$portVersion"
 PATCHES="libftdi-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86_64"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion="2.5.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```
Begasus, line 21: @korli compat string for this is fine for libftdi1 but libftddipp1 provides a so link with compat = 3 (libftdipp1.so.3 -> libftdipp1.so.2.5.0), should there be some change for this?

korli: @Begasus it's OK so. Alternatively one can use "3.0.0 >= 3". It's only important if some other package uses this library..

Begasus: Thanks @korli , I thought it was OK also, but wanted to double check ;) :+1:

PR #5662, opened 2021-02-04
pyenchant: Add recipe

JakobDev:
pyencahnt provides Python bindings to the enchant library.

### DIFF
--- a/dev-python/pyenchant/pyenchant-3.2.0.recipe
+++ b/dev-python/pyenchant/pyenchant-3.2.0.recipe
@@ -8,7 +8,6 @@ LICENSE="GNU LGPL v2.1"
 REVISION="1"
 SOURCE_URI="https://github.com/pyenchant/pyenchant/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="3195bfc5cfbdeb8a60f9484390c2c982e59dd220d6f9a5f8418ba67551ead01a"
-SOURCE_DIR="pyenchant-$portVersion"
 
 ARCHITECTURES="any"
 
@@ -17,7 +16,7 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku
-	enchant
+	lib:libenchant_2$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
--- a/dev-python/pyenchant/pyenchant-3.2.0.recipe
+++ b/dev-python/pyenchant/pyenchant-3.2.0.recipe
@@ -1,0 +1,63 @@
+SUMMARY="Python bindings for the Enchant spellchecker"
+DESCRIPTION="This package provides a set of Python language \
+bindings for the Enchant spellchecking library. For more \
+information, visit the project website."
+HOMEPAGE="http://pyenchant.github.io/pyenchant/"
+COPYRIGHT="2020 Dimitri Merejkowsky"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://github.com/pyenchant/pyenchant/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="3195bfc5cfbdeb8a60f9484390c2c982e59dd220d6f9a5f8418ba67551ead01a"
+SOURCE_DIR="pyenchant-$portVersion"
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	$portName = $portVersion
+	"
+REQUIRES="
+	haiku
+	enchant
+	"
+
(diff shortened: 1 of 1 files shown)


dev-python/pyenchant/pyenchant-3.2.0.recipe
```diff
@@ -0,0 +1,63 @@
...
+DESCRIPTION="This package provides a set of Python language \
+bindings for the Enchant spellchecking library. For more \
+information, visit the project website."
+HOMEPAGE="http://pyenchant.github.io/pyenchant/"
+COPYRIGHT="2020 Dimitri Merejkowsky"
+LICENSE="GNU LGPL v2.1"
+REVISION="1"
+SOURCE_URI="https://github.com/pyenchant/pyenchant/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="3195bfc5cfbdeb8a60f9484390c2c982e59dd220d6f9a5f8418ba67551ead01a"
+SOURCE_DIR="pyenchant-$portVersion"
```
korli, line 11: you can remove SOURCE_DIR when it's already the default

dev-python/pyenchant/pyenchant-3.2.0.recipe
```diff
@@ -0,0 +1,63 @@
...
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	$portName = $portVersion
+	"
+REQUIRES="
+	haiku
+	enchant
```
korli, line 20: this probably depends on the library: lib:libenchant_2$secondaryArchSuffix

JakobDev:
@korli Changes are now done

dev-python/pyenchant/pyenchant-3.2.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+CHECKSUM_SHA256="3195bfc5cfbdeb8a60f9484390c2c982e59dd220d6f9a5f8418ba67551ead01a"
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	$portName = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libenchant_2$secondaryArchSuffix
```
Begasus, line 19: As "lib:libenchant_2$secondaryArchSuffix" isn't "any" architecture, I guess this means this can't be "any" architecture, see failed build: https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=1159&viewMode=expanded

korli: Yes, that makes sense.

JakobDev: I just testet it with the first requirement and I had used the suggestion from korli without testing it again. Sorry for that.

korli: to be honest, I don't expect everything to work out of the box on x86, because python3 is only x86 secondary arch. It means that it should load x86 secondary arch libraries.
@Begasus  Maybe we could check the effectiveTargetArchitecture, and add a suffix in this case as a workaround.

Begasus: Not a big fan for these "mixed" architectures with the python recipes, but an option to check yes

Begasus: I pushed a small change to: https://github.com/haikuports/haikuports/pull/5675
This fixes the build for me, but still seems to fail on the buildbot :(

dev-python/pyenchant/pyenchant-3.2.0.recipe
```diff
@@ -0,0 +1,62 @@
...
+PYTHON_VERSIONS=(3.7 3.8 3.9)
+for i in "${!PYTHON_PACKAGES[@]}"; do
+pythonPackage=${PYTHON_PACKAGES[i]}
+pythonVersion=${PYTHON_VERSIONS[$i]}
+eval "PROVIDES_${pythonPackage}=\"\
+	${portName}_$pythonPackage = $portVersion\
+	\"; \
+REQUIRES_$pythonPackage=\"\
+	haiku\n\
+	cmd:python$pythonVersion\
```
Begasus, line 36: The base package now only has a requirement for libenchant_2, do the subpackages require that too?
If so, maybe move this here as a requirement?

JakobDev: pyenchant is a python wrapper for enchant, so all packages need it.

PR #5664, opened 2021-02-04
certifi: bump

JakobDev:

PR #5665, opened 2021-02-04
requests: Add certifi as requirement

JakobDev:
requests needed certifi. Please note that #5664 added packages for python38 and python39 so make sure to merge them before.


korli:
We patch requests to use system certificates, see https://github.com/haikuports/haikuports/blob/master/dev-python/requests/requests-2.25.1.recipe#L61
It doesn't work for you?

JakobDev:
requests itself works without any problem. But I have installed a program that requires requests with a setup.py. When I try to start this program, I get a pkg_resources.DistributionNotFound Error. If requests is patched to use the system certificates, it might be a good idea the patch the requirement of certifi out of the setup.py of requests.

korli:
Good idea! Modified in 1470583c0ef5d6b0d14dda272c7b3e44baf206dd.

JakobDev:
@korli That fixed it for me. Thanks. I'm going to close this PR now.

PR #5694, opened 2021-02-09
bash_completion: bump

JakobDev:
New Version of bash completions. btw. I noticed that bash:completion not getting sourced at start. You have to do this by hand. Maybe we need to fix the bash recipe (which contains the line that should start bash_completion).  I also think we should try to upstream the completions of pkgman and finddir.


diversys:
Please `git mv` older recipe instead of creating new one.

JakobDev:
@diversys How can I do that when I already opened a PR?

app-shells/bash_completion/patches/bash_completion-2.11.patchset
```diff
@@ -0,0 +1,102 @@
+From c83f6af3e75d67aa7789c5656f872c4b3aaf49cc Mon Sep 17 00:00:00 2001
+From: JakobDev <jakobdev@gmx.de>
```
Begasus, line 2: Would be nice if the credits were kept as they were (quick look doesn't show any real diffs):
https://github.com/haikuports/haikuports/blob/f85fb2d8ddbc31ce149a3395a48ff1cca61c05c9/app-shells/bash_completion/patches/bash_completion-2.7.patchset#L2

diversys:
> @diversys How can I do that when I already opened a PR?

I think you can just force push.

korli:
> @diversys How can I do that when I already opened a PR?

Or just push a commit which removes the old recipe.

PR #5575, opened 2021-01-05
Embree3: bump, cpucount

extrowerk:
Required by Blender 2.9x

PR #5695, opened 2021-02-10
embree: bump to 3.12.2

kenmays:
Ref: #5575


kenmays:
Seems OK to migrate to 3.12.2. Reverted to #5575  for 32-bit work request.

Issue #5504, opened 2020-12-23
Blender: crashes on quit


diversys:
```
Debug information for team /boot/system/apps/Blender/Blender (2853):
CPU(s): 2x Intel Core™ i7-3635QM
Memory: 1.49 GiB total, 718.19 MiB used
Haiku revision: hrev54805 Dec 22 2020 07:49:38 (x86_64)

Active Threads:
	thread 2854: pthread func 
	thread 2855: pthread func 
	thread 2856: SDLTimer 
	thread 2857: SDLApplication 
	thread 2860: _BMediaRoster_ 
	thread 2861: Blender control 
	thread 2862: pthread func 
	thread 2868: pthread func 
	thread 2869: pthread func 
	thread 2935: pthread func 
	thread 2936: team 2853 debug task 
	thread 2853: Blender (main)
		state: Exception (General protection fault)

		Frame		IP			Function Name
		-----------------------------------------------
		0x7f7b226834b0	0x1294efdfa9b	BMessage::HasSpecifiers() const + 0xb 
			Disassembly:
				BMessage::HasSpecifiers() const:
				0x000001294efdfa90:         488b5710  mov 0x10(%rdi), %rdx
				0x000001294efdfa94:             31c0  xor %eax, %eax
				0x000001294efdfa96:           4885d2  test %rdx, %rdx
				0x000001294efdfa99:             7409  jz 0x4efdfaa4
				0x000001294efdfa9b:           8b4208  mov 0x8(%rdx), %eax <--

			Frame memory:
				[0x7f7b226834a8]  B..N)...   42 e2 fd 4e 29 01 00 00
		0x7f7b22683670	0x1294efde23d	BLooper::MessageReceived(BMessage*) + 0x1d 
		0x7f7b22683680	0x461b97c7b5	GHOST_ContextSDL::~GHOST_ContextSDL() + 0x55 
		0x7f7b226836a0	0x461b97c80c	GHOST_ContextSDL::~GHOST_ContextSDL() + 0xc 
		0x7f7b226836c0	0x461b97bb2c	GHOST_WindowSDL::~GHOST_WindowSDL() + 0xc 
		0x7f7b226836f0	0x461b97d2af	GHOST_System::disposeWindow(GHOST_IWindow*) + 0x9f 
		0x7f7b22683720	0x461af52357	wm_window_free + 0x117 
		0x7f7b22683750	0x461af32e49	wm_close_and_free + 0x49 
[...]

Issue #5500, opened 2020-12-22
Blender is broken on 32bit


Begasus:
https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/870/builds/1490.log

Think this is the root cause: undefined reference to `___tls_get_addr'

extrowerk:
export LDFLAGS=“-lssp”
maybe?

Begasus:
Wanted to have a go at it yesterday, but it's a monster to build, only thing left is to push changes and see if the buildbot is ok with it (from my side, and if no one steps up check do a local build) :)

Begasus:
Pushed the change, lets see how it goes ...

Begasus:
Didn't work, so have to revert these last changes on a next push ...

Begasus:
Maybe -DCMAKE_SKIP_RPATH=ON

_Originally posted by @korli in https://github.com/haikuports/haikuports/issues/3061#issuecomment-422427687_

PR #5696, opened 2021-02-10
blender: respin

kenmays:
Ref  #5500
Ref #5504
Builds on x86.

### DIFF
--- a/media-gfx/blender/blender-2.79b.recipe
+++ b/media-gfx/blender/blender-2.79b.recipe
@@ -72,6 +72,7 @@ REQUIRES="
 	lib:libopenvdb$secondaryArchSuffix
 	lib:libosdCPU$secondaryArchSuffix
 	lib:libosdGPU$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


extrowerk:
About the switch to python 3.7: did you tested it? There was compatibility problems afair with newer python versions with 2.79.
Please try to answer as simple as possible.

extrowerk:
Ref: https://github.com/haikuports/haikuports/pull/4465/files

kenmays:
@extrowerk - Didn't run testsuite. Using internal python. Tested only compilation on Python 3.7 - since R1B2 is now installs for python3 versus Python 3.6. I built blender on x86 & x86_64. Crashes on exit on both platforms. Also, some GUI display issues now seen on both platform builds.

kenmays:
Cancel this PR. X512 is moving towards Blender 2.82+

X547:
Blender 2.79 don't work with python 3+. See https://github.com/haikuports/haikuports/issues/4459.

PR #5708, opened 2021-02-13
osl: fixed x86 patching

kenmays:


Begasus:
Still fails to build on the buildbot: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/1190/builds/2075.log

PR #5714, opened 2021-02-15
[tuxpaint]  Update tuxpaint enable svg 

mazbrili:
enable svg , so svg stamp can be loaded


Begasus:
Thanks for the PR @mazbrili could you check with the updated librsvg ?

mazbrili:
> Thanks for the PR @mazbrili could you check with the updated librsvg ?

tested.. work as expected on x86_64

PR #5724, opened 2021-02-16
mac: bump to 6.12

enzo1982:
Here's another bump for the Monkey's Audio package as #5717 failed because the upstream author removed the 6.11 zip.


Begasus:
Would indeed be nice to have at least a few past versions around before removing them, thanks here!

PR #5733, opened 2021-02-21
tuxpaint, revbump for rebuild (trying to fix svg loading)

Begasus:


Begasus:
@mazbrili "should" be fixed, could you check on x86_64?
Still not really shore why it didn't work before, my guess is something with the shared_mime_info

mazbrili:
> @mazbrili "should" be fixed, could you check on x86_64?
> Still not really sure why it didn't work before, my guess is something with the shared_mime_info

@Begasus  confirmed ok on x86_64 thanks!!

Begasus:
Thanks @mazbrili for reporting back! :+1:

PR #5735, opened 2021-02-21
New recipe: MidiSynth v1.8

humdingerb:


haiku-apps/midisynth/additional-files/install_chord_keymapping.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+settingsDir="$(finddir B_USER_SETTINGS_DIRECTORY)/MidiSynth/"
```
korli, line 3: Would it work to use template settings instead?
https://github.com/haikuports/haikuports/blob/15f115b19035027b6f848d886093a89f035c3fe3/net-misc/openssh/openssh-8.4p1.recipe#L40

humdingerb: I'm not that familiar with USER_SETTINGS_FILES and 'templates'. I've updated the recipe.
The Keymappings and Chords are in the package in the data dir, but are not copied to ~/config/settings/midisynth on install. Do I still need the post-install script? If so, what's gained by using those template-ma-thingie?

korli: Yes, it's not automatic on activation, see https://git.haiku-os.org/haiku/plain/docs/develop/packages/BuildingPackages.rst
The usecase is for the application to copy the template. Alternatively this can be done with a post install script.

humdingerb: Alright. I reinstated the post install script for now. Merge?

PR #5739, opened 2021-02-25
FFMPEG: enable dav1d

extrowerk:


extrowerk:
Seems to be working ok on 64 bit, MediaPlayer plays AV1 videos without a hitch with this change.

extrowerk:
Youtube started to use AV1 for their videos, which was unable to play back even if we download them with ytdl, until now.

PR #5741, opened 2021-02-25
FFMPEG: enable some 3rdparty libs

extrowerk:
None requires the --enable-gpl switch, seems to be working ok on 64 bit.

### DIFF
--- a/media-video/ffmpeg/ffmpeg-4.2.2.recipe
+++ b/media-video/ffmpeg/ffmpeg-4.2.2.recipe
@@ -73,6 +73,7 @@ if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
 		lib:libfontconfig$secondaryArchSuffix
 		lib:libfreetype$secondaryArchSuffix
 		lib:libfribidi$secondaryArchSuffix
+#		lib:libgdk_pixbuf_2.0$secondaryArchSuffix
 		lib:libgme$secondaryArchSuffix
 		lib:libgmp$secondaryArchSuffix
 #		lib:libglib_2.0$secondaryArchSuffix
@@ -116,9 +117,9 @@ REQUIRES_avdevice="
 
 PROVIDES_tools="
 	ffmpeg${secondaryArchSuffix}_tools = $portVersion
-	cmd:ffmpeg = $portVersionCompat
-	cmd:ffplay = $portVersionCompat
-	cmd:ffprobe = $portVersionCompat
+	cmd:ffmpeg$secondaryArchSuffix = $portVersionCompat
+	cmd:ffplay$secondaryArchSuffix = $portVersionCompat
+	cmd:ffprobe$secondaryArchSuffix = $portVersionCompat
 	"
 REQUIRES_tools="
 	haiku$secondaryArchSuffix
@@ -189,11 +190,13 @@ if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
 #		devel:libbluray$secondaryArchSuffix
 #		devel:libbs2b$secondaryArchSuffix
 #		devel:libcaca$secondaryArchSuffix
+#		devel:libcairo$secondaryArchSuffix
 		devel:libdav1d$secondaryArchSuffix
 		devel:libfdk_aac$secondaryArchSuffix
 		devel:libfontconfig$secondaryArchSuffix
 		devel:libfreetype$secondaryArchSuffix
 		devel:libfribidi$secondaryArchSuffix
+#		devel:libgdk_pixbuf_2.0$secondaryArchSuffix
 		devel:libgme$secondaryArchSuffix
 		devel:libgmp$secondaryArchSuffix
 		devel:libopenjp2$secondaryArchSuffix
--- a/media-video/ffmpeg/ffmpeg-4.2.2.recipe
+++ b/media-video/ffmpeg/ffmpeg-4.2.2.recipe
@@ -6,7 +6,7 @@ HOMEPAGE="https://ffmpeg.org/"
 COPYRIGHT="2000-2003 Fabrice Bellard
 	2003-2019 the FFmpeg developers"
 LICENSE="GNU LGPL v3"
-REVISION="8"
+REVISION="9"
 SOURCE_URI="https://ffmpeg.org/releases/ffmpeg-$portVersion.tar.xz"
 CHECKSUM_SHA256="cb754255ab0ee2ea5f66f8850e1bd6ad5cac1cd855d0a2f4990fb8c668b0d29c"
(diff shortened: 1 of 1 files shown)


pulkomandy:
Please test with x86_gcc2. Probably we need to keep a special version of the ffmpeg recipe for gcc2 because it will not be possible to build all dependencies there.

extrowerk:
The new deps enabled only for x86 and for x86_64.
Maybe @Begasus can make a testrun on 32 bit.

Adrien Destugues <notifications@github.com>:
> Please test with x86_gcc2. Probably we need to keep a special version of the ffmpeg recipe for gcc2 
> because it will not be possible to build all dependencies > there.

Begasus:
Give me some time until the weekend, I could run a build on x86_gcc2 there then

media-video/ffmpeg/ffmpeg-4.2.2.recipe
```diff
@@ -158,9 +185,33 @@ BUILD_REQUIRES="
...
 		devel:libdav1d$secondaryArchSuffix
+		devel:libfdk_aac$secondaryArchSuffix
+		devel:libfontconfig$secondaryArchSuffix
+		devel:libfreetype$secondaryArchSuffix
+		devel:libfribidi$secondaryArchSuffix
+		devel:libgme$secondaryArchSuffix
+		devel:libgmp$secondaryArchSuffix
+		devel:libopenjp2$secondaryArchSuffix
 		devel:libopenmpt$secondaryArchSuffix
+#		devel:librsvg_2$secondaryArchSuffix
```
Begasus, line 201: Since I did the update for librsvg, did you check why the build needs/fails on it?

extrowerk: I just don't see the point enabling librsvg for ffmpeg. Is it really a use case to open svg files with MediaPlayer? Or to save a bitmap image (because factually ffmpeg gives a non-vctor output) to svg with ffmpeg? Or is there any real-world use case for that?
We could enable tessaract or tensorflow, or many other things (ok, we don't have tensorflow port yet), but we need to check f that makes any sense.
Other argumentation coulkd be: it is possible to do something, so it must be allowed to do that. I can accept that kind of argumentation, but i still don't see the point.

Begasus: Didn't check on how ffmpeg uses it, just asking :)

Begasus:
For gcc2 the build is ok, for gcc8 the commands in PROVIDES_tools need "$secondaryArchSuffix" added

media-video/ffmpeg/ffmpeg-4.2.2.recipe
```diff
@@ -63,9 +63,36 @@ REQUIRES="
...
+		lib:libass$secondaryArchSuffix
+#		lib:libbluray$secondaryArchSuffix
+#		lib:libbs2b$secondaryArchSuffix
+#		lib:libcaca$secondaryArchSuffix
+#		lib:libcairo$secondaryArchSuffix
 		lib:libdav1d$secondaryArchSuffix
+		lib:libfdk_aac$secondaryArchSuffix
+		lib:libfontconfig$secondaryArchSuffix
+		lib:libfreetype$secondaryArchSuffix
+		lib:libfribidi$secondaryArchSuffix
```
Begasus, line 75:
```diff
+#		lib:libgdk_pixbuf_2.0$secondaryArchSuffix
```

media-video/ffmpeg/ffmpeg-4.2.2.recipe
```diff
@@ -158,9 +185,33 @@ BUILD_REQUIRES="
...
 	BUILD_REQUIRES+="
+		devel:libass$secondaryArchSuffix
+#		devel:libbluray$secondaryArchSuffix
+#		devel:libbs2b$secondaryArchSuffix
+#		devel:libcaca$secondaryArchSuffix
 		devel:libdav1d$secondaryArchSuffix
+		devel:libfdk_aac$secondaryArchSuffix
+		devel:libfontconfig$secondaryArchSuffix
+		devel:libfreetype$secondaryArchSuffix
+		devel:libfribidi$secondaryArchSuffix
```
Begasus, line 196:
```diff
+#		devel:libgdk_pixbuf_2.0$secondaryArchSuffix
```

media-video/ffmpeg/ffmpeg-4.2.2.recipe
```diff
@@ -158,9 +185,33 @@ BUILD_REQUIRES="
 	"
 if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
 	BUILD_REQUIRES+="
+		devel:libass$secondaryArchSuffix
```
Begasus, line 188:
```diff
+#		lib:libcairo$secondaryArchSuffix
```

Begasus (commented):
These additions would be needed to use "--enable-librsvg" (just a suggestion) :)

Begasus:
> For gcc2 the build is ok, for gcc8 the commands in PROVIDES_tools need "$secondaryArchSuffix" added

@extrowerk did you have a chance to look at this? No need to look for adding librsvg (although it seems to be working ok), fixing that first one would be enough to merge on my side ;)

media-video/ffmpeg/ffmpeg-4.2.2.recipe
Begasus: FWIW, you can use the commandline "ffmpeg" to convert images from svg to for instance png, you can join them as a group into a video file that can be played with MP (last check I did however was so fast you couldn't see the starting images, but there's an option for delay IIRC) :)

extrowerk:
I did built ffmpeg with all the currently disabled deps enabled. While the build is ok, MediaPlayer crashes silently with them, i had no time to track the badly behaving dep down, so i disabled all the deps with questionable use.

Begasus:
That's ok, I'm fine with that, but for 32bit Haiku the commands in tools package would still need $secondaryArchSuffix if this will be build for gcc8, other then that it's fine :+1:

media-video/ffmpeg/ffmpeg-4.2.2.recipe
```diff
@@ -63,9 +63,37 @@ REQUIRES="
 	"
 if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
 	REQUIRES+="
+		lib:libass$secondaryArchSuffix
+#		lib:libbluray$secondaryArchSuffix
```
Begasus, line 67: FWIW @extrowerk I think I had problems with ffmpeg crashes on libbluray also (didn't check on this PR for this though)

PR #5751, opened 2021-02-28
blender: enforce boost 1.69 depend only build

kenmays:
Enforces usage of boost 1.69 on buildbots - versus boost 1.70 which breaks build.
Recipe needed to block the installation of boost 1.70 devel package.

[blender-2.79b_buildlog.txt](https://github.com/haikuports/haikuports/files/6062673/blender-2.79b_buildlog.txt)

### DIFF
--- a/media-gfx/blender/blender-2.79b.recipe
+++ b/media-gfx/blender/blender-2.79b.recipe
@@ -89,6 +89,7 @@ REQUIRES="
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	boost169${secondaryArchSuffix}_devel
 	devel:eigen$secondaryArchSuffix
 	devel:libalembic$secondaryArchSuffix
 	devel:libavcodec$secondaryArchSuffix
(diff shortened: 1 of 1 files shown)


extrowerk:
Keeping the other boost libs in the BUILD_DEPS list will result in a situation where HP mounts happily both the old and the new boost packages into the chroot (if they have no CONFLICTS).
You should pick one and keep only that.

korli:
This is not how one enforces a version. Please use version requirement.

korli (changes requested):
Change.

Begasus:
FWIW, I fired up a build yesterday and it finished (sometime today while I was out for work) fine, starts up ok also, so it looks ok to me, @extrowerk and @korli you guys are the experts here, if it's ok this could be merged?

korli:
@kenmays   it should be ">= 1.69" (1.69.1 is compatible with 1.69.0)

kenmays:
@korli - The boost 1.70 devel package gets pulled in during compile on the buildbot from the '>=1.69' versus '==1.69'.  Needs a tighter constraint.

```
        /boot/home/haikuports/packages/boost169_x86-1.69.0-2-x86_gcc2.hpkg
       /boot/home/haikuports/packages/boost170_x86-1.70.0-3-x86_gcc2.hpkg
       /boot/home/haikuports/packages/boost170_x86_devel-1.70.0-3-x86_gcc2.hpkg
       /boot/home/haikuports/packages/boost_x86_program_options-1.57.0-3-x86_gcc2.hpkg
```

PR #5763, opened 2021-03-06
nginx: version and feature bump

ihatemyisp:


dev-lang/php/patches/php-7.4.16.patchset
```diff
@@ -0,0 +1,311 @@
+From be47a7ecf2972547120a3f55ea3c177644b5921a Mon Sep 17 00:00:00 2001
```
extrowerk, line 1: Why is this file here? It is php

dev-lang/php/php-7.4.16.recipe
```diff
@@ -0,0 +1,141 @@
+SUMMARY="A popular general-purpose scripting language"
```
extrowerk, line 1: This is php

extrowerk (changes requested):
Clean up the unrelated files

ihatemyisp:
Sorry about that, I really shouldn't work on things late a night.

extrowerk:
Looks better. Will do a testbuild this weekend.

extrowerk:
Could you please remove the old recipe and patches?

extrowerk:
Builds fine on 64 bit.
```
grabbing nginx-1.19.7-2-x86_64.hpkg and moving it to /Ports/haikuports/packages/nginx-1.19.7-2-x86_64.hpkg
```

ihatemyisp:
I'll wipe the old patch and recipe in the morning.

ihatemyisp:
Old receipt and patches removed.

dev-lang/php/php-7.4.16.recipe
Begasus: Still not resolved?

www-servers/nginx/patches/nginx-1.19.7.patchset
```diff
@@ -1,7 +1,7 @@
-From 0d0bb4ae9552f7a9a2c87a27e910b3ad1a4a92bc Mon Sep 17 00:00:00 2001
-From: Sergei Reznikov <diver@gelios.net>
-Date: Wed, 24 Jul 2019 01:43:25 +0300
-Subject: Haiku fixes
+From 9bb9a4bd5106c08fece6ef89052be877c31cdbb8 Mon Sep 17 00:00:00 2001
+From: Ian Corbitt <ian@ihatemyisp.net>
+Date: Fri, 5 Mar 2021 21:59:00 +0000
+Subject: Haiku fixes based off Sergei Reznikov <diver@gelios.net>
```
Begasus, line 4: Would be nice if you could keep the credits from prior committer

ihatemyisp: Ah, so the from can be manually changed. I've changed it.

dev-lang/php/php-7.4.16.recipe
ihatemyisp: Fixed again, sorry.

www-servers/nginx/nginx-1.19.7.recipe
```diff
@@ -11,10 +11,11 @@ COPYRIGHT="2002-2019 Igor Sysoev
 LICENSE="BSD (2-clause)"
 REVISION="2"
 SOURCE_URI="https://nginx.org/download/nginx-$portVersion.tar.gz"
-CHECKSUM_SHA256="5e333687464e1d6dfb86fc22d653b99a6798dda40093b33186eeeec5a97e69ec"
+CHECKSUM_SHA256="7ae4dd020c41d3a5e1e6a8578fcc60e508e3e27e7668e845ddc87a05a775b50e"
 PATCHES="nginx-$portVersion.patchset"
+ADDITIONAL_FILES="userlaunch.in"
 
-ARCHITECTURES="?x86_gcc2 ?x86 ?x86_64"
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
```
Begasus, line 18: This can be build for secondary architecture (32bit) with:
```diff
+SECONDARY_ARCHITECTURES="x86"
```
Not sure how to use this, but at least I had to create a user and group "nobody" to have it started from Terminal, also had to manually create the directory /system/var/nginx/logs ...

ihatemyisp: I'll add it x86. I haven't had a lot of luck getting a 32bit VM to spin up.

The default install requires some configuration before you can launch it. The default configuration has the parent process run as "user" and launch child processes as "nobody". As the default Haiku install doesn't have "nobody", it has to be created or the configuration changed. I'll figure something up to make this more seamless.

For the logs folder, it appears that I need to do like the lighttpd port and have a post-install script create the folders (several in $localStateDir). I'll work on this some this weekend.

Begasus: For the post-install script only "/system/var/nginx/logs" is needed, the other ones are created by nginx
For the configuration, had played a bit with it (like "usr nobody nogroup" and some things I found on the net) only thing that worked for as far as I checked was groupadd and useradd, thanks for looking into this!

ihatemyisp: I can't figure out why the folder isn't already there. It's in the built package file (when I expand it).

I think I'm missing something in how packages work, but I've only been poking around with Haiku for maybe 2 weeks now so I've got a bit of learning left to do.

Begasus:
Another note ;) I see nginx-1.19.8 is released, maybe bump to the latest one in this PR while you are at it :)
Added the post-install script here locally and adjusted the recipe to use it, that part is working already (should be an easy fix)

Begasus:
Changing the path to the webpages shows me it's working :)
nginx

www-servers/nginx/nginx-1.19.7.recipe
```diff
@@ -11,10 +11,11 @@ COPYRIGHT="2002-2019 Igor Sysoev
 LICENSE="BSD (2-clause)"
 REVISION="2"
```
Begasus, line 12: This should be:
```diff
- REVISION="2"
+ REVISION="1"
```

Begasus:
One last (hope so) ;) point, I see the html pages are installed in $docDir now, the directory is read-only so it would be hard to make some changes if you would host a website/page that you need to edit.
IMHO it would be better if this would be installed in $localStateDir/www/nginx (keep the subdirectory so it doesn't collide with other webservers that would use $localeStateDir/www
A sed command for changing the paths in the config file could used also (They also did this in ARCH: https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/nginx-mainline#n109)

extrowerk:
Packages can create users and groups, but idk if it really works.

ihatemyisp:
> Packages can create users and groups, but idk if it really works.

I'll give it a shot with the post install script and see if that works. I've gotta figure out how to test if a user already exists. I think a good option would be to follow what other OSs do and create/use the user "www-data" or something similar.

www-servers/nginx/nginx-1.19.8.recipe
```diff
@@ -0,0 +1,120 @@
...
+	2011-2019 Nginx, Inc."
+# "it's complicated", cf. http://packages.gentoo.org/package/www-servers/nginx
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://nginx.org/download/nginx-$portVersion.tar.gz"
+CHECKSUM_SHA256="308919b1a1359315a8066578472f998f14cb32af8de605a3743acca834348b05"
+PATCHES="nginx-$portVersion.patchset"
+ADDITIONAL_FILES="
+	userlaunch.in
+	setup-nginx.sh"
```
Begasus, line 18: Move double-qoute to next line:
```diff
- 	setup-nginx.sh"
+ 	setup-nginx.sh
+ 	"
```

www-servers/nginx/nginx-1.19.8.recipe
```diff
@@ -0,0 +1,120 @@
...
+	cmd:make
+	cmd:ranlib
+	cmd:sed
+	"
+
+BUILD()
+{
+	./configure --prefix=$prefix \
+		--sbin-path=$binDir/nginx \
+		--doc-prefix=$docDir \
```
Begasus, line 64: With this the default html pages are still installed in $docdir (index.html and 50.html)
Maybe better move them to $localStateDir here also:
```diff
- 		--doc-prefix=$docDir \
+ 		--doc-prefix=$localStateDir/www/nginx \
```
**EDIT** like this you won't have to create the directory in INSTALL but you should add "var/www/nginx" as directory keep old in GLOBAL_WRITABLE_FILES

www-servers/nginx/nginx-1.19.8.recipe
```diff
@@ -0,0 +1,120 @@
...
+
+	mkdir -p \
+		$localStateDir/log/nginx \
+		$localStateDir/tmp/nginx \
+		$localStateDir/www/nginx \
+		$localStateDir/nginx \
+		$localStateDir/run \
+		$dataDir/user_launch \
+		$binDir \
+		$docDir
```
Begasus, line 98: If you used the above suggestion there is no need for $docDir
```diff
- 		$docDir
```

www-servers/nginx/patches/nginx-1.19.8.patchset
```diff
@@ -33,34 +33,34 @@ index d884487..fb487d7 100644
...
-index 521c976..b695ba6 100644
+index 80be906..8aac0bc 100644
 --- a/auto/options
 +++ b/auto/options
 @@ -13,6 +13,7 @@ NGX_CONF_PATH=
  NGX_ERROR_LOG_PATH=
  NGX_PID_PATH=
  NGX_LOCK_PATH=
-+NGX_DOC_PREFIX=
++NGX_DOC_PATH=
```
Begasus, line 43: This doesn't reflect the $value used in --doc-prefix?

ihatemyisp:
I think I've got it working.

- User/group nginx is added if they don't exist
- Build bumped to 1.19.8, 
- root = /var/www/nginx
- logs in /var/logs/nginx/
- temporary files in /var/tmp/nginx
- config in /system/settings/nginx
- pid/lock in /var/run/nginx

www-servers/nginx/patches/nginx-1.19.8.patchset
```diff
@@ -33,34 +33,34 @@ index d884487..fb487d7 100644
...
  NGX_USER=
  NGX_GROUP=
  NGX_BUILD=
-@@ -187,6 +188,7 @@ do
+@@ -190,6 +191,7 @@ do
          --sbin-path=*)                   NGX_SBIN_PATH="$value"     ;;
          --modules-path=*)                NGX_MODULES_PATH="$value"  ;;
          --conf-path=*)                   NGX_CONF_PATH="$value"     ;;
-+        --doc-prefix=*)                  NGX_DOC_PREFIX="$value"    ;;
++	--doc-prefix=*)                  NGX_DOC_PREFIX="$value"    ;;
```
Begasus, line 51: Ok, not sure what this should be doing? Be sure to keep the lines aligned at the same tabs, also NGX_DOC_PREFIX="$value"  is not the same as previous (changed) NGX_DOC_PATH=

www-servers/nginx/nginx-1.19.8.recipe
ihatemyisp: I did it this way so that the files that come with nginx source are always up-to-date. It's unlikely that anyone installing nginx is going to leave the default pages and will likely delete them. Removing the symlinks ensures the source files are still available.

Begasus: Didn't though about that, good point :)

www-servers/nginx/patches/nginx-1.19.8.patchset
ihatemyisp: Oh wow, not even sure how that happened. I'll fix in a few.

www-servers/nginx/patches/nginx-1.19.8.patchset
ihatemyisp: I used tab, the source uses spaces. Will fix shortly.

www-servers/nginx/additional-files/setup-nginx.sh
```diff
@@ -0,0 +1,33 @@
...
+	groupadd nginx
+	notify --group nginx \
+		--icon /system/kernel_`uname -p` \
+		"Created group: nginx"
+fi
+
+mkdir -p var/www/nginx
+
+if [ ! -e var/www/nginx/index.nginx.html ]; then
+	ln -s /system/documentation/packages/nginx/html/index.html var/www/nginx/index.nginx.html
```
Begasus, line 21: On 32bit this path creates broken links as they are installed in secondary architecture:

``` 	ln -s /system/documentation/packages/nginx_x86/html/index.html var/www/nginx/index.nginx.html```

Comment (ihatemyisp) [reply www-servers/nginx/additional-files/setup-nginx.sh:21]: Thanks. I'll build in a check for 32bit.

Comment (Begasus) [reply www-servers/nginx/additional-files/setup-nginx.sh:21]: A possible solution could be to use a variable like they do for the rdef files (in the recipe)
```
local DOCDIR="$docDir"
sed -i "s|@DOCDIR@|$DOCDIR|g" $portDir/additional-files/setup-nginx.sh
```

ihatemyisp: I went with the sed/DOCDIR approach. I'm still not able to get a 32bit VM running but hopefully this works.

ihatemyisp:
I changed all the doc roots to the always point to the system var folder vice the $localStateDir to prevent issues package upgrades that still use default config files.

www-servers/nginx/additional-files/setup-nginx.sh
Begasus: > I went with the sed/DOCDIR approach

Did the same here, need to recheck but so far it's looking ok to me :)

Begasus:
Thanks! :+1:

PR #5767, opened 2021-03-06
Re-enable compiler hardening and zlib check.

daztucker:
The configure tests have been fixed to work upstream.
Switch URL to canonical CDN URL which should be faster in most cases.

### DIFF
--- a/net-misc/openssh/openssh-8.4p1.recipe
+++ b/net-misc/openssh/openssh-8.4p1.recipe
@@ -16,8 +16,8 @@ ssh-keyscan, ssh-keygen and sftp-server."
 HOMEPAGE="http://www.openssh.com/"
 COPYRIGHT="2005-2020 Tatu Ylonen et al."
 LICENSE="OpenSSH"
-REVISION="1"
-SOURCE_URI="https://ftp.fr.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-$portVersion.tar.gz"
+REVISION="2"
+SOURCE_URI="https://cdn.openbsd.org/pub/OpenBSD/OpenSSH/portable/openssh-$portVersion.tar.gz"
 CHECKSUM_SHA256="5a01d22e407eb1c05ba8a8f7c654d388a13e9f226e4ed33bd38748dafa1d2b24"
 PATCHES="openssh-$portVersion.patchset
 	0001-Fix-EOF-command-not-found-error-in-ssh-copy-id.patch
@@ -126,9 +126,7 @@ BUILD()
 		--with-default-path="$defaultPath" \
 		--with-md5-passwords \
 		--disable-utmpx \
-		--with-libedit \
-		--without-zlib-version-check \
-		--without-stackprotect
+		--with-libedit
 	make $jobArgs
 }
 


Begasus:
Builds ok on 32bit Haiku R1B2, test fails:
```
BUILDDIR=`pwd`; \
cd ./regress || exit $?; \
EGREP='/bin/grep -E' \
make \
        .OBJDIR="${BUILDDIR}/regress" \
        .CURDIR="`pwd`" \
        BUILDDIR="${BUILDDIR}" \
        OBJ="${BUILDDIR}/regress/" \
        PATH="${BUILDDIR}:${PATH}" \
        TEST_ENV=MALLOC_OPTIONS="" \
        TEST_MALLOC_OPTIONS="" \
        TEST_SSH_SCP="${BUILDDIR}/scp" \
        TEST_SSH_SSH="${BUILDDIR}/ssh" \
        TEST_SSH_SSHD="${BUILDDIR}/sshd" \
        TEST_SSH_SSHAGENT="${BUILDDIR}/ssh-agent" \
        TEST_SSH_SSHADD="${BUILDDIR}/ssh-add" \
        TEST_SSH_SSHKEYGEN="${BUILDDIR}/ssh-keygen" \
        TEST_SSH_SSHPKCS11HELPER="${BUILDDIR}/ssh-pkcs11-helper" \
        TEST_SSH_SSHKEYSCAN="${BUILDDIR}/ssh-keyscan" \
        TEST_SSH_SFTP="${BUILDDIR}/sftp" \
        TEST_SSH_PKCS11_HELPER="${BUILDDIR}/ssh-pkcs11-helper" \
        TEST_SSH_SK_HELPER="${BUILDDIR}/ssh-sk-helper" \
        TEST_SSH_SFTPSERVER="${BUILDDIR}/sftp-server" \
        TEST_SSH_PLINK="plink" \
        TEST_SSH_PUTTYGEN="puttygen" \
        TEST_SSH_CONCH="conch" \
        TEST_SSH_IPV6="yes" \
        TEST_SSH_UTF8="yes" \
        TEST_SSH_ECC="yes" \
        TEST_SHELL="sh" \
        EXEEXT="" \
        t-exec && echo all t-exec passed
make[1]: Entering directory '/sources/openssh-8.4p1/regress'
run test connect.sh ...
Missing privilege separation directory: /packages/openssh-8.4p1-2/.self/data/openssh/empty
FATAL: sshd_proxy broken
make[1]: *** [Makefile:214: t-exec] Error 1
make[1]: Leaving directory '/sources/openssh-8.4p1/regress'
make: *** [Makefile:693: t-exec] Error 2
```

daztucker:
> Missing privilege separation directory: /packages/openssh-8.4p1-2/.self/data/openssh/

The required privsep chroot directory does not exist.  I suspect this is because the package has not been installed yet.

I think you'll need to either install the package or, if you want to run the tests without installing, change --with-privsep-path in the recipe to point to an existing empty, root-owned, non-writable directory somewhere else.

daztucker:
Got a chance to look into this some more:
 - the test seems to fail the same way at HEAD and on x86_86 too.
 - if you run the test command in the shell, it'll work (at least if the package is installed so the directory exists)
 - it seems to be something about the setup of the chroot the test runs in.  I hacked the test script inside openssh (regress/test-exec.sh) and got it to ls the privsep path, and it showed that the privsep directory is indeed missing from the chroot:
 
run test connect.sh ...
+ ls -ld /packages/openssh-8.4p1-2/.self/data/openssh/empty
ls: cannot access '/packages/openssh-8.4p1-2/.self/data/openssh/empty': No such file or directory
+ ls -ld /packages
lrwxrwxrwx 1 user root 26 Mar  7 11:39 /packages -> /boot/system/package-links
+ ls -l /boot/system/package-links/openssh-8.4p1-2
ls: cannot access '/boot/system/package-links/openssh-8.4p1-2': No such file or directory

Begasus:
Don't think it would be a show stopper, but as this is a fairly important package I'll leave this up to the experts :)

PR #5772, opened 2021-03-08
lynx: update patch for Haiku > r1b2

ihatemyisp:
A Haiku [commit](https://review.haiku-os.org/plugins/gitiles/haiku/+/e04970ad6fb948e03aeb803b00f535a2e712a526) hrev54720 and forward broke Lynx and this fixes things.


Begasus:
Maybe revbump the recipe?

ihatemyisp:
Are you referring to to the file name or the rev number in the recipe?

I didn't think to change anything other than the patch.

Begasus:
"REVISION" should be bumped for this to be rebuild

ihatemyisp:
Gotcha, will update shortly.

ihatemyisp:
Revision bumped.

korli:
this patch isn't interesting for us, because:
1. the buildbot is on r1b2
2. this is a workaround to fix the build, doesn't explain the issue, doesn't fix the issue
3. it should be fixed on the Haiku side to build without ifdef(HAIKU)

ihatemyisp:
Ok...

ihatemyisp:
I'm still looking into fixing this in Haiku itself but that may take me a while.

I doubt there are many users going after this port specifically, but the resolv.h issue might affect more than just this one (and php 7).

Begasus:
@ihatemyisp if you created a bug report upstream could you paste a link to this and the PR for php?
As on R1B2 it seems to be OK, it would be nice to get this out of the way before R1B3 would hit ground :)

ihatemyisp:
> @ihatemyisp if you created a bug report upstream could you paste a link to this and the PR for php?
> 
> As on R1B2 it seems to be OK, it would be nice to get this out of the way before R1B3 would hit ground :)

I'll get it submitted later tonight.

ihatemyisp:
@korli @Begasus I misspoke on PHP7 being related to this workaround. I have opened an issue for this upstream.

https://dev.haiku-os.org/ticket/16851

korli:
thanks, it will be handled at the Haiku bugtracker.

PR #5776, opened 2021-03-11
libsndfile: bump to 1.0.31

enzo1982:
libsndfile and libsamplerate are now maintained by the libsndfile project. Unfortunately, this went largely unnoticed as the previous websites have not been updated to mention this.

This bumps the HaikuPorts recipe to the current libsndfile 1.0.31 release which also fixes some of the quirks dealt with in the old recipe:

- the pkgconfig file has been fixed upstream, so it's no longer necessary to edit it in the INSTALL step
- the new release no longer ships HTML docs, so the patch to adjust the HTML doc path is no longer needed


korli:
Same question as for libsamplerate (cmake and TEST()). Otherwise we can merge as is.

enzo1982:
Alright, I'll give cmake a try and see if I can add the TEST section.

enzo1982:
Pushed the commit to switch to cmake.

libsndfile tests are only supported with static libraries, though. So as we are building shared libs and cmake will not build both, static and shared, we cannot make use of the tests.

media-libs/libsndfile/libsndfile-1.0.31.recipe
```diff
@@ -41,63 +41,51 @@ PROVIDES_devel="
...
 {
-	libtoolize --force --copy --install
-	aclocal -I M4
-	autoconf
-	automake
 	if [ $effectiveTargetArchitecture = x86_gcc2 ]; then
 		export CFLAGS=-O1
 	fi
-	runConfigure ./configure
+	cmake -DCMAKE_INSTALL_PREFIX=$prefix \
```
korli, line 71: $cmakeDirArgs could be used I think

media-libs/libsndfile/libsndfile-1.0.31.recipe
```diff
@@ -41,63 +41,51 @@ PROVIDES_devel="
...
-	cmd:autoconf
-	cmd:automake
-	cmd:find
+	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
-	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
+	cmd:python
```
korli, line 63: should be python2 or python3

enzo1982:
> $cmakeDirArgs could be used I think
> should be python2 or python3

Both done.

PR #5777, opened 2021-03-11
libsamplerate: bump to 0.2.1

enzo1982:
libsndfile and libsamplerate are now maintained by the libsndfile project. Unfortunately, this went largely unnoticed as the previous websites have not been updated to mention this.

This bumps the HaikuPorts recipe to the current libsamplerate 0.2.1 release. The new release no longer ships HTML docs, so the patch to adjust the HTML doc path is no longer needed.


korli:
Would you mind switching to cmake? Maybe also a TEST section with "make test".
Otherwise we can also merge as is.

enzo1982:
> Would you mind switching to cmake? Maybe also a TEST section with "make test".

Alright, I'll give cmake a try and see if I can add the TEST section.

enzo1982:
Pushed the commit to switch to cmake and add the TEST section.

media-libs/libsamplerate/libsamplerate-0.2.1.recipe
```diff
@@ -41,29 +40,29 @@ BUILD_REQUIRES="
...
-	cmd:libtoolize$secondaryArchSuffix
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
 BUILD()
 {
-	autoreconf -fi
-	runConfigure ./configure
+	cmake -DCMAKE_INSTALL_PREFIX=$prefix \
```
korli, line 52: Please try to use $cmakeDirArgs

enzo1982:
> Please try to use $cmakeDirArgs

Done.

PR #5809, opened 2021-04-02
package for python fasteners

phw:
Package for the Python fasteners module, see https://pypi.org/project/fasteners/

Will be needed for an update of MusicBrainz Picard.


dev-python/fasteners/fasteners-0.16.recipe
```diff
@@ -0,0 +1,58 @@
...
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	"
+
+PYTHON_PACKAGES=(python python3 python38 python39)
+PYTHON_VERSIONS=(2.7 3.7 3.8 3.9)
```
korli, line 24: 2.7 isn't supported by this version.

phw: True, removed it

PR #5812, opened 2021-04-03
mtxclient, rename libmatrix_client to mtxclient, fix build for x86_gcc2

Begasus:


Begasus:
Following the Gentoo layout and repology reports: https://repology.org/project/mtxclient/versions

PR #5815, opened 2021-04-04
afl++, rename, fix build

Begasus:


Begasus:
Enabling LLVM gives a fatal error "list not found" -> #include <list>

Begasus:
Patch has been merged upstream, think we are good to go :)

Begasus:
Both build failed, need to investigate :(

PR #5818, opened 2021-04-05
Try to fix the build on the buildbot

Begasus:


Begasus:
Not sure what happens at the buildbot, local builds are ok (R1B2 32bit) here ...

PR #5824, opened 2021-04-09
afl++, revbump, trying to fix build on the buildbot for both arch's

Begasus:


Begasus:
A bit lost here atm, can't find why it builds ok on local (R1B2 hrev54154+123) and fails to build on 32bit at the buildbot, 64bit build seems to be ok now.
Log for failed build on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/1361/builds/2380.log

PR #5834, opened 2021-04-14
[Luckybackup] enable extended attribute

mazbrili:
Luckybackup enable extended attribute backup


Begasus:
You need to add "cmd:rsync" in REQUIRES also (didn't had it installed and I got an error for that one when trying to sync) :)

mazbrili:
> You need to add "cmd:rsync" in REQUIRES also (didn't had it installed and I got an error for that one when trying to sync) :)

yes.. done. thanks @begasus

Issue #4005, opened 2019-07-17
iasl: acpidump does not work


probonopd:
`acpidump` in the package `iasl` does not work:

```
~/Desktop> acpidump  -o DSDT.dat
Cannot open directory - /sys/firmware/acpi/tables
Could not get ACPI tables, AE_NOT_FOUND
```

Apparently `/sys/firmware/acpi/tables` is a Linux thing. Is there a Haiku eqivalent?

waddlesplash:
It would be nicer if it just used `/dev/mem` to read the IO space directly...

probonopd:
Is there a Haiku-ish way to get to that table? It must be somewhere in memory. Obviously the kernel ACPI driver does it. For userland, `/dev/mem` can be used to read physical pages, so maybe that?

probonopd:
Fwiw, on the Mac they can be read with `ioreg -lw0`.

OscarL:
Closing. `acpidump` works now (since c3bdcfa79b0760ff559104ae103e5f52d1bfdc4a):

```bash
> acpidump -b -n DSDT -z

Intel ACPI Component Architecture
ACPI Binary Table Dump Utility version 20230628
Copyright (c) 2000 - 2023 Intel Corporation

Writing [DSDT] to binary file: dsdt.dat 0x101DE (66014) bytes
```

PR #5849, opened 2021-04-22
iasl: bump

extrowerk:
Tested only on 64 bit


extrowerk:
We should drop acpidump, as it expects the tables to be exposed in the /dev/mem or in the /sys/firmware/acpi/tables/dynamic/* tree.

Begasus:
Builds ok on 32bit, test run shows no errors

Begasus:
> We should drop acpidump, as it expects the tables to be exposed in the the /sys/firmware/acpi/tables/* tree.

This is mentioned in an issue also: https://github.com/haikuports/haikuports/issues/4005

PR #5855, opened 2021-04-27
gcc: add GCC 11.1.0 recipe

kenmays:
Official release: https://gcc.gnu.org/gcc-11/changes.html

NOTE:
1. Migrated the legacy Haiku GCC 8.3/8.4 patchset. 
2. Configuration test passed on Haiku R1B2 x86_64.
3. Work in progress.
4. Due to 'madvise' requirement, build/test on a recent Haiku nightly > R1B2.

### DIFF
--- a/sys-devel/gcc/gcc-11.1.0_2021_04_27.recipe
+++ b/sys-devel/gcc/gcc-11.1.0_2021_04_27.recipe
@@ -1,0 +1,462 @@
+SUMMARY="C/C++ compiler for target ${effectiveTargetMachineTriple}"
+DESCRIPTION="The standard compiler for non-legacy Haiku (i.e. for all \
+architectures other than x86_gcc2)."
+HOMEPAGE="https://gcc.gnu.org/"
+COPYRIGHT="1988-2021 Free Software Foundation, Inc."
+LICENSE="GNU GPL v3
+	GNU LGPL v3"
+REVISION="1"
+gccVersion="${portVersion%%_*}"
+SOURCE_URI="https://ftpmirror.gnu.org/gcc/gcc-$gccVersion/gcc-$gccVersion.tar.xz
+	https://ftp.gnu.org/gnu/gcc/gcc-$gccVersion/gcc-$gccVersion.tar.xz"
+CHECKSUM_SHA256="4c4a6fb8a8396059241c2e674b85b351c26a5d678274007f076957afa1cc9ddf"
+SOURCE_DIR="gcc-$gccVersion"
+PATCHES="gcc-$portVersion.patchset"
+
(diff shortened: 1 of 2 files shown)


extrowerk:
Remove old so we can see what changed.

extrowerk:
Maybe you should not waste your time for one of the most complex building block. But if you insists to do it, you should spend some time with git first, otherwise it will be impossible to see what you did, as you tend to delete and recreate branches, which means all the comments will gets forgotten.

I don't really think it makes any sense to attempt this.

But about the 
```
/sources/gcc-11.1.0/gcc/config/haiku.h:206:9: error: attempt to use poisoned "NO_IMPLICIT_EXTERN_C"
 #define NO_IMPLICIT_EXTERN_C
         ^
```
error:
the GCC guys said we can safely remove that line as NO_IMPLICIT_EXTERN_C became default.

extrowerk:
It would be also a good idea to update mpfr, mpc, gmp and the binutils recipe to the version recommended for GCC11.1.0

kenmays:
Thanks. I knew about 'NO_IMPLICIT_EXTERN_C'. Haiku implemented/improved 'madvise' after Haiku R1B2 so you'll need to review moreso on the nightly builds. GCC 8.4.0 built fine on Haiku R1B2 - why not review GCC 8.4.0 build on the buildbots?

kenmays:
Issue: Haiku hrev55063 - os/BeBuild.h provides '__GNUC__ <=10' on line 52 for B_HAIKU_ABI. Need to review for '__GNUC__  <= 12' change.

extrowerk:
> GCC 8.4.0 built fine on Haiku R1B2 - why not review GCC 8.4.0 build on the buildbots?

Cant build it on nightly -> can't check.

Somebody correct me if i am wrong, but i don1t think we "review" ports on the builders. It is not some kind of CI (at least not yet). One should make sure the port actually builds and works. Mistakes happens, but we still should try our best.

extrowerk:
GCC8.4.0 is too small jump from my POV. A GCC change will require a whole repository-rebuild, as most ports depends on GCCs libs. I don't think it is economical for a minor version bump. If bump, then directly to 11 or whatever.

pulkomandy:
You are correct. Don't submit recipes and pull requests here if you have not tested them. Especially when touching core system components like gcc, test it carefully.
Once it's merged, it automatically gets packaged and installed on everyone machine. If you make us all unable to use gcc because you shipped a broken version, no one can compile things anymore. Sounds like a bad idea.

extrowerk:
```
/boot/system/develop/headers/os/BeBuild.h:55:9: error: #error Unsupported compiler!
   55 | #       error Unsupported compiler!
```
Proposed fix: https://review.haiku-os.org/c/haiku/+/3890

kenmays:
Superseded by #5855.

PR #5869, opened 2021-05-05
OpenSSH: Remove work-around script that creates user.

agmsmith:
Now that the package daemon properly runs post-install actions
in hrev54986, such as creating users, we can remove the
work-around script files that did the same job.


korli:
this can't be merged at the moment: the current stable is still beta2.

agmsmith:
> this can't be merged at the moment: the current stable is still beta2.

Good point.  Should be safe to do it soon after Beta3 comes out and doesn't do any harm (other than being inefficient) if it's delayed a bit more.

agmsmith:
Oops, accidentally deleted master branch while trying to force it to update.  Will have to regenerate the pull request...

agmsmith:
Now see https://github.com/haikuports/haikuports/pull/6014

PR #5878, opened 2021-05-08
ruby, revbump for newer libedit

Begasus:


Begasus:
@extrowerk build failed on 64bit, could you have a look? https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/5684/builds/19032.log

Issue #5845, opened 2021-04-21
python uses libreadline instead of libedit


pulkomandy:
Haiku ships with libedit by default (used by Debugger cli). We picked libedit because it is not GPL licensed.

Python can use either libedit or readline, but the haikuports version is built with readline. It should use libedit instead to reduce the number of dependencies, especially because python is shipped in Haiku release images.

pulkomandy:
Possibly needs to update Python to 3.8 or later?
https://bugs.python.org/issue13501

pulkomandy:
Also, on 64bit systems, python3 uses readline7 instead of the latest readline.

Begasus:
Maybe bump libedit to the latest version before checking it out (builds ok on R1B2 for both gcc2 and gcc8)

pulkomandy:
I tried to rebuild Python but got a crash in gcc (compiler internal error, and then another crash while trying to display that).

Begasus:
Just added a PR for libedit, will do a testrun on python (did you do the 3.7 version)?

pulkomandy:
Yes, python 3.7. I now notice that I had made another change: replace cmd:python with cmd:python3 in the REQUIRES (I have removed Python2 from my system and don't want to reinstall it). Maybe that caused the problem somehow?

PR #5880, opened 2021-05-08
python, use libedit over libreadline

Begasus:


Begasus:
Fixes: https://github.com/haikuports/haikuports/issues/5845
@pulkomandy could you have a look, build tested on R1B2 32bit and all were ok

Issue #5865, opened 2021-05-04
recipe for chafa


korli:
https://github.com/hpjansson/chafa/

media-gfx/chafa

LGPL-3+

PR #5881, opened 2021-05-08
chafa, new recipe

Begasus:


Begasus:
Fixes: https://github.com/haikuports/haikuports/issues/5865
The patch is probably wrong (no other solution found so far) @korli could you have a look?

korli:
A typedef u_int32_t __uint32_t would be simpler, maybe.

Begasus:
> A typedef u_int32_t __uint32_t would be simpler, maybe.

Doesn't seem to solve the problem, checked something comparable earlier, still getting the error then:
```
chafa-popcnt.c:45:10: error: unknown type name '__int32_t'; did you mean 'u_int32_t'?
```

Begasus:
Commented on the issue where they added the support for 32bit: https://github.com/hpjansson/chafa/issues/5

Begasus:
Builds fine on 64bit without the patch

korli:
"typedef int32_t __int32_t" maybe?

Begasus:
Think this should do it, could you have another look @korli ?

PR #5894, opened 2021-05-10
blastem_libretro: new recipe

kwyxz:
A port of the BlastEm Sega Genesis/Mega Drive emulator to libretro. This emulator is focused on being both extremely accurate and extremely fast. While it is not as mature as Genesis Plus GX, its extreme accuracy makes it capable of running many demos and test ROMs that choke other emulators, including the notoriously complex Titan Overdrive I and II demos.


kwyxz:
Looks like the Makefile is using `uname -m` to figure out what arch to build for. Not sure how exactly what the easiest way to fix that would be... short of disabling the 32bit entirely, that is.

kwyxz:
To be entirely honest, I am not sure I know how to fix this. Would it be ok if I submitted a recipe with 32bit disabled? @Begasus @diversys ?

kwyxz:
My understanding is that on an actual 32bit system, the Makefile would recognize a 32bit system when running `uname -m` and the package would build. It doesn't on the buildbot as it detects a 64bit CPU, I guess? I don't know how to get this to work unfortunately.

pulkomandy:
The buildbot is an actual 32bit system, it's nothing special. Something is forcing CPU=x86_64 in this log, where does this come from? I also see OS=Linux so I assume it is some kind of "I have no idea, let's use the default values" thing going on?

What do you check for in uname -m exactly? On 32bit Haiku it will return "BePC" so you may need to adjust for that if you expect something similar to what Linux does.

kwyxz:
Thanks for the feedback @pulkomandy. What's odd is that on 64bit Haiku, `uname -m` returns `x86_64` (as it does on Linux) which is what the Makefile is testing for. It's possible that it's confused by "BePC" and then defaults to `x86_64`. I don't have a 32bit Haiku system to test so that makes things a bit more complex, but I'll take a look.

kwyxz:
All right, looks like building for 32bit would require *a lot* of work.
- upstream code uses gnu99 standard which gcc 2.95 doesn't seem to be a big fan of
- overall, gcc-2.95 coughs on a lot of things that are enabled by default
```
cc1: Invalid option `-Werror=return-type'
cc1: Invalid option `-Werror=implicit-function-declaration'
cc1: Invalid option `-Wno-unused-value'
cc1: Invalid option `-Werror=pointer-arith'
cc1: unknown C standard `gnu99'
cc1: Invalid option `-flto'
```
- disabling all these, it's unable to build most probably because of GNU99 coding standards:
```
In file included from /boot/home/blastem/libblastem.c:6:
/boot/home/blastem/vdp.h:248: field `vdpmem' has incomplete type
In file included from /boot/home/blastem/genesis.h:11,
                 from /boot/home/blastem/libblastem.c:9:
/boot/home/blastem/m68k_core.h:100: field `ram_code_flags' has incomplete type
In file included from /boot/home/blastem/genesis.h:15,
                 from /boot/home/blastem/libblastem.c:9:
/boot/home/blastem/z80_to_x86.h:91: field `ram_code_flags' has incomplete type
In file included from /boot/home/blastem/ym2612.h:13,
                 from /boot/home/blastem/genesis.h:17,
                 from /boot/home/blastem/libblastem.c:9:
/boot/home/blastem/vgm.h:7: warning: malformed `#pragma pack'
/boot/home/blastem/vgm.h:68: warning: malformed `#pragma pack'
```
and that's barely a quarter of it.

Fixing all this would be way above my skills and take way too much of my time, sorry. I'm just going to disable 32bit as a valid target for this recipe.

pulkomandy:
You are not supposed to use gcc2 for anything unless it's a library used by BeOS applications.

You can use a modern gcc for 32bit but it requires some changes to the recipe, mainly adding $secondaryArchSuffix to several places (you can look at other recipes for examples).

The reason for this is that our 32bvit version is compiled with gcc2 as "primary" and gcc8 as "secondary" compiler (gcc2 is "primary" because of BeOS compatibility reasons, but for haikuports really, these days gcc8 is the main useful thing and as few recipes as possible should handle gcc2).

kwyxz:
Oh I see... I got confused because indeed, "gcc" is -2.95 and I had not realized "gcc-x86" is a different version. Going to take one more stab at it then. Please disregard #5899 for now. Thanks!

kwyxz:
Alright, it builds with gcc-8 now. Going to submit a patch upstream and will update the recipe accordingly. Thanks again @pulkomandy this has been invaluable information!

PR #5930, opened 2021-05-18
gemrb git recipe (SDL2)

Nugrud:
building against SDL2 compared to SDL1.2 in previous recipes fixes window-resizing and fullscreen


korli:
0.8.7 was released in August 2020. Why not just pick this one?

Nugrud:
> 0.8.7 was released in August 2020. Why not just pick this one?
> The correct version for this one is 0.8.8~git

When launched with --version option it reports 0.8.7-git not 0.8.8~git. This is the changelog since August 2020 release:
    - gameplay content for the demo
    - SDLAudio caching
    - lightning bolts, iwd2 stealth, iwd effects, pst levelup
    - scripting, audio backends, logging
    - build system, portability and performance improvements
    - externalized summoning limit, trap limit, narration text speed
    - bugfixes

korli:
> > 0.8.7 was released in August 2020. Why not just pick this one?
> > The correct version for this one is 0.8.8~git
> 
> When launched with --version option it reports 0.8.7-git not 0.8.8~git. This is the changelog since August 2020 release:

I understand, but the versions in the package manager are considered 0.8.7 > 0.8.7~git (or ~alpha or ~beta). If someone updates the 0.8.6 recipe to 0.8.7. It would override 0.8.7~git.

It's also possible to version with the number of commits since the last release: 0.8.7+549 (AUR does this I think).

Nugrud:
> It's also possible to version with the number of commits since the last release: 0.8.7+549 (AUR does this I think).

I get a warning that there is invalid character in micro version string

korli:
> > It's also possible to version with the number of commits since the last release: 0.8.7+549 (AUR does this I think).
> 
> I get a warning that there is invalid character in micro version string

Right, 0.8.7.549 would be possible (it comes after 0.8.7). What also works is for example dev-util/gn/gn-0.1891.dfcbc6fe.recipe, which is never tagged upstream, 1891 is the commits count, dfcbc6fe is the short hash for the commit.

Nugrud:
I don't know if you got the number of 549 commits from github, but I didn't know how to easily obtain that info, so I renamed the recipe with numbers of days after the 0.8.7 tag.

games-engines/gemrb/gemrb-0.8.7.270.recipe
```diff
@@ -0,0 +1,100 @@
...
+
+INSTALL()
+{
+	cd build
+	make install
+}
+
+TEST()
+{
+	cd /boot/system/settings/gemrb
```
korli, line 93: is it $settingsDir/gemrb?

games-engines/gemrb/gemrb-0.8.7.270.recipe
```diff
@@ -0,0 +1,100 @@
...
+DESCRIPTION="GemRB (Game Engine Made with preRendered Background) \
+is a port (actually a new implementation) of the original Infinity \
+Engine (the one of Baldur's Gate, Icewind Dale, Planescape: Torment, ...) \
+to Linux/Unix, MacOS X and Windows with some enhancements. Would \
+you like to create a game like Baldur's Gate? \
+(Originals games' data is not included)"
+HOMEPAGE="https://gemrb.org"
+COPYRIGHT="2003-2021 The GemRB Team"
+LICENSE="MIT"
+REVISION="2"
```
korli, line 11:
```suggestion
REVISION="1"
```

Nugrud:
I commited your suggestions manually.

PR #5939, opened 2021-05-22
zlib-ng, fix x86 build

Begasus:


Begasus:
Forgot to mark x86_gcc2 as broken, on version bump this could be done :)

PR #5910, opened 2021-05-14
lapack, bump version

Begasus:
@extrowerk could you check? This probably means an obsolete cblas recipe? https://github.com/haikuports/haikuports/pull/4046
**EDIT** all tests passed on 32bit

PR #5940, opened 2021-05-22
new Scriptureguide release wich fixes ScriptureguideManager

Paradoxianer:


Begasus:
Could you remove the previous version while bumping to this one?

Paradoxianer:
> Could you remove the previous version while bumping to this one?

I can do this but may i ask why because with the former bump i didnt need to delete the old one. Just out of curiosity :)

Begasus:
For some recipes it's useful to have an older version around, for this, it fixes things that are broken in the previous version (probably overseen with the previous version bump) :)
So there wouldn't be a need to keep the previous versions around at haikuports (keeps git log smaller as we only see what's changed in the recipe)

Begasus:
You can see it in my PR for lapack here: https://github.com/haikuports/haikuports/pull/5910/files

Begasus:
Good to go?

Paradoxianer:
👍

PR #5958, opened 2021-06-04
flac: import gentoo patch

extrowerk:


korli:
This seems to be fixed with some other way upstream: https://github.com/xiph/flac/commit/5f5035d2093421c43770eb0f0628e2f60aea61a5

unless you're in a hurry, I would wait on the release: https://github.com/xiph/flac/issues/239

Begasus:
Doesn't seem to be any new release there, maybe use a $srcGitRev checkout to fix this?

korli:
> Doesn't seem to be any new release there, maybe use a $srcGitRev checkout to fix this?

The bug is said to affect flac compiled with gcc 9.2+ (unclear in which gcc version a fix is included).

korli:
superseded by 34dce7bc0246808abba2d1f81716fc466fda6b98 (I suppose).

Issue #5960, opened 2021-06-05
Recent python versions depend on python2.7?


Begasus:
Looking at the recipes they mention `cmd:python` in the `BUILD_PREREQUIRES`
Summining @korli for this one :)

Begasus:
From Arch https://github.com/archlinux/svntogit-packages/blob/packages/python/trunk/PKGBUILD:

```
depends=('bzip2' 'expat' 'gdbm' 'libffi' 'libnsl' 'libxcrypt' 'openssl' 'zlib')
makedepends=('tk' 'sqlite' 'bluez-libs' 'mpdecimal' 'llvm' 'gdb' 'xorg-server-xvfb' 'ttf-font')
```

korli:
Please try without?

Begasus:
> Please try without?

Already did, python39 builds/packages OK without `cmd:python`

Begasus:
Seems to be introduced in commit https://github.com/haikuports/haikuports/commit/8aeb2cfbf972cbbb197e6acc8e211471d107fef6#diff-97dbb585afbff18e3304932005686557385e126e17272a51d033bff0b34963bf ?

Begasus:
So far builds for python3 python36 and python39 finished fine on x86_64 here, gonna try some builds on x86_gcc2 with other recipes depending on it (mainly python3)

korli:
Ok cmd:python can be removed then.

Begasus:
Will do a checkout on my latest PR's for numpy and scipy, if I get around to it I'll push changes later :)

Begasus:
Fixed in https://github.com/haikuports/haikuports/pull/5962

PR #5962, opened 2021-06-05
python recipes don't depend on python2.7

Begasus:


Begasus:
As reported in https://github.com/haikuports/haikuports/issues/5960

Begasus:
Checked builds for lapack, numpy and scipy (open PR's), all seem to be OK. If issues arise this could be reverted

PR #4861, opened 2020-04-09
t1utils: new recipe

jmairboeck:
Now that we have a guile1 package, I wanted to give porting Lilypond a try. It turned out that this is not so easy (I have suspected that already, as I have already fixed the openSUSE package for Lilypond and am kind-of a co-maintainer for that).

There are a few missing dependencies yet, the biggest of which is Texlive. Another missing tool is t1asm (part of t1utils), which is provided with this pull request.

Tested building on x86_64 and x86_gcc2, it works on both. (yes, using gcc2!)

This is my first HaikuPorts PR, let's see what you think.

PR #5963, opened 2021-06-06
Texlive: bump to 2021

jmairboeck:
I finally managed to get TeX Live 2021 to build using haikuporter after 14 months of trying! (I started with the 2020 version of course)

Changes from the 2014 version:
 * install-tl and the tlpdb is not fetched from subversion any more but from source tarballs
 * the package uses system libraries where feasible (everything except libkpathsea and libptexenc which are tex-specific, and libxpdf, which is library version fork of Xpdf that is not available upstream). Until 2020, poppler could be used instead of xpdf, but they dropped support for it in 2021 because of the maintenance effort.
 * t1utils is not built from texlive, but a separate package is used for it (see #4861). This could probably be also done for other included 3rd-party tools.
 * the package is (very coarsely) split into a main package and _texmf and _texmf_doc subpackages (both "any" architecture). This way, Haiku issue #12116 can be avoided.

The package and the recipe probably need some cleanup before this can be merged, but it produces a somewhat working package (tested on x86_64). It needs quite a few hours to build though, so be warned!

Also, make sure to clean the installation directory when retrying. Otherwise `make texlinks` creates symlinks pointing to themselves. If you have built such broken packages, this can be worked around with locally by creating the needed links manually in non-packaged/bin after the installation.


app-text/texlive/texlive-2021.recipe
```diff
@@ -0,0 +1,958 @@
...
+}
+
+BUILD()
+{
+	mkdir -p ../build && cd ../build
+
+	if [ $targetArchitecture = x86_gcc2 ] || [ $targetArchitecture = x86 ]; then
+		HOST=i586-pc-haiku
+	elif [ $targetArchitecture = x86_64 ]; then
+		HOST=x86_64-unknown-haiku
```
pulkomandy, line 840: this could be HOST=$targetArchitecture-unknown-haiku and work for all non-x86-32bit architectures (when we start building packages for them)

pulkomandy:
Thanks for working on this! The recipe looks clean enough to me. But I'll let haikuports experts review it in case I missed something.

app-text/texlive/texlive-2021.recipe
```diff
@@ -0,0 +1,958 @@
...
+CHECKSUM_SHA256="7aefd96608d72061970f2d73f275be5648ea8ae815af073016d3106acc0d584b"
+CHECKSUM_SHA256_2="ff12d436c23e99fb30aad55924266104356847eb0238c193e839c150d9670f1c"
+CHECKSUM_SHA256_3="74eac0855e1e40c8db4f28b24ef354bd7263c1f76031bdc02b52156b572b7a1d"
+CHECKSUM_SHA256_4="21d218dcf56160081e8c376ec9d6b4cfc750e2a0fea14ee24884b1d5039f2566"
+SOURCE_DIR="texlive-$fullVersion-source"
+SOURCE_DIR_2="texlive-$fullVersion-texmf"
+SOURCE_DIR_3="install-tl-20210324"
+SOURCE_DIR_4=""
+PATCHES="texlive-$portVersion.patchset"
+# ADDITIONAL_FILES="tlutils.pm.patch"
```
extrowerk, line 24: You could remove this if it is not used anymore.

app-text/texlive/texlive-2021.recipe
```diff
@@ -0,0 +1,958 @@
...
+# is used in many distros.  In this case, set VARTEXFONTS like this
+#VARTEXFONTS = $TEXMFVAR/fonts
+# and do not mention it in TEXMFDBS (but _do_ mention TEXMFVAR).
+#
+# Remove $VARTEXFONTS from TEXMFDBS if the VARTEXFONTS directory is below
+# one of the TEXMF directories (avoids overlapping ls-R files).
+fixKSearchPaths()
+{
+	TEXMFCNF=$sourceDir/texk/kpathsea/texmf.cnf
+	sed -i -e "s@TEXMFROOT =.*@TEXMFROOT = $dataDir/texlive@" $TEXMFCNF
```
extrowerk, line 806: You can combine this lines

app-text/texlive/texlive-2021.recipe
```diff
@@ -0,0 +1,958 @@
...
+	elif [ $targetArchitecture = x86_64 ]; then
+		HOST=x86_64-unknown-haiku
+	fi
+
+	runConfigure --omit-buildspec --omit-dirs "binDir sbinDir" \
+		$sourceDir/configure \
+		--bindir=$prefix/bin --sbindir=$prefix/bin \
+		--disable-native-texlive-build \
+		--enable-shared --disable-static \
+		--with-banner-add="/Haiku" --with-editor='lpe +%d "%s"' \
```
extrowerk, line 848: Why Pe? Can't we just use 
```
open +%d "%s"
```
here instead to use the sys-default editor?

app-text/texlive/texlive-2021.recipe
jmairboeck: I have not changed this part from the 2014 recipe. Maybe I should delete it, now that this version "works" so that one can see the differences better?

extrowerk: Yep, pls delete the old recipe then.

app-text/texlive/texlive-2021.recipe
pulkomandy: open is non-blocking and will not wait for the editor to exit. This is what lpe is for (for example this is used in git when waiting for the user to enter a commit message).

I don't know what this is used for, but maybe it is important to have this behavior.

app-text/texlive/texlive-2021.recipe
```diff
@@ -6,27 +6,35 @@ are free software, including support for many languages around the world."
...
+CHECKSUM_SHA256="7aefd96608d72061970f2d73f275be5648ea8ae815af073016d3106acc0d584b"
+CHECKSUM_SHA256_2="ff12d436c23e99fb30aad55924266104356847eb0238c193e839c150d9670f1c"
+CHECKSUM_SHA256_3="74eac0855e1e40c8db4f28b24ef354bd7263c1f76031bdc02b52156b572b7a1d"
+CHECKSUM_SHA256_4="21d218dcf56160081e8c376ec9d6b4cfc750e2a0fea14ee24884b1d5039f2566"
+SOURCE_DIR="texlive-$fullVersion-source"
+SOURCE_DIR_2="texlive-$fullVersion-texmf"
+SOURCE_DIR_3="install-tl-20210324"
+SOURCE_DIR_4=""
+PATCHES="texlive-$portVersion.patchset"
+ADDITIONAL_FILES="texlive.profile"
```
extrowerk, line 25: Shouldn't we add:
```
commandBinDir=$binDir
then
	commandBinDir=$prefix/bin
	commandSuffix=
fi
```
here?

extrowerk: If yes, we should add `$commandSuffix` to every provides and the `configure` invocation should be also adjusted with `$commandBinDir`.

app-text/texlive/texlive-2021.recipe
```diff
@@ -617,36 +821,53 @@ fixKSearchPaths()
...
 	fi
 
-	# don't use secondary arch bindir
-	runConfigure --omit-buildspec --omit-dirs "binDir sbinDir" $sourceDir/configure \
-		--bindir=$prefix/bin --sbindir=$prefix/bin --disable-native-texlive-build --disable-cxx-runtime-hack \
-		--enable-shared --disable-static --with-banner-add="/Haiku" --with-editor='lpe +%d "%s"' --with-system-icu \
-		--without-cairo --with-system-freetype2 --with-system-libpng --with-system-zlib --without-x -C \
+	runConfigure --omit-buildspec --omit-dirs "binDir sbinDir" \
+		$sourceDir/configure \
+		--bindir=$prefix/bin --sbindir=$prefix/bin \
```
jmairboeck, line 847: @extrowerk see here, this is solved differently

app-text/texlive/texlive-2021.recipe
```diff
@@ -617,36 +820,53 @@ fixKSearchPaths()
...
```
Begasus, line 846: This could be fixed as reported by @extrowerk 
```diff
- 		--bindir=$prefix/bin --sbindir=$prefix/bin \
+ 		--bindir=$commandBindDir --sbindir=$commandBindDir \

**Comment by Begasus** [comment]
Realy should check this one out, thanks for looking into it! :+1:

File: app-text/texlive/texlive-2021.recipe [hunk 8]
```diff
@@ -6,27 +6,35 @@ are free software, including support for many languages around the world."
...
+CHECKSUM_SHA256_3="74eac0855e1e40c8db4f28b24ef354bd7263c1f76031bdc02b52156b572b7a1d"
+CHECKSUM_SHA256_4="21d218dcf56160081e8c376ec9d6b4cfc750e2a0fea14ee24884b1d5039f2566"
+SOURCE_DIR="texlive-$fullVersion-source"
+SOURCE_DIR_2="texlive-$fullVersion-texmf"
+SOURCE_DIR_3="install-tl-20210324"
+SOURCE_DIR_4=""
+PATCHES="texlive-$portVersion.patchset"
+ADDITIONAL_FILES="texlive.profile"
 
 ARCHITECTURES="x86 !x86_gcc2 x86_64"
```
Begasus, line 26: No "x86" architecture for primary, so better use:
```diff
- ARCHITECTURES="x86 !x86_gcc2 x86_64"
+ ARCHITECTURES="!x86_gcc2 x86_64"
```

app-text/texlive/texlive-2021.recipe
```diff
@@ -6,27 +6,35 @@ are free software, including support for many languages around the world."
...
-PATCHES="texlive-2014.patchset"
-ADDITIONAL_FILES="tlutils.pm.patch"
-
-ARCHITECTURES="x86 !x86_gcc2 x86_64"
+REVISION="1"
+fullVersion=20210325
+SOURCE_URI="ftp://tug.org/historic/systems/texlive/$portVersion/texlive-$fullVersion-source.tar.xz"
+SOURCE_URI_2="ftp://tug.org/historic/systems/texlive/$portVersion/texlive-$fullVersion-texmf.tar.xz"
+SOURCE_URI_3="ftp://tug.org/historic/systems/texlive/$portVersion/install-tl-unx.tar.gz"
+SOURCE_URI_4="ftp://tug.org/historic/systems/texlive/$portVersion/texlive-$fullVersion-tlpdb-full.tar.gz"
```
Begasus, line 14: Maybe rename the recipe's version to the fullVersion so you could use $portVersion instead?
`texlive-20210325.recipe` ,also done by some others as shown at repology: https://repology.org/project/texlive/versions

jmairboeck: I am not really convinced. The "official" version number is just the year, and we would need to hardcode it in the directory names of the URLs instead (where $portVersion is used now). Also, if you look at the SOURCE_DIR for install-tl, it uses a slightly different date.

Begasus: Didn't look that far, thanks for the explanation :)

app-text/texlive/texlive-2021.recipe
```diff
@@ -659,66 +879,65 @@ INSTALL()
...
 	if [ ! -e $prefix/bin/mktexfmt ] ; then
-		ln -s $prefix/bin/fmtutil $prefix/bin/mktexfmt
+		ln -s fmtutil $prefix/bin/mktexfmt
 	fi
 
-	# remove the documents directory; it's too big for package create to handle!
-	# see https://dev.haiku-os.org/ticket/12116
-	#rm -rf $dataDir/texlive/texmf-dist/doc
+	# needed post-install steps apparently not done by install-tl
+	$binDir/fmtutil-sys --all
```
Begasus, line 917: This needs to be $prefix/bin instead of $binDir for 32bit:
```diff
- 	$binDir/fmtutil-sys --all
+ 	$prefix/bin/fmtutil-sys --all
```

jmairboeck: sorry, I missed that one

app-text/texlive/texlive-2021.recipe
```diff
@@ -384,77 +453,112 @@ PROVIDES="
...
+	cmd:xhlatex
+	cmd:xindex
+	cmd:xml2pmx
 	cmd:xmltex
+	cmd:yplan
 	lib:libkpathsea$secondaryArchSuffix
 	lib:libptexenc$secondaryArchSuffix
+	lib:libsynctex$secondaryArchSuffix
+	lib:libtexlua53$secondaryArchSuffix
+	lib:libtexluajit$secondaryArchSuffix
```
Begasus, line 476: Maybe add $libversion + $libVersionCompat here?
For libkpathsea it should be 6.3.3 (compat 6)
For libptexenc it should be 1.3.9 (compat 1)
For libtexlua53 it should be 5.3.6 (compat 3)
For libtexluajit and libsynctex it should be 2.1.0 (compat 2)

Begasus:
Nice, the base package is only 108.78 MiB (packaging still in progress but looking good so far) (not checked with one of the apps using LaTex yet :)

jmairboeck:
> Nice, the base package is only 108.78 MiB (packaging still in progress but looking good so far) (not checked with one of the apps using LaTex yet :)

on 64 bit it is now also just 109.68 MiB because I deleted the duplicated "custom" binaries. For testing, make sure that `make texlinks` has been called only once, otherwise it creates broken symlinks (as I already noted above).

Begasus:
> otherwise it creates broken symlinks (as I already noted above).

Yep, found out ;)

app-text/texlive/texlive-2021.recipe
jmairboeck: I will add these later (still have work to do for the day job first ...)

I think I can't really use variables for that because the values are nearly all different?

Begasus: A bit more work, but you could look at the next for an example :) 

https://github.com/haikuports/haikuports/blob/14c2cab5428145b93232cb69683a67bbe68a9f06/app-text/poppler/poppler-20.11.0.recipe#L232

extrowerk:
Can we further reduce the package size with debug subpackage?

jmairboeck:
> Can we further reduce the package size with debug subpackage?

The main package for sure, if we figure out which of the hundreds of commands are actually binaries and which are scripts (which seems a very tedious task). But the main package is now just a bit over 100 MiB. The _texmf and _texmf_doc subpackages which are the big ones, contain just data and scripts (hence the "any" architecture), so they wouldn't benefit.

extrowerk:
> > Can we further reduce the package size with debug subpackage?
> 
> The main package for sure, if we figure out which of the hundreds of commands are actually binaries and which are scripts (which seems a very tedious task). But the main package is now just a bit over 100 MiB. The _texmf and _texmf_doc subpackages which are the big ones, contain just data and scripts (hence the "any" architecture), so they wouldn't benefit.

One can extract the main package and look at the contents.

jmairboeck:
> Can we further reduce the package size with debug subpackage?

I have checked the binaries with `file` and they show as "stripped". Does a debug package help anything in this case? Or is the `file` output wrong for Haiku binaries?

extrowerk:
try to copy the binaries in an own folder and try to strip them manually then compare the resulting  file size with the original.

jmairboeck:
> try to copy the binaries in an own folder and try to strip them manually then compare the resulting file size with the original.

`du` shows the same size before and after stripping (58M for the bin folder). I used `strip -s *`.

Begasus:
Last try I ended up with broken links still :(

app-text/texlive/patches/texlive-2021.patchset
```diff
@@ -0,0 +1,145 @@
+From b5b20e0315b1e713cbe2b616996ac003f0d01c49 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Joachim=20Mairb=C3=B6ck?= <j.mairboeck@gmail.com>
```
korli, line 2: This should be credited to Jessica Hamilton <jessica.l.hamilton@gmail.com>

app-text/texlive/additional-files/texlive.profile
```diff
@@ -0,0 +1,26 @@
...
+instopt_portable 0
+instopt_write18_restricted 1
+tlpdbopt_autobackup 1
+tlpdbopt_backupdir tlpkg/backups
+tlpdbopt_create_formats 1
+tlpdbopt_desktop_integration 1
+tlpdbopt_file_assocs 1
+tlpdbopt_generate_updmap 0
+tlpdbopt_install_docfiles 1
+tlpdbopt_install_srcfiles 1
```
Begasus, line 21: Maybe for now set these as "0" ?

app-text/texlive/patches/texlive-2021.patchset
jmairboeck: How do I do that? I have mentioned her in the commit message.

jmairboeck:
> Last try I ended up with broken links still :(

My test build is now finished and works.

app-text/texlive/patches/texlive-2021.patchset
jmairboeck: should I amend the commits and change the author?

Begasus:
Added the "wip" label for now, will see when I get around to it to do a new build here on 32bit :) Thanks so far! :+1:

jmairboeck:
If there is anything (easy) that I should change about this, I can do it of course. I have thought a bit how to split up the package. This could maybe be done somehow by parsing the "collection-*" tlpdb files and generating the file lists in the recipe for it with `eval`. I'm not sure when I'll have the time to try that or if this is even feasible.

Also, should I do try these experiments in this PR or would you merge this first?

Begasus:
> I'll have the time to try that or if this is even feasible.

Don't worry, there is no rush here, I won't merge (or probably someone else even) before we know we have a working port that can be used by apps using latex I guess, like texstudio/texmaker ...) :)

app-text/texlive/texlive-2021.recipe
```diff
@@ -384,77 +464,112 @@ PROVIDES="
...
+PROVIDES_texmf="
+	texlive${secondaryArchSuffix}_texmf = $portVersion
+	"
+REQUIRES_texmf="
+	texlive$secondaryArchSuffix == $portVersion base
+	"
+
+ARCHITECTURES_texmf_doc="any"
+PROVIDES_texmf_doc="
+	texlive${secondaryArchSuffix}_texmf = $portVersion
```
Begasus, line 534: I 'think' this should be _texmf_doc (got an error for a not matching provides earlier), doing a new check atm :)

Begasus: Maybe it should even be `texlive_texmf_doc` (without `$secondaryArchSuffix` as this is a `any` package)?

jmairboeck: You are right, this is a copy-paste bug. Why didn't that show up with my test builds? (I used `haikuporter -S`). Is the resulting package actually called `texlive_texmf_doc` on 32 bit despite being a subpackage of `texlive_x86`? (and the same for `texlive_texmf` above)

jmairboeck: And if it is the same package for all architectures, can I even put a `REQUIRES=texlive$secondaryArchSuffix` there? This would break for at least one architecture then because `$secondaryArchSuffix` expands to different values.

Begasus: That's the next check ;) For now let's fix this one :)

Begasus:
Success! :1st_place_medal: 
Checked with TeXstudio and looking good :+1: :)

jmairboeck:
I have pushed another change, fixing the remaining issues with `install-tl`. It now runs without ignoring errors. I have replaced the `tlpdb` tarball with the `extra` tarball, which also contains the tlpdb file and other files which were missing before.

I have also added a special workaround to the SELFAUTO patched parts, replacing the hardcoded paths from the patchset again in the patch function with `$prefix/bin` and `$dataDir`. They expand to /packages-based paths, so that this code also works correctly inside haikuporter's chroot and can be used during INSTALL by `install-tl`. It uses this for finding the `tlpostcode` scripts to execute after installing the TeX packages.

jmairboeck:
I have changed the author of the patchset commits and marked the architectures as "untested".

app-text/texlive/texlive-2021.recipe
```diff
@@ -384,77 +464,112 @@ PROVIDES="
...
+	devel:libptexenc$secondaryArchSuffix = $libptexencVersionCompat
+	devel:libsynctex$secondaryArchSuffix = $libsynctexVersionCompat
+	devel:libtexlua53$secondaryArchSuffix = $libtexluaVersionCompat
+	devel:libtexluajit$secondaryArchSuffix = $libtexluajitVersionCompat
 	"
 REQUIRES_devel="
 	texlive$secondaryArchSuffix == $portVersion base
 	"
 
+ARCHITECTURES_texmf="any"
```
Begasus, line 524: Maybe disable this one too (largest package also) :)
```diff
-ARCHITECTURES_texmf="any"
+ARCHITECTURES_texmf="?any"
```

Begasus:
Thanks! :+1:

PR #5965, opened 2021-06-06
Samba4: new recipe

extrowerk:
A huge pile of hack, because i am too simple minded to solve basic problems.


extrowerk:
I still see the famous:
```
messaging_dgm_init: socket failed: Address family not supported by protocol family
messaging_dgm_ref: messaging_dgm_init returned Address family not supported by protocol family
messaging_dgm_init failed: Address family not supported by protocol family
messaging_dgm_ref failed: Address family not supported by protocol family
```
errors.

Begasus:
@extrowerk stepping in here for a  bit, removed the now obsolete packages (those are already included in the repo).
So far I have a build for samba4, needs finetuning/cleaning up first.
Thanks for the work so far!

net-fs/samba/samba4-4.14.5.recipe
```diff
@@ -0,0 +1,592 @@
...
+	export PATH="$portPackageLinksDir/cmd~python3/bin:$PATH"
+	pythonVersion=$(python3 --version 2>&1 | sed 's/Python //' | head -c3)
+	installLocation=$prefix/lib/python$pythonVersion/vendor-packages/
+
+	mkdir -p $installLocation
+	mv $prefix/non-packaged/lib/python$pythonVersion/site-packages/* $installLocation
+	rm -rf $prefix/non-packaged
+
+	# HACK
+	ln -s $libDir/samba/*.so* $libDir
```
Begasus, line 566: @extrowerk remember why this was needed? (link the libraries back into $libDir)

extrowerk:
AFAIR the samba cli tools werent able to locate their libs as they werent installed in the standard libpath.
Test the cli tools to see if it is still the case.
My hack is probably not a proper solution (therefore the comment), maybe one can pass or adjust the RPATH or some similar variable which affects the default searching paths the runtime-linker uses. I never had time to research RPATH and how it works, so maybe it is not even the correct variable. If so, forgive me.

Schrijvers Luc írta 2024. júl.. 13, Szo-n 10:17 órakor:

Begasus:
It's all to early to tell if anything works at all, so thanks for getting back to me, lot's of those *samba4* libraries are not in the install, so I removed them, when I'm done polishing (cleaning) I'll push here if you don't mind, no need to open another PR for this.
If nothing else someone could pick up and go frem there.

Begasus:
Pushed recipe/patchset for 4.20.2 here, left it disabled as I don't have any real means to check it, some directories mentioned for /var I created manually (could be improved). :)

Begasus:
Marked both 4 versions as untested, merging for now.

PR #5987, opened 2021-06-12
dsfmt: new recipe, v2.2.5

vazub:


Begasus:
Could you see if this builds on 32bit also?

vazub:
Sorry, @Begasus, no can do. Don't have necessary setup ready and frankly not interested enough in that arch to get one. This port is intended for x86_64 only, deliberately, but please feel free to take over and update it with further support options as you see fit.

Begasus:
@vazub no problem, we'll merge and if need be one can still look into creating the changes to make it build on x86_gcc2 ;)

PR #5988, opened 2021-06-13
libqt5pas, new recipe

Begasus:


dev-qt/libqt5pas/libqt5pas-1.2.9.recipe
```diff
@@ -0,0 +1,66 @@
...
+	devel:libQt5Pas$secondaryArchSuffix = $portVersion
+	"
+REQUIRES_devel="
+	libqt5pas$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	lib:libqt5Core$secondaryArchSuffix
+	lib:libqt5Gui$secondaryArchSuffix
```
korli, line 41: Are you sure with lib:?

Begasus: jikes :) strangly enough it didn't error out there ;)

Begasus: Thanks ! Fixed now :)

dev-qt/libqt5pas/libqt5pas-1.2.9.recipe
```diff
@@ -0,0 +1,66 @@
+SUMMARY="Pascal wrapper for Qt5"
+DESCRIPTION="libqt5pas is an interface between Lazarus and the Qt5 libraries.
+Not all Qt5 functions are available, only those necessary for Lazarus \
+functionality."
+HOMEPAGE="https://wiki.freepascal.org/Qt5_Interface"
+COPYRIGHT="2005-2013 by Jan Van hijfte"
```
Begasus, line 6: Not sure here, I see Jan as the original maintainer for libqt4pas, but I think Željan should be added as the current one (can't find it atm)

Begasus: Found it :)

dev-qt/libqt5pas/libqt5pas-1.2.9.recipe
```diff
@@ -0,0 +1,77 @@
...
+}
+
+INSTALL()
+{
+	#hack to fix the install path
+	if [ "effectiveTargetArchitecture" = x86_64 ]; then
+		sed -i "s|/boot/system/lib||" Makefile
+	else
+		sed -i "s|/boot/system/lib/x86||" Makefile
+	fi
```
Begasus, line 66: `target.path = $$[QT_INSTALL_LIBS]` seems to hardcode the patch to the Makefile, haven't found another way to fix this for now

Begasus:
Not the best solution atm, but it's working for now ...

PR #6006, opened 2021-06-19
ytdl_gui: Bump to 2.5

win8linux:
Update YouTubeDL-GUi to 2.5, which contains mainly bugfixes. It also adds [subtitles](https://github.com/JaGoLi/ytdl-gui/commit/640277e40e44eda357bb1107fbc52e5a4d21696e) and [Ogg Vorbis](https://github.com/JaGoLi/ytdl-gui/commit/6b3813fa8d49c2406f0bfd7f8de10fa31e94362c) support.


net-misc/ytdl-gui/ytdl_gui-2.5.recipe
```diff
@@ -0,0 +1,81 @@
...
+	devel:libQt5Svg$secondaryArchSuffix
+	devel:libQt5Widgets$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:lrelease$secondaryArchSuffix >= 5
+    cmd:cmake
```
Begasus, line 54: Use tab instead of spaces here

net-misc/ytdl-gui/ytdl_gui-2.5.recipe
```diff
@@ -0,0 +1,81 @@
...
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:qmake$secondaryArchSuffix >= 5
+	"
+
+BUILD()
+{
+	cmake -B build
+    make -C build
```
Begasus, line 63: Same here

Begasus:
Could you remove the previous version while bumping version?

Begasus:
Should be ok now, need to run a test here to make sure :) (nice to see that --all-subs has made it in) :)

win8linux:
Looking through the changes, it seems that `qmake` got replaced by `cmake`. I'll try changing the deps accordingly and test it.

net-misc/ytdl-gui/ytdl_gui-2.5.recipe
Begasus: Not resolved ;) (and further also)

win8linux: Resolved for real. :P

win8linux:
Not going to rename the package despite the binary name change, to avoid breaking update continuity of existing installs. I'd be willing to do it, if there are reassurances that doing so will not break updates of current installs.

net-misc/ytdl-gui/ytdl_gui-2.5.recipe
```diff
@@ -51,20 +51,21 @@ BUILD_PREREQUIRES="
...
+	cmake -B build $cmakeDirArgs \
+		-DCMAKE_INSTALL_BINDIR=$appsDir
 	make build
 }
 
 INSTALL()
 {
-	mkdir -p $appsDir
-	mv ./src/youtubedl-gui $appsDir/ytdl-gui
+	make install -C build
```
Begasus, line 69: The files in $dataDir are useless to us (we are not linux) so best to remove them from the build:
```diff
+ 	rm -rf $dataDir
```

PR #6014, opened 2021-06-22
OpenSSH: Remove work-around script that creates user.

agmsmith:
Now that the package daemon properly runs post-install actions
in hrev54986, such as creating users, we can remove the
work-around script files that did the same job.  Make this
active for R1Beta3 for minor compatibility reasons.

Replaces https://github.com/haikuports/haikuports/pull/5869


Begasus:
Pinging @korli can this be merged now? Seeing buildmaster is already running on Beta3?

korli:
@begasus imo a beta2 branch should first be created from master

PR #6015, opened 2021-06-23
BeShare: bump version

agmsmith:
New recipe for BeShare, now actually works.  Can do file uploads and resize the window dividers.  Has the new package info data column too.

Works by making packages out of existing executables for all platforms.  If there's Haiku for 32 bit PPC, this will build a package for it too, assuming that BeOS PPC executables can run on it :-)


kallisti5:
Is this one ready? I need a ~victim~ guinea pig to test the haikuporter buildmasters running the r1/beta3 build environment :smiling_imp:

agmsmith:
It's ready, but the build step is a no-operation.  So it won't test much other than repackaging the contents of a .zip file into an .hpkg file.

diversys:
Is there no source code available for this version?

agmsmith:
Yes, there is source code:
SOURCE_URI="http://web.ncf.ca/au829/BeOS/BeShare_3.04.zip"

Though it doesn't build the annoying source package, unless you really want one?

kallisti5:
the source package is self-generating.  Why would it need to be disabled?

agmsmith:
I'll see what enabling the source package does...

agmsmith:
Now with a source package too!

diversys:
I mean why does it need to be in a form of precompiled binaries and not build from source code as every other recipe does?

kallisti5:
oh.. you're 100% correct diversys.  Why is this recipe consuming binaries?  Generally this stuff should be compiled by HaikuPorter. 

extremely complex things like rust use pre-compiled binaries since the bootstrap process is hard to replicate within haikuporter. This seems like just a basic application build though?

agmsmith:
Because of a lack of time mostly and the desire to get a more recent and working BeShare to the R1B3 users.  If you want to update the sources to work with the latest compiler, feel free to have at it.  Also there's no PPC compiler :-)

pulkomandy:
Please compile it from inside the haikuports recipe properly. We do not normally accept pre-built binary packages in the haikuports repository.

Also, are all the issues with the new Muscle versions fixed? We had stayed for a while on BeShare 2 because the 3.x versions had problems uploading/downloading files. However, at some point we had a recipe for it, which should be possible to extract from the Git history and restore, if these issues are fixed now.

agmsmith:
OK, will have a look at getting it to compile.

agmsmith:
Now compiles the code in the Haikuporter recipe.  Tested on gcc2 and x86_64 systems.

diversys:
It would be great if you could rename beshare recipe (instead of deleting and creating a new one) so that a diff could be seen.

agmsmith:
> It would be great if you could rename beshare recipe (instead of deleting and creating a new one) so that a diff could be seen.

Good idea, I'll do a bit of git-foo and retcon a bit of continuity into the history.

pulkomandy:
Git does not track renames, it only detects if files are similar enough when showing a diff and decides "oh, this must be a rename".

If you do too much changes, it decides instead that they are different files. There is not much that can be done about it.

diversys:
Right, but is it possible to guess when this is done my git or it wasn't a rename in the first place?

agmsmith:
Well, after redoing it as a rename, git changes it back to a delete and create new file when I squash the commits down to one.  Can do it if I submit two commits.  Back in a few minutes, redoing the rename yet again.

agmsmith:
Now with two commits so when you view the main one on GitHub, you see the differences with the previous recipe.

haiku-apps/beshare/beshare-3.04.recipe
```diff
@@ -0,0 +1,85 @@
...
+REVISION="3"
+SOURCE_URI="http://web.ncf.ca/au829/BeOS/BeShare_3.04.zip"
+CHECKSUM_SHA256="37b9e590d2f3f5ce1395fdb6a31e8849ce0d286a79a53908476ea70f942987a2"
+SOURCE_FILENAME="BeShare_3.04.zip"
+SOURCE_DIR="BeShare"
+
+ADDITIONAL_FILES="beshare-post-install.sh"
+POST_INSTALL_SCRIPTS="$relativePostInstallDir/beshare-post-install.sh"
+
+# Note this source code is portable and should compile on all platforms (PPC,
```
diversys, line 37: I guess this comment is not needed anymore?

haiku-apps/beshare/beshare-3.04.recipe
```diff
@@ -0,0 +1,85 @@
...
+# all x86 varieties), and for both BeOS and Haiku.  AGMS20210629
+ARCHITECTURES="x86_gcc2 ?x86 x86_64 ?arm ?ppc"
+
+PROVIDES="
+	beshare$secondaryArchSuffix = $portVersion
+	app:BeShare = $portVersion
+	"
+
+# We can use Haiku versions back to when package management was created, but
```
diversys, line 46: Same as above.

haiku-apps/beshare/beshare-3.04.recipe
```diff
@@ -0,0 +1,85 @@
...
+		CXXFLAGS="-DMUSCLE_64_BIT_PLATFORM"
+	fi
+	export CXXFLAGS="-D_BSD_SOURCE $CXXFLAGS"
+	cd source/beshare
+	make $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p -v "$appsDir"
```
diversys, line 74: I think quotes are superfluous here and below.

haiku-apps/beshare/beshare-3.04.recipe
agmsmith: Needed to prevent bit-rot by novice programmers accidentally making things non-portable.

haiku-apps/beshare/beshare-3.04.recipe
agmsmith: Makes for buggy code, which doesn't handle spaces in path names.  But then everyone else does the same thing.

haiku-apps/beshare/beshare-3.04.recipe
agmsmith: Yup, obsolete, will be removed.

agmsmith:
Now without the unnecessary obsolete comment and without quote marks around file paths.

haiku-apps/beshare/beshare-3.04.recipe
```diff
@@ -0,0 +1,79 @@
...
+	mkdir -p -v $appsDir
+	cp -av $sourceDir/source/beshare/BeShare $appsDir/BeShare
+	chmod -v a+rx $appsDir/BeShare
+	addAppDeskbarSymlink $appsDir/BeShare
+
+	mkdir -p -v $postInstallDir
+	cp -av $portDir/additional-files/beshare-post-install.sh $postInstallDir
+	chmod -v a+rx $postInstallDir/beshare-post-install.sh
+
+	mkdir -p -v $docDir
```
diversys, line 77: I think it is preferred to have self-contained folder with everything, so 
```
mkdir -p $appsDir/BeShare/apps
cp -r documentation/* $appsDir/BeShare/apps
```

diversys (approved):

diversys:
It takes 16 seconds for BeShare to start now. :(

haiku-apps/beshare/beshare-3.04.recipe
agmsmith: On the other hand, the documentation should be in the standard documentation folder, so there's only one thing in the apps directory, and then apps/beshare/beshare looks redundant.  Not that the user ever sees it unless they're trying to start the program from the command line.  But yes, if there was more stuff then the extra directory would be used.

Hmmm, should really have a menu item in the program that shows the documentation...

agmsmith:
> It takes 16 seconds for BeShare to start now. :(

Sounds like a DNS timeout, starts up in 0.1 seconds over here.  Make sure it isn't trying to log into an old dead BeShare server by default.

Anyway, works nicely, other than a BTextView scroll offset glitch that goes away if you resize the window.  Happens only in Haiku (resize a pane to make wrap width change, scroll back, do "/msg butler catsup" and see text scroll to the wrong position, resize whole window and it gets the correct scroll offset again).  Works fine in BeOS with the same code.

PR #6019, opened 2021-06-25
Update to perl 5.34.0 and major rework of perl recipe

thibaultduponchelle:
* Clean patchs that are no longer needed (merged upstream)
* Configure defaults are already all good.
* Do not use/install sitecustomize
* Remove sed command in perl recipe
* Simplify drastically configure arguments
* No global writable directory
* Change site prefix


thibaultduponchelle:
@olta

olta:
@thibaultduponchelle - I'm sorry, but could you please refrain from pulling me into these discussions? I have given up work on Haiku many years ago and I do not intend to restart it.

pulkomandy:
Please delete the previous version of the patch and recipe. Git will then detect that the new version is "moved" from the previous one, and github will show a diff from the previous to the current version, making it possible to review the changes more easily.

dev-lang/perl/patches/perl-5.34.0.patchset
```diff
@@ -0,0 +1,406 @@
+From 3410e7e2e37488e45d2914ca6cf02b6d1dbe4584 Mon Sep 17 00:00:00 2001
```
thibaultduponchelle, line 1: I removed all sitecustomize patchs, the removal of hints/ (I use it), the configure patchs (they are merged upstream)

dev-lang/perl/perl-5.34.0.recipe
```diff
@@ -0,0 +1,127 @@
...
+	cmd:ptardiff = $portVersion compat >= 5
+	cmd:ptargrep = $portVersion compat >= 5
+	cmd:shasum = $portVersion compat >= 5
+	cmd:splain = $portVersion compat >= 5
+	cmd:streamzip = $portVersion compat >= 5
+	cmd:xsubpp = $portVersion compat >= 5
+	cmd:zipdetails = $portVersion compat >= 5
+	# vendor_perl refers to the path to packaged perl modules, which includes
+	# the short perl version, so it is backwards compatible up this one.
+	vendor_perl = $perlShortVersion compat >= 5
```
thibaultduponchelle, line 64: I changed this, because I have problems when switching between perl 5.32 and perl 5.34 (conflicting version for other packages using vendor_perl and pkgman was asking to uninstall packages e.g. git)

dev-lang/perl/perl-5.34.0.recipe
```diff
@@ -0,0 +1,127 @@
...
+	cmd:gcc
+	cmd:ld
+	cmd:make
+	cmd:sed
+	cmd:grep
+	"
+
+perlArchName="$(uname -m)-haiku"
+
+GLOBAL_WRITABLE_FILES="
```
thibaultduponchelle, line 85: This was not empty (because sitecustomize)

dev-lang/perl/perl-5.34.0.recipe
```diff
@@ -0,0 +1,127 @@
...
+	"
+
+BUILD_PACKAGE_ACTIVATION_PHASE=INSTALL
+
+BUILD()
+{
+	./Configure -des -Dprefix=$prefix -Dsiteprefix=$prefix/non-packaged
```
thibaultduponchelle, line 92: This is now almost empty, reusing almost all perl defaults

dev-lang/perl/perl-5.34.0.recipe
```diff
@@ -0,0 +1,127 @@
...
+	make $jobArgs
```
thibaultduponchelle, line 93: I removed a sed command, not sure why it was there

dev-lang/perl/perl-5.34.0.recipe
```diff
@@ -0,0 +1,127 @@
...
+INSTALL()
+{
+	make install
+	chmod a+x $binDir/{perl,perlthanks}
+}
+
+TEST()
+{
+#x86_64
+#Failed 19 tests out of 2439, 99.22% okay.
```
thibaultduponchelle, line 105: I do not updated this command, but there is still a bunch of failing tests but the list has evolved

thibaultduponchelle (commented):
> @thibaultduponchelle - I'm sorry, but could you please refrain from pulling me into these discussions? I have given up work on Haiku many years ago and I do not intend to restart it.

OK sorry @olta I do not wanted to disturb you but just politely ping you. Well noted 👍🏼 😄

dev-lang/perl/perl-5.34.0.recipe
```diff
@@ -83,47 +83,20 @@ BUILD_PREREQUIRES="
...
-		-Dlibc="'$(finddir B_SYSTEM_LIB_DIRECTORY)/libroot.so'" \
-		-Dlibs=-lnetwork -Dcc=gcc -Dld=gcc \
-		-Ud_link -Ddont_use_nlink -Ud_syserrlst \
-		-Dldlibpthname=LIBRARY_PATH -Dstartperl="#! perl" \
-		-Dccdlflags="-Wl,-fno-stack-protector,-rpath=$libDir/perl5/$portVersion/$perlArchName/CORE" \
-		-Dusesitecustomize \
-		-de
-
-	# force-remove path from sed
-	sed -i -e 's/LOC_SED\s*"\/bin\/sed/LOC_SED "sed/' config.h
```
thibaultduponchelle, line 115: I removed this, not sure why it was there, I'm maybe wrong

thibaultduponchelle (commented):
> Please delete the previous version of the patch and recipe. Git will then detect that the new version is "moved" from the previous one, and github will show a diff from the previous to the current version, making it possible to review the changes more easily.

Done :)

dev-lang/perl/perl-5.34.0.recipe
pulkomandy: The idea of this dependency is that perl modules are specific to a given Perl version. That's because of paths like "/boot/system/non-packaged/lib/perl5/site_perl/$Config{version}/$Config{archname}" which have the complete version in it.

So, modules built for one version will not work with another.

pkgman is correct when it wants to uninstall things: they will not work out of the box with the new Perl version because this path needs to be changed (and probably the perl modules need to be rebuilt).

We have the same problem for Python, and what we did there is allow multiple packages to be installed (so we have python3, python38 and python39 packages) to allow progressive migration from one to another (after rebuilding all packages).

extrowerk:
Builds fine on 64 bit.

dev-lang/perl/perl-5.34.0.recipe
thibaultduponchelle: Ok then I was just wrong and will reuse old value for compat

Thank you

thibaultduponchelle: But then I have this: 

Screenshot from 2021-06-25 11-12-37

Is it because the new version of perl is not in official haikuports repo?

pulkomandy: If you recompile git locally it should use the new Perl version. Then it shuold be possible to install that new Git version.

As I said, we made some changes to Python packages to handle this case (there are multiple packages with different names, and they can be installed side by side). You can look at how it's done there, maybe. Or we can just rebuild everything that depends on Perl and have only the new version remain in the repository (with all perl-dependant packages adjusted to rely on the new version)

thibaultduponchelle: OK I tested what you said and pushed a new commit to revert this change

waddlesplash:
@korli, you updated perl last, do you have any input here?

Begasus:
Conflicting file now, could this be looked into?

thibaultduponchelle:
Updated to remove the conflict 👍🏼

pulkomandy:
> Do not use/install sitecustomize

Can you clarify how this works now? The "site customize" setup allowed to have perl packages coming both from our repositories (as hpkg files) and from cpan (installed in /system/non-packaged). How is this handled now? Was this code moved upstream?

thibaultduponchelle:
> > Do not use/install sitecustomize
> 
> Can you clarify how this works now? The "site customize" setup allowed to have perl packages coming both from our repositories (as hpkg files) and from cpan (installed in /system/non-packaged). How is this handled now? Was this code moved upstream?

No, I removed this feature by mistake

Begasus:
Any progress here?

thibaultduponchelle:
I'm going to close this, it's not complete

PR #6021, opened 2021-06-27
smooth: bump to 0.9.8

enzo1982:


Begasus:
Maybe next time add some description, like smooth needs to go before boca and boca needs to be done before freac?

enzo1982:
> Maybe next time add some description, like smooth needs to go before boca and boca needs to be done before freac?

Alright, will do. Sorry for the trouble!

PR #6029, opened 2021-06-30
fpc_bin, change recipe, enable x86_64

Begasus:


Begasus:
Both arch's have been checked and build ok so far, need to check for fpc-3.2.2 on both arch's

Issue #2746, opened 2018-06-27
BeZilla 2.0.0.22-7: Not supported on many popular websites (EoL/Decommision?)


kenmays:
Please review decommissioning BeZilla 2.0.0.22 as most popular websites do not support this version and it renders many business and media-related websites incorrectly (versus WebPositive 1.1).

NOTE: Mainly, does not work at all with certain websites (i.e. unsupported) or will misrender them worse than when viewed with WebPositive 1.1..

Although the source release is 10 years old, the application's main rendering functionality needed for browsing/supporting today's web servers is the main issue.

scottmc:
We should at least update the HaikuDepot description to note the age of this if it hasn't already been noted there.

extrowerk:
In the same time we should mark it and the Mail companion as untested for every arch, so if somebody really needs them, could still build it. I would remove them from depot too.

diversys:
I just want to note that when last year(?) when it so happened that both Web+ and Qt browsers were broken Firefox was the only browser that still worked which I used it to find and download previous version of Qt to get QupZilla working again :)

pulkomandy:
> I just want to note that when last year(?) when it so happened that both Web+ and Qt browsers were broken Firefox was the only browser that still worked which I used it to find and download previous version of Qt to get QupZilla working again :)

I would recommend NetSurf in that case!

Anyways, rather than removing old software from the depot, I would
rather rely on user ratings and comments so people can make an informed
decision on it. And it's not called Firefox so it's unlikely that people
would find it by accident anyways.

kenmays:
Thanks everyone. I'll close this ticket as I now agree with keeping around this old browser - since I've experienced a few legacy commercial apps that  misrender on newer Firefox browsers.

As mentioned, since it is not labeled as Firefox, most people will consider it a BeOS legacy app
due to remaining info on the Web like: https://www-archive.mozilla.org/ports/beos/

PR #6032, opened 2021-07-01
Bezilla: disabled build for R1B3 x86 and x86_64

kenmays:
Tested on Haiku R1B3 (hrev55181+16) x86_gcc2 and hrev55204 x86_64. Now disabled for R1B3+. 
NOTE: Compiles successfully with GCC 2.95.3 on x86_gcc2. 
Ref# #2746 - decommision bezilla from future R1B3+ build efforts.


diversys:
@kenmays let me quote you from #2746

> Thanks everyone. I'll close this ticket as I now agree with keeping around this old browser - since I've experienced a few legacy commercial apps that misrender on newer Firefox browsers.

Why did you decide to remove it now all of a sudden? It did't hurt keeping it available in the repo even if it is rarely used. I think we should revert it.

kenmays:
@diver - The release of Haikuwebkit 1.8.2+ (i.e. Webkit 612.1.21). I use haikuwebkit with the current Webkit 612.1.25 updates.

Realize it is 13 years since the Firefox 2.0.0.20 release, BeZilla was considered unsupported by Mozilla since Y2012 (see below).

extrowerk mentioned to mark the BeZilla recipe as 'untested' on every arch. Pulkomandy mentioned using NetSurf versus BeZilla. Korli mentioned something similar with NetSurf. 

The Otter Browser/NetSurf/Web+ browsers are current;y suited for most web browser tasks done by Haiku's non-developers.

NOTE: As of post Y2011 - Ref: https://support.mozilla.org/en-US/questions/911431
`There are only 2 currently updated and supported Released versions, Firefox 9.0.1 and Firefox 3.6.25. Support for the Firefox 3.6.x versions is likely to be discontinued in the first quarter of 2012. `

Also: 'Firefox 2.0.0.20 does not include Phishing Protection.' - Mozilla

X547:
I also disagree with Bezilla removal. It has historical importance and it often works better than Netsurf. It is feature complete compared to WebPositive and can be used as reference.

kenmays:
Resolved: https://build.haiku-os.org/buildmaster/master/x86_gcc2/

PR #6050, opened 2021-07-08
New recipe: PonyExpress

wattoc:
This is an app I've been working on for a little while now which lets you connect your Dropbox to Haiku and sync files, etc.

I'd like to get some more people using it and testing it at this point, although I'd recommend perhaps creating a new Dropbox account to use with it as it's still pretty early in development.

Comments, etc. welcomed and would appreciate if someone could test on x86 and gcc2 (I suspect GCC2 won't work as I've been a bit sloppy with the dev, assuming the pre-reqs work though I can try to get it building on GCC2 also.)

### DIFF
--- a/haiku-apps/ponyexpress/ponyexpress-0.1.1.recipe
+++ b/haiku-apps/ponyexpress/ponyexpress-0.1.1.recipe
@@ -1,0 +1,47 @@
+SUMMARY="A cloud file syncing application"
+DESCRIPTION="PonyExpress is a cloud file syncing application. \
+It currently supports DropBox full folder syncing.
+The application is still in Alpha and requires more testing \
+so please take care using it for important data."
+HOMEPAGE="https://github.com/wattoc/PonyExpress"
+COPYRIGHT="2021 Craig Watson"
+LICENSE="BSD (3-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/wattoc/PonyExpress/archive/refs/tags/v$portVersion.tar.gz"
+CHECKSUM_SHA256="eed85af52c400511298d9d5adcfa5b418dd7a7c4049a70462903945d1ceed9a1"
+SOURCE_DIR="PonyExpress-$portVersion"
+
+ARCHITECTURES="?x86_gcc2 ?x86 x86_64"
+
+USER_SETTINGS_FILES="settings/ponyexpress.settings"
+
+PROVIDES="
+	ponyexpress = $portVersion
+	app:ponyexpress = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libcurl$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libcurl$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:gcc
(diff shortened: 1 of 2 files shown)


Begasus:
Build fails with either gcc2 or gcc8, submitted an issue on your github page with the details, best bet would be to go with gcc8 (this would mean adding SECONDARY_ARCHITECTURE="x86" etc) :)

wattoc:
Thanks for checking @Begasus - from the output looks like it's complaining about Curl includes (not sure why, looks as if Curl is on gcc2) and normal old C++ standards stuff.

The gcc8 build is complaining about function pointer casts which appears to be in the Haiku headers...

I'll try to get an x86 environment up as it'll be a bit easier to figure out from there.

Begasus:
You already made it happen on 64bit (with gcc8), might as well go for gcc8 on 32bit too :)

wattoc:
Got it working apparently on both gcc2 and gcc8, will update the recipe and a new tag after some further testing.

GCC2 issues were mainly down to declarations so it was fairly trivial to fix, plus needed some include paths added.

Begasus:
I'll clone your work later and give some build tests here if you are ready, building another thing atm that could take some time :)

wattoc:
Picking up some bugs in this environment, so may be a while before I get an update out...

wattoc:
The remaining issues are:
1. The Deskbar replicant not appearing when compiled in gcc8 (I guess this may be normal?)
2. A temporary file seems to be added when eg. saving Pe in the Dropbox folder - this is getting notified which is then getting queued to add - by the time it gets to the add the file is already gone and things get confused (it'll just loop trying to upload the missing file.)  Since this is a potential legit issue elsewhere, I will aim to fix it.  I'm not sure why the behaviour is different between x64 and x86 here though, as I never saw this on x64.
3. None of the messaging back to the Deskbar icon seems to be received - it is registering with the main App correctly but I can't seem to get anything back through to the Deskbar icon.  This means that you get no indication of any activity from the Deskbar icon.  Not an issue on x64 so another mystery one.

wattoc:
Fixed the issue with the Deskbar indicators not working.  Looked to be a ref passing thing, I guess later GCC is better at detangling those.

There's still a bit of an issue with the updates where it seems to be intermittently missing changes - think this is due to my "fix" for the temp files problem, so it'll need more work.

I can confirm it does build on x86, gcc2, x64 without issues but the x86 (gcc8) does not seem to add to Deskbar which I'm assuming is a gcc2 Deskbar vs gcc8 replicant thing.

Begasus:
Did you bump the source package with the latest changes? I don't see any new release there :)
Other then that, confirmed build ok with gcc2 and no Deskbar icon with gcc8, so this should be ok now :)

wattoc:
I thought I had, may have messed it up with the tag though.

I'll check the Deskbar for issues too, did it crash or kill the Deskbar on exit?

I want to get the final bug resolved, hopefully I can figure it out today and get another update done.

Begasus:
> I thought I had, may have messed it up with the tag though.
My bad, should have checked locally :)

I'll check the Deskbar for issues too, did it crash or kill the Deskbar on exit?
With a build in Terminal it crashed, with haikuporter it's ok and doesn't crash :) Merging :+1:

PR #6060, opened 2021-07-15
pugixml, bump version

Begasus:


Begasus:
In preparation to bump OSL -> OCIO -> OIIO
Used by:
- [ ] osl
- [ ] photivo
- [ ] mudlet
- [ ] mame
- [ ] gerbera (crash reported in the issues here)

PR #6075, opened 2021-07-18
new recipe: extractpdfmark

jmairboeck:
This is a tool for optimizing PDFs generated by TeX.
It is an optional dependency for LilyPond (but actually using it when building
LilyPond breaks its build because of problems with GhostScript).

Note: this does not actually depend on texlive, so it can be built and provided
in HaikuPorts.


jmairboeck:
note that `gs` and `diff` are technically only required for tests but putting them into `TEST_REQUIRES` doesn't work, presumably because the configure script checks for them and causes the tests to be actually skipped.

jmairboeck:
tested on x86_64 and x86_gcc2: works on both

PR #6077, opened 2021-07-18
libpurple, symlink libjabber in libs

JadedCtrl:
The XMPP plugin (libxmpp) requires libjabber as a shared lib, but it's default install path ($libs/purple-2/) isn't used in this case― a symlink does the trick just fine, but I don't know if that's good practice…


Begasus:
I've been trying to find libxmpp but can't find it here, is it part of the build process for Cardie?

dev-libs/libpurple/libpurple-2.14.5.recipe
```diff
@@ -168,6 +168,8 @@ INSTALL()
 		$libDir/purple-2/*.la \
 		$libDir/lib*.la
 
+	ln -s $libDir/purple-2/libjabber.so.0 $libDir/libjabber.so.0
```
Begasus, line 171: atm libjabber.so links to libjabber.so.0.0.0, maybe symlink that one into $libDir? (not tested here)

JadedCtrl: Just pushed― either's working fine on my system :)

Begasus:
Merging until we can find a proper solution for it, thanks @JadedCtrl

PR #6078, opened 2021-07-18
new recipe: dblatex

jmairboeck:
This is a python script to convert DocBook to LaTeX.
It is another (optional) dependency of LilyPond.

Note: this depends on texlive and can therefore not be built as long as that is disabled.

### DIFF
--- a/app-text/dblatex/dblatex-0.3.12.recipe
+++ b/app-text/dblatex/dblatex-0.3.12.recipe
@@ -1,0 +1,64 @@
+SUMMARY="DocBook to LaTeX/ConTeXt Publishing"
+DESCRIPTION="dblatex is a program that transforms your SGML/XML DocBook \
+documents to DVI, PostScript or PDF by translating them into pure LaTeX as a \
+first process. MathML 2.0 markups are supported, too. It started as a clone \
+of DB2LaTeX."
+HOMEPAGE="http://dblatex.sourceforge.net/"
+COPYRIGHT="2003–2020 Benoit Guillon"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/project/dblatex/dblatex/dblatex-$portVersion/dblatex3-$portVersion.tar.bz2"
+CHECKSUM_SHA256="16e82786272ed1806a079d37914d7ba7a594db792dc4cc34c1c3737dbd4da079"
+SOURCE_DIR="dblatex3-$portVersion"
+
+ARCHITECTURES="?any"
+
+PROVIDES="
+	dblatex = $portVersion
+	cmd:dblatex
+	"
+REQUIRES="
+	haiku
+	cmd:convert
+	cmd:epstopdf
+	cmd:kpsewhich
+	cmd:latex
+	cmd:makeindex
+	cmd:pdflatex
+	cmd:python3
+	texlive_texmf # TODO: this must be changed when the texlive_texmf package is split up
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	cmd:convert
+	cmd:epstopdf
+	cmd:kpsewhich
+	cmd:latex
+	cmd:makeindex
+	cmd:pdflatex	
+	setuptools_python3
+	texlive_texmf
+	"
+BUILD_PREREQUIRES="
+	cmd:python3
+	cmd:sed
+	"
+
+PATCH()
+{
+	sed -i "s|share/dblatex|$relativeDataDir/dblatex|
+		s|share/doc/dblatex|$relativeDocDir|
+		s|share/man|$relativeManDir|" setup.py
+}
+
+BUILD()
+{
+	rm -rf build
+	python3 setup.py build
+}
+
+INSTALL()
+{
+	python3 setup.py install --prefix="$prefix" --root=/
+}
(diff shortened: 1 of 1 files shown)


app-text/dblatex/dblatex-0.3.12.recipe
```diff
@@ -0,0 +1,64 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	cmd:convert
+	cmd:epstopdf
+	cmd:kpsewhich
+	cmd:latex
+	cmd:makeindex
+	cmd:pdflatex	
+	setuptools_python3
+	texlive_texmf
```
Begasus, line 41: texlive_texmf is gcc8 so architecture can't be "any"
```diff
- 	texlive_texmf
+ 	texlive${secondaryArchSuffix}_texmf
```

jmairboeck: texlive_texmf technically is "any", but its name is different on x86. Maybe we should add an additional PROVIDES to texlive${secondaryArchSuffix}_texmf without the "secondaryArchSuffix" part if it is not empty?

To be honest, this could even be viewed as a "bug" in haikuporter. Why does the package name of the subpackage have to contain "_x86" if it is architecture "any"?

Or could something like `SECONDARY_ACHITECTURES_texmf=""` be set in the texlive recipe (explicitly clearing the secondary architectures for the subpackages)?

Begasus: Need to search if there is some example here on haikuports that uses such an extention, at this moment I don't know if it's possible

jmairboeck: Maybe libreoffice? I didn't look specifically at it, but I think it contains "any" subpackages.

Begasus: Provides there is also secondary:
https://github.com/haikuports/haikuports/blob/14c2cab5428145b93232cb69683a67bbe68a9f06/app-office/libreoffice/libreoffice-6.4.7.2.recipe#L146

Begasus: So far I haven't found anything that could work here, I did something similar for the fpc package, but that included a separate package, so it would mean that in this case you should split up texlive and texlive_texmf into seperate packages ... one benefit maybe, the base package including the commands for latex could be included in haikuports as active (I guess there are more packages around the size of +180MiB already :)

jmairboeck: I will try splitting the texlive recipe in two, that is probably a good idea anyway. But please be patient, this will take a while ;)

jmairboeck:
I have now updated the recipe to use the split texlive package names from #6229. If that could get merged, we could activate and merge this one too probably.

Maybe I should start writing the actual Lilypond recipe in the meantime ...

jmairboeck:
I just noticed that this doesn't work correctly when packaged. `dblatex --version` prints the following error message:

```
/packages/python3-3.7.12-1/.self/lib/python3.7/vendor-packages> dblatex --version
  File "/bin/dblatex", line 11, in <module>
    dblatex.main(base=package_base)
  File "/packages/python3-3.7.12-1/.self/lib/python3.7/vendor-packages/dbtexmf/dblatex/dblatex.py", line 43, in main
    command.main()
  File "/packages/python3-3.7.12-1/.self/lib/python3.7/vendor-packages/dbtexmf/core/dbtex.py", line 688, in main
    version = run.get_version()
  File "/packages/python3-3.7.12-1/.self/lib/python3.7/vendor-packages/dbtexmf/core/dbtex.py", line 204, in get_version
    f = open(os.path.join(self.topdir, "xsl", "version.xsl"), "rt", encoding="latin-1")
FileNotFoundError: [Errno -2147459069] No such file or directory: '/boot/system/lib/python3.7/vendor-packages/packages/dblatex-0.3.12-1/.self/data/dblatex/xsl/version.xsl'
```

It seems to use a relative path as base path that should be absolute instead.

jmairboeck:
the above is just an example, when running it from the home directory, it searches in `/boot/home/packages/dblatex-0.3.12-1/.self/data/dblatex/xsl/version.xsl`.

Only when running it from the root directory, it works correctly.

jmairboeck:
> the above is just an example, when running it from the home directory, it searches in `/boot/home/packages/dblatex-0.3.12-1/.self/data/dblatex/xsl/version.xsl`.
> 
> Only when running it from the root directory, it works correctly.

I have solved the problem by removing `--root=/` from the `setup.py install` call.

korli:
> > the above is just an example, when running it from the home directory, it searches in `/boot/home/packages/dblatex-0.3.12-1/.self/data/dblatex/xsl/version.xsl`.
> > Only when running it from the root directory, it works correctly.
> 
> I have solved the problem by removing `--root=/` from the `setup.py install` call.

The install location should come first in PYTHONPATH, see for instance: https://github.com/haikuports/haikuports/blob/master/dev-python/aiodns/aiodns-2.0.0.recipe#L50

jmairboeck:
What is the purpose of `--root=/` anyway?

The "problem" was the following: the `setup.py` script of dblatex creates a launcher script that gets installed in `$prefix/bin`. This uses `package_base` to find its files. The problematic function here is `_strip_root` (see [here](https://sourceforge.net/p/dblatex/dblatex/ci/default/tree/setup.py#l91), which removes the root dir if it is a prefix of the `package_path`, therefore stripping the initial '/' and resulting in an incorrect relative path.

Begasus:
Needs merging for texlive update first: https://github.com/haikuports/haikuports/pull/6818

PR #6094, opened 2021-07-24
openblas: bump with patch

wattoc:
Run full build successfully, also launches digiKam, etc. without crashing.

I've also re-enabled AVX since this should be supported now, original issue for disabling it was here: https://github.com/xianyi/OpenBLAS/issues/1916 - this was due to Haiku not supporting AVX at all at the time.

It should be detecting the processor and using AVX if the processor supports it at runtime (basically compiles a bunch of different versions for various x64 cpu types.)  If anyone is concerned about this it can always be taken out again with some performance hit, although if we can find someone with a Prescott x64 CPU, eg. a Pentium 4F and derivative to test, this would be helpful.

Patch should be upstreamed to openblas project so will hopefully not be needed in future.  I can attempt a pull for this unless someone else wants to volunteer.

### DIFF
--- a/sci-libs/openblas/openblas-0.3.17.recipe
+++ b/sci-libs/openblas/openblas-0.3.17.recipe
@@ -62,7 +62,8 @@ case "$effectiveTargetArchitecture" in
 	*) target=;;
 esac
 
-BUILD_CONF="NO_LAPACK=1 \
+BUILD_CONF="NO_STATIC=1 \
+		NO_LAPACK=1 \
 		NO_LAPACKE=1 \
 		NO_AFFINITY=1 \
 		NO_WARMUP=1 \
@@ -74,7 +75,8 @@ BUILD()
 {
 	make MAKE_NB_JOBS=${jobArgs#-j} \
 		$BUILD_CONF \
-		${target:+TARGET=$target}
+		${target:+TARGET=$target} \
+		libs shared
 }
 
 INSTALL()
--- a/sci-libs/openblas/openblas-0.3.17.recipe
+++ b/sci-libs/openblas/openblas-0.3.17.recipe
@@ -4,11 +4,12 @@ version."
 HOMEPAGE="http://www.openblas.net/"
 COPYRIGHT="2011-2021 The OpenBLAS Project"
 LICENSE="BSD (3-clause)"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="http://github.com/xianyi/OpenBLAS/archive/v$portVersion.tar.gz"
 CHECKSUM_SHA256="df2934fa33d04fd84d839ca698280df55c690c86a5a1133b3f7266fce1de279f"
 SOURCE_FILENAME="OpenBLAS-$portVersion.tar.gz"
 SOURCE_DIR="OpenBLAS-$portVersion"
+PATCHES="openblas-$portVersion.patchset"
 
 ARCHITECTURES="!x86_gcc2 x86 x86_64"
 SECONDARY_ARCHITECTURES="x86"
@@ -61,11 +62,10 @@ case "$effectiveTargetArchitecture" in
 	*) target=;;
 esac
 
-BUILD_CONF="NO_STATIC=1 NO_LAPACK=1 \
+BUILD_CONF="NO_LAPACK=1 \
 		NO_LAPACKE=1 \
 		NO_AFFINITY=1 \
 		NO_WARMUP=1 \
-		NO_AVX=1 \
 		NUM_THREADS=64 \
 		DYNAMIC_ARCH=1 \
 		USE_OPENMP=1"
@@ -74,8 +74,7 @@ BUILD()
 {
 	make MAKE_NB_JOBS=${jobArgs#-j} \
 		$BUILD_CONF \
-		${target:+TARGET=$target} \
-		libs shared
+		${target:+TARGET=$target}
 }
 
 INSTALL()
--- a/sci-libs/openblas/patches/openblas-0.3.17.patchset
+++ b/sci-libs/openblas/patches/openblas-0.3.17.patchset
@@ -1,0 +1,67 @@
+From d05e2ec0713ba49796faa04641644156e171e4c5 Mon Sep 17 00:00:00 2001
+From: Craig Watson <watsoncraigjohn@gmail.com>
+Date: Sat, 24 Jul 2021 13:36:09 +0000
+Subject: Include Haiku in processor count checks
+
+
+diff --git a/driver/others/memory.c b/driver/others/memory.c
+index 6e654cc..aac8341 100644
+--- a/driver/others/memory.c
++++ b/driver/others/memory.c
+@@ -428,7 +428,7 @@ extern int openblas_goto_num_threads_env();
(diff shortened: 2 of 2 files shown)


sci-libs/openblas/openblas-0.3.17.recipe
```diff
@@ -74,8 +74,7 @@ BUILD()
 {
 	make MAKE_NB_JOBS=${jobArgs#-j} \
 		$BUILD_CONF \
-		${target:+TARGET=$target} \
-		libs shared
```
korli, line 78: This should stay as is. Tests go in TEST()

sci-libs/openblas/openblas-0.3.17.recipe
```diff
@@ -61,11 +62,10 @@ case "$effectiveTargetArchitecture" in
 	*) target=;;
 esac
 
-BUILD_CONF="NO_STATIC=1 NO_LAPACK=1 \
```
korli, line 64: we don't want static libs. please revert.

sci-libs/openblas/openblas-0.3.17.recipe
```diff
@@ -61,11 +62,11 @@ case "$effectiveTargetArchitecture" in
...
+BUILD_CONF="NO_STATIC=1 \
+		NO_LAPACK=1 \
 		NO_LAPACKE=1 \
 		NO_AFFINITY=1 \
 		NO_WARMUP=1 \
-		NO_AVX=1 \
```
korli, line 68: AVX is only on x86_64 available. NO_AVX=1 should stay for x86.

wattoc: This should be detected during Build on x86.  https://github.com/xianyi/OpenBLAS/blob/1a8b6134c24f66c649ca342c74d551055b98cbc6/Makefile.system  - 32-bit binaries should have these flags added by the build system

Have you tried building it on x86?

korli: This won't be detected during build.

wattoc: I don't understand why.  Are you saying the 32 bit build system will not be detected and then somehow try to build AVX features?  The build process should use the provided GCC to detect whether it can generate 32-bit vs 64-bit.  It sets the BINARY env variable which then, if it's = 32 will provide the NO_AVX, etc. itself.

NO_AVX is implicit for 32 bit binaries.  It should not need to be provided separately... If so it should be NO_AVX, NO_AVX2, NO_AVX512, etc.

korli: I mean the build won't detect that 32-bit Haiku has no AVX support. If OpenBLAS disables AVX for every 32-bit system (which would be worrying), then OK.

wattoc: 32 bit OpenBLAS has no AVX support, it's not possible anyway:

```
ifeq ($(BINARY), 32)
GETARCH_FLAGS	+= -DNO_AVX -DNO_AVX2 -DNO_AVX512
NO_AVX512 = 1
endif
...
ifeq ($(ARCH), x86)
CCOMMON_OPT	+= -DNO_AVX
endif
```

korli: So it's not detected, but hardcoded disabled. Should be OK.

brada4:
Please check on real AVX-2 and AVX-512 CPUs, they have bigger xsave file at least.

wattoc:
I only have access to Zen2 CPUs here (CPUs which support some sort of AVX anyway) - I've been testing with those.  Not sure if there's anyone with Skylake X/Cannon Lake or newer to test AVX512 about.

korli:
@wattoc AVX-512 can be disabled, untested.

brada4:
`NO_AVX512=1` then

wattoc:
@korli this seems a bit extreme when you recently pushed out an update for this that was not working at all.  I'm not sure why stringent testing requirements seem to only apply when I'm pushing packages here.

I would suggest leaving it enabled and wait for someone to report an issue.

korli:
@wattoc fine, not that I care much. I'm the one who implemented AVX support on Haiku x86_64, the reason why I argue.

wattoc:
If you want it disabled it is one line, it's easy enough to do - I would prefer evidence of it not working vs. speculation that it might not work.  If I'd have done the AXV support I'd rather people have a chance to attempt to use it so I can be sure it works.  

This package isn't really for my personal use, so is not based on some personal investment I'm just trying to help fix some bugs here - you're asking a lot of testing potentially (there are multiple kernels for various processor types) from someone doing this on a hobby basis who is naively trying to be helpful.  It was literally broken for many people since 0.3.9 which went completely ignored for some time - so clearly has not been tested that well anyway.  I'm not sure how I'm making it worse enabling an option that as far as I'm aware should work (I'm sure you can confirm that given you implemented AVX support) and if it doesn't is probably a great way to discover issues with the OS itself.

brada4:
@wattoc Previous time there was no AVX in haiku kernel, I just suggested to check incrementally the new (register space) additions so that casual users do not return with "openblas crashes my gimp" sort of thing.

pulkomandy:
We have tested AVX support in Haiku in other places (for example the AVIF and Dav1d decoders use it if available). It's working fine. I don't know about AVX512 because no one with compatible hardware has tested Haiku on that, as far as I know.

The only thing to make sure is that openblas detects the instruction set before using it, and switches to another codepath if it isn't. But I understand that this is the case, so there should be no problem here.

wattoc:
@brada4 I appreciate what you and korli are saying - as far as I know korli added AVX support about a year ago, so surely that's enough time to have spotted some issues - we have some other things using it already as far as I'm aware.  The main issue we have is lack of users with such hardware, so we need someone with a later Skylake X or similar to confirm AVX512 at some point, but that issue is not limited to OpenBLAS.  If an issue occurs it's an easy place to toggle it off (we can ask the reporter to test a non-AVX512 target through the command line anyway) and then look at the issues on an OS level if need be.  Since I now know a bit more about OpenBLAS I'm happy to try to assist if that happens.

I don't think it's a good approach to put artificial limits on things because we speculate there might be problems (although as far as anyone knows there shouldn't be) or we simply haven't tested that exact thing yet.  It'd be like putting a 16GB limit on Haiku because nobody tested it with more than that any maybe something goes wrong.

Anyway we don't have GIMP yet :)  The use case I know of here is something like digiKam (because that's how I became aware of the issue with someone raising a bug on it) using OpenCV and therefore OpenBLAS, which has been tested by myself as best I can with the resources I have.  There are probably other things that potentially use OpenBLAS I haven't tested - but at some point you need to be reasonable.  I committed a trivial bug fix that means we now count processors and don't create MAX_THREADS and enabled some options that SHOULD work.

brada4:
@wattoc i back @korli , current XSAVE code in haiku supports only AVX2 (and SSE, and x87) , and no AVX-512

wattoc:
@brada4 I'll bring in a new pull request with NO_AVX512 defined in that case.

PR #6096, opened 2021-07-24
libjxl, new recipe

Begasus:


Begasus:
Just remembered why I didn't push it, checksums seems to fail on new downloads .. don't merge until it's solved

wattoc:
Doesn't look like those googlesource links will work with a checksum - no matter what commit I grab the files have today's date.  Unless I'm missing something.

wattoc:
Probably want to put those binaries into a tools package?

Begasus:
> Probably want to put those binaries into a tools package?

That's an easy fix :) will do it later today

Begasus:
Using "git+https..." can work, but I don't think it's encouraged here

Begasus:
tools package included now :)

media-libs/libjxl/libjxl-0.3.7.recipe
```diff
@@ -0,0 +1,142 @@
...
+	devel:libOpenEXRUtil_3_0$secondaryArchSuffix
+	devel:libjpeg$secondaryArchSuffix
+	devel:libpng16$secondaryArchSuffix
+	devel:libwebp$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:git
```
korli, line 103: cmd:git should be removed.

CodeforEvolution:
I don't think GitLab is the home of libjxl anymore, and is now based here: https://github.com/libjxl/libjxl

wattoc:
If we can sort out the recipe for this I've got a Translator pretty much ready to go now.

Begasus:
> I don't think GitLab is the home of libjxl anymore, and is now based here: https://github.com/libjxl/libjxl

That isn't an issue on the downloads, that checksup is fine (and on par with the one at github) :)

media-libs/libjxl/libjxl-0.3.7.recipe
Begasus: If I get around to it I'll have another look at this this week(end) :)

Begasus: Re-working the recipe so it will download the submodules, for this `git` is needed, haikuporter doesn't play nice so I have to delete the `.git` directory so it fetches the `submodules` ...

Begasus: There seems to be an issue (maybe related to this) at haikuporter at: https://github.com/haikuports/haikuporter/issues/62

korli: I see Homebrew manages to build without git.
https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/jpeg-xl.rb

Begasus: Others also do it without git, but they don't use the checksum on those or use "SKIP" to by-pass it, the offending checksum fail is especially for skcms (googlesource), I could leave that checksum out and go back to the previous version, that would work without git.

pulkomandy:
I think the best thing to do is to create an archive (zip or tar or whatever) of the sources and host that somewhere, and then use that for building.

We really want to be sure that we always use the same sources to build something, and git and submodules don't make this really convenient to do, that's why we avoid using them. There is no way to do a checksum on the whole thing.

media-libs/libjxl/libjxl-0.3.7.recipe
korli: I was talking about submodules

media-libs/libjxl/libjxl-0.5.recipe_copy
```diff
@@ -0,0 +1,140 @@
...
+REVISION="1"
+SOURCE_URI="https://gitlab.com/wg1/jpeg-xl/-/archive/v$portVersion/jpeg-xl-v$portVersion.tar.bz2"
+CHECKSUM_SHA256="43ae213b9ff28f672beb4f50dbee0834be2afe0015a62bf525d35ee2e7e89d6c"
+srcGitRev_2="48e5364ef48ec2408f44c727657ac1b6703185f8"
+SOURCE_URI_2="https://github.com/lvandeve/lodepng/archive/$srcGitRev_2.tar.gz"
+CHECKSUM_SHA256_2="c47c48c77a205f1af484b7b5a847290af65de3ea6f15817aa27c5ec7cc5208fd"
+SOURCE_FILENAME_2="lodepng-$srcGitRev_2.tar.gz"
+srcGitRev_3="64374756e03700d649f897dbd98c95e78c30c7da"
+SOURCE_URI_3="https://skia.googlesource.com/skcms/+archive/$srcGitRev_3.tar.gz"
+#checksum fails on every time it's checked
```
Begasus, line 21: This is the only one failing checksum, have been thinking about moving/copying it over to my account on github, but I don't know if that's legal to do?

korli: or use an existing copy: https://github.com/workerVA/skcms/

Begasus: That could work: https://github.com/workerVA/skcms/commit/64374756e03700d649f897dbd98c95e78c30c7da
Will check on it :) Thanks for the pointer :+1:

Begasus:
Recipe updated with the fixed URL @korli mentioned, added some copyright and license lines to it (some come from the submodules, libjxl itself should be fine with BSD and Zlib afaict), if these are OK this should be good to merge

PR #6101, opened 2021-07-27
bash: bump

kenmays:
Tested on Haiku R1B3.


app-shells/bash/bash-5.1.8.recipe
```diff
@@ -63,7 +63,7 @@ PATCH()
 BUILD()
 {
 	CPPFLAGS="-DUSE_MKTEMP=1 -DUSE_MKSTEMP=1" runConfigure ./configure \
-		--without-bash-malloc --with-installed-readline
+		--without-bash-malloc --disable-rpath --enable-disabled-builtins --with-installed-readline
```
extrowerk, line 66: What does the `--enable-disabled-builtins` do?

pulkomandy: > --enable-disabled-builtins
> Allow builtin commands to be invoked via ‘builtin xxx’ even after xxx has been disabled using ‘enable -n xxx’. See Bash Builtins, for details of the builtin and enable builtin commands.

Sounds quite useful to me, why remove it?

korli:
there is also a pending patch for readline: https://mirror.easyname.at/gnu/readline/readline-8.1-patches/readline81-001

kenmays:
@korli: https://github.com/haikuports/haikuports/pull/6106

PR #6105, opened 2021-07-28
Openblas bump, remove AVX512

wattoc:
This will disable AVX512 kernels from building on x86_64 as it appears we do not support AVX512 in Haiku yet.


korli:
Thanks!

PR #6106, opened 2021-07-29
readline: revision bump

kenmays:
Tested on Haiku R1B3.
Fixed recipe and added readline patches.


sys-libs/readline/readline-8.1.recipe
```diff
@@ -60,30 +60,27 @@ defineDebugInfoPackage readline$secondaryArchSuffix \
 
 BUILD()
 {
-	libtoolize --force --copy --install
+	libtoolize -fci
 	aclocal
 	autoconf
 	runConfigure ./configure \
-		--with-curses
+		--with-curses --enable-multibyte --disable-static
```
korli, line 67: why --enable-multibyte?

kenmays: Already a default, just enforces usage.

PR #6108, opened 2021-07-30
gcc: added 11.2.0 recipe and patchset

kenmays:
Tested on Haiku R1B3


waddlesplash:
I'm sorry, @kenmays, but I don't think we trust you to properly update patchsets at this point.

kenmays:
@waddlesplash - do as U will.

PR #6110, opened 2021-07-31
lagrange: bump to 1.6.1

sikmir:
[Changelog](https://github.com/skyjake/lagrange/releases)


net-misc/lagrange/lagrange-1.6.1.recipe
```diff
@@ -53,7 +57,9 @@ BUILD()
 	cd build
 	cmake .. \
 		-DCMAKE_INSTALL_PREFIX=$appsDir/Lagrange \
-		-DENABLE_MPG123=ON
+		-DENABLE_MPG123=ON \
+		-DENABLE_HARFBUZZ_MINIMAL=OFF \
+		-DENABLE_FRIBIDI_BUILD=OFF
```
Begasus, line 62: Any specific reason (not tested here) why they are disabled if you added them to the requirements (I think leaving them out does the same as cmake wouldn't pick them up anyway)?

Begasus:
Does it fix some of the open issues atm?

net-misc/lagrange/lagrange-1.6.1.recipe
sikmir: These flags just say to use the system libs (fribidi and harfbuzz) instead of building them from sources.

sikmir:
> Does it fix some of the open issues atm?

No, the problem with resizing main window still exists. But everything else seems to work for me.

Begasus:
Thanks for commenting on the issues :+1:

PR #6113, opened 2021-08-04
ripgrep: bump to 13.0.0

Crestwave:
Note: Untested on x86, though it should be fine as it's just a version bump


extrowerk (changes requested):
It seems you forgot about the checksums.

Crestwave:
> It seems you forgot about the checksums.

I forgot to push the changes, whoops. Fixed it

Crestwave:
Tested on x86 now, all good

Issue #1767, opened 2017-11-01
Texlive: split up to avoid a 800MB main package


korli:
Lyx needs a medium TexLive installation for instance.

Texlive common package should have "any" as an architecture, and for each architecture a binaries package.

PR #6115, opened 2021-08-07
Texlive: split texmf part into its own recipe

jmairboeck:
This allows the texmf part to be a real "any" package without the secondaryArchSuffix as part of the package name.

Other changes:
 * make a debuginfo package for the main package (instead of using `make install-strip`)
 * fix tests for the main package
 * add missing `fixPkgConfig`
 * the /packages-path hack has been solved differently in the texmf package, not using kpsewhich anymore at all during install-tl. Because the paths are different in the second recipe, this doesn't work any more.
 * move `fmtutil-sys --all` (and `updmap-sys`) into a post-install script for the same reason
 * remove the copying of the `tlpkg/README` file which is missing somehow


Begasus:
Checking builds atm and will report back later :)

Begasus:
Checked a build for a sample.tex file in TeXstudio, didn't work (for the base install) needs checking on what's missing here :)

Begasus:
Looks like the sample I used had some graphics in it, a plain tex file worked fine with the base package (without installing the texlive_texmf package)

jmairboeck:
I am surprised that even anything works without the texlive_texmf package ...

When trying to build LilyPond without it, it seems to hang when invoking mf (MetaFont), but I still have to investigate it what happens exactly.

jmairboeck:
> When trying to build LilyPond without it, it seems to hang when invoking mf (MetaFont), but I still have to investigate it what happens exactly.

It says that it can't find the input file `mfplain` and asks interactively for another one instead. So it seems Llilypond requires at least that to build.

jmairboeck:
I have now further split the texlive_texmf package into individual packages per "collection" (runtime, source and documentation files each), plus a "full" metapackage that depends on all collection packages.

app-text/texlive_texmf/additional-files/createSubpackageInfos.pl
```diff
@@ -0,0 +1,95 @@
...
+		my $depPkg = $tlpdb->get_package($depName);
+		push @runfiles, $depPkg->runfiles();
+		push @docfiles, $depPkg->docfiles();
+		push @srcfiles, $depPkg->srcfiles();
+	}
+	if (@runfiles) {
+		print "SUMMARY_$collection=\"TeXLive Collection: $shortdesc\"\n";
+		print "DESCRIPTION_$collection=\"$longdesc\"\n" if $longdesc;
+		print "PROVIDES_$collection=\"\n";
+		print "\ttexlive_texmf_$collection = \$portVersion\n";
```
jmairboeck, line 51: Should we add PROVIDES for the individual TeX packages in the collection here as well, so that other packages can later depend on them instead of just the whole collection.

How should these be called? Should (or can) we introduce a new prefix (e.g. "tex:")?

Note: I would only do this for the runtime files. This is where it makes the most sense.

The generated file would become ca. 4000 lines longer. I have already implemented it locally but not tested yet.

jmairboeck:
@korli has suggested [in the forum](https://discuss.haiku-os.org/t/texlive-on-x86-64/10548/33) to only make one subpackage for runtime, docs and source files per collection, and not 3 separate ones. What do you think?

jmairboeck:
The biggest collection is "fontsextra" at 724.4 MiB (runtime files), the docs for this are 114.67 MiB, the sources 15.6 MiB. Putting these together would amount to 854.67 MiB.

app-text/texlive_texmf/additional-files/createSubpackageInfos.pl
```diff
@@ -0,0 +1,95 @@
+#!/bin/perl
+
```
jmairboeck, line 2: Should I add a license header here?

As to which one, I suppose it would have to be GPL v2 because it depends on such code. But I am open to MIT too, if that is possible.

app-text/texlive_texmf/additional-files/getPackageEntries.pl
```diff
@@ -0,0 +1,34 @@
+#!/bin/perl
+
```
jmairboeck, line 2: License header here too?

jmairboeck:
I am thinking about swapping the names around, calling the texlive recipe "texlive_bin" and texlive_texmf just "texmf". The collection packages could be called "texlive_collection_$foo" too. Which would also solve the problem of what the PROVIDES for other packages should be called if we decide to do that. They could just use "texlive_$packagename" and no new prefix would be needed.

app-text/texlive_texmf/texlive_texmf-2021.recipe
```diff
@@ -0,0 +1,264 @@
...
+SOURCE_DIR_3="texlive-$fullVersion-extra"
+ADDITIONAL_FILES="
+	createSubpackageInfos.pl
+	getPackageEntries.pl
+	subpackageInfos.recipe
+	texlive.profile
+	texlive_postinstall.sh
+	"
+
+POST_INSTALL_SCRIPTS="$relativePostInstallDir/texlive_postinstall.sh"
```
jmairboeck, line 28: this doesn't work as such now. When only texlive_texmf is installed (no collection subpackages), it fails because something is missing apparently. I suppose it should at least be moved to the _basic subpackage. Ideally it should be called whenever one of the runtime subpackages is installed, but only once if multiple are installed at the same time.

However, this needs a bit more advanced logic because there no post-transaction scripts in Haiku (like %posttrans in RPM).

waddlesplash:
This should fix #1767 once merged.

korli:
> The biggest collection is "fontsextra" at 724.4 MiB (runtime files), the docs for this are 114.67 MiB, the sources 15.6 MiB. Putting these together would amount to 854.67 MiB.

Seems correct seeing the size at https://archlinux.org/packages/extra/any/texlive-fontsextra/
You can make one exception in your generating script, if that makes sense to you.

korli:
> I am thinking about swapping the names around, calling the texlive recipe "texlive_bin" 

The _bin suffix (at Haikuports) generally means the package is repacked from a binary archive and not built from sources.

korli:
> Adding a special exception feels like a "hack" to me. Right now, the script doesn't even assume wh

This recipe is a hack anyway, the generating script will need to be maintained. Do other repositories split Texlive the same? I see Arch is doing differently, why?

jmairboeck:
Debian seems to separate the documentation too: https://packages.debian.org/search?keywords=texlive&searchon=names&suite=stable&section=all

jmairboeck:
I think following Debian would not be the worst choice. It seems that their main package maintainer is rather involved upstream too (if I am not mistaken). At least Debian is the only distro that links to an internal page on the [official website](https://tug.org/texlive/distro.html).

But they too have a whole repository with quite a few scripts and other complex infrastructure to build the packages. https://github.com/debian-tex/texlive-nonbin
I don't think we need or should use something that complex.

Also, I think you were mistaken about Arch: https://wiki.archlinux.org/title/TeX_Live#Package_documentation

Note that I am not completely against using single packages for runtime files, docs and sources, I just think that most users probably won't need the latter. If you or others still think that it is better with "consolidated" packages, I can also change that. I'm just not really convinced, and the current implementation works too.

korli:
If packaging docs and sources can be skipped, why not doing the same?

jmairboeck:
> > I am thinking about swapping the names around, calling the texlive recipe "texlive_bin"
> 
> The _bin suffix (at Haikuports) generally means the package is repacked from a binary archive and not built from sources.

Debian calls the package "texlive-binaries". Would that be a suitable name?

Begasus:
Maybe it's time to create a seperate texlive_texmf PR for this, so far the base packages provide the needed binaries (afaikt) for TexStudio?

jmairboeck:
I have removed the texlive_texmf splitting into a new branch (PR will come later) so that this can be merged now and at least the main package can be provided in HaikuDepot soon (although it will mostly not work without the other package).

The former main package has been renamed to texlive_core to reflect Gentoo's naming. The texmf package is now just called "texlive".

jmairboeck:
I have rebased the branch to resolve the conflict because of #6189.

I think this can be merged now (I have not tested the renamed package though).

jmairboeck:
> This should fix #1767 once merged.

This will not be fixed (any more) with this PR but with a follow-up which contains the actual split into multiple subpackages.

Begasus:
@korli @waddlesplash do you think we can merge in it's current state?

korli:
I'm afraid, I lost the thread on this PR, so from my point of view, no objection.

Begasus:
Merging (after some talks here and on IRC so @jmairboeck can go along with the work on splitting up the correct packages.

PR #6122, opened 2021-08-08
GMSH: new recipe

extrowerk:


Begasus:
LGTM, can't check a build atm, merging, check buildmaster to see if things are fine on both arch's :)

PR #6124, opened 2021-08-11
guile: move the guile interpreter tools to a 'tools' subpackage like guile1

jmairboeck:
further changes (for both guile1 and guile2):
 * move man page also to tools as it concerns the interpreter
 * define debuginfo for guile and the other than main libraries
   - note: defining multple debuginfo packages is not possible using the predefined shell scriptlet
 * add/fix compat versions for libraries


jmairboeck:
Reason for this: the interpreter in `bin/guile` is the same file name in both versions. By splitting it into a tools subpackage also for guile2, the library part of both versions can be installed at the same time.

PR #6125, opened 2021-08-12
guile: fix build on 32 bit x86

jmairboeck:


Begasus:
Missed bumping the revision for a rebuild on the buildmaster :)

PR #6128, opened 2021-08-14
guile1: add version-less symlinks for the libraries to facilitate dynamically loading

### LINKED ISSUE #5993: guile1: error invoking scheme script: "in procedure dynamic-link"
Now that texlive is mostly "done" (see #5963), I thought I'll try going further towards my goal of porting [LilyPond](http://lilypond.org).

However, this currently fails pretty early in the build process because I suppose, the guile1 package is somewhat broken on Haiku. When trying to invoke [lilypond-invoke-editor](http://git.savannah.gnu.org/gitweb/?p=lilypond.git;a=blob;f=scripts/lilypond-invoke-editor.scm;h=507345b1f89bcd2a26c42a02b757585bb6900ab7;hb=HEAD) during the build process, it fails with this error message:

```
ERROR: In procedure dynamic-link:
ERROR: file: "libguile-srfi-srfi-1-v-3", message: "file not found"
```

The package `guile1_devel` is installed, and it contains the mentioned library in `develop/lib` (and the runtime counterpart is in `lib` from the `guile1` package). I suspect that guile isn't prepared to look for development libs in the `develop` directory, but that is just a guess from my side.

`dynamic-link` is apparently a built-in function provided by guile, so I suppose the `guile1` package needs to be fixed.

Note: I could try building LilyPond with guile2 instead, but this is [not officially supported and not recommended](http://lilypond.org/doc/v2.22/Documentation/contributor/requirements-for-running-lilypond) yet.

jmairboeck:
Fixes #5993

Also, fix the version number of libguile-readline in provides.

### DIFF
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -147,12 +147,9 @@ INSTALL()
 
 	fixPkgconfig
 
-	# add version-less symlinks of the libraries (which are needed because they are loaded dynamically)
-	ln -s libguile.so.$libguileVersion "$libDir"/libguile.so
+	# add version-less symlinks of the libraries which are loaded dynamically
 	ln -s libguilereadline-v-17.so.17.0.3 "$libDir"/libguilereadline-v-17.so
 	ln -s libguile-srfi-srfi-1-v-3.so.3.0.2 "$libDir"/libguile-srfi-srfi-1-v-3.so
-	ln -s libguile-srfi-srfi-13-14-v-3.so.3.0.1 "$libDir"/libguile-srfi-srfi-13-14-v-3.so
-	ln -s libguile-srfi-srfi-4-v-3.so.3.0.1 "$libDir"/libguile-srfi-srfi-4-v-3.so
 	ln -s libguile-srfi-srfi-60-v-2.so.2.0.2 "$libDir"/libguile-srfi-srfi-60-v-2.so
 
 	packageEntries tools \
--- a/dev-scheme/guile/guile1-1.8.8.recipe
+++ b/dev-scheme/guile/guile1-1.8.8.recipe
@@ -7,7 +7,7 @@ the same) and users to use them to have an application fit their needs."
 HOMEPAGE="https://www.gnu.org/software/guile/"
 COPYRIGHT="1993-2018 Aubrey Jaffer, George Carrette, et al."
 LICENSE="GNU LGPL v3"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://ftpmirror.gnu.org/guile/guile-$portVersion.tar.gz
 	https://ftp.gnu.org/gnu/guile/guile-$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


jmairboeck:
I have tested this on both 64 bit and 32 bit, using the lilypond-invoke-editor script (not yet with a full Lilypond against this though).

jmairboeck:
It also builds against libffi.so.8, so if this gets merged, this would also fix guile1 again with the newer libffi, without needing a separate revision bump.

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -147,6 +147,14 @@ INSTALL()
 
 	fixPkgconfig
 
+	# add version-less symlinks of the libraries (which are needed because they are loaded dynamically)
+	ln -s libguile.so.$libguileVersion "$libDir"/libguile.so
```
jmairboeck, line 151: this could be left out probably

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -147,6 +147,14 @@ INSTALL()
...
+	ln -s libguilereadline-v-17.so.17.0.3 "$libDir"/libguilereadline-v-17.so
+	ln -s libguile-srfi-srfi-1-v-3.so.3.0.2 "$libDir"/libguile-srfi-srfi-1-v-3.so
+	ln -s libguile-srfi-srfi-13-14-v-3.so.3.0.1 "$libDir"/libguile-srfi-srfi-13-14-v-3.so
```
jmairboeck, line 154: and this also

dev-scheme/guile/guile1-1.8.8.recipe
```diff
@@ -147,6 +147,14 @@ INSTALL()
...
+	ln -s libguile-srfi-srfi-4-v-3.so.3.0.1 "$libDir"/libguile-srfi-srfi-4-v-3.so
```
jmairboeck, line 155: and this also

Begasus:
Anything needed here to get it merged?

jmairboeck:
Should I remove the technically unneeded ones? OpenSUSE includes all of them in the main package, however. On the other hand, being Linux, it doesn't have a separate $developLibDir, so the devel-package would contain them anyway.

Begasus:
Are the links needed at all seeing they are provided with the devel package?

jmairboeck:
In the devel package, they are in the wrong directory ($developLibDir instead of $libDir), where they are not found when loaded dynamically because that is not in LIBRARY_PATH by default. Also, a normal non-development application should not depend on a devel-package for normal functionality. That is in essence what the whole issue #5993 is about.

Begasus:
OK, seems reasonable enough to me, if you have some changes left, just push them and we can merge (if no one objects) :)

jmairboeck:
> (if no one objects)

@pulkomandy kind of objected [here](https://github.com/haikuports/haikuports/issues/5993#issuecomment-898966104) if I understand him correctly, but I don't see another feasible solution, honestly.

jmairboeck:
I have removed the unneeded ones, can this be merged now?

PR #6133, opened 2021-08-15
purple_discord: add recipe

JadedCtrl:
Had a go at porting a libpurple plugin, probably got a good bit wrong here.


korli:
Build for a dependency failed on x86: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/1678/builds/2908.log

PR #6135, opened 2021-08-21
mercurial: update to 4.9.1

congocongocongo:


korli:
Remove old recipe. Please explain what was tested.

korli:
What about the 5.x versions and switching to python 3?

congocongocongo:
I tested with mozilla-unified, clone/pull/update works fine.  And it did not pass all unit tests like 4.8.1, mostly same tests.  Looks like something is broken in py2 datetime.datetime or around datetime.

congocongocongo:
@korli 5.8 version crashes python3.7 on unit tests and consumes 100% cpu on mozilla-unified. Its better to stay on 4.9.1

PR #6146, opened 2021-08-23
glib: bump to 2.68.4

linkmauve:
On an unrelated note, is it useful to keep all of the old versions here?  glib 2.x is fully backwards compatible from 2.0 (released in 2002) up until latest master, so maybe we could drop those?  I’m also not sure whether glib 1.x is still used anywhere, it has been removed from all Linux distributions I checked and isn’t maintained since 2002.

### DIFF
--- a/dev-libs/glib/glib2-2.68.4.recipe
+++ b/dev-libs/glib/glib2-2.68.4.recipe
@@ -128,6 +128,7 @@ BUILD()
 {
 	CFLAGS="-D_BSD_SOURCE" \
 		LDFLAGS="-lbsd -lgnu -lnetwork" meson build \
+		-D glib_debug=disabled --buildtype=debugoptimized \
 		-Diconv=external --prefix=$prefix --includedir=$includeDir \
 		--libdir=$libDir --datadir=$dataDir --bindir=$binDir \
 		--localedir=$dataDir/locale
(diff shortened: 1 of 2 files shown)


pulkomandy:
glib 1.x seems to be used by lilypond if I follow things correctly (@jmairboeck can you confirm?)

Recipes that provide the same "libVersionCompat" as the current one can be deleted (ABI is compatible, so the new package replaces the old one without problems). If there are different library versions and sonames, we need to make sure no packages are built against the older library, and there is currently no easy way to check that (especially as it could be packages in other repositories outside of haikuports...)

jmairboeck:
> glib 1.x seems to be used by lilypond if I follow things correctly (@jmairboeck can you confirm?)
No, Lilypond doesn't require glib at all. I think you confused it with guile. It is recommended to still use guile 1 (the scheme interpreter) because of performance issues with newer versions. Lilypond is written in C++, not C ...

pulkomandy:
> No, Lilypond doesn't require glib at all. I think you confused it with guile (the scheme interpreter)

Yes, sorry.

Well we should check other recipes to make sure nothing needs it, then

korli:
Well, building for x86 isn't each time tested, so it can be easier to not remove the old recipe. Especially when there is low interest.

Begasus:
> Well, building for x86 isn't each time tested, so it can be easier to not remove the old recipe. Especially when there is low interest.

I've tested builds for x86_gcc2 in the past, one of the glib dependencies is gobject-introspection directly related to the glib version, as it isn't used atm it's no deal-breaker though, I can check later to bump there if needed

waddlesplash:
I see Arch's PKGBUILD sets: `    -D glib_debug=disabled`. Do we need to do that here?

Also, it looks like Meson generates debug builds by default. We probably want `debugoptimized`, right? We probably need to specify that, too?

PR #6149, opened 2021-08-24
radare2: update from 4.5.1 to 5.4.0

trufae:


Begasus:
Could you remove the previous recipe in this PR? That way we can see the changes made in this one

korli:
Does it change the SPNAME?

Begasus:
> mv: cannot stat '/packages/radare2-5.4.0-1/.self/lib/libr_shlr.*': No such file or directory

Is you can see in the install part from the log there is no libr_shlr.* file, hence the build failed

PR #6163, opened 2021-08-30
new recipe: Postal

wattoc:


games-action/postal/patches/postal-1.0.0.patchset
```diff
@@ -0,0 +1,401 @@
...
+diff --git a/CompileOptions.h b/CompileOptions.h
+index 57b1799..e5dae0f 100644
+--- a/CompileOptions.h
++++ b/CompileOptions.h
+@@ -360,7 +360,7 @@
+ 				#endif
+ 			#endif
+ 		#else
+-			#define PREFS_FILE	"POSTAL.INI"
++			#define PREFS_FILE	"/boot/system/settings/POSTAL.INI"
```
korli, line 16: Preferences are usually in /boot/home/config/settings

games-action/postal/postal-1.0.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+Please note you will need the original game files to play this game.
+
+To play: copy the Linux data folders (res and title) to: 
+/boot/home/config/non-packaged/data/PostalPlus
+"
+HOMEPAGE="https://runningwithscissors.com/games/postal/"
+COPYRIGHT="1997-2003 Running With Scissors"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/classiccoding/postal-1-open-source/archive/refs/heads/master.zip"
```
korli, line 13: This should use a tag or a commit.

games-action/postal/postal-1.0.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+
+BUILD()
+{
+	make HAIKU=1 $jobArgs
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/Postal
+	cp bin/postal1 $appsDir/Postal/postal
```
korli, line 56: If you don't intend to install more than a binary, it can be located in $appsDir.

wattoc:
Comments above noted, there was a lot of ongoing discussion about this in the forums so I capitulated and threw a recipe together rather quickly.  Thanks for the feedback @korli will sort it as soon as I can.

games-action/postal/patches/postal-1.0.0.patchset
wattoc: This seems to be done in other packages, including games, but I have changed it anyway

games-action/postal/postal-1.0.0.recipe
```diff
@@ -0,0 +1,58 @@
...
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="1b5d7b9ffe0af4b20ca6825a7471a810ab6a6ef9"
+SOURCE_URI="https://github.com/classiccoding/postal-1-open-source/archive/$srcGitRev.zip"
+CHECKSUM_SHA256="adbfe09eeaf67527703d68c43989ddabecdcfca5bfcdc57857f36b282e90e703"
+SOURCE_DIR="postal-1-open-source-$srcGitRev"
+PATCHES="postal-$portVersion.patchset"
+ADDITIONAL_FILES="postal.rdef"
+
+ARCHITECTURES="!x86_gcc2 ?x86 x86_64"
```
Begasus, line 20: Maybe add support for a 32bit build?
```diff
+SECONDARY_ARCHITECTURES="x86"
```

wattoc: It'll probably work but not had much time to test.

Begasus: Build is OK :) 
Is it possible to disable fullscreen startup? (don't have game files to check but it launches fine without errors here on 32bit)

wattoc: I believe you can pass in -windowed on the command line:

postal1 -windowed

Begasus: Works fine in Terminal (not through Deskbar/Tracker (as expected)), now I see the message saying it's missing the game files ;)

tbn97:
Awesome! How do I build the recipe from here? I cannot find it in https://github.com/haikuports/haikuports/tree/master/games-action (still finding my way around this side of things)

wattoc:
I'll do a little testing here on 32-bit then update the recipe.  Assuming no complaints someone will merge it soon enough.

Otherwise you'll need to checkout my branch from github and build it locally.  There are some good wiki articles on getting haikuporter up and running: https://github.com/haikuports/haikuports/wiki/A-Gentle-Introduction-to-HaikuPorter-%231

games-action/postal/patches/postal-1.0.0.patchset
```diff
@@ -0,0 +1,446 @@
...
+ ////////////////////////////////////////////////////////////////////////////////
+@@ -765,6 +771,44 @@ int main(int argc, char **argv)
+         return 1;
+     #endif
+ 
++#ifdef __HAIKU__
++		//pref file should be copied from data if missing
++		//will then be used from user non-packaged for all writes
++		BPath systemDataPath, userDataPath;
++		if(find_directory(B_USER_NONPACKAGED_DATA_DIRECTORY, &userDataPath) == B_OK)
```
korli, line 86: the directory won't be created by default. 

https://git.haiku-os.org/haiku/tree/headers/os/storage/FindDirectory.h#n206

wattoc: The USER_NONPACKAGED_DATA_DIRECTORY?

The PostalPlus folder is created on launch if the app is run with nothing else, but the user will need to copy the files across as instructed for any of it to work anyway.

korli: it's irrelevant because the code under doesn't even check whether the folder PostalPlus exists.

korli: BTW have you a reason not to use B_USER_SETTINGS_DIRECTORY ?

wattoc: It's irrelevant? What are you talking about?  The game won't work at all without the data files being copied over.  This is the same for loads of ports.  If the folder is missing and therefore the data files are missing then the game doesn't work.  Would you like me to also code something in so it can access your GOG or Steam account to download all the game assets and install them for you?

There is code elsewhere that creates the PostalPlus folder, etc.  All I'm doing is working around the packaging system and conventions which you insist on.  I doubt you have looked at the code at all as a whole and are simply commenting on patch, recipe, which is useless.  There are checks elsewhere in the code.  All I have done is mimic the existing functionality and rework some stuff so it'll run at all on Haiku.

This type of feedback is useless.  One line comments don't inform anything, linking to a header for a function?  Is this some ego thing?

It does not use the B_USER_SETTINGS_DIRECTORY because it's not a Haiku app, but a port of a game from the mid-90s where such things didn't exist.  There's a few functions which literally all the file system calls in the program use which would take too much time and effort to rework for my level of interest.

If you want to rework it so it makes separate calls for prefs file, save games, etc. and can then follow the Haiku conventions you are welcome to.  But I would start with the 90% of Haiku Ports that also don't follow the conventions properly.

I have better things to do than tit for tat with you.

pulkomandy: Please calm down. People reviewing your code need to make sure they understand what you are doing, and so they ask questions. It is not an attack on what you did.

We will collectively be responsible for maintaining this, and so we need to understand how it works. So we make sure of that before merging it. That's all.

wattoc: If you want people to contribute then you need to make meaningful feedback.  Telling me something is "irrelevant" seems needlessly aggressive and unhelpful but I need to calm down.

I don't understand why I seem to be singled out by certain individuals for additional scrutiny.  This is not the first time and makes me wonder why I bother at all.

Comments as above are not meaningful, useful, etc. and show that the commenter hasn't done much other than a shallow review.  If you want people to work with this project then put some effort in.  If there are additional rules, conventions, etc. that need to be followed then these should be documented up front.  Frankly the majority of my base for recipes is from finding other recipes that look similar and adapting them.  Then I am told that something in a similar recipe is not correct.  This is the basis for my work since the documentation is not really good enough.

If it is wrong then you can simply point at the rule that is documented.  Meanwhile I can point at all the other ports where this was apparently fine and wonder why I merit special attention.

threedeyes: Instead of specifying a constant path to resource files, I would make a launcher (see Arx-Libertatis, Julius, dhewm3 for example). It will be very easy to add support for a new game to this launcher - https://github.com/EXL/BeGameLauncher

screenshot48

wattoc: The main problem with this is the game almost hardcodes everything in to using the PostalPlus folder.  It has one function it calls for everything that effectively assumes everything will sit under that folder.  You can redirect some things within the INI file but will later hit issues with it loading from the file system direct and it won't get past the load screen when starting a new level.  Effectively it requires some sort of constant path because it's basically hardcoded to one - the only thing I've done is make it use a Haiku friendly path.

To do things differently would require reworking that function and also detecting the type of file it is trying to access.

There's also no point in storing the POSTAL.INI somewhere else since it will recreate it under the PostalPlus folder anyway.

These are all issues that can be resolved by someone else if they have the interest.

korli: I'm asking because a standard user won't look for a settings file in USER_NONPACKAGED_DATA_DIRECTORY.

I take this is acceptable to everyone then.

PR #6174, opened 2021-09-04
kiwisolver, enable secondary architecture for 32bit

Begasus:


Begasus:
Needed for matplotlib

PR #6185, opened 2021-09-13
radare2, fix missing/duplicate entries

Begasus:


trufae:
Could you take a look at this package and update it for 5.7.0? it builds without any change, but will require to update some checksums and i don't have the environment ready for doing it right now. is there a better way to get packages updated in haiku?

Begasus:
I'll take a look at it once I get around to finishing up wip atm for other projects (if no one beats me to it), thanks for pointing out.

PS if you are sure the build is OK you could push a PR from any OS that has access to github (and can use git cmd's),  if you have the correct checkum it's a matter of renaming the recipe's version and updating the checksum (I sometimes do it from Ubuntu also).

Begasus:
@trufae Updated to version 5.7.2 just now in: https://github.com/haikuports/haikuports/pull/6998

PR #6190, opened 2021-09-15
Libnova 0.16.0

extrowerk:
New recipe


korli:
It looks like a dead project, is this library needed?

extrowerk:
> It looks like a dead project, is this library needed?

xygrib depends on it: https://opengribs.org/en/xygrib which i already got it built and run on Haiku. Will provide a recipe for that too. I don't know any other multiplatform meteorological program with the same feature set.

Begasus:
Repology still mentions this: https://repology.org/projects/?search=Libnova

PR #6200, opened 2021-09-20
New recipe: JXL Translator

wattoc:
@Begasus please check this one :)

### DIFF
--- a/haiku-apps/jxl_translator/jxl_translator-0.1.0.recipe
+++ b/haiku-apps/jxl_translator/jxl_translator-0.1.0.recipe
@@ -1,0 +1,47 @@
+SUMMARY="JPEG-XL Translator addon"
+DESCRIPTION="JPEG-XL Translator is a translator add-on for reading and writing \
+the JPEG-XL format.
+
+Using this add-on will allow any application using Haiku Data Translations \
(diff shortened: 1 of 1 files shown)


Begasus:
Will try it out when I get a chance to it, in the meantime all that could check could do the same :) Thanks so far! :+1:

Begasus:
Build is OK as is the install part, one question, on 32bit this is installed in the x86 subdirectory, whey I try to select a "save as" in ShowImage the translator doesn't show up? (ps, not checked on 64bit yet)

korli:
> Build is OK as is the install part, one question, on 32bit this is installed in the x86 subdirectory, whey I try to select a "save as" in ShowImage the translator doesn't show up? (ps, not checked on 64bit yet)

Then the translator will only show up on 32-bit for x86 applications.

wattoc:
The gcc2 exclusion is inherited from the libjxl recipe - I suspect the translator itself would compile OK on gcc2 but the dependency wouldn't be available.  I'm not aware of any way to make it work but someone else might know.

Begasus:
Some other translators are suffering from the same, no gcc2 possible, don't think it should really matter at this point.
As for me this could be merged.

PR #6220, opened 2021-10-01
sdl2_pango, new recipe

Begasus:


Begasus:
This is one needed when TuxPaint goes SDL2

PR #6229, opened 2021-10-03
texlive: split the package into multiple subpackages

### LINKED ISSUE #1767: Texlive: split up to avoid a 800MB main package

jmairboeck:
This splits the texlive data package into multiple subpackages, based on package collections as provided by TeX Live.

Some scripts are used to generate parts of the recipe (package infos and packageEntries for the subpackages). The subpackageInfos are pregenerated and sourced from the main recipe because for generating them, the source tarballs need to be already extracted and some stuff put in place at least (as is done by the BUILD function now). The packageEntries script is actually invoked during the install phase.

When trying to install the packages, I got an error from pkgman:

```
failed to commit transaction: Failed to open directory "/boot/system/packages/administrative/writable-files/texlive-2021-1/var/texlive/fonts/conf": No such file or directory
```

It seems that the `texlive-fontconfig.conf` and its parent directories is missing somehow.

This contains also an experimental version of createSubpackageInfos for now which adds provides for the individual packages. I think it should be removed before this is merged (if we don't want to use it in some form). So I would suggest adding the "wip-no-merge" tag for now.

Fixes #1767


app-text/texlive/additional-files/createSubpackageInfos_provides.pl
```diff
@@ -0,0 +1,108 @@
+#!/bin/perl
```
jmairboeck, line 1: remove this (experimental) file before merging

app-text/texlive/additional-files/createSubpackageInfos.pl
```diff
@@ -0,0 +1,98 @@
+#!/bin/perl
+
```
jmairboeck, line 2: Re-adding review comments from #6115

Should I add a license header here?

As to which one, I suppose it would have to be GPL v2 because it depends on such code. But I am open to MIT too, if that is possible.

app-text/texlive/additional-files/getPackageEntries.pl
```diff
@@ -0,0 +1,34 @@
+#!/bin/perl
+
```
jmairboeck, line 2: Re-adding review comments from #6115

License header here too?

jmairboeck:
I have now removed the alternate version of the "createSubpackageInfos" script. I can maybe put it somewhere if someone still needs or wants them.

If my other inline comments (about license headers) need not be addressed, this could be merged from my standpoint. The recipe is marked stll marked as "untested"

app-text/texlive/additional-files/createSubpackageInfos.pl
```diff
@@ -0,0 +1,98 @@
...
+my $tlpdb = TeXLive::TLPDB->new(root => $ARGV[0]);
+my @collections = $tlpdb->collections();
+
+for my $collectionPkgName (@collections) {
+	my $collection = substr $collectionPkgName, 11;
+	my $collectionPkg = $tlpdb->get_package($collectionPkgName);
+	my $shortdesc = $collectionPkg->shortdesc();
+	my $longdesc = $collectionPkg->longdesc();
+	if (length $shortdesc >= 48) { # SUMMARY must not exceed 80 characters in total
```
pulkomandy, line 31: I think we have upgraded to 100 characters now? Or does that not apply to the SUMMARY?

app-text/texlive/additional-files/createSubpackageInfos.pl
pulkomandy: MIT is fine.

The way GPL is "viral" is often misunderstood: it never adds further restrictions on things that are under a more open license and grant more rights.

So, if someone was to download the GPL'ed package, and then copy just some bits from this script that is under MIT license, and use that as allowed by the MIT license, that's fine.

It's not like this makes a big difference for a script anyway, since most of the extra things in GPL is about distributing sourcecode alongside binaries. In this case, there are no binaries because this is a script.

app-text/texlive/additional-files/createSubpackageInfos.pl
jmairboeck: That restriction comes from haikuporter, see https://github.com/haikuports/haikuporter/blob/226a85daf6255ad9b8528a349cea5214c36d97eb/HaikuPorter/Port.py#L393

Begasus: That could use an update too I guess on haikuporter side as this has been changed not too long ago in haikuports guidelines: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#general-tips

Begasus: On the other hand, if we use SUMMARY in rdef files it would probably look strange in FileTypes?

Begasus:
So far it looks OK, if no one objects we could merge this (it's still disabled - as requested, due to it's size).
Been tested by @jmairboeck and myself.

PR #6232, opened 2021-10-05
xpdf, bump version, remove tools provided by poppler (Arch approach)

Begasus:


Begasus:
Having poppler installed (more common) and after installing xpdf the commands from xpdf take precedence over the ones from poppler, hence only the ones not provided by poppler are left in (pdftopng (checked on IRC and poppler doesn't provide it, not even in their latest release)).

PR #6236, opened 2021-10-05
xapian, bump version

Begasus:


Begasus:
Bumped to latest stable, last one had some KDL iirc, maybe could be checked on x86_64 (no KDL when running the tests on 32bit)

PR #6253, opened 2021-10-08
Spout: New recipe

win8linux:
*Spout* is a simple game originally developed by kuni wherein the player flies through an endless cave in a box-shaped ship and must use the ship's exhaust to clear the way.

This recipe is based on a [fork](https://github.com/win8linux/spout/tree/patched) of the [Unix version by Nick White](https://njw.me.uk/spout/) that integrates the following patches from Debian:

* [Fix hardening errors from unused result](https://salsa.debian.org/smlx-guest/spout/-/blob/debian/sid/debian/patches/0002-Fix-hardening-errors-from-unused-result.patch)
* [Fix call to open with missing argument](https://salsa.debian.org/smlx-guest/spout/-/blob/debian/sid/debian/patches/0003-Fix-call-to-open-with-missing-argument.patch)
* [Fix segfault on 64-bit platforms due to incorrect types](https://salsa.debian.org/smlx-guest/spout/-/blob/debian/sid/debian/patches/0004-Fix-segfault-on-64-bit-platforms-due-to-incorrect-ty.patch)
* [Remove hard-coded library search path /usr/lib](https://salsa.debian.org/smlx-guest/spout/-/blob/debian/sid/debian/patches/0007-Remove-hard-coded-library-search-path-usr-lib.patch)
* [Remove redundant out-of-range comparison](https://salsa.debian.org/smlx-guest/spout/-/blob/debian/sid/debian/patches/0008-Remove-redundant-out-of-range-comparison.patch)
* [Add exit instructions on scroll screen](https://salsa.debian.org/smlx-guest/spout/-/blob/debian/sid/debian/patches/0009-Add-exit-instructions-on-scroll-screen.patch)

Currently a work in progress, as the recipe fails to build for now.

### DIFF
--- a/games-arcade/spout/spout-1.4b.recipe
+++ b/games-arcade/spout/spout-1.4b.recipe
@@ -23,7 +23,8 @@ REVISION="1"
 SOURCE_URI="https://github.com/win8linux/spout/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="3f3c1d9e9c67e5512e786fc94eaeaf3c6adabd5abe08653adbaff4bb03cfaf66"
 
-ARCHITECTURES="all"
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	spout$secondaryArchSuffix = $portVersion
@@ -50,7 +51,7 @@ BUILD()
 
 INSTALL()
 {
-	mkdir -p $appsDir/spout
+	mkdir -p $appsDir
 	mv ./spout $appsDir/spout
 	addAppDeskbarSymlink $appsDir/spout "Spout"
 }
--- a/games-arcade/spout/spout-1.4b.recipe
+++ b/games-arcade/spout/spout-1.4b.recipe
@@ -1,0 +1,59 @@
+SUMMARY="A simple caveflying game"
+DESCRIPTION="Spout was originally written for a handheld by kuni, and soon \
+afterwards was ported to Windows using Cygwin and SDL and released under the \
+MIT license.
+
+In 2004 a 'Unix version' was released, which mostly just slapped autotools \
+into the Windows version and infringed the license.
+
+This is a new Unix version, based on the original Windows code by kuni, \
+which aims to add useful features and simplify the code.
+
+Controls:
+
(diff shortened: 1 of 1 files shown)


korli:
> > Only happens on Haiku AFAIK, since it does build successfully on Linux 64-bit.
> 
> Unfortunate, time.h is included for a few types. Please submit a bug report.

As a workaround you could patch spout.c to use another local variable name as time.

games-arcade/spout/spout-1.4b.recipe
```diff
@@ -0,0 +1,56 @@
...
+Esc: Pause
+Shift-Esc: Quit"
+HOMEPAGE="https://njw.me.uk/spout/"
+COPYRIGHT="2001-2006 Kuni, 2010 Nick White"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/win8linux/spout/archive/$portVersion.tar.gz"
+CHECKSUM_SHA256="3f3c1d9e9c67e5512e786fc94eaeaf3c6adabd5abe08653adbaff4bb03cfaf66"
+
+ARCHITECTURES="all"
```
Begasus, line 26: Doesn't build for gcc2, so you need to enable secondary architecture for 32bit

Begasus:
> I've tried replacing `time` with another variable name, but the same error occurs still.

Just replace `time` with `_time` (in all cases in spout.c) and the build will work ;)

win8linux:
Thanks, can confirm that doing so does indeed let the build finish. However, the game itself is unplayable with graphical aberrations and instantly going to the game over screen upon user interaction. Will prolly need to investigate further in the future, see what needs to be fixed.

Begasus:
> Should I still upload those changes remotely?

Yes please, the build in itself should be OK with those changes, we'll hold off on merging until you can confirm this is running as expected.

Begasus:
@win8linux any progress on this? If not maybe merge as disabled for all arch's?

win8linux:
I'll just close this, since apparently even Debian has stopped shipping this in the repos of the last couple stable versions due to how broken it has become.

Begasus:
Thanks for getting back on this, if need be someone could proceed from this to check it out. +1

PR #6255, opened 2021-10-09
xaos, bump version, move to x11-misc following Gentoo's layout

Begasus:
Build disabled as it launches XaoS during build, issue filed upstream for this at: https://github.com/xaos-project/XaoS/issues/221


Begasus:
Moved the old recipes to the new location (don't know if that means a rebuild for them will be done on buildmaster, but we'll find out - some changes needed there)

PR #6263, opened 2021-10-11
pogger: new recipe

JadedCtrl:
New recipe for an RSS/Atom program.


Begasus:
PoggerShaddow crashes on first launch from Pogger, the second time it seems to be OK.

JadedCtrl:
Oop, thanks for the report! I was able to replicate that one, should be taken care of with latest commit.

Begasus:
Merging, maybe some works still needed but works fine as is, updated the srcGitRev to latest commit. Thanks! Hope to see you around again. :)

PR #6269, opened 2021-10-12
Dragengine - New feature

LordOfDragons:
Updated to version 1.6 . This one includes now also "libdelauncher.so", which can be used to create launchers for Haiku.


Begasus:
Could you squash the commits? Also it would be nice if you master branch was kept on par with haikuports master branch and work with featured branches, this would mean not all prior commits would bin in the PR (like now). Thanks :)

PR #6271, opened 2021-10-13
opencl_headers: bump to 3.0

kenmays:
Tested on Haiku R1B3.


Begasus:
Did you check a build for the packages depending on it during BUILD()?
- [ ] opensubdiv
- [ ] opencv3
- [ ] opencv

I know it don't interfere with the installed packages, but would be good to know if those still build with this PR

PR #6274, opened 2021-10-14
dragengine release 1.6

LordOfDragons:

PR #6276, opened 2021-10-15
Drag[en]gine 1.6: Fixed issue with implicit casting of nullptr to function pointers

LordOfDragons:
Follow up of https://github.com/haikuports/haikuports/pull/6274#issuecomment-944051888


Begasus:
Crossing fingers :)

Begasus:
Argh, no revision bump ...

LordOfDragons:
I used sha256sum to calculate the checksum. Not good if this does not produce the right results :(

LordOfDragons:
Okay. It looks like I need to explicit cast all nullptr usage in the vulkan code. Not a big deal... just annoying.

PR #6278, opened 2021-10-15
replaced all vulkan specific use of nullptr with VK_NULL_HANDLE.

LordOfDragons:
this #define makes a distinction between 64bit and 32bit.
this should now compile on 32bit.


Begasus:
Thanks, will do a follow up build here on 32bit before we merge :)

PR #6279, opened 2021-10-16
stress-ng: bump to 0.13.05

davidkaroly:
Stress-ng now compiles without any patchset.

It won't build on gcc2 so architectures are set to "all !x86_gcc2." I tried building for x86 as secondary architecture but the binary won't start, so some more investigation is needed for 32-bit.


extrowerk:
Builds / packages fine on 64 bit.

PR #6283, opened 2021-10-16
new recipe: fig2dev

jmairboeck:
This is an optional dependency of dblatex (see #6078).

Quite a number of tests fail when run via haikuporter that work when running it without. I don't know why.

Tested (building the package) on 64 bit (nightly) and 32 bit (beta 3)

### DIFF
--- a/media-gfx/fig2dev/fig2dev-3.2.8b.recipe
+++ b/media-gfx/fig2dev/fig2dev-3.2.8b.recipe
@@ -56,7 +56,7 @@ TEST_REQUIRES="
 "
 
 defineDebugInfoPackage fig2dev$secondaryArchSuffix \
-	$binDir/fig2dev
+	$commandBinDir/fig2dev
 
 BUILD()
 {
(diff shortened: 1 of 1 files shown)


extrowerk:
Adding 
```
	cmd:aclocal
	cmd:gzip
	cmd:sed
	cmd:gs
	cmd:pnmtotiff
#	cmd:latex
	cmd:gm
	cmd:convert
```
to the REQUIRES / BUILD REQUIRES / TEST_REQUIRES section solves most problems, but it seems something wrong with the eps test, ghostscript just dies (test 94/95). This happens outside of haikuporter too.

```
 94: eps with tiff preview                           FAILED (bitmaps.at:125)
 95: eps with tiff preview, use "-" for stdin        FAILED (bitmaps.at:132)
```

The commands running interactively results:

TEST 94:
```
Error calling ghostscript: gs -q -dSAFER -sDEVICE=tifflzw -r72 -g32x14 -o '/tmp//f2dprevhhsYEv' /tmp//f2deps9TEjgs
No preview will be produced
```

TEST 95:
```
no output????
```

Begasus:
Maybe you could enable transfig too? See https://github.com/archlinux/svntogit-community/blob/packages/fig2dev/trunk/PKGBUILD#L24

jmairboeck:
I have enabled transfig now. I have left it out initially because it is apparently considered obsolete. I have used the version number from its manpage (3.2.4) for its PROVIDES.

PR #6293, opened 2021-10-17
gettext: bump

extrowerk:
The -presumably- gcc2 patches doesn't maintained. If needed, i can try to apply them, but i am unable to test it.
Question: why the gettext package doesn't have _devel subpackage?

### DIFF
--- a/sys-devel/gettext/gettext-0.21.recipe
+++ b/sys-devel/gettext/gettext-0.21.recipe
@@ -1,0 +1,130 @@
+SUMMARY="GNU tools for native language support"
+DESCRIPTION="GNU 'gettext' is an important step for the GNU Translation \
+Project, as it is an asset on which we may build many other steps. This \
+package offers to programmers, translators, and even users, a well integrated \
+set of tools and documentation. Specifically, the GNU 'gettext' utilities are \
+a set of tools that provides a framework to help other GNU packages produce \
+multi-lingual messages. These tools include a set of conventions about how \
+programs should be written to support message catalogs, a directory and file \
+naming organization for the message catalogs themselves, a runtime library \
+supporting the retrieval of translated messages, and a few stand-alone \
+programs to massage in various ways the sets of translatable strings, or \
+already translated strings. A special GNU Emacs mode also helps interested \
+parties in preparing these sets, or bringing them up to date."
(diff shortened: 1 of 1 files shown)


extrowerk:
FAIL: test-supersede
FAIL: test-posix_spawn_file_actions_adddup2

Begasus:
> Question: why the gettext package doesn't have _devel subpackage?

Good catch, I guess no one checked it out to do so?

Begasus:
Too many parse errors on gcc2 (not included the previous patchset, but those files are not involved anyway)

Begasus:
Maybe keep the old version for x86_gcc2 only and bump this one for x86 and x86_64

Begasus:
228 hits on libintl, this is being used by a lot of of recipes, should check a x86 build and see if anything gets broken

waddlesplash:
```
problem 1: package gettext_x86-0.19.8.1-7 requires
gettext_x86_libintl==0.19.8.1, but none of the providers can be
```

You need to upgrade the subpackages at the same time.

Begasus:
> ```
> problem 1: package gettext_x86-0.19.8.1-7 requires
> gettext_x86_libintl==0.19.8.1, but none of the providers can be
> ```
> 
> You need to upgrade the subpackages at the same time.

I'm not sure the list is complete, those are only the installed ones here

Begasus:
@waddlesplash aside from my comments is this good to merge?

waddlesplash:
I don't know, it sounds like we might need to rebuild glib2 at the same time?

Begasus:
OK, packages installed now, but a bit strange.
Rebuild this version for gettext, rebuild glib2 (against the packages for gettext in .../haikuports/packages)
Dropped all packages for these 2 in /system/packages
Getting the notice mentioned above for libintl (canceled that)
Rebooted to see if nothing was broken
Installed the new gettext packages by dropping them in /system/packages (asked me to install new and uninstall old)
Things seem fine now ...

Begasus:
Did a local build again (added a devel package to it), still the same issue trying to install libintl resolves in uninstalling several packages, for instance Qt (Otter is removed also ...), trying to install Qt5 again results in the need to downgrade gettext, so that is atleast one that needs a revbump.
Would be nice if this could be fixed (not sure I'm the man to do this), our current version is from 2016!

Issue #6259, opened 2021-10-10
Create Haiku-style icon for Kristall


win8linux:
[Kristall's recipe](https://github.com/haikuports/haikuports/pull/5359) has an icon, but it doesn't quite fit in with the rest of Haiku.

Kristall icon for reference:
![image](https://user-images.githubusercontent.com/11584387/136697121-fcca62d2-1977-4f17-8a2a-0abbdb1c826d.png)
[SVG](https://github.com/MasterQ32/kristall/blob/master/src/icons/kristall.svg)

pulkomandy:
The icon for Ruby here may be used as a base: https://haikuarchives.github.io/ZumisIcons/myhaiku/3rd%20party%20apps/index.html but needs recoloring

Handmaus:
Hello, i have made an icon for this program.
Already tried recoloring zuMi's Ruby icon, as Pulkomandy suggested, but it doesn't felt really ok, as it looked too good and a bit out of place, as this program is a browser for protocols that aim to be a return to a simpler internet.
So i decided to go the extra mile and create an icon that adapts to Haiku's style, but that also respects the program's spirit trying to retain the Kristall logo simplicity.
It's also my first successful attempt at creating something with Icon-o-Matic that at least don't make anyone's eyes bleed.
I'm such a brick using this program -_-
Hope you like it :)

![Kristall HaikuStyle B 03 Clear 03](https://user-images.githubusercontent.com/92672828/137625639-0a970cec-569c-447b-a8cb-fe2965150030.png)

BTW: I'm a total beginner in regards to this git thing (signed up in GitHub just today), and naturally there's a lot of doubts that come to my mind.
But there's this thing that strikes me the most after reading the contributing guidelines:
 is it really needed to download the thousands of sources contained in the entire Haikuports repository (isn't that what happens when you "fork" and "clone"?), then do one of this "pull requests", just to contribute an icon to one recipe?
Maybe in the end is not that big of a deal, but it sure feels a bit overwhelming to a non-coder like me.
As i don't want to make you wait another year, for me to learn how to do it the proper way, in the meantime here's the hvif attached to this message:
[Kristall.hvif.zip](https://github.com/haikuports/haikuports/files/7360143/Kristall.hvif.zip)
If something else is needed, please ask.

pulkomandy:
[...]

PR #6305, opened 2021-10-19
Kristall.iom Icon-O-Matic icon file

Handmaus:
This is the source .iom file for the .hvif icon I clumsily attached to a comment to issue #6259.
It has all the names for paths, shapes and styles, so that anyone can grasp what's what in case someone needs to edit or reuse it.

(For personal reference, original file name: Kristall HaikuStyle B 03 Clear 03.iom)


humdingerb:
I missed that #6297 was opened by win8linux...

Those named objects are not stored in attributes of IOM files, so they wouldn't get lost. I could imagine that you originally exported to HVIF (thereby losing those names), then opened that HVIF in IOM and saved as .IOM file which then didn't have those names anymore.  If you cannot find the original IOM with named objects, it's no biggie either.

WRT github, you'll fare best with using git within Terminal instead of a browser under Haiku in the long run.
See the wiki at https://github.com/haikuports/haikuports/wiki , esp. the DevelopmentModel page. Do join us on IRC #haiku for realtime assistance. :)

Handmaus:
> ...checking it out properly, I see the named objects...

Yay! :D

> So, everything would be OK, if the .iom was in the additional-files sub-folder... :P

Not yay.... :<

I had a slight suspicion that someone had put that folder there for some reason ^_^¡

Handmaus:
Wohoooo! My first contrib to Haiku!
@win8linux , @pulkomandy ,
  @humdingerb ,
    thank you!
`-------v------------´
. . .      \ (^ o ^) /

win8linux:
So my PR gets merged first since that has the new icon, then this one to add the correct IOM file. Is that what's going to happen or should this one get another commit to add the new icon in instead?

humdingerb:
Damn. I confused the two PRs... Pfft. will fix the IOM file location in another commit...

PR #6321, opened 2021-10-24
libxo: new recipe

extrowerk:


dev-libs/libxo/libxo-1.6.0.recipe
```diff
@@ -0,0 +1,86 @@
...
+
+ARCHITECTURES="all"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+
+libVersion="0.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	libxo$secondaryArchSuffix = $portVersion
+	cmd:libxo_config = $libVersionCompat
```
Begasus, line 21: Maybe move the config file to the devel package?

Begasus:
Parse errors with gcc2, so you can disable `x86_gcc2` :)

PR #6326, opened 2021-10-24
coreutils: show creation time in stat and ls output

### LINKED ISSUE #5628: stat command does not show Birth entry
Hello,

I have created a .txt file on desktop. In default file manager under list view (manually add "Created" to column) birth time = creation time is displayed.

But when I do with `stat` command it shows me this. Birth entry is not displayed, it shows an empty character -.

Access: 2021-01-21 02:24:51.235050163 +0000

Modify: 2021-01-21 02:24:49.608383509 +0000

Change: 2021-01-21 02:24:49.608383509 +0000

Birth: -

The problem can be fixed when using coreutils 8.31 and later, because since version 8.31 `stat` command can display the creation time when the file system stores it. ​https://lists.gnu.org/archive/html/coreutils/2019-03/msg00042.html

Since version 8.32 it can also display the birth entry via `ls` command
http://savannah.gnu.org/forum/forum.php?forum_id=9693

It would be good if always the latest coreutils version is used in Haiku.

volo-droid:
Fixes #5628

This would allow `stat` and `ls` to display file creation (aka birth) time 

Screenshot 2021-10-24 at 22 37 44


Begasus:
@volo-droid could you add `cmd:basenc` to the recipe and rm the others mentioned before in the `INSTALL` section as @korli  replied to it?

korli:
@Begasus it's OK to merge with the warnings (not a regression)

Begasus:
@korli OK didn't want to touch build system files, if this OK it can be merged then.

volo-droid:
@korli thanks for taking care of those commands in the subsequent commits

volo-droid:
@korli Do you think it's worth to upstream the usage of `st_crtim` to the coreutils project?

p.s.: thanks for taking care of those missing commands in the subsequent commits

korli:
@volo-droid yeah if it's small enough. ideally this would happen in https://github.com/coreutils/gnulib/blob/master/m4/stat-time.m4 but better to ask the coreutils devs.

Issue #6284, opened 2021-10-17
Ghostscript crashes


jmairboeck:
Ghostscript crashes in two testcases of fig2dev (see #6283)
[gs-6083-debug-16-10-2021-15-44-27.txt](https://github.com/haikuports/haikuports/files/7359602/gs-6083-debug-16-10-2021-15-44-27.txt)

Begasus:
There were no crashes on 32bit with the current ghostscript version

jmairboeck:
Version 9.55.0 shows the same crashes (and other issues) as version 9.50.

jmairboeck:
ghostscript8 is not available for x86_64, only for x86. Should we enable it also for other architectures (and maybe update it a bit as well)?

Is the reason why this exists the license?

pulkomandy:
No, the reason is it is needed by old BeOS apps. There is no need to port it over to x86_64 please. Use up to date versions there.

PR #6337, opened 2021-10-27
ghostscript: update to version 9.55.0, remove old versions

### LINKED ISSUE #6312: Update ghostscript to the lates version.
A simple update can solve the problem #6284 , but no guarantee.

https://github.com/ArtifexSoftware/ghostpdl-downloads/releases

jmairboeck:
Fixes #6312, but untested, so not sure yet if it helps with #6284 (or the other issues with ghostscript in Lilypond)


Begasus:
Was looking into this also before I got side-tracked with poppler (again) :)
There is a typo patch floating around that other OS's use also: https://src.fedoraproject.org/rpms/ghostscript/blob/rawhide/f/ghostscript-9.55-fix-typo-in-gsicc-lcms2.patch
They also rm leptonica and tesseract in their recipes: https://github.com/archlinux/svntogit-packages/blob/packages/ghostscript/trunk/PKGBUILD#L32

jmairboeck:
The reason why it built before without that patch is because it actually still used the bundled version (lcms2mt). I have fixed that and included the patch.

I have also looked what to do with leptonica and tesseract: it seems that the configure script doesn't try to use the system version of these libs, presumably because they use modified versions. See https://ghostscript.com/blog/ocr.html for details. I have therefore retained these directories for now.

I also added openjpeg (to use the system library instead of the bundled one) and libiconv to dependencies to fix the build of a driver. The detection of dlopen in the configure script has also been fixed.

jmairboeck:
ghostscript 9.55.0 (and fig2dev) builds fine on 32 bit (using gcc11). It leads to the same crashes as on 64 bit ...

PR #6340, opened 2021-10-28
 vmware_addons: add shared folders support for x64 arch

volo-droid:


volo-droid:
@Begasus yes, removing `-Werror` allows the build to proceed. How could we keep it for 64bit, but disable for 32?

Begasus:
> @Begasus yes, removing -Werror allows the build to proceed. How could we keep it for 64bit, but disable for 32?

@volo-droid disabling it for 64bit also shouldn't harm imho

volo-droid:
@Begasus yes, that what I've come to in the end. I've discovered it's also possible to use `CC_VER` from the `makefile-engine` to check for gcc2 (which is probably a better approach then splitting by 32/64-bit architecture, cause those warnings are gcc2 specific), and I had to use in another place to link the correct libraries, but for the `enhanced_backdoor` module that looked like overkill.

Issue #6327, opened 2021-10-24
coreutils ln: hardlink creation fails with exit code 0, potentially breaking scripts


barctor:
# System Information
Haiku R1/beta3 (Revision hrev55181+62) x86_64

# Expected behaviour

Using `ln` to create a hardlink should fail and the binary should exit with a non-zero status code (this behaviour confirmed on Linux).  

# Actual behaviour

No hardlink is created, but no errors are printed to stderr and the program exits with status code 0.  

# Reproducing

* Create a file
* Attempt to create a hardlink to this file using ln
* Observe the exit status

In bash this may be achieved with `touch empty; ln empty empty2; echo $?`  
This should print 0 to the terminal.  

# Why might this be a problem

Any script that tries to create a hardlink and uses the exit code to decide if this worked can fail in potentially unpredictable ways.  
To observe this in practice, [download](http://zbar.sourceforge.net/download.html) version 0.10 of the zbar library and attempt to build it.  
To build zbar extract the archive, and run `./configure --disable-video --without-gtk --without-python --without-qt --prefix /boot/home/Software/libzbar/ --build x86_64-unknown-haiku` then `make`.  

The build fails because `libtool` attempts to hardlink a couple of object files using `ln`, which causes a later build step to fail:
```
copying selected object files to avoid basename conflicts...
libtool: link: ln zbar/processor/zbar_libzbar_la-null.o zbar/.libs/libzbar.lax/lt1-zbar_libzbar_la-null.o || cp zbar/processor/zbar_libzbar_la-null.o zbar/.libs/libzbar.lax/lt1-zbar_libzbar_la-null.o
libtool: link: ln zbar/window/zbar_libzbar_la-null.o zbar/.libs/libzbar.lax/lt2-zbar_libzbar_la-null.o || cp zbar/window/zbar_libzbar_la-null.o zbar/.libs/libzbar.lax/lt2-zbar_libzbar_la-null.o
[...]

PR #6352, opened 2021-10-31
GCC 11: enable for every arch

extrowerk:
It seems @pulkomandy have forgot about the 64 bit users. Lets fix it.


waddlesplash:
I guess there's little reason not to do this at this point...

waddlesplash:
Well, I guess someone did report that GCC build fails due to https://github.com/haikuports/haikuports/issues/6327. So maybe that should be considered first.

extrowerk:
@pulkomandy : makes sense, sorry for the misunderstanding.

PR #6356, opened 2021-11-01
gerbera, bump version

Begasus:


Begasus:
Tests breaks the build (I guess our port for gtest isn't working as expected (hence I hardly use it)), so disabled it ...
Build on 32bit with gcc11 also seemed to have fixed the crash.

PR #6359, opened 2021-11-02
netperf: bump to 2.7.0

davidkaroly:
Build verified on 32-bit and 64-bit.
I've added some fixups for the select(), now at least I'm able to get some numbers out of netperf.


korli:
@davidkaroly please try to upstream at least part of the patchset

PR #6362, opened 2021-11-03
gcc: enable build for arm

davidkaroly:
one more change that seems to be required for a successful arm bootstrap


pulkomandy:
You can just remove "arm" altogether and it will be enabled thanks to the "all" entry.

davidkaroly:
yep, you're right.
I've updated it accordingly.

PR #6372, opened 2021-11-07
vifm, bump version, fix build

Begasus:


Begasus:
Haiku doesn't provide the `fdatasync` function, so needed to replace it with `fsync`, upstream mentioned at: https://github.com/vifm/vifm/issues/737#issuecomment-962586089

PR #6378, opened 2021-11-08
SRB2: Icon shadow and gradient fixes

win8linux:
Updates icon with fixed shadow and gradients from @threedeyes:
image


pulkomandy:
You need to update the REVISION in the recipe, otherwise it won't be rebuilt.

Issue #6382, opened 2021-11-08
GCC11.2: -fstack-protector doesn't works on x86


extrowerk:
Build fails:
https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/2066/builds/77002.log

korli:
yeah sorry this was scattered among a few commits. This one might help.
https://github.com/haikuports/haikuports/commit/558523994f3ab645b82e257a0f40f5b3f9940469

extrowerk:
@korli i don't really see what should we do now. The referred commit removes patches, but doesn1t add new ones.
Originally 11.1 didint had this patch, i added it in 11.2, but it is still broken. Could you explain what kind of change is needed? I can provide the required patches.

korli:
@extrowerk please look at the commit (the recipe)

pulkomandy:
This seems to be the part that needs to be added into the nez recipe:

	# merge libssp_nonshared.a in libgcc.a on x86
	if [ "$targetArchitecture" = "x86_gcc2" ]; then
		(mkdir temp_libgcc; cd temp_libgcc; ar x ../$gccLibDir/libssp_nonshared.a;
			ar x ../$gccLibDir/libgcc.a; cd ..)
		ar -qc $developLibDir/libgcc.a temp_libgcc/*.o
		cp $developLibDir/libgcc.a $gccLibDir/libgcc.a
		rm -rf temp_libgcc
	fi

korli:
Thanks, fixed in e2db3969ce3b2ab09476db6552423532e9f18b6c

PR #6383, opened 2021-11-09
GCC: fix -fstack-protector on x86

extrowerk:
linking stack-protected objects on x86 would fail if -fstack-protector
wasn't passed to the linker. Only one symbol is missing, provided by
libssp_nonshared. So merge it in libgcc.a, which is always linked in.


extrowerk:
Fixes #6382

Untested, but sopposedly should work.

PR #6384, opened 2021-11-09
Kottan: new maintenance release 0.14.3 

andimachovec:
- links against libcolumnlistview directly 
- added new and updated language translations


pulkomandy:
you need to increment the REVISION, otherwise the package won't be rebuilt

korli:
> you need to increment the REVISION, otherwise the package won't be rebuilt

not for a new version

andimachovec:
Thanks korli for clarifying. Seems more logical to me than having to increment the revision for a new version.

pulkomandy:
Ah yes sorry, I didn't notice the filename had changed.

andimachovec:
No problem ;-) Thanks Pulkomandy!

PR #6385, opened 2021-11-10
gcc-8.3.0: build bootloader libraries

davidkaroly:
This change is meant to (at least partially) tackle issue [#16763](https://dev.haiku-os.org/ticket/16763) and the comments from review [#4703](https://review.haiku-os.org/c/haiku/+/4703)

The libraries libgcc-boot.a, libgcc_eh-boot.a, libsupc++-boot.a are built and packaged similarly to the gcc_bootstrap recipe from haikuports.cross.

I checked it on x86_64 and x86 secondary architecture. The build completes successfully and the new *-boot.a files are added to gcc_syslibs_devel package.

### DIFF
--- a/sys-devel/gcc/gcc-8.3.0_2019_05_24.recipe
+++ b/sys-devel/gcc/gcc-8.3.0_2019_05_24.recipe
@@ -285,7 +285,7 @@ BUILD()
 	mv saved/* .
 	rmdir saved
 
-	echo "### libsupc++-kernel ###"
+	echo "### libsupc++-boot###"
 
 	# build bootloader version of libsupc++.a (without threads or TLS)
 	cd $objectsDir/$effectiveTargetMachineTriple/libstdc++-v3/libsupc++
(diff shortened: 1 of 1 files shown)


sys-devel/gcc/gcc-8.3.0_2019_05_24.recipe
```diff
@@ -251,13 +256,57 @@ BUILD()
...
+	make clean
+	ln -sfn "$sourceDir/libgcc/gthr-single.h" gthr-default.h
+	make $jobArgs CFLAGS="-O2 $bootCcFlags" CXXFLAGS="-O2 $bootCcFlags"
+	mv libgcc.a libgcc-boot.a
+	mv libgcc_eh.a libgcc_eh-boot.a
+	ln -sfn "$sourceDir/libgcc/gthr-posix.h" gthr-default.h
+	mv saved/* .
+	rmdir saved
+
+	echo "### libsupc++-kernel ###"
```
korli, line 288:
```suggestion
	echo "### libsupc++-boot###"
```

davidkaroly: oops... you're right

waddlesplash:
Probably these patches should be applied to GCC 11 instead...

davidkaroly:
> Probably these patches should be applied to GCC 11 instead...

that's what I'm trying to do next

PR #6393, opened 2021-11-12
opencolorio1, blender, disable build

Begasus:


Begasus:
@korli this should be OK to disable the build for opencolorio and blender?

PR #6396, opened 2021-11-12
vulkan_headers: bump to 1.2.198

kenmays:


diversys:
https://github.com/haikuports/haikuports/commit/4ddada76bde13ac37366efa3cbcec0a4cbe64cdf includes vulcan headers. Not sure what is better. @X547 thoughts?

kenmays:
The vulkan recipe is OK for the newer header info. We can just fix that one for further Vulkan header/lib updates.

PR #6400, opened 2021-11-13
hiredis, new recipe

Begasus:


Begasus:
Test fail, but ccache builds fine with it.

PR #6401, opened 2021-11-13
ccache, bump version

Begasus:


Begasus:
Some tests still fail, but so it did with previous version.

dev-util/ccache/ccache-4.4.2.recipe
```diff
@@ -0,0 +1,81 @@
...
+	cmd:perl
+	cmd:sed
+	"
+
+BUILD()
+{
+	cmake -B build -S . \
+		$cmakeDirArgs \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_INSTALL_BINDIR=$commandBinDir
```
Begasus, line 68: Previous version used configure script with the line: ```--sysconfdir="`finddir B_SYSTEM_SETTINGS_DIRECTORY`"```
I think this should be covered with `$cmakeDirArgs` now? (with the variable `-DCMAKE_INSTALL_SYSCONFDIR`)

waddlesplash: Indeed it should, yes.

PR #6402, opened 2021-11-13
gcc-11.2.0: build bootloader libraries

davidkaroly:
second attempt, revised in line with the comment from @korli 

i'll do some additional checks to see if the previously failing recipes are any better with this change

please let me know if there's any other suggestion for verifying this.


davidkaroly:
I checked package contents after rebuilding with the revised recipe:
in gcc package rev. 4 the contents of gthr-default.h and c++config.h is overwritten
in new rev. 5 they are not overwritten any more

Trying to build retroshare:
it fails on gcc-11 package rev. 4 with an error message similar to the one seen in buildmaster logs
on rev. 5 it builds successfully

davidkaroly:
a similar update will be needed for gcc-8.3.0

PR #6413, opened 2021-11-17
drag[en]gine release 1.7.1

LordOfDragons:


games-engines/dragengine/dragengine-1.7.1.recipe
```diff
@@ -15,9 +15,9 @@ COPYRIGHT="2020-2021 DragonDreams"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
 REVISION="3"
```
diversys, line 17: Revision needs to be reset.

games-engines/dragengine/dragengine-1.7.1.recipe
```diff
@@ -14,10 +14,10 @@ HOMEPAGE="https://dragondreams.ch
 COPYRIGHT="2020-2021 DragonDreams"
 LICENSE="GNU GPL v3
 	GNU LGPL v3"
-REVISION="3"
-HAIKU_PATCH="-haiku3"
+REVISION="1"
+HAIKU_PATCH=""
```
diversys, line 18: Is it still needed?

LordOfDragons: Not in this recipe. I kept the value in case I need to make a patch version in an upcoming recipe. If though this is a problem I can remove it.

diversys: No other recipe use it so I think it would be cleaner to drop it.

Issue #6428, opened 2021-11-23
stress-ng: stress-icache does not work on x86


davidkaroly:
If I build stress-ng for x86 secondary architecture, I get a defunct executable:

```
~/haikuports/app-benchmarks/stress-ng/work-x86-0.13.05/sources/stress-ng-0.13.05> stress-ng
runtime_loader: /boot/home/haikuports/app-benchmarks/stress-ng/work-x86-0.13.05/sources/stress-ng-0.13.05/stress-ng: Could not map image: Bad data
~/haikuports/app-benchmarks/stress-ng/work-x86-0.13.05/sources/stress-ng-0.13.05>

```

The issue is caused by stress-icache as it's linked in its own section stress_icache_cal aligned on 64K boundary.
This results in a gap greater than 8K between two sections in the executable file, therefore runtime_loader gives as B_BAD_DATA.
see: https://github.com/haiku/haiku/blob/master/src/system/runtime_loader/images.cpp#L335

```
	// Check whether the segments have an unreasonable amount of unused space
	// inbetween.
	if (reservedSize > length + MAX_PAGE_SIZE * 2)
		return B_BAD_DATA;

```

No such error on x86_64 as MAX_PAGE_SIZE is defined as 2M there.

korli:
x86 uses an older binutils, might also be related

davidkaroly:
attaching "bad" and "good" section list generated with "readelf -S"

the original binary contains section stress_icache_cal aligned on 64K

if I remove the 64K alignment (but leave in the next alignment in the sources which is 16K) i get a binary where stress_icache_cal is aligned on 16K and now it's possible to load it on 32-bit Haiku

[stress_ng_bad.txt](https://github.com/haikuports/haikuports/files/7609791/stress_ng_bad.txt)
[stress_ng_good.txt](https://github.com/haikuports/haikuports/files/7609794/stress_ng_good.txt)

davidkaroly:
...and this seems to be a good solution, as the 64K-aligned section is only used if the page size is 64K which is not the case on x86.

davidkaroly:
#6435

Begasus:
![DSC_0387](https://user-images.githubusercontent.com/16057090/144702289-9c7f59c6-0d10-43f1-965b-aa2d96cd3e48.JPG)

KDL when trying to run one of the examples there ... (32bit R1B3)

diversys:
This should be reported in Haiku's Trac.

[...]

PR #6429, opened 2021-11-23
stress-ng: add x86 as secondary architecture

davidkaroly:
Disabling stress-icache on x86 as runtime_loader complains
about an "unreasonable amount of unused space."

see also issue #6428 

No such problem on x86_64 so stress-icache is still enabled there.

### DIFF
--- a/app-benchmarks/stress-ng/stress_ng-0.13.05.recipe
+++ b/app-benchmarks/stress-ng/stress_ng-0.13.05.recipe
@@ -38,25 +38,30 @@ SOURCE_URI="https://github.com/ColinIanKing/stress-ng/archive/V$portVersion.zip"
 SOURCE_DIR="stress-ng-$portVersion"
 CHECKSUM_SHA256="442c07327c5534566670a85dc67d50b1c643250aa09cc7129c793705d294cbeb"
 
+if [ $effectiveTargetArchitecture = x86 ]; then
+	PATCHES="stress_ng_x86-$portVersion.patchset"
+fi
+
 ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
-	stress_ng = $portVersion
-	cmd:stress_ng = $portVersion
+	stress_ng$secondaryArchSuffix = $portVersion
+	cmd:stress_ng$secondaryArchSuffix = $portVersion
 	"
 REQUIRES="
-	haiku
-	gcc_syslibs
-	lib:libz
+	haiku$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


app-benchmarks/stress-ng/stress_ng-0.13.05.recipe
```diff
@@ -38,25 +38,31 @@ SOURCE_URI="https://github.com/ColinIanKing/stress-ng/archive/V$portVersion.zip"
...
-	cmd:stress_ng = $portVersion
+	stress_ng$secondaryArchSuffix = $portVersion
+	cmd:stress_ng$secondaryArchSuffix = $portVersion
 	"
 REQUIRES="
-	haiku
-	gcc_syslibs
-	lib:libz
+	haiku$secondaryArchSuffix
+	gcc${secondaryArchSuffix}_syslibs
```
korli, line 54: I think this one can be removed

korli:
on x86, the bin should go in $prefix/bin

app-benchmarks/stress-ng/stress_ng-0.13.05.recipe
davidkaroly: gcc_syslibs is required because of libatomic, otherwise we get a warning during package build. (error if using strict mode)

gcc_syslibs_devel can be removed from BUILD_REQUIRES.

davidkaroly:
> on x86, the bin should go in $prefix/bin

the package contents on x86 look like this with the current recipe:

```
bin                                      0  2021-11-23 14:35:21  drwxr-xr-x
  stress-ng-x86                          0  2021-11-23 14:35:21  lrwxrwxrwx  -> x86/stress-ng
  x86                                    0  2021-11-23 14:35:21  drwxr-xr-x
    stress-ng                       955340  2021-11-23 14:35:21  -rwxr-xr-x
```

app-benchmarks/stress-ng/stress_ng-0.13.05.recipe
korli: then declare the library itself: lib:libatomic

korli:
> > on x86, the bin should go in $prefix/bin
> 
> the package contents on x86 look like this with the current recipe:

there won't be a package for x86_gcc2. the binary can be installed directly in $prefix/bin

app-benchmarks/stress-ng/stress_ng-0.13.05.recipe
davidkaroly: ok that makes sense

davidkaroly:
thx :)

PR #6430, opened 2021-11-24
wireshark: bump to version 2.6.20

davidkaroly:
We need to move around the include files as glib.h cannot be included from inside extern "C" anymore.
Also add realpath to initialization so wireshark can know where the binary is installed, otherwise it could not find dumpcap when started from the deskbar.


davidkaroly:
crash on exit is still there unfortunately

PR #6436, opened 2021-11-26
make: fix build after dirent refactor

davidkaroly:


korli:
this is only a build problem on nightlies, correct?

davidkaroly:
> this is only a build problem on nightlies, correct?

i think so

i'll have a try on beta3

davidkaroly:
yep, make builds successfully on beta3.
the issue is there on nightly.

waddlesplash:
It'll still build fine on beta3 though with this fix. So we probably should merge it.

PR #6459, opened 2021-12-04
fpcupdeluxe, new recipe

Begasus:


Begasus:
So far this is only for 32bit ...

sys-devel/fpcupdeluxe/fpcupdeluxe-2.2.0.recipe
```diff
@@ -0,0 +1,53 @@
...
+	non-packaged/bin/fpcupdeluxe-i386-haiku-qt5 keep-old
+	"
+
+PROVIDES="
+	fpcupdeluxe$secondaryArchSuffix = $portVersion
+	cmd:fpcupdeluxe$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libQt5pas$secondaryArchSuffix
```
korli, line 43: with binaries it is usually a good idea to specify the versions based on the output of "readelf -d"

sys-devel/fpcupdeluxe/fpcupdeluxe-2.2.0.recipe
```diff
@@ -0,0 +1,53 @@
...
+	"
+
+INSTALL()
+{
+	installDir="$prefix/non-packaged/bin"
```
korli, line 48: this is a bad idea

Begasus: hence the tag "help wanted" couldn't find an example on how to install there

korli: with a post-install script

Begasus: Thanks, when I get around to it I'll give it another go :)

sys-devel/fpcupdeluxe/fpcupdeluxe-2.2.0.recipe
Begasus: Updated the libQt5Pas recipe to use libVersion(Compat), this should be the next then?

```diff
- 	lib:libQt5pas$secondaryArchSuffix
+ 	lib:libQt5pas$secondaryArchSuffix >= 1
```

korli:
```suggestion
	lib:libQt5pas$secondaryArchSuffix >= 1.2.9
```

Begasus: Thanks, will push that change tomorrow :)

sys-devel/fpcupdeluxe/fpcupdeluxe-2.2.0.recipe
Begasus: Still haven't figured this out :(

Begasus:
OK, think the base should be fine now, any thoughts @korli ?

korli:
Looks fine @Begasus

Begasus:
Thanks, adding some rdef info into it and then it should be good to go (need a check on 64bit first)

sys-devel/fpcupdeluxe/additional-files/install_fpcupdeluxe.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+installDir="$(finddir B_USER_NONPACKAGED_DIRECTORY)/bin/"
+binaryDir="$(finddir B_SYSTEM_BIN_DIRECTORY)"
+mkdir -p $installDir
+
+cp -f "$binaryDir/fpcupdeluxe" "$installDir"
```
Begasus, line 7: @korli is there a way to "move" this from $binaryDir? Launching it from Terminal (will use $prefix/bin/fpcupdeluxe) won't work. Tried with `mv` instead of `cp -f`, but that doesn't actually move the binary.

Begasus:
Launch from Deskbar, not from Terminal :)

PR #6460, opened 2021-12-04
OIIO: bump

extrowerk:


media-libs/openimageio/openimageio-2.3.10.0.recipe
```diff
@@ -102,6 +102,7 @@ BUILD_REQUIRES="
 BUILD_PREREQUIRES="
 	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
+	cmd:git
```
korli, line 105: for tests, this would go to TEST_REQUIRES

extrowerk: Ah, true.

Begasus: Looks like it's needed here: https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/6900/builds/77092.log

PR #6472, opened 2021-12-14
portaudio: enable build for all architectures

TheZeldakatze:
I'm currently porting the game globulation2, which requires the portaudio library (or sys/soundcard.h). Portaudio is already being built for x86_gcc4 and compiles just fine under x86_64 and x86_32_gcc2, so I think it should be safe to enable on all architectures.

For some reason the patch file used only contains some configure stuff. There is a second patch file which contains a file which contains Haiku-specific code


TheZeldakatze:
> I'm currently porting the game globulation2, which requires the portaudio library (or sys/soundcard.h). 
(I just noticed that Haiku has soundcard.h)

waddlesplash:
We don't have soundcard.h, it's only there as a private header for the OSS drivers port. It isn't usable as such.

waddlesplash (changes requested):
The patch appears to be incomplete and will not actually work, so this shouldn't be enabled.

Begasus:
Superseded by https://github.com/haikuports/haikuports/pull/6494

PR #6473, opened 2021-12-15
globulation2: new recipe

TheZeldakatze:
I've changed out `#include <sys/soundcard.h>` with `#include <private/audio/soundcard.h>` as I don't know how else to replace it without using portaudio. If necessary though I'll have a closer look at the file and try to come up with a proper replacement.

Also I've disabled gl except if explicitly enabled with the argument `-g` as opengl makes the game lock up and flicker between the loading screen and the main menu (`-test-games` just makes it show a white screen when using gl). This seems to only happen on Haiku and after spending some time in the debugger, I still could not find the error, so I just disabled it.

The reason release is set to 0 is because the only thing it seems to change is adding the gcc `-O2` flag, but there seems to be an optimization bug which affects the saving of the config file,   which creates an approximatly 280mb configuration file before the game crashes. To make it run again you'd need to remove the config file so it doesn't try to parse a 280mb file. Disabling the optimization fixes this bug.

I used the patches from debian as a guideline to fix the compile errors, the scons files were updated with 2to3


games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,92 @@
...
+	lib:libSDL_image_1.2$secondaryArchSuffix
+	lib:libSDL_net_1.2$secondaryArchSuffix
+	lib:libspeex$secondaryArchSuffix
+	lib:libvorbis$secondaryArchSuffix
+	lib:libogg$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libfribidi$secondaryArchSuffix
+	lib:libglu$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libboost_thread$secondaryArchSuffix
```
Begasus, line 35: Sort alphabetically (also in BUILD_REQUIRES)

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,92 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku_devel$secondaryArchSuffix
+	devel:libSDL$secondaryArchSuffix
+	devel:libSDL_ttf$secondaryArchSuffix
+	devel:libSDL_image$secondaryArchSuffix
+	devel:libSDL_net$secondaryArchSuffix
```
Begasus, line 43: libSDL* needs versioning the same as in REQUIRES

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,92 @@
...
+	lib:libogg$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	lib:libfribidi$secondaryArchSuffix
+	lib:libglu$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libboost_thread$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku_devel$secondaryArchSuffix
```
Begasus, line 39: Change to:
```diff
- 	haiku_devel$secondaryArchSuffix
+ 	haiku${secondaryArchSuffix}_devel
```

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,92 @@
...
+	devel:libogg$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	devel:libfribidi$secondaryArchSuffix
+	devel:libglu$secondaryArchSuffix
+	boost170${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
+	cmd:scons
+	cmd:g++
```
Begasus, line 55: Needs secondaryArchSuffix (also sort alphabetically)
```diff
-	cmd:gcc
	cmd:scons
-	cmd:g++
+	cmd:gcc$secondaryArchSuffix
	cmd:scons
+	cmd:g++$secondaryArchSuffix
```

Begasus (changes requested):

Begasus:
Tried to build it on 32bit but it doesn't seem to find any of the libraries needed?

TheZeldakatze:
> Tried to build it on 32bit but it doesn't seem to find any of the libraries needed?

I somehow managed to screw up the suff with the $secondaryArchSuffix.

After correctly setting up secondary_arch in haikuporter, I'll try to compile it on x86 in a virtual machine.

TheZeldakatze:
> scons is build with gcc2, so no $secondaryArchSuffix here (see the recipe for scons, it's not listed as `scons$secondaryArchSuffix` in `PROVIDES` there) :)

just noticed that aswell, I should maybe check before I commit and push. I've just pushed another commit with the correct names (hopefully) and I'll try to compile it on x86_32

Begasus (approved):

kenmays:
@TheZeldakatze - Built and tested x86_64 build. No major issues.
Suggestion: For recipe, disable 32-bit build and restrict boost >=1.69.

TheZeldakatze:
It looks like scons searches the gcc2_x86 part instead of the x86_gcc4 one. Because after installing the speex_devel (x86_gcc2) package the error that it could not find the library disappeared

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,91 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="http://mirror.netcologne.de/savannah/glob2/0.9.4/glob2-$portVersion.tar.gz"
+SOURCE_FILENAME="glob2-$portVersion.tar.gz"
+SOURCE_DIR="glob2-$portVersion"
+CHECKSUM_SHA256="0f4d898ec6b05ce27b4a12ef242cc26571304b90d2509932a4743c71311314b8"
+PATCHES="globulation2-0.9.4.4.patchset"
+ADDITIONAL_FILES="globulation2.rdef.in"
+
+ARCHITECTURES="all !x86_gcc2 !x86"
```
Begasus, line 15: Changes:
```diff
- ARCHITECTURES="all !x86_gcc2"
+ SECONDARY_ARCHITECTURES="x86"
```

Begasus:
> It looks like scons searches the gcc2_x86 part instead of the x86_gcc4 one.

This should be fixed by adding SECONDARY_ARCHITECTURES

games-strategy/globulation2/globulation2-0.9.4.4.recipe
TheZeldakatze: I actually did do it like that previosly, it still looked for the libraries in the x86_gcc2 parts

Begasus: Although it still doesn't work the change would eliminate a build for 32bit (x86_gcc2 primary and x86 secondary architecture)

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,91 @@
...
+
+PROVIDES="
+	globulation2$secondaryArchSuffix = $portVersion
+	app:globulation2
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
```
Begasus, line 22: Should need libboost_date_time too (looking at what it's looking for?
```diff
+ 	lib:libboost_date_time$secondaryArchSuffix
```

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,91 @@
...
+	lib:libSDL_image_1.2$secondaryArchSuffix
+	lib:libSDL_net_1.2$secondaryArchSuffix
+	lib:libspeex$secondaryArchSuffix
+	lib:libvorbis$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	boost170${secondaryArchSuffix}_devel
```
Begasus, line 39: Replace:
```diff
- 	boost170${secondaryArchSuffix}_devel
+ 	lib:libboost_date_time$secondaryArchSuffix >= 1.70.0
+ 	lib:libboost_thread$secondaryArchSuffix >= 1.70.0
```

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -0,0 +1,91 @@
...
+	devel:libSDL_net_1.2$secondaryArchSuffix
+	devel:libspeex$secondaryArchSuffix
+	devel:libvorbis$secondaryArchSuffix
+	devel:libz$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:g++$secondaryArchSuffix
+	cmd:scons
+	python3$secondaryArchSuffix
```
Begasus, line 55: No secondaryArchSuffix here:
```diff
- 	python3$secondaryArchSuffix
+ 	python3
```

Begasus:
OK, checked on 64bit, same results here, can't find any libraries (not even if I change boost version to 1.69.0)

TheZeldakatze:
> OK, checked on 64bit, same results here, can't find any libraries (not even if I change boost version to 1.69.0)

In my testing, it did find all the dependencies, should I rebuild it and send a buildlog for comparison? Maybe it's because of package version mismatches

Begasus:
> > OK, checked on 64bit, same results here, can't find any libraries (not even if I change boost version to 1.69.0)
> 
> In my testing, it did find all the dependencies, should I rebuild it and send a buildlog for comparison? Maybe it's because of package version mismatches

The 64bit install here runs nightly, so packages there should be on par with the depot, but yeah, I could compare with the local ones here.

TheZeldakatze:
I finally found the reason for the x86 version failing to compile: scons has it's own path seperate from the system path. In it there are no x86 folders and it couldn't find the gcc executable. Adding the x86 folders to the search path seems to make  it work, at least the libraries are being found. I'll report back when the game is (hopefully) compiled.

TheZeldakatze:
Now the recipe should work, it builds fine on both of my machines for 32 and 64 bit, though someone should check if the build works for them aswell. I changed scons to always use the system path which should not break the build for other operating systems. I also don't understand where there would be  whitespace like lint reports, could someone help me with that?

Just in case this list is needed, these packages were used for compiling globulation on 64-bit, I put the list into a text file
[glob2_buildPackages.txt](https://github.com/haikuports/haikuports/files/8645464/glob2_buildPackages.txt)

games-strategy/globulation2/globulation2-0.9.4.4.recipe
```diff
@@ -86,6 +86,5 @@ INSTALL() {
 	
 	addResourcesToBinaries $portDir/globulation2.rdef $appsDir/Globulation2
 	
-	addAppDeskbarSymlink $appsDir/Globulation2
-	
-}
+	addAppDeskbarSymlink $appsDir/Globulation2
```
augiedoggie, line 89: There is an extra space at the end of the line

TheZeldakatze:
I removed the lines only containing whitespaces, now it works. I thought that only the last whitespace would matter, my bad

kenmays:
@TheZeldakatze - Builds fine on Haiku x86 (R1B3).

PR #6490, opened 2021-12-22
sdlinvaders: fix recipe

TheZeldakatze:
I fixed the recipe and thus incremented the revision counter by 1


TheZeldakatze:
The pr seems to contain all prior commits aswell, is that a problem or fine?

Begasus:
> The pr seems to contain all prior commits aswell, is that a problem or fine?

Looks OK, I'll do a check build tomorrow and let you know (if no one beats me to it)

TheZeldakatze:
> > The pr seems to contain all prior commits aswell, is that a problem or fine?
> 
> Looks OK, I'll do a check build tomorrow and let you know (if no one beats me to it)

ok, thank you
And sorry for not properly testing the recipe last time

kenmays:
@TheZeldakatze - Built and tested on Haiku R1B3 x86 (GCC2). No issues.

Begasus:
in `main.c` change fullscreen to 0 (line 56), otherwise it changes the screenresolution, also when I launch it close it (with q) it crashes.

Begasus:
At the moment highscores are saved into /boot/system/settings/sdlinvaders, could you change this to ~/config/settings/sdlinvaders? (this are user settings and imho should go to the home config directory rather then the system one.
PS, also checked a build on 64bit, runs fine, thanks!

pulkomandy:
It makes sense to save highscorts in a shared directory. Then multiple users on the same system can compare their scores

TheZeldakatze:
> It makes sense to save highscorts in a shared directory. Then multiple users on the same system can compare their scores

> At the moment highscores are saved into /boot/system/settings/sdlinvaders, could you change this to ~/config/settings/sdlinvaders? (this are user settings and imho should go to the home config directory rather then the system one.
> PS, also checked a build on 64bit, runs fine, thanks!

I used the system wide directory because it's the same directory used by tecnoballz for its highscore file, but it would be easy to change, should I?

games-arcade/sdlinvaders/sdlinvaders-0.7.6.recipe
```diff
@@ -42,13 +42,7 @@ BUILD_PREREQUIRES="
 	"
 	
 BUILD() {
-	if [ $targetArchitecture == "x86"  ] ||  [ $targetArchitecture == "x86_gcc2" ]; then
-		# for some reason, the x86 build doesn't like clang
-		make
-	else
-		# use clang instead of gcc, as gcc fails to compile
-		make CC=clang
```
waddlesplash, line 50: If we're now build with GCC you may as well remove cmd:clang from BUILD_PREREQUIRES.

TheZeldakatze: oh, I somehow forgot to remove that, sorry

PR #6510, opened 2022-01-02
poppler-data, bump version, remove old

Begasus:


Begasus:
wip, don't merge just yet, as discussed on IRC the .pc file shouldn't go into $prefix/lib but rather to $dataDir

PR #6528, opened 2022-01-09
avisynthplus: update to 3.7.1

qyot27:
3.7.1 was released a little over a week ago, a bit overdue.  The patches that were needed for 3.7.0 are not needed for 3.7.1, and since the build process complained about CMAKE_BUILD_TYPE not being set, do so.

The extra arches just bring it mostly into line with what we have internally in `avs/config.h` ([here](https://github.com/AviSynth/AviSynthPlus/blob/master/avs_core/include/avs/config.h#L47)).  ppc64 is the missing variant there, but I didn't add it since that's not listed under ARCHITECTURES for any other recipe.


diversys:
You can drop the patch files if they aren't needed anymore.

qyot27:
Done.  I'd removed them from the recipe but forgotten to delete them.

PR #6532, opened 2022-01-11
Add libphysfs to REQUIRES list for raceintospace

rodneyrod:
physfs is required to run this application however it doesn't get pulled in as a dependency in pkgman.
This resolves that issue by adding lib:libphysfs$secondaryArchSuffix to the REQUIRES list.


waddlesplash:
You will need to bump the REVISION in order for the package to be rebuilt.

And, welcome to HaikuPorts. :)

rodneyrod:
Cheers think that's done now, and thanks always happy to help.

korli:
lib:libphysfs comes before lib:libpng

PR #6534, opened 2022-01-12
 fontconfig: bump version to 2.13.94

kenmays:
Tested on Haiku R1B3 w/GCC2 and GCC11.


kenmays:
Outdated. May move to 2.13.96, if needed.

Issue #3749, opened 2019-04-07
LLVM can't find its libs


extrowerk:
```
CMake Error at /boot/system/lib/cmake/llvm/LLVMExports.cmake:1255 (message):
  The imported target "LLVMDemangle" references the file

     "/boot/system/lib/libLLVMDemangle.a"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.
```

It is in /boot/system/develop/lib
The cmake files should be adjusted.

Begasus:
Still valid (checked with llvm7 and llvm9)

Begasus:
Had the same problem trying to get this to work for Sourcetrail ...
![Sourcetrail2](https://user-images.githubusercontent.com/16057090/72350623-166bec80-36df-11ea-9b3f-2ef3ccb144e6.png)

waddlesplash:
cc @korli. Ran into this while trying to use the new Cmake build system for qt creator.

augiedoggie:
I did start toying around with a possible fix for this a long time ago but got lazy and never actually tested it.  I'm not sure if there's a preferred way to handle this but I was thinking of just using a sed command in the recipe like ...

```diff
diff --git a/sys-devel/llvm/llvm9-9.0.1.recipe b/sys-devel/llvm/llvm9-9.0.1.recipe
index c4e0bab05..94e97e1e5 100644
--- a/sys-devel/llvm/llvm9-9.0.1.recipe
+++ b/sys-devel/llvm/llvm9-9.0.1.recipe
@@ -31,7 +31,7 @@ other than the ones listed above.
 HOMEPAGE="https://www.llvm.org/"
 COPYRIGHT="2003-2019 University of Illinois at Urbana-Champaign"
 LICENSE="Apache v2 with LLVM Exception"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://github.com/llvm/llvm-project/releases/download/llvmorg-$portVersion/llvm-$portVersion.src.tar.xz"
 CHECKSUM_SHA256="00a1ee1f389f81e9979f3a640a01c431b3021de0d42278f6508391a2f0b81c9a"
 SOURCE_DIR="llvm-$portVersion.src"
@@ -549,6 +549,10 @@ INSTALL()
 		liblld* \
 		libRemarks
 
+	# cmake library paths need to be adjusted after prepareInstalledDevelLibs
[...]

PR #6536, opened 2022-01-14
llvm12: adjust cmake library paths

augiedoggie:
This is related to issue #3749. The fixed packages were tested on x86_64 by
building a couple of language servers(ccls and irony-server) which use these
library paths.

### DIFF
--- a/sys-devel/llvm/llvm12-12.0.0.recipe
+++ b/sys-devel/llvm/llvm12-12.0.0.recipe
@@ -31,7 +31,7 @@ other than the ones listed above.
 HOMEPAGE="https://www.llvm.org/"
 COPYRIGHT="2003-2019 University of Illinois at Urbana-Champaign"
 LICENSE="Apache v2 with LLVM Exception"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/llvm/llvm-project/releases/download/llvmorg-$portVersion/llvm-$portVersion.src.tar.xz"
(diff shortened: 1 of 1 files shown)


korli:
Remark: such a workaround can't be upstreamed.

sys-devel/llvm/llvm12-12.0.0.recipe
```diff
@@ -551,6 +551,11 @@ INSTALL()
 		liblld* \
 		libRemarks
 
+	# adjust cmake library paths because prepareInstalledDevelLibs moved them
+	sed -i 's,\${_IMPORT_PREFIX}/lib/,\${_IMPORT_PREFIX}/develop/lib/,' \
```
Begasus, line 555: I "think" this won't work on 32bit (didn't check there yet) because the libraries are installed in */develop/lib/x86 ?
If so a check for targetarchitecture would be needed?

augiedoggie: Probably.  I haven't looked at the package on 32 bit but I would guess that it's still broken.

augiedoggie: > I "think" this won't work on 32bit (didn't check there yet) because the libraries are installed in */develop/lib/x86 ? If so a check for targetarchitecture would be needed?

I finally got around to looking at this so I could extend the workaround for x86 but it isn't needed.  The paths in the files are correct and I was able to use it to compile some other software.  The only immediate change that I need to make is that I missed one of the files.  I need to add $libDir/cmake/lld/LLDTargets-release.cmake

PR #6537, opened 2022-01-14
freetype: enabled WOFF 2.0

kenmays:
WOFF File Format 2.0 - The WOFF 2.0 specification is implemented in all major browsers, and is widely used on production websites. It supports the entirety of the TrueType and OpenType specifications, including Variable fonts, Chromatic fonts, and font Collections.

### DIFF
--- a/media-libs/freetype/freetype-2.11.0.recipe
+++ b/media-libs/freetype/freetype-2.11.0.recipe
@@ -16,7 +16,7 @@ PATCHES="
 #PATCHES="freetype-subpixel.patchset" # enable patented subpixel rendering
 
 ARCHITECTURES="all"
-SECONDARY_ARCHITECTURES="x86"
+SECONDARY_ARCHITECTURES="x86_gcc2 x86"
 
 PROVIDES="
 	freetype$secondaryArchSuffix = $portVersion
--- a/media-libs/freetype/freetype-2.11.0.recipe
+++ b/media-libs/freetype/freetype-2.11.0.recipe
@@ -24,10 +24,10 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
+	lib:libbrotlidec$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


media-libs/freetype/freetype-2.11.0.recipe
```diff
@@ -16,7 +16,7 @@ PATCHES="
 #PATCHES="freetype-subpixel.patchset" # enable patented subpixel rendering
 
 ARCHITECTURES="all"
-SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
```
Begasus, line 19: Not needed atm, there is no x86_gcc2 for secondary architecture (yet), but there are other recipes still mentioning it)

media-libs/freetype/freetype-2.11.0.recipe
```diff
@@ -75,7 +78,7 @@ defineDebugInfoPackage freetype$secondaryArchSuffix \
 BUILD()
 {
 	unset meson
-	meson setup build . --buildtype release \
+	meson setup build . --buildtype release  \
```
Begasus, line 81: Extra space added? Shouldn't be needed also.

Begasus (changes requested):

media-libs/freetype/freetype-2.11.0.recipe
kenmays: Fixed

korli:
Why would freetype need to encode with brotli?

kenmays:
Moreso, to reduce web font/text latency when fonts are transferred from a web server to a web client over a network connection.

waddlesplash:
I think we should disable this. There's not a lot of point in decoding WOFF2 directly in FreeType; WebKit decodes WOFF(2) to TTF before handing it off to the rendering engines anyway, so it can handle brotli decoding. Otherwise this will introduce another dependency to the base system.

PR #6540, opened 2022-01-16
llvm12: update to 12.0.1 and add missing provides

augiedoggie:
The packages now build cleanly with 'haikuporter --strict-policy'.  Built and tested on x86_64 only.


augiedoggie:
Test suite now executes on x86_64(nightly) and x86(B3).  I have left the target as "check-llvm" for now, but there is a "check-all" ninja target which should execute the clang test suite in addition.

korli:
Looks good

Issue #6541, opened 2022-01-16
7kaa: duplicate data folder


humdingerb:
There seems to be a duplicate of the data folder packaged in 7kaa:

![duplicated_data_dir](https://user-images.githubusercontent.com/6107158/149655844-ca7b7a9a-688b-45e8-9f41-8bfdb53532e8.png)

Maybe @TheZeldakatze  can have another look?

Begasus:
Nice catch! Looking into it also, atleast the problem with curl can be fixed with adding "cmd:pkg_config$secondaryArchSuffix" in BUILD_PREREQUIRES, it also checks for libiconv and libintl (gettext) so might as well add them where needed ...

Begasus:
Even nicer with music :)

TheZeldakatze:
> cmd:pkg_config$secondaryArchSuffix

I somehow managed to copy the data folder twice, because I excpected make to not put it into the correct path and not checking it afterwards. I also found a way to make the game run in window mode, you have to create a config file for that. I'll try to create it automatically with a post install script.

Begasus:
Nice, I'll add a recipe for the music files then ;)

Begasus:
PR created at: https://github.com/haikuports/haikuports/pull/6542
If you want to check it just add **7kaa_music** to REQUIRES (after a local build if it's not merged yet) :)

TheZeldakatze:
> PR created at: #6542
> If you want to check it just add **7kaa_music** to REQUIRES (after a local build if it's not merged yet) :)

I finally figured out how to run the post install scripts. (I constantly installed the wrong package until I noticed that I had changed the revision). Sorry for taking so much time because of that. The PR is #6544 . The game now also runs in windowed mode by default
and requires the music

Begasus:
Fixed in the meantime.

PR #6544, opened 2022-01-16
7kaa: improve/fix recipe

TheZeldakatze:
This should fix issue #6541 (duplicate data directory). It also creates a config file so the game runs in windowd mode instead of fullscreen which would clash with the system tray (I never moved my mouse there so I didn't notice before). It also now uses the music package from pull request #6542 . I'll test if everything builds correctly tomorrow.

### DIFF
--- a/games-strategy/7kaa/7kaa-2.15.5.recipe
+++ b/games-strategy/7kaa/7kaa-2.15.5.recipe
@@ -8,11 +8,15 @@ buildings with spies, or offering opponents money for their kingdom."
 HOMEPAGE="https://www.7kfans.com/"
 COPYRIGHT="Enlight Software Ltd."
 LICENSE="GNU GPL v2"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/the3dfxdude/7kaa/releases/download/v$portVersion/7kaa-$portVersion.tar.xz"
 CHECKSUM_SHA256="350a2681985feb4b71d20677d1a6d11921b9e3d97facbc94e4f020e848d8ab2b"
 PATCHES="7kaa-$portVersion.patchset"
-ADDITIONAL_FILES="7kaa.rdef.in"
+ADDITIONAL_FILES="7kaa.rdef.in
+	7kaaLauncher.sh.in
+	7kaaLauncherWrapper.c
+	yoshiLauncher
+	"
 
 ARCHITECTURES="all !x86_gcc2"
 SECONDARY_ARCHITECTURES="x86"
@@ -20,6 +24,7 @@ SECONDARY_ARCHITECTURES="x86"
 PROVIDES="
 	7kaa$secondaryArchSuffix
 	app:SevenKingdoms
+	cmd:7kaa$secondaryArchSuffix
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
@@ -27,6 +32,7 @@ REQUIRES="
 	lib:libenet$secondaryArchSuffix
 	lib:libopenal$secondaryArchSuffix
 	lib:libSDL2_2.0$secondaryArchSuffix
+	7kaa_music
 	"
 
 BUILD_REQUIRES="
@@ -75,12 +81,28 @@ INSTALL() {
 
 	mkdir -p $appsDir
 	mkdir -p $dataDir
+	mkdir -p $dataDir/7kaa
 
-	mv data/ $dataDir/7kaa
+	# install the launcher
+	cp $portDir/additional-files/yoshiLauncher $dataDir/7kaa/
 
-	mv $binDir/7kaa $appsDir/"Seven Kingdoms"
+	# use the correct executable in the launcher
+	local archSuffix=""
+	if [ ! -z $secondaryArchSuffix ]
+	then
+		archSuffix="-`echo "$secondaryArchSuffix" | cut -c 2-`"
+	fi
+	sed -e "s|@archSuffix@|$archSuffix|" \
+		$portDir/additional-files/7kaaLauncher.sh.in > $dataDir/7kaa/7kaaLauncher.sh
+
+	chmod +x $dataDir/7kaa/7kaaLauncher.sh
+
+	# compile the launcher
+	gcc -o $appsDir/"Seven Kingdoms" $portDir/additional-files/7kaaLauncherWrapper.c
+	chmod +x $appsDir/"Seven Kingdoms"
 
 	addResourcesToBinaries $portDir/7kaa.rdef $appsDir/"Seven Kingdoms"
+	addResourcesToBinaries $portDir/7kaa.rdef $binDir/7kaa
 	addAppDeskbarSymlink $appsDir/"Seven Kingdoms"
 
 }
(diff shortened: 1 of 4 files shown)


games-strategy/7kaa/7kaa-2.15.4p1.recipe
```diff
@@ -36,6 +38,7 @@ BUILD_REQUIRES="
 	devel:libcurl$secondaryArchSuffix
 	devel:libSDL2_2.0$secondaryArchSuffix
 	devel:libenet$secondaryArchSuffix
+	7kaa_music
```
korli, line 41: The music is a runtime requirement, not a build requirement, correct?

games-strategy/7kaa/7kaa-2.15.4p1.recipe
```diff
@@ -76,7 +79,10 @@ INSTALL() {
 	mkdir -p $appsDir
 	mkdir -p $dataDir
 	
-	mv data/ $dataDir/7kaa
+	
+	mkdir -p $postInstallDir
+	cp createHaikuConfig.sh $portDir/7kaa_postinstall.sh
+	install -t $postInstallDir -m 755 $portDir/7kaa_postinstall.sh
```
korli, line 85: the usual way of doing this is with:
`USER_SETTINGS_FILES="
        settings/kyua directory
        settings/kyua/kyua.conf template $dataDir/kyua/examples/kyua.conf"`

games-strategy/7kaa/7kaa-2.15.4p1.recipe
TheZeldakatze: I somehow manged to mess that up. I was rather tired yesterday

games-strategy/7kaa/7kaa-2.15.4p1.recipe
TheZeldakatze: ok, I'll replace that later

TheZeldakatze: the template is not automatically installed on pacakage installation. So I just re-added the post install script for now. Though this would probably clash if there would ever be multiuser support. Do you know how to do this without a script?

korli: Yeah sorry for that. USER_SETTINGS_FILES is a declaration with no automatic actions. GLOBAL_WRITABLE_FILES is what should be used for system settings files, not for user settings. 
The usual way to do this is in a launcher: the user runs the launcher, the config is copied if needed, the app is launched. With multiple users, there is no mechanism which would copy a template config to each possible user settings directory.

Begasus:
Still needs some work, not a lot of time in the next days, so will put a wip label on this for now, one thing (as mentioned earlier) add "cmd:pkg_config$secondaryArchSuffix" in BUILD_PREREQUIRES (this solves the check for curl) also add libintl and libiconv in the right places (you'll see configure checking for them in the process) ...

games-strategy/7kaa/7kaa-2.15.4p1.recipe
TheZeldakatze: I now used the system() function to call the script on start and it seems to work

And it seems to not overwrite a modified config

TheZeldakatze:
> Still needs some work, not a lot of time in the next days, so will put a wip label on this for now, one thing (as mentioned earlier) add "cmd:pkg_config$secondaryArchSuffix" in BUILD_PREREQUIRES (this solves the check for curl) also add libintl and libiconv in the right places (you'll see configure checking for them in the process) ...

I somehow did forgot to do that, though I was sure I'd have done that

games-strategy/7kaa/7kaa-2.15.4p1.recipe
TheZeldakatze: > Yeah sorry for that. USER_SETTINGS_FILES is a declaration with no automatic actions. GLOBAL_WRITABLE_FILES is what should be used for system settings files, not for user settings.
> The usual way to do this is in a launcher: the user runs the launcher, the config is copied if needed, the app is launched. With multiple users, there is no mechanism which would copy a template config to each possible user settings directory.

no problem, I could've also just read the documentation

TheZeldakatze:
I should have more time now so that I can finally fix those recipes. I now added libiconf and libintl (and accidentially pushed half of haikuports, which is the reason for the force-push).

Now that I think about it, the game automatically creating a configure file is maybe not a great idea, especially for those who want to play it fullscreen as editing the configs is rather unintuitive. Should I keep it or just remove it?

games-strategy/7kaa/7kaa-2.15.4p1.recipe
```diff
@@ -36,13 +45,16 @@ BUILD_REQUIRES="
 	devel:libcurl$secondaryArchSuffix
 	devel:libSDL2_2.0$secondaryArchSuffix
 	devel:libenet$secondaryArchSuffix
+	devel:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
```
korli, line 49: libintl shouldn't be required directly.

korli:
What's the alternative to the config file?

TheZeldakatze:
> What's the alternative to the config file?

I thought of some alternative but I have forgotten that. Maybe a launcher script that prompts the user about the resolution would make sense

But currently I have have to spend my time studying as I'll write my finals shortly. But after they are over I'll try to finish these PRs (this one, globulization and I'll try to submit a port of libpoco so mumble can be ported)

korli:
The question is more: why is it only a problem on Haiku? How does it go on other platforms?

TheZeldakatze:
> The question is more: why is it only a problem on Haiku? How does it go on other platforms?

There is no ingame setting so on other plattforms, it is just fullscreen. On Haiku the windowed fullscreen conflicts with the deskbar if autoraise is enabled

korli:
> > The question is more: why is it only a problem on Haiku? How does it go on other platforms?
> 
> There is no ingame setting so on other plattforms, it is just fullscreen. On Haiku the windowed fullscreen conflicts with the deskbar if autoraise is enabled

Could it be considered a Haiku bug?

TheZeldakatze:
> > > The question is more: why is it only a problem on Haiku? How does it go on other platforms?
> > 
> > 
> > There is no ingame setting so on other plattforms, it is just fullscreen. On Haiku the windowed fullscreen conflicts with the deskbar if autoraise is enabled
> 
> Could it be considered a Haiku bug?

As 7kaa uses a desktop full screen mode and the autoraise feature should rise to the top when the mouse is at the deskbar area I'd say it's kind of annoying but working as intended. One could turn off the autoraise if necessary.

TheZeldakatze:
I changed the package to now include a launcher made with bash and yoshi, run by a wrapper c program which has the rdef file applied to it. The launcher asks the user about the window resolution (it has to be 4:3, different aspect ratios just produce black bars) and if the game should be run in fullscreen. It then creates a config file and starts the game itself, which is now in the binDir folder. I'm currently checking if it runs on x86 as well, I'll report back shortly.

Begasus:
A new version is already available here, could you rebase on that one?

TheZeldakatze:
I've copied the launcher stuff to the new version. I couldn't test it yet, as I have some problem with my computers memory which makes gcc crash, so I'll report back tomorrow

Begasus:
Updated recipe: https://github.com/haikuports/haikuports/pull/9754 (closing this one), thanks!

PR #6553, opened 2022-01-22
fdupes: update to 2.1.2

augiedoggie:
custom fdupes license file was removed because it contained  the exact text of the MIT license.  I did do a `git mv` from the old recipe to the new but apparently enough of the file changed that git considers it a completely new file.


augiedoggie:
Pushed a new version because I forgot to add the line continuations when I changed the line wrapping to 100 columns.

PR #6572, opened 2022-01-25
Wine: add recipe

X547:


korli:
Nice, fast merge without review.

app-emulation/wine/wine-6.23.recipe
```diff
@@ -0,0 +1,89 @@
...
+	cmd:make
+	cmd:sed
+	cmd:flex
+	cmd:bison
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:mkdepend
+	cmd:gettext
+	llvm12
+	llvm12_clang
+	llvm12_lld
```
waddlesplash, line 75: Why do we need LLVM to build?

X547: Clang is needed to compile PE executables.

PR #6573, opened 2022-01-26
cdogs-sdl: update to 1.2.0

augiedoggie:
Needs an icon.  Sound and gamepad support are untested because I don't have them enabled in either of my VMs.

I have left the really old games-arcade/cyberdogs_sdl recipe in place for now.  I think it might have a level editor?


waddlesplash:
May as well just remove the old version, so people are not confused? We can package the editor later.

PR #6578, opened 2022-01-28
mercurial: update to 6.0 and switch to python3

augiedoggie:


dev-vcs/mercurial/patches/mercurial-6.0.patch
```diff
@@ -0,0 +1,28 @@
...
+--- a/contrib/python-zstandard/setup_zstd.py
++++ b/contrib/python-zstandard/setup_zstd.py
+@@ -178,8 +178,8 @@ def get_c_extension(
+         extra_args.append("-DZDICTLIB_VISIBILITY=")
+         extra_args.append("-DZSTDERRORLIB_VISIBILITY=")
+ 
+-        if compiler_type == "unix":
+-            extra_args.append("-fvisibility=hidden")
++#        if compiler_type == "unix":
++#            extra_args.append("-fvisibility=hidden")
```
waddlesplash, line 12: Why's this needed?

augiedoggie: It's an old gcc2 fix that I haven't removed from the patch

waddlesplash: We should remove it since we are not building with GCC2 anymore, and let the symbols get their proper visibility then.

augiedoggie: I'm having some git issues. I'm just going to open a new pull request.

PR #6580, opened 2022-01-28
attr: add recipe for 2.5.1

augiedoggie:


sys-apps/attr/attr-2.5.1.recipe
```diff
@@ -0,0 +1,86 @@
...
+GLOBAL_WRITABLE_FILES="
+	settings/xattr.conf keep-old
+	"
+
+PROVIDES="
+	attr = $portVersion
+	cmd:attr
+	cmd:getfattr
+	cmd:setfattr
+	lib:libattr
```
korli, line 23: version for the lib is missing

augiedoggie: I've added the versions and redone the patch to be a bit more proper.

PR #6581, opened 2022-01-28
ninja: fix build and add zsh subpackage

augiedoggie:
Maybe there is a better way to enforce CMAKE_BUILD_TYPE.  haikuporter gets very unhappy even if you use the command as a build/install tool.


augiedoggie:
I've noticed some other packages just put all of the shell completion files in the base package.  It seems like I had a discussion about this many years ago with someone and they wanted subpackages.  Should I get rid of the subpackage and include bash/zsh/etc.. in the base package?

korli:
Could we stop merging without actual review?

Issue #6584, opened 2022-02-01
fre:ac doesn't launch: symbol not found


humdingerb:
```
runtime_loader: /boot/system/lib/libsmooth-0.9.so.0: Could not resolve symbol 'libiconvctl'
resolve symbol "libiconvctl" returned: -2147478780
runtime_loader: /boot/system/lib/libsmooth-0.9.so.0: Troubles relocating: Symbol not found

```

This is on hrev55827, 64bit.
@enzo1982 , maybe you can have a look?

enzo1982:
smooth needs iconvctl/libiconvctl (which one is decided by the iconv.h system header). So far it didn't link to libiconv on Haiku, which usually means that iconvctl from the standard C library should be used.

However, the symbol smooth imported was libiconvctl, which usually means stand-alone libiconv.so. fre:ac still runs when I LD_PRELOAD libiconv.so.2.

I guess what happened is that libiconv was loaded by default previously (through a dependency in a system library) and is no longer loaded automatically now.

The solution is to re-build smooth, linking to libiconv. I'll adjust the recipe and the upstream sources.

humdingerb:
Works, thanks for the quick fix.
JFYI, If you add a "Fixes #6584" at the end of your commit message, github will automatically close that ticket.

PR #6585, opened 2022-02-01
smooth: Fix libiconv linkage

enzo1982:
See #6584


dev-libs/smooth/smooth-0.9.9.recipe
```diff
@@ -81,6 +81,7 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
+	export LDFLAGS="-liconv"
```
Begasus, line 84: Maybe this could be upstreamed here:

https://github.com/enzo1982/smooth/blob/master/Makefile#L169

or here?

https://github.com/enzo1982/smooth/blob/master/Makefile#L226

PR #6595, opened 2022-02-05
jo: add recipe for 1.6

augiedoggie:


app-text/jo/jo-1.6.recipe
```diff
@@ -0,0 +1,62 @@
...
+	cmd:autoreconf
+	cmd:awk
+	cmd:find
+	cmd:gcc
+	cmd:grep
+	cmd:install
+	cmd:make
+	cmd:pkg_config
+	cmd:sed
+	"
```
Begasus, line 39: Are all these needed? pkg-config for instance is only needed to find pandoc (which we don't have): https://github.com/jpmens/jo/blob/master/configure.ac#L44

augiedoggie: pkg-config can be removed but I believe the rest are needed.

augiedoggie: Actually, the build system requires pkg-config or it fails.  I can remove autoheader though.

Begasus: Haven't checked it (yet), thanks for looking into it :)

PR #6604, opened 2022-02-08
LibreOffice: fix building with newer bison

augiedoggie:
patch was taken directly from the upstream 7.2 branch of the libreoffice repo

I was able to successfully build, package, install, and run on x86_64.  I can also look into the x86 build issues but I'm not sure when I'll have the time.  Could be tomorrow or later in the week.


Begasus:
Had been planning a build for it yesterday but still kept hanging on "waiting for the build to activate", glad it's fixed (at least on 64bit) thanks!

Issue #6547, opened 2022-01-19
GIMP


waddlesplash:
We need a number of dependencies first:

 * [x] babl
 * [x] gegl
 * etc...

waddlesplash:
Perhaps @extrowerk or @Begasus can work on this one, now that there is a functioning GTK3 recipe? I added a template GIMP recipe in 10ad62ea9af76ab6527edf02fd452eed00613c4a.

extrowerk:
I can definetely spend some hours with this at the upcoming weekend, maybe earlier.

threedeyes:
I already have working recipes for these libraries

extrowerk:
> I already have working recipes for these libraries

I am always late of the show. Thanks!

threedeyes:
I made them a long time ago, I just did not publish them for lack of use

threedeyes:
babl, gegl, gexiv2 - done
i have appstream_glib recipe also, but it requered some broken libs (xinerama for example) for build

waddlesplash:
There really is no way to build appstream_glib without Xinerama? I can patch it, if so. Submit the recipe anyway even if it's disabled and I can take a look.

waddlesplash:
Anyone working on GIMP itself, then? If nobody else does, I can next week.

extrowerk:
> Anyone working on GIMP itself, then? If nobody else does, I can next week.

I can look into it. But each and every gtk sample program/ hello world crashes here on my side. Have to recheck.

waddlesplash:
Not even the ones included with the package? That may be due to some misconfiguration. Make sure you upgrade gdk_pixbuf to the latest version in case it isn't. Upload a crash report and I'll take a look. (Make sure you don't have libX11 installed but `xlibe` instead.)

waddlesplash:
The crashes may be fixed by https://github.com/haikuports/haikuports/commit/f366bf6118862812c481cd11227c95e1c15ffb52.

Begasus:
gegl needs to be fixed (as mentioned) but json_glib also (for secondary architecture on 32bit)

Begasus:
PR for appstream_glib and gtk3 have been filed, build for GIMP succeeds on 32bit but fails to launch:

```
~> GIMP 
libX11: unhandled Property: _NET_WM_PID<XA_CARDINAL>
libX11: unhandled Property: WM_CLIENT_LEADER<XA_WINDOW>
[...]

PR #6619, opened 2022-02-13
gtk3, enable 32bit

Begasus:


Begasus:
In preparation for GIMP on 32bit ... (builds OK)

Begasus:
As mentioned in https://github.com/haikuports/haikuports/issues/6547 GIMP builds ok but fails to launch, doing a check for inkscape (which needs gtkmm3) atm (32bit)

x11-libs/gtk3/gtk3-3.24.30.recipe
```diff
@@ -5,14 +5,14 @@ ranging from small one-off projects to complete application suites."
 HOMEPAGE="https://www.gtk.org"
 COPYRIGHT="GNOME Desktop"
 LICENSE="GNU LGPL v2"
-REVISION="5"
+REVISION="6"
```
waddlesplash, line 8: Not needed.

waddlesplash: Bumping the revision will cause rebuilds everywhere. Merely enabling the architecture suffices.

Begasus: Done

PR #6642, opened 2022-02-20
libzip: bump

extrowerk:


dev-libs/libzip/libzip-1.8.0.recipe
```diff
@@ -0,0 +1,110 @@
...
+	rm -rf $prefix/include
+
+	prepareInstalledDevelLibs libzip
+	fixPkgconfig
+
+	if [ -z "$secondaryArchSuffix" ]; then
+		maybe_manDir_man3=$manDir/man3
+	else
+		maybe_manDir_man3=
+		rm -rf $binDir
```
Begasus, line 96: This fixes the build for secondary architecture on 32bit:
```diff
- 		rm -rf $binDir
+ 		rm -rf $prefix/bin
```

korli:
the old recipe (1.2.0? ) should be removed.

PR #6661, opened 2022-02-27
NetCDF: bump

extrowerk:
base lib: bupm, lobotomized previous version
cxx: no new version available upstream
fortran: bump, nothing uses it yet, therefore no previous version kept.


sci-libs/netcdf/netcdf47-4.7.2.recipe
```diff
@@ -0,0 +1,90 @@
+SUMMARY="Scientific library and interface for array oriented data access"
```
waddlesplash, line 1: Why not name this recipe `netcdf4.7`?

Begasus: I don't think you can have a dot in a filename (filename should be the same as in PROVIDES also).
Same as the ones for eg python, lua ...

waddlesplash: You can, and we should here. I think PROVIDES can have dots as well.

Begasus: Stand corrected, works fine by renaming it to netcdf4.7 (tested on 32bit nightly)

PR #6677, opened 2022-02-28
x265: fix detection of cpu cores

eblanca:
Allow for x265 proper detection of actual number of cpu cores, instead of using default (2).

### DIFF
--- a/media-libs/x265/patches/x265-3.4.patchset
+++ b/media-libs/x265/patches/x265-3.4.patchset
@@ -108,7 +108,7 @@ index 0df90c8..e07fa91 100644
       * uses only one processor as the scheduler doesn't have enough work to utilize all processors */
      return sysconf(_SC_NPROCESSORS_CONF);
 -#elif __unix__
-+#elif __unix__ || __HAIKU__
++#elif defined(__unix__) || defined(__HAIKU__)
      return sysconf(_SC_NPROCESSORS_ONLN);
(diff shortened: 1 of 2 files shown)


extrowerk:
Looks good, but you have to bump the revision number in the recipe too. This signs for the build system: the package needs to be rebuilt. Just add +1 to the revision number.

media-libs/x265/patches/x265-3.4.patchset
```diff
@@ -91,3 +91,27 @@ index 2db7a14..0df90c8 100644
...
+diff --git a/source/common/threadpool.cpp b/source/common/threadpool.cpp
+index 0df90c8..e07fa91 100644
+--- a/source/common/threadpool.cpp
++++ b/source/common/threadpool.cpp
+@@ -636,7 +636,7 @@ int ThreadPool::getCpuCount()
+     /* Return the number of processors configured by OS. Because, most embedded linux distributions
+      * uses only one processor as the scheduler doesn't have enough work to utilize all processors */
+     return sysconf(_SC_NPROCESSORS_CONF);
+-#elif __unix__
++#elif __unix__ || __HAIKU__
```
korli, line 111:
```suggestion
+#elif defined(__unix__) || defined(__HAIKU__)
```

eblanca:
Ooops! Again, I missed to bump the version number for the recipe, but I learnt my lesson, guys - thank you!

PR #6680, opened 2022-03-01
YAB: bugs fix

lorglas:


dev-lang/yab/yab-1.8.1.recipe
```diff
@@ -5,22 +5,23 @@ HOMEPAGE="https://software.besly.de/"
...
-REVISION="3"
+REVISION="1"
 SOURCE_URI="https://github.com/lorglas/yab/archive/$portVersion.tar.gz"
-CHECKSUM_SHA256="947560212c2ab7b56da4ebd9b732afc6c4839adc7667c7ab13cfd2ef5c982a76"
+CHECKSUM_SHA256="abeeb6d5d525c8b06f28c78bec52bcac6688967ae27c612a9a13cbbe4b17c933"
 SOURCE_FILENAME="yab-$portVersion.tar.gz"
 SOURCE_DIR="yab-$portVersion"
 
-ARCHITECTURES="all !x86"
+ARCHITECTURES="any!x86"
```
korli, line 16: all was correct before

dev-lang/yab/yab-1.8.1.recipe
```diff
@@ -5,22 +5,23 @@ HOMEPAGE="https://software.besly.de/"
...
 
 PROVIDES="
 	yab = $portVersion
 	cmd:yab = $portVersion
 	lib:libyab = $portVersion
 	lib:libyab_1.7.8 = 1.7.8
 	lib:libyab_1.7.9 = 1.7.9
+	lib:libyab_1.8.0 = 1.8.0
```
korli, line 24: actually these three old libs are only available on x86_gcc2 and x86_64

dev-lang/yab/yab-1.8.1.recipe
Begasus: Given your next comment then x86_gcc2 and x86_64 should only be in?
```
 ARCHITECTURES="x86_gcc2 x86_64" 
```

korli: Not sure actually, the two ifs are only present because of binaries located at https://github.com/lorglas/yab/tree/master/oldlib64 and https://github.com/lorglas/yab/tree/master/oldlib32
But the project should otherwise build on all archs.
I'd recommend renaming oldlib64 to oldlib_x86_64, oldlib32 to oldlib_x86_gcc2, there are several 64bit arches coming soon ;)

dev-lang/yab/yab-1.8.1.recipe
Begasus: @lorglas are these "old" libraries still needed?

dev-lang/yab/yab-1.8.1.recipe
Begasus: OK, for one, you can't mix any with other architectures?
```
Error: /Opslag/haikuports/dev-lang/yab/work-x86-1.8.1/port.recipe specifies both 'any' and other architectures
```

pulkomandy: > If i type all by Architektur i can't use haikuporter -S because i get the message that the recipe is broken. with any it works

Make sure your version of haikuporter is up to date.

"any" and "all" are not the same thing.

"any" means a single hpkg file can be used for all architectures. For example a package that contains only data, or scripts. This is not the case here, the package contains .so files which will have to be compiled for each architecture.

"all" means the package can be built for all existing architectures. This is just a shortcut to not need to list them all. Which we did before, but soon it will look like "x86_gcc2 x86_64 sparc ppc arm arm64" and would need to be changed everytime we create a new architecture. So we now have the shorter "all", unless there is a special case for some architectures.

lorglas: @pulkomandy, thanks for clarification between any and all. That was not clear for me and my version of Haikuporter was not update. After updating, its working.

PR #6687, opened 2022-03-05
Harfbuzz: bump

extrowerk:
Possible soname change, but it seems no symbols got removed: https://github.com/harfbuzz/harfbuzz/releases


sikmir:
[5.1.0](https://github.com/harfbuzz/harfbuzz/releases) is the latest

PR #6692, opened 2022-03-06
npm, disable 32bit

Begasus:


Begasus:
npm needs nodejs which isn't available on 32bit

PR #6698, opened 2022-03-06
binutils: bump to binutils 2.38 for x86_64

kenmays:
Binutils 2.38 for Haiku x86_64:
- Cleaned out old and non-working recipes
- Tested Binutils 2.38 on Haiku x86 and x86_64
- Fixed 'hp --test' issue on Haiku x86 and x86_64
- Added legacy PE/COFF support for cross-platform development support


extrowerk:
This needs to be tested before merging. Patchset seems to be too short.

pulkomandy:
I think @kallisti5 has worked on upstreaming our changes to binutils, which would result in shorter patchsets.

extrowerk:
Can we enable gold?

PR #6699, opened 2022-03-07
Update optipngtranslator to 0.1.0

noryb009:
Update the OptiPNGTranslator to version 0.1.0

This release adds translations.

image


haiku-apps/optipngtranslator/optipngtranslator-0.1.0.recipe
```diff
@@ -6,12 +6,12 @@ be available in any native application. You find settings like the \
 optimization level in the DataTranslations preferences. The higher the \
 level, the longer it takes."
 HOMEPAGE="https://github.com/noryb009/OptiPNGTranslator"
-COPYRIGHT="2013-2018 Luke (noryb009)"
+COPYRIGHT="2013-2022 Luke (noryb009)"
 LICENSE="MIT"
-REVISION="5"
```
pulkomandy, line 11: when creating a new version (renaming the file) you can reset REVISION to 1 instead of incrementing it

haiku-apps/optipngtranslator/optipngtranslator-0.1.0.recipe
```diff
@@ -6,12 +6,12 @@ be available in any native application. You find settings like the \
...
-srcGitRev="22a9172ddd2035f1770ac0bd32e858d9d33cef27"
+REVISION="6"
+srcGitRev="b5b6f25e05992a062ebde02079a8ab73ecab5e0e"
```
pulkomandy, line 12: is there a git tag on the repository? That could be used here instead of a SHA1, allowing to more easily find which commit it is

pulkomandy (commented):
Hello, thanks for your update. I suggested two changes to make the recipe a bit cleaner, let us know what you think of it :)

haiku-apps/optipngtranslator/optipngtranslator-0.1.0.recipe
Begasus: Looking at the release there it is from the same tag, so you could use the source archive instead of the srcGitRev one: https://github.com/noryb009/OptiPNGTranslator/releases/tag/v0.1.0 ?

noryb009: I've updated this. I had to change the naming somewhat (for some reason the tar file Github gives don't include the `v`), let me know if there's a better way to work around that issue.

PR #6701, opened 2022-03-10
openal: Strip non-existent "include" directory

Cacodemon345:
This is intended to make linking against the `OpenAL::OpenAL` target work without errors. Previously attempting to link against the `OpenAL::OpenAL` target would cause an error in the generate step in CMake due to the configuration files referencing the "include" directory which doesn't exist in Haiku.


korli:
Applied fix from upstream in 1c339a1792b8e86e9fae4aab375efefd1a8a1be6

PR #6702, opened 2022-03-11
Wine: bump

extrowerk:
Patchset updated with all the latest patches from https://github.com/X547/wine
Seems to be working correctly. Tested with Irfanview and TotalCommander.


waddlesplash:
nice.

korli:
@extrowerk I suppose it shouldn't be installable on beta3.

PR #6707, opened 2022-03-13
lua: bump

extrowerk:


Begasus:
lua builds ok on 32bit, lua54 (gcc2) needs a fix for this: https://github.com/lua/lua/blob/fc6c74f1004b5f67d0503633ddb74174a3c8d6ad/luaconf.h#L668
lua54_x86 is fine without (so only needed for gcc2)
Added `&& !defined(__HAIKU__)` at the end of that line to fix the build, but maybe there is another (cleaner) sollution?

extrowerk:
@Begasus : could you please check?

Begasus:
Builds are ok now, thanks @extrowerk 
Penny thought, atm the commands are bundled in the base package, installing both version will conflict with the `lua` and `luac` symlinks installed, maybe move them to the `devel package` or move them into a separate `tool package with a conflict `then?

Begasus:
Rebuild caveexpress with lua54 works fine, running it doesn't require the cmd's, so I think it should be save to move the cmd's to the devel package (that already has a conflict stated)

augiedoggie:
> Builds are ok now, thanks @extrowerk Penny thought, atm the commands are bundled in the base package, installing both version will conflict with the `lua` and `luac` symlinks installed, maybe move them to the `devel package` or move them into a separate `tool package with a conflict `then?

Or we could do the same as with python where the commands get a version appended to the name.  My Arch Linux box has several versions of lua installed and they have names like `lua5.4` and `luac5.4`

Begasus:
> Or we could do the same as with python where the commands get a version appended to the name. My Arch Linux box has several versions of lua installed and they have names like `lua5.4` and `luac5.4`

We already have them, they are build by default, atm they are linked as lua and luac (looking into the recipe there's also liblua), maybe we should drop the symlinks?

PR #6709, opened 2022-03-13
zsh: bump

extrowerk:


Begasus:
This needs `util_linux` for `cmd:col` (so this conflicts with `libuuid` (needed for `gimp`)), also parse error on gcc2 ...

PR #6710, opened 2022-03-13
rtmidi: new recipe

Cacodemon345:
This ports the popular RtMidi library to Haiku.

### DIFF
--- a/media-libs/rtmidi/rtmidi-5.0.0.recipe
+++ b/media-libs/rtmidi/rtmidi-5.0.0.recipe
@@ -37,7 +37,7 @@ PROVIDES_devel="
 	"
 REQUIRES_devel="
 	haiku$secondaryArchSuffix
-	lib:librtmidi$secondaryArchSuffix == $portVersion
+	rtmidi$secondaryArchSuffix == $portVersion
 	"
 		
 BUILD_REQUIRES="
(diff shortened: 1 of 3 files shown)


media-libs/rtmidi/rtmidi-5.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	rtmidi${secondaryArchSuffix}_devel = $portVersion
+	devel:librtmidi$secondaryArchSuffix = $portVersion
+	"
+REQUIRES_devel="
+	haiku$secondaryArchSuffix
+	lib:librtmidi$secondaryArchSuffix == $portVersion
```
korli, line 40:
```suggestion
	rtmidi$secondaryArchSuffix == $portVersion
```

korli:
one comment in REQUIRES_devel, otherwise looks good.

korli:
@Cacodemon345  build failed on x86: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/2492/builds/77654.log

Cacodemon345:
The GCC2 compiler isn't supported since it requires C++11.

pulkomandy:
There is  a mix of int32 and int types.
On 64bit systems, int32 is defined as int. But on 32bit it is defined as long int.

Cacodemon345:
https://github.com/haikuports/haikuports/pull/6711
@pulkomandy @korli 

Pardon me for not testing the patch with 32-bit since I don't have extra media handy right now for that.

pulkomandy:
That's ok, but in that case you can mark it untested, for example:

    SECONDARY_ARCHITECTURES = "?x86"

Then someone can test and enable it later.

Cacodemon345:
I will do so next time, thanks.

PR #6714, opened 2022-03-14
pip: enable Python 3.10

extrowerk:
Somehow a pip2 materialized in the binfolder, i had no better idea than deleting it.

### DIFF
--- a/dev-python/pip/pip-19.3.1.recipe
+++ b/dev-python/pip/pip-19.3.1.recipe
@@ -3,7 +3,7 @@ DESCRIPTION="Install Python packages, replacement for easy_install"
 HOMEPAGE="https://pypi.python.org/pypi/pip"
 COPYRIGHT="2006-2019 Python Packaging Authority"
 LICENSE="Python"
-REVISION="2"
+REVISION="3"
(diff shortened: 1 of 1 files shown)


dev-python/pip/pip-19.3.1.recipe
```diff
@@ -20,8 +20,8 @@ BUILD_REQUIRES="
 	haiku_devel
 	"
 
-PYTHON_PACKAGES=(python python3 python38 python39)
-PYTHON_VERSIONS=(2.7 3.7 3.8 3.9)
+PYTHON_PACKAGES=(python python3 python38 python39 python310)
```
waddlesplash, line 23: Can we please use `python3.10` for python 3.10, at least, even if we can't for the other versions?

jmairboeck: Apparently, there is a `PACKAGE_NAME` recipe attribute which could be used for this in the future: https://github.com/haikuports/haikuporter/blob/03ed35d7e7bb6f0fb6607450dee7f795f66c1b84/HaikuPorter/RecipeAttributes.py#L187-L193

This seems to be implemented correctly, if I understood the code correctly, but support for it in the `packageEntries` shell scriptlet could be improved.

PR #6715, opened 2022-03-14
Xonsh: bump

extrowerk:
It triggers some bugs in Python3.7, but seems to be working okayinsh with Py3.10


korli:
It doesn't look like the builder picked it up...

PR #6718, opened 2022-03-14
php8: fix for 32 bit

TheZeldakatze:
I'm currently checking if it still builds on 64-bit. I'll report soon.


TheZeldakatze:
still compiles on x64

TheZeldakatze:
I just noticed I forgot to change the revision

PR #6723, opened 2022-03-15
Fontconfig 2.13.96

extrowerk:


extrowerk:
Needs to be checked on gcc2, i am nsure about the new Python3 dependency.

Begasus:
Getting a some parse errors in fcserialize.c with gcc2, first one starts here (not sure how to solve this one): https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/main/src/fcserialize.c#L123

Begasus:
Just checked a build for current version 2.14.0 which builds fine on secondary architecture, maybe add !x86_gcc2 into ARCHITECTURES? And keep (as it is atm) the older one for gcc2 builds?
Parse errors are the same for gcc2, and given the one above mentioned that I can't seem to fix, I'm guessing there will be plenty following after that one ...

PR #6724, opened 2022-03-15
JPEG: bump

extrowerk:


waddlesplash:
Actually we should get rid of this package entirely, and use `libjpeg-turbo` only, I think?

Begasus:
libjpeg-turbo has been in favor for a while now

extrowerk:
Still, we could have an up-to-date version of this in the ports tree, right?

Begasus:
I guess, I think when on linking time you'd have to see to use the correct libVersion (as it is done in most cases by now)

PR #6735, opened 2022-03-17
fluidsynth2: MidiKit2 support and MediaKit backend improvements

Cacodemon345:
The changes in question add a MidiKit2 backend to FluidSynth2.

I also got rid of the period size mangling code in the MediaKit backend since it doesn't actually seem to be needed and causes high latency issues in audio output and also added support for 32-bit float format.


pulkomandy:
There is already code wiring fluidsynth with the MIDI kit in Haiku side (using Fluidsynth to run the system MIDI synth).
Can you elaborate on what this new implementation adds?

Cacodemon345:
Is there anything else not understood? I already explained my reasons for adding MidiKit2 support.

pulkomandy:
It's ok for me. I let it here for some time to see if anyone else had comments, but it seems not. So let's merge it.

PR #6737, opened 2022-03-18
gcc-11.2: true up to buildtools repo; add missing arch ports

kallisti5:


kallisti5:
so.. people have been patching haikuports gcc and *not* buildtools gcc. This creates a problem since the patches here are supposed to reflect builtools gcc.

This patch aligns the haikuports gcc patch to match the gcc 11.2 patches in buildtools.  Some notable things missing in the existing haikuports patchset were the entire riscv64 and sparc architecture ports.

With that said, now we need to investigate differences going the other way.

For clairity, i've prepared this showing the changes to the gcc tree between the old and new gcc 11.2 patchsets.

https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L277


Here are things that need investigated:

https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L277
https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L612
https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L671
https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L841
https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L868
https://gist.github.com/kallisti5/7a088456bac66dbeefb849aaa7c90213#file-diff-diff-diff-diff-diff-L941
-        enable_libstdcxx_filesystem_ts=yes


I think most are arm related?

kallisti5:
@davidkaroly

extrowerk:
@kallisti5 :

> I personally think we should leave it disabled. why provide experimental namespace functions?

Afaik it is not experimental anymore and it is unconditionally enabled for most oerating systems: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/acinclude.m4#L4504

kallisti5:
> @kallisti5 :
> 
> > I personally think we should leave it disabled. why provide experimental namespace functions?
> 
> Afaik it is not experimental anymore (it is part of c++17, which is meant to be supported by gcc11) and it is unconditionally enabled for most operating systems: https://github.com/gcc-mirror/gcc/blob/master/libstdc%2B%2B-v3/acinclude.m4#L4504

ok.  If that's the case i'll re-add that bit to the new patchset.  the std::experimental namespace is throwing things off.

kallisti5:
Per https://review.haiku-os.org/c/buildtools/+/5114, i broke the Apple darwin arm fixes out into a separate commit since they're unrelated to the haiku port.

The fixes were also damaged during the gcc 11.2 import, so i theoretically repaired them.  5114 above should be monitored to see if the apple darwin fixes need updated differently.

PR #6760, opened 2022-03-26
Dav1d: bump, add gutted previous version for soname compatibility

extrowerk:


korli:
I usually name the recipe with the SONAME, because if it has to bumped, the version is in the name.

PR #6772, opened 2022-03-26
artpaint: bump version

kenmays:
Build tested on Haiku R1B3 x86_gcc2 and x64.


humdingerb:
I think we should wait for @dsizzle to fix that very annoing window resizing bug and make a new official point release of ArtPaint.

dsizzle:
yeah, I think it should only be a few more days on that issue.  Then a point release would be great!

Begasus:
> I think we should wait for @dsizzle to fix that very annoing window resizing bug and make a new official point release of ArtPaint.

Second that, this is still a wip and it would be good until it's ready for an official release.

PR #6781, opened 2022-03-28
pip, use python310 instead of python3.10

Begasus:
As is is done in: https://github.com/haikuports/haikuports/commit/53e2d007fef2b4bc8e2e8b9927b19460630092bd#diff-f01e5034d05b9fbfc6fb752eac65a49cd183598edd1dff81e9c77c36fa7fea0fR30


korli:
you could change in setuptools I suppose

korli:
Well it seems bash doesn't understand a dot in a variable name, so I don't see how to declare subpackages with dot in their names.

waddlesplash:
There's definitely a way to do it because environment variables can have . in them. I'll take a look.

PR #6783, opened 2022-03-28
pkgconfig, new python recipe

Begasus:


dev-python/pkgconfig/python_pkgconfig-1.5.5.recipe
```diff
@@ -0,0 +1,61 @@
...
+PYTHON_VERSIONS=(3.7 3.8 3.9 3.10)
+for i in "${!PYTHON_PACKAGES[@]}"; do
+pythonPackage=${PYTHON_PACKAGES[i]}
+pythonVersion=${PYTHON_VERSIONS[$i]}
+eval "PROVIDES_${pythonPackage}=\"\
+	${portName}_$pythonPackage = $portVersion\
+	\"; \
+REQUIRES_$pythonPackage=\"\
+	haiku\n\
+	cmd:pkg_config\n\
```
jmairboeck, line 35: After the discussion in #10737: I don't think that this requirement makes any sense here, because `cmd:pkg_config` is architecture-dependent (with `$secondaryArchSuffix`), and this is an "any" arch package. So anything that depends on this needs to depend on `cmd:pkg_config$secondaryArchSuffix` separately anyway (presuming that there are no "any" arch packages which provide pkg-config files and could be found with either pkg-config version).

Begasus: Makes sense now :) Mentions @OscarL

Begasus: I think @jmairboeck reply is the way to go, just nuke out "cmd:pkg_config" in here, no need to make python_pkgconfig arch aware.

jmairboeck: > Not sure that's what @jmairboeck is suggesting Begasus, unless I'm not following.

Yes, that is what I suggested. Having it like this is wrong for x86, and adding `cmd:pkg_config_x86` explicitly just for x86 is also wrong because then the package isn't architecture-independent any more.
 
However, I think the better solution would be to make this architecture-dependent then. The contents of the package would still be the same, but the metadata aren't.

In a way, the situation here is similar to alien_sdl, which I thought of making "any" but decided against it because it contains some configuration data with paths containing secondary arch suffixes, see https://github.com/haikuports/haikuports/blob/b2ff1616efbbd3ea0af1905d52155db82622d87b/dev-perl/alien_sdl/alien_sdl-1.446.recipe#L77-L78

Another alternative would be adding `SECONDARY_ARCHITECTURES="x86"` but keep "any", but as @pulkomandy wrote in https://github.com/haikuports/haikuports/pull/9436#discussion_r1623524876, "this doesn't make sense" (and he is right of course).

pulkomandy:
24 juillet 2024 à 19:40 "Joachim Mairböck" ***@***.***> a écrit:



> 
> Another alternative would be adding SECONDARY_ARCHITECTURES="x86" but keep "any", but as @pulkomandy https://github.com/pulkomandy  wrote in #9436 (comment) https://github.com/haikuports/haikuports/pull/9436#discussion_r1623524876 , "this doesn't make sense" (and he is right of course).
> 
Well, if it works for now, it could be good enough.

It's unfortunate that we can't get the paths on x86 to agree with the other architectures, but until we do, there aren't many good solutions.

So, this setup would be a bit surprising, but should be fine.

The only problem is that the "any" package will also be installable on x86_gcc2 systems, and I see no way to prevent that. Unless it has some missing dependencies that make it not built there, but that isn't very clean.

PR #6785, opened 2022-03-28
Betex modernized

Sylvain78:
-Preferences are now dynamic (not need to save to take them into account)
- About window doesn't crash anymore
- compiles on x86_64
- color_space corrected
- Go to line now working
- templates    

and many more i can't remember!


augiedoggie:
I don't like having a homepage that points to HaikuArchives while the source is downloaded from his own fork.  I personally think he should rename his fork to something else.

Begasus:
> I don't like having a homepage that points to HaikuArchives while the source is downloaded from his own fork. I personally think he should rename his fork to something else.

That hold me back, and I already pointed this out on IRC, the PR upstream only needs a few changes, once that is merged a new release could be made official there, and corrected here, for now this version is good enough and tested quite intense so for me it was reason enough to merge this one here atm.

PR #6800, opened 2022-03-31
BetterSpades: Add recipe and icon

win8linux:
Adds a recipe for [BetterSpades](https://github.com/xtreme8000/BetterSpades), an *Ace of Spades* 0.75 client made with low-end systems in mind. The game does work when ran directly from its app directory:
[](https://postimg.cc/T5V3BspN)

However, the game refuses to run from anywhere else (including the deskbar). I've attempted to symlink the `config.ini` file from somewhere else into the app dir, with no success. As far as I can tell, there is no way to directly tell the program to look for the config file elsewhere.


pulkomandy:
There are many ported software which are patched for this already.

You need to add something like this to the start of the main function:

    chdir(dirname(argv[0]));

This will make the game start in the directory of the executable. By default in Haiku, executables run from the home directory instead (unlike on other OS) unless run from Terminal.

games-action/betterspades/betterspades-0.1.6~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+
+INSTALL()
+{
+	mkdir -p $appsDir/BetterSpades
+	mkdir -p $settingsDir/BetterSpades
+	cp -r build/BetterSpades/* $appsDir/BetterSpades
+	mv $appsDir/BetterSpades/config.ini $settingsDir/BetterSpades/config.ini
+	mv $appsDir/BetterSpades/cache $settingsDir/BetterSpades/cache
+	rm -f $settingsDir/BetterSpades/cache/.gitignore
+
```
Begasus, line 79: Trailing whitespace here (strange that it doesn't show up on github)

games-action/betterspades/betterspades-0.1.6~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+DESCRIPTION="BetterSpades is an Ace of Spades client targeted at low end \
+systems (GL/ES 1.1). Runs on your grandmother's rig!
+
+* Replicates of the great game Ace of Spades (classic voxlap)
+* Runs on very old systems back to OpenGL 1.1 (OpenGL ES support too)
+* Shares similar if not even better performance to voxlap
+* Can run on "embedded" systems like a Steam Link
+* Open for future expansion
+* Easy to use
+"
```
Begasus, line 11: Move double quote to the end of previous line:
```diff
-* Easy to use
-"
+* Easy to use"
```

games-action/betterspades/betterspades-0.1.6~git.recipe
```diff
@@ -0,0 +1,94 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+srcGitRev="5438d7bbd0f6a6cd7565c5cc2118fc4935205a3c"
+SOURCE_URI="https://github.com/win8linux/BetterSpades/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="bf1ea77c57eb983149e51c961b126f50c5a89fadb1a58b76f031625c62bd1135"
+SOURCE_DIR="BetterSpades-$srcGitRev"
+ADDITIONAL_FILES="betterspades.rdef.in"
+GLOBAL_WRITABLE_FILES="
+	settings/BetterSpades/config.ini auto-merge
+	"
```
Begasus, line 23: GLOBAL_WRITABLE_FILES go between ARCHITECTURES and PROVIDES, see: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

Begasus:
Thoughts ... betterspades pulls quite a number of files (and the source for cglm) during the build, there is no way to do a checksum on those, should we consider this to go into haikuports? summons @korli @waddlesplash 

https://github.com/xtreme8000/BetterSpades/blob/e9bd4551e17ab53a4327fffffdb9d8cbcd4f1ecf/src/CMakeLists.txt#L36

games-action/betterspades/betterspades-0.1.6~git.recipe
win8linux: Weird that Pe didn't trim that, even though it is set to trim trailing whitespaces. Fixing it locally right now.

korli:
@Begasus, you can zip all these files once downloaded by the build, then keep this archive in additional-files, and unpack before calling cmake.

Begasus:
> @Begasus, you can zip all these files once downloaded by the build, then keep this archive in additional-files, and unpack before calling cmake.

That can be done, leaves the download from the cglm source, not sure if we can overcome that ... investigating as we go

win8linux:
> There are many ported software which are patched for this already.
> 
> You need to add something like this to the start of the main function:
> 
> ```
> chdir(dirname(argv[0]));
> ```
> 
> This will make the game start in the directory of the executable. By default in Haiku, executables run from the home directory instead (unlike on other OS) unless run from Terminal.

Thanks for the suggestion @pulkomandy, but unfortunately it didn't resolve the problem. The game still tries to load the config from the same place.  Perhaps it works better somewhere else in the source?

Begasus:
Think the patchset can be dropped and reference in the recipe removed?

win8linux:
Not quite, since the game still doesn't launch from DeskBar. Still haven't figured out where to put `chdir(dirname(argv[0]));` in, as it doesn't work at the start of the main function (as previously suggested). Might consider closing this PR instead.

Begasus:
> Not quite,

From looking at the srcGitRev it takes the upstream patch already, atm the patch is already commented out, so I think "that" could be left out.



> Still haven't figured out where to put `chdir(dirname(argv[0]));` in, as it doesn't work at the start of the main function (as previously suggested). Might consider closing this PR instead.

That is another issue and can be looked at later. :) Don't close yet.

Begasus:
Pushed changes for the downloaded files and added a fix to launch from Tracker/Deskbar, thanks @win8linux for the PR!

PR #6802, opened 2022-03-31
tmux: Fix suffix on x86_gcc2

t6:
It should be available as `tmux` not `tmux-x86`.


app-misc/tmux/tmux-3.1c.recipe
```diff
@@ -42,7 +49,7 @@ BUILD_PREREQUIRES="
 BUILD()
 {
 	autoreconf -fi
-	runConfigure ./configure
+	runConfigure ./configure --bindir=$commandBinDir
```
augiedoggie, line 52: should be:
```
runConfigure --omit-dirs binDir ./configure --bindir="$commandBinDir"
```

according to https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#using-the-commandbindir-user-variable

t6: Fixed. Thanks.

PR #6803, opened 2022-03-31
kakoune: some improvements

t6:
- add debuginfo package
- set B_BACKGROUND_APP to avoid it briefly flashing up at least once in the Deskbar whenever it is launched.


Begasus:
According to Arch kak-lsp is an "optional" dependency for this, did you try that?

t6:
Yes, I tried kakoune+kak-lsp+clangd together. Everything seems fine.

Can I add optional or suggested dependencies to the package somehow?

Begasus:
> Yes, I tried kakoune+kak-lsp+clangd together. Everything seems fine.
> 
> Can I add optional or suggested dependencies to the package somehow?

Haven't checked a build yet, but I guess you could add `cmd:kak_lsp` in `REQUIRES` here (or in `BUILD_PREREQUIRES`), need to check that too :)

PR #6812, opened 2022-04-03
mtxclient, bump version, coeurl new recipe

extrowerk:


dev-libs/mtxclient/mtxclient-0.7.0.recipe
```diff
@@ -25,8 +25,12 @@ REQUIRES="
 	lib:libboost_iostreams$secondaryArchSuffix
 	lib:libboost_system$secondaryArchSuffix
 	lib:libboost_thread$secondaryArchSuffix
```
deepbluev7, line 27: Why is boost still part of the dependencies?

Begasus: Was checking the source on github when I saw you removed it from the dependencies, should indeed be removed.

Begasus:
According to nephele the patchset should be re-enabled for disabling QSingleApplication, atm conflicts as nheko has been disabled (crashes reported by nephel)

PR #6818, opened 2022-04-09
texlive: update to 2022

jmairboeck:
just texlive_core for now, I will add texlive (the texmf part) next.

I just renamed the patchset from 2021 and it still applied cleanly (which is a bit surprising). According to [the high-level changelog](https://tug.org/texlive/doc/texlive-en/texlive-en.html#news), it could be that the "SELFAUTO" patch is not actually needed any more, but we will only see that after the upgrade is finished and this is tested. It will probably not be harmful though, so we can just keep it anyway, I think.

I tested the build on x86_64 (nightly). Could someone please check x86?


app-text/texlive/texlive-2022.recipe
```diff
@@ -195,7 +195,7 @@ GLOBAL_WRITABLE_FILES="
 	"
 
 PATCH() {
-	sed -i "s|\`kpsewhich -var-value=SELFAUTOPARENT\`|\"$dataDir/texlive\"|" $sourceDir/tlpkg/TeXLive/TLUtils.pm
+	sed -i "s|\`kpsewhich -var-value=TEXMFROOT\`|\"$dataDir/texlive\"|" $sourceDir/tlpkg/TeXLive/TLUtils.pm
```
jmairboeck, line 198: This is apparently what was meant with the change from SELFAUTOPARENT to TEXMFROOT, not the actual code in kpsewhich.

jmairboeck:
I noticed that the linting CI job failed because it was tried to find a package called "subpackageInfos" which doesn't exist obviously. I used the `.recipe` file extension on it to get correct synatax highlighting and because it is an actual recipe snippet (not a full recipe on its own). Should I rename the file regardless, or could the linter be changed so that it ignores files in additional-files, or should we just ignore it?

Begasus:
I think it's safe to ignore for now

Begasus:
Missing dependencies for lib:libbrotlydec and lib:libbz2 in REQUIRES

jmairboeck:
shouldn't there be a policy check for missing dependencies? I didn't get any warnings besides some missing cmd: PROVIDES, which I have added.

Begasus:
> shouldn't there be a policy check for missing dependencies? I didn't get any warnings besides some missing cmd: PROVIDES, which I have added.

I got POLICY ERRORS on them for 3 cmd in the PROVIDES section (running haikuporter in strict mode)

Begasus:
This fixes the freetype detection here: https://gitweb.gentoo.org/repo/gentoo.git/tree/app-text/texlive-core/texlive-core-2021-r1.ebuild#n183

Begasus:
PS, latest version atm is 20220403 ;)

jmairboeck:
> This fixes the freetype detection here: https://gitweb.gentoo.org/repo/gentoo.git/tree/app-text/texlive-core/texlive-core-2021-r1.ebuild#n183
freetype-config not being found seems not to be a problem.

> **EDIT** not sure if it's related to this, at first I got some undefined references (-lnetwork), adding `LDFLAGS="-lnetwork"` got pass that but now stuck with `undefined reference to `getpass'`, so not there yet ... (32bit nightly)

this means that the reautoconf script from the patch function didn't work correctly. It runs autoconf internally, but that only does something if the `configure` script isn't newer than the corresponding `configure.ac` file. The best way to fix this temporarily is to

1. make sure the patchset is applied
2. run `touch texk/{dvipdfm-x,web2c}/configure.ac` in the sourceDir (to make sure the patched configure.ac scripts are current)
3. run `reautoconf` again

jmairboeck:
>     2. run `touch texk/{dvipdfm-x,web2c}/configure.ac` in the sourceDir (to make sure the patched configure.ac scripts are current)

should i add that to the recipe before calling reautoconf in the patch function just to make sure? Right now, it only works when building "cleanly", i.e. when the patchset is applied before running the patch function.

jmairboeck:
> **EDIT** not sure if it's related to this, at first I got some undefined references (-lnetwork), adding `LDFLAGS="-lnetwork"` got pass that but now stuck with `undefined reference to `getpass'`, so not there yet ... (32bit nightly)

`getpass` is defined in `libbsd`, so `LDFLAGS="-lnetwork -lbsd" would also work probably.

jmairboeck:
Before merging this, I need to investigate why xetex doesn't find the DejaVu fonts even though they are installed (see https://github.com/haikuports/haikuports/pull/6836#issuecomment-1107855046). On MacOS, there are similar problems apparently: https://tex.stackexchange.com/a/426682/52759.

jmairboeck:
I found out what the problem was: xetex uses system fonts and finds them using fontconfig. For fontconfig to find the fonts, its cache needs to be current. The cache needs to be rebuilt with `fc-cache -f`.

PR #6823, opened 2022-04-11
Add recipe for MuPDF 1.19.1

ivanholmes:
Recipe for MuPDF 1.19.1 (latest version) providing libmupdf.so in addition to mutool and muraster linked to this dynamic library.
The older recipes alter the Makefile to cause it to be built as a shared object but recently this has become officially supported so the only patch I needed was to add Haiku detection to the Makerules file. I also switched to downloading the source as a tar.xz because it is about 20 megabytes smaller.

One potential issue I can see is that the API of libmupdf is not stable across versions, so really the .so file should have a version suffix but I am not sure how to implement this, perhaps someone more experienced could weigh in.

I have successfully built PyMuPDF based on this library and will contribute a recipe for that soon.


ivanholmes:
Apologies for messing up the commit here, I updated mine to remove trailing whitespace and tried to squash it back down into one. Hopefully someone more experienced can sort this out without too much headache!

ivanholmes:
I've cleared up the mess now. Hopefully this is okay to merge, let me know...

ivanholmes:
It should be noted that this breaks pdftranslator because it expects libmupdf 1.8 or thereabouts which is pretty old by now.

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,110 @@
...
+{
+	make USE_SYSTEM_FREETYPE=yes USE_SYSTEM_HARFBUZZ=yes \
+		USE_SYSTEM_JBIG2DEC=yes USE_SYSTEM_JPEGXR=no \
+		USE_SYSTEM_LCMS2=no USE_SYSTEM_LIBJPEG=yes \
+		USE_SYSTEM_MUJS=no USE_SYSTEM_OPENJPEG=yes \
+		USE_SYSTEM_ZLIB=yes USE_SYSTEM_GLUT=yes \
+		USE_SYSTEM_CURL=yes USE_SYSTEM_GUMBO=no \
+		build=release shared=yes install \
+		prefix=$prefix \
+		bindir=$binDir \
```
Begasus, line 99: Since you're using $commandSuffix this should be:
```diff
- 		bindir=$binDir \
+ 		bindir=$commandBinDir \
```

ivanholmes:
> > It should be noted that this breaks pdftranslator because it expects libmupdf 1.8 or thereabouts which is pretty old by now.
> 
> Try building pdftranslator with your libmupdf, if that works out ok pdftranslator needs a revbump (REVISION="*+1")

I tried it, it doesn't work as quite a few function names have changed (the API is not stable at all). I could try to update pdftranslator I suppose, it might be simple enough if it really is just name changes.
My motivation for building mupdf here was to use it with pymupdf - I haven't ever used it directly. I will give the update a go anyway.

app-text/mupdf/mupdf-1.19.1.recipe
ivanholmes: Thanks, I just carried this stuff over from the original recipe because I didn't really understand it. I'll make the changes and commit

Begasus:
Been fiddling with this today, some parts can be adjusted (will come back on that later), for primary arch this can't be build on 32bit as there is no harfbuzz port for gcc2
Trying to see if a build works without harfbuzz :)
pdftranslator 1.0.3 seems broken anyway, I got mupdf_x86 installed and pdftranslator_x86 installs fine without any error needing the previous mupdf port (summons @humdingerb who bumped pdftranslator, @kallisti5 who maintains the port for it)

ivanholmes:
All sounds good. I use x86_64 myself so all this stuff kind of bypassed me. Interested to see how it all works out.

I still think we ought to suffix the shared object file with the major and minor version (1.19) but I am not really sure how to do this (of course I could apply a dirty hack to the makefile but I wonder if there is a clean way that won't require maintenance). At least in that case I suppose pdftranslator could continue to link against 1.8 until someone updates it.

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: cmd's are placed in `$prefix/bin/x86` with `$binDir`, with `$commandBinDir` they are placed in `$prefix/bin `

ivanholmes: Ah, understood. Will you make this change as part of your work on the recipe or shall I?

Begasus:
For starters, gumbo and lcms2 can be left out for the pkg-config search (in the patch), gumbo hasn't been developed over years so probably good to use the bundled one (there is no port for it at haikuports either), lcms2 has been forked by mupdf and got some patches that are not in the official release for lcms2, so use the bundled one there too (getting errors not finding a header with the one currently in haikuports).

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: I'll make the suggestions, the rest is on you to change it in the recipe here. :)

ivanholmes: Works for me. New commit incoming shortly, cheers

Begasus (changes requested):

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,110 @@
...
+HOMEPAGE="https://www.mupdf.com/"
+COPYRIGHT="2006-2022 Artifex Software, Inc"
+LICENSE="AGPL-3.0"
+REVISION="1"
+SOURCE_URI="https://mupdf.com/downloads/archive/mupdf-$portVersion-source.tar.xz"
+CHECKSUM_SHA256="b5eac663fe74f33c430eda342f655cf41fa73d71610f0884768a856a82e3803e"
+SOURCE_DIR="mupdf-$portVersion-source"
+PATCHES="mupdf-$portVersion.patchset"
+
+ARCHITECTURES="all ?x86_gcc2"
```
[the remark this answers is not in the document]
Begasus: You pushed while I was still writing ;) Resolved

app-text/mupdf/patches/mupdf-1.19.1.patchset
```diff
@@ -44,19 +35,7 @@ index 0fdaecb..926291a 100644
...
-+	SYS_LEPTONICA_CFLAGS := $(shell pkg-config --cflags lept)
-+	SYS_LEPTONICA_LIBS := $(shell pkg-config --libs lept)
-+  endif
-+
-+  HAVE_SYS_TESSERACT := $(shell pkg-config --exists 'tesseract >= 4.0.0' && echo yes)
-+  ifeq ($(HAVE_SYS_TESSERACT),yes)
-+	SYS_TESSERACT_CFLAGS := $(shell pkg-config --cflags tesseract)
-+	SYS_TESSERACT_LIBS := $(shell pkg-config --libs tesseract)
-+  endif
-+
```
Begasus: I think they could be preserved, works for me here to have these 2 included :)

app-text/mupdf/mupdf-1.19.1.recipe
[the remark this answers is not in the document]
ivanholmes: Yes I did sorry, leave it with me and I'll implement all the suggestions and push again :)

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,110 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libharfbuzz$secondaryArchSuffix
+	lib:libjbig2dec$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libopenjp2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
```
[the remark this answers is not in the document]
ivanholmes: liblept and libtesseract are only needed for some kind of OCR capability I think which requires extra files, I've removed them from pkg_config now so won't add them here.

app-text/mupdf/mupdf-1.19.1.recipe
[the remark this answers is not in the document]
Begasus: Make sure you test the changes locally too (new clean builds ...)

app-text/mupdf/mupdf-1.19.1.recipe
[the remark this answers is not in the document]
Begasus: Ok, works for me

app-text/mupdf/patches/mupdf-1.19.1.patchset
ivanholmes: I will test with them deleted from the thirdparty directory, pretty sure they aren't needed though

Begasus: Will do a check here too, for the meantime I looked at a few other recipes around used by other OS's, they keep leptonica and tesseract also in thirdparty:
1. https://packages.altlinux.org/en/p10/srpms/mupdf/specfiles/#line-54
2. https://github.com/archlinux/svntogit-community/blob/packages/mupdf/trunk/PKGBUILD#L32

Begasus: During build with leptonica and tesseract thirdparty included, libraries not included in the recipe, I don't see any use of them during build, so I guess it's safe to move forward without them :)

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,110 @@
...
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+PROVIDES="
+	mupdf$secondaryArchSuffix = $portVersion
+	lib:libmupdf$secondaryArchSuffix = $portVersion
+	cmd:muraster$commandSuffix
+	cmd:mutool$commandSuffix
```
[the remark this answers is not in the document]
ivanholmes: I've attempted this, not sure if there is something else that needs doing with it

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,125 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage mupdf$secondaryArchSuffix \
+	"$libDir"/libmupdf.so
```
Begasus, line 94: Didn't check a new build with your latest changes but you should be able to add the cmd's here also, for an example look here: https://github.com/haikuports/haikuports/blob/4f180bdb94db9749ae800820b271e262a12754aa/media-libs/libheif/libheif-1.12.0.recipe#L90

app-text/mupdf/patches/mupdf-1.19.1.patchset
```diff
@@ -0,0 +1,62 @@
+From 17ebe02ea6bfe59922c74de9b954958895515df7 Mon Sep 17 00:00:00 2001
+From: Ivan Holmes <ivan@ivanholmes.co.uk>
```
Begasus, line 2: Looking at previous patchset's it looks like (with some changes) a lot like the first one in line, maybe keep credits for the one that did the original patch?
https://github.com/haikuports/haikuports/blob/master/app-text/mupdf/patches/mupdf-1.6.patchset#L61

ivanholmes: So it does, I did it freshly myself though by copy-pasting the Linux section below and removing bits that didn't apply.
Quite happy to credit PulkoMandy though if you wish

app-text/mupdf/mupdf-1.19.1.recipe
ivanholmes: I gave this a go before, searching for something similar in other packages but didn't find the right variables for the path to the tools so I gave up. Added now

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,127 @@
...
+
+SUMMARY_tools="Tools to convert, examine and rasterise PDF files"
+PROVIDES_tools="
+	mupdf${secondaryArchSuffix}_tools = $portVersion
+	cmd:muraster$commandSuffix
+	cmd:mutool$commandSuffix
+	"
+REQUIRES_tools="
+	haiku$secondaryArchSuffix
+	mupdf$secondaryArchSuffix == $portVersion
```
Begasus, line 62: Same here as line 51

ivanholmes: Done
Didn't mean to request a review, sorry about that. The button text colour is the same as the background in WebPositive so I couldn't see what I was clicking...

Begasus:
Thanks for the changes, will report back tomorrow :)

app-text/mupdf/mupdf-1.19.1.recipe
ivanholmes: For my curiosity, what is the meaning of 'base' here?

ivanholmes:
Cheers, enjoy your evening

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: Don't know the insides but it refers to the base package, see for a short reply from way back :) 
https://github.com/haikuports/haikuports/pull/994#discussion_r95056494

Begasus (approved):

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,127 @@
...
+	mupdf$secondaryArchSuffix == $portVersion base
+	lib:libcrypto$secondaryArchSuffix
+	lib:libcurl$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libharfbuzz$secondaryArchSuffix
+	lib:libjbig2dec$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libopenjp2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
```
Begasus, line 71: libcurl and libglib_2.0 not needed here:
```diff
	lib:libcrypto$secondaryArchSuffix
-	lib:libcurl$secondaryArchSuffix
	lib:libfreetype$secondaryArchSuffix
-	lib:libglib_2.0$secondaryArchSuffix
	lib:libharfbuzz$secondaryArchSuffix
	lib:libjbig2dec$secondaryArchSuffix
	lib:libjpeg$secondaryArchSuffix
	lib:libopenjp2$secondaryArchSuffix
	lib:libz$secondaryArchSuffix
```

app-text/mupdf/patches/mupdf-1.19.1.patchset
Begasus: Would be nice yes :)

app-text/mupdf/mupdf-1.19.1.recipe
ivanholmes: Removed. Probably unrelated but the build failed because it couldn't find python3 so I have added that to the BUILD_PREREQUIRES

app-text/mupdf/patches/mupdf-1.19.1.patchset
ivanholmes: Changed credit now

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: Jikes, you removed them in all places (also in the patchset)? These 2 are not needed for the tools package, for the base and develop package you can leave them in (and also the search for them in the patch)

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,122 @@
...
+
+defineDebugInfoPackage mupdf$secondaryArchSuffix \
+	"$libDir"/libmupdf.so \
+	"$(getPackagePrefix tools)"/bin/mutool \
+	"$(getPackagePrefix tools)"/bin/muraster
+
+BUILD()
+{
+	# remove bundled and use system libaries where possible
+	rm -rf thirdparty/{curl,freetype,harfbuzz,jbig2dec,leptonica,libjpeg,openjpeg,tesseract,zlib}
```
Begasus, line 96: You can leave leptonica and tesseract in there (doesn't harm the build):
```diff
- 	rm -rf thirdparty/{curl,freetype,harfbuzz,jbig2dec,leptonica,libjpeg,openjpeg,tesseract,zlib}
+ 	rm -rf thirdparty/{curl,freetype,harfbuzz,jbig2dec,libjpeg,openjpeg,zlib}
```

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,122 @@
...
+		mandir=$manDir
+
+	prepareInstalledDevelLib libmupdf
+	fixPkgconfig
+
+	# tools package
+	packageEntries tools $commandBinDir
+
+	# devel package
+	packageEntries devel $developDir
```
Begasus, line 121: devel package before tools package

app-text/mupdf/mupdf-1.19.1.recipe
ivanholmes: Okay I'll revert this and then remove them only for the tools package

ivanholmes: I'm curious as to how you could tell they weren't needed here? Sorted anyway

Begasus:
So far things are OK, only leaves the versioned shared library, in the previous version it was done with some sed magic, I'm no good there but maybe someone else could have a look.
https://github.com/haikuports/haikuports/blob/master/app-text/mupdf/mupdf-1.8.recipe#L67

Begasus:
So far things are OK, only leaves the versioned shared library, in the previous version it was done with some sed magic, I'm no good there but maybe someone else could have a look.
https://github.com/haikuports/haikuports/blob/master/app-text/mupdf/mupdf-1.8.recipe#L67

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,125 @@
...
+		mandir=$manDir
+
+	prepareInstalledDevelLib libmupdf
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel $developDir
+
+	# tools package
+	packageEntries tools $commandBinDir
```
Begasus, line 124: With the cmd's now in the tools package, also move the manpages for them there:
```diff
- 	packageEntries tools $commandBinDir
+ 	packageEntries tools $commandBinDir \
+ 	 	$manDir
```

ivanholmes: Will push this soon.

ivanholmes:
> So far things are OK, only leaves the versioned shared library, in the previous version it was done with some sed magic, I'm no good there but maybe someone else could have a look.
> https://github.com/haikuports/haikuports/blob/master/app-text/mupdf/mupdf-1.8.recipe#L67

So I had a look at what the sed command is actually doing here - it basically just changes the name of the output from libmupdf.a to libmupdf.1.8.so, sets up the options for building as a shared library (which we don't need to do as upstream have done it now) and adds the COPYING file to the list of files to install.

To this end I tried changing the line
`MUPDF_LIB = libmupdf.{$SO}` to
`MUPDF_LIB = libmupdf.{$SO}.1.19` and the result is that the library file is not built at all and so linking the tools to it fails.

I also tried changing the definiton of `{$SO}` from `.so` to `.so.1.19` which works better (libmupdf is generated and linked) but then linking the tools fails with the message 'cannot find -lmupdf'.

I'll continue looking into this...

ivanholmes:
I think I've got it... based on the patch I added (to remove some code that changes the library detection if building as a shared library) which essentially undoes this https://github.com/ArtifexSoftware/mupdf/commit/7e6d92e63e31b55a397a5c7e9f37f92e74834524 it links properly and works. 
Perhaps it would be more proper to alter the pattern matching but I am not sure if it actually matters for our use case? Certainly it was not there in earlier versions so perhaps it isn't relevant to Haiku.

I also added the manDir to the tools package.

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,134 @@
...
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+# get the library version number (excluding patch rev) from port version
+libVersion=$(echo $portVersion | cut -d"." -f1,2)
```
Begasus, line 30: You now get a $libVersion = 1.19, I think this should be 1.19.1 (prior version was 1.8, but $portVersion was also 1.8)

Begasus: Checked some things on my end here with changes the $libVersion to use $portVersion, also adding libVersionCompat:
```diff
- libVersion=$(echo $portVersion | cut -d"." -f1,2)
+ libVersion="$portVersion"
+ libVersionCompat="$libVersion compat >= ${libVersion%.**}"
```
Using compat = 1.19 here, seems there are some compat breakage between prior versions, from where I could check (thanks @extrowerk ) it ony goes up to 1.17.0: https://abi-laboratory.pro/index.php?view=timeline&l=mupdf

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,134 @@
...
+
+PROVIDES="
+	mupdf$secondaryArchSuffix = $portVersion
+	lib:libmupdf$secondaryArchSuffix = $portVersion
```
Begasus, line 34: Use libVersionCompat here and in devel package:
```diff
- 	lib:libmupdf$secondaryArchSuffix = $portVersion
+ 	lib:libmupdf$secondaryArchSuffix = $libVersionCompat
```

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,134 @@
...
+	lib:libharfbuzz$secondaryArchSuffix
+	lib:libjbig2dec$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libopenjp2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	mupdf${secondaryArchSuffix}_devel = $portVersion
+	devel:libmupdf$secondaryArchSuffix
```
Begasus, line 51: Same here:
```diff
- 	devel:libmupdf$secondaryArchSuffix
+ 	devel:libmupdf$secondaryArchSuffix = $libVersionCompat
```

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: Hold on on changes atm, last check to import fitz yield a missing library ...

ivanholmes: > You now get a $libVersion = 1.19, I think this should be 1.19.1 (prior version was 1.8, but $portVersion was also 1.8)

This was my intent but as you say it's better that this is libVersionCompat. With mupdf the API is only compatible within minor versions (1.19, 1.8 etc), at 1.12.0 they introduced patch versions as well which are compatible with each other.

ivanholmes: > Hold off on changes atm, last check to import fitz yield a missing library ...

I noticed this after I pushed, oops! I think it is because libmupdf.so symlink in system/lib doesn't exist. We should create it (not sure how, I thought I had but the one in the recipe goes in system/develop/lib I think) but also link pymupdf with the versioned file directly

ivanholmes: Changes you suggested have been pushed. Just need to work out this last kink

Begasus:
As mentioned in one of the earlier comments, importing fitz fails when libVersion is set (with current changes and the ones I wanted to check). Maybe we should drop libVersion'n totally and use a variable like $portVersionCompat to version the library?

ivanholmes:
> As mentioned in one of the earlier comments, importing fitz fails when libVersion is set (with current changes and the ones I wanted to check). Maybe we should drop libVersion'n totally and use a variable like $portVersionCompat to version the library?

Perhaps so, I am happy with whatever really. It could be a problem if/when apps start using the library as we will have to keep multiple different versions around. The versioning scheme is certainly a bit strange, I haven't seen any other lib with 3 version numbers as a suffix to .so.

Begasus:
> I haven't seen any other lib with 3 version numbers as a suffix to .so.

I thing the majority of the libraries use 3 version numbers in their suffix (hence libVersion%%.*) or libVersion="*.*.*" or libVersion="$portVersion" ..., checking a build now with portVersionCompat where I use: portVersionCompat="$portVersion compat >= ${portVersion%.**}"

This doesn't bread the libversioning but you could use that compat for instance for pymupdf in: devel:libfreetype$secondaryArchSuffix >= 1.19

ivanholmes:
> > I haven't seen any other lib with 3 version numbers as a suffix to .so.
> 
> I thing the majority of the libraries use 3 version numbers in their suffix (hence libVersion%%._) or libVersion="_._._" or libVersion="$portVersion" ..., checking a build now with portVersionCompat where I use: portVersionCompat="$portVersion compat >= ${portVersion%.**}"
Yep I just looked and found loads so not sure what I was on about... let me know how this goes
> 
> This doesn't break the libversioning but you could use that compat for instance for pymupdf in: devel:libmupdf$secondaryArchSuffix >= 1.19
Makes sense although would this allow 1.20? It shouldn't of course
> 
> Need to take a break I think ;)
Too right! This has been hurting my head for a while ;)

Begasus:
> Makes sense although would this allow 1.20? It shouldn't of course

When it comes to that I guess we can just delete the 1.19.1 version and move on to 1.20.*, rev-bumping the needed recipes then (probably in this case only pymupdf)

ivanholmes:
> > Makes sense although would this allow 1.20? It shouldn't of course
> 
> When it comes to that I guess we can just delete the 1.19.1 version and move on to 1.20.*, rev-bumping the needed recipes then (probably in this case only pymupdf)

Works for me.

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+# get the library version number (excluding patch rev) from port version
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%.**}"
```
Begasus, line 31: Checked with a local build here and with python3.7 and python3.10 works OK for pymupdf:
```diff
-# get the library version number (excluding patch rev) from port version
-libVersion="$portVersion"
-libVersionCompat="$libVersion compat >= ${libVersion%.**}"
+portVersionCompat="$portVersion compat >= ${portVersion%.**}"
```

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+
+PROVIDES="
+	mupdf$secondaryArchSuffix = $portVersion
+	lib:libmupdf$secondaryArchSuffix = $libVersionCompat
```
Begasus, line 35:
```diff
- 	lib:libmupdf$secondaryArchSuffix = $libVersionCompat
+ 	lib:libmupdf$secondaryArchSuffix = $portVersionCompat
```

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:find
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage mupdf$secondaryArchSuffix \
+	"$libDir"/libmupdf.so.$libVersion \
```
Begasus, line 96:
```diff
- 	"$libDir"/libmupdf.so.$libVersion \
+ 	"$libDir"/libmupdf.so \
```

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+		shared=yes build=release install \
+		prefix=$prefix \
+		bindir=$commandBinDir \
+		docdir=$docDir \
+		libdir=$libDir \
+		incdir=$includeDir \
+		mandir=$manDir
+
+	ln -s libmupdf.so.$libVersion $libDir/libmupdf.so
+
```
Begasus, line 125:
```diff
-	ln -s libmupdf.so.$libVersion $libDir/libmupdf.so
-
```

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,135 @@
...
+	"$(getPackagePrefix tools)"/bin/muraster
+
+BUILD()
+{
+	# remove bundled and use system libaries where possible
+	rm -rf thirdparty/{curl,freetype,harfbuzz,jbig2dec,libjpeg,openjpeg,zlib}
+
+	# add version number to library
+	sed "s/SO := so$/SO := so.${libVersion}/1" -i Makerules
+
```
Begasus, line 107:
```diff
-	# add version number to library
-	sed "s/SO := so$/SO := so.${libVersion}/1" -i Makerules
-
```

Begasus (commented):
Looking at other recipes on repology they don't provide a soname version for their ports, probably it was intended this way, with $portVersionCompat we have something to compare to.

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: For this I changed requirement for pymupdf to use >= 1.19 for the devel:libmupdf and seems to work fine ;)

ivanholmes:
> Looking at other recipes on repology they don't provide a soname version for their ports, probably it was intended this way, with $portVersionCompat we have something to compare to.

Great I'll sort everything to work this way and then push

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: Looking at things this is not resolved? This needs to be done so pymupdf can pick up the correct version for mupdf.

app-text/mupdf/mupdf-1.19.1.recipe
```diff
@@ -0,0 +1,128 @@
...
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libharfbuzz$secondaryArchSuffix
+	lib:libjbig2dec$secondaryArchSuffix
+	lib:libjpeg$secondaryArchSuffix
+	lib:libopenjp2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	mupdf${secondaryArchSuffix}_devel = $portVersionCompat
```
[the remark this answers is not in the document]
ivanholmes: I thought I had! Just testing these changes now

app-text/mupdf/mupdf-1.19.1.recipe
Begasus: $portVersion for pymupdf isn't the same as mupdf ;)
```
devel:libmupdf$secondaryArchSuffix >= 1.19
```
Should work

ivanholmes: $portVersion isn't the same but ${portVersion%.**} should be the same, because the major and minor versions of pymupdf track those of mupdf exactly. Will try what you suggested anyway, cheers

Begasus: Reminder, that is for pymupdf, not for mupdf

ivanholmes: On further examination I think this failed because I hadn't set the version for devel:libmupdf properly in the recipe for mupdf. Will continue this at #6824

Begasus:
Will do a full re-check tomorrow, for now, uninstall all you got in /system/packages for mupdf (don't remove all there!!!) ;)
Re-build pymupdf with the packages still in ~/haikuports/packages (don't install them, the build for pymupdf will pick those up there too), try to install one of the sub-packages for pymupdf (python3 for instance), it should give you a notice that it can't be installed due to missing mupdf, install base package for mupdf, install pymupdf, check import with python3 ...

ivanholmes:
> Will do a full re-check tomorrow, for now, uninstall all you got in /system/packages for mupdf (don't remove all there!!!) ;) Re-build pymupdf with the packages still in ~/haikuports/packages (don't install them, the build for pymupdf will pick those up there too), try to install one of the sub-packages for pymupdf (python3 for instance), it should give you a notice that it can't be installed due to missing mupdf, install base package for mupdf, install pymupdf, check import with python3 ...

Tried exactly steps you described with python3.7, all seems to work properly/as expected :)

Begasus:
I don't see any use for libmupdf in haikuports, maybe remove the older recipes/patches while we are in the process of bumping, think this should be OK to merge now.

ivanholmes:
> I don't see any use for libmupdf in haikuports, maybe remove the older recipes/patches while we are in the process of bumping, think this should be OK to merge now.

I think PDFTranslator depends on it? Either it could be updated or a legacy version of MuPDF 1.8 provided

Begasus:
> > I don't see any use for libmupdf in haikuports, maybe remove the older recipes/patches while we are in the process of bumping, think this should be OK to merge now.
> 
> I think PDFTranslator depends on it? Either it could be updated or a legacy version of MuPDF 1.8 provided

My bad, DocumentViewer is also using it, will try a re-build for both with this version, if things are ok we'll merge and revbump those 2 :)

ivanholmes:
> > > I don't see any use for libmupdf in haikuports, maybe remove the older recipes/patches while we are in the process of bumping, think this should be OK to merge now.
> > 
> > 
> > I think PDFTranslator depends on it? Either it could be updated or a legacy version of MuPDF 1.8 provided
> 
> My bad, DocumentViewer is also using it, will try a re-build for both with this version, if things are ok we'll merge and revbump those 2 :)

Okay, DocumentViewer might work, I have already tried PDFTranslator and there are loads of missing symbols. Hopefully it can be updated without too much hassle if the names have just changed without the underlying functionality being much different...

ivanholmes:
https://github.com/HaikuArchives/DocumentViewer/pull/17 seems to be the last time DocumentViewer was updated to include the latest MuPDF API changes

Begasus:
I see -fpic being used by previous version (and on some of the other recipes at repology, maybe we should still use it?
https://github.com/haikuports/haikuports/blob/4f180bdb94db9749ae800820b271e262a12754aa/app-text/mupdf/mupdf-1.8.recipe#L77

ivanholmes:
> I see -fpic being used by previous version (and on some of the other recipes at repology, maybe we should still use it?
> 
> https://github.com/haikuports/haikuports/blob/4f180bdb94db9749ae800820b271e262a12754aa/app-text/mupdf/mupdf-1.8.recipe#L77

It seems Linux From Scratch does use it, not sure if it is necessary here or implied by some other option somewhere but it can't hurt. I'll add it
https://www.linuxfromscratch.org/blfs/view/svn/pst/mupdf.html

ivanholmes:
> > [HaikuArchives/DocumentViewer#17](https://github.com/HaikuArchives/DocumentViewer/pull/17) seems to be the last time DocumentViewer was updated to include the latest MuPDF API changes
> 
> Those were for version 1.7 (probably 1.8 was OK with them too), build fails now (with latest commit checkout there)

Yep, was a long time ago. I'll have a look tonight at which symbols are missing and how easy it will be to update

ivanholmes:
> > > [HaikuArchives/DocumentViewer#17](https://github.com/HaikuArchives/DocumentViewer/pull/17) seems to be the last time DocumentViewer was updated to include the latest MuPDF API changes
> > 
> > 
> > Those were for version 1.7 (probably 1.8 was OK with them too), build fails now (with latest commit checkout there)
> 
> Yep, was a long time ago. I'll have a look tonight at which symbols are missing and how easy it will be to update

The way a few functions are called has changed and there isn't support for progressive loading any more (stuff related to this was giving the missing symbols). Now PDFTranslator is working with the latest MuPDF, as you can see I have opened a PR https://github.com/threedeyes/PDFTranslator/pull/5 with my changes. Now on to DocumentViewer...

ivanholmes:
Having trouble getting DocumentViewer to work as you can see from my WIP PR https://github.com/HaikuArchives/DocumentViewer/pull/41
Any help would be appreciated.

Begasus:
> Having trouble getting DocumentViewer to work as you can see from my WIP PR [HaikuArchives/DocumentViewer#41](https://github.com/HaikuArchives/DocumentViewer/pull/41) Any help would be appreciated.

I'm no use there, hope someone will chip in.

PR #6824, opened 2022-04-11
pymupdf: add recipe for 1.19.6

ivanholmes:
As alluded in my other PR #6823 here is a recipe for PyMuPDF. Working well on hrev 56001.

### DIFF
--- a/dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
+++ b/dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
@@ -16,13 +16,13 @@ index 475bc5d..4647f52 100644
 +    
 +elif sys.platform.startswith("haiku"):
 +    include_dirs = [
-+        "/boot/system/develop/headers/mupdf",
 +        "/boot/system/develop/headers/x86/mupdf",
-+        "/boot/system/develop/headers/freetype2",
-+        "/boot/system/develop/headers/x86/freetype2"
++        "/boot/system/develop/headers/mupdf",
++        "/boot/system/develop/headers/x86/freetype2",
++        "/boot/system/develop/headers/freetype2"
 +    ]
 +    libraries = ["mupdf"]
-+    library_dirs = ["/boot/system/lib", "/boot/system/lib/x86"]
++    library_dirs = ["/boot/system/lib/x86", "/boot/system/lib"]
 +    extra_link_args = []
  
  else:
--- a/dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
+++ b/dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
@@ -1,0 +1,30 @@
+From fdefbb7634650726393dcae61c14842f8d3ea891 Mon Sep 17 00:00:00 2001
+From: Ivan Holmes <ivan@ivanholmes.co.uk>
+Date: Mon, 11 Apr 2022 16:54:27 +0100
+Subject: Add Haiku platform to setup.py
+
+
+diff --git a/setup.py b/setup.py
+index 475bc5d..8ee0d72 100644
+--- a/setup.py
++++ b/setup.py
+@@ -104,7 +104,15 @@ elif sys.platform.startswith(("darwin", "freebsd", "openbsd")):
+     libraries = ["mupdf", "mupdf-third"]
+     library_dirs = ["/usr/local/lib", "/opt/homebrew/lib"]
+     extra_link_args = []
+-
++    
++elif sys.platform.startswith("haiku"):
++    include_dirs = [
++        "/boot/system/develop/headers/mupdf",
++        "/boot/system/develop/headers/freetype2",
++    ]
++    libraries = ["mupdf"]
++    library_dirs = ["/boot/system/lib"]
++    extra_link_args = []
+ 
+ else:
+     include_dirs = [
+-- 
+2.30.2
+
--- a/dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
+++ b/dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
@@ -1,14 +1,14 @@
-From fdefbb7634650726393dcae61c14842f8d3ea891 Mon Sep 17 00:00:00 2001
+From c65cf1b6eb9791e523e348946c15be8621e22744 Mon Sep 17 00:00:00 2001
 From: Ivan Holmes <ivan@ivanholmes.co.uk>
-Date: Mon, 11 Apr 2022 16:54:27 +0100
+Date: Wed, 13 Apr 2022 13:07:48 +0100
 Subject: Add Haiku platform to setup.py
 
 
 diff --git a/setup.py b/setup.py
-index 475bc5d..8ee0d72 100644
+index 475bc5d..4647f52 100644
 --- a/setup.py
 +++ b/setup.py
-@@ -104,7 +104,15 @@ elif sys.platform.startswith(("darwin", "freebsd", "openbsd")):
+@@ -104,7 +104,17 @@ elif sys.platform.startswith(("darwin", "freebsd", "openbsd")):
      libraries = ["mupdf", "mupdf-third"]
      library_dirs = ["/usr/local/lib", "/opt/homebrew/lib"]
      extra_link_args = []
@@ -17,10 +17,12 @@ index 475bc5d..8ee0d72 100644
 +elif sys.platform.startswith("haiku"):
 +    include_dirs = [
 +        "/boot/system/develop/headers/mupdf",
++        "/boot/system/develop/headers/x86/mupdf",
 +        "/boot/system/develop/headers/freetype2",
++        "/boot/system/develop/headers/x86/freetype2"
 +    ]
 +    libraries = ["mupdf"]
-+    library_dirs = ["/boot/system/lib"]
++    library_dirs = ["/boot/system/lib", "/boot/system/lib/x86"]
 +    extra_link_args = []
  
  else:
--- a/dev-python/pymupdf/pymupdf-1.19.6.recipe
+++ b/dev-python/pymupdf/pymupdf-1.19.6.recipe
@@ -1,0 +1,78 @@
+SUMMARY="Python bindings for MuPDF's rendering library"
+DESCRIPTION="PyMuPDF is a Python binding with support for MuPDF, a lightweight PDF, XPS, and \
+E-book viewer, renderer, and toolkit, which is maintained and developed by Artifex Software, Inc.
+MuPDF can access files in PDF, XPS, OpenXPS, CBZ, EPUB and FB2 (e-books) formats, and it is known \
+for its top performance and high rendering quality.
+With PyMuPDF you can access files with extensions like \".pdf\", \".xps\", \".oxps\", \".cbz\", \
+\".fb2\" or \".epub\". In addition, about 10 popular image formats can also be handled like \
+documents: \".png\", \".jpg\", \".bmp\", \".tiff\", etc..
+"
+HOMEPAGE="https://pypi.org/project/PyMuPDF/
+	https://github.com/pymupdf/PyMuPDF/"
+COPYRIGHT="2022 Jorj X. McKie and Ruikai Liu"
+LICENSE="AGPL-3.0"
+REVISION="1"
+pypi="032d24e0c774e67742395fda163a172c60e4d0f9875785d5199eb2956d5e"
+SOURCE_URI="https://files.pythonhosted.org/packages/9f/1d/$pypi/PyMuPDF-$portVersion.tar.gz"
+CHECKSUM_SHA256="ef3d13e27f1585d776f6a2597f113aabd28d36b648b983a72850b21c5399ab08"
+SOURCE_DIR="PyMuPDF-$portVersion"
+PATCHES="pymupdf-$portVersion.patchset"
+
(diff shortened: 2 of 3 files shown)


korli:
1/ License not found
2/ x86 won't work

ivanholmes:
I'm a bit confused as to why the license wasn't found. It is there on my local machine, also it's the same license (AGPL-3.0) as MuPDF (#6823) which passed linting about an hour before.

R.e. x86 not working, should I replace:
```
ARCHITECTURES="all ?x86_gcc2"
SECONDARY_ARCHITECTURES="x86"
```
with simply
`ARCHITECTURES="x86_64" `

or something else like
`ARCHITECTURES="all !x86"`

I am not really sure how the system works in this case.

Thanks for your assistance.

ivanholmes:
I see. As you can tell I'm new to this, thanks for your patience.
I will add the license to this port and commit shortly

ivanholmes:
License added as you can see. I didn't notice any issue before uploading as on my Haiku system the file AGPL-3.0 is present in /boot/system/data/licenses. I wonder why this is not the case on the build server?

Begasus:
> License added as you can see. I didn't notice any issue before uploading as on my Haiku system the file AGPL-3.0 is present in /boot/system/data/licenses. I wonder why this is not the case on the build server?

You probably had one or some packages that ship with the license, so haikuporter will find that one.
These are the default licenses in Haiku: https://github.com/haiku/haiku/tree/master/data/system/data/licenses
The ones that are not present there should be shipped with the recipe.

ivanholmes:
> > License added as you can see. I didn't notice any issue before uploading as on my Haiku system the file AGPL-3.0 is present in /boot/system/data/licenses. I wonder why this is not the case on the build server?
> 
> You probably had one or some packages that ship with the license, so haikuporter will find that one. These are the default licenses in Haiku: https://github.com/haiku/haiku/tree/master/data/system/data/licenses The ones that are not present there should be shipped with the recipe.

Ah yes, that makes sense, because of course I have mupdf installed as it is required by this recipe. Out of interest, Is there any reason the AGPL is not included with Haiku? I count 6 packages using it including this one.
Thanks for the info!

pulkomandy:
> Ah yes, that makes sense, because of course I have mupdf installed as it is required by this recipe. Out of interest, Is there any reason the AGPL is not included with Haiku?

The Haiku package only includes licences for code contained in the Haiku package and its dependencies I think. Haikuporter started using these as a convenience, but the plan was not to provide every popular license in the Haiku package.

Haikuports could consider providing a package with more common licenses if it seems desirable. Or they could be added to Haiku, but that means extra disk space and extra download time used at every Haiku package update.

ivanholmes:
> Haikuports could consider providing a package with more common licenses if it seems desirable. Or they could be added to Haiku, but that means extra disk space and extra download time used at every Haiku package update.

Makes sense. It is quite minor as the files are quite small anyway but I think a license package for Haikuports could be a good idea.

Begasus:
> Haikuports could consider providing a package with more common licenses if it seems desirable.

Well it could be done I suppose, but you'd need an extra dependency for those that need the license in REQUIRES?
Would take some search but if there is a need for it ... :)

pulkomandy:
According to https://pulkomandy.tk/stats_hp/files.html the two most popular licenses not available in the Haiku package are ISC (22 copies in haikuports) and GNU FDL v1 (17 copies), then Python (6 copies) and Unlicense (5 copies).

ivanholmes:
Perhaps there is a case for a licenses package then. One for you long-term contributors to decide I think!

dev-python/pymupdf/pymupdf-1.19.6.recipe
```diff
@@ -0,0 +1,78 @@
+SUMMARY="Python bindings for MuPDF's rendering library"
+DESCRIPTION="PyMuPDF is a Python binding with support for MuPDF, a lightweight PDF, XPS, and \
+E-book viewer, renderer, and toolkit, which is maintained and developed by Artifex Software, Inc.
+MuPDF can access files in PDF, XPS, OpenXPS, CBZ, EPUB and FB2 (e-books) formats, and it is known \
+for its top performance and high rendering quality.
+With PyMuPDF you can access files with extensions like \".pdf\", \".xps\", \".oxps\", \".cbz\", \
+\".fb2\" or \".epub\". In addition, about 10 popular image formats can also be handled like \
+documents: \".png\", \".jpg\", \".bmp\", \".tiff\", etc..
+"
```
Begasus, line 9: Move closing double-qoute to the end of the previous line:
```diff
-documents: \".png\", \".jpg\", \".bmp\", \".tiff\", etc..
-"
+documents: \".png\", \".jpg\", \".bmp\", \".tiff\", etc.."
```

dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
```diff
@@ -0,0 +1,30 @@
...
+-
++    
++elif sys.platform.startswith("haiku"):
++    include_dirs = [
++        "/boot/system/develop/headers/mupdf",
++        "/boot/system/develop/headers/freetype2",
++    ]
++    libraries = ["mupdf"]
++    library_dirs = ["/boot/system/lib"]
++    extra_link_args = []
```
Begasus, line 24: This won't work for 32bit secondary architecture, "/boot/system/develop/headers/mupdf" vs "/boot/system/develop/headers/`x86`/mupdf"

Begasus: Works if I add the correct paths to these:
```diff
elif sys.platform.startswith("haiku"):
    include_dirs = [
        "/boot/system/develop/headers/mupdf",
+        "/boot/system/develop/headers/x86/mupdf",
        "/boot/system/develop/headers/freetype2",
+        "/boot/system/develop/headers/x86/freetype2",
    ]
    libraries = ["mupdf"]
    library_dirs = ["/boot/system/lib"]
+    library_dirs = ["/boot/system/lib/x86"]
    extra_link_args = []
```

Begasus:
Installed the pymupdf_x86_python3 but somehow python3 can't seem to find the module.
Steps to check, in Terminal run:
1. python3
2. import pymupdf

dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
korli: You might try to use pkgconfig, for instance: https://stackoverflow.com/questions/60174152/how-do-i-add-pkg-config-the-setup-py-of-a-cython-wrapper/60893447#60893447

Begasus: Nice one, will check it out

ivanholmes: I'll give it a go shortly...

ivanholmes:
> Installed the pymupdf_x86_python3 but somehow python3 can't seem to find the module. Steps to check, in Terminal run:
> 
> 1. python3
> 2. import pymupdf
> 
> **EDIT** Maybe it's not meant to use `pymupdf`, `import fitz` ... `help (fitz)` works ok :)

Yes, for the record the correct import is fitz

dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
ivanholmes: I've had a look into this. PyMuPDF doesn't use pkg-config for this on any other platforms so I think the patch would be simpler, shorter and easier to understand if we didn't. To this end I have pushed with the changes made by @Begasus.
I'm not opposed to using pkg-config in theory though.

dev-python/pymupdf/pymupdf-1.19.6.recipe
ivanholmes: Done

dev-python/pymupdf/patches/pymupdf-1.19.6.patchset
korli: In the end the patch should be upstreamed. Why it is even hardcoded is over my head.

ivanholmes: I agree in principle. It's probably hardcoded just because it's simple to begin with... I haven't seen many python modules' code so I'm not sure what the normal practice is.

dev-python/pymupdf/pymupdf-1.19.6.recipe
```diff
@@ -0,0 +1,77 @@
...
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	$portName = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libmupdf$secondaryArchSuffix
+	lib:libfreetype$secondaryArchSuffix
```
Begasus, line 29: Both `libmupdf` and `libfreetype` are linked in the `pythonPackages` so they should be moved to `REQUIRES_$pythonPackage`
I had pymupdf_python3 installed when I removed mupdf, it didn't ask me to remove pymupdf_python3, atm only the base (empty) package require both libs which isn't needed.

Begasus: https://github.com/haikuports/haikuports/pull/6825#pullrequestreview-943304456

ivanholmes: Holding off on this until we decide what to do about versioning

ivanholmes:
Latest push tested on all versions of python from 3.7 to 3.10
The ${portVersion%.**} should always work because pymupdf 1.19.x needs mupdf 1.19.x (the version numbers excluding the patch must always be equal). It saves this having to be manually when 1.20.x comes.

dev-python/pymupdf/pymupdf-1.19.6.recipe
```diff
@@ -25,13 +25,11 @@ PROVIDES="
...
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libmupdf$secondaryArchSuffix
-	lib:libfreetype$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	devel:libmupdf$secondaryArchSuffix
+	devel:libmupdf$secondaryArchSuffix >= ${portVersion%.**}
```
Begasus, line 32: As mentioned, don't use $portVersion here:
```diff
- 	devel:libmupdf$secondaryArchSuffix >= ${portVersion%.**}
+ 	devel:libmupdf$secondaryArchSuffix >= 1.19
```

ivanholmes: of course plain `$portVersion` wouldn't work as it could have a different patch revision but that is why I'm using the substitution

Begasus: If you check on other recipes (eg boost related), you would see none of them use $portVersion variables there, it's common to use the depending version (even if in this case it's the same) (PS, didn't check this change), but what if pymupdf bumpt to 1.20.* before mupdf does?

ivanholmes: In this case I am quite sure that won't happen - it would be meaningless to for the maintainers to release pymupdf 1.20.* without mupdf 1.20.*
Anyway I am happy to conform to the convention used by other ports so I'll push the change in a moment :)

ivanholmes: > it's common to use the depending version (even if in this case it's the same)
My point here is that the depending version is **absolutely always** the same between pymupdf and mupdf - pymupdf 1.19.x won't work with mupdf 1.18.x or 1.20.x or any other version of mupdf except for 1.19.x, because mupdf breaks compatibility with every minor version release 
That's why it seems logical to me to do this, but I'm happy to go with whatever you think is best and get these ports merged :)

Begasus: ATM pymupdf is already at 1.19.6 while mupdf is still at 1.19.1, I can't tell what the developers agree there, but better safe then sorry, dependency should rely on requires, not on the own version (now I'm closing down for today) :)

ivanholmes: Cheers, have a good evening

korli: pymupdf uses the static library of mupdf on Arch Linux, this is surely another "good" option  ...

ivanholmes: > pymupdf uses the static library of mupdf on Arch Linux, this is surely another "good" option ...

Yes I suppose it is a decent option - the unstable API is the reason MuPDF didn't support building as a shared library until recently. I tend to prefer the shared library though as it is more efficient and MuPDF is quite big.

korli: Sure, depends on the number of users, and how often it is updated.

Begasus: Well you could provide both shared and static library for mupdf, not sure if we want to go there, so far from my checks pymupdf seems ok with the shared library.

ivanholmes: Yeah I don't really want to go there. I am happy to keep MuPDF and pymupdf updated together and if anyone uses it and an issue arises look into providing a static lib as well.

PR #6833, opened 2022-04-16
licenses, first draft to add common licenses

Begasus:


Begasus:
Not sure how if or how to add PROVIDES for the separate licenses yet. Help welcome :)

waddlesplash (changes requested):
I don't think we should do this, for a variety of reasons.

I think we previously discussed that we may need to ensure the license text is always distributed with the package, which would be a change at HaikuPorter, and then the licenses may be stored with it.

Begasus:
Fair enough, it was just from a question on another PR.

pulkomandy:
> I think we previously discussed that we may need to ensure the license text is always distributed with the package, which would be a change at HaikuPorter, and then the licenses may be stored with it.

I still don't agree with that. All that matters is that the license is distributed "with" the software, that is, on the same repository. It doesn't need to be in the same package.

kallisti5:
On an alternative note, I sent a message to Creative Commons to see if they could offer a public API to generate licenses.  

Potentially, haikuporter could detect CC- and contact Creative Commons to get the official text of various configurations.  We then could pack the non-standard licenses with the package automatically without storing hundreds of CC combinations in our repos.

PR #6834, opened 2022-04-16
mdp: new recipe

hoanga:
the following recipe adds support for mdp.

sample run below:
```
> uname -a
Haiku shredder 1 hrev56016 Apr 15 2022 06:47:33 x86_64 x86_64 Haiku

> mdp --version
mdp 1.0.15
Copyright (C) 2018 Michael Goehler
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Michael Goehler and others, see <https://github.com/visit1985/mdp/blob/master/AUTHORS>.
```

### DIFF
--- a/app-text/mdp/mdp-1.0.15.recipe
+++ b/app-text/mdp/mdp-1.0.15.recipe
@@ -1,7 +1,7 @@
 SUMMARY="A markdown presentation tool"
 DESCRIPTION="A command-line based markdown presentation tool"
 HOMEPAGE="https://github.com/visit1985/mdp"
-COPYRIGHT="2016 visit1985"
+COPYRIGHT="2018 Michael Goehler and others"
 LICENSE="GNU GPL v3"
 REVISION="1"
 SOURCE_URI="https://github.com/visit1985/mdp/archive/refs/tags/$portVersion.tar.gz"
--- a/app-text/mdp/mdp-1.0.15.recipe
+++ b/app-text/mdp/mdp-1.0.15.recipe
@@ -21,7 +21,7 @@ REQUIRES="
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	devel:libncurses$secondaryArchSuffix
+	devel:libncursesw$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
--- a/app-text/mdp/mdp-1.0.15.recipe
+++ b/app-text/mdp/mdp-1.0.15.recipe
@@ -1,0 +1,46 @@
+SUMMARY="A markdown presentation tool"
+DESCRIPTION="A command-line based markdown presentation tool"
+HOMEPAGE="https://github.com/visit1985/mdp"
+COPYRIGHT="2016 visit1985"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/visit1985/mdp/archive/refs/tags/$portVersion.tar.gz"
+CHECKSUM_SHA256="3edc8ea1551fdf290d6bba721105e2e2c23964070ac18c13b4b8d959cdf6116f"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	mdp$secondaryArchSuffix = $portVersion
+	cmd:mdp$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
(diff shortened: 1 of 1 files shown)


app-text/mdp/mdp-1.0.15.recipe
```diff
@@ -0,0 +1,46 @@
...
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	mdp$secondaryArchSuffix = $portVersion
+	cmd:mdp$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
```
korli, line 19: Isn't it libncursesw?

hoanga: i have pushed an update that uses `ncursesw` rather than `ncurses`. thanks!

app-text/mdp/mdp-1.0.15.recipe
```diff
@@ -0,0 +1,46 @@
...
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://github.com/visit1985/mdp/archive/refs/tags/$portVersion.tar.gz"
+CHECKSUM_SHA256="3edc8ea1551fdf290d6bba721105e2e2c23964070ac18c13b4b8d959cdf6116f"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	mdp$secondaryArchSuffix = $portVersion
```
Begasus, line 14: Install puts the binary in: `install -d $(DESTDIR)$(PREFIX)/bin` which means this should be:
```
cmd:mdp = $portVersion
```

app-text/mdp/mdp-1.0.15.recipe
```diff
@@ -0,0 +1,46 @@
...
+	haiku${secondaryArchSuffix}_devel
+	devel:libncursesw$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+TEST_REQUIRES="
+	"
```
Begasus, line 32: Not required:
```diff
-TEST_REQUIRES="
-	"
```

app-text/mdp/mdp-1.0.15.recipe
```diff
@@ -0,0 +1,46 @@
...
+
+BUILD()
+{
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $binDir
```
Begasus, line 41: Shouldn't be needed, make install takes care of that:
```diff
- 	mkdir -p $binDir
```

app-text/mdp/mdp-1.0.15.recipe
```diff
@@ -0,0 +1,46 @@
...
+	mkdir -p $manDir
+
+	make install PREFIX=$prefix
+	mv $prefix/share/man/* $manDir
```
Begasus, line 45: POLICY error for $prefix/share, so remove it from the package after manpage move:
```diff
+ 	rm -rf $prefix/share/
```

Begasus (changes requested):
Checked a build for 32bit, some changes needed, some things not required

Begasus:
Created a upstream PR for which you could apply for the most part on this release also, this simplifies the install targets :)

Begasus:
PS, a nice thing to have, works pretty well (only needs another background color (theme) for Terminal then the default white color) :)

app-text/mdp/mdp-1.0.15.recipe
korli: REQUIRES still pending.

hoanga: i pushed an update to handle the REQUIRES. i had updated the BUILD_REQUIRES and somehow missed the REQUIRES, thanks for catching that!

hoanga:
i have pushed updates that should put the recipe more in line with suggestions. i was pleasantly surprised how usable this was under Haiku (am using a dark terminal background so the default theme seems to require less knob turning for myself). thanks for setting up the upstream PR!

Begasus:
Could you remove the previous recipe for this? AFAICT it's not used by anything atm :)

Begasus (approved):

Begasus:
Upstream PR merged, this will make it easier on the next release :)

PR #6837, opened 2022-04-18
libwmf, new recipe

Begasus:


media-libs/libwmf/libwmf-0.2.12~git.recipe
```diff
@@ -0,0 +1,143 @@
...
+BUILD()
+{
+	autoreconf -fi
+	runConfigure --omit-dirs binDir ./configure \
+		--bindir=$commandBinDir \
+		--with-docdir=$docDir \
+		--with-fontdir=$dataDir/libwmf/fonts \
+		--with-xtrafontmap=$dataDir/libwmf/fonts/fontmap #\
+#		--with-gsfontdir=$fontsDir/otfonts\
+#		--with-gsfontmap=$dataDir/ghostscript/9.55.0/lib/Fontmap.GS
```
Begasus, line 116: Still a bit unsure here, documentation isn't installed, so setting this isn't needed maybe, couldn't find any fontmap on my system (ghostscript_gpl and the urw fonts installed) ...

Begasus: From looking at some other recipes at repology I also see that they remove the fonts all together because they should be provided by gsfonts (which afaik doesn't exists in haikuports)?

jmairboeck: perhaps the URW base fonts (`urw_base35_fonts`) are meant here? They come from Ghostscript IIRC.

Begasus: Thought so too, but those are installed in `$fontsDir/otfonts`, the ones in here are similar to the ones in `$fontsDir/otfonts` and don't provide a "fontmap"?
The extentions of the ones here are `t1` and `afm`, the ones in `$fontsDir/otfonts` are `otf`, then there are `afm` fonts in `$fontsDir/psfonts` ...

media-libs/libwmf/libwmf-0.2.12~git.recipe
```diff
@@ -0,0 +1,143 @@
...
+	cmd:libwmf_config$commandSuffix
+	cmd:libwmf_fontmap$commandSuffix
+	devel:libwmf$secondaryArchSuffix
+	devel:libwmf_0.2$secondaryArchSuffix = $libWMFVersionCompat
+	devel:libwmflite$secondaryArchSuffix
+	devel:libwmflite_0.2$secondaryArchSuffix = $libWMFLITEVersionCompat
+	"
+REQUIRES_devel="
+	libwmf$secondaryArchSuffix == $portVersion base
+	brotli${secondaryArchSuffix}_devel
```
korli, line 56:
```suggestion
	devel:libbrotlidec$secondaryArchSuffix
```

korli:
Should be libwmf-0.2.13~git.recipe (0.2.12 is already released)

Begasus:
> Should be libwmf-0.2.13~git.recipe (0.2.12 is already released)

Was thinking about that too, but libwmf-config --version still tells me 0.2.12, so I kept it as such?

pulkomandy:
> Was thinking about that too, but libwmf-config --version still tells me 0.2.12, so I kept it as such?

You can't play with the rules this way if you want updates to work reliably.

We don't use ~ for "fun" or arbitrary complications. The ~ means a pre-version that's before the corresponding version.

For example: 1.2.1 is after 1.2. But 1.2~1 is before 1.2

So, either you use 0.2.12.git, or you use 0.2.13~git in this case. But 0.2.12~git means this is a version that's before 0.2.12, and that's not correct. So people using the 0.2.12 release would not get this as an update.

media-libs/libwmf/libwmf-0.2.12~git.recipe
Begasus: Done :) Thanks!

Begasus:
> So, either you use 0.2.12.git, or you use `0.2.13~git` in this case. But `0.2.12~git` means this is a version that's before 0.2.12, and that's not correct. So people using the 0.2.12 release would not get this as an update.

Already got a patch in the making to bump the internal version to 0.2.13, so I'll rename the recipe after that to 0.2.13~git.

korli:
An alternative is to use 0.2.12 as version. Less headaches.

Begasus:
> An alternative is to use 0.2.12 as version. Less headaches.

On option too, but since this contains fixes from upstream and is ahead of 0.2.12 wouldn't it be strange?
Well, we can discuss this to no end, I'll rename the recipe for now to 0.2.12 ;)

Begasus:
Builds OK on 32bit and 64bit, if no objections I think it could be merged?

media-libs/libwmf/libwmf-0.2.12~git.recipe
korli: Otherwise add a TODO.

Begasus: Will add a comment there, thanks!

PR #6855, opened 2022-04-27
xpdf, bump version, fix font detection

Begasus:


Begasus:
Current version (when launched from Terminal) will produce errors not finding the fonts (courier, helvetica ...), this patchset fixes that, idea came fro archlinux: https://github.com/archlinux/svntogit-community/blob/packages/xpdf/trunk/xpdf-4.00-xpdfrc.patch

PR #6882, opened 2022-05-05
SDL2: patch to only create BApplication if one doesn't already exist

ivanholmes:
I've created this PR in response to a comment I saw in the mGBA recipe, which uses SDL 1.2 in place of SDL2 because the former checks for the existence of a BApplication before creating one whereas the latter does not.

This causes an issue with applications that use both Qt and SDL2 (such as mGBA's Qt frontend), because Qt has already created a BApplication by the time SDL2 tries to.

After adding this simple patch (the same as in the code of SDL1.2), mGBA works properly with SDL2. 

I think this change should eventually be upstreamed (if it is valid) but I thought I'd submit this PR in case there is any reason why the change hasn't been done already.


pulkomandy:
Please submit SDL2 patches upstream. It's fine to have them here until SDL2 makes a new release, but having the code upstream makes works easier for everyone involved.

ivanholmes:
> Please submit SDL2 patches upstream. It's fine to have them here until SDL2 makes a new release, but having the code upstream makes works easier for everyone involved.

I will gladly do this, the patch is so simple (and present in SDL 1.2) I just wondered if there's a reason it hasn't been done already, so I thought I would check here first

ivanholmes:
https://github.com/libsdl-org/SDL/issues/5385
On a second search I found this. Seems there may be a good reason it hasn't been done. Will investigate further though.

ivanholmes:
So for anyone who comes across this issue - this patch is not right because the BApplication used by SDL2 is a subclass SDL_BApp, which implements extra methods. 
I think this issue needs solving because I have seen multiple apps that use SDL2 and Qt at the same time, but I'm not sure how it should be solved.

Feel free to close this PR.

waddlesplash:
> Please submit SDL2 patches upstream.

This one is experimental enough that we should probably play with it here first.

> this patch is not right because the BApplication used by SDL2 is a subclass SDL_BApp, which implements extra methods.

Those extra methods should be optional and disable-able if there's an already-existing BApplication.

Our SDL1 port handles this correctly already, indeed we should make SDL2 do the same.

> perhaps by dynamically casting be_app to SDL_BApp if it already exists, something tells me that might be a bad idea but I really don't know (I don't have much experience with C++).

That is a bad idea, it won't work. Instead we should just be able to gracefully fall back and either return errors or do nothing if SDL_BApp does not exist and cannot be made.

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
```diff
@@ -1265,3 +1265,66 @@ index c5026de..868512e 100644
...
+-    int32 cookie = 0;
+-    if (get_next_image_info(B_CURRENT_TEAM, &cookie, &info) == B_OK) {
+-        BFile f(info.name, O_RDONLY);
+-        if (f.InitCheck() == B_OK) {
+-            BAppFileInfo app_info(&f);
+-            if (app_info.InitCheck() == B_OK) {
+-                char sig[B_MIME_TYPE_LENGTH];
+-                if (app_info.GetSignature(sig) == B_OK)
+-                    signature = strndup(sig, B_MIME_TYPE_LENGTH);
++    if (!be_app) {
```
waddlesplash, line 1297: Rather than indenting everything, just put this check at the start of the function. Or, better yet, do not invoke `StartBeApp` at all if `be_app` is already set.

ivanholmes: Okay, I'll keep working on it and give this a go

ivanholmes:
> > Please submit SDL2 patches upstream.
> 
> This one is experimental enough that we should probably play with it here first.

That was my thinking when creating this PR.

> > this patch is not right because the BApplication used by SDL2 is a subclass SDL_BApp, which implements extra methods.
> 
> Those extra methods should be optional and disable-able if there's an already-existing BApplication.
> 
> Our SDL1 port handles this correctly already, indeed we should make SDL2 do the same.
> 

How can I disable them? Check if be_app refers to an SDL_BApp or a plain BApplication in every place they are called?
Looking at the methods it seems many of them won't be needed in a case where Qt or another toolkit is used, which is promising. 

> > perhaps by dynamically casting be_app to SDL_BApp if it already exists, something tells me that might be a bad idea but I really don't know (I don't have much experience with C++).
> 
> That is a bad idea, it won't work. Instead we should just be able to gracefully fall back and either return errors or do nothing if SDL_BApp does not exist and cannot be made.

Alright that makes sense.

waddlesplash:
> How can I disable them? Check if be_app refers to an SDL_BApp or a plain BApplication in every place they are called?

Yes.

ivanholmes:
After looking in more detail - the extra public methods of SDL_BApp are:
reimplementation of RefsReceived - not needed if running with another toolkit
reimplementation of MessageReceived - not needed if running with another toolkit?
GetID - used by `_InitWindow` in src/video/haiku/SDL_bwindow.cc, seems not needed if running with another toolkit
ClearID - used by `HAIKU_DestroyWindow` in src/video/haiku/SDL_bwindow.cc, same as above I think
GetSDLWindow - only used by private methods of SDL_BApp
GetCurrentContext, SetCurrentContext - only used in destructor and RemoveGLView of SDL_BWin, which is not needed if using an external toolkit

If we can assume that nobody will use the SDL functions for creating windows if they are using another toolkit (native, Qt etc.) in the application (which seems to make sense), then I think there is no problem with the simple patch...

waddlesplash:
>  GetID - used by _InitWindow in src/video/haiku/SDL_bwindow.cc, seems not needed if running with another toolkit
> ClearID - used by HAIKU_DestroyWindow in src/video/haiku/SDL_bwindow.cc, same as above I think
>  GetCurrentContext, SetCurrentContext - only used in destructor and RemoveGLView of SDL_BWin, which is not needed if using an external toolkit

SDL windows may be used in tandem with non-SDL windows. We likely should make these workable even if there is no SDLBApp. Probably by making them static.

ivanholmes:
> SDL windows may be used in tandem with non-SDL windows. We likely should make these workable even if there is no SDLBApp. Probably by making them static.

Okay. I didn't notice before that `GetID` and `ClearID` use `GetSDLWindow` as well. `GetSDLWindow` is an accessor for the vector of `SDL_Window`s `_window_map` which is a private member of `SDL_BApp`, so I don't think they can be made static. I'll see if there is anything similar in SDL 1.2.

In addition `GetCurrentContext` and `SetCurrentContext` need to access the private member `_current_context`. In this case of course there will be no example in 1.2 because it doesn't use OpenGL iirc.

waddlesplash:
> which is a private member of SDL_BApp

That member should be made static along with the methods.

> In addition GetCurrentContext and SetCurrentContext need to access the private member _current_context.

This one may be a bit trickier. Possibly the _current_context shouldn't be a member variable at all.

ivanholmes:
> > which is a private member of SDL_BApp
> 
> That member should be made static along with the methods.
> 
> > In addition GetCurrentContext and SetCurrentContext need to access the private member _current_context.
> 
> This one may be a bit trickier. Possibly the _current_context shouldn't be a member variable at all.

I've made all the non-virtual methods (and the two variables we discussed) of SDL_BApp static. It compiles without complaint and I've tried a few apps/games, all work fine.

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
```diff
@@ -1265,3 +1265,145 @@ index c5026de..868512e 100644
...
+
+diff --git a/src/main/haiku/SDL_BeApp.cc b/src/main/haiku/SDL_BeApp.cc
+index a5e1428..3b0f784 100644
+--- a/src/main/haiku/SDL_BeApp.cc
++++ b/src/main/haiku/SDL_BeApp.cc
+@@ -84,10 +84,12 @@ SDL_InitBeApp(void)
+ {
+     /* Create the BApplication that handles appserver interaction */
+     if (SDL_BeAppActive <= 0) {
++    	if (!be_app) {
```
waddlesplash, line 1284: Combine this check with the previous one (i.e. `SDL_BeAppActive <= 0 && !be_app`)

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
```diff
@@ -1265,3 +1265,145 @@ index c5026de..868512e 100644
...
+ 
+-    void _PushBackWindow(SDL_Window *win) {
++    static void _PushBackWindow(SDL_Window *win) {
+         _window_map.push_back(win);
+     }
+ 
+ 
+     /* Members */
+-    std::vector<SDL_Window*> _window_map; /* Keeps track of SDL_Windows by index-id */
++    static std::vector<SDL_Window*> _window_map; /* Keeps track of SDL_Windows by index-id */
```
waddlesplash, line 1382: Looking into things more I actually think we shouldn't be using statics here. Instead we should put these in the private `driverdata` field that is present in `SDL_VideoDevice`, which is passed in as _THIS to all the implementation functions. That may be more work, though, and might require some more memory management. I can work on that if you are unsure how to do it.

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
ivanholmes: Ah yeah that would be cleaner, will do.

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
ivanholmes: I am definitely not sure how to do it, but I'll have a look - I won't learn anything otherwise!
Will let you know if/when I decide it is beyond my ability.

ivanholmes: Okay, I am really not sure how to do this. Apologies for creating work for you...

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
waddlesplash: Yeah, we shouldn't interfere with the working directory in that case.

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
waddlesplash: Well, if you want to learn, I can probably help you through it via IRC/Matrix if you want. Otherwise, it's fine, as this was on my TODO list anyway :)

ivanholmes: Ok, I definitely do want to learn, I'll log on to IRC shortly :)

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
ivanholmes: Makes sense, thanks

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
```diff
@@ -1267,143 +1267,532 @@ index ca9dcad..1ac3201 100644
...
-     if (SDL_BeAppActive <= 0) {
-+    	if (!be_app) {
-         SDL_AppThread = SDL_CreateThreadInternal(StartBeApp, "SDLApplication", 0, NULL);
-         if (SDL_AppThread == NULL) {
-             return SDL_SetError("Couldn't create BApplication thread");
-         }
-+    	}
+     /* Initialize the Be Application, if it's not already started */
+-    if (SDL_InitBeApp() < 0) {
++    if (SDL_InitBeApp(NULL) < 0) {
```
ivanholmes, line 1285: I am pretty sure this will cause a problem (literally passing NULL as the value of a pointer...). I didn't realise that `SDL_InitBeApp` was called in the audio driver code when I made it take the `SDL_VideoDevice` as an argument. Need to find a better way of making the `devicedata` visible to the BApp.

media-libs/libsdl2/patches/libsdl2-2.0.20.patchset
```diff
@@ -1267,143 +1267,532 @@ index ca9dcad..1ac3201 100644
...
+diff --git a/src/video/haiku/SDL_bwindow.cc b/src/video/haiku/SDL_bwindow.cc
+index 587affa..ca18c62 100644
+--- a/src/video/haiku/SDL_bwindow.cc
++++ b/src/video/haiku/SDL_bwindow.cc
+@@ -71,7 +71,7 @@ static int _InitWindow(_THIS, SDL_Window *window) {
+         return -1;
+ 
+     window->driverdata = bwin;
+-    int32 winID = _GetBeApp()->GetID(window);
++    int32 winID = _GetBeApp()->driverdata->GetID(window);
```
ivanholmes, line 1783: This ought to be _this->driverdata... but I had to cast driverdata to an SDL_VideoData* to get it to compile. Looking in SDL_sysvideo.h driverdata is defined as void* so that is no great surprise. Many backends call the custom struct type they use for this SDL_VideoData, so that part is ok at least.

waddlesplash:
@korli superseded by https://github.com/haikuports/haikuports/commit/d96a8717a6b5e3ec8cf9a130ea9ee79a110cf7cd ?

korli:
> @korli superseded by [d96a871](https://github.com/haikuports/haikuports/commit/d96a8717a6b5e3ec8cf9a130ea9ee79a110cf7cd) ?

Yes, unless upstream objects somehow.

waddlesplash:
The PR was merged.

ivanholmes:
Nice to see this resolved! Thanks

Issue #6689, opened 2022-03-05
Port sdl12-compat


extrowerk:
An SDL-1.2 compatibility layer that uses SDL 2.0 behind the scenes.
https://github.com/libsdl-org/sdl12-compat

pulkomandy:
As long as we don't have 3d acceleration, this is only going to make sdl1.2 software use more cpu to do the same thing (because sdl2 does everything with opengl). Not sure if it's a good idea.

waddlesplash:
It's actually quite probable that Mesa software rendering is faster than SDL1.2's software code, as it uses a JIT compiler tuned for precisely the CPU it is running on.

pulkomandy:
There is no "software code" in SDL 1.2. All it does is let the app draw into a bitmap and then copies that bitmap into the framebuffer using a BDirectWindow. If games use an 8 bit colorspace it performs a conversion, but it is pretty well optimized. I don't think OpenGL is optimized for that kind of task, which don't really need a JIT compiler, anyways.

PR #6883, opened 2022-05-05
mGBA: bump to 0.9.3 and update to use sdl2

ivanholmes:
This depends on #6882


ivanholmes:
Based on https://github.com/haikuports/haikuports/issues/6689 maybe the update to SDL2 is not desirable. The performance on my hardware (ThinkPad X230) was fine but I haven't done any in-depth evaluation of performance. 
Happy to alter the PR to update but carry on using SDL 1.2 if this is felt to be better.

ivanholmes:
Closing, see #6882
Will create another PR just for revbump

PR #6896, opened 2022-05-08
qpageview: new recipe

jmairboeck:


dev-python/qpageview/qpageview-0.6.2.recipe
```diff
@@ -0,0 +1,65 @@
...
+	qpageview = $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+PROVIDES_python38="
+	qpageview_python38 = $portVersion
+	"
+REQUIRES_python38="
+	haiku
```
augiedoggie, line 26: Seems like `haiku` would be duplicated here because it also includes `$REQUIRES` which has `haiku` in it.  Not a big deal.

augiedoggie (approved):

PR #6901, opened 2022-05-09
nanosaur: replace mini icon

ivanholmes:
The previous mini icon was scaled down from the 32x32 (large) icon. 
I had to boot up my G4 Mac mini into OS 9 to get the icons for the other Pangea ports, so I got the 16x16 icon for Nanosaur at the same time.


augiedoggie:
You should increment the REVISION in the recipe so that the package will be rebuilt.

ivanholmes:
> You should increment the REVISION in the recipe so that the package will be rebuilt.

Good point! Getting a bit late here. Done now.

PR #6905, opened 2022-05-10
ocr, new recipe

Begasus:


media-fonts/ocr/additional-files/fontconfig-relaod-cache.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+fc-cache -f
```
jmairboeck, line 3: This should probably check whether fc-cache is available before calling it because our font packages don't depend on fontconfig (it is optional on Haiku and not used by native apps IIRC).

So, probably something like this:

```sh
type fc-cache >&/dev/null && fc-cache -f
```

Begasus: If fc-cache isn't available it wouldn't do anything I guess like that?
Maybe a fontconfig requirement in REQUIRES to make sure it's installed?

Begasus: Updated per suggestion :)

jmairboeck: If fontconfig is not installed, the post-install script would fail. I don't know how the package kit would handle that case ...

Begasus: If one would put it in REQUIRES it would be downloaded and installed before the build starts, so that way you can ensure that it is installed:
```
REQUIRES="
	cmd:fc_cache
	"
```

jmairboeck: that would be an alternative, but I don't know if that is wanted in general. app_server doesn't need or use fontconfig, is that correct?

Begasus: No idea there, but trying to "uninstall" it leads to several packages wanting to uninstall (among haiku_devel package)

jmairboeck: As I said, this is probably not relevant at all for native apps (like Pe) but only for ported apps that actually use fontconfig (like xetex).

augiedoggie:
There is a typo in the name of the script "fontconfig-relaod-cache.sh"

media-fonts/ocr/additional-files/fontconfig-relaod-cache.sh
```diff
@@ -0,0 +1,3 @@
+#!sh
```
korli, line 1:
```suggestion
#!/bin/sh
```

Begasus: Got a local change for that already, will push later :)

Begasus: Fixed

Begasus:
> There is a typo in the name of the script "fontconfig-relaod-cache.sh"

Fixed, thanks :)

media-fonts/ocr/ocr-0.2.recipe
```diff
@@ -0,0 +1,48 @@
...
+REQUIRES="
+	"
+
+BUILD_REQUIRES=""
+BUILD_PREREQUIRES=""
+
+BUILD()
+{
+	true
+}
```
korli, line 35:
```suggestion
```

Begasus: Had that thought too, wasn't sure :)

Begasus: Done, PS using git_x86-2.36.1 atm here, so far working OK :)

media-fonts/ocr/additional-files/fontconfig-reload-cache.sh
```diff
@@ -0,0 +1,3 @@
+#!sh
```
korli, line 1:
```suggestion
#!/bin/sh
```

PR #6912, opened 2022-05-12
tcl/tk, bump version, enable secondary architecture

Begasus:


Begasus:
Not tested on 64bit yet, tk_x is disabled for now, tk (SDL2 version) enabled (gitk starts up on both, with tk_x there is a window but it's all distorted, with tk the window pops up for a second and then crashes with an error: `CRITICAL: Error in startup script: bad screen distance "5.0"`, not sure if it's related to SDL/tk/git atm)
Both tcl and tk* are updated to use secondary architecture on 32bit.

dev-lang/tk/tk_x-8.6.12.recipe
```diff
@@ -0,0 +1,103 @@
+SUMMARY="A windowing toolkit for use with Tcl"
```
waddlesplash, line 1: I would omit this recipe as it isn't ready for use yet. See: https://github.com/waddlesplash/xlibe/issues/2

dev-lang/tk/tk-8.6.10.recipe
```diff
@@ -93,10 +110,25 @@ INSTALL()
...
 
 	# add standard symlinks
-	ln -s $binDir/wish8.6 $binDir/wish
+	ln -s $commandBinDir/wish8.6 $commandBinDir/wish
 
-	prepareInstalledDevelLibs libtkstub8.6
+	prepareInstalledDevelLib libtkstub8.6
 	fixPkgconfig
+
+	# since we pretend to be regular tk
```
Begasus, line 118: So far this works for building the _tkinter module for python3.10, haven't had any luck so far enabling this for prior python versions

Begasus:
If there are no issues reported here I think this can be merged.

PR #6917, opened 2022-05-14
bump cantata to 2.5.0

ivanholmes:


media-sound/cantata/cantata-2.5.0.recipe
```diff
@@ -69,6 +69,7 @@ BUILD_PREREQUIRES="
 	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
 	cmd:ld$secondaryArchSuffix
+	cmd:lrelease$secondaryArchSuffix
```
augiedoggie, line 72:
```suggestion
	cmd:lrelease$secondaryArchSuffix >= 5
```
Almost all of the recipes specify a version requirement for `lrelease`.  The package built fine without this change but we might as well keep it consistent with the other recipes.

augiedoggie (changes requested):
Copyright needs an update as well.

media-sound/cantata/cantata-2.5.0.recipe
ivanholmes: Agreed, changed in latest push along with the copyright.

PR #6836, opened 2022-04-18
new recipe: lilypond

jmairboeck:
~Tests fail because of missing `extractbb` command (which should be provided by `texlive_core`).~ fixed in 29e1618 in #6818 

It also cannot be built right now as long as `texlive` is not available (should it be marked as untested?). Texlive is only required for building, not at runtime.

The libgs API is used instead of the GhostScript executable because that crashes on Lilypond's PostScript output (see #6284).

PR #6924, opened 2022-05-14
frescobaldi: new recipe

jmairboeck:
note: this doesn't technically depend on Lilypond (see #6836) but is not really useful without it. (is there a way to suggest other packages?)

MIDI playback doesn't work because [PortMidi](https://github.com/PortMidi/portmidi) is not available. That would need a dedicated implementation for Haiku.


media-sound/frescobaldi/patches/frescobaldi-3.2.patchset
```diff
@@ -0,0 +1,89 @@
...
++    "data/hunspell",
+ ]
+ 
+ 
+@@ -68,7 +69,7 @@ def settings():
+ def directories():
+     """ Yields a list of existing paths based on config """
+     # prefixes to look in for relative paths
+-    prefixes = ['/usr/', '/usr/local/']
++    prefixes = ['/usr/', '/usr/local/', '/boot/system/', '/boot/home/config/']
```
jmairboeck, line 24: should the non-packaged variants also be added here?

media-sound/frescobaldi/patches/frescobaldi-3.2.patchset
```diff
@@ -0,0 +1,89 @@
...
+diff --git a/frescobaldi_app/lilydoc/manager.py b/frescobaldi_app/lilydoc/manager.py
+index 9bb39df..4849e14 100644
+--- a/frescobaldi_app/lilydoc/manager.py
++++ b/frescobaldi_app/lilydoc/manager.py
+@@ -123,6 +123,8 @@ def urls():
+         '/usr/local',
+         '/usr/share/doc',
+         '/usr/doc',
++        '/boot/system',
++        '/boot/home/config',
```
jmairboeck, line 37: and here?

media-sound/frescobaldi/patches/frescobaldi-3.2.patchset
waddlesplash: Yes, and /usr ones removed. Same below I think.

jmairboeck: I just added the Haiku specific paths for potentially upstreaming the changes (I have not tried that yet ...). But I think for the patchset you are right, Haiku needs only the paths relevant for it.

jmairboeck: There are a few more places where I added the Haiku specific paths instead of replacing the existing ones. Should these be changed too?

PR #6927, opened 2022-05-17
mc: use slang, fix deps, improve Haiku patches

augiedoggie:


app-misc/mc/mc-4.8.28.recipe
```diff
@@ -49,11 +51,14 @@ BUILD_REQUIRES="
 	devel:libiconv$secondaryArchSuffix
 	devel:libintl$secondaryArchSuffix
 	devel:libncursesw$secondaryArchSuffix
```
korli, line 53: What's the point of having libncurses?

augiedoggie: It provides some termcap functions that slang uses. Without it trying to run `mc` results in an unresolved symbol error.  I noticed that `mc` is linked to both ncurses and slang on my Linux box.

korli: at least https://archlinux.org/packages/community/x86_64/mc/ doesn't list ncurses in the soname list

Begasus: Maybe because slang is used first if found?
https://github.com/MidnightCommander/mc/blob/7fe3f4c898d029e87faa439ebd7297214059f710/doc/INSTALL#L207

augiedoggie: > at least https://archlinux.org/packages/community/x86_64/mc/ doesn't list ncurses in the soname list

I know and I haven't figured out where the problem is.  Running `ldd` on my Linux box shows both ncurses and slang...
```
ldd /bin/mc
        linux-vdso.so.1 (0x00007ffd97908000)
        libslang.so.2 => /usr/lib/libslang.so.2 (0x00007efd9bfca000)
        libgpm.so.2 => /usr/lib/libgpm.so.2 (0x00007efd9bfc2000)
        libe2p.so.2 => /usr/lib/libe2p.so.2 (0x00007efd9bfb6000)
        libssh2.so.1 => /usr/lib/libssh2.so.1 (0x00007efd9bf75000)
        libgmodule-2.0.so.0 => /usr/lib/libgmodule-2.0.so.0 (0x00007efd9bf6e000)
        libglib-2.0.so.0 => /usr/lib/libglib-2.0.so.0 (0x00007efd9be30000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007efd9bc21000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007efd9bc1c000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007efd9bb35000)
        libncursesw.so.6 => /usr/lib/libncursesw.so.6 (0x00007efd9bac1000)
        libssl.so.1.1 => /usr/lib/libssl.so.1.1 (0x00007efd9ba2a000)
        libcrypto.so.1.1 => /usr/lib/libcrypto.so.1.1 (0x00007efd9b749000)
        libz.so.1 => /usr/lib/libz.so.1 (0x00007efd9b72d000)
        libpcre.so.1 => /usr/lib/libpcre.so.1 (0x00007efd9b6b6000)
        /lib64/ld-linux-x86-64.so.2 => /usr/lib64/ld-linux-x86-64.so.2 (0x00007efd9c435000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007efd9b6b1000)
```

augiedoggie: I believe the Arch slang package is missing a dependency on ncurses, not for the library but for the terminfo database.  I have successfully built this package without a direct link to libncurses by redoing parts of the slang patch but `mc` will still have an indirect dependency on ncurses for the terminfo files.

korli: interessant. It might be a good idea to add a PROVIDES for the terminfo db in ncurses. Then one can depends on it.

augiedoggie: > It might be a good idea to add a PROVIDES for the terminfo db in ncurses. Then one can depends on it.

Yes, but it appears to be broken at the moment and I am not interested in working on that recipe at all.

PR #6854, opened 2022-04-27
gdal: add recipe for 3.4.2

gillins:
Update to latest version. Also added `curl` and `geos` support to enable more functionality. 

I think I've incorporated all the recent changes to the existing recipe.

Please let me know if anything needs changing.

PR #6930, opened 2022-05-20
opencv: update recipe for 4.5.5

gillins:
As mentioned in https://github.com/haikuports/haikuports/pull/6854 `opencv` needs to be rebuilt as part of `gdal` version bump.


extrowerk:
Builds fine on 64 bit.

Begasus:
With this some packages also needs looking into: https://github.com/haikuports/haikuports/search?q=lib%3Alibopencv

PR #6946, opened 2022-05-26
lua*, rename recipes, fix conflict for cmd's in lua5.3 and lua5.4

Begasus:


Begasus:
Should be OK

Begasus:
Lua5.3 has seen about the EOL (last release 2 years ago), when lua5.4 becomes default some changes would be needed

PR #6947, opened 2022-05-26
latex2html, new recipe

Begasus:


dev-tex/latex2html/latex2html-2022.recipe
```diff
@@ -0,0 +1,83 @@
...
+	cmd:texexpand = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	netpbm
+	texlive_core$secondaryArchSuffix
```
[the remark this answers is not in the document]
jmairboeck: Just specifying one of them would also implicitly add all requirements, as long as the packages are not split. But using `cmd:foo` is usually preferred to raw package names, I thought.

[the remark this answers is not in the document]
Begasus: > But using `cmd:foo` is usually preferred to raw package names, I thought.

Granted, I'll update in a bit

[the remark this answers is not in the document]
Begasus: > I was surprised at first that you had it enabled at all ;). I didn't check it, but I would not be surprised if gcc2 can't build it.

Didn't even try it (as texlive_core is gcc11 only) ;)

dev-tex/latex2html/latex2html-2022.recipe
```diff
@@ -0,0 +1,83 @@
...
+	1997 Marek Rouchal
+	1996-1999 Jens Lippmann
+	1995-2002 Ross More"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/latex2html/latex2html/archive/refs/tags/v$portVersion.tar.gz"
+CHECKSUM_SHA256="9b3ba484226a2e39fb20695729370372b355f71bd65eeb4dd14f2c78699ed59a"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
```
jmairboeck, line 39:
```suggestion
ARCHITECTURES="any"
```

The files in $binDir are only perl scripts, not actual binaries, and the rest are also perl modules and tex data files.

dev-tex/latex2html/latex2html-2022.recipe
```diff
@@ -0,0 +1,119 @@
...
+	cmd:dvipng
+	cmd:ps2pdf
+	cmd:pdfcrop
+	devel:libgif$secondaryArchSuffix
+#	devel:libgs$secondaryArchSuffix
+	devel:libpng16$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:cmp
+	cmd:gcc$secondaryArchSuffix
```
jmairboeck, line 97:
```suggestion
```

dev-tex/latex2html/latex2html-2022.recipe
```diff
@@ -0,0 +1,119 @@
...
+	cmd:make
+	cmd:perl
+	"
+
+BUILD()
+{
+	runConfigure --omit-dirs "binDir dataRootDir docDir" ./configure \
+		--bindir=$prefix/bin \
+		--shlibdir=$libDir/latex2html \
+		--with-tex=$dataDir/texlive/texmf-dist/tex/latex/html
```
jmairboeck, line 107:
```suggestion
	runConfigure --omit-dirs "dataRootDir docDir" ./configure \
		--shlibdir=$libDir/latex2html \
		--with-texpath=$dataDir/texlive/texmf-dist/tex/latex/html
```
You could probably also use $dataDir/latex2html for --shlibdir, not sure what is better there.

dev-tex/latex2html/latex2html-2022.recipe
```diff
@@ -0,0 +1,119 @@
...
+	1995-2002 Ross More"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/latex2html/latex2html/archive/refs/tags/v$portVersion.tar.gz"
+CHECKSUM_SHA256="9b3ba484226a2e39fb20695729370372b355f71bd65eeb4dd14f2c78699ed59a"
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	latex2html$secondaryArchSuffix = $portVersion
```
jmairboeck, line 41: remove all references to $secondaryArchSuffix (not used without SECONDARY_ARCHITECTURES)

dev-tex/latex2html/latex2html-2022.recipe
```diff
@@ -0,0 +1,119 @@
...
+	cmd:latex2html = $portVersion
+	cmd:pstoimg = $portVersion
+	cmd:texexpand = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
```
jmairboeck, line 47: this needs at least `cmd:perl`, probably more (texlive etc.)

jmairboeck, line 47: this needs at least `cmd:perl`, probably more (texlive etc.)

dev-tex/latex2html/latex2html-2022.recipe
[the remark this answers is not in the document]
Begasus: Had been tinkering about that too, not sure what the correct way would be, I guess I should check this on linux and see how texlive etc are packaged compared to this?

PR #6953, opened 2022-05-28
mc: use slang, fix deps, improve Haiku patches

### LINKED ISSUE #6849: [Midnight Commander] Suggestion: switch from ncurses to slang-2.3.2
Now that Haiku's Terminal [supports 24 bit colors](https://git.haiku-os.org/haiku/log/?qt=range&q=hrev56010..hrev56011), maybe Midnight Commander could switch from being built with `--with-screen=ncurses` to `--with-screen=slang` in order to be able to use [true-color themes](https://midnight-commander.org/ticket/3724).

[This issue](https://dev.haiku-os.org/ticket/17719) might get in the way, but seems fairly minor.

BTW... any idea about why box-drawing characters are not used/enabled/working on the current `mc` Haiku port?

augiedoggie:
fixes #6849


augiedoggie:
Redone to remove the libncurses dependency(or any direct ncurses dependency)

PR #6969, opened 2022-05-30
picard: remove patches

phw:
No longer needed since recent fixes to Haiku's wcsxfrm implementation in https://github.com/haiku/haiku/commit/218604196afb8f7bc142ab94af6e9cafdf5327b2


phw:
@waddlesplash Thanks for the fix in libroot. This issue was driving me crazy, but I also never noticed that Python's `locale.strxfrm`  actually uses `wcsxfrm` and not `strxfrm`, so at first that fix seemed unrelated to my problems. In retrospection this now makes sense ;)

korli:
@phw  you want to keep this patch until the builder switched to a Haiku version with the wcsxfrm bugfix.

PR #6974, opened 2022-06-04
libgit2, bump version, drop x86_gcc2

Begasus:


Begasus:
Part of the patchset is created upstream by @nielx (still waiting to be merged), other part is still from the previous versions, @pulkomandy is that part still needed if gcc2 isn't involved anymore?

pulkomandy:
I don't know, if you don't get errors about "free_func" when building, it's possible that the issue is already solved elsewhere (maybe a zlib or openssl update), or maybe for some reason it only affected gcc2, in that case it can be removed.

Begasus:
> I don't know, if you don't get errors about "free_func" when building, it's possible that the issue is already solved elsewhere (maybe a zlib or openssl update), or maybe for some reason it only affected gcc2, in that case it can be removed.

Thanks, I'll do another round of building it without that part of the pachset. :)

PR #6980, opened 2022-06-07
sys-process/htop: update to Haiku-supported version 2.3.0

Low-power:
The original version 2.0.2 builds, but almost completely useless on Haiku; as it didn't contain any Haiku-specific code, but uses the `unsupported` skeleton code, which can't be used to view system information and process list.

This change updates sys-process/htop to my customized version, which actually works and useful for Haiku.

Has been tested on Haiku R1/beta3.


pulkomandy:
Nice! Do you plan to submit your changes to the original htop developers?

Low-power:
> I think we would prefer to keep Haiku-support patches in-tree in HaikuPorts as a patchset instead of using a forked version where it is generally possible.

Well, since I finished the Haiku support in my fork in an early Haiku release first, before I discovered the haikuports, the code or patch is written for the fork and therefore would not easily applying to other version.

Begasus:
The original repository has been archived, any reason to keep us from merging? (PS, didn't check a 32bit build so far)

PR #6981, opened 2022-06-08
sane_backends: fixed x86 build

kenmays:
Tested on Haiku R1B3 x86.

### DIFF
--- a/media-gfx/sane_backends/sane_backends-1.1.1.recipe
+++ b/media-gfx/sane_backends/sane_backends-1.1.1.recipe
@@ -82,6 +82,7 @@ BUILD_PREREQUIRES="
 	cmd:libtoolize
 	cmd:make
 	cmd:pkg_config$secondaryArchSuffix
+	cmd:python3
 	"
 
 BUILD()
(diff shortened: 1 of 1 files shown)


media-gfx/sane_backends/sane_backends-1.1.1.recipe
```diff
@@ -41,7 +41,6 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	cmd:python3$secondaryArchSuffix
```
Begasus, line 44: AFAICT python doesn't provide a $secondaryArchsuffix, if this is required then it would be:
```
cmd:python3
```

Begasus: Is it required at runtime or build-time? (a bit confused atm) :)

kenmays: Build.

Begasus: Will merge, watch buildmaster on eventual failed builds.

PR #6982, opened 2022-06-09
cargo-c, bump version

Begasus:


Begasus:
Needs check on 64bit

nielx:
This is also a known issue. What's happening is that libssh2-sys is building its own copy of the SSH2 library, and it is linking against libcurl.so which dynamically links to libssh2.so. Try to set the LIBSSH2_SYS_USE_PKG_CONFIG environment variable.

dev-rust/cargo_c/cargo_c-0.9.10.recipe
```diff
@@ -25,40 +22,44 @@ PROVIDES="
...
 
 BUILD()
 {
-	cp $sourceDir2/Cargo.lock Cargo.lock
+	pkg-config --libs --cflags libgit2 'libgit2 >= 1.4.0' 'libgit2 < 1.5.0'
+
+	cargo generate-lockfile
 
 	cargo fetch --locked
+	cargo build --release --frozen
```
Begasus, line 53: @nielx could this part be skipped, from what I see the part in "INSTALL" takes care of it, running it here and in INSTALL seems to build it twice?

Begasus: PS, seems 64bit build was fine, now the install part is still running.

nielx: So what's necessary is more about style than it is about what you need. Technically, all you need is `cargo install cargo-c` to fetch the source, fetch all dependencies, build and install. Let's call this the **supersimple** variant. However, there are downsides, mainly that the dependencies are resolved at the time of building, which has the potential advantage of getting the latest compatible versions with the most recent bugfixes, but the distinct disadvantage that you download newer dependencies that break the build for whatever reason.

So for the purposes of having a stable, repeatable build, you want the dependencies to be stable. There are two variants here. One is the variant where as part of the patching of the source tree, we provide a Cargo.lock file that was generated. A lockfile 'locks' the dependencies, and therefore gives you reasonably repeatable builds. The only potential issue you might run into then, is when an original crate is 'yanked' and no longer exists on crates.io. That should be rare though, and can be fixed by updating Cargo.lock.

The alternative can be found in various other recipes that use the cargo build system. In this variant, all the dependencies are individually specified as source files to be downloaded by Haikuporter. I call this the **fundamentalist** approach, because in my opinion it is driven by the ideology that the packager needs to be in charge of everything that goes into the recipe, and also that build and/or install steps should not download files from the internet. I am doubtful of the added benefit of this approach, but there are scripts available in a PR for Haikuporter that help automate this, which is useful for packages that have tenths if not hundreds of dependencies.

Begasus: Had been playing around for a recipe for git2 while adding several crates in the list as some of the other packages do, around 50 crates I gave up on that idea.
cargo-c fetches about 134 crates if I look at the build process and compiles 198 packages  in the end.
Still not giving up yet. First attempt would be to use the Cargo.lock file from upstream, if that also fails maybe start adding the individual crates as in some of the current used recipes available. (Gentoo seems to do this also, https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-util/cargo-c/cargo-c-0.9.9.ebuild)

Begasus: Giving up for now, need to get away from this for some time ;)

kenmays:
This is fixed for Haiku nightly build (tested on Haiku hrev57134 x86).

PR #6989, opened 2022-06-14
Updated Android File Transfer to version 4.2

jsteinaker:
Bumped to version 4.2. Had to remove that readline thing on CMakeLists.txt because it was trying to include it even when it didn't find readline. There's probably a better way to resolve this, but I don't really know CMake, I'm open to suggestions.

4.2 is working for me with the same hit/miss ratio than 3.9 currently on the repo (it will work fine one time and throw a kernel panic the next one). Libusb and Haiku's USB stack need a lot of love.


waddlesplash:
Is the kernel panic already reported?

jsteinaker:
It wasn't, for a moment I tought it could be the same one I reported one or two days ago using the phone as a tethering device, but they seem to be different. Now reported on https://dev.haiku-os.org/ticket/17799

Begasus:
> Libusb and Haiku's USB stack need a lot of love.

This seems a known issue, even reported before the build starts, Gentoo disables it: https://gitweb.gentoo.org/repo/gentoo.git/tree/sys-fs/android-file-transfer-linux/android-file-transfer-linux-4.2.ebuild#n75

Doing some checks here too, ps, for readline I think you need both libncursesw and libreadline added to the recipe?

pulkomandy:
I have a version of this on my machine where I have replaced libusb with a native backend talking directly with our USBKit. This seems to work much more reliably for me.

I have also patched it to build and run with libedit instead of readline, since we ship libedit by default in Haiku.

I did not release it yet because I wanted to get it working with FUSE and UserlandFS, however I found that it needs some APIs that we don't implement yet in our FUSE. But the CLI is working. I did not try the GUI in my version.

Remind me to upload my changes somewhere next time I power up my computer at home.

pulkomandy:
I have opened a pull request to upstream with my changes: https://github.com/whoozle/android-file-transfer-linux/pull/309/files

jsteinaker:
Beautiful. I think we should wait to see what happens with this upstream, since 4.2 didn't seem to bring anything new to the table to us at least.

Begasus:
> I have opened a pull request to upstream with my changes: https://github.com/whoozle/android-file-transfer-linux/pull/309/files

Shouldn't `develop/headers/x86/editline/readline.h` etc. be added to: https://github.com/whoozle/android-file-transfer-linux/pull/309/files#diff-37259873c84f7a0675df28102b4973148ea067de728b150e0883e893d79f855c ?

pulkomandy:
Possibly yes, I didn't check the 32bit build for this.

waddlesplash:
Upstream change was merged, so this PR can be updated with that now.

Begasus:
Created a PR to support detection on 32bit for libreadline/libedit at: https://github.com/whoozle/android-file-transfer-linux/pull/312

Begasus:
@jsteinaker could you bump to a recent commit (srcGitRev) for this?

jsteinaker:
@Begasus updated. I did a few minor fixes as well, one of them should probably go upstream (the use of a readline variable which is non-set when readline is not detected, probably no one noticed since most systems have readline). But, and there's always a but, it's not working for me. The first run seems alright, detects the phone (I can see it on the terminal output) and shows the "retry" window, since I have to give permission from my phone. Problem is after I give permission and hit the retry button it hangs on the message "Control read(type 80 req 6 value 300 index 0 size 255) with timeout 1000…" and does nothing. I kill the app, but from that point on the behaviour is the same until I restart the machine, and so I get the phone detected again (just for one time).

Can someone give a try? Could be a bug I'm facing and you're not.

I checked the source code and that message comes from the haiku backend @pulkomandy has written, so the good thing is that it is being used (and the problem should be on our side).

pulkomandy:
Timeouts are not implemented correctly. So it just waits forever instead of timing out after a second.

You don't need to restart the whole machine in this case, just disconnect and reconnect the phone should be enough.

waddlesplash:
Should we merge this as-is? Or are there other changes needed?

Begasus:
Build is fine on 32bit, merging.

pulkomandy:
This version should not be labelled as 4.3, there is no 4.3 release yet. It could be named 4.3~git instead to note that it is a pre-release. Otherwise, what will we do when an actual version 4.3 is released?

PR #7009, opened 2022-06-29
krb5: Added recipe for krb5-1.20

trungnt2910:
Recipe for krb5-1.20 added.

Build and tests are running well on x86_64 and x86_gcc2 (with secondary x86 architecture). Building on gcc2 is currently failing due to the compiler not supporting some flags.

Thanks to @jessicah for some of the patches in order to get this building properly.

I haven't made any recipe for more than 2 years, sorry if I did something wrong...

### DIFF
--- a/net-misc/krb5/krb5-1.20.recipe
+++ b/net-misc/krb5/krb5-1.20.recipe
@@ -1,0 +1,148 @@
+SUMMARY="Kerberos: The Network Authentication Protocol"
+DESCRIPTION="Kerberos is a network authentication protocol. It is designed to \
+provide strong authentication for client/server applications by using secret-key \
+cryptography. A free implementation of this protocol is available from the \
+Massachusetts Institute of Technology. Kerberos is available in many commercial \
+products as well."
+HOMEPAGE="https://web.mit.edu/kerberos/"
+COPYRIGHT="1985-2022 by the Massachusetts Institute of Technology and its contributors"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://kerberos.org/dist/krb5/$portVersion/krb5-$portVersion.tar.gz"
+CHECKSUM_SHA256="7e022bdd3c851830173f9faaa006a230a0e0fdad4c953e85bff4bf0da036e12f"
+PATCHES="krb5-$portVersion.patchset"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+GLOBAL_WRITABLE_FILES="
+	settings/krb5.conf auto-merge
+	settings/gss directory auto-merge
+	var/krb5kdc directory auto-merge
+	"
+
+PROVIDES="
+	krb5$secondaryArchSuffix = $portVersion
+	cmd:compile_et$secondaryArchSuffix
+	cmd:gss_client$secondaryArchSuffix
+	cmd:gss_server$secondaryArchSuffix
+	cmd:k5srvutil$secondaryArchSuffix
+	cmd:kadmin$secondaryArchSuffix
+	cmd:kadmin.local$secondaryArchSuffix
+	cmd:kadmind$secondaryArchSuffix
+	cmd:kdb5_util$secondaryArchSuffix
+	cmd:kdestroy$secondaryArchSuffix
+	cmd:kinit$secondaryArchSuffix
+	cmd:klist$secondaryArchSuffix
+	cmd:kpasswd$secondaryArchSuffix
+	cmd:kprop$secondaryArchSuffix
+	cmd:kpropd$secondaryArchSuffix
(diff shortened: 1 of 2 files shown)


net-misc/krb5/krb5-1.20.recipe
```diff
@@ -0,0 +1,138 @@
...
+
+GLOBAL_WRITABLE_FILES="
+	settings/krb5.conf auto-merge
+	settings/gss directory auto-merge
+	var/krb5kdc directory auto-merge
+	"
+
+PROVIDES="
+	krb5$secondaryArchSuffix = $portVersion
+
```
Begasus, line 26: No empty line here.

net-misc/krb5/krb5-1.20.recipe
```diff
@@ -0,0 +1,138 @@
...
+	cmd:kswitch$secondaryArchSuffix
+	cmd:ktutil$secondaryArchSuffix
+	cmd:kvno$secondaryArchSuffix
+	cmd:sclient$secondaryArchSuffix
+	cmd:sim_client$secondaryArchSuffix
+	cmd:sim_server$secondaryArchSuffix
+	cmd:sserver$secondaryArchSuffix
+	cmd:uuclient$secondaryArchSuffix
+	cmd:uuserver$secondaryArchSuffix
+
```
Begasus, line 54: No empty line here.

net-misc/krb5/krb5-1.20.recipe
```diff
@@ -0,0 +1,138 @@
...
+{
+	cd src
+
+	make install
+
+	mkdir -p $settingsDir/gss
+	cp $dataDir/examples/krb5/krb5.conf $settingsDir/krb5.conf
+	awk '{sub("/usr/local/var", "/var")}1' $dataDir/examples/krb5/kdc.conf > $localStateDir/krb5kdc/kdc.conf
+
+	fixPkgconfig
```
Begasus, line 129: I'm pretty sure this will error out when running haikuporter in strict mode, think you need to add `prepareInstalledDevelLibs` (and the names of the libraries) `before fixPkgconfig.`

Begasus (dismissed):

Begasus:
Haven't checked the build, but left a few comments.

net-misc/krb5/krb5-1.20.recipe
trungnt2910: `prepareInstalledDevelLibs` breaks the build as the binaries included depend on the unversioned symlink of the libraries.

I've checked the recipe in strict mode (`haikuporter -S`) and it gave no errors.

pulkomandy: This probably means the libraries don't have a soname set or it is not set correctly. This can result in problems if the library ABI changes, it will not be possible to install multiple versions of it side-by-side.

If the soname is correctly set to the versionned name of the library, the binaries will not need the unversionned files anymore.

trungnt2910: Yeah, I can see the soname problem now.
Fixed it, will test a recipe with the new patchset later.

trungnt2910: Checked, everything is working.

I initially thought the linked name was the behavior of the project itself, but it was actually caused by a faulty patch...

extrowerk:
Huh, i recall itried to port kerberos to Haiku, but i don1t remember anymore why. Maybe for the GSSAPI support for mutt?

trungnt2910:
Python tests are mysteriously failing now. I'll reopen this after testing and fixing any potential bugs from a clean Haiku image.

trungnt2910:
Nope, this is Haiku's fault for failing to resolve the hostname "shredder".

trungnt2910:
The recipe is ready. Build and tests run well in HaikuPorter.

A summary of the patchset included:
1. Regular `autoconf` stuff for Haiku and some misc header patches to get builds working (this is similar to many other recipes).
2. Some workarounds for Haiku's POSIX networking functions weird behavior, including:
- `gethostbyname` and `getaddrinfo` does not resolve the local hostname (usually "shredder") on Haiku. (It is said to be related to a lack of mDNS support on this platform).
- `bind` does not fail when being used on a socket that has already been `bind`-ed, but calls `unbind` and re-`bind`s the socket to the new address.
- `select` fails when passing `FD_SETSIZE` as the first argument. Currently, `FD_SETSIZE` is 1024 while in the kernel, `select` fails if the first argument is greater than the fd table size, usually 256.
- Most Haiku functions that take `struct iovec*` as an argument do not accept `struct iovec` instances with `iov_base` set to `NULL`.

I don't know if some stuff in the second group should be fixed in Haiku instead, but at least with these patches krb5 works normally.

Begasus:
Thanks!

PR #7010, opened 2022-06-29
agar, bump version

Begasus:


Begasus:
Disabled for now, can't find a way to run some tests here.
manpages are not installed (they do when running from Terminal), no clue where to fix this.

PR #7020, opened 2022-07-04
build, new python recipe

Begasus:


Begasus:
Seems to be working ok, so far only for python3.7

Begasus:
Depends on: https://github.com/haikuports/haikuports/pull/7022

PR #7021, opened 2022-07-04
installer, new python recipe

Begasus:


Begasus:
Included upstream (waiting to merge) PR: https://github.com/pypa/installer/pull/103 as patch, works ok, only downside is that it still insists on installing in site-packages instead of vendor-packages (could be related to our python port?)

Begasus:
Some explanation, in the screenshot you can see in the cutout tomli build with the `destdir` destination, this will fail as the packages are not moved to the correct location, in the middle you see tomli package build with the `prefix` destination, that way it installs in the correct destination but in `site-packages` rather then `vendor-packages`, in the recipe for tomli I've moved them to the correct path and removed the site-packages from the package as you can see in the right package for tomli.

python-installer2

PR #7024, opened 2022-07-05
flit-core, new python recipe

Begasus:


Begasus:
Depends on:
- [ ] installer https://github.com/haikuports/haikuports/pull/7021

PR #6741, opened 2022-03-19
LLVM: bump

extrowerk:
Applied 12.0.1 patches, some already merged upstream, therefore unnecessary.
Build fails.

PR #7036, opened 2022-07-06
llvm-libgcc: Add recipe.

trungnt2910:
Added recipe for `llvm-libgcc`, based on LLVM 14.0.6.
This recipe might be a bit controversial, so I'll write a detailed description here.

## What is this?
From the official [description](https://github.com/llvm/llvm-project/blob/main/llvm-libgcc/docs/LLVMLibgcc.rst#motivation):
> To solve this problem, libunwind needs libgcc "front" that is, link the necessary functions from compiler-rt and libunwind into an archive and shared object that advertise themselves as libgcc.a, libgcc_eh.a, and libgcc_s.so, so that glibc’s baked calls are diverted to the correct objects in memory. Fortunately for us, compiler-rt and libunwind use the same ABI as the libgcc family, so the problem is solvable at the llvm-project configuration level: no program source needs to be edited. Thus, the end result is for a distro manager to configure their LLVM build with a flag that indicates they want to archive compiler-rt/unwind as libgcc. We achieve this by compiling libunwind with all the symbols necessary for compiler-rt to emulate the libgcc family, and then generate symlinks named for our "libgcc" that point to their corresponding libunwind counterparts.

Basically, this is a library containing LLVM's `libunwind` and `compiler-rt`, and is binary compatible to GCC's `libgcc_s.so`. It can replace the existing library from `gcc_syslibs` without affecting any Haiku programs.

## Why is `llvm-libunwind` needed?
Projects such as the [.NET runtime](https://github.com/dotnet/runtime/issues/55803) [requires](https://github.com/dotnet/runtime/issues/55803#issuecomment-884804726) an unwinding library, such as HP libunwind or LLVM libunwind. These libraries provide a wider interface than libgcc_s.

## Why should we replace `libgcc_s.so.1`? Why can't `libunwind` stand on its own?
To acquire unwind information, libunwind [tries to](https://github.com/llvm/llvm-project/blob/511a7eef937ebc1e25efc90bf41f12b52ad1dff3/libunwind/src/AddressSpace.hpp#L429) find a `PT_GNU_EH_FRAME` header by [iterating](https://github.com/llvm/llvm-project/blob/511a7eef937ebc1e25efc90bf41f12b52ad1dff3/libunwind/src/AddressSpace.hpp#L489) through the program headers of loaded objects, using a [`dl_iterate_phdr`](https://github.com/llvm/llvm-project/blob/511a7eef937ebc1e25efc90bf41f12b52ad1dff3/libunwind/src/AddressSpace.hpp#L605) call. From the info provided in this header, the unwind libraries can work out the address of the `.eh_frame` section.

However, most Haiku binaries do not contain a `PT_GNU_EH_FRAME` (or [`PT_EH_FRAME`](https://xref.landonf.org/source/xref/haiku/headers/os/kernel/elf.h#347), as defined in Haiku headers). To work around this, during initialization, Haiku binaries call [`__register_frame_info`](https://github.com/haiku/buildtools/blob/493aaad4dc61224ff101425c283a36960fb6841b/gcc/libgcc/crtstuff.c#L487), a function declared in `libgcc_s.so.1`, to tell libgcc_s about the [location of the `.eh_frame` section](https://github.com/haiku/buildtools/blob/493aaad4dc61224ff101425c283a36960fb6841b/gcc/libgcc/crtstuff.c#L265).

This way, libgcc_s (and **only** libgcc_s), knows enough frame information for unwind operations (exception handling and stuff like that). Other libraries have no means of obtaining the same information on Haiku (only the `__register_frame_info` in libgcc_s is called by Haiku binaries). This makes libunwind, whether standalone as `libunwind.so` or statically linked to other binaries, virtually useless (it cannot unwind without frame information).

## Why can't other approaches be used?
Currently, another approach exists: Instead of linking to `libgcc_s.so.1`, Haiku can have its `libroot.so` dynamically (or even statically, as in [this patch](https://review.haiku-os.org/c/haiku/+/5105)) linked to libunwind instead. This causes a few problems:
- Haiku is currently stuck on C++98 for GCC2 compatibility, while libunwind, as a part of LLVM, requires [C++14](https://llvm.org/docs/CMake.html#rarely-used-cmake-variables) or higher.
- Only support for x86_64 is implemented, support for other architectures is still pending.
Following this approach makes the Haiku source split into two groups: Using libgcc_s.so on unsupported platforms, and statically link to libunwind on x86_64. This makes it hard to maintain the codebase.

In contrast, as a HPKG, llvm-libgcc can be installed on supported platforms when needed, and also uninstalled to restore the original libgcc.

## Why is this recipe separate from the LLVM [recipe](https://github.com/haikuports/haikuports/tree/master/sys-devel/llvm)?
**1. Flexibility**
`llvm-libgcc` is made of `libunwind` and `compiler-rt`, both of which are fairly portable. It is easy to build on Haiku (CompilerRT requires no patches), and does not depend on much of the LLVM infrastructure. `llvm-libgcc` based on LLVM 14.0.6 can already be built on Haiku, while there is much more work to be done for [the rest of LLVM](https://github.com/haikuports/haikuports/pull/6741).
Freeing `llvm-libgcc` from the rest of LLVM allows this library to be updated more often, and be built more quickly and with less disk space.
**2. Build requirements**
`llvm-libgcc` "requires being built in a monorepo layout with libunwind available", according to the [sources](https://github.com/llvm/llvm-project/blob/cf7502a1ebe635c10e69ff70bc590313474b5277/llvm-libgcc/CMakeLists.txt#L4). However, the [recipe](https://github.com/haikuports/haikuports/blob/e6ec930ae2f8729cfee41f20c68036b945d6afda/sys-devel/llvm/llvm12-12.0.1.recipe#L528-L538) in HaikuPorts builds up a weird repo format, with the `llvm` folder as the root, `llvm/tools` for additional tools, and `llvm/projects` for additional projects. It is hard to build `llvm-libgcc` in this directory format.

Furthermore, `llvm-libgcc` is a special library that requires being built [separately](https://github.com/llvm/llvm-project/blob/f8e026457e5100a9144b5864d17e8c8b3ae7cd24/runtimes/CMakeLists.txt#L207) from both CompilerRT and libunwind. The existing recipe is already [building](https://github.com/haikuports/haikuports/blob/e6ec930ae2f8729cfee41f20c68036b945d6afda/sys-devel/llvm/llvm12-12.0.1.recipe#L46) CompilerRT, making `llvm-libgcc` incompatible.

## Why are there so many warnings about missing tools and libraries when building?
Tools such as `cmd:git` or libraries such as `devel:zlib` is required for building other parts of LLVM (versioning and stuff,...) but is not required for libunwind or CompilerRT. `llvm-libgcc` builds fine and functions properly with the configuration in this recipe.

## Why is it only available on x86_64?
While libunwind supports a wide range of platforms, Haiku-specific signal frame unwinding is currently only available for x86_64.

## How is the package intended to be used? How does it affect Haiku and programs that run on Haiku?
This package is not intended for users to install directly. It will only be pulled as a dependency of any application that needs LLVM's libunwind (such as the .NET runtime, if it gets properly ported). Installing the package should not affect any other programs (all `libgcc_s.so.1` functions are properly implemented). The package can be uninstalled to restore the official libgcc_s.

## What does the patchset contain?
1. The initial LLVM libunwind port for Haiku, done by @X547, as in [this Haiku patch](https://review.haiku-os.org/c/haiku/+/5105). It has been rebased on libunwind version 14.0.6. Most of this patch is trivial configuration, except for the signal frame unwinding support on Haiku x86_64 in `UnwindCursor.hpp`.
2. LLVM's [official](https://github.com/llvm/llvm-project/commit/c5a20b518203613497fa864867fc232648006068) `llvm-libgcc` support. This commit is already on LLVM's trunk, probably ready for release in LLVM vNext, but it is not available on LLVM 14.0.x. However, it is quite easy to backport the patch to earlier versions ([this branch](https://github.com/trungnt2910/llvm-project/tree/dev/trungnt2910/test-libgcc-v12) even contains the patch backported to LLVM 12).
3. Some misc configuration in the build system for Haiku.
4. An improved implementation in libunwind, for platforms that uses the EH frame registry (such as Haiku). This prevents the long initialization and de-initialization times for binaries that X512's initial libunwind port introduces.

Thanks for reviewing this pull request. If there's anything wrong, I'll try to fix as soon as possible.

### DIFF
--- a/sys-devel/llvm-libgcc/licenses/Apache v2 with LLVM Exception
+++ b/sys-devel/llvm-libgcc/licenses/Apache v2 with LLVM Exception
@@ -1,0 +1,279 @@
+==============================================================================
+The LLVM Project is under the Apache License v2.0 with LLVM Exceptions:
+==============================================================================
+
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+    1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
+      "Contribution" shall mean any work of authorship, including
+      the original version of the Work and any modifications or additions
+      to that Work or Derivative Works thereof, that is intentionally
+      submitted to Licensor for inclusion in the Work by the copyright owner
+      or by an individual or Legal Entity authorized to submit on behalf of
+      the copyright owner. For the purposes of this definition, "submitted"
+      means any form of electronic, verbal, or written communication sent
+      to the Licensor or its representatives, including but not limited to
+      communication on electronic mailing lists, source code control systems,
+      and issue tracking systems that are managed by, or on behalf of, the
+      Licensor for the purpose of discussing and improving the Work, but
+      excluding communication that is conspicuously marked or otherwise
+      designated in writing by the copyright owner as "Not a Contribution."
+
+      "Contributor" shall mean Licensor and any individual or Legal Entity
+      on behalf of whom a Contribution has been received by Licensor and
+      subsequently incorporated within the Work.
+
+    2. Grant of Copyright License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      copyright license to reproduce, prepare Derivative Works of,
+      publicly display, publicly perform, sublicense, and distribute the
+      Work and such Derivative Works in Source or Object form.
+
+    3. Grant of Patent License. Subject to the terms and conditions of
+      this License, each Contributor hereby grants to You a perpetual,
+      worldwide, non-exclusive, no-charge, royalty-free, irrevocable
+      (except as stated in this section) patent license to make, have made,
+      use, offer to sell, sell, import, and otherwise transfer the Work,
+      where such license applies only to those patent claims licensable
+      by such Contributor that are necessarily infringed by their
+      Contribution(s) alone or by combination of their Contribution(s)
+      with the Work to which such Contribution(s) was submitted. If You
+      institute patent litigation against any entity (including a
+      cross-claim or counterclaim in a lawsuit) alleging that the Work
+      or a Contribution incorporated within the Work constitutes direct
+      or contributory patent infringement, then any patent licenses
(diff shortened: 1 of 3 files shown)


waddlesplash:
> However, most Haiku binaries do not contain a PT_GNU_EH_FRAME (or [PT_EH_FRAME](https://xref.landonf.org/source/xref/haiku/headers/os/kernel/elf.h#347), as defined in Haiku headers). To work around this, during initialization, Haiku binaries call [__register_frame_info](https://github.com/haiku/buildtools/blob/493aaad4dc61224ff101425c283a36960fb6841b/gcc/libgcc/crtstuff.c#L487), a function declared in libgcc_s.so.1, to tell libgcc_s about the [location of the .eh_frame section](https://github.com/haiku/buildtools/blob/493aaad4dc61224ff101425c283a36960fb6841b/gcc/libgcc/crtstuff.c#L265).

Why don't we solve this problem by just changing our compiler specs to always include PT_EH_FRAME, then? Isn't that the simplest solution?

trungnt2910:
> Why don't we solve this problem by just changing our compiler specs to always include PT_EH_FRAME, then? Isn't that the simplest solution?

If that's the case, the unwind libraries would only work for newly compiled binaries. If a newly-compiled binary somehow reference an old binary without a `PT_EH_FRAME` header, and then throw an exception with a stack that contains frame from the old binary, then the unwind library would still go "boom".

waddlesplash:
So long as the unwind library can identify what library lacks EH_FRAME information and include that appropriately in the crash message, I think that's acceptable; we'll rebuild all of HaikuPorts for beta4 anyway.

trungnt2910:
> We don't need this now since we have taken a different solution, yes?

Probably. I'm testing a build for llvm-libunwind on the new Haiku. I'll open a new pull request when that succeeds, and will close this one when libunwind is available on HaikuPorts.

trungnt2910:
Close as standalone libunwind from #7045 has been merged.

PR #7041, opened 2022-07-08
trackgit, update srcGitRev with the latest changes

Begasus:
32bit broken, issue filed upstream: https://github.com/HaikuArchives/TrackGit/issues/25


Begasus:
Add-on is installed in /boot/system/add-ons/x86/Tracker on 32bit, it doesn't show up in the sub-menu when right-click is used?

pulkomandy:
This is atracker add-on so it has to be built with gcc2 for the 32bit version, otherwise it won't work.

Begasus:
This should be ok for 64bit, 32bit broken and mentioned upstream.

PR #7044, opened 2022-07-09
airstrike: new recipe

TheZeldakatze:
I tried porting over airstrike from debian. The source I found has object files in it, that's the reason why there is a 'make clean'. It compiles on both x86-32 and 32 bit gcc11.

For some reason the game doesn't close properly, which has something to do with the sound, though if necessary it can be compiled without sound if the problem is not fixable.


TheZeldakatze:
> Maybe mark the architectures as `?` it launches here but not really playable.

It works on both x86 and and x86_64 for me. What's the part that's not workingß Physics should be patched

Begasus:
> It works on both x86 and and x86_64 for me.

Maybe it's just me, remember not getting it also from back in the BeOS days :)

PR #7048, opened 2022-07-11
sbase: new recipe

extrowerk:
While the recipe builds fine, it fails at the install stage which i can't really understand. Please give me some hints.

Thank You!

### DIFF
--- a/sys-apps/sbase/sbase-0.0~git.recipe
+++ b/sys-apps/sbase/sbase-0.0~git.recipe
@@ -1,0 +1,251 @@
+SUMMARY="Portable core utilities"
+DESCRIPTION="A  collection of  unix  tools that  are inherently  portable \
+across UNIX and UNIX-like systems.
+
+The following tools are implemented:
+
+  basename cal cat chgrp chmod chown chroot cksum cmp cols comm cp cron cut \
+  date dirname du echo ed env expand expr false find flock fold getconf grep \
+  head hostname install join kill link ln logger logname ls md5sum mkdir mkfifo \
+  mktemp mv nice nl nohup od paste pathchk printenv printf pwd readlink renice \
+  rm rmdir sed seq setsid sha1sum sha224sum sha256sum sha238sum sha512sum \
+  sha512-224sum sha512-256sum sleep sort split sponge strings sync tail tar \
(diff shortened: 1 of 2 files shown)


extrowerk:
Ah, it uses the freshly built mkdir and it doesn't want to create any directory if any parent folder is RO.

Begasus:
When using `make sbase-box $jobArgs` in `BUILD()` and `make sbase-box-install PREFIX=$prefix MANPREFIX=$manDir` in `INSTALL()` (and some changes to the recipe and the Makefile) I got the package to build.
Using sbase-box will create one binary and links the other "binaries" to sbase-box.

sys-apps/sbase/sbase-0.0~git.recipe
```diff
@@ -0,0 +1,251 @@
+SUMMARY="Portable core utilities"
+DESCRIPTION="A  collection of  unix  tools that  are inherently  portable \
```
diversys, line 2: Extra spaces

extrowerk:
> ~When using `make sbase-box $jobArgs` in `BUILD()` and `make sbase-box-install PREFIX=$prefix MANPREFIX=$manDir` in `INSTALL()` (and some changes to the recipe and the Makefile) I got the package to build. Using sbase-box will create one binary and links the other "binaries" to sbase-box.~
> 
> **EDIT** strike that, it worked a minute ago, same issue now :(

Check what package got actvated in the chroot. Probably the previously built sbase causes the problem.

Begasus:
> Check what package got actvated in the chroot. Probably the previously built sbase causes the problem.

Moving it out of the way did the trick. :)

korli:
> Should we really add a recipe for this if it could potentially cause confusion in pkgman or HaikuPorter when looking for `cmd` provides?

it's a bad idea, at least not a HaikuPorts, this looks like a support nightmare.

pulkomandy:
Not only a support nightmare, but wasn't the goal of Haiku to provide a unified system and avoid the modern GNU situation of "you have a dozen kernels, 10 different C libraries and 50 different implementations of the core utilities, please pick one of each to assemble your system"?

We should try to avoid that and pick one implementation for each tool. We can discuss which implementation is the best, but given this recipe can't even install itself, it seems not the best in the current state at least?

PR #7062, opened 2022-07-21
tiff, bump version

Begasus:


Begasus:
gcc2 fails with parse errors, so marked primary as broken, one of the recipes using tiff is sdl2_image, in this case it could still use the tiff version 4.2.0 for primary and 4.4.0 for secondary architecture.
Update on tiff was needed for wxgtk update (coming up): https://github.com/wxWidgets/wxWidgets/commit/45e43b4527ea1886aa51964f3cbd710503b6c43e

PR #7074, opened 2022-07-25
Icon for Dukto

Handmaus:
Hi, I made an icon for the Dukto program, hope you like it.

Dukto 02d


Begasus:
Looks nice, can you add the icon to the binary? (if not I can have a look at it later (or someone else)) :)

Handmaus:
You mean, uploading the modified program file I have, (to some folder)?
I guess not...
The way I have personalized icons applied to programs in my system is via copying the program file from "_system/apps_" to an _apps_ folder under _non-packaged_, and then invoking the FileTypes dialog, dropping the .iom in the **Icon** area, putting a link to that modified program file in a LaunBox pad and calling it a day :)

If you mean updating the recipe, I plea you to handhold me a bit through the process, as I'm not a programmer.
Looking the recipe of some other program that already has an icon, I have seen that it just has an additional line at he end of the first block that says:
ADDITIONAL_FILES="lagrange.rdef.in"
That will be it? Exporting from Icon-O-Matic in HVIF RDef format with ".rdef.in" extension and adding such a line in the recipe?

Begasus:
If you need some help updating the recipe I (or someone else) could guide you through it, maybe this could be done over IRC (doesn't clobber the mailing list to much then).
And yes, you should create an rdef.in file and add the needed information in the recipe, I guess there are plenty of recipes you could look into to create this.

PR #7087, opened 2022-08-05
olm: bump version

TheZeldakatze:
I hope I did everything correctly. I renamed the file, changed the version number in the file and added 
`-DCMAKE_BUILD_TYPE=RelWithDebInfo` to the cmake command, as that was required to build it.


korli:
The old version recipe should still be removed.

TheZeldakatze:
> The old version recipe should still be removed.

The old recipe should be removed now, I'm a bit confused why it is shown in the commit but not the changed files tab

PR #7109, opened 2022-08-17
geos: fix libVersion

sikmir:
Fixup #7107

### DIFF
--- a/sci-libs/geos/geos-3.11.0.recipe
+++ b/sci-libs/geos/geos-3.11.0.recipe
@@ -26,7 +26,7 @@ else
 	commandBinDir=$prefix/bin
 fi
 
-libVersion="1.16.0"
+libVersion="1.17.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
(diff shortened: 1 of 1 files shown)


Begasus:
some packages already exist at /var/buildmaster/haikuports/packages, revision bump required

https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=3029&viewMode=expanded

Maybe needed to get the correct libVersion for other packages that will use it?

sikmir:
> some packages already exist at /var/buildmaster/haikuports/packages, revision bump required

Sure

> Maybe needed to get the correct libVersion for other packages that will use it?

As far as I see only gdal and qgis use geos now. They don't have any requirements for geos version. Or what do you mean?

Begasus:
The fix for the libVersion hasn't been rebuild, so if a package pulls in geos it will still see `libVersion="1.16.0"` atm.

sikmir:
Well, am I right that all we need is to bump geos revision now?

PR #7110, opened 2022-08-17
fish: bump version

ivanholmes:
Thanks to the good work recently by waddlesplash on FIFOs and trungnt2910 on locale_t, the latest version of fish shell now compiles and runs on Haiku with minimal changes.

I tried porting extrowerk's patch to fix the crash on exit from fish 2.7.1, but it doesn't seem to work so I removed it.

I'm submitting this recipe anyway in case someone else can fix it, or it can be lived with for now.

### DIFF
--- a/app-shells/fish/fish-3.5.1.recipe
+++ b/app-shells/fish/fish-3.5.1.recipe
@@ -42,9 +42,9 @@ BUILD_REQUIRES="
 	devel:libpcre2_32$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
+	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
 	cmd:make
-	cmd:cmake
 	"
 
 defineDebugInfoPackage fish$secondaryArchSuffix \
@@ -57,12 +57,7 @@ BUILD()
 	mkdir -p build && cd build
 	cmake .. \
 		-DCMAKE_BUILD_TYPE=Release \
-		-DCMAKE_INSTALL_PREFIX=$prefix \
-		-DCMAKE_INSTALL_DATADIR=$dataDir \
-		-DCMAKE_INSTALL_SYSCONFDIR=$settingsDir \
-		-DCMAKE_INSTALL_DOCDIR=$docDir \
-		-DCMAKE_INSTALL_MANDIR=$manDir \
-		-DCMAKE_INSTALL_LOCALEDIR=$dataDir/locale
+		$cmakeDirArgs
 
 	make $jobArgs
 }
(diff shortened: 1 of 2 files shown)


app-shells/fish/fish-3.5.1.recipe
```diff
@@ -36,17 +34,17 @@ REQUIRES="
...
 	devel:libgettextlib$secondaryArchSuffix
 	devel:libiconv$secondaryArchSuffix
 	devel:libncursesw$secondaryArchSuffix
 	devel:libpcre2_32$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
-	cmd:awk
 	cmd:gcc$secondaryArchSuffix
 	cmd:make
+	cmd:cmake
```
augiedoggie, line 47: Entries should be sorted alphabetically

app-shells/fish/fish-3.5.1.recipe
```diff
@@ -56,12 +54,15 @@ defineDebugInfoPackage fish$secondaryArchSuffix \
...
-		--without-included-pcre2
+	mkdir -p build && cd build
+	cmake .. \
+		-DCMAKE_BUILD_TYPE=Release \
+		-DCMAKE_INSTALL_PREFIX=$prefix \
+		-DCMAKE_INSTALL_DATADIR=$dataDir \
+		-DCMAKE_INSTALL_SYSCONFDIR=$settingsDir \
+		-DCMAKE_INSTALL_DOCDIR=$docDir \
+		-DCMAKE_INSTALL_MANDIR=$manDir \
+		-DCMAKE_INSTALL_LOCALEDIR=$dataDir/locale
```
waddlesplash, line 65: Use $cmakeDirArgs instead of specifying these manually.

waddlesplash:
Can you start two Fish shells? IIRC when I was testing you could not, the second one hung.

Also this can't be merged yet as it won't build on beta3, which is what the buildmaster currently targets.

app-shells/fish/fish-3.5.1.recipe
ivanholmes: Thanks, done in latest commit

app-shells/fish/fish-3.5.1.recipe
ivanholmes: Ah yes I'm a bit rusty, thanks

ivanholmes:
> Can you start two Fish shells? IIRC when I was testing you could not, the second one hung.
> 
> Also this can't be merged yet as it won't build on beta3, which is what the buildmaster currently targets.

I hadn't tried this at the time of submitting the PR but I have now and you can indeed start two or more and use them both at the same time. I tested with pkgman and iperf.

Begasus:
@waddlesplash  Since buildmaster moved to beta4 could this be merged?

waddlesplash:
yes.

PR #7117, opened 2022-08-18
JOSM: new recipe

sikmir:
**[JOSM](https://josm.openstreetmap.de/)** is an extensible editor for [​OpenStreetMap](https://www.openstreetmap.org/) (OSM).

[Packaging status](https://repology.org/project/josm/versions)

JOSM warns about using Java 8:
> JSOM soon will no longer support this version of java, you should update to version 11 or better!

I've tried to use Java 11 and 12, but it doesn't work for me, JOSM stops with error while startup. Looks like some Haiku specific problem. At the same time JOSM with Java 8 works fine on Haiku.


sci-geosciences/josm/josm_bin-18531.recipe
```diff
@@ -0,0 +1,46 @@
...
+	cat << EOF > $josmDir/josm.sh
+#!/bin/sh -l
+APPDIR="\$(dirname -- "\$(readlink -f -- "\${0}")" )"
+CLASSPATH="\${APPDIR}/java/josm.jar"
+\${JDK8_HOME}/bin/java -cp "\${CLASSPATH}" -Djosm.restart=true org.openstreetmap.josm.gui.MainApplication "\$@"
+EOF
+	chmod 755 $josmDir/josm.sh
+
+	addattr -t icon -f $portDir/additional-files/josm.hvif \
+	  "BEOS:ICON" "$josmDir/josm.sh"
```
Begasus, line 44: No chance on converting the hvif icon into an rdef file and use that one to add it to the binary?
Not sure it's needed to provide the svg icon (not checked but I guess it's part of the source)?

sikmir: > Not sure it's needed to provide the svg icon (not checked but I guess it's part of the source)?

Yes, svg icon is part of the upstream source repo and I've used it to convert into hvif icon.

Begasus: OK, then you can drop it in here, when I can I also use the svg icon to export it to rdef, doing so also enables you to add some mimetype info.

PR #7125, opened 2022-08-22
Bump MuPDF to version 1.20.3 and update packages that depend on it

ivanholmes:
This PR brings the latest version of MuPDF along with PyMuPDF, DocumentViewer and PDFTranslator.

Notes:

- switched to using the .tar.gz again for mupdf because they have stopped using .xz in favour of .lz (probably a good thing) but lzip is not installed by default
- a new patch was necessary for pymupdf's setup.py because the relevant part was rewritten upstream


Begasus:
Latest version is 1.20.3? ;)

ivanholmes:
> Latest version is 1.20.3? ;)

According to
https://mupdf.com/releases/index.html
I'm pretty sure it is? Now you have me worrying...

Begasus:
> > Latest version is 1.20.3? ;)
> 
> According to https://mupdf.com/releases/index.html I'm pretty sure it is? Now you have me worrying...

Eeps, misread some of the commit, you changed an older version to 1.20.2 and added 1.20.3 in here also :)

Begasus:
Build failures:
64bit:
https://build.haiku-os.org/buildmaster/master/x86_64/?buildrunDir=&viewMode=expanded
32bit:
https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=&viewMode=expanded

ivanholmes:
It looks to me like the build failures are the result of trying to build the newest version of DocumentViewer against MuPDF 1.19.1. I had the 'opposite' errors when trying to compile the non-updated source of DocumentViewer against MuPDF 1.20.3. 

This makes me think the buildmaster is still using 1.19.1 and the problem will be fixed when 1.20.3 is installed, but I know nothing about the way it operates.

Begasus:
Maybe first try to update the mupdf recipe's, then tackle the other ones?

ivanholmes:
> Maybe first try to update the mupdf recipe's, then tackle the other ones?

I suppose that could work, it would be a bit of a hassle to split it up though. Is it possible to have the buildmaster retry the build of a specific package?

korli:
> What's the best strategy to sort this out? I could submit a PR raising the revision number

Updating the recipes will trigger a build.

When submitting everything in one PR, you should make sure that the build dependencies also specify the required version. Then the build master will notice using the newly built dependency is required, will wait and resolve the build packages correctly.

ivanholmes:
> > What's the best strategy to sort this out? I could submit a PR raising the revision number
> 
> Updating the recipes will trigger a build.

Okay good, I'll do that now.

> When submitting everything in one PR, you should make sure that the build dependencies also specify the required version. Then the build master will notice using the newly built dependency is required, will wait and resolve the build packages correctly.

Right, I hadn't thought ot that, it makes sense though. I learn more about how all of this works with every recipe and update. Thank you for the help :)

PR #7126, opened 2022-08-23
pymupdf: add $secondaryArchSuffix to swig to fix 32-bit build

ivanholmes:
This fixes one of the issues with #7125 and should allow pymupdf to build on 32-bit.

### DIFF
--- a/dev-python/pymupdf/pymupdf-1.20.2.recipe
+++ b/dev-python/pymupdf/pymupdf-1.20.2.recipe
@@ -31,7 +31,7 @@ BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	devel:libmupdf$secondaryArchSuffix >= 1.20
 	devel:libfreetype$secondaryArchSuffix
-	swig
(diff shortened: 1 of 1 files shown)


dev-python/pymupdf/pymupdf-1.20.2.recipe
```diff
@@ -31,7 +31,7 @@ BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	devel:libmupdf$secondaryArchSuffix >= 1.20
 	devel:libfreetype$secondaryArchSuffix
-	swig
+	swig$secondaryArchSuffix
```
Begasus, line 34: There is no $secondaryArchSuffix for swig :)

https://github.com/haikuports/haikuports/blob/4f180bdb94db9749ae800820b271e262a12754aa/dev-lang/swig/swig-4.0.2.recipe#L33

Begasus: Strike that, my bad :)

PR #7127, opened 2022-08-23
documentviewer: add version requirement for mupdf

ivanholmes:
This adds a version requirement for libmupdf to the DocumentViewer recipe, which should force it to rebuild as per @korli's advice on #7125.


haiku-apps/documentviewer/documentviewer-0.3~git.recipe
```diff
@@ -22,7 +22,7 @@ REQUIRES="
 	lib:libfreetype$secondaryArchSuffix
 	lib:libjbig2dec$secondaryArchSuffix
 	lib:libjpeg$secondaryArchSuffix
-	lib:libmupdf$secondaryArchSuffix
+	lib:libmupdf$secondaryArchSuffix >= 1.20
```
korli, line 25: This one should be automatically set by haikuporter

ivanholmes: Okay, done in latest commit, thanks

PR #7129, opened 2022-08-25
check, bump version, drop older recipes

Begasus:


Begasus:
From the looks of it not used in any other recipe, checked and ran tests fine.

PR #7138, opened 2022-08-27
libatomic_ops: bump to 7.6.14

ivmai:


Begasus:
Thanks! Packages relying on this one seem to be ecl and boehm_gc (maybe need a revbump after this)

PR #7140, opened 2022-08-27
globulation2: move config folder

### LINKED ISSUE #7137: Globulation2 settings stored in ~/.glob2
PR merged here: https://github.com/haikuports/haikuports/pull/6473

@TheZeldakatze maybe you could have a look there now this has been merged? :)

TheZeldakatze:
This moves the game folder to B_USER_SETTINGS_DIRECTORY.

This should fix #7137

### DIFF
--- a/games-strategy/globulation2/patches/globulation2-0.9.4.4.patchset
+++ b/games-strategy/globulation2/patches/globulation2-0.9.4.4.patchset
@@ -1,4 +1,4 @@
-From 6bcfcb4cca73089051c56666d3a80be956e919ed Mon Sep 17 00:00:00 2001
+From c067937f7812485e83380c951a9f17d5c2718442 Mon Sep 17 00:00:00 2001
 From: Victor Gamper <victor@wenzeslaus.de>
 Date: Tue, 14 Dec 2021 22:52:40 +0100
 Subject: globulation2: use 2to3 to update the scons files
@@ -267,10 +267,10 @@ index 76f19fc..55ac205 100644
  		function = scanNsisContent,
  		argument = None,
 -- 
-2.30.2
+2.37.3
 
 
-From d9b8f94f723cc17c11c7e21d2a0f86a71837925e Mon Sep 17 00:00:00 2001
+From b3d5eaac359b62699b4b0cc89fe6d3bb833b71a8 Mon Sep 17 00:00:00 2001
 From: Victor Gamper <victor@wenzeslaus.de>
 Date: Tue, 14 Dec 2021 22:54:22 +0100
 Subject: Change SConstruct to use Variables instead of the deprecated Options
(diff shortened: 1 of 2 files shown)


games-strategy/globulation2/patches/globulation2-0.9.4.4.patchset
```diff
@@ -514,5 +514,52 @@ index d0ca0ef..dd6e33b 100644
...
++#endif
++
+ namespace GAGCore
+ {
+ 	FileManager::FileManager(const char *gameName)
+ 	{
+ 		#ifndef WIN32
++		#ifndef __HAIKU__
+ 		const char *homeDir = getenv("HOME");
++		#else
```
Begasus, line 546: Won't this affect the windows build? Maby `#elif defined(__HAIKU__)` ?

TheZeldakatze: it shouldn't, the `#ifndef WIN32` part is in the original aswell, but I had some problems with the newlines created being added into the source file, so this line is in there aswell.
I could try and set up a windows build environment to check that though

TheZeldakatze: I redid the patch and replaced `#ifndef WIN32` with `#if !(defined(WIN32) || defined(__HAIKU__))`  . The if clause contains the code that adds the unix folders to the path, I added a Haiku part above it that adds the config directory to the path instead

Begasus: No time to check this, but it still looks like a bit overkill? Maybe someone from the haikuports team can chime in?

augiedoggie: It doesn't look right to me. I don't think you need to replace the `#ifndef WIN32`.  Adding the `|| defined(__HAIKU__)` part will exclude this from Linux and other operating systems.

Although it's not required, I probably would've used the C++ version of [`find_directory()`](https://www.haiku-os.org/legacy-docs/bebook/BDirectory.html#find_directory).  This simplifies things because you can pass `NULL` to get the boot volume instead of looking it up.

games-strategy/globulation2/patches/globulation2-0.9.4.4.patchset
```diff
@@ -514,5 +514,74 @@ index d0ca0ef..dd6e33b 100644
...
+ 	FileManager::FileManager(const char *gameName)
+ 	{
+-		#ifndef WIN32
++		#ifdef __HAIKU__
++		BPath path;
++		find_directory(B_USER_SETTINGS_DIRECTORY, &path);
++		path.Append(gameName);
++		addDir(path.Path());
++		#endif
++		#if !(defined(WIN32) || defined(__HAIKU__))
```
augiedoggie, line 581:
```suggestion
+		#elif !defined(WIN32)
```

TheZeldakatze:
the game crashes if the folder does not exist, I'll fix that

games-strategy/globulation2/patches/globulation2-0.9.4.4.patchset
TheZeldakatze: I changed it to this, thank you
(I hope I implemented that correctly)

TheZeldakatze:
> the game crashes if the folder does not exist, I'll fix that

now it should work

augiedoggie:
I noticed the game crashes if it's installed to the user package directory(/boot/home/config/packages).  There is another find_directory() in part of the existing patch that could be changed to fix this.  Ideally the code should be modified to use the DATADIR value that is passed to scons at configure time. This path($dataDir) will be consistent regardless of which package directory the hpkg is in.

PR #7147, opened 2022-09-02
intel-wifi-firmwares: Update to use firmwares from OpenBSD 7.2 release

sgulgas:
The main priority for this was because the 7.2 release added support for my AX210 WiFi card with the iwx driver. (tested and working on nightly from yesterday's build) 

Looks like a few iwx capable firmwares are new since 7.1. 
Can't test iwm firmwares because I don't have the hardware but did bump that to the 7.2 version for the sake of uniformity.


diversys:
I guess you also need to update the date in the recipe filename.

sgulgas:
What should I name it to? 

I kept the original file name because i matched up with the date on the iwm firmwares (which haven't changed) and the iwx firmwares are marked 2022-01-10, a day before the iwm ones which I thought might cause some weird conflict. 

Being new to haikuports I left it alone and it built the package just fine so i figured it was good enough.

waddlesplash:
The recipe should be named according to the latest firmware package it contains, so yes, rename it.

waddlesplash:
In addition to the new firmwares I will also have to import the new driver from OpenBSD into Haiku's trunk. I can look at that in the next few days, please open a ticket at the main Haiku issue tracker for that.

waddlesplash:
Ah ... actually it appears the relevant changes are already in Haiku because I used a version of the driver from OpenBSD's development tree. So indeed this change may be all that is needed.

sys-firmware/intel-wifi-firmwares/intel_wifi_firmwares-2022_01_10p0.recipe
```diff
@@ -5,18 +5,18 @@ contained within the base system."
 HOMEPAGE="https://www.intel.com/content/www/us/en/support/articles/000005511/wireless.html"
 COPYRIGHT="2006-2022 Intel Corporation"
 LICENSE="Intel WiFi Firmware"
-REVISION="2"
+REVISION="3"
```
waddlesplash, line 8: reset this to 1 and drop the `p0` from the filename

korli: The version goes from 2022_01_11 to 2022_01_10p0, isn't this a downgrade?

waddlesplash: Ah, yeah, whoops, I missed that.  
So indeed the old filename should stand after all, and the revision bump stay.

sgulgas:
Okay i reverted the filename change commit, so it should be the old filename version and revision bump.

PR #7157, opened 2022-09-04
sqlite, bump version, no soname changes

Begasus:


Begasus:
Couldn't figure out how to sed the correct version for `sqliteVersion`, the version should be `3390200` so I added it to the version as `${sqliteVersion}00 `for now.

korli:
the recipe should be named sqlite-3.39.2.0.recipe

Begasus:
Will look into it later today, thanks :)

PR #7228, opened 2022-09-25
libsoup, bump version

Begasus:


Begasus:
@threedeyes the build is OK without the patch, any reason to keep it around?

Begasus:
Keeping the patch for now.

PR #7229, opened 2022-09-25
libspiro, bump version

Begasus:


jmairboeck:
this has broken fontforge, and therefore lilypond.

Begasus:
Actualy fontforge fails to build on an error for locale_t (reverting the old patch builds it fine)

Begasus:
Should be fixed in https://github.com/haikuports/haikuports/pull/7348

PR #7237, opened 2022-10-01
python310: bumped version to 3.10.7

### LINKED ISSUE #7183: [Python] Enhancement: Enable line-editing support (libedit or readline) for the interactive interpreter
The REPL on the current Python packages (tested with 3.7.12 and 3.10.6) doesn't supports line editing features.

Attempting to use the cursor keys seems to confuse the interpreter with spurious A/B/C/D chars after pressing up/down/right/left respectively. Using backspace to edit the line results in garbled input being sent to the interpreter.

Assuming there was some miss-configuration, I've tried importing rtcompleter and readline to see if those worked. Was able to import the first, but the later fails.

Looking at the recipes, ~~iteems~~ it seems that Python builds are missing a `--with-readline=editline` or `--with-readline=readline` for that to actually work. See; https://github.com/python/cpython/blob/88a7f661ca02c0eb76b8f19234b8293b70f171e2/configure#L1879

The first uses the BSD libedit/editline library, the later depends on GNU's readline.

Thanks in advance!

Edit: fixed typo.

OscarL:
Also: build with "--with-readline=editline", which enables line-editting, history, and tab-completion on the interactive interpreter prompt.

Closes #7183.


Begasus:
While you are at it, could you move libedit before libexpat in REQUIRES and BUILD_REQUIRES? (seems I missed that one earlier) :)

Begasus:
Confirmed working fine on 32bit :+1:

PR #7247, opened 2022-10-02
iperf3: new recipe

davidkaroly:
There are a few kludges, I hope you'll still like it:
* name clash for `get_system_info()` resulting in stack overflow. I just renamed it to `do_get_system_info()`
* iperf3 uses IPv6 by default which doens't really work on Haiku so I changed default to IPv4
* the usual variable declaration issues on gcc2


Begasus:
Maybe add a develop package here to move the header there?

davidkaroly:
yep valid suggestion

PR #7249, opened 2022-10-03
hstr: bump to version 2.5

OscarL:
Built and tested on 64-bits only.


Begasus:
Build tested on 32bit fine too.

PR #7262, opened 2022-10-08
PortMidi: new recipe

jmairboeck:


jmairboeck:
tested on x86_gcc2, x86 (beta3) and x86_64 (nightly)

korli:
Looks good. Though you should probably select a java version, as we don't know which one the builder will select.

jmairboeck:
It should work with any version. Which one should I choose? (I only tested with openjdk 11)

media-libs/portmidi/portmidi-2.0.3~git.recipe
```diff
@@ -0,0 +1,155 @@
...
+}
+
+INSTALL()
+{
+	cd build
+	make install
+	cd ..
+
+	if [ "$effectiveTargetArchitecture" != x86_gcc2 ]; then
+		cp pm_java/pmdefaults/pmdefaults.jar $libDir
```
jmairboeck, line 100: Is this the correct location for a jar file?

jmairboeck:
Should the test programs also be packaged, maybe as a tools subpackage? OpenSUSE provides them in the portmidi package.

pulkomandy:
If you want to package them, put them in a _tools package or in the _devel package. This way, apps that just need the lib don't bring them in as dependencies.

jmairboeck:
Note: the implementation of the Haiku backend is incomplete and some of the (manually invoked) test suite fails. See https://github.com/PortMidi/portmidi/pull/38#issuecomment-1279744313 for details. The basic implementation (including Midi output in Frescobaldi https://github.com/haikuports/haikuports/pull/6924) works though.

media-libs/portmidi/portmidi-2.0.3~git.recipe
waddlesplash: No idea. Where are the other .jars that come with java placed?

jmairboeck: OpenJDK itself is intalled in $libDir/openjdk$majorVersion.

I looked at other recipes in the dev-java category. abcl uses $libDir, ant_core uses $libDir/ant, rxtx uses the lib/ext directory of openjdk8 itself (hard-coded version and only for x86 it seems).

So I think, according to these examples, $libDir is probably okay.

jmairboeck:
> Looks good. Though you should probably select a java version, as we don't know which one the builder will select.

It apparently compiled with OpenJDK 14, but the package now requires openjdk13_jre. This should be fixed, probably by requiring a fixed version. Which one should be used?

PR #7263, opened 2022-10-08
wxqt: update download url

davidkaroly:
> If you intend to build wxWidgets from sources (which is recommended), please do NOT download the files using the "Source code" links just above, which are automatically generated by GitHub and don't contain the submodules sources which are necessary for building wxWidgets.

see:
https://github.com/wxWidgets/wxWidgets/releases/tag/v3.1.3


Begasus:
Didn't check this, but have been issues building scummvm_tools with this, hence I switched to wxgtk3 iirc

davidkaroly:
our wxqt is somewhat outdated, we need to start using the updated download link if we want to move to a newer version.
3.1.x is development branch so eventually we should move to 3.2.x (like it's already done for wxgtk)

Begasus:
Will do a check build for the scummvm-tools gui when this can be installed, last time I checked with this wxwidgets it didn't work where as it did for the wxgtk3 package. Thanks!

davidkaroly:
one more thing, this wxqt recipe has SDL disabled, that might cause an issue with scummvm

Begasus:
Haven't checked recently, but I think the base for wxgtk was based on the wxqt recipe?

PR #7270, opened 2022-10-10
hstr: fix configuration instructions and other minor changes.

OscarL:
* Use the proper configuration instructions for bash in Haiku. Big thanks to madmax for finding the issue!
* Reworked the patch about the history files locations.
* Removed the TEST() section, as "make check" had "nothing to do".
* Removed "-I m4" from aclocal, as it was not necessary.

The issue was that, on platforms that lack the TIOCSTI ioctl, a workaround is needed to be able to insert text on the shell's command line. The workaround relies on defining a shell function, and keybinding said function so it gets handled by readline.

Now, following the "hstr --show-configuration" instructions will result in C-r working as expected, at least for bash.

It still won't under zsh (for the lack of a similar workaround, but implemented as a zsh function).

### DIFF
--- a/app-shells/hstr/hstr-2.5.recipe
+++ b/app-shells/hstr/hstr-2.5.recipe
@@ -1,10 +1,10 @@
 SUMMARY="Shell history suggest box"
 DESCRIPTION="Easily view, navigate, search and use your command history with \
-shell history suggest box for Bash and ZSH."
+shell history suggest box for Bash."
 HOMEPAGE="http://me.mindforger.com/projects/hh.html"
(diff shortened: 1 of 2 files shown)


humdingerb:
If it doesn't work for ZSH, it shouldn't be mentioned in the desceiptioniption

humdingerb:
> Edit 2: Ah... you mean removing the mention of ZSH from Haiku's package description, right? that makes sense :-D

+1 :)

humdingerb:
LGTM. Thanks!

PR #7273, opened 2022-10-11
haiku-format: update to latest commit

humdingerb:
Use the provided build.sh

### DIFF
--- a/sys-devel/haiku-format/additional-files/haiku-format_install_config.sh
+++ b/sys-devel/haiku-format/additional-files/haiku-format_install_config.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+settingsDir="$(finddir B_USER_SETTINGS_DIRECTORY)"
+
+configSysDataDir="$(finddir B_SYSTEM_DATA_DIRECTORY)/haiku-format"
+configUserDataDir="$(finddir B_USER_DATA_DIRECTORY)/haiku-format"
+
+if [ -e "$configSysDataDir/.haiku-format" ]; then
+	cp -a "$configSysDataDir/.haiku-format" "$settingsDir"
+elif [ -e "$configUserDataDir/.haiku-format" ]; then
+	cp -a "$configUserDataDir/.haiku-format" "$settingsDir"
+else
+	exit 1 # couldn't find .haiku-format
+fi
+
+chmod +w "$settingsDir/.haiku-format"
--- a/sys-devel/haiku-format/haiku_format-10.0.1.recipe
+++ b/sys-devel/haiku-format/haiku_format-10.0.1.recipe
@@ -2,19 +2,27 @@ SUMMARY="Code formatter for Haiku coding style"
 DESCRIPTION="Format C++ code following the Haiku Coding Guidelines.
 
 This is a fork of clang-format with customizations to implement the coding style followed by Haiku \
-and several applications running on Haiku."
+and several applications running on Haiku.
+
+haiku-format looks for a configuration file '.haiku-format' or '_haiku-format', going up the file \
+hierarchy starting with the folder of the file you want to format. It defaults to ~/config/settings/.haiku-format."
 HOMEPAGE="https://github.com/owenca/haiku-format"
 COPYRIGHT="2003-2019 University of Illinois at Urbana-Champaign
-	2019-2021 Owen Pan, Saloni Goyal, Adrien Destugues"
+	2019-2022 Owen Pan"
 LICENSE="Apache v2 with LLVM Exception"
-REVISION="1"
-srcGitRev="bcc5a44f23e6663ade208840fe5509c6da5f2459"
+REVISION="2"
+srcGitRev="859d67b9d2f6395f20d424b94aa63b75db932c7d"
 SOURCE_URI="https://github.com/owenca/llvm-project/archive/$srcGitRev.tar.gz"
 SOURCE_DIR="llvm-project-$srcGitRev"
-CHECKSUM_SHA256="b9b923099224a238187560fdf5bd2181d9df73a624040cc83c6db26f6aae3f77"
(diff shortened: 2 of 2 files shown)


humdingerb:
Instead of having build.sh download the llvm and clang sources via wget, this recipe downloads them, see SOURCE_URL_2/3. build.sh never gets to use wget. Is cmd:wget still needed as a build requirement?

pulkomandy:
It would be a lot simpler to make the recipe build directly from https://github.com/owenca/llvm-project/tree/haiku-format which is a version of clang and llvm already patched with haiku-format patches.

In any case, wget should not be needed.

humdingerb:
Oh, I thought it'd be easier for @owenca (or anyone) to maintain those patches and build script than provide the whole patched clang/llvm code. Should I revert that?

pulkomandy:
For me it's easier to work with the full repository at least in the context of haikuports packaging. I will let owenca say if he has any preference.

owenca:
> Instead of having build.sh download the llvm and clang sources via wget, this recipe downloads them, see SOURCE_URL_2/3. build.sh never gets to use wget.

I can simplify build.sh or add a command-line option if you want to use it in the recipe.

> Should I revert that?

Yeah, or edit the recipe after I add the [feature](https://github.com/owenca/haiku-format/issues/25#issuecomment-1275249111) you wanted and fix the tabs in _haiku-format.

owenca:
> I will let owenca say if he has any preference.

Thanks, though I don't know which will be easier until after updating haiku-format to the latest LLVM that Haiku supports. I will keep both repositories in sync meanwhile.

BTW the [copywrite](https://github.com/haikuports/haikuports/blob/99a07a0a2e2d53945000bdb889cf7cd31f0cb898/sys-devel/haiku-format/haiku_format-10.0.1.recipe#L14) is including `Saloni Goyal, Adrien Destugues`, but none of their work has been merged. I asked @saloniig to create pull requests but haven't heard back.

humdingerb:
> I will keep both repositories in sync meanwhile.

Alright. Then I'll revert to how it was done before and we'll if something needs to change when you made an update.

> BTW the [copywrite](https://github.com/haikuports/haikuports/blob/99a07a0a2e2d53945000bdb889cf7cd31f0cb898/sys-devel/haiku-format/haiku_format-10.0.1.recipe#L14) is including Saloni Goyal, Adrien Destugues, but none of their work has been merged. I asked @saloniig to create pull requests but haven't heard back.

OK, we'll remove them then for now.
I'll wait with the recipe update for your ~/config/setting/.haiku-format update. No hurry.

humdingerb:
@owenca, is there a chance to put the _haiku-format config file into the [llvm]https://github.com/owenca/llvm-projectl) repo? Then we'd just have to pull in one repo in the recipe.

owenca:
@humdingerb The _haiku-format [file](https://github.com/owenca/llvm-project/blob/haiku-format/_haiku-format) has always been there and is the same as the one in https://github.com/owenca/haiku-format. :) Did you switch branch to `haiku-format` in the https://github.com/owenca/llvm-project repo?

Anyway, I need to fix the tabs in _haiku-format as noted [above](https://github.com/haikuports/haikuports/pull/7273#issuecomment-1275608822) before you can use it.

humdingerb:
> Did you switch branch to haiku-format in the https://github.com/owenca/llvm-project repo?
Of course not... I was wondering why I couldn't find the commit hash of the old recipe in master. :)
Give us a shout when you've finished your changes.

owenca:
@humdingerb I'm done with updating https://github.com/owenca/llvm-project/tree/haiku-format. Now that the default config file goes into ~/config/settings, I recommend that you use _haiku-format, not .haiku-format, as the file name.

I have noticed that the package name of haiku-format is haiku_format. Not sure it was on purpose or an oversight.

humdingerb:
> Now that the default config file goes into ~/config/settings, I recommend that you use _haiku-format, not .haiku-format, as the file name.

Does anything speak against naming it "haiku-format.settings" or "haiku-format.config". That would seem to fit better with toher setting files.

Also, wouldn't it be better to use find_directory(B_USER_SETTINGS_DIRECTORY) instead of hard-coding the settings folder? Granted, it seems unlikely to change, but who knows...

> I have noticed that the package name of haiku-format is haiku_format. Not sure it was on purpose or an oversight.

Package names aren't allowed dashes, they are replaced with underscores. I will, however, see to it's visible name in HaikuDepot showing "haiku-format".

owenca:
> Does anything speak against naming it "haiku-format.settings" or "haiku-format.config". That would seem to fit better with toher setting files.
>
> Also, wouldn't it be better to use find_directory(B_USER_SETTINGS_DIRECTORY) instead of hard-coding the settings folder? Granted, it seems unlikely to change, but who knows...

I had actually thought of both but settled on doing neither. I didn't use `B_USER_SETTINGS_DIRECTORY` because I usually develop haiku-format on a non-Haiku OS. As to the naming of the config file, I don't want to change the current clang-format behavior of looking for .clang-format (and if not found, then _clang-format) up the directory hierarchy. (The only reason that I renamed the executable and config files to haiku-format and _haiku-format etc. is to allow clang-format and haiku-format to coexist on Haiku.) It just feels weird to me to have config files named e.g. haiku-format.config in various source directories if one needs different formatting configurations than the settings in the default config file in ~/config/settings.

On second thought, you can keep the name as `.haiku-format` as there is at least one precedence: `.lesshst`. Alternatively, I can go with either `haiku_format` (similar to `bash_history`, `time_dststatus`, etc) or `haiku-format` (similar to `wget-hsts`).

Anyway, I want to minimize the diff between clang-format and haiku-format to ease maintainability.

humdingerb:
Alright. Let's keep it named .haiku-format to make it distinguishablefrom the executable.

Anyone, feel free to merge, if it looks OK.

Issue #7259, opened 2022-10-07
blender fatal error at launch


haroldsoixante:
blender just doesnt starts at all

showing this message

~> /boot/system/apps/Blender/Blender
runtime_loader: Cannot open file libOpenImageIO.so.2.2 (needed by /boot/system/apps/Blender/Blender): No such file or directory
~> /boot/system/apps/Blender/Blender.sh 
runtime_loader: Cannot open file libOpenImageIO.so.2.2 (needed by /boot/system/apps/Blender/Blender): No such file or directory
~> 

![screenshot5](https://user-images.githubusercontent.com/102438052/194671105-f71eb5e1-c11f-460b-be5e-a9171c936155.png)

kenmays:
1. pkgman full-sync
2. pkgman install blender3

kenmays:
To fix the libosdGPU.so.3.4.3 issue:
Reinstall opensubdiv 3.4.3-1.

kenmays:
To fix the libosdGPU.so.3.4.3 issue:
- Reinstall opensubdiv 3.4.3-1.

To fix the libopenimageio.so.2.2 issue:
- Reinstall openimageio - 2.2.16.0-1

scottmc:
@haroldsoixante does this work for you now after the 3.3.1 rev bump and updating the files listed above?

kenmays:
@scottmc - Need to merge the new PR. The buildbots were updated this weekend to fix python update issues so the PR will build on it now.

haroldsoixante:
> @haroldsoixante does this work for you now after the 3.3.1 rev bump and updating the files listed above?

please indicate me by quote which specific commands I should enter, Im not used yet of haiku commands :o :)

PR #7277, opened 2022-10-13
 blender: bump to 3.3.1

kenmays:
Tested on hrev56524 x64.
Ref: https://github.com/haikuports/haikuports/issues/7259


extrowerk:
Patches could be upstreamed.

PR #7280, opened 2022-10-13
lighttpd: rebuild

kenmays:
- added modern authentication and data security methods
- added compression for HTTP content 
- unit tested on hrev56529

### DIFF
--- a/www-servers/lighttpd/lighttpd-1.4.53.recipe
+++ b/www-servers/lighttpd/lighttpd-1.4.53.recipe
@@ -9,7 +9,7 @@ solution for every server that is suffering load problems."
 HOMEPAGE="https://www.lighttpd.net/"
 COPYRIGHT="2003-2019 lighttpd authors"
 LICENSE="BSD (3-clause)"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-$portVersion.tar.xz"
 CHECKSUM_SHA256="3bdfce1cf3e9650a556a8c26fb15342c5717c63f530c54693db632b0371dcb78"
 ADDITIONAL_FILES="
@@ -20,7 +20,7 @@ ADDITIONAL_FILES="
 	quickstart-geoip.txt
 	"
 
-ARCHITECTURES="all !x86_gcc2 ?arm ?ppc"
+ARCHITECTURES="all !x86_gcc2"
 if [ "$targetArchitecture" = x86_gcc2 ]; then
 SECONDARY_ARCHITECTURES="x86"
 fi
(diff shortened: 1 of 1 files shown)


www-servers/lighttpd/lighttpd-1.4.53.recipe
```diff
@@ -20,7 +20,7 @@ ADDITIONAL_FILES="
 	quickstart-geoip.txt
 	"
 
-ARCHITECTURES="all !x86_gcc2 ?arm ?ppc"
```
nephele-gh, line 23: did you test this on power?

kenmays: An unsupported platform? No.

nephele-gh: "all !x86_gcc2" means *every* arch defined except for x86_gcc2, I'm asking for the reason you marked arm and powerpc as tested if you haven't tested them at all.

Wether repos exist is irrelevant to wether recipes could be build for those arches.

kenmays: - Lighttpd is known to work/verified on BSD/Linux-ppc/ppc64/arm platforms. 

Why document on unsupported variants or a port known to work on certain processors or supported cross platform OSes similar to Haiku? We are discussing POWER in which there is no arch support yet for Haiku. Even for ppc, there are no image releases at all (just the CI kernel builds).  As for arm, lighttpd packages exist for BSD/Linux so why create more effort documenting a "maybe/maybe not" condition with "supported" ways to build/verify on Haiku-arm?

augiedoggie: The majority of the recipes at HaikuPorts make no mention of the other arches.  Obviously they haven't been tested either but I don't think we should go through and add the untested flag to all of them.  I seem to recall having a conversation with someone(maybe @waddlesplash) who said that we assume it works until we hear otherwise.

waddlesplash: Yes. Unless the recipe has some kind of per-architecture status (e.g. GCC) then `all` should always be used, with specific exceptions (likely `!x86_gcc2`) listed, not each individual architecture listed.

waddlesplash: So that is to say, the change as it currently exists is correct.

Begasus:
Build failure on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=3204&viewMode=expanded

kenmays:
Check for libattr_x86 build.

augiedoggie:
> Check for libattr_x86 build.

libattr should not be used for this purpose.  The attr functions have been intentionally disabled as they are already provided by libgnu.  The only reason i added the recipe is to have the command line tools available.

augiedoggie:
It looks like they have improved their support for xattr functions in the newer versions.  It should be possible to bump the version and add `-lgnu` to the LIBS variable in the recipe.

PR #6728, opened 2022-03-15
X265: bump

extrowerk:
Lobotomized old version for VLC
enabled 5, 10 and 12 bit depth
debuginfo

PR #7292, opened 2022-10-13
flac: bump to 1.4.2

enzo1982:
Updates patchset to maintain GCC 2.95 compatibility (variables not declared at the top of blocks).

**Attention: SONAME change!**

There are minor changes to the libFLAC 1.4 API and ABI, so dependencies (~35 recipes) will have to be rebuilt. I assume bumping the revision number of those recipes is still necessary for this. Should that be a separate pull request or added as a commit here?


diversys:
I think either way is fine.

enzo1982:
OK, this will take a while. Some recipes need version bumps, others are broken due to build system changes or changes in other packages.

I'm at 13 out of ~35 now. Will try to work through the rest over the weekend.

korli:
If it's not compatible, please add a recipe for the old version, so that everything keeps working.

korli (changes requested):
please add a recipe for the old version, so that everything keeps working.

enzo1982:
Well, if keeping the old version around is a way to get this merged without having to adjust all the recipes depending on FLAC, I agree. I'll update the PR.

I'm still going to work through the recipes and try to make them compile again. I'll post that work as a separate PR.

The new FLAC release is generally compatible, but needs a recompile due to an ABI change (size of struct members changed). There is one renamed function, too, but it's rarely used and so far I have not come across anything using that function in the Haikuports repository.

The breakages I encountered so far are not due to FLAC. Problems encountered include:

- Source URL no longer available
- Configure script rejects GCC 11
- Cmake wants CMAKE_BUILD_TYPE defined
- Meson not available under the expected path
- Autotools failing to rebuild from ancient configure.ac

enzo1982:
Added a recipe for the old FLAC libraries as flac13, so both versions can be installed side-by-side.

Is this the correct way to handle this?

enzo1982:
Hm, when both, flac13 and flac, packages have been built, and I build a package that depends on libflac, Haikuporter builds against the older libs. Why doesn't it choose the newer version? How to avoid this?

korli:
> Hm, when both, flac13 and flac, packages have been built, and I build a package that depends on libflac, Haikuporter builds against the older libs. Why doesn't it choose the newer version? How to avoid this?

You can remove the devel package for the old recipe, it isn't needed.

Begasus:
@enzo1982 if you are having problems with the latest comment by @korli to remove the devel package from the older recipe, have a look here: https://github.com/haikuports/haikuports/pull/6728/files#diff-d017bd334afa89cedbc217895c3a49e4cab37e4bb1e1351cb4dab5a4ca69085c

enzo1982:
@Begasus, thanks, that link is helpful! I did not have time to continue with this immediately and then got tangled up in other things. But I'm at it again now.

Removing the devel package from flac13 breaks libsndfile_devel, so I guess that should be bumped immediately after this. That's no problem, I already prepared the recipes.

libsndfile should be the only recipe with a devel package depending on devel:libflac.

I'll update this soon and will also bump the FLAC recipe to 1.4.2 while I'm at it.

korli (approved):

Begasus:
Buildmaster is failing to build packages atm: https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=3254&viewMode=expanded
Related/reported issue: https://github.com/haikuports/haikuporter/issues/233

korli:
Yeah many issues, could be release related so...

PR #7297, opened 2022-10-17
wxqt: remove unnecessary patches

davidkaroly:


davidkaroly:
The patches for sckaddr and lnetwork are not needed any more as of wxqt-3.1.3
We still need to disable SDL because of a linking issue (sound_sdl.cpp references sound.cpp which is not built)
We also still need the patch for QPainterPath, that can be removed once we upgrade to 3.1.4 or newer

@Begasus I think this is not enough to help with ScummVM but should make it easier to move to newer wxqt version soon.

Begasus:
Haven't checked a build, as scummvm is in the process of releasing it's new version I'll wait until I have time to check this up on scummvm_tools, thanks so far!

PR #7309, opened 2022-10-19
mesa: fixed, updated recipe

kenmays:
Tested build on hrev56536 x64.


waddlesplash:
dependencies added to Mako instead.

PR #7313, opened 2022-10-21
python3.10: bump version to 3.10.8.

OscarL:
* Use "make altinstall".
* Documented problematic tests, and skipped the ones that hang/stall.
* Enable optimizations.
* Added a note regarding tests that invoke the crash dialog.

Running "hp --test python310" now gives the same results[*] for both non-optimized, and optimized builds, so let's just enable optimizations for 3.10.8.

[*] As there's some randomness involved in the test runs, YMMV.

Anecdotal/additional info:

Building and testing with "--disable-test-modules" might be considered? Python's docs read:

   "The test package is meant for internal use by Python only."

And that those are used for regression testing mostly. See: https://docs.python.org/3/library/test.html#module-test

Personal experience: I got more consistent results with that flag enabled, than without it. I left it out, trying to avoid hiding problems (a.k.a.: failing silently. A big NO-NO in QA/QC), and because I assume that packaging Python counts as "internal use".

Typos on the commit message might or might not be addressed on follow up commits :-P


dev-lang/python/python3.10-3.10.8.recipe
```diff
@@ -11,28 +11,28 @@ LICENSE="Python"
...
 commandBinDir=$binDir
+commandSuffix=$secondaryArchSuffix
 if [ "$targetArchitecture" = x86_gcc2 ]; then
-	commandSuffix=
 	commandBinDir=$prefix/bin
+	commandSuffix=
 fi
 
 PROVIDES="
-	python310$secondaryArchSuffix = $portVersion compat >= 3.10
```
Begasus, line 35: I think you need to keep this one (see the python3.9 recipe), this is picked up for recipes depending on `PYTHON_PACKAGES=(python3 python38 python39 python310)`.

Begasus:
GLOBAL_WRITABLE_FILES should be moved up before PROVIDES (see: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering)
Probably a remain from early recipes.

dev-lang/python/python3.10-3.10.8.recipe
Begasus: > I was assuming that was what the REPLACES was for.

I think in this case there isn't anything to replace as there wouldn't be anything providing python310?

> Is there any documentation/guidelines about how to handle situations like this?

No documentation, but as in the case with `-` vs `_` in recipes, here you can't use dot's in `PYTHON_PACKAGES` :)

pulkomandy:
>  PROVIDES="
> -	python310$secondaryArchSuffix = $portVersion compat >= 3.10
> 
> > I think in this case there isn't anything to replace as there wouldn't be anything providing python310?
> 
> Ok. I was assuming that on a situation like this: "package_X" requires "python310", but that got renamed to "python3.10", the package manager will see that no one provides "python310", BUT that the new "python3.10 guy" do replaces that.
> 
> If that's not the case, keeping python310 on PROVIDES makes sense, but then.... what would be the purpose of adding a REPLACES section in cases like this one (as @korli suggested on that "rename Python 3.9 and make it the default one" change)... just as a reference / documentation thing?

I don't know if what you suggest works, sounds logical but I don't know.

What I know works, is, when doing a pkgman full-sync, if a package
replaces another, pkgman will be hinted to uninstall the old one and
install the new one instead.

If you want more details, these things are actually implemented by
libsolv, the same dependency solver used by OpenSUSE, and I think the
documentation will be there if there is any.

PR #7314, opened 2022-10-22
wxqt: bump to version 3.1.7

davidkaroly:


davidkaroly:
3.1.7 is latest from 3.1.x series

we still need to disable dialupman, sound, SDL because of known upstream issues
we need to packport a patch from 3.2.0 to fix the build for uilocale.cpp

all samples and demos build successfully, except for demos/life.cpp which has an upstream issue that is reproducible also on Fedora (I'd like to open a ticket on that soon)

PR #7318, opened 2022-10-23
blender: updated for x64 buildbot

kenmays:
closes: https://github.com/haikuports/haikuports/issues/7259


media-gfx/blender/blender3-3.3.1.recipe
```diff
@@ -23,7 +23,7 @@ ADDITIONAL_FILES="
 	"
 
 ARCHITECTURES="all !x86_gcc2"
-SECONDARY_ARCHITECTURES="x86"
+SECONDARY_ARCHITECTURES="!x86"
```
waddlesplash, line 26: Why?

kenmays: We don't support Blender3 on x86.

kenmays: There are no Blender3 x86 builds.

PR #7326, opened 2022-10-26
dblatex: change REQUIRES to individual tex packages

jmairboeck:


Begasus:
Doesn't this conflict when texlive ... packages would be enabled?

jmairboeck:
this should be merged before enabling texlive (otherwise this would also need a revbump). This depends on #7325, which is already merged.

PR #7327, opened 2022-10-26
wireshark: bump to version 4.0.0

davidkaroly:


diversys:
Sure, if it works with qt6 why not. We're slowly migrating other recipes to it anyway.

davidkaroly:
as I usually say: first results are promising.

some more testing, then probably I'll raise one more PR.

image

davidkaroly:
well... maybe I was a bit too eager
Qt6.4 is not supported in wireshark-4.0.0, see https://www.wireshark.org/lists/wireshark-dev/202210/msg00009.html

so while it compiles and starts successfully, I'll hold back with the move to Qt6 for now.

PR #7328, opened 2022-10-26
wxqt: fix wxBITMAP() definition

davidkaroly:


davidkaroly:
We need this patch from upstream to fix wxBITMAP() macro on wxqt. This way all demos build successfully, including life.cpp.

PR #7331, opened 2022-10-27
fennel: add new recipe

dokutan:
I've added a new recipe for fennel (https://fennel-lang.org/), a lisp that runs on Lua.

### DIFF
--- a/dev-lang/fennel/fennel-1.2.1.recipe
+++ b/dev-lang/fennel/fennel-1.2.1.recipe
@@ -1,0 +1,38 @@
+SUMMARY="Lisp-like language that compiles to Lua"
+DESCRIPTION="
+Fennel is a programming language that brings together the speed, simplicity, \
+and reach of Lua with the flexibility of a lisp syntax and macro system."
+HOMEPAGE="https://fennel-lang.org"
+COPYRIGHT="2016-2022 Calvin Rose and contributors"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://fennel-lang.org/downloads/fennel-$portVersion.tar.gz"
+CHECKSUM_SHA256="27aa830ec70bb56a559fa367f385c2a2f4e4c1f03c8d165770e4921b7ede1d46"
+
+ARCHITECTURES="all"
+
+PROVIDES="
+	fennel = $portVersion
+	cmd:fennel = $portVersion
+	"
+REQUIRES="
+	haiku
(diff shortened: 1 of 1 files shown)


dev-lang/fennel/fennel-1.2.1.recipe
```diff
@@ -0,0 +1,38 @@
+SUMMARY="Lisp-like language that compiles to Lua"
+DESCRIPTION="
+Fennel is a programming language that brings together the speed, simplicity, \
```
Begasus, line 3: Keep DESCRIPTION start on one line:
```diff
-DESCRIPTION="
-Fennel is a programming language that brings together the speed, simplicity, \
+DESCRIPTION="Fennel is a programming language that brings together the speed, simplicity, \
```

dev-lang/fennel/fennel-1.2.1.recipe
```diff
@@ -0,0 +1,38 @@
...
+
+ARCHITECTURES="all"
+
+PROVIDES="
+	fennel = $portVersion
+	cmd:fennel = $portVersion
+	"
+REQUIRES="
+	haiku
+	lua
```
Begasus, line 20: No specific lua version required?

dokutan: The fennel website says that Lua 5.1, 5.2, 5.3, 5.4, or LuaJIT are supported. Since haikuports contains no other version of Lua, i didn't specify a version.

I also tried ``lua >= 5.1`` and ``lua >= 5.1.0``, but haikuporter failed to build the package in both cases. What would the correct way to specify the version be, if it is needed?

Begasus: Checked a build here, seems to work fine (although I can't tell how it's being used) :)
lua defaults to lua5.3, if you want to add support for other versions you can use lua5.1, lua5.2 or lua5.4

Begasus: I'm OK with it, just leaves the issue for DESCRIPTION (minor change) :)

dev-lang/fennel/fennel-1.2.1.recipe
dokutan: Fixed in 1753cdda8ef547455177509651a671bc28a3e5bd.

Begasus:
Thanks! :+1:

PR #7332, opened 2022-10-28
ne: update to version 3.3.2

OscarL:
Besides fixes/improvements due to the version update, we now have:

* Working syntax-highlighting.
* Settings stored under ~/config/settings/ne.
* Optional fix to the "backspace not working" issue.
* Dropped dependency on aspell (as is only required for an example macro file).

We now use the "low-level" makefiles to buid "ne". Otherwise the value assigned to NE_GLOBAL_DIR  was not being respected (causing "ne" to not find its syntax files, for example).

See "/boot/system/data/ne/fix_backspace.keys" for the backspace workaround.


augiedoggie (changes requested):

augiedoggie:
I haven't looked at the code but I suspect it still needs the termcap/terminfo database at runtime and should have a `ncurses6$secondaryArchSuffix` dependency in the REQUIRES.

augiedoggie:
Doesn't seem to use the database at runtime so I'll just add an entry to the BUILD_REQUIRES and merge it.

Thanks for the extra work on this one.

app-editors/ne/ne-3.3.2.recipe
```diff
@@ -0,0 +1,95 @@
...
+	ne$secondaryArchSuffix = $portVersion
+	cmd:ne$commandSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
```
augiedoggie, line 35:
```suggestion
	ncurses6 # for the terminfo/termcap database
	"
```

augiedoggie:
> I don't think it needs the terminfo (and thus ncurses6) at all in this case. At least... it builds just fine for me after removing any mention of ncurses from the recipe.

I think it might extract some keys at build time but I didn't take a close enough look.  Ncurses is a weird package because it's almost impossible to build a recipe and not have that package active.  Even if the recipe doesn't mention ncurses there are so many other things that use it that it will end up being pulled in.

PR #7338, opened 2022-10-29
wxqt: bump to 3.2.1

davidkaroly:


davidkaroly:
We can backport the patch for sound from upstream. This allows us to enable SDL.
All samples and demos build successfully.

PR #7346, opened 2022-10-31
ne: change default action for backspace key, and build against ncurses to fix redrawing issues.

### LINKED ISSUE #7333: ne: post-install the backspace workaround?
I almost gave up on "ne", because backspace and del didn't seem to work correctly.
Then I saw the latest commit message:

> See "/boot/system/data/ne/fix_backspace.keys" for the backspace
workaround.

@OscarL (or anyone): Is there any situation where anyone wouldn't want the "backspace workaround"?

If not, IMO that workaround should be hardwired (patched) or via post_install put into the ne settings.

At the very least, the above quoted hint needs to be in the recipe description or nobody will know...

OscarL:
~~A simple "is the workaround already applied" check is performed before making any change. Not ultra-sophisticated, but seems to work.~~

Should close #7333.

~~Tested with system and home package installations, without a `.keys` file in `~/config/settings/ne`, with one already containing the fix, and with an existing `.keys` with user content, but not the workaround.~~

~~Worked OK for me on these 6 cases. Albeit I've only tested these on 64 bits, I would expect it to work ok on 32 bits too.~~

~~This is my first time using post-install scripts, so... any suggestions/criticisms will be more than welcomed.~~

Thanks! 

Edit: workaround now applied directly on the executable, and not as a post-install keybinding setting.

### DIFF
--- a/app-editors/ne/ne-3.3.2.recipe
+++ b/app-editors/ne/ne-3.3.2.recipe
@@ -2,12 +2,14 @@ SUMMARY="The 'nice editor'"
 DESCRIPTION="ne is one of the few text editors which are \
 both easy to use for beginners and powerful enough for powerusers. \
 It supports UTF-8 and multi-column characters as well as syntax highlighting. \
-Is also has fully configurable key bindings and supports macros."
+Has a drop-down top menu (accesible with ESC, or F1), default keybindings like \
+CTRL+Q, CTRL+S, CTRL+O, etc, that resemble GUI editors. Keybindings are fully \
+configurable, and it also supports recording/replaying macros."
 HOMEPAGE="https://ne.di.unimi.it/"
 COPYRIGHT="1993-1998 Sebastiano Vigna
 	1999-2022 Todd M. Lewis and Sebastiano Vigna"
 LICENSE="GNU GPL v2"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/vigna/ne/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="9b8b757db22bd8cb783cf063f514143a8c325e5c321af31901e0f76e77455417"
 PATCHES="ne-$portVersion.patchset"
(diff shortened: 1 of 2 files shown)


Begasus:
First run it didn't seem to copy the .keys file, second run (install) it did and works OK on 32bit.

humdingerb:
It works here on 64bit. I tried:
* cleanly -> .keys is correctly copied
*  having an existing .keys file with random contents -> the BACKSPACE/DEL entries are added at the end
* having an existing .keys file with those same BACKSPACE/DEL entries -> nothing gets copied or added
All good.

Nevertheless, for the post-install script, I suggest to test for both "KEY" entries to be there, and if one or both are missing, adding them explicitly. If there is no .keys file it gets created with the "echo" command.
The .keys file in the $dataDir is just a backup, if the users messes things up... (though users could also just un/reinstall "ne")
```
if ! grep -q 0x7f ".keys" ; then
	echo -e "KEY\t0x7f\tBackSpace" >> ".keys"
fi
if ! grep -q 0x115 ".keys" ; then
	echo -e "KEY\t0x115\tDeleteChar" >> ".keys"
fi
```

humdingerb:
While I have you on, @OscarL...
As you're using ne on other plattforms as well, does copy&pasting of multiplelines?
At least on Haiku, only the first line is pasted, then the cursor jumps into the input line at the bottom of the screen.

humdingerb:
The pasting thing may be a Haiku problem. I filed a ticket:
https://dev.haiku-os.org/ticket/18029

korli:
The workaround should be set by default in the executable, not as part of the packaging.

app-editors/ne/ne-3.3.2.recipe
```diff
@@ -2,12 +2,14 @@ SUMMARY="The 'nice editor'"
 DESCRIPTION="ne is one of the few text editors which are \
 both easy to use for beginners and powerful enough for powerusers. \
```
humdingerb, line 3: A good place to have a line-break:

```suggestion
both easy to use for beginners and powerful enough for powerusers.
```

humdingerb (commented):
Nice! Small suggestions above.

app-editors/ne/ne-3.3.2.recipe
```diff
@@ -28,11 +47,12 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	ncurses6 # for the terminfo/termcap database
+	ncurses6${secondaryArchSuffix}_devel
```
korli, line 55: should be devel:libncurses$secondaryArchSuffix

korli:
Looks good!

humdingerb:
Very good. Now even the backspace/del issue in .py files is solved (that was the syntax higlighting thing, rigth?)
Well done! Let's merge!

humdingerb:
:+1:

PR #7347, opened 2022-10-31
audacity-2.3.3: new recipe, almost works

davidkaroly:


davidkaroly:
the good part:
- build, packaging, install successful
- it starts up, I'm able to load, select, cut&paste, save

the bad part:
- an audio editor with no play or record - as portaudio does not have a Haiku backend :(
- flickering
- strftime assertions
- had to disable a bunch of things like midi, portmixer, nyquist, ffmpeg

image

diversys:
Very cool! Have you tried to build it with wxqt?

davidkaroly:
> Very cool! Have you tried to build it with wxqt?

I think I tried and it failed quite early. But probably I'll take one more look.

I don't have such high hopes though, see this thread: audacity/audacity#877


Is flickering a known issue with wxgtk over xlibe? Maybe wayland could be better?

diversys:
Flickering is a known issue with xlibe and gtk, although there is no issue about that in https://github.com/waddlesplash/xlibe.
AFAIK there is no flickering in wayland.

PR #7352, opened 2022-11-01
lighttpd: fixed for x86

kenmays:


www-servers/lighttpd/lighttpd-1.4.53.recipe
```diff
@@ -231,10 +229,10 @@ defineDebugInfoPackage lighttpd$secondaryArchSuffix "${debugList[@]}"
 
 BUILD()
 {
-	LIBS="-lintl" \
+	LIBS="-lintl -lgnu" \
```
augiedoggie, line 232:
```suggestion
	LIBS="-lintl" \
```

After lighttpd is updated to a slightly newer version then `-lgnu` can be added along with `--with-attr` to the configure args

PR #7353, opened 2022-11-01
drumcircle: new recipe

davidkaroly:


davidkaroly:
I'd like some help with this one: DrumCircle expects profiles to use mime type `application/x-vnd.SY.DrumCircle-Profile`
This seems to be included in the hpkg file correctly but when I activate the package, the file system somehow forgets these extra attributes.

humdingerb:
Works here. All profile files in /boot/system/settings/drumcircle/ have application/x-vnd.SY.DrumCircle-Profile as BEOS:TYPE and the app seems to work.

davidkaroly:
Probably it was a build environment issue on my side. After rebuilding the package from 1.0.4 tag, I get the correct MIME types.

media-sound/drumcircle/drumcircle-1.0.4.recipe
```diff
@@ -0,0 +1,41 @@
+SUMMARY="Simple sequencer with Hippie AI"
+DESCRIPTION="A simple MIDI sequencer with a knack for simulating Hippie Drum Circles. \
+Tweek the Tweekables and have fun!"
+HOMEPAGE="https://github.com/steffenyount"
```
augiedoggie, line 4: https://github.com/steffenyount/DrumCircle  would be better

augiedoggie:
I just noticed that this recipe was put in `media-sound` instead of `haiku-apps`.  I kind of like having the native apps all in one place under `haiku-apps` but I don't feel strongly about it.  Should this be moved or should we stop using `haiku-apps`?

PR #7355, opened 2022-11-01
nginx: bump version

kenmays:
** SECURITY UPDATES
- http://[nginx.org/en/CHANGES](http://nginx.org/en/CHANGES)


Begasus:
Maybe enable secondary architecture? Builds OK on 32bit and launches fine.

PR #7359, opened 2022-11-02
Fix zziplib cmake build error

danboid:
Fix zziplib cmake build error by specifying build type.


korli:
Please bump the revision.

danboid:
Done.

PR #7366, opened 2022-11-04
wxgtk: enable SDL

davidkaroly:


davidkaroly:
This change allows us to have proper sound output from wxGTK.
before: samples/sound builds successfully but it doesn't do anything (no sound output neither in synch nor in asynch mode)
after: samples/sound is able to play wav files

PR #7266, opened 2022-10-08
new recipe: libgrapheme

hoanga:
the following recipe adds support for the [libgrapheme](https://libs.suckless.org/libgrapheme/)

checked install on 
```
Haiku shredder 1 hrev56521 Oct  7 2022 07:25:26 x86_64 x86_64 Haiku
```

PR #7368, opened 2022-11-04
libgrapheme: bump to 2.0.2 and further simplify the build

FRIGN:
I was glad to see that a recipe for libgrapheme was pushed (see PR #7266), but saw some mistakes.

The version used is not the latest from upstream (2.0.2) and the patch to unset LDCONFIG wasn't necessary.

The build does not require sed(1) and there was a mistake in the description. I also reordered the make(1)-arguments as required by the standard.

Thanks in advance for your review!

Signed-off-by: Laslo Hunhold <dev@frign.de>


dev-libs/libgrapheme/libgrapheme-2.0.2.recipe
```diff
@@ -47,8 +44,8 @@ BUILD()
 
 INSTALL()
 {
-	make install LIBPREFIX=$libDir INCPREFIX=$includeDir \
-		MANPREFIX=$manDir $jobArgs
+	make $jobArgsLIBPREFIX=$libDir INCPREFIX=$includeDir \
```
augiedoggie, line 47:
```suggestion
	make $jobArgs LIBPREFIX=$libDir INCPREFIX=$includeDir \
```

augiedoggie:
Thanks!

FRIGN:
> Thanks!

Thank you for fixing my small mistake, @augiedoggie, which I overlooked. Keep up the great work! :)

PR #7378, opened 2022-11-06
audacity: update to version 2.4.2. Still very experimental.

davidkaroly:


davidkaroly:
continuation of the work started in #7347

some improvement over previous 2.3.3 recipe:
* now we're able to build with nyquist, LV2, portmixer
* not so many crash on exit
* the strftime() assertion seems gone - not entirely sure about this as I was also playing around with fixing this on wxGTK side

most of the limitations are still there unfortunately:
* no Haiku backend for portaudio so no audio devices are detected. An audio editor without sound :(
* flickering
* no ffmpeg, no midi
* keyboard shortcuts don't work very well

image

PR #7382, opened 2022-11-07
haikuwebkit: fix build

lonemadmax:
Just enough to make WebPositive run again in master until we have a shiny new version.

**Only tested in master x86_64**


haiku-libs/haikuwebkit/haikuwebkit-1.8.4.recipe
```diff
@@ -85,6 +85,8 @@ BUILD_PREREQUIRES="
 	cmd:which
 	"
 
+PATCHES="haikuwebkit-$portVersion.patchset"
```
waddlesplash, line 88: wrong place, goes at the top block

PR #7385, opened 2022-11-08
Calendar: update to latest git-rev.

OscarL:
Fixes the latest "missing symbols" issue, and also makes it possible to build on Haiku after hrev55770 (changes on BBitmap's ImportBits() methods).

Tested on x86_64.


Begasus:
As the build has been fixed, I think this one can be closed until there is a newer version around?

PR #7386, opened 2022-11-09
BePDF: switch to latest git commit.

OscarL:
Switching to latest git commit, and dropping freetype_config in favor of pkg_config fixes the builds on newer Haiku revisions.

Builds and runs OK on both 64 and 32 bits (hrev56584).


humdingerb:
Thanks!

PR #7401, opened 2022-11-12
64tass: Add updated recipe for 1.57.2900

soci64:
Tested with x86_gcc2, should work everywhere else in theory. Changed to upstream URL, corrected copyright and licensing. The makefile doesn't need patching with overriding the path variables. The removed licenses were replaced with symlinks in the documentation directory. Text version of the manual is not included (redundant and is missing Unicode characters). Binary is stripped now.


dev-lang/64tass/64tass-1.57.2900.recipe
```diff
@@ -0,0 +1,66 @@
...
+LICENSE="GNU GPL v2
+	GNU LGPL v2
+	GNU LGPL v2.1
+	my_getopt"
+REVISION="1"
+SOURCE_URI="https://downloads.sourceforge.net/project/tass64/source/64tass-$portVersion-src.zip"
+CHECKSUM_SHA256="e4f0dcb1b00d5206ccb391eb5b8b43f4d95f064d63af739a71f5c822bbfc56a4"
+SOURCE_DIR="64tass-$portVersion-src"
+
+ARCHITECTURES="all ?arm ?ppc ?sparc"
```
augiedoggie, line 19:
```suggestion
ARCHITECTURES="all !x86_gcc2"
```
Does the port build with gcc2? If not then it should be excluded from the primary architectures.  The other untested arches can be removed.

augiedoggie (changes requested):

dev-lang/64tass/64tass-1.57.2900.recipe
```diff
@@ -0,0 +1,66 @@
...
+HOMEPAGE="https://sourceforge.net/projects/tass64/"
+COPYRIGHT="Daniel Berlin
+	2001 Benjamin Sittler
+	2010 Franck Bui-Huu
+	Alain Magloire
+	2002-2022 Zsolt Kajtar"
+LICENSE="GNU GPL v2
+	GNU LGPL v2
+	GNU LGPL v2.1
+	my_getopt"
```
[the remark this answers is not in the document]
soci64: Actually that's a MIT license. For some reason the original packager put it in the licenses directory and I kept it that way. However now that I thought about it more it's better to just refer to the MIT license is the recipe and put the license of my_getopt to the documentation directory. After all it's a variant with a specific attribution not the generic template.

dev-lang/64tass/64tass-1.57.2900.recipe
soci64: Yes it does :) When it was packaged last time there were incompatibilities however these were resolved since then. That's one of the reasons why I'd like to have it updated. So I've changed it to "all".

dev-lang/64tass/64tass-1.57.2900.recipe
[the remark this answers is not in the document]
augiedoggie: Sorry, I was only half awake when I started reviewing this pull request.  I thought this was a new recipe instead of an update.

dev-lang/64tass/64tass-1.57.2900.recipe
augiedoggie: I think this is going to cause a conflict on 32 bit systems.  If you install `64tass` and `64tass_x86` then both packages will try to put the commands in /bin

dev-lang/64tass/64tass-1.57.2900.recipe
```diff
@@ -2,16 +2,21 @@ SUMMARY="Multi pass optimizing macro assembler for the 65xx series of processors
...
-	my_getopt"
+COPYRIGHT="Daniel Berlin
+	2001 Benjamin Sittler
+	2010 Franck Bui-Huu
+	Alain Magloire
+	2002-2022 Zsolt Kajtar"
+LICENSE="GNU LGPL v2
+	GNU LGPL v2.1
+	MIT
+	GNU GPL v2"
```
augiedoggie, line 13:
```suggestion
LICENSE="GNU GPL v2
	GNU LGPL v2
	GNU LGPL v2.1
	MIT"
```

Might as well sort the licenses while we're cleaning it up

dev-lang/64tass/64tass-1.57.2900.recipe
soci64: Installing both versions is not a good idea as all files would conflict. I've changed it as suggested, the newer GCC makes better code. Who really wants a gcc2 version can just force it.

PR #7403, opened 2022-11-13
botan, add patch for configure.py for cmd:python3

Begasus:


Begasus:
Build failure on 32bit buildmaster, local build on 32bit (nightly, most recent) was OK, looks like a conflicting type for` uint32`?

PR #7405, opened 2022-11-13
qputty: clean up recipe

davidkaroly:
* fix putty download url
* build debuginfo package
* show putty version in about box
* fix duplicate variable definition in qtwin.cpp


Begasus:
Build failure on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/3297/builds/78852.log

davidkaroly:
ah yes I can see it now... there's some trouble with `uint32` and `uint32_t`

I'll try to come up with a patch for that.

PR #7417, opened 2022-11-14
DrumCircle: update to 1.0.5

OscarL:
Tested on both 64 bits and 32 bits (x86_gcc2).

Patchset dropped, as it was merged upstream.


davidkaroly:
> I was wondering why it wasn't there already :-D

'twas my mistake I guess

PR #7421, opened 2022-11-16
audacity: build with ninja

davidkaroly:


media-sound/audacity/audacity-2.4.2.recipe
```diff
@@ -71,24 +71,24 @@ BUILD_PREREQUIRES="
...
-	cd build
-	cmake .. $cmakeDirArgs -DCMAKE_BUILD_TYPE=Release \
+	cmake -S . -B build -G Ninja \
+		$cmakeDirArgs \
+		-DCMAKE_BUILD_TYPE=Release \
 		-DUSE_MIDI=OFF \
 		-Daudacity_use_pa_oss=NO -Daudacity_use_pa_jack=off \
 		-Daudacity_use_ffmpeg=off -Daudacity_use_midi=off -Daudacity_use_pch=no
-	make $jobArgs
+	cmake --build build
```
augiedoggie, line 88: Have you tested this? Running cmake like this usually results in an error about not specifying the build type.  I don't think that "bug" has been fixed

Begasus: Haven't checked this, but should be good with `ninja -C build` ? (same for install)

davidkaroly: Yes I just had a successful build with this updated recipe.

augiedoggie: > Yes I just had a successful build with this updated recipe.

In strict mode?

davidkaroly: Do you mean strict mode for haikuporter? In that case we'll see it soon, I started a rebuild with `haikuporter -S` option enabled.

davidkaroly: OK I can confirm it, rebuild in strict mode was successful. First I removed work-2.4.2 folder to rebuild completely from a clean slate. Then I invoked haikuporter like this: `haikuporter --no-source-packages --get-dependencies -S audacity` and I got a successful pacakge build, no policy warnings/errors.

augiedoggie: Yeah, looks like @waddlesplash fixed this a year ago but the packaged version of haikuporter was too old.  You should keep the `$jobArgs ` though.  I restrict the number of parallel jobs when I do a build and don't want ninja automatically using all cores on my computer.

augiedoggie, line 88:
```suggestion
	cmake --build build $jobArgs
```

media-sound/audacity/audacity-2.4.2.recipe
```diff
@@ -71,24 +71,24 @@ BUILD_PREREQUIRES="
...
 }
 
 INSTALL()
 {
-	make -C build install
+	cmake --install build
```
augiedoggie, line 93:
```suggestion
	cmake --install build $jobArgs
```

davidkaroly:
thx :+1:

PR #7430, opened 2022-11-18
chicken, bump version, cleanup

Begasus:


dev-scheme/chicken/chicken-5.3.0.recipe
```diff
@@ -55,6 +55,7 @@ BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	"
 BUILD_PREREQUIRES="
+	cmd:chicken
```
Begasus, line 58: Seems chicken requires chicken to build now, if this one is OK it should be good to merge

PR #7456, opened 2022-11-20
Calendar: fix build

lonemadmax:


lonemadmax:
> (Wonder for how long as Calendar been broken on gcc2)

I don't see it in [Haiku Depot web](https://depot.haiku-os.org/#!/?repos=haikuports&arch=x86_gcc2&viewcrttyp=ALL&srchexpr=calendar), so I guess since at least the last recipe update?

I don't even have a VM with the 32bits version setup anymore, so I don't know if this last change in architecures is correct or what would be wanted. Maybe being just a bandage for the libshared tokens shake up, only x86_64 should be included.

Begasus:
As mentioned by @OscarL the recipe needs the $secondaryArchSuffix in the correct places to be able to build on 32bit.
Any change on upstreaming the patchset?

lonemadmax:
> "backported" that change to 0.1 here, as an interim solution until newer Calendar gets more work done (or dropping gcc2 gets decided).

Yup. The BBitmap change is already in upstream. The library order change is not needed there, I compiled their HEAD without it, maybe libnetservices is not used anymore in the new code.

PR #7468, opened 2022-11-22
(WebKit_GTK) Use jobs variable instead of hard coded thread count.

SamuraiCrow:
Just as it says on the tin.  $jobs is the number of threads requested from the Haikuporter script.


threedeyes:
I know. This is done on purpose to regulate memory consumption during build. My PC uses over 20 gigabytes of RAM when built with 16 jobs.

net-libs/webkit-gtk/webkit_gtk-2.38.2.recipe
```diff
@@ -231,7 +231,7 @@ BUILD()
 		-DUSE_WPE_RENDERER=OFF \
 		$cmakeDirArgs
 
-	make -C build -j4
+	make -C build -j $jobs
```
korli, line 234:
```suggestion
	make -C build -j $((jobs>4 ? 4 : jobs))
```

threedeyes: I have great doubts that our buildmaster will do even with these parameters. Last time I tried to build webkit-gtk, it hung and still does not respond. Maybe it makes sense to build packages on more powerful hardware and put ready-made binary packages in the ports, as it is done for qtwebengine, wine, rust?

SamuraiCrow:
Ok, Never mind.

net-libs/webkit-gtk/webkit_gtk-2.38.2.recipe
korli: I don't see how it is related to this patch.

korli: then disable the recipe

SamuraiCrow:
Ok.  Never mind.  I'll do changes locally to make my build faster.

pulkomandy:
A comment in the recipe to explain why it is done this way would be nice. It will avoid other people trying to change this again?

Begasus:
Would be nice if in the description it would mention what this is for, now no-one sees it's about webkit-gtk unless looking at the changes. (Other PR also)

SamuraiCrow:
@Begasus
Done.

net-libs/webkit-gtk/webkit_gtk-2.38.2.recipe
waddlesplash: It looks like the HaikuWebKit recipe does not actually use any job-args. Instead it just uses Ninja, which automatically scales jobs based on memory and CPU cores. Those builds take a few hours but it's not too excessive. So, can we just make whatever changes are needed to use Ninja instead of Make, like the HaikuWebKit recipe does?

SamuraiCrow:
@waddlesplash 
Agreed.  I've resolved the merge conflict on my end so it should work now.

PR #7493, opened 2022-11-27
Icons for Edgar, Furnace, SuperTux and Bochs.

Handmaus:
Added icon files for Edgar, Furnace, SuperTux and Bochs.
Also added the vector_icon data to the rdef.in files that already existed and created new ones that were missing, but the ones created by me just contain icon data and nothing else. I guess they need to be completed?
Also edited the recipe files to include a line that points to the rdef file, but I suspect that maybe there's also some other needed changes that I missed, because I have seen other recipes that mention the rdef file in more than one place.


Begasus:
Thanks! When I get around to it I'll check each one of them in a build and report back. :+1:

Begasus:
Looks not to bad, iom files have been supplied, so if any one would update the icons they could, merging for now. Thanks!

PR #7511, opened 2022-12-03
audacity: add icon

davidkaroly:
Haiku-style icon for audacity by @threedeyes


Begasus:
Please bump REVISION, otherwise this won't get rebuild on buildmaster. :)

davidkaroly:
> Please bump REVISION, otherwise this won't get rebuild on buildmaster. :)

The recipe's still marked as untested so it won't build yet anyway. There's still at least a startup stability issue that needs to be investigated. Not to mention folder usage. Only for the adventurous at this point :wink:

PR #7521, opened 2022-12-05
filezilla: x86 secondary arch, icon, set XDG dirs

davidkaroly:
It's possible to build filezilla for x86 secondary architecture without any additional patching.

Unfortunately there's still one ingredient missing before we can enable it in HaikuDepot: currently if we start FileZilla from the command line it uses the correct folders in `~/config` but if we start it from deskbar then it uses the wrong folders `~/.cache` and `~/.config` because XDG_* environment variables are not set.


davidkaroly:
Icon from zuMi - see https://discuss.haiku-os.org/t/just-some-icons/12721/27
XDG initialization adapted from Epiphany

all in all I think this recipe is in a pretty good shape now

PR #7529, opened 2022-12-06
Nitroshare: added icon

kenmays:


net-wireless/nitroshare/patches/nitroshare-0.4.0~git.patchset
```diff
@@ -0,0 +1,32 @@
...
++++ b/libnitroshare/src/util/signalnotifier.cpp
+@@ -24,12 +24,18 @@
+ 
+ #include <QSocketNotifier>
+ 
++#ifdef __HAIKU__
++#  include <posix/signal.h>
++#  include <posix/sys/socket.h>
++#  include <unistd.h>
++#else
```
waddlesplash, line 19: `posix` in header path should not be needed. So, what's the problem here? Do we not have `sys/signal.h`? Just ifdef that one out then, don't duplicate this.

PR #7541, opened 2022-12-12
Remove static libraries from packages

Begasus:


Begasus:
> The changes on the broli and gettext recipes are doing more than just removing static libraries. Shouldn't those be separate changes instead? (edit: or do separate PRs for those documenting the other changes).
> 
> (might not matter much, but can help when searching for specific changes in commit logs)

For brotli the other change is to move from python3.7 to (the default) python3.9, for gettext it only affects the gcc2 version and is about the same as done for the new version?

Begasus:
No problem @OscarL maybe it's better to revert the change for brotli and not use the python changes, been looking into bzip2 recipe, that has some more changes and won't be included in here (and indeed plan to do a mass PR  for those static libs, so any input is welcome) :)

PR #7542, opened 2022-12-12
bzip2, fix some install paths, remove static library

Begasus:


Begasus:
Only checked on 32bit, would be nice if someone could check this on 64bit :)

humdingerb:
Builds and the tests say it's OK.

PR #7561, opened 2022-12-14
wput, disable x86_gcc2, don't rebuild (build is broken)

Begasus:


Begasus:
Cut down on the need to use the old gnutls on x86_gcc2

PR #7564, opened 2022-12-14
libpng, remove static libraries

Begasus:


media-libs/libpng/libpng16-1.6.38.recipe
```diff
@@ -77,8 +77,12 @@ INSTALL()
 {
 	make install
 
+	# remove libtool file
 	rm $libDir/libpng*.la
 
+	# remove static libraries
+	rm $libDir/libpng*.a
+
```
waddlesplash, line 85: same comment here as in the other PR: put removals in the same line.

PR #7582, opened 2022-12-17
libsdl, remove static library libSDL.a

Begasus:


Begasus:
This doesn't touch libSDLmain.a

PR #7607, opened 2022-12-20
qemu: update to 6.2.0, enable GTK frontend

davidkaroly:
successful build, no policy errors in haikuporter strict mode
smoke test looks OK

### DIFF
--- a/app-emulation/qemu/patches/qemu-6.2.0.patchset
+++ b/app-emulation/qemu/patches/qemu-6.2.0.patchset
@@ -1,20 +1,11 @@
-From e2fc63ba3e0f33018a2b0623e50feca92d234556 Mon Sep 17 00:00:00 2001
+From e4399a97f16a6a8b23ce95bafdec4df0d9c70ef9 Mon Sep 17 00:00:00 2001
 From: Alexander von Gluck IV <kallisti5@unixzen.com>
 Date: Tue, 18 May 2021 16:49:20 -0500
-Subject: [PATCH] haiku: fixes and patches, rebased from qemu 3.x
+Subject: haiku: fixes and patches, rebased from qemu 3.x
 
----
- disas/nanomips.h         |  4 ++++
- os-posix.c               | 10 ++++++++++
- slirp/src/tftp.c         |  5 +++++
- target/mips/mips-semi.c  |  4 ++++
- util/main-loop.c         |  2 ++
- util/notify.c            |  4 ++++
- util/qemu-thread-posix.c |  4 ++++
- 7 files changed, 33 insertions(+)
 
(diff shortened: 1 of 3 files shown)


khallebal:
@davidkaroly The 7.1.0 is already out, tested it on haiku it's working.

davidkaroly:
> @davidkaroly The 7.1.0 is already out, tested it on haiku it's working.

eventually we should go for 7.2.0 but that will need some more testing because of devendored libraries

khallebal:
> because of devendored libraries?

What is that? i'm unfamiliar with this term?
To be honest i only built it with the terminal not haikuporter, the only thing that didn't work was the keyboard shortcuts which is a wayland issue.

pulkomandy:
"vendored" libraries are when a project copies and pastes sourcecode from other libraries into their own repository, either to maintain their own patched versions, or to make sure their code is built against that one very specific version.

This is a hroblem for Haiku packaging because we often need our own patches to such libraries, so whenever possible we do the opposite known as "devendoring": reconfiguring the project to use our packaged libraries instead of their version. But sometimes the "vendored" version has patches or is newer

khallebal:
Aah ok! vendored/devendored, i didn't make the connection, thanks for the clarification.

Begasus:
@davidkaroly build in both 64bit and 32bit OK? (didn't check a build myself)

davidkaroly:
Haven't tried it on 32bit yet. Probably I'll do some testing on my 32bit Haiku VM over the weekend.

Begasus:
> Haven't tried it on 32bit yet. Probably I'll do some testing on my 32bit Haiku VM over the weekend.

If you can could you also check it builds with: https://github.com/haikuports/haikuports/pull/7674

Tried qemu-3.1.1.1_x86 on 32bit and build has been successful, only change there was "cmd:qemu_system_riscv32"
But I never really used qemu (for years) so it would be nice if it is OK with the changes for dtc (current version in haikuports fails).

davidkaroly:
Update: it builds successfully for x86 secondary architecture, even though there are a few policy violations. (no matching self provides for qemu_x86_blablabla)

Begasus:
> Update: it builds successfully for x86 secondary architecture, even though there are a few policy violations. (no matching self provides for qemu_x86_blablabla)
> 
Could you fix or copy the policy warnings (here)?

> Package names look a bit funny e.g. qemu_x86_arm but the best is qemu_x86_x86. Is there a way to remove the secondary arch suffix? Unless we'd like to build qemu with gcc2 but I'm not sure if that's a valid goal.

No valid goal for gcc2 (there is no GTK backend there.

PR #7618, opened 2022-12-23
w2c2:added Makefile patch so GCC2 can handle it

SamuraiCrow:
In addition to the commit mentioned in the title, I added an indication to the build that strdup is supplied by the runtimes so even the original version for x86_64 needs updating.

### DIFF
--- a/dev-lang/w2c2/w2c2-0.1~20221219.recipe
+++ b/dev-lang/w2c2/w2c2-0.1~20221219.recipe
@@ -13,7 +13,7 @@ Among its features are:
 HOMEPAGE="https://github.com/turbolent/w2c2"
 COPYRIGHT="2021 Bastian Mueller"
 LICENSE="MIT"
-REVISION="1"
+REVISION="2"
 srcGitRev="cd90cd995501d83ce41f57bfb11b1a52de092d20"
 SOURCE_URI="https://github.com/turbolent/w2c2/archive/$srcGitRev.zip"
 CHECKSUM_SHA256="8bc9699771b03f92237980e24805f74e4b866d57130089e968c038d364b0de2a"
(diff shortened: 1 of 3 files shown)


korli:
Could you send this patch upstream?

SamuraiCrow:
It's already been accepted upstream.  I can redo the recipe accordingly if you like.

korli:
Please do. It's always better to bump the git revision.

SamuraiCrow:
Ok.  I've been experimenting with storing fixed source archives in my tag releases on my fork on GitHub.  Would that be better yet?  If so, would .tar.xz be preferred over .zip?

SamuraiCrow:
## Update

The author just started adding CMake build support which is even better yet!  Hopefully I can build with Ninja Build instead soon.

dev-lang/w2c2/w2c2-0.1~20221219.recipe
```diff
@@ -37,10 +37,13 @@ BUILD_PREREQUIRES="
 	cmd:gcc
 	cmd:make
 	"
+PATCHES="
```
Begasus, line 40: PATCHES should come after CHECKSUM_SHA256, see: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#ordering

media-fonts/caligula/caligula-1.0.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="A script font"
```
Begasus, line 1: Is this supposed to be in here?

SamuraiCrow: I moved that to a different branch and tried to reset the branch.  Please revert.

PR #7624, opened 2022-12-24
guile, remove static libraries, switch SOURCE_URI, fix build

Begasus:


dev-scheme/guile/patches/guile1-1.8.8.patchset
```diff
@@ -55,3 +55,30 @@ index 2ecd8ae..fdaaad9 100644
 -- 
 2.24.1
 
+From 255c61ebe04c52c6ca24b1ace370eb34120e2555 Mon Sep 17 00:00:00 2001
+From: begasus <begasus@gmail.com>
+Date: Sat, 24 Dec 2022 10:59:32 +0000
+Subject: build fix
```
Begasus, line 61: Similar error as with boost: would be nice if someone could check if it's OK. :)

PR #7657, opened 2022-12-25
portaudio, disable static library, remove libtool files

Begasus:


Begasus:
@threedeyes static library is preventing a build, also don't think the *.la files are not needed, just asking as you did a great job on getting this to work.

PR #7659, opened 2022-12-26
dooble: bump version

kenmays:
Dooble 2022.12.25
Tested on hrev56673 x64.


Begasus:
@kenmays could you squash your commits before a merge (when done), no need to add a separate commit for every change in the PR. Thanks :)

Begasus:
On a second commit you can always use `git commit --amend`, this is enough to keep it to one commit (that's how I use it). :)

Begasus:
@kenmays Buid failure on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=&viewMode=expanded

kenmays:
QtWebEngine not available on x86.

kenmays:
For now, yes. - we can disable the x86 building.

PR #7661, opened 2022-12-26
Caligula: a Public Domain script font

SamuraiCrow:


media-fonts/caligula/caligula-1.0.recipe
```diff
@@ -0,0 +1,46 @@
...
+COPYRIGHT="1994 or earler"
+LICENSE="Public Domain"
+REVISION="1"
+SOURCE_URI="https://github.com/FontCollection/Caligula/releases/download/initial/Caligula.tar.xz"
+CHECKSUM_SHA256="dde42b6f09b85e92c2723d695fce907f0a07a74dff35f6b6094571fa43ac125c"
+SOURCE_FILENAME="Caligula.tar.xz"
+SOURCE_DIR="Caligula"
+
+ARCHITECTURES="any"
+DISABLE_SOURCE_PACKAGE="yes"
```
Begasus, line 18: Don't think it needed to be disabled?

Begasus (changes requested):

media-fonts/caligula/caligula-1.0.recipe
```diff
@@ -0,0 +1,46 @@
+SUMMARY="A script font"
+DESCRIPTION="\
```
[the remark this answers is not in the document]
SamuraiCrow: I'll fix it tomorrow.  Thanks.

media-fonts/caligula/caligula-1.0.recipe
SamuraiCrow: This part was copied from another font's recipe.  I can't really comment but I noticed something else that needs changing anyway.

SamuraiCrow: Fixed.

media-fonts/caligula/caligula-1.0.recipe
```diff
@@ -0,0 +1,43 @@
...
+
+ARCHITECTURES="any"
+
+PROVIDES="
+	caligula = $portVersion
+	"
+
+BUILD_PREREQUIRES="
+	cmd:cp
+	cmd:mkdir
```
Begasus, line 23: Haven't checked a build, but I guess these 2 are provided by default and not needed?

SamuraiCrow: You are correct.  I just tried it without these and it worked.  Once again, copied from an existing font recipe.  Oh well.

Begasus: No problem, depending on the recipe you used as a base it could be an older one. :) Thank for checking :+1:

media-fonts/caligula/caligula-1.0.recipe
```diff
@@ -0,0 +1,41 @@
...
+
+BUILD()
+{
+	md2html --github -f --fpermissive-autolinks --fstrikethrough --ftables --ftasklists --funderline --fwiki-links README.md --output=index.html
+}
+
+INSTALL()
+{
+	FONTDIR=$fontsDir/ttfonts
+	DOCDIR=$documentationDir/packages/Caligula
```
jmairboeck, line 33: You could just use `$docDir` instead (that uses the package name in lower case though).

SamuraiCrow: Fixed!  The DOCDIR variable has been replaced by the $docDir environment variable.

SamuraiCrow:
@Begasus 
This should be ready to go now.  (I hope.)

media-fonts/caligula/caligula-1.0.recipe
```diff
@@ -0,0 +1,39 @@
+SUMMARY="A script font"
+DESCRIPTION="Caligula is a crude representation of what was originally a \
+calligraphic bitmap script font autotraced by Fontographer 3.2. There are no \
+kerning pairs in this font and it should probably be used only in sizes 24pt \
+or larger."
+HOMEPAGE="https://github.com/FontCollection/Caligula"
+COPYRIGHT="1994 or earler"
```
Begasus, line 7: Still a bit puzzled on the COPYRIGHT here, it's only a date, I'll let other chime in if they have any thoughts on this. Thanks so far!

Begasus (approved):

media-fonts/caligula/caligula-1.0.recipe
SamuraiCrow: The only reason I made a note of it is that Creative Commons licenses weren't available at the time this design was written.  No author is stated and, apart from the documentation requirement, it was declared Public Domain in the U.S.  Would a relicense be appropriate in such a situation?

SamuraiCrow: If the author is unknown, I suppose there is no reason to list the copyright date at all.  Should I remove it altogether?

Begasus: No one stepped up, so guess we can merge this.

PR #7665, opened 2022-12-26
icu: enable icu67 for all non-gcc2 architectures

pulkomandy:
Currently the x86 architectures use icu66 while other architectures (arm, sparc, risc-v) are already migrated to icu 67. This complicates icu version management in Haiku. To make things simpler, enable icu 67 for everyone.

Confirmed booting without problems on x86_64 after doing the corresponding changes on Haiku side.

### DIFF
--- a/dev-libs/icu/icu67-67.1.recipe
+++ b/dev-libs/icu/icu67-67.1.recipe
@@ -41,18 +41,18 @@ wrapping when displaying the text."
 HOMEPAGE="http://www.icu-project.org"
 COPYRIGHT="1995-2020 IBM Corporation and others."
 LICENSE="ICU"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/unicode-org/icu/releases/download/release-67-1/icu4c-67_1-src.tgz"
 SOURCE_URI_2="https://github.com/unicode-org/icu/releases/download/release-67-1/icu4c-67_1-data.zip#noarchive"
-#SOURCE_URI_3="http://www.iana.org/time-zones/repository/releases/tzdata2019c.tar.gz#noarchive"
+SOURCE_URI_3="http://www.iana.org/time-zones/repository/releases/tzdata2022f.tar.gz#noarchive"
 CHECKSUM_SHA256="94a80cd6f251a53bd2a997f6f1b5ac6653fe791dfab66e1eb0227740fb86d5dc"
 CHECKSUM_SHA256_2="7c16a59cc8c06128b7ecc1dc4fc056b36b17349312829b17408b9e67b05c4a7e"
+CHECKSUM_SHA256_3="9990d71f675d212567b931fe8aae1cab7027f89fefb8a79d808a6933a67af000"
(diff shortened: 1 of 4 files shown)


pulkomandy:
Relaved changes on Haiku side: https://review.haiku-os.org/c/haiku/+/5966

waddlesplash:
We need to leave the older version around for a while before everything is rebuilt to avoid breaking things. I would also prefer we jump straight to the latest version; either we can patch WebKit if needed, or we can just build WebKit against the older version for the time being.

pulkomandy:
If we build WebKit against the older version, we have to ship that in Haiku anyway since WebKit will depend on it. The patch on WebKit is just replacing "TRUE" with "true", should not be a problem.

pulkomandy:
I tried ICU 72 and Haiku does not boot. Until this can be fixed (I have no idea how to debug it) it would be nice to switch to 67 at least?

waddlesplash:
> Until this can be fixed (I have no idea how to debug it) it would be nice to switch to 67 at least?

Anything in syslog? Presumably something crashed during launch_roster startup, this should appear in the syslog (for nightly builds at least.) Try booting in QEMU with serial out enabled.

Begasus:
@pulkomandy could you see if things are still fine with ICU70? If not add a conflict to ICU67 for ICU66 and ICU70 devel packages?
There are also 2 conflicts now with libxml2 (which I think our current status should be fine) and for harfbuzz.

dev-libs/icu/icu67-67.1.recipe
```diff
@@ -129,8 +129,8 @@ BUILD()
 	cd source
 
 	# Created during build and prevent rebuilds.
-	#rm -rf tools/tzcode/vanguard
-	#cp ../../../sources-3/tzdata*.tar.gz tools/tzcode/
+	rm -rf tools/tzcode/vanguard
+	cp ../../../sources-3/tzdata*.tar.gz tools/tzcode/
```
Begasus, line 133: These were disabled, is there any reason to enable this now? (maybe they should be enabled for future versions if there is a benefit from it)  :)

pulkomandy:
Using any version newer than 69 to build Haiku will resuilt in a non-bootable system. If this is not solved and haikuports starts using a newer ICU, we will have to ship two versions of ICU in the next release.

It seems no one on Haiku side is interested in fixing this at the moment. My patches to update to ICU 69 (the pest we can get) were rejected because we should update directly to the very latest version (which does not work).

Probably moving forward on haikuports will force the Haiku side to look at this problem and fix the things that need to be fixed. It is probably another case of ICU trying to use a C library function that we implement using ICU.

dev-libs/icu/icu67-67.1.recipe
korli: on older icu version the included tzdata is too old. so the recipe uses another source for tzdata and puts it in the right place for the build to take into account. This isn't needed on newest versions because they include up to date tzdata.

jmairboeck:
I guess this can be closed now.

PR #7678, opened 2022-12-28
speedtest_cli: update to latest release, and switch to Python 3.9.

OscarL:
Also: drop unsupported Python versions (and only provide a package for the default Python on beta4).

I saw a commit related to Python 3.10 support on the speedtest-cli repo, but it were newer than the 2.1.3 release, so... let this recipe update stick to 3.9 (works for me on beta4 with Python 3.9).


Begasus:
> Grrr, the original recipes has this line:
> 
> `pythonVersions=${PYTHON_VERSIONS[$i]}`
> 
> Notice the PLURAL in pythonVersion**s**. And then uses `$pythonVersion` (no **s**). Oh well, at least I found it before going even crazier :-D

That was the one, ran a build on 32bit and checked the binary, all good here :)

Begasus:
Only needed to change the pythonVersions vs pythonVersion. That worked, what other change did you do?

PR #7683, opened 2022-12-28
openldap, bump version, disable static library

Begasus:


Begasus:
Some packages rely on previous version and probably need a revbump for this:

- [ ] postgresql11 (12?)
- [ ] libreoffice
- [ ] pam_yubico
- [ ] bitlbee
- [ ] gnupg
- [ ] claws-mail
- [ ] apache
- [ ] lighttpd

Begasus:
postgresql11 and 12 fail to build with this is this doesn't contain libldap_r anymore, a fix was found at: https://postgrespro.com/list/thread-id/2558992

Unfurtunatly this didn't do the trick as the build fails in the end with: 
```
/boot/system/develop/tools/x86/bin/../lib/gcc/i586-pc-haiku/11.2.0/../../../../i586-pc-haiku/bin/ld: auth.c:(.text+0x25bc): undefined reference to `ldap_err2string'```

Begasus:
Updated to the latest release:

* disable static libraries
* added conflict for the tools package and the base package (2.4.*) on primary 32bit (both contain the binaries)
* added conflict for the devel$secondaryArchSuffix package
* renamed recipe to openldap2.6

This should not interfere with the existing base package so nothing gets broken, any feedback would be welcome. :)

PR #7685, opened 2022-12-28
lilypond: bump version

jmairboeck:
Note: the recipe is still disabled because it requires texlive to build. I will rebase #7316 when this is merged.


jmairboeck:
The tests were running successful (on x86_64).

PR #7687, opened 2022-12-28
smpeg2: fix build with modern GCC.

OscarL:
Cleaned up the recipe somehow. The resulting "plaympeg" is still non-functional, at least with the sample files I have:

On 64 bits: Shows a white, empty window.
On 32 bits: Shows a black, empty window.

That after properly recognizing the file as a video file, give the correct details of resolution and duration, but repeating the message:

"Warning: picture block before sequence header block".

Previous revisions (smpeg2-2.0.0-4) showed same behavior. Only difference is that old version (on 64 bits at least) loads "add-ons/opengl/Software Pipe" and prints:

"GalliumContext: CreateDisplay: Using llvmpipe (LLVM 12.0.1, 128 bits) driver"


Begasus:
Will give it a run in a bit, in the meantime, maybe disable the cmd all in all? (if possible)

Begasus:
Thanks!

PR #7691, opened 2022-12-29
libxmu, disable x86_gcc2, disable static libraries

Begasus:


Begasus:
@waddlesplash this is the first of the X libraries, marked x86_gcc2  as broken, if this is the way to go, I'll proceed with the others like this.

x11-libs/libxmu/libxmu-1.1.3.recipe
```diff
@@ -8,12 +8,12 @@ COPYRIGHT="1987, 1988 by Digital Equipment Corporation
...
 	1989, 1998 The Open Group"
 LICENSE="MIT (no promotion)"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://www.x.org/releases/individual/lib/libXmu-$portVersion.tar.bz2"
 CHECKSUM_SHA256="9c343225e7c3dc0904f2122b562278da5fed639b1b5e880d25111561bac5b731"
 SOURCE_DIR="libXmu-$portVersion"
 
-ARCHITECTURES="all"
+ARCHITECTURES="all !x86_gcc2"
```
waddlesplash, line 16: Don't bother adding this constraint. The recipe will still build if you add certain constraints to other recipes.

waddlesplash: It will currently fail to build on GCC2 but that's not a constraint of the software itself so it shouldn't be declared here, just allowed to implicitly fail.

PR #7697, opened 2022-12-29
fluidlite, disable static library

Begasus:


korli:
The Fluidlite static library  is used by the Haiku build: https://cgit.haiku-os.org/haiku/tree/build/jam/BuildFeatures#n262

Begasus:
I can revert the change, but wouldn't buildbot fail? (hmm ... guess not, it doesn't use the strict policy right?) pokes @waddlesplash

waddlesplash:
It doesn't use the strict policy yet but will do so eventually.

What should be done is manually `mv`ing the library and giving it a different name, probably `libfluidlite-static`, which an explanation of why it exists.

Begasus:
This should be changed for the Haiku build also I guess then? Will do a PR for that here tomorrow morning.

PR #7699, opened 2022-12-30
fluidlite, re-enable and rename static library used by Haiku builds

Begasus:
This addresses: https://github.com/haikuports/haikuports/pull/7697#issuecomment-1367583768


Begasus:
On the other hand, can't the Haiku build use the shared library from fluidlite instead of the static one?

korli:
The point of the static library was including it in libmidi.so.

PR #7705, opened 2022-12-31
zsh: rev-bump to fix libgdbm dependency

OscarL:
Needed after the update on the libgdbm recipe (#7704).

Tested on 64 bits.

~Not sure if this change is enough, or specifying that the compat is with "4.0" version only.~

~If that's the case, suggestions (or links to a good example) on what's the right syntax for that... are more than welcome.~


Begasus:
Maybe a revbump is needed to build with the regular libgdbm?

PR #7714, opened 2023-01-01
pycryptodome: new recipe

OscarL:
~First try. Missing secondary arch support.~ Works on 64 and 32 bits (x86, GCC11).

Apparently, from the same code base a package named `pycryptodomex` (mind the final "x") can be also created, that do not conflicts with `pycrypto`.

Edit:

Added a second recipe for the `pycryptodomex` version, that installs as the `Cryptodome` Python module (and thus does not conflicts nor replaces `Crypto`).

Not sure if combining the two recipes would be possible, or desirable. For now, this is a good first attempt.

### DIFF
--- a/dev-python/pycryptodome/pycryptodomex-3.16.0.recipe
+++ b/dev-python/pycryptodome/pycryptodomex-3.16.0.recipe
@@ -41,9 +41,6 @@ REQUIRES_$pythonPackage=\"\
 	lib:libgmp$secondaryArchSuffix\n\
 	cmd:python$pythonVersion\
 	\""
-eval "CONFLICTS_$pythonPackage='
-	pycrypto_$pythonPackage
-	'"
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	devel:libgmp$secondaryArchSuffix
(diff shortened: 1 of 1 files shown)


dev-python/pycryptodome/pycryptodomex-3.16.0.recipe
```diff
@@ -0,0 +1,88 @@
...
+LICENSE="BSD (2-clause)
+	Public Domain
+	Unlicense
+	"
+COPYRIGHT="2014-2022 Legrandin"
+REVISION="1"
+SOURCE_URI="https://pypi.python.org/packages/source/p/pycryptodome/pycryptodome-$portVersion.tar.gz"
+CHECKSUM_SHA256="0e45d2d852a66ecfb904f090c3f87dc0dfb89a499570abad8590f10d9cffb350"
+PATCHES="pycryptodomex-$portVersion.patchset"
+
```
Begasus, line 18: Needs: SOURCE_DIR="pycryptodome-$portVersion"
I would rather see the two of them put together in one recipe. :)

dev-python/pycryptodome/pycryptodomex-3.16.0.recipe
```diff
@@ -0,0 +1,97 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libgmp$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:cc$secondaryArchSuffix
```
Begasus, line 36: Maybe move this below in the `for` `done` statement?

PR #7721, opened 2023-01-03
tcl: update to version 8.16.13

jmairboeck:
The so version is the same as in 8.6.12 therefore the recipe was just renamed. OpenCASCADE's DRAW still runs.
Should I remove version 8.6.11?

Tests don't run successfully. Some tests hang, and one test even crashes with an assertion in free.


jmairboeck:
Debug report of tcltest [tcltest-2064-debug-03-01-2023-14-09-15.txt](https://github.com/haikuports/haikuports/files/10337883/tcltest-2064-debug-03-01-2023-14-09-15.txt)

PR #7725, opened 2023-01-04
R, bump version

Begasus:


Begasus:
@waddlesplash build failure on buildmaster for bott arch's here: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/3659/builds/79370.log

I think this is an issue with `cairo `vs `xcairo`, with cairo the build runs fine here, with xcairo I'm getting the same error as on buildmaster (on another check with xcairo I also got `undefined reference to XUnmapSubwindows`.

PR #7726, opened 2023-01-04
BeBook: fix mime types

lonemadmax:
The original files have the type attribute set already, with CSS files taken as text/x-source-code.


humdingerb:
@lonemadmax : That didn't appear to have fixed the issue. I think, it's because the buildbots don't yet have the fixed sniffer rule for CSS files. Until they do, the userguide uses this workaround:

```
	# workaround for incorrect sniffing
	settype -t "text/css" $docsDir/userguide/*.css
```

lonemadmax:
You are so right. Builder is at hrev56578+31, while the rule is in hrev56578+64.

PR #7737, opened 2023-01-05
BeBook: fix CSS mime type

lonemadmax:
This time for real, whatever sniffing rules the builder may have.


pulkomandy:
Yes, haikuporter already runs mimeset on all files in packages so that wouldn't have helped.

lonemadmax:
In this case not even the haikuporter run would help, as it doesn't force updating and the files already have a type. I guess we should fix the original files, but I've not found where to report on those.

pulkomandy:
haiku-files is owned by the haiku project so we can change that if we need to. I have access there and I can put up a new archive if someone provides one.

PR #7746, opened 2023-01-07
qemu, revbump for newer dtc package

Begasus:


Begasus:
New version not enabled, so this needs to be rebuild.

PR #7748, opened 2023-01-07
mpfr library: added 4.2.0 recipe

jeffsauer:
this is the latest version at this time.


jeffsauer:
qalculate_qt uses libqalculate which is dependent on mpfr library, hence updating each to their current latest versions

Begasus:
lib:libmpfr$secondaryArchSuffix is being used by other packages, have these been checked? https://github.com/haikuports/haikuports/search?q=lib%3Alibmpfr%24secondaryArchSuffix

If so maybe remove the old recipes so we can see the defferences in the packages. Thanks.

jeffsauer:
> lib:libmpfr$secondaryArchSuffix is being used by other packages, have these been checked? https://github.com/haikuports/haikuports/search?q=lib%3Alibmpfr%24secondaryArchSuffix
> 
> If so maybe remove the old recipes so we can see the defferences in the packages. Thanks.

I tested mpfr 4.2.0 with the following and all passed:

wcalc-2.5 tested and passed
mpc-1.2.1 self checks passed
older libqalculate-4.10, qalculate-4.1.0 worked fine with newer mpfr 4.2.0
kcalc self checked with knumbertest and passed

korli:
gcc also uses libmpfr AFAICT

PR #7752, opened 2023-01-08
audacity: update to 3.2.3

davidkaroly:
Audacity starting to take a good shape
* update to 3.2.3
* enable build for x86_64 and x86 secondary arch
* fix crash on exit - we need to release the sLocale smart pointer
* there's one folder that still doesn't look entirely right, we'll need to set XDG_STATE_HOME - not sure what its default value should be in Haiku
* there was something strange with portmidi on x86 secondary arch but I'm not sure and I was not able to reproduce the issue

image


Begasus:
> * there was something strange with portmidi on x86 secondary arch but I'm not sure and I was not able to reproduce the issue

This should be fixed now.

jmairboeck:
The portmidi port is also not functionally complete, see https://github.com/PortMidi/portmidi/pull/38#issuecomment-1279696894. I didn't implement everything completely.

PR #7718, opened 2023-01-02
libevent: update to 2.1.12, disable static libs, depend on python3

OscarL:
Tested on both x86_64 and x86_32 (GCC11).

PR #7754, opened 2023-01-08
libevent: fix-ups to previous PR (#7718)

OscarL:


korli:
The CONFLICTS_devel doesn't look good...

Begasus:
Did you check if there is "actualy" a $mandir in the 2.* version?

Begasus:
Sounds reasonable, make sure you do local builds before though ;)

PR #7755, opened 2023-01-08
R, put all in one package

Begasus:


korli:
@Begasus  an alternative is to use SUPPLEMENTS, so that subpackages are proposed when installing R.

Begasus:
@korli haven't used SUPPLEMENTS  before, could you point me to some examples where this is being used?

korli:
@Begasus example: https://github.com/haikuports/haikuports/blob/6b0baf342359b300794b7fc0fb7295f1c8363a0f/dev-libs/openssl/openssl-1.1.1s.recipe#L59 
a little pointless on openssl because it's a base package.

Begasus:
OK, thanks for t the link, think I need some more time to investigate this :) (back in RL with work so have to split up time for haikuports and other things 👍

PR #7756, opened 2023-01-08
dtc: re-add libfdt symlink

davidkaroly:
We need to reapply this workaround unfortunately. I was a bit too enthusiastic with #7607 and #7674 - while it's true that I was able to get a sucessful qemu build with new libfdt but I tested it only on old libfdt :disappointed:


pulkomandy:
Looks like the SONAME aren't set in the library quite right, and so prepareInstalledDevelLibs can't do its job correctly? Fixing that (passing the right linker flags) would fix this more cleanly and allow multiple apps to use different libfdt versions. But it may not be worth the effort.

Begasus:
I need to get a closer look into this, din't check libVersionCompat I think for this version, getting back when I get around to it (if  no one beats me to it (don't have as much time as I had in the past few weeks).

Begasus:
Looks like I missed this one when updating, thanks! (haven't checked a build for qemu so not sure why it should depend on the $libVersionCompat one (but as it  was in the previous recipe I guess it's OK to have it in)).

PR #7758, opened 2023-01-10
libevent: revbump packages that depend on libevent.

OscarL:
This is needed after the changes in the libevent package.

----

Opening as Draft, pending the merge of #7757.

`links`, `tmux` and `tor` were built and tested on beta4, both 64 and 32 bits (_x86), by me (using locally built `libevent-2.1.12-2` packages).

@Begasus did a build for `transmission` confirming it only needed a rev-bump to make it work with libevent-2.1.12.

Didn't included here a rev-bump for the remaining package that depends on libevent (`qtwebengine_bin`), as I think we not only need to rev-bump it, but also actually rebuild the binary packages against the updated libevent.


Begasus:
Transmission fixed, don't have time to look into "links" atm, but should be fairly simple to fix this ;)

PR #7759, opened 2023-01-10
library mpfr-4.2.0: added recipe

jeffsauer:
Add new recipe for mpfr-4.2.0 library.

The following test passed:

wcalc-2.5 tested and passed
mpc-1.2.1 self checks passed
libqalculate-4.1.0, qalculate-4.1.0 worked
libqalculate-4.5.1, qalculate-4.5.1 worked
kcalc self checked with knumbertest and passed
gcc-11.2.0_2021_07_28 compiles clean, not sure what additional tests to run?

Fails:
sci-mathematics/cgal (4.1x recipes circa 2019 pre-date mpfr 4.2.0) would not compile clean.  Trying to port cgal-5.5.1 release as that release should use the newer mpfr 4.2.0, but having difficulty porting.


korli:
please remove the old recipe

PR #7767, opened 2023-01-13
cudatext, 32bit version

Begasus:


Begasus:
Might as well push this by now, had been sitting around for a while here. :-)
Could use some checkes as libQt5pas has seen an update, so adding wip label for now.

Begasus:
@roired could you do the same for the 64bit version?

PR #7771, opened 2023-01-14
tcpdump: update to 4.99.3

davidkaroly:
The patch for `aclocal.m4` is still needed - could be upstreamed.

make check: 1 failed test case `nflog-e`, this is better test result than on any previous version so far.


davidkaroly:
see also the-tcpdump-group/tcpdump#1028

infrastation:
Thank you for taking care of this update. Please note that tcpdump does not require awk in `BUILD_PREREQUIRES` since release 4.99.2.

infrastation:
Correction: CMake does not require awk and `mkdep` no longer requires awk, but `configure` still seems to require it, so it looks like the dependency still stands.

PR #7774, opened 2023-01-14
x265: enabled assembly for 10bit & 12bit encoding libs

eblanca:
The recipe for x265 3.5 contained a statement -DENABLE_ASSEMBLY=OFF which caused encoding libraries for both 10 bit and 12 bit to be built without assembly optimizations.
I removed that statement and now nasm is used to build faster versions of the libs.

### DIFF
--- a/media-libs/x265/x265-3.5.recipe
+++ b/media-libs/x265/x265-3.5.recipe
@@ -7,7 +7,7 @@ the bit rate. x265 is a free software project implementing that standard."
 HOMEPAGE="http://x265.org/"
 COPYRIGHT="2013-2021 x265 Project"
 LICENSE="GNU GPL v2"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://bitbucket.org/multicoreware/x265_git/get/$portVersion.tar.gz"
 CHECKSUM_SHA256="5ca3403c08de4716719575ec56c686b1eb55b078c0fe50a064dcf1ac20af1618"
 # BitBucket sucks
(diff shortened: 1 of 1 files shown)


pulkomandy:
@extrowerk it seems you introduced this in the recipe when updating to version 3.5, do you remember what was the reason for it?

extrowerk:
> @extrowerk it seems you introduced this in the recipe when updating to version 3.5, do you remember what was the reason for it?

I can't recall, sorry.

extrowerk:
> @extrowerk it seems you introduced this in the recipe when updating to version 3.5, do you remember what was the reason for it?

I can't recall, sorry.

pulkomandy:
Ok then, let's merge and see if anyone finds a problem with it :)

Begasus:
Build failure on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/3684/builds/79405.log

waddlesplash:
Does 32-bit have a too-old version of nasm?

korli:
> Does 32-bit have a too-old version of nasm?

it's the same version as x86_64

korli:
"x265 does not support 32-bit builds when HIGH_BIT_DEPTH is enabled."  from https://bitbucket.org/multicoreware/x265_git/issues/198/assembly-code-fails-on-x32

korli:
Seems like Arch didn't build 10bit & 12bit encoding libs on x86: https://github.com/archlinux/svntogit-packages/commit/a82b1bf5e7e90f3049e0ef783b10be9d8ce587b7#diff-37538beb61ff63edebbf735dfcf39e5d732f49183d6beb097169d971875ca422L31

PR #7778, opened 2023-01-15
portmidi: bump version

jmairboeck:
also fix the requirement of cmd:javac (actually openjdk_default)


jmairboeck:
Technically, this looks like an upstream bug, because the version wasn't updated for the tag (but only later), but I think it is still better to fix the recipe instead of patching it :)

PR #7780, opened 2023-01-15
icu57, reanable base package (needed for libreoffice, calligra ...?) …

Begasus:
…on 32bit

### DIFF
--- a/dev-libs/icu/icu-57.2.recipe
+++ b/dev-libs/icu/icu-57.2.recipe
@@ -40,7 +40,7 @@ wrapping when displaying the text."
 HOMEPAGE="http://www.icu-project.org"
 COPYRIGHT="1995-2019 IBM Corporation and others."
 LICENSE="ICU"
-REVISION="3"
+REVISION="4"
 SOURCE_URI="https://github.com/unicode-org/icu/releases/download/release-57-2/icu4c-57_2-src.tgz"
 SOURCE_URI_2="http://www.iana.org/time-zones/repository/releases/tzdata2019c.tar.gz#noarchive"
 SOURCE_URI_3="https://github.com/unicode-org/icu/releases/download/release-57-2/icu4c-57_2-data.zip#noarchive"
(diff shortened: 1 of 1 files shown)


Begasus:
Related to: https://discuss.haiku-os.org/t/install-libreoffice-on-x86/12922

waddlesplash:
LibreOffice should be rebuilt on 32-bit instead.

Begasus:
Already tried that with a revbump on the buildmaster, seems that for some reason it still seems to depend on boot166?

waddlesplash:
Probably because boost166 is still available and is used by default. Its own recipe should be disabled, or rebuilt against a newer ICU.

Begasus:
Tried rebuilding boost166 localy, failed with errors, didn't investigate further at that time.

pulkomandy:
We should also disable the older boost recipes and update everything that still uses them.

But the question remains: is there a way to query the package repo to know what depends on a package? Can pkgman do this or do we need tools in the web interface? It would help a lot for preparing these transitions and making them smoothly.

Begasus:
Tried rebuilding boost166 localy, failed with errors, didn't investigate further at that time.

> Did you try `pkgman search -r "lib:libicuuc>=57"` ?

Doesn't reveil any matches?

Begasus:
Boost166 fails with: error: no matching function for call to 'decay_copy(void (&)(rlimit*) noexcept)' (so far no idea how to solve this)

pulkomandy:
> Did you try pkgman search -r "lib:libicuuc>=57" ?

It doesn't seem to work when specifying a version.

> Boost166 fails with: error: no matching function for call to 'decay_copy(void (&)(rlimit*) noexcept)' (so far no idea how to solve this)

We can't provide long term maintenance for old boost versions. They often don't work with newer compilers. So we should just make sure everything does use the newer versions.

If we can't find out what depends on boost 1.66, maybe we should remove boost 1.66 from the repository and see what the next complaining package is? But really this isn't a great way to solve this.

Begasus:
LO for 32bit is  building atm, should fix this so it can be closed when all things are OK.

Begasus:
Fixed with revbump for LO

PR #7789, opened 2023-01-19
qemu-7.2.0: new recipe

davidkaroly:
Successful build for both 64-bit and 32-bit, no policy warnings :heavy_check_mark: 
Running Haiku inside Haiku :heavy_check_mark: 
Network access with devendored libslirp :heavy_check_mark: 

Wonky mouse pointer :x:
Ctrl/Alt/Shift doesn't work :x:

Edit:
These two issues were there also on 6.2.0
It seems to be much worse on qemu-3.1.1.1 (unless I'm doing something wrong?)

### DIFF
--- a/app-emulation/qemu/patches/qemu-7.2.0-haiku.patchset
+++ b/app-emulation/qemu/patches/qemu-7.2.0-haiku.patchset
@@ -1,0 +1,1549 @@
+From 573976c4edb7df7a10f46fdd2daf6f6c7faa086a Mon Sep 17 00:00:00 2001
+From: David Karoly <david.karoly@outlook.com>
+Date: Sat, 4 Feb 2023 13:39:11 +0100
+Subject: introduce Haiku ui
+
+
+diff --git a/meson.build b/meson.build
+index 5c6b5a1c75..9600599fc1 100644
+--- a/meson.build
++++ b/meson.build
+@@ -387,6 +387,7 @@ hvf = not_found
+ midl = not_found
+ widl = not_found
+ pathcch = not_found
++haiku = not_found
+ host_dsosuf = '.so'
+ if targetos == 'windows'
+   midl = find_program('midl', required: false)
+@@ -412,6 +413,11 @@ elif targetos == 'haiku'
+   socket = [cc.find_library('posix_error_mapper'),
+             cc.find_library('network'),
+             cc.find_library('bsd')]
++  if not get_option('haiku').disabled()
++    haikulibs = [cc.find_library('be'),
++                 cc.find_library('game')]
++    haiku = declare_dependency(dependencies: haikulibs)
++  endif
+ elif targetos == 'openbsd'
+   if get_option('tcg').allowed() and target_dirs.length() > 0
+     # Disable OpenBSD W^X if available
+@@ -1826,6 +1832,7 @@ config_host_data.set('CONFIG_BRLAPI', brlapi.found())
+ config_host_data.set('CONFIG_COCOA', cocoa.found())
+ config_host_data.set('CONFIG_FUZZ', get_option('fuzzing'))
+ config_host_data.set('CONFIG_GCOV', get_option('b_coverage'))
++config_host_data.set('CONFIG_HAIKU', haiku.found())
+ config_host_data.set('CONFIG_LIBUDEV', libudev.found())
+ config_host_data.set('CONFIG_LZO', lzo.found())
+ config_host_data.set('CONFIG_MPATH', mpathpersist.found())
+@@ -3895,6 +3902,9 @@ if targetos == 'darwin'
+   summary_info += {'Cocoa support':           cocoa}
+   summary_info += {'vmnet.framework support': vmnet}
+ endif
++if targetos == 'haiku'
++  summary_info += {'Haiku support':   haiku}
++endif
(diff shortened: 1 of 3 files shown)


khallebal:
> Wonky mouse pointer x
Ctrl/Alt/Shift doesn't work x

The first issue is likely dependent on the pointing device you choose, i did not experience it using `-usb -device usb-tablet`, the second issue is a wayland problem as i mentioned it in the other PR [.7607](https://github.com/haikuports/haikuports/pull/7607)

davidkaroly:
yep it works much better with `-usb -device usb-tablet`

Begasus:
Good to merge?

davidkaroly:
> Good to merge?

I wanted to do one more update as the patch for `os-posix.c` is not needed anymore.

davidkaroly:
> Good to merge?

it's marked as `"?all"` but I think it can be meged now as it isn't any worse than qemu-6.2.0

we can workaround the wonky mouse pointer with `-usb -device usb-tablet`

we can workaround the lack of Ctrl/Alt/Shift modifier keys with `-console sdl` (unfortunately they don't work on GTK UI)

davidkaroly:
> The first issue is likely dependent on the pointing device you choose, i did not experience it using `-usb -device usb-tablet`, the second issue is a wayland problem as i mentioned it in the other PR [.7607](https://github.com/haikuports/haikuports/pull/7607)

yep you're right, it seems related to GTK-on-Wayland on Haiku
* qemu with SDL UI works fine
* qemu with GTK-on-Wayland does not recognize Crtl/Alt/Shift modifier keys as their scancodes are not dispatched to `gd_key_event()` handler
* qemu with GTK UI on Fedora works fine (probably it's GTK-on-X but it would be good to try it also with Wayland eventually)

davidkaroly:
ok, the issue with ctrl/alt/shift seems to be related to the xkb keymap generation logic in Wayland inproc server.

The key codes for Ctrl/Alt/Shift are dispatched from Wayland server to GTK but they are not recognized by GTK as they are not mapped to a valid keysym in the xkb keymap. So GTK does not generate any key event for them.

I can get Ctrl/Alt/Shift working in qemu if I add the mapping for codes 0x4b, 0x56, 0x5c, 0x5d, 0x5f, 0x60 in `WriteSymbol()` (map them to Shift_L, Shift_R, Control_L, Alt_L, Alt_R, Control_R respectively)

khallebal:
If you've tested it and it works, you can create a PR at [@X547](https://github.com/X547/wayland-server/pulls), he is probably busy doing other stuff.

davidkaroly:
> If you've tested it and it works, you can create a PR at [@X547](https://github.com/X547/wayland-server/pulls), he is probably busy doing other stuff.

I tried it only with qemu. Some more testing will be needed with other applications.

khallebal:
AWESOME, thanks for working on this.

davidkaroly:
> AWESOME, thanks for working on this.

credit goes to @mmlr

mmlr:
This is indeed very cool to see live on. I wouldn't have expected it to still be applicable after this long a time and am glad that it could still be useful. Thanks for your work!

waddlesplash:
Can we build just the SDL2 and Haiku-native backends, then, and skip the GTK3 one for now if the key shortcuts are kind of broken? Plus that will cut out a lot of dependencies...

davidkaroly:
Sure we can disable GTK UI for the time being

khallebal:
Is the SDL2 backend here just an alternative or is it somehow tied to the haiku backend?

davidkaroly:
Well i think it doesn't get much better than this... we can use either SDL display+audio or Haiku native display+audio depending on command line args. (however take care not to mix them, Haiku display with SDL audio results in a crash)
GTK display disabled for now, because of the Wayland keymap issue with the Ctrl/Alt/Shift modifier keys.

Still I think it works better than version 3.1.1.1 and we can get up to date qemu feature set.

khallebal:
Did you notice any difference in performance when using the haiku backend compared to the other backends? CPU load?

pulkomandy:
Nice, thanks for working on this :)

Is it useful to keep all the older recipes for qemu as well, or should we delete them?

Should we keep the SDL2 backend enabled, or should we leave only the Haiku one to reduce dependencies?

Any plans for upstreaming the patch?

davidkaroly:
> Did you notice any difference in performance when using the haiku backend compared to the other backends? CPU load?

I haven't really noticed any difference in performance, it's rather slow either way :man_shrugging: emulation only, no HW acceleration...


> Is it useful to keep all the older recipes for qemu as well, or should we delete them?

I don't really see much use in the older recipes, and anyway github will remember them in the commit history - so I think they can be removed.

> Should we keep the SDL2 backend enabled, or should we leave only the Haiku one to reduce dependencies?
> 
> Any plans for upstreaming the patch?

At least a few people should try the native backend first.

davidkaroly:
Any further review comments?

khallebal:
I'd say just go ahead and commit  it so that it can be tested by more users, a lot of the users don't know about it, you might even make an announcement on the forum to get more feedback.

Begasus:
Any more work needed here (didn't check a build myself)?

PR #7793, opened 2023-01-21
app-arch/ancient: Version bump + added library

mywave82:
As usual, all comments are welcome!

This is needed for the future update for opencubicplayer

### DIFF
--- a/app-arch/ancient/ancient-2.0.0.recipe
+++ b/app-arch/ancient/ancient-2.0.0.recipe
@@ -1,0 +1,88 @@
+SUMMARY="Modern decompressor for old data compression formats"
+DESCRIPTION="This is a collection of decompression routines for old formats \
+popular in the Amiga, Atari computers and some other systems from 80's \
+and 90's as well as some that are currently used which were used in a some \
+specific way in these old systems."
+HOMEPAGE="https://github.com/temisu/ancient"
+COPYRIGHT="2017-2022, Teemu Suutari and contributors"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/temisu/ancient/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="9c6b31e4d968ec3adb25596a0b30405d1080d3de0546f1485ecb0ac2eca6261d"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libVersion="$portVersion"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	ancient$secondaryArchSuffix = $portVersion
+	cmd:ancient$commandSuffix = $portVersion
+	lib:libancient$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	ancient${secondaryArchSuffix}_devel = $portVersion
+	devel:libancient$secondaryArchSuffix = $libVersionCompat
+	"
+REQUIRES_devel="
+	ancient$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	autoconf_archive
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoreconf
+	cmd:automake
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage ancient$secondaryArchSuffix \
+	"$libDir"/libancient.so.$libVersion
+
+BUILD()
+{
+	./autogen.sh
+
+	runConfigure --omit-dirs binDir ./configure \
+		--disable-static \
+		--bindir=$commandBinDir
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	rm -f "$libDir"/libancient.la
+
+	prepareInstalledDevelLib libancient
+	fixPkgconfig
+
+	# devel package
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
+}


dev-libs/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+
+ARCHITECTURES="all ?x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+libVersion="2.0.0"
+libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
+
+PROVIDES="
+	ancient$secondaryArchSuffix = $portVersion
+	cmd:ancient = $portVersion
```
augiedoggie, line 19: This should be `cmd:ancient$secondaryArchSuffix` or the recipe should be set up for `$commandBinDir` style of `cmd` entries

dev-libs/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,76 @@
+SUMMARY="A collection of old decompression routines"
+DESCRIPTION="This is a collection of decompression routines for various old formats that \were populare on Amiga, Atari computers and other old systems. A simple \command-line application is provided aswell."
+HOMEPAGE="https://github.com/temisu/ancient"
+COPYRIGHT="2017-20222, Teemu Suutari and contributors"
```
augiedoggie, line 4:
```suggestion
COPYRIGHT="2017-2022, Teemu Suutari and contributors"
```

dev-libs/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,76 @@
+SUMMARY="A collection of old decompression routines"
+DESCRIPTION="This is a collection of decompression routines for various old formats that \were populare on Amiga, Atari computers and other old systems. A simple \command-line application is provided aswell."
```
augiedoggie, line 2: The newlines need to be fixed here.

augiedoggie (changes requested):
The portage overlays don't have a listing for this project but I wonder if it would be better in `app-arch/libancient`.  I don't have strong feelings about it though.

dev-libs/ancient/ancient-2.0.0.recipe
Begasus, line 2: DESCRIPTION line up to 100 characters, see: https://github.com/haikuports/haikuports/wiki/HaikuPorter-Guidelines#description

dev-libs/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+HOMEPAGE="https://github.com/temisu/ancient"
+COPYRIGHT="2017-20222, Teemu Suutari and contributors"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/temisu/ancient/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="9c6b31e4d968ec3adb25596a0b30405d1080d3de0546f1485ecb0ac2eca6261d"
+SOURCE_DIR="ancient-$portVersion"
```
Begasus, line 9: Not checked, but looking at the recipe name and the version SOURCE_DIR shouldn't be needed?

dev-libs/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+REQUIRES_devel="
+	ancient$secondaryArchSuffix == $portVersion base
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	autoconf_archive
+	pkgconfig
```
Begasus, line 39: It's `cmd:pkg_config$secondaryArchSuffix`

mywave82:
> app-arch/libancient

I totally missed app-arch/ancient which is an older version of the same library - need to merge new version, this recipe and comments into a version update instead

mywave82:
I did not keep the commandBinDir logic from the v1.0 recipe. I do not need if this is still needed and do not have the experience to test it.

dev-libs/ancient/ancient-2.0.0.recipe
mywave82: Worked great without it

dev-libs/ancient/ancient-2.0.0.recipe
mywave82: Updated

dev-libs/ancient/ancient-2.0.0.recipe
mywave82: I used $secondaryArchSuffix for now, unsure if $commandBinDir is needed. I do not know how to test

mywave82: (it went missing in my commit, it will be updated very soon - just doing a build test first)

app-arch/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,79 @@
...
+and 90's as well as some that are currently used which were used in a some \
+specific way in these old systems."
+HOMEPAGE="https://github.com/temisu/ancient"
+COPYRIGHT="2017-2022, Teemu Suutari and contributors"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/temisu/ancient/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="9c6b31e4d968ec3adb25596a0b30405d1080d3de0546f1485ecb0ac2eca6261d"
+
+ARCHITECTURES="all ?x86_gcc2"
```
Begasus, line 13: Doesn't build with gcc2, so:
```diff
-ARCHITECTURES="all ?x86_gcc2"
+ARCHITECTURES="all !x86_gcc2"
```

dev-libs/ancient/ancient-2.0.0.recipe
Begasus: While this is fine for primary architecture on 64bit where `$binDir` is the same as `$prefix/bin`, on 32bit you need to use secondary architecture where `$binDir` is the same as `$prefix/bin/x86`, so best to use $commandBinDir here, if you need guidance just give a yell :)

app-arch/ancient/ancient-2.0.0.recipe
mywave82: I will probably look at it tomorrow, and also flatten down the commits to one single.


The v1.0 recipe had this chunk, which matches the mini-guide in the wiki, which I add back in
```
ARCHITECTURES="?all !x86_gcc2"
SECONDARY_ARCHITECTURES="x86"

commandBinDir=$binDir
	commandSuffix=
	commandBinDir=$prefix/bin
fi
```
So even if x86_gcc is never built, the copy-paste block is a standard code block.



PROVIDES should then provide
cmd:ancient$commandSuffix = $portVersion

and in INSTALL, after `make install`, I need to do a mv if `$binDir/ancient` and `$commandBinDir/ancient$commandSuffix` does not match


or do like bzip2, do not provide commands in secondaryArch
```
	#remove bin commands and man pages when building for secondary arch
	if [ -n "$secondaryArchSuffix" ];then
		rm -rf $prefix/bin
		rm -rf $manDir
	fi
```

Begasus: > or do like bzip2, do not provide commands in secondaryArch

There is a difference if the primary binary is build, then there is no need for it in secondary architecture (as in this case for bzip2)

mywave82: Updated with !x86_gcc2 details, and flattened the commit log

mywave82: @Begasus does recipe look ok in its current shape?

app-arch/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,87 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+commandSuffix=$secondaryArchSuffix
+commandBinDir=$binDir
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	commandSuffix=
+	commandBinDir=$prefix/bin
+fi
+
+libVersion="2"
```
Begasus, line 23: In this case libVersion is the same as portVersion (2.0.0), so:
```diff
-libVersion="2"
+libVersion="$portVersion"
```

app-arch/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,87 @@
...
+		--bindir=$commandBinDir
+	make $jobArgs
+}
+
+INSTALL()
+{
+	make install
+	rm -f "$libDir"/libancient.la
+
+	fixPkgconfig
```
Begasus, line 77: Install de develop library:
```diff
-	fixPkgconfig
+	prepareInstalledDevelLib libancient
+	fixPkgconfig
```

Begasus (changes requested):

app-arch/ancient/ancient-2.0.0.recipe
mywave82: SO files only have "2" in them

app-arch/ancient/ancient-2.0.0.recipe
mywave82: Fixed, thank you

app-arch/ancient/ancient-2.0.0.recipe
Begasus: If you look in the devel package you'll see `libancient.so.2`, `libancient.so.2.0.0` and `libancient.so` in there.

Begasus:
```
/boot/system/lib/x86> tree | grep libancient
├── libancient.so.2 -> libancient.so.2.0.0
├── libancient.so.2.0.0
```

mywave82: I just tested again something compiled against libancient, and did a lddtree, and libancient.so.2 is the only file name the resulting binary searches for.

But if is preferred to rather use 2.0.0 the provides, I can update recipe later when I return from work trip

Begasus: As you can see in the above tree search, `libancient.so.2` is linked against `libancient.so.2.0.0`

mywave82: This has now been updated

app-arch/ancient/ancient-2.0.0.recipe
```diff
@@ -0,0 +1,88 @@
...
+	autoconf_archive
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:aclocal
+	cmd:autoconf
+	cmd:autoreconf
+	cmd:automake
+	cmd:libtoolize$secondaryArchSuffix
+	cmd:make
+	cmd:g++$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
```
Begasus, line 56: Please sort alphabeticly

mywave82: Fixed :-)

Begasus (approved):

Begasus:
@augiedoggie if this looks ok to you I'm fine here to merge :)

augiedoggie (approved):

Begasus:
Thanks  @mywave82 !

PR #7796, opened 2023-01-22
libexttextcat, bump version, disable static, move cmd to tools package

Begasus:


Begasus:
Builds fine with gcc2 also, but as there is no need for it I didn't enably it.

PR #7799, opened 2023-01-24
fpc: link against libiconv

hfsfox:


dev-lang/fpc/fpc-3.2.2.recipe
```diff
@@ -115,6 +116,7 @@ REQUIRES_examples="
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	libiconv_devel
```
korli, line 119:
```suggestion
	devel:libiconv$secondaryArchSuffix
```

Begasus:
@hfsfox any chance on getting this fix upstreamed? Would be nice to have that fixed when building lazarus with fpcupdeluxe.

PR #7806, opened 2023-01-27
tenacity-1.3~beta: new recipe

davidkaroly:
Tenacity 1.3 beta is out - so I'm posting a preliminary version of this recipe here.
Build/install/smoketest seems OK on x86_64.

image

### DIFF
--- a/media-sound/tenacity/tenacity-1.3~beta.recipe
+++ b/media-sound/tenacity/tenacity-1.3~beta.recipe
@@ -1,0 +1,148 @@
+SUMMARY="An easy-to-use, cross-platform multi-track audio editor/recorder"
+DESCRIPTION="Tenacity is an easy-to-use, cross-platform multi-track \
+audio editor/recorder for Windows, macOS, Linux and other operating \
+systems and is developed by a group of volunteers as open-source software."
+HOMEPAGE="https://www.tenacityaudio.org"
+COPYRIGHT="2021-2023 by Tenacity Team"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="v1.3-beta"
+srcGitDate="22 Jan 2023"
+SOURCE_URI="https://codeberg.org/tenacityteam/tenacity/archive/${srcGitRev}.tar.gz"
+CHECKSUM_SHA256="10577327867fc5463b72e19f94064feaf541b081ce6c882dbc0801311b80470d"
+SOURCE_DIR="tenacity"
+PATCHES="tenacity-$portVersion.patchset"
+ADDITIONAL_FILES="tenacity.rdef.in"
+
+ARCHITECTURES="?all !x86_gcc2"
+SECONDARY_ARCHITECTURES="?x86"
+
+PROVIDES="
+	tenacity$secondaryArchSuffix = $portVersion
+	app:Tenadacity$secondaryArchSuffix = $portVersion
+	cmd:tenacity
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	wxgtk$secondaryArchSuffix
+	lib:libatk_1.0$secondaryArchSuffix
+	lib:libatomic$secondaryArchSuffix
+	lib:libavcodec$secondaryArchSuffix
+	lib:libavformat$secondaryArchSuffix
+	lib:libavutil$secondaryArchSuffix
+	lib:libcairo$secondaryArchSuffix
+	lib:libexpat$secondaryArchSuffix
+	lib:libflac$secondaryArchSuffix
+	lib:libgdk_pixbuf_2.0$secondaryArchSuffix
+	lib:libglib_2.0$secondaryArchSuffix
+	lib:libgtk_3$secondaryArchSuffix
+	lib:libharfbuzz$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:libid3tag$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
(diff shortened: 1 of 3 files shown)


davidkaroly:
cc @n0toose

n0toose:
AWESOME!

media-sound/tenacity/tenacity-1.3~beta.recipe
```diff
@@ -0,0 +1,145 @@
...
+}
+
+INSTALL()
+{
+	cmake --install build
+
+	mkdir -p $appsDir
+	ln -s $binDir/tenacity $appsDir/Tenacity
+
+	local APP_SIGNATURE="application/x-vnd.tenacity"
```
n0toose, line 129: Unless if this is explicitly Haiku-specific (sorry, bit rusty), the signature should likely be different (something like `org.tenacityaudio.Tenacity`).

pulkomandy: it has to start with "application/x-vnd." because it's a MIME type. Since "x-" means "this is nonstandard" and "vnd" means "this is vendor specific", there are unfortunately no rules to how this namespace is used. It's usually "vendor name.app name" or something like that.

Not that it really matters, as long as it is unique, it should be fine.

n0toose: Yeah, was aware about `"application/x-vnd.", it's the capitalization that kind bothered me and I have had some issues specifically related to the application signature before -- they mostly concerned plugins and creating new windows in conjuction with Audacity, but it's safe to presume that an Audacity port will not contain such a conflict.

I was just trying to double check as I had some problems with creating new windows -- or that this signature is *Haiku-specific* in the sense that it does not affect the functionality of the application (which it shouldn't, as it gets baked into the resource file). I was worried as I had some issues in the past as I was testing OS-wide default applications when using Tenacity at the same time as Audacity on all platforms, but this should not be the case here because Haiku does things nicer.

n0toose: Anyways, I'm probably not making much sense and I glossed over this PR, please do not consider my comments and my attempts at giving extra context as some sort of a blocker.

media-sound/tenacity/tenacity-1.3~beta.recipe
```diff
@@ -0,0 +1,145 @@
...
+PATCHES="tenacity-$portVersion.patchset"
+ADDITIONAL_FILES="tenacity.rdef.in"
+
+ARCHITECTURES="?all !x86_gcc2"
+SECONDARY_ARCHITECTURES="?x86"
+
+PROVIDES="
+	tenacity$secondaryArchSuffix = $portVersion
+	app:Tenadacity$secondaryArchSuffix = $portVersion
+	cmd:Tenacity$secondaryArchSuffix
```
Begasus, line 23: Should be:
```diff
-	cmd:Tenacity$secondaryArchSuffix
+	cmd:tenacity$secondaryArchSuffix
```

media-sound/tenacity/tenacity-1.3~beta.recipe
```diff
@@ -0,0 +1,145 @@
...
+	cmd:gettext$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:ninja
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:which
+	cmd:wx_config$secondaryArchSuffix
+	"
+
+defineDebugInfoPackage tenacity$secondaryArchSuffix \
+	$binDir/tenacity$secondaryArchSuffix
```
Begasus, line 107: No need for $secondaryArchSuffix here:
```diff
-	$binDir/tenacity$secondaryArchSuffix
+	$binDir/tenacity
```

Begasus (changes requested):

Begasus:
Fails to launch (32bit) with: `runtime_loader: Cannot open file lib-ffmpeg-support.so`

davidkaroly:
> Fails to launch (32bit) with: `runtime_loader: Cannot open file lib-ffmpeg-support.so`

sounds like RPATH issue

Begasus:
> > Fails to launch (32bit) with: `runtime_loader: Cannot open file lib-ffmpeg-support.so`
> 
> sounds like RPATH issue

Crashes on launch atm (on 32bit), linking error seems to be solved.

Begasus (approved):

davidkaroly:
> Crashes on launch atm (on 32bit), linking error seems to be solved.

I tried it on 32-bit beta4. Launch, open MP3 file, cut&paste, export to WAV seems to work.

Did you try it on beta4 or nightly?

Begasus:
> > Crashes on launch atm (on 32bit), linking error seems to be solved.
> 
> I tried it on 32-bit beta4. Launch, open MP3 file, cut&paste, export to WAV seems to work.
> 
> Did you try it on beta4 or nightly?

R1B4, maybe something left behind from previous launch here?

davidkaroly:
> R1B4, maybe something left behind from previous launch here?

Have you tried deleting `config/settings/tenacity.cfg`?

Begasus:
> > R1B4, maybe something left behind from previous launch here?
> 
> Have you tried deleting `config/settings/tenacity.cfg`?

No such file here, if it works for you then maybe it could something on my end and we can merge, if others experience the same thing a seperate issue can be filed for it.

davidkaroly:
I also tried a build on a 32-bit nightly fresh install, it "works for me"

PR #7809, opened 2023-01-28
lilypond: require libgs >= 9

jmairboeck:
Ghostscript >= 9.03 is actually [required](http://lilypond.org/doc/v2.24/Documentation/contributor/requirements-for-running-lilypond).

This is untested, but a rather trivial change.

A revision bump shouldn't be needed here, correct?

PR #7811, opened 2023-01-29
texinfo 6.7: fix gcc2 build

jmairboeck:
Using texinfo 6.7 fixes the `info` files build failure of lilypond reported [here](https://github.com/haikuports/haikuports/pull/7809#issuecomment-1407463711)

Should we delete texinfo-6.1 now?

### DIFF
--- a/sys-apps/texinfo/patches/texinfo-6.7.patchset
+++ b/sys-apps/texinfo/patches/texinfo-6.7.patchset
@@ -1,4 +1,4 @@
-From 6e83e938550563738f6d2adc24dccc3c209b3692 Mon Sep 17 00:00:00 2001
+From 8c565d815ba2f11f772a762bf38685290b3c2171 Mon Sep 17 00:00:00 2001
 From: Jerome Duval <jerome.duval@gmail.com>
 Date: Tue, 22 Dec 2015 11:52:48 +0000
 Subject: Haiku patch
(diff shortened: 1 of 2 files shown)


pulkomandy:
I think there are problems compiling gcc2 while using newer texinfo versions (it fails to generate some manpages), so you may want to check that and make sure it still works before replacing texinfo-6.1 for gcc2. (I don't remember if the problem was already fixed or not)

jmairboeck:
Maybe it would be better to switch texinfo 6.7 to secondary arch and use that for lilypond?

jmairboeck:
I'll try building gcc2 when the lilypond build is finished.

jmairboeck:
the gcc2 recipe uses makeinfo4.13, so this should be fine:

https://github.com/haikuports/haikuports/blob/14c2cab5428145b93232cb69683a67bbe68a9f06/sys-devel/gcc/gcc-2.95.3_2017_07_20.recipe#L57

Begasus:
Builds fine for primary architectures on 32bit,  @pulkomandy any examples where it could fail creating manpages?

pulkomandy:
It would fail when building gcc2 itself, but it looks like it's makeinfo (not the same thing as texinfo?) so there shouldn't be a problem?

jmairboeck:
> It would fail when building gcc2 itself, but it looks like it's makeinfo (not the same thing as texinfo?) so there shouldn't be a problem?

`cmd:makeinfo` is part of `texinfo`, but there is `texinfo4` (providing `cmd:makeinfo_4.13`), which is apparently used by gcc2.

PR #7822, opened 2023-02-01
tcpdump: migrate to gcc11

davidkaroly:
Let's move to the updated toolchain so we don't have to maintain a downstream gcc2 patch.

We still need a patch for `configure.ac` in this version but once the next version of tcpdump gets released, most likely we'll be able to do a build from pristine source tree as the patch has been already upstreamed.


net-analyzer/tcpdump/tcpdump-4.99.3.recipe
```diff
@@ -79,6 +80,11 @@ INSTALL()
 	make install
 	strip $binDir/tcpdump
 	rm $binDir/tcpdump.$portVersion
+
+	if [ $effectiveTargetArchitecture = x86 ]; then
```
korli, line 84: Should be done with:
`runConfigure --omit-dirs binDir ./configure --bindir="$prefix/bin"`

davidkaroly: done

Begasus: Or with $commandBinDir as it has been done for a while now: https://github.com/Begasus/haikuports/wiki/Useful-variables-for-recipes#commandbindir-and-commandsuffix

PR #7823, opened 2023-02-04
nmap: migrate to gcc11

davidkaroly:
similar to #7822

### DIFF
--- a/net-analyzer/nmap/nmap-3.81.recipe
+++ b/net-analyzer/nmap/nmap-3.81.recipe
@@ -18,36 +18,42 @@ scan results (Ndiff), and a packet generation and response analysis tool \
 HOMEPAGE="https://nmap.org/"
 COPYRIGHT="1996–2005 The Nmap Project"
 LICENSE="GNU GPL v2"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://nmap.org/dist-old/nmap-$portVersion.tgz"
 CHECKSUM_SHA256="4ab4044b2c8a44466409f668919156b4d8429e0449c6c482ea2063730d727293"
 PATCHES="nmap-$portVersion.patchset"
 
-ARCHITECTURES="all"
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
-	nmap = $portVersion
+	nmap$secondaryArchSuffix = $portVersion
 	cmd:nmap = $portVersion
 	"
 REQUIRES="
-	haiku
-	lib:libpcap
-	lib:libpcre
-	lib:libssl
+	haiku$secondaryArchSuffix
+	lib:libpcap$secondaryArchSuffix
+	lib:libpcre$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
-	haiku_devel
-	devel:libpcap
-	devel:libpcre
-	devel:libssl
+	haiku${secondaryArchSuffix}_devel
+	devel:libpcap$secondaryArchSuffix
+	devel:libpcre$secondaryArchSuffix
+	devel:libssl$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
-	cmd:gcc
-	cmd:ld
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
 	cmd:libtoolize
 	cmd:make
 	"
+if [ $effectiveTargetArchitecture = x86 ]; then
+	REPLACES="
+		nmap
+		"
+fi
 
 BUILD()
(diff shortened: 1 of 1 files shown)


net-analyzer/nmap/nmap-3.81.recipe
```diff
@@ -56,8 +62,9 @@ BUILD()
 	libtoolize --force --copy --install
 	popd
 
-	runConfigure --omit-dirs "dataRootDir docDir" \
-		./configure  --without-nmapfe LDFLAGS=-lnetwork
+	runConfigure --omit-dirs "binDir dataRootDir docDir" \
+		./configure --bindir="$prefix/bin" \
```
Begasus, line 66: Same here for $commandBinDir: https://github.com/Begasus/haikuports/wiki/Useful-variables-for-recipes#commandbindir-and-commandsuffix

davidkaroly: even better there should be a similar helper for CMake as well (which needs some rework anyway as currently `cmakeDirArgs` are all absolute which does not work so well on some recipes)

Begasus: This should be dealth with at haikuporter I guess, @davidkaroly good to go?

davidkaroly: i think so

Begasus:
Build failure on both buildmaster :/
https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/9312/builds/80379.log

davidkaroly:
Strange that it failed on the x86_64 as well, it should have been only a rebuild for 64bit without any change.

pulkomandy:
The error is (for future reference, I think the buildmaster won't keep the log forever):

{{{
g++ -lnetwork -Lnbase -Lnsock/src/  -o nmap main.o nmap.o targets.o tcpip.o nmap_error.o utils.o idle_scan.o osscan.o output.o scan_engine.o timing.o charpool.o services.o protocols.o nmap_rpc.o portlist.o NmapOps.o TargetGroup.o Target.o FingerPrintResults.o service_scan.o NmapOutputTable.o MACLookup.o  -lnbase -lnsock -lpcre -lpcap -lssl -lcrypto -lm
/boot/system/develop/tools/bin/../lib/gcc/x86_64-unknown-haiku/13.2.0/../../../../x86_64-unknown-haiku/bin/ld: output.o:/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/x86_64-unknown-haiku/bits/gthr-default.h:250:(.data.rel.local+0x0): undefined reference to `__gthrw_pthread_cancel(_pthread_thread*)'
}}}

This function is supposed to be a weak alias to pthread_cancel defined in libgcc (buildtools/gcc/libgcc/gthr-posix.h).

That file is included in the libgcc build from gcc/libgcc/configure with a snippet generated from gcc/config/gthr.m4 which is invoked from libgcc/configure.ac. The parameter is target_thread_file which is initialized from the "Thread model:" value printed by gcc -v.

There are also some defines that decide wether the function is enabled or not in gthread-posix.h, mainly depending on SUPPORTS_WEAK and GTHREAD_USE_WEAK. Could it be that one of those changed values at some point between building libgcc and compiling nmap? 

Earlier in the log we see:

```
In file included from /boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/x86_64-unknown-haiku/bits/gthr.h:148,
                from /boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/ext/atomicity.h:35,
                from /boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/bits/cow_string.h:37,
                from /boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/bits/basic_string.h:51,
                from /boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/string:54,
                from output.cc:112:
/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/x86_64-unknown-haiku/bits/gthr-default.h: At global scope:
/boot/system/develop/tools/lib/gcc/x86_64-unknown-haiku/13.2.0/include/c++/x86_64-unknown-haiku/bits/gthr-default.h:112:1: warning: 'int __gthrw_pthread_cancel(pthread_t)' used but never defined
 112 | __gthrw(pthread_cancel)
```

Here we are in libstdc++ (notice the "c++" in the include path) and it is similar to ghtread_posix, but it uses different defines to control wether the __gthr_ weak alias are generated: __GXX_WEAK__ and _GLIBCXX_GTHREAD_USE_WEAK. Do these have the same value than the corresponding ones in libgcc?

PR #7835, opened 2023-02-07
texlive: add notification when post-installing

humdingerb:
As it was mentioned on the forums that running the post-install script can take a while. Let's not make the user wonder if installation has failed and show a notification.

I did **not** test this by actually building/installing texlive...

### DIFF
--- a/app-text/texlive/additional-files/texlive_update.sh
+++ b/app-text/texlive/additional-files/texlive_update.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
 
+notify --group "TexLive" --title "Installation in progress" --messageID texliveInstallID --timeout 10000 "Updating TeXLive filename database…"
 mktexlsr
+
+notify --group "TexLive" --title "Installation in progress" --messageID texliveInstallID --timeout 10000 "Updating TeXLive format files…"
 fmtutil-sys --all
+
 #mtxrun --generate
+notify --group "TexLive" --title "Installation in progress" --messageID texliveInstallID --timeout 10000 "Updating TeXLive font maps…"
 updmap-sys
 
+notify --group "TexLive" --title "Installation finished" --messageID texliveInstallID  "Took long enough...!"


korli:
"Updating TeXLive filename database..."  "Updating TeXLive format files..." "Updating TeXLive font maps..."

Begasus:
I don't think it adds up to the current recipe, it's not being rebuild (and shouldn't for my count), don't need to add extra revision builds for this, this is one downside one has to take into consideration?

humdingerb:
I amended showing several notifications, one for each executed command. Thanks korli.

The "texlive" package that's impacted by this is only 4.5 MiB. Updating this would provide all users that install the package from now on with these notifications. All users that already have it (and the gigabytes of sub-packages) won't be affected.

I think. You guys are the experts with this monster. :)
You decide if I should also bump the revision.

Begasus:
Still not sure if this will get rebuild too? https://github.com/haikuports/haikuports/blob/9c3f1bb1de7ccd5b51c4621b8809d0ca6c19fc1a/app-text/texlive/texlive-2022.recipe#L21

humdingerb:
Dunno... I just started to build texlive, but when it started to download some 3.4 GiB file I had to abort. :)
Maybe merge this as is and wait for a rev bump until some poor soul has to update the package in the future... :)

jmairboeck:
> Still not sure if this will get rebuild too?
> 
> https://github.com/haikuports/haikuports/blob/9c3f1bb1de7ccd5b51c4621b8809d0ca6c19fc1a/app-text/texlive/texlive-2022.recipe#L21

createSubpackageInfos.pl is not used during the build at all (maybe it should be removed from the recipe). It is just used to generate subpackageInfos.sh.

Ideally, it would be used during the build to recreate the subpackage infos dynamically, but haikuporter needs to know which packages are built before-hand, so this doesn't work (see https://github.com/haikuports/haikuports/blob/9c3f1bb1de7ccd5b51c4621b8809d0ca6c19fc1a/app-text/texlive/texlive-2022.recipe#L67-L68 how that would look like).

humdingerb:
Your porting voodoo is above my paygrade... :)
Feel free to merge this PR - it doesn't trigger a rebuild. Then at least the notification stuff is in, once you figure when/if to rebuild after some/no needed other changes.

Begasus:
Thanks!

PR #7836, opened 2023-02-08
Haikuwebkit: 1.9.3

ghost:
Merge this once the haikuports builders are updated to beta4 (that is beta4 hrev56578+69 !)


jmairboeck:
If no explicit version is specified, the version available at build time is used as the minimum implicitly, so just using a new enough builder should be sufficient.

PR #7841, opened 2023-02-11
wxgtk: update to 3.2.2

davidkaroly:
update to latest version
we need to re-add the dependency on xlibe otherwise gcc is complaining
successful build on 32&64-bit (incl. all demos and samples)

GL is still not enabled, I think there are not so many applications that would actually use it.


pulkomandy:
Kicad will probably need GL support? Not sure how theis rendering is done in the current versions

davidkaroly:
> Kicad will probably need GL support? Not sure how theis rendering is done in the current versions

I remember I had a successful wxgtk build with GL some time ago so I can try enabling it in this recipe. I had some build issues with one of the GL examples though.

davidkaroly:
Some issues with OpenGL... even though I'm able to build the package and samples with GL enabled, none of the samples work. I'd rather we merge it like this as I have neither the time nor the skills to investigate the GL stuff.

pulkomandy:
Yes, that's fine, we can look into it later

PR #7844, opened 2023-02-12
qr_code_generator, bump version

Begasus:


Begasus:
@diversys any reason to keep the static library for this? (previous version only provided the static one, this one only provides the shared libraries for the C and C++ version)

PR #7861, opened 2023-02-18
media-sound/ocp: Version bump

mywave82:
version 0.2.103


augiedoggie:
It would be better to do a `git mv` when updating these.  It makes reviewing much easier when you can see the changes between the old and new recipes.

I think you can `git rm` the old versions of this recipe.  It should be safe...

mywave82:
I will fix in the future

mywave82:
rename + update in the same commit

PR #7862, opened 2023-02-18
Doxygen: 1.9.6

nephele-gh:
Required to test changes to the user api docs. 1.9.5 has a bug where it will style scrollbars that can't be overiden in css.


Begasus:
Latest pachset got upstreamed earlier today, with the new changes to filesystem.hpp and a pending PR there patching shouldn't be needed anymore :)
PS, had this version build earlier today also, so things should be fine for this version 👍

PR #7865, opened 2023-02-19
gulrak_filesystem, new recipe

Begasus:


dev-cpp/gulrak-filesystem/gulrak_filesystem-1.5.12.recipe
```diff
@@ -0,0 +1,62 @@
...
+SOURCE_FILENAME="gulrak_filesystem-v$portVersion.tar.gz"
+SOURCE_DIR="filesystem-$portVersion"
+PATCHES="163.patch"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	gulrak_filesystem$secondaryArchSuffix = $portVersion
+	devel:filesystem$secondaryArchSuffix = $portVersion
```
Begasus, line 28: Maybe I should use gulrak_filesystem here too?
```diff
-	devel:filesystem$secondaryArchSuffix = $portVersion
+	devel:gulrak_filesystem$secondaryArchSuffix = $portVersion
```

jmairboeck: isn't a `lib:` or `devel:` dependency always the library name?

Begasus: In this case there is no library involved, it only provides the headers and cmake files for detection.

dev-cpp/gulrak-filesystem/gulrak_filesystem-1.5.12.recipe
```diff
@@ -0,0 +1,62 @@
...
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	"
+BUILD_PREREQUIRES="
+	cmd:cmake
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+PATCH()
```
korli, line 42: missing newline before PATCH()

Begasus: Will correct on a new release, thanks for noticing! (patch has been upstreamed aside from this (disabling -Werror)) :)

PR #7871, opened 2023-02-21
Add llvm15

jessicah:
* libunwind is now a runtime component, so doesn't have its own package any more.
* switched to the llvm-project tarball instead of downloading individual components.
* switched to $cmakeDirArgs for configuring, simplifying the recipe in many places.

### DIFF
--- a/sys-devel/llvm/llvm15-15.0.7.recipe
+++ b/sys-devel/llvm/llvm15-15.0.7.recipe
@@ -1,0 +1,739 @@
+SUMMARY="Modular and reuseable compiler and toolchain technologies"
+DESCRIPTION="LLVM is a collection of modular and reuseable compiler and and \
+toolchain technologies.The LLVM compiler system for C and C++ includes the \
+following:
+
+- Front-ends for C, C++, Objective-C, Fortran, etc. based on the GCC 4.2 \
+parsers. They support the ANSI-standard C and C++ languages to the same \
+degree that GCC supports them. Additionally, many GCC extensions are \
+supported.
+- A stable implementation of the LLVM instruction set, which serves as both \
+the online and offline code representation, together with assembly (ASCII) \
+and bytecode (binary) readers and writers, and a verifier.
+- A powerful pass-management system that automatically sequences passes \
+(including analysis, transformation, and code-generation passes) based on \
+their dependences, and pipelines them for efficiency.
+- A wide range of global scalar optimizations.
+- A link-time interprocedural optimization framework with a rich set of \
+analyses and transformations, including sophisticated whole-program pointer \
+analysis, call graph construction, and support for profile-guided optimizations.
+- An easily retargettable code generator, which currently supports X86, \
+X86-64, PowerPC, PowerPC-64, ARM, Thumb, SPARC, Alpha, CellSPU, MIPS, MSP430, \
+SystemZ, and XCore.
+- A Just-In-Time (JIT) code generation system, which currently supports X86, \
+X86-64, ARM, AArch64, Mips, SystemZ, PowerPC, and PowerPC-64.
+- Support for generating DWARF debugging information.
+- A C back-end useful for testing and for generating native code on targets \
+other than the ones listed above.
+- A profiling system similar to gprof.
+- A test framework with a number of benchmark codes and applications.
+- APIs and debugging tools to simplify rapid development of LLVM components."
+HOMEPAGE="https://www.llvm.org/"
+COPYRIGHT="2003-2019 University of Illinois at Urbana-Champaign"
+LICENSE="Apache v2 with LLVM Exception"
+REVISION="1"
+SOURCE_URI="https://github.com/llvm/llvm-project/releases/download/llvmorg-$portVersion/llvm-project-$portVersion.src.tar.xz"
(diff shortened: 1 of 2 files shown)


sys-devel/llvm/llvm15-15.0.7.recipe
```diff
@@ -0,0 +1,739 @@
...
+	lib:libclang_cpp$secondaryArchSuffix
+	lib:libLLVM_15$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
+	"
+CONFLICTS="
+	llvm$secondaryArchSuffix
+	llvm6$secondaryArchSuffix
+	llvm7$secondaryArchSuffix
+	llvm8$secondaryArchSuffix
+	llvm9$secondaryArchSuffix
```
augiedoggie, line 320: Do these CONFLICTS stanzas need llvm12 entries?

sys-devel/llvm/llvm15-15.0.7.recipe
```diff
@@ -0,0 +1,739 @@
...
+- A Just-In-Time (JIT) code generation system, which currently supports X86, \
+X86-64, ARM, AArch64, Mips, SystemZ, PowerPC, and PowerPC-64.
+- Support for generating DWARF debugging information.
+- A C back-end useful for testing and for generating native code on targets \
+other than the ones listed above.
+- A profiling system similar to gprof.
+- A test framework with a number of benchmark codes and applications.
+- APIs and debugging tools to simplify rapid development of LLVM components."
+HOMEPAGE="https://www.llvm.org/"
+COPYRIGHT="2003-2019 University of Illinois at Urbana-Champaign"
```
augiedoggie, line 32: Might need an update to the copyright year

augiedoggie:
It seems all of the LLVM recipes use a mix of `cmd:foo` and `cmd:foo = $portVersion`.  It might be nice to clean that up at some point.  I never got around to it when I was messing with the LLVM12 recipe.

jessicah:
@augiedoggie ah thanks for that. I had assumed the $portVersion stuff was missing on purpose, can add.

sys-devel/llvm/llvm15-15.0.7.recipe
```diff
@@ -0,0 +1,739 @@
...
+	# such as Mesa
+	cmake -DCMAKE_BUILD_TYPE=Release $cmakeDirArgs \
+		-DCMAKE_SKIP_RPATH=NO $cmakeFlags \
+		-DLLVM_ENABLE_RTTI=ON -DLLVM_LINK_LLVM_DYLIB=YES \
+		-DLIBUNWIND_ENABLE_STATIC=OFF \
+		-DLLVM_ENABLE_PROJECTS="clang;clang-tools-extra;lld;compiler-rt" \
+		-DLLVM_ENABLE_RUNTIMES="libunwind" \
+		-G Ninja ..
+
+	ninja
```
davidkaroly, line 555: Maybe `$jobArgs` can be added here?

jessicah: ninja already uses all available, unless want to explicitly limit it?

sys-devel/llvm/llvm15-15.0.7.recipe
jessicah: llvm doesn't seem to update it, so I'll leave as-is.

sys-devel/llvm/llvm15-15.0.7.recipe
pulkomandy: yes, it's better to let ninja guess for itself. I think it uses a measurement of the system load rather than just one task per CPU, so it auto-adjusts nicely (for example if there is a lot of IO in one job, it can spawn more jobs to keep the CPUs busy)

diversys:
Probably need to add patches from https://github.com/X547/llvm-project/tree/llvm-haiku

sys-devel/llvm/llvm15-15.0.7.recipe
korli: $jobArgs shouldn't be ignored, otherwise the one can't locally use the CPU for anything else.

sys-devel/llvm/llvm15-15.0.7.recipe
```diff
@@ -0,0 +1,739 @@
...
+	packageEntries python3 \
+		$binDir/lit \
+		$prefix/lib/python*
+}
+
+TEST()
+{
+	export LIT_ARGS="-v"
+	if [ "$targetArchitecture" = x86_gcc2 ];then
+		export LIBRARY_PATH=$LIBRARY_PATH:$sourceDir/build/lib/x86
```
jmairboeck, line 734: this could use `$relativeLibDir` at the end, replacing the whole if, or `$secondaryArchSubDir`

khallebal: Sorry for hijacking this PR, but any chance that one of you guys will add support for and enable the sanitizers either in llvm-15 or gcc-12?

Begasus:
Superseded and imported in: https://github.com/haikuports/haikuports/pull/9141
Thanks here too!

PR #7879, opened 2023-02-25
media-fonts/noto_emoji: bump recipe

n0toose:
Related to https://dev.haiku-os.org/ticket/17328.

This was not tested on a running system, but I tested the URLs and ran the scripts in the root directory of the repository, validated the URLs and paths and ran `sha256sum` against the downloaded archive. I haven't done this in a bit, I apologize in advance if there's any silly errors :)

cc: @nephele-gh  @pulkomandy


pulkomandy:
The source repository is here: https://github.com/googlefonts/noto-emoji

But the monochrome version (the one which currently works in Haiku) is currently missing from it: https://github.com/googlefonts/noto-emoji/issues/390

I would suggest we don't update until this is fixed, since the color one currently doesn't work in Haiku.

pulkomandy:
> This was not tested on a running system

Stop doing this, please. Test your changes. You are wasting some one else's time when you don't do it.

pulkomandy:
We don't really need them, no. But in any case the recipe needs to be adjusted to still have the non-color version somehow. And it would be nice to know the sources are available somewhere in case we want to make some changes :)

n0toose:
I had this discussion with PulkoMandy, this route makes the most sense. I'll get on that.

n0toose:
Question: I'm dealing with a problem here. This recipe exports `noto_emoji_color` and `noto_emoji` right now, but this assumes that the color version is more "special" than the monochrome version, which is essentially deprecated. I'd assume that it's supposed to be the other way around: `noto_emoji` should be the version with colors, whereas something like `noto_emoji_monochrome` or some similar name should be in charge of providing the deprecated monochrome font, until it will eventually phase out. How do I deal with this? Is it possible to deal with this, without affecting the user experience or introducing breaking changes?

pulkomandy:
We don't get to name these fonts, Google decides how they're named. They named the monochrome one noto_emoji and later one made a color one named noto_emoji_color. Let's just keep their names, unless they renamed things since then?

n0toose:
> unless they renamed things since then?

Was not aware of that, apologies.

n0toose:
Succeeded by https://github.com/haikuports/haikuports/pull/7947.

PR #7895, opened 2023-03-02
libfilezilla: update to 0.42.2

davidkaroly:
* <s>disable static library</s>
* update patchset
* update command line for running tests

successful build on 32-bit & 64-bit nightly

### DIFF
--- a/dev-libs/libfilezilla/libfilezilla-0.42.2.recipe
+++ b/dev-libs/libfilezilla/libfilezilla-0.42.2.recipe
@@ -8,17 +8,17 @@ extremely efficient
 which simplifies dealing with timestamps originating from different sources
 - Simple process handling for spawning child processes with redirected I/O"
 HOMEPAGE="https://lib.filezilla-project.org/"
-COPYRIGHT="2015-2022 Tim Kosse"
+COPYRIGHT="2015-2023 Tim Kosse"
 LICENSE="GNU GPL v2"
-REVISION="2"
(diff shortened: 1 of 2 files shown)


dev-libs/libfilezilla/libfilezilla-0.41.1.recipe
```diff
@@ -8,17 +8,17 @@ extremely efficient
...
 SOURCE_URI="https://download.filezilla-project.org/libfilezilla/libfilezilla-$portVersion.tar.bz2"
-CHECKSUM_SHA256="c0843c62d1512f6aec57c6a27806b491ec8cf40b8c2ed259faba9f6907820c72"
+CHECKSUM_SHA256="a4d55f82224057c8940a01565483b056bab321df140f9e64f20018f986200574"
 PATCHES="libfilezilla-$portVersion.patchset"
 
 ARCHITECTURES="all !x86_gcc2"
 SECONDARY_ARCHITECTURES="x86"
 
-libVersion="33.0.0"
+libVersion="35.0.0"
```
korli, line 21: This means dependent packages should be rebuilt, correct?

davidkaroly: Yes i think so. There will be a similar change for filezilla-3.63.2.

korli: I mean, in between, filezilla won't be installable.

PR #7901, opened 2023-03-04
wesnoth: Use Boost 1.70

robxnano:


games-strategy/wesnoth/wesnoth-1.12.6.recipe
```diff
@@ -61,13 +62,19 @@ REQUIRES="
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
-	devel:libboost_filesystem$secondaryArchSuffix
+	devel:libboost_filesystem$secondaryArchSuffix >= 1.70.0
+	devel:libboost_iostreams$secondaryArchSuffix >= 1.70.0
+	devel:libboost_locale$secondaryArchSuffix >= 1.70.0
+	devel:libboost_program_options$secondaryArchSuffix >= 1.70.0
+	devel:libboost_regex$secondaryArchSuffix >= 1.70.0
+	devel:libboost_system$secondaryArchSuffix >= 1.70.0
```
waddlesplash, line 70: Does a version need to be specified (and in other recipes, too) or can we not just let HaikuPorter pick?

robxnano: The problem is that when no version is specified, HaikuPorter (always?) chooses the earliest version. Since there are 6 old versions of Boost (1.57, 1.63, 1.64, 1.65 ,1.66 and 1.69), some packages will break if any of them are removed from the repositories. That's why I'm trying to move everything that uses 1.66 or below to the latest version.

waddlesplash: I suppose that's fair enough.

korli:
build failed https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/8142/builds/78833.log

PR #7902, opened 2023-03-04
mupen64plus: Fix build and use Boost 1.70

robxnano:


Begasus:
Build failure on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/3802/builds/79773.log

robxnano:
Sorry, I don't know why it's not working. Since the problem seems to be with LTO, maybe it would be okay to just disable it for the x86 build?

diversys:
Either that or you could try to disable ASLR with ```export DISABLE_ASLR=1```.

robxnano:
> Either that or you could try to disable ASLR with `export DISABLE_ASLR=1`.

Thanks, that fixed it.

PR #7904, opened 2023-03-04
mlpack: use Boost 1.70

robxnano:


Begasus:
Build on 32bit fails also here with "out of memory" ...

robxnano:
The mlpack build failing due to lack of memory seems to be a [common](https://github.com/mlpack/mlpack/issues/2800) [issue](https://github.com/mlpack/mlpack/issues/2049). Presumably that's why the x86 recipe was marked as untested.

PR #7905, opened 2023-03-04
mudlet: Use Boost 1.70

robxnano:
This builds and runs but the main screen complains about a load of missing lua modules. Not sure what needs to be done to fix it, but the old version had the same errors.


Begasus:
> This builds and runs but the main screen complains about a load of missing lua modules. Not sure what needs to be done to fix it, but the old version had the same errors.

`make install` takes care of that but isn't used in the recipe atm. :) (ps haven't checked it running)

PR #7907, opened 2023-03-04
innoextract: Update to 1.9, Use Boost 1.70

robxnano:
Using Boost 1.69 as 1.70 failed to work. EDIT: Found an alternative solution to use 1.70.

### DIFF
--- a/app-arch/innoextract/innoextract-1.9.recipe
+++ b/app-arch/innoextract/innoextract-1.9.recipe
@@ -2,16 +2,16 @@ SUMMARY="A tool to unpack installers created by Inno Setup"
 DESCRIPTION="Inno Setup is a tool to create installers for Microsoft Windows \
 applications. innoextract allows to extract such installers under non-Windows \
 systems without running the actual installer using wine. innoextract currently\
- supports installers created by Inno Setup 1.2.10 to 5.6.0.
+ supports installers created by Inno Setup 1.2.10 to 6.0.5.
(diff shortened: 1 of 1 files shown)


app-arch/innoextract/innoextract-1.8.recipe
```diff
@@ -55,6 +55,8 @@ BUILD()
 {
 	export CXXFLAGS="-D_BSD_SOURCE"
 	cmake . $cmakeDirArgs \
+		-DBoost_NO_BOOST_CMAKE=ON \
+		-DCMAKE_BUILD_TYPE=Release \
 		-DCMAKE_INSTALL_BINDIR=$commandBinDir
 	make $jobArgs
 }
```
Begasus, line 62: Nice catch! Suggestion:
```
	export CXXFLAGS="-D_BSD_SOURCE"
	cmake -Bbuild -S. $cmakeDirArgs \
		-DCMAKE_BUILD_TYPE=Release \
		-DCMAKE_INSTALL_BINDIR=$commandBinDir \
		-DBoost_NO_BOOST_CMAKE=ON \
	make -C build $jobArgs
```
First system variables, then user variables (I know there is no straight guideline for this (yet). :)
In INSTALL() you could use `make -C build install` then.

robxnano: Thanks! I've made the suggested changes, and also updated to version 1.9. It adds support for building on newer Boost versions which could be useful in the future.

PR #7913, opened 2023-03-04
lilypond: fix secondaryArch (x86) build (2nd try)

jmairboeck:
I can't really test this unfortunately ...


jmairboeck:
@korli the revision bump wasn't really necessary because the x86_64 had already worked before and the x86 build not. The only difference will be that the empty `documentation/packages` directory will not be part of the main package now. That wouldn't have needed a rev-bump ;-)

korli:
> @korli the revision bump wasn't really necessary because the x86_64 had already worked before and the x86 build not. The only difference will be that the empty `documentation/packages` directory will not be part of the main package now. That wouldn't have needed a rev-bump ;-)

It changed enough, that it should be tested on x86_64.

PR #7928, opened 2023-03-08
Update Scintilla and Koder

KapiX:


x11-libs/lexilla/lexilla-5.2.3.recipe
```diff
@@ -0,0 +1,104 @@
...
+
+BUILD()
+{
+	local SCINTILLA_INCLUDE=$sourceDir/../scintilla/include
+	mkdir -p $SCINTILLA_INCLUDE
+	cp -R $sourceDir3/include/* $SCINTILLA_INCLUDE
+
+	#local packageLinksDir=$(dirname $portPackageLinksDir)
+	#local includeDir="$packageLinksDir/$portName-$portFullVersion/devel~libscintilla$secondaryArchSuffix/$relativeIncludeDir/scintilla"
+	#ls -la $packageLinksDir/$portName-$portFullVersion/
```
KapiX, line 75: I'd prefer to use packageLinks instead of downloading scintilla source, but for whatever reason devel~libscintilla points to `'?'`. Any tips?

korli: No, but I suppose `$portPackageLinksDir/devel~libscintilla$secondaryArchSuffix/` should work.

KapiX: I couldn't make it work. It might be a bug in haikuporter.

PR #7933, opened 2023-03-10
smooth: add patch to fix crash in CPUID code

enzo1982:
Add a patch to the smooth library to fix a crash when querying CPU information. Unfortunately, this issue was found only now and fre:ac 1.1.7 is unusable on Haiku without this.

This patch is already in the upstream smooth repository, so it will be needed only until the next release.

I have opened a PR at the libcpuid repository as well: anrieff/libcpuid#186 There is no HaikuPorts recipe for libcpuid yet, so smooth uses a bundled version of the library.

### DIFF
--- a/dev-libs/smooth/patches/smooth-0.9.10.patchset
+++ b/dev-libs/smooth/patches/smooth-0.9.10.patchset
@@ -1,0 +1,23 @@
+From 3d39dc4f9a3d3603ea8110e56f5fb46d518595bd Mon Sep 17 00:00:00 2001
+From: Robert Kausch <robert.kausch@freac.org>
+Date: Fri, 10 Mar 2023 01:20:01 +0000
+Subject: Fix crash querying CPU info on systems without affinity setting
+ support in libcpuid.
+
+
+diff --git a/libraries/libcpuid/cpuid_main.c b/libraries/libcpuid/cpuid_main.c
+index 39aaa57..3893f2c 100644
+--- a/libraries/libcpuid/cpuid_main.c
++++ b/libraries/libcpuid/cpuid_main.c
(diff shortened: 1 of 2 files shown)


Begasus:
I felt a chalange here so I started to work on some draft recipe for libcpuid, given that you mentioned a crash I pulled in the latest commit with srcGitRev, launching cpuid_tool-x86 still crashes here, so thought I mentioned here.
Output from Terminal:
```
~> cpuid_tool-x86 
Your system is not supported by libcpuid -- don't know how to set the CPU affinity.
Writing raw CPUID dump to 'raw.txt'
Kill Thread
```

The raw.txt is empty.

enzo1982:
> launching cpuid_tool-x86 still crashes here, so thought I mentioned here.

You need to apply my patch from this PR or anrieff/libcpuid#186. The `cpuid_serialize_raw_data_internal()` function in libcpuid assumes that at least one record exists and thus crashes without this patch.

Begasus:
Thought that was merged, my bad :)

Begasus:
PR for libcpuid created at: https://github.com/haikuports/haikuports/pull/7941

Could you check if this is still ok with smooth? ;)

enzo1982:
Thanks @Begasus!

Yes, smooth still works fine with that after adding the dependency and calling `make` with `config="systemlibcpuid"`.

I could update the smooth recipe to use the libcpuid package after your PR is merged. Or I would just do it with the next regular update.

I also noticed that there is a package for libwebp here. So smooth could use that one too.

Begasus:
Updating smooth should be fine on the next release, maybe by then your patch is merged and a new release for libcpuid too :)

PR #7936, opened 2023-03-10
libzim, rebuild with updated meson (part 2)

Begasus:


Begasus:
Previous rebuild was still done with revision 1 for meson.

Issue #7531, opened 2022-12-07
Python 3.10: problems with the modules installation paths


OscarL:
Opening an issue for this so the problem (and its cause) doesn't gets buried...

As mentioned on #7519, Python 3.10 has some issues with the paths it's using for installing modules:

"system" modules can't be installed, as the "site-package" directory it tries to use (under `/system/lib/pyton3.10/`) is read-only.  So it reverts to user installations.

But those user installations are using paths as `~/.local/bin/`, `~/.local/lib/`, instead of the expected `~/config/non-packaged/bin/` and `~/config/non-packaged/lib/`. This makes it impossible to import "successfully" installed modules, and also prevents uninstalling them with pip.

Python 3.7 to 3.9 seem to have the proper paths[1] for installations, and we should carry that to 3.10 and up.

----

[1] Small caveat.. 3.7 to 3.9 seem to have some left-over code looking at `~/.local/bin/`. This is easy to see doing the following:

```bash
> pip3.10 install --user pyserial
> pip3.9 install --user pyserial
> pip3.9 uninstall pyserial
```

The first line "succeeds", but has the problems mentioned above.

The last line is the interesting one, because now pip3.9 offers to delete the following

```
/boot/home/.local/bin/pyserial-miniterm
/boot/home/.local/bin/pyserial-ports
/boot/home/config/non-packaged/bin/pyserial-miniterm
/boot/home/config/non-packaged/bin/pyserial-ports
/boot/home/config/non-packaged/lib/python3.9/site-packages/pyserial-3.5.dist-info/*
/boot/home/config/non-packaged/lib/python3.9/site-packages/serial/*
```

when the first two were installed by pip3.10 and NOT by pip3.9 (3.7 and 3.8 behaves the same here).

It might be worth to investigate this issue too.

korli:
> I accidentally deleted a previous comment where I suggested that we should also patch sysconfig.py's `_getuserbase()` also for Python versions 3.7 to 3.9 in the same way it is done for site.py's `_getuserbase()`.
> 
> Thoughts?

Yeah, both `_getuserbase()` should be synched. Good idea.

Begasus:
Food for thoughts, when using pip to install pyserial, the binaries are installed in `bin`, I'm guessing the latest python installed version will have the binaries available and will overwrite the existing ones?
Won't that lead to conflicts?

korli:
@Begasus does this work on other platforms?

Begasus:
[...]

PR #7942, opened 2023-03-10
python3.7: update to 3.7.16, sync both _getuserbase() copies.

OscarL:

Similar PRs for the rest of the versions will follow.


korli:
You need to update to 3.7.16 or bump the revision.

PR #7944, opened 2023-03-11
python3.8: update to 3.8.16, sync both _getuserbase() copies.

OscarL:

Also: renamed the package to be consistent with the rest of the Python versions.


dev-lang/python/python3.8-3.8.16.recipe
```diff
@@ -32,7 +32,7 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
 fi
 
 PROVIDES="
-	python38$secondaryArchSuffix = $portVersion compat >= 3.8
+	python3.8$secondaryArchSuffix = $portVersion compat >= 3.8
```
korli, line 35: REPLACES="python38$secondaryArchSuffix" is missing

PR #7946, opened 2023-03-11
python3.10: update to 3.10.10, sync both _getuserbase() copies.

OscarL:

Syncronizing these functions means  that now `pip3.10` can be used for `--user` installations, as it will now install things under `~/config/non-packaged/`, and not under `~/.local/` as was the case before.


Begasus:
Checked on 32bit and looks fine!

PR #7947, opened 2023-03-11
Split noto emoji and noto color emoji

nephele-gh:
Also has the newer version of noto emoji

### DIFF
--- a/media-fonts/noto_emoji/noto_emoji-20230311.recipe
+++ b/media-fonts/noto_emoji/noto_emoji-20230311.recipe
@@ -1,0 +1,28 @@
+SUMMARY="The Noto emoji font"
+DESCRIPTION="Fonts containing Unicode emoji characters"
+HOMEPAGE="https://fonts.google.com/download?family=Noto%20Emoji"
+COPYRIGHT="2015-2023 Google Inc."
+LICENSE="SIL Open Font License v1.1"
+REVISION="1"
+SOURCE_URI="https://fonts.google.com/download?family=Noto%20Emoji"
+SOURCE_FILENAME="noto-emoji.zip"
+CHECKSUM_SHA256="d1c824c449a22fcc14360fe1d23cd0840e83a6621c85a2cf32380663d36e721e"
+SOURCE_DIR=""
+
+ARCHITECTURES="any"
+DISABLE_SOURCE_PACKAGE="yes"
+
+PROVIDES="
+	noto_emoji = $portVersion
+	"
+
+
+INSTALL()
+{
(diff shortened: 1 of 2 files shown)


korli:
Reason is missing.

korli:
> The reason for updating the font?

The reason for splitting

korli:
> for what purpose? I'm not sure what you want me to clarify in the commit message?

If you don't have a reason for splitting, then this PR is useless...

korli:
> It updates noto emoji and noto color emoji. How is that useless?
> 
> I see no reason to keep these two unrelated fonts in the same recipe as they no longer share an upstream source.

So I should update your commit message, because you don't want to?

pulkomandy:
The comment from Korli has not been addressed, the commit message should explain what happened so we can keep this information later on.

If you don't know what to write, here is a suggestion:

---

Put Noto Emoji and Noto Color Emoji into different packages.

They used to be developped and shipped together by Google, but they are now two independant projects.
We can ship them as two different packages, which allows to ship only the monochrome version in Haiku,
since color emojis are not supported by app_server yet anyways.

n0toose:
Commit messages are more permanent and a more reliable source of storing information than GitHub PRs, you can't `git blame` that and Haiku could hypothetically agree to moving to a self-hosted forge.

n0toose:
Additional context: Most archives of repositories on the Internet (incl. GitHub's own initiative, most probably) do not store information in bug trackers or in pull requests, just a `git clone` copy of it.

pulkomandy:
It's not just about "most archives", it's about making the info easily accessible (in git log, git blame etc) without referring to an external tool, and we have to assume that Github won't be an eternal solution for us and we will eventually move elsewhere (because no tool lasts forever). Likely only the git history will be properly migrated then.

n0toose:
> and we have to assume that Github won't be an eternal solution for us and we will eventually move elsewhere (because no tool lasts forever).

Yep, I tried to bring that up myself too :)

> likely only the git history will be properly migrated then.

Many (official / unofficial) tools support transitioning issues and pull requests these days support transitioning Issues and PRs and projects like Forgejo are working on standardized-ish formats ("Friendly Forge Format") so as to prevent these situations in the future - just letting you know in case your involvement lasts longer than mine - but it's always best when you can be 100% sure about these things! :)

pulkomandy:
We don't need advertisement for your other projects here… I know there are options for this

n0toose:
> We don't need advertisement for your other projects here

Was just trying to share something I'm personally excited about and I'm only marginally affiliated / not representing it in any capacity, sorry for the off-topic traveling salesman attitude.

PR #7949, opened 2023-03-12
libass: update to 0.17.1, enable assembly and cleanup

TheOneric:
- cleanup dependencies
- enable x86 assembly usage in builds
- disable static library build *(apparently required for all packages now?)*
- update to 0.17.1

### DIFF
--- a/media-libs/libass/libass-0.16.0.recipe
+++ b/media-libs/libass/libass-0.16.0.recipe
@@ -76,7 +76,7 @@ defineDebugInfoPackage libass$secondaryArchSuffix \
 BUILD()
 {
 	autoreconf -fi
-	runConfigure ./configure --with-pic
+	runConfigure ./configure --with-pic --disable-static
 	make $jobArgs
 }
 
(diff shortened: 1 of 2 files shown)


waddlesplash:
Welcome!

> enable x86 assembly usage in builds

libass has hand-written x86 assembly?! o_O

> disable static library build (apparently required for all packages now?)

It isn't allowed anymore to ship packages with both static and shared libraries of the same name, no. Only one or the other (preferably the shared) should be distributed.

TheOneric:
> libass has hand-written x86 assembly?! o_O

yup, since 0.11.0 (2014); rasterizing and blurs can be expensive. configure also emits a  `Install nasm for a significantly faster libass build` warning if building for x86 without NASM.
The unreleased next version will also have aarch64 assembly, albeit this doesn't use NASM (since NASM is x86-only) but GNU’s or llvm’s assembler.

TheOneric:
> while at it, could you remove transitive build dependencies

Done.

Begasus:
LGTM, if @korli is fine this should be good to merge. Thanks!

korli:
> 
> Done.

I only meant build requirements.

TheOneric:
> I only meant build requirements.

Restored transitive dependencies *(and libintl comment addition)* in `REQUIRES=`. Does this look good?

PR #7960, opened 2023-03-13
ddnet: fix recipe

n0toose:
recipe unexpectedly broke because of ~git suffix, tested on x86_64

See: https://github.com/haikuports/haikuports/pull/7951
Build logs of build mysteriously failing: https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/8219/builds/78935.log


n0toose:
keep in mind, this omits `~git` from the .rdef file

PR #7961, opened 2023-03-13
pycryptodome/pycryptodomex: update to 3.17

OscarL:
- Dropped patchset, fix was made upstream.
- Also build for Python 3.10.
- Mention package differences in their descriptions.

----

I think it might be possible to combine these two in a single recipe, but:

- I haven't been able to do so (still too new to recipes),
- Not sure if its entirely desirable/needed.

---

Smoke-Tested on beta4 64 bits (packages installed system wide, can import them in both Python versions, and call `help(Crypto)` and `help(Cryptodome)` at least).


dev-python/pycryptodome/pycryptodomex-3.17.0.recipe
```diff
@@ -4,19 +4,20 @@ primitives.
...
-COPYRIGHT="2014-2022 Legrandin"
+COPYRIGHT="2014-2023 Legrandin"
 LICENSE="BSD (2-clause)
 	Public Domain
 	Unlicense
 	"
 REVISION="1"
-SOURCE_URI="https://pypi.python.org/packages/source/p/pycryptodome/pycryptodome-$portVersion.tar.gz"
-CHECKSUM_SHA256="0e45d2d852a66ecfb904f090c3f87dc0dfb89a499570abad8590f10d9cffb350"
+SOURCE_URI="https://github.com/Legrandin/pycryptodome/archive/refs/tags/v$portVersion.tar.gz"
```
korli, line 17: Is this the same as https://pypi.org/packages/source/p/pycryptodomex/pycryptodomex-$portVersion.zip ?

korli: If things aren't stable, then don't bother. The next update will probably break anyway.

PR #7970, opened 2023-03-17
x265: fix build on 32 bit & version string

eblanca:
This PR fixes build in 32 bit haiku, where asm is now disabled for hdr libraries - on 64 bit asm is enabled for all libraries.
Further, fixed the missing version number for those hdr libs.

### DIFF
--- a/media-libs/x265/x265-3.5.recipe
+++ b/media-libs/x265/x265-3.5.recipe
@@ -130,7 +130,7 @@ BUILD()
 		-DLINKED_12BIT=ON \
 		-DCMAKE_INSTALL_PREFIX:PATH=$prefix	\
 		-DLIB_INSTALL_DIR:PATH=$relativeLibDir	\
-		-DBIN_INSTALL_DIR:PATH=$relativeBinDir	\
+		-DBIN_INSTALL_DIR:PATH=bin	\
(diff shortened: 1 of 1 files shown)


media-libs/x265/x265-3.5.recipe
```diff
@@ -38,7 +38,7 @@ REQUIRES_devel="
 
 PROVIDES_bin="
 	x265${secondaryArchSuffix}_bin = $portVersion
-	cmd:x265 = $portVersion
+	cmd:x265$secondaryArchSuffix = $portVersion
```
korli, line 41: should be installed $prefix/bin

eblanca: Done.

korli:
```suggestion
	cmd:x265 = $portVersion
```

eblanca: Wouldn't this brign to [this](https://discuss.haiku-os.org/t/policy-error-from-haikuporter/13230) ?

korli: no, the bin is installed in $prefix/bin

PR #7982, opened 2023-03-18
poco: new recipe

TheZeldakatze:
The recipe as of now doesn't get packaged probably, it's a WIP. The library itself compiles, however haikuporter errors with dependencies, which are in the package itself, I'm not sure how you solve that.
Also I've not yet tested x86, I'll do that shortly.

An example for the dependency errors would be
`Warning: POLICY ERROR: "/packaging/poco/bin/arc" needs library "libPocoFoundation.so.94", but the package doesn't seem to declare that as a requirement`


Begasus:
Quite some changes needed (also for 32bit), and the buildsystem they use don't use $cmakeDirArgs, so a bit tricky, but I managed to get a full build with the parts in the right places without errors, you can check and use the one I posted at: https://github.com/Begasus/haikuports/blob/981e076ab87182d6b08707bb3d33b14fdb856d55/dev-libs/poco/poco-1.12.4.recipe

Begasus:
Thanks on the work so far, there is no hurry, just give a yell once ready ;)

TheZeldakatze:
The reason the tests failed for foundation were that it could not find the `TestLibrary.so`, which I for now fixed by adding the library to the LIBRARY_PATH, the test `testCreate`, and the test  `testThreadStackSize` failed aswell, though I remeber reading in the BeBook that it is impossible to set the stack size, is that true? (It throws a `System exception: cannot set thread stack size`). TestCreate also throws a `System exception: Cannot create shared memory object: /hi`. Util-testrunner only fails on `testProperties` because system.env.HOME is not set, running it manually, it succedes. Net still fails a lot and both MongoDB and Redis fail, because there is no way to start a instance of either of those databases.

TheZeldakatze:
Had a look at net again: I compiled the library with `-DCMAKE_CXX_FLAGS_RELWITHDEBINFO="-g3"` so I could step through it with the debugger. testOptions fails at `StreamSocket::getLinger()` on `SocketTest.cpp:412`. Apparently it fails on getsockopt() (returning -1) and I personally don't know how to fix it. `testBroadcast` relies on setsockopt() and I suppose that that's what's ultimatly stops the test, as it is stuck on  `receiveBytes()`. But the basic Net tests work.

PR #8004, opened 2023-03-20
py: fix versioning.

OscarL:
Removed the `cmd:git` dependency, as it is `setuptools_scm` the one that actually requires that one.


Begasus:
Will be fixed in the other PR, thanks!

PR #8020, opened 2023-03-20
Disable boost versions 1.65.1 and 1.66.0, no longer installable, and …

Begasus:
…buildmaster shouldn't try to use them.


Begasus:
These are not installable anymore, and I had no luck rebuilding them, disabled for now (maybe all in all remove?)

PR #8035, opened 2023-03-21
pycares: drop support for 3.7, add it for 3.9 and 3.10.

OscarL:
Only `net-im/poezio` depends on this in HaikuPorts. Following PRs will update that too.

(tested on beta4 64 bits).


korli:
cffi_python39 missing on x86

PR #8036, opened 2023-03-21
slixmpp: update to 1.8.2, switch to Python 3.9+

OscarL:
`net-im/poezio` is the sole user of this in-tree. This version is needed to update poezio to 0.14 (PR for that will follow).

### DIFF
--- a/dev-python/slixmpp/slixmpp-1.8.2.recipe
+++ b/dev-python/slixmpp/slixmpp-1.8.2.recipe
@@ -11,7 +11,7 @@ LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/poezio/slixmpp/archive/slix-$portVersion.tar.gz"
 SOURCE_DIR="slixmpp-slix-$portVersion"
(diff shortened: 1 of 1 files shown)


korli:
aiohttp_python310 missing

korli:
> Can we do locally with haikuporter the same dependency check that the buildmaster does? (would like to avoid this in the future).

the buildmaster also checks the requires, which shouldn't be needed, per se. So it's easily missed when building locally. 
Don't be mistaken, this is totally OK, to have such failures :)

korli:
Well if they don't need a C/C++ compiler or a shared library, they should stay any. On x86 you can make python packages also provide the non-x86 provides, so that any and not-any packages work together.

korli:
aiohttp_x86_python39 can provide aiohttp_python39
example: https://github.com/haikuports/haikuports/blob/6c4ed285d5108836e9fe213f939b8bbf708bc1cc/dev-python/cython/cython-0.29.32.recipe#L42

PR #8040, opened 2023-03-22
cegui, missing CMAKE_BUILD_TYPE

Begasus:


Begasus:
@korli didn't revbump here (saw you did just after it), because I didn't see a cause to let it rebuild, should I revbump others aswell if I run into them?

korli:
@Begasus I would say to make sure it actually builds. BTW the cmake macro could add a default CMAKE_BUILD_TYPE, this would spare us this kind of commits.

Begasus:
@korli OK, I test the builds before I push (on 32bit), so maybe yes, should've revbumped :)
For the cmake macro that would indeed solve most of the issues that will pop up, but that's out of my leage ;)

PR #8046, opened 2023-03-22
multidict: switch to x86, drop support for Python 3.7, add it for 3.10.

OscarL:
Build was broken on x86_gcc2 (requires compiler with C99 support).


dev-python/multidict/multidict-4.5.2.recipe
```diff
@@ -36,6 +37,11 @@ REQUIRES_$pythonPackage=\"\
 	haiku\n\
```
korli, line 37:
```suggestion
	haiku$secondaryArchSuffix\n\
```

PR #8092, opened 2023-03-24
imutils: retire recipe.

OscarL:
Not used by anything in-tree, attepts to update it to 0.5.4 ended
with installable, but non-working modules.

Upstream repo possibly unmaintained.


korli:
imutils can also be dropped.

korli:
> > imutils can also be dropped.
> 
> Done.

we might as well disable it instead (sorry). maybe @extrowerk wants to fix it.

extrowerk:
LAtely i don't have time for HP, so please do what you see to fit in this case. Thanks for the understanding.

korli:
OK. The recipe can be revived anytime from the history.

PR #8112, opened 2023-03-26
unicorn: provide Python bindings in a separate recipe.

OscarL:
This change allows to:

- Decouple the building of the Python bindings (otherwise needs to rebuild full libunicorn.so for each build of the bindings).
- Be able to provide bindings for multiple Python versions.

The tiny patch is the same as one already applied upstream 3 weeks ago, so it could be dropped on next release/update.

Also:

- Removed static library.
- Disabled building of tests.


korli:
@OscarL I don't know any recipes actually using unicorn, what's the usecase for python bindings?

PR #8113, opened 2023-03-26
Texlive: bump to 2023

jmairboeck:
Building this requires ICU 70 or newer and building libxml2 (an indirect dependency) against ICU 70 to get rid of icu66_devel in the build environment. These changes are done in separate commits. I can also split them into separate Pull Requests if needed.

ConTeXt is not included in TeXLive 2023. If it is needed, it should probably be built as a separate package (the `mtxrun` call was commented out from texlive_update.sh before, so it probably didn't work before anyway). See https://tug.org/texlive/bugs.html for some details or [this](https://tug.org/pipermail/tex-live/2023-March/048936.html) mailing list thread.

Because of ICU 70, this is not ready to be merged. I just enabled x86_64 _for testing_, I didn't fully enable it. This is something for the experts to look into.

A possible alternative would be to try to use the bundled ICU instead of the system one (see the arguments to `configure` in texlive_core), but because our version of ICU is patched, I suppose that we would need to apply these patches to the texlive copy too, if we use that.

For libxml2, I noticed that it includes a python subpackage for python 2.7. This could be modernized or cleaned up some time ...


jmairboeck:
I moved the needed dependency updates to their own PRs (#8117 and #8118).

jmairboeck:
I rebased this on top of the ghostscript branch. Please see #8320 for the ghostscript related changes.

jmairboeck:
I rebased this again on master.

As soon as this, #8393 and #8394 are merged, ghostscript 9 can be removed.

This one still depends on ICU >= 70, though.

Issue #2569, opened 2018-05-18
 libxml2: enable the ICU feature 


kenmays:
Enable the ICU feature of libxml2 for Qt WebEngine 5.10.1.
Enhancement: allows proper integration of libxml2 with Qt Web Engine 5.10.1
Flag: --enable-icu

diversys:
Please create a PR.

PR #8117, opened 2023-03-26
libxml2: remove ICU dependency

jmairboeck:
It wasn't used anyway because the wrong option was passed to configure.

### DIFF
--- a/dev-libs/libxml2/libxml2-2.9.13.recipe
+++ b/dev-libs/libxml2/libxml2-2.9.13.recipe
@@ -10,7 +10,7 @@ available in other environments."
 HOMEPAGE="http://www.xmlsoft.org/"
 COPYRIGHT="1998-2013 Daniel Veillard.  All Rights Reserved."
 LICENSE="MIT"
-REVISION="4"
+REVISION="5"
 SOURCE_URI="https://download.gnome.org/sources/libxml2/2.9/libxml2-$portVersion.tar.xz"
 CHECKSUM_SHA256="276130602d12fe484ecc03447ee5e759d0465558fbc9d6bd144e3745306ebf0e"
 PATCHES="libxml2-$portVersion.patchset"
@@ -37,7 +37,6 @@ if [ -z "$secondaryArchSuffix" ]; then
 fi
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libicuuc$secondaryArchSuffix
 	lib:libz$secondaryArchSuffix
 	"
 
@@ -48,7 +47,6 @@ PROVIDES_devel="
 	"
 REQUIRES_devel="
 	libxml2$secondaryArchSuffix == $portVersion base
-	devel:libicuuc$secondaryArchSuffix
 	devel:libz$secondaryArchSuffix
 	"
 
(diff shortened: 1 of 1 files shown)


jmairboeck:
see #8113

jmairboeck:
I noticed that it includes a python subpackage for python 2.7. This could be modernized or cleaned up some time ... (@OscarL)

Begasus:
This has been checked with current version in the depot, lddtree doesn't show any ICU links in the library, building it with `--with-icu` actually binds it to ICU, so this shouldn't effect any other libraries.

Begasus:
> I noticed that it includes a python subpackage for python 2.7. This could be modernized or cleaned up some time ... (@OscarL)

I think removing the support for python2.7 in total should be fine, don't think anything is using it (aside itstool)?

jmairboeck:
> I think removing the support for python2.7 in total should be fine, don't think anything is using it (aside itstool)?

Should I do that here in this PR, or separately (or someone else)?

pulkomandy:
Apparently this had been enabled by request here: https://github.com/haikuports/haikuports/issues/2569

But since it is a request from kenmays I have no trust in it being based on any real software needs. And since we have since built QWebEngine without any problems, this seems safe to disable.

Begasus:
Seems https://github.com/haikuports/haikuports/pull/2570 didn't realy do anything there, configure doesn't use `--enable-icu` but it uses `--with-icu`, so it was never build with ICU support. ;)

pulkomandy:
Yes, I saw that, and we still got QWebEngine anyway, so that's what I am saying: this original merge request was not backed by reality or became obsolete since then, and there is no problem removing ICU from libxml2. I just wanted to check that before we break something :)

jmairboeck:
As an alternative, we could also build it with ICU 70 (on gcc11 platforms) if using ICU is still desirable.

korli:
how is texlive dependent on libxml2?

pulkomandy:
It is part of Haiku core dependencies through freetype or fontconfig. So all haikuports chroots have libxml2 in.

Maybe we should continue splitting the haiku package in smaller parts and put app_server in a separate package to avoid this?

korli:
> It is part of Haiku core dependencies through freetype or fontconfig. So all haikuports chroots have libxml2 in.

I'm talking about libxml2_devel obviously. fontconfig_devel depends on libxml2_devel. It seems like cairo_devel brings fontconfig_devel.

jmairboeck:
> > It is part of Haiku core dependencies through freetype or fontconfig. So all haikuports chroots have libxml2 in.
> 
> I'm talking about libxml2_devel obviously. fontconfig_devel depends on libxml2_devel. It seems like cairo_devel brings fontconfig_devel.

That is exactly what is happening, and that then causes both icu66_devel and icu70_devel to be included (before this PR) which doesn't work.

korli:
> That is exactly what is happening, and that then causes both icu66_devel and icu70_devel to be included (before this PR) which doesn't work.

IMO switching fontconfig from libxml2 to expat would be better. It looks Arch Linux and Gentoo use expat so that should be a safe change.

PR #8118, opened 2023-03-26
icu70: enable for x86_64 and x86

jmairboeck:
This is needed for texlive_core 2023 (#8113).

Also add a conflict for the devel package with icu66_devel. Others, like icu67 should be added as well, when they are made available.

### DIFF
--- a/dev-libs/icu/icu70-70.1.recipe
+++ b/dev-libs/icu/icu70-70.1.recipe
@@ -41,7 +41,7 @@ wrapping when displaying the text."
 HOMEPAGE="http://www.icu-project.org"
 COPYRIGHT="1995-2020 IBM Corporation and others."
 LICENSE="ICU"
-REVISION="1"
+REVISION="2"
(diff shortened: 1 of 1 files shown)


Begasus:
ICU66 and ICU70 base packages don't conflict with eachother, for the devel package there should be a conflict (uninstalling icu66_devel doesn't take that many `devel packages` down on my system, so should be fine)

waddlesplash:
Why don't we upgrade to ICU 72 directly? Indeed it's high time we change versions though.

korli:
At least it would be interesting to get parts of the ICU patchset upstreamed.

jmairboeck:
Is ICU 70 obsoleted by #8631?

Begasus:
> At least it would be interesting to get parts of the ICU patchset upstreamed.

Partially created a PR upstream: https://github.com/unicode-org/icu/pull/2463

PR #8134, opened 2023-03-27
docutils: clean up, support Python 3.10.

OscarL:
- Use loops to generate the package/version specific PROVIDES.
- Use version suffixes for the provided cmds, except for the ones from the package for the default Python version (3.9 currently).

Note:

~Moved suffixes after the ".py", as names like "rst2html4310.py" looked weird, and we can't use "rst2html4-3.10.py" due to the "no dashes in PROVIDES or REQUIRES" rule.~

As names like `rst2html4310.py` look weird, use the easier to read: `rst2html4-3.10.py` naming style. See comments below.

### DIFF
--- a/dev-python/docutils/docutils-0.18.1.recipe
+++ b/dev-python/docutils/docutils-0.18.1.recipe
@@ -11,7 +11,7 @@ LICENSE="Public Domain
 	BSD (2-clause)
 	GNU GPL v3
 	Python"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://pypi.io/packages/source/${portName:0:1}/$portName/$portName-$portVersion.tar.gz"
 CHECKSUM_SHA256="679987caf361a7539d76e584cbeddc311e3aee937877c87346f31debc63e9d06"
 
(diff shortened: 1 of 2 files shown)


korli:
rst2html4-3.10.py becomes rst2html4_3.10.py in PROVIDES

korli:
@OscarL the question is about command suffixes, right?
Normally there is no need to install commands for each python version, because one needs only one command...
Otherwise always dotted: script_name3.10
Actually IMO this recipe should create symlinks like script_name -> script_name.py,  script_name3.10 -> script_name.py3.10, ..
I don't remember any recipes with script_name310, but should be the exception.

korli:
> Regarding the installation of multiple versions...

> Of the two, I believe the second is better: The user might have no need/use for Python 3.9, but desires to use 3.10 (or other) version, along with its correspondent scripts, so they will install, say, `docutils_python3.10`, and would like use the scripts it provides. (otherwise needing to install another package just for the scripts).

A command is a tool, having the same tool in different python versions is questionable, because in the end it's the same tool. It's like if we provide pkgman and pkgman-x86, because the user might want to use the x86 binary instead of x86_gcc2.
Surely there are usecases, for example pip, where the mapping to the python version is obvious.

korli:
@OscarL  is this PR ready to go?

korli:
the recipe is OK for me actually. Symlinks are only nice to have.

dev-python/docutils/docutils-0.18.1.recipe
```diff
@@ -79,10 +100,16 @@ INSTALL()
...
+		for f in $binDir/*; do
+			mv $f ${f%.py}-${pythonVersion}.py
+		done
+
+		# And provide suffix-less symlinks for the default version
+		if [ $pythonVersion = $defaultVersion ]; then
 			for f in $binDir/*; do
-				mv $f ${f%.py}${commandSuffix/_/-}.py
+				scriptName=`basename "$f"`
+				ln -s $scriptName $binDir/${scriptName%-$pythonVersion.py}
```
korli:
```suggestion
				ln -sr $f ${f%-$pythonVersion.py}
```
Untested :)

korli (approved):
Thanks!

PR #8143, opened 2023-03-28
pygments: cleanup, update to version 2.14.0

OscarL:
Move the PROVIDES_* inside the loop, fix script suffixes for 3.10.

Install from the wheel file instead of using the tarball for faster package buids (1.1 MB vs 4.4 MB file download).

### DIFF
--- a/dev-python/pygments/pygments-2.14.0.recipe
+++ b/dev-python/pygments/pygments-2.14.0.recipe
@@ -12,11 +12,11 @@ ANSI sequences
 * it is usable as a command-line tool and as a library
 * ... and it highlights even Brainf*ck!"
 HOMEPAGE="https://pygments.org/"
-COPYRIGHT="2006-2022 by the Pygments team"
+COPYRIGHT="2006-2023 by the Pygments team"
 LICENSE="BSD (2-clause)"
-REVISION="4"
+REVISION="1"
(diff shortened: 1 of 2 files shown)


korli:
Usually we keep the tarball because the tests, the manpage, the docs are otherwise not provided.

korli:
It looks good to me. You might install the manpage, if it's missing.

dev-python/pygments/pygments-2.14.0.recipe
```diff
@@ -50,32 +53,36 @@ BUILD_PREREQUIRES="$BUILD_PREREQUIRES
...
 			--root=/ --prefix=$prefix
 
-		if [ $pythonPackage = python38 ]; then
-			mv $binDir/pygmentize $binDir/pygmentize3.8
+		mkdir -p $manDir/man1
+		cp doc/pygmentize.1 $manDir/man1
+
+		if [ "$pythonVersion" = "$commandSuffix" ]; then
+			for f in $binDir/*; do
+				mv $f ${f}$commandSuffix
```
Begasus, line 79: This is somehow broken? 
```
Warning: POLICY ERROR: no matching provides "cmd:pygmentize" for "bin/pygmentize"
```

Begasus: Fixed in: https://github.com/haikuports/haikuports/pull/9852

PR #8148, opened 2023-03-28
coverage: update to version 7.2.2.

OscarL:
- Drop support for Python 2.7 and 3.7.
- Add it for 3.9 and 3.10.
- Include man page and README.html files.

----

I did used `git mv` to rename the recipe, but I guess the changes were too substantial for git to keep it as a rename + changes?

Also. I've named the man page "coverage.1" (from `python_coverage.1`) to match the "cmd:" names. Can revert that if desired.


dev-python/coverage/coverage-7.2.2.recipe
```diff
@@ -0,0 +1,125 @@
...
+
+	eval "REQUIRES_$pythonPackage=\"
+		haiku$secondaryArchSuffix
+		lib:libpython$pythonVersion$secondaryArchSuffix
+		\""
+	eval "BUILD_REQUIRES+=\"
+		setuptools_$pythonPackage
+		cmd:groff
+		cmd:rst2html.py
+		cmd:rst2man.py
```
korli, line 62: These should go in BUILD_PREREQUIRES. IMO they should be out of the loop. cmd:gcc$secondaryArchSuffix too.

PR #8155, opened 2023-03-29
libiodbc, add libVersionCompat, disable and remove static library

Begasus:


Begasus:
@waddlesplash this one produced a static library (libodbc.a) and shared library (libodbc.so) without an error from haikuporter, a flaw here?

dev-libs/libiodbc/libiodbc-3.52.12.recipe
```diff
@@ -49,14 +52,21 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	runConfigure ./configure --with-iodbc-inidir=$settingsDir/etc
+	runConfigure ./configure --with-iodbc-inidir=$settingsDir/etc \
+		--disable-static
```
korli, line 56:
```suggestion
		--disable-static --disable-libodbc
```

korli:
Could be bumped to https://github.com/openlink/iODBC/archive/refs/tags/v3.52.15.tar.gz

dev-libs/libiodbc/libiodbc-3.52.12.recipe
Begasus: OK, could add this suggestion in the browser, but I know I would mess up getting things on par. Checking. :)

korli:
Sure. Just suggesting, when only the minor version number is incremented, it shouldn't be a problem to build correctly :)

Begasus:
Leaving as wip for a while so @waddlesplash can investigate :)

waddlesplash:
Well, without seeing an exact version of the recipe with the problem, there isn't much for me to investigate.

Begasus:
> Well, without seeing an exact version of the recipe with the problem, there isn't much for me to investigate.

As mentioned on IRC, you can build the current one in haikuports that still has the old changes and libodbc in prepareInstalledDevelLibs, only need to add --disable-static for the default libraries, no need to check this, here it is fixed.

PR #8159, opened 2023-03-29
libnslog, don't build/install the static library

Begasus:


dev-libs/libnslog/libnslog-0.1.3.recipe
```diff
@@ -49,15 +49,15 @@ TEST_REQUIRES="
 
 BUILD()
 {
-	for linkage in lib-static lib-shared; do
+	for linkage in lib-shared; do
```
jmairboeck, line 52: The loop could have been removed, now that there is only one kind of linkage (below too).

PR #8160, opened 2023-03-29
libnsutils, don't build/install the static library

Begasus:


dev-libs/libnsutils/libnsutils-0.1.0.recipe
```diff
@@ -41,15 +41,15 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
-	for linkage in lib-static lib-shared; do
+	for linkage in lib-shared; do
```
jmairboeck, line 44: here the same as in #8159

PR #8161, opened 2023-03-29
tqdm: update to version 4.65.0

OscarL:
- Switch to "any" arch (as this is pure Python code).
- Added support for Python 3.10.
- Added bash completion file for the default "cmd:tqdm".
- Added man page.

---

Used "git mv" + modifications, but seems the result was too different to register as a renamed file.


dev-python/tqdm/tqdm-4.65.0.recipe
```diff
@@ -0,0 +1,95 @@
...
+is an abbreviation for *I love you so much* in Spanish *te quiero demasiado*.
+Instantly make your loops show a smart progress meter - just wrap any iterable with \
+``tqdm(iterable)``, and you're done!"
+HOMEPAGE="https://github.com/tqdm/tqdm"
+COPYRIGHT="2015-2023 tqdm contributors"
+LICENSE="MIT
+	MPL v2.0"
+REVISION="1"
+pypiVersion="3d/78/81191f56abb7d3d56963337dbdff6aa4f55805c8afd8bad64b0a34199e9b"
+SOURCE_URI="https://files.pythonhosted.org/packages/$pypiVersion/$portName-$portVersion.tar.gz"
```
korli, line 12: Doesn't work with https://github.com/tqdm/tqdm/archive/refs/tags/v4.65.0.tar.gz ?

Begasus: You should take care with using `$portNam`e also, if this wasn't `any` it would become `tqdm_x86` then (32bit secondary arch)

Begasus: > Yup, noticed that on other recipes. Would be better to just not use `$portName` there ever, just in case?

IMHO nothing architecture should be used in SOURCE_URI (my 2 cents)

Begasus (approved):

Begasus:
LGTM

PR #8164, opened 2023-03-29
ripgrep: fix install path for bash completions.

OscarL:
There was an final "s" on "$dataDir/bash-completion/[...]" that wasn't supposed to be there.

---

Untested (too many dependencies, + rust, for my poor PC/connection).

I didn't rev-bump it, as the fix is small, and will be pick up on future updates. Meanwhile user can manually move the files to the proper path.


Begasus:
I don't think this will be updated anytime soon, so maybe best revbump it?

PR #8172, opened 2023-03-29
psutil: move to Python 3.9, add TEST().

OscarL:
Package is non-functional: on both 3.7 and 3.9 it gives an error when trying to "import psutil".

The patchset is WIP, so attempting to update to psutil 5.9.4 won't do much.

Still, moving away from 3.7 makes sense.


korli:
What's the error when trying to "import psutil"?

korli:
Right, unfinished/unmaintained. :)

PR #8207, opened 2023-03-31
colm, fix recipe, disable static library (that doesn't conflict when …

Begasus:
…--disable-static isn't used?


waddlesplash:
I fixed the lack of an error from HaikuPorter for this one in https://github.com/haikuports/haikuporter/commit/df3c0a88bcd57b46ac1cb7b42c876ec9a0bffe60.

PR #7051, opened 2022-07-12
Fontforge: Initial GUI support

extrowerk:

PR #8222, opened 2023-04-01
fontforge, bump version

Begasus:


Begasus:
Should be tested. Supersedes version in: https://github.com/haikuports/haikuports/pull/7051
This one is still build without GUI support, keeping the old layout.

PR #8226, opened 2023-04-01
libsdl2: remove cmake file for libsdl2_test.a

nephele-gh:
The cmake file for libSDL2_test.a was still there, causing build failures for dependant proijects (because of broken cmake file?)


korli:
Please bump the revision number.

korli (approved):

PR #8256, opened 2023-04-03
monsterz: retire recipe.

OscarL:
Game was Python 2.x only (2.4 in fact!).

Tried to update its dependencies and even gave it a go with 2to3.py, plus minor manual hacking... no go.

Good bye, Monsterz, we hardly knew you!


pulkomandy:
There is a Python 3 patch from the Debian maintainers: https://salsa.debian.org/games-team/monsterz/-/tree/master/debian/patches

Can we restore the recipe using this?

PR #8287, opened 2023-04-05
glew, fixes and use libVersionCompat

Begasus:
GLEW_PREFIX is being used to set the path in pkgconfig file, moving the pkgconfig file to $libDir instead of $developLibDir enables fixPkgconfig to do it's work to set the correct path for the headers.

### DIFF
--- a/media-libs/glew/glew-2.2.0.recipe
+++ b/media-libs/glew/glew-2.2.0.recipe
@@ -10,7 +10,7 @@ COPYRIGHT="2007 The Kronos Group Inc.
 	2002 Lev Povalahev
 	1999-2007 Brian Paul"
 LICENSE="BSD (3-clause)"
-REVISION="2"
+REVISION="3"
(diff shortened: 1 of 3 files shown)


Begasus:
Could use some checking:

Before:
```
prefix=/usr
exec_prefix=${prefix}
libdir=/packages/glew_x86-2.2.0-2/.self/lib/x86
includedir=${prefix}/include

Name: glew
Description: The OpenGL Extension Wrangler library
Version: 2.2.0
Cflags: -I${includedir} 
Libs: -L${libdir} -lGLEW -lGL
Requires: glu
```

With this change:
```
prefix=/packages/glew_x86-2.2.0-3/.self
exec_prefix=${prefix}
libdir=${prefix}/develop/lib/x86
includedir=${prefix}/develop/headers/x86

Name: glew
Description: The OpenGL Extension Wrangler library
Version: 2.2.0
Cflags: -I${includedir} 
Libs: -L${libdir} -lGLEW -lGL
Requires: glu
```

Begasus:
Checked with a few recipes.

PR #8301, opened 2023-04-07
libp8_platform, cleanup, fix cmake configuration install path

Begasus:


Begasus:
Using `fixCMake` here only creates an empty `$developLibDir/cmake`.

PR #8310, opened 2023-04-07
libnice, use libVersionCompat, fix build (needs 'unset meson' ?)

Begasus:


net-libs/libnice/libnice-0.1.18.recipe
```diff
@@ -60,25 +63,30 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
+	unset meson
```
Begasus, line 66: Not quite sure why this is needed, without it I'm getting an error that the buildtype isn't set (although it is).

Begasus: I added the buildtype first, cleaned ... but still had an error that the buildtype wasn't set (allthough in the summary it showed it was there), without unset meson I couldn't get the build started.

korli: compile and install aren't supported by the scriptlet...

PR #8320, opened 2023-04-09
ghostscript_gpl: bump version, remove ghostscript8

jmairboeck:
This requires rebuilding ports that use libgs, because the so version changed from 9 to 10. Currently these are:

 * bghostview
 * texlive_core (see #8113)
 * lilypond (this requires both libgs and a working texlive to build)


jmairboeck:
@pulkomandy @Sylvain78 please review.

I'll rebase the texlive PR on this one.

app-text/ghostscript_gpl/patches/ghostscript_gpl-10.01.1.patchset
```diff
@@ -90,66 +67,66 @@ index 3312477..f627cbf 100644
...
              CC_VISIBILITY_FLAGS_TO_TRY="-fvisibility=hidden"
          fi
          ;;
 -- 
-2.30.2
+2.37.3
 
 
-From 8fd422a85278e54141866273a61d3eeb0d2245b1 Mon Sep 17 00:00:00 2001
+From 978e4572d5186fc354aa99258cdeba29f3682ebc Mon Sep 17 00:00:00 2001
```
jmairboeck, line 119: This could be removed together with the next commit. The next one below reverts this one.

korli:
I would keep a ghostscript_gpl library only package for version 9.

jmairboeck:
> I would keep a ghostscript_gpl library only package for version 9.

I'll reintroduce a `ghostscript_gpl9` compatibility package with just the library.

korli:
Is ghostscript8 still needed?

jmairboeck:
> Is ghostscript8 still needed?

@pulkomandy explained to me before that it is needed for BeOS compatibility. It is therefore only enabled for x86.

pulkomandy:
It was needed for BGhostView, I don't remember if BGhostView was migrated to a newer/more standard API or not. I'm not aware of anything else needing it.

jmairboeck:
Ghostscript only uses C, no C++, so the ABI shouldn't matter if I understand that correctly.

jmairboeck:
> Ghostscript only uses C, no C++, so the ABI shouldn't matter if I understand that correctly.

Apparently, I'm wrong. The BeOS driver uses C++, and doesn't build with gcc2:

```
/sources/ghostscript-8.14/src/gdevbe.cpp:99: member `struct rgb_color {anonymous union}::rgb' with copy assignment operator not allowed in union
```

pulkomandy:
The C++ ABI is not the only problem you will face when mixing versions of gcc. There are changes in libgcc that will cause problems without quite a bit of hackery. We do it for ffmpeg, but that's pretty much the only place.

The needed things aren't there in Ghostscript 8, so that wouldn't work. 

Also, it does use C++: the whole point of that Ghostscript 8 recipe was to restore support for output to BBitmap, which is what was used by BGhostView. This is in the patchset as src/gdevbe.cpp, which makes use of the Be API and is thus written in C++. There is no hope to build that part with gcc 11 and then make use of it in a gcc2 app.

As for BGhostView, it was indeed modified to use the newer 'gsapi' interface to ghostscript, instead of the old 'gsdll' with BeOS patches: https://github.com/HaikuArchives/BGhostView/commit/e2095fa55c95754cb2dacc53d67afb5dec95f111 so there is no problem here, it should be able to use newer versions of Ghostscript.

This is also confirmed by the fact that the recipe is already updated to use Ghostscript 9: https://github.com/haikuports/haikuports/blob/master/haiku-apps/bghostview/bghostview-20190401.recipe

There is really no need to keep this Ghostscript 8 around. Not for BeOS, and not for BGostview either.

jmairboeck:
I removed ghostscript8. Thanks for the explanation.

jmairboeck:
I rebuilt both Lilypond and BGhostview locally with Ghostscript 10. Should I add them to this PR or separately? (I built Lilypond with texlive 2023 but it shouldn't matter, texlive is only a build requirement for lilypond).

Begasus:
Good to go?

app-text/ghostscript_gpl/ghostscript_gpl9-9.55.0.recipe
```diff
@@ -23,75 +25,54 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
...
-	cmd:ps2pdf12
-	cmd:ps2pdf13
-	cmd:ps2pdf14
-	cmd:ps2pdfwr
-	cmd:ps2ps
-	cmd:pv.sh
-	cmd:unix_lpr.sh
-	cmd:wftopfa
-	lib:libgs$secondaryArchSuffix = 8.14 compat >= 8
+	cmd:gsx
```
korli, line 30: commands should be removed.

jmairboeck: done (although untested)

PR #8324, opened 2023-04-10
agg, disable static library

Begasus:


pulkomandy:
Static libraries won't change anything there? The tempates are in .h files and recompiled everytime, no matter how the other parts of the library are linked.

PR #8335, opened 2023-04-11
libupnp, disable static libraries, fix libVersion

Begasus:


Begasus:
Need some more checking here as libVersion changed for libupnp.

PR #8337, opened 2023-04-11
libxcb, disable static library

Begasus:


Begasus:
Marked x86_gcc2 as broken (no libcheck there).

PR #8343, opened 2023-04-11
libcaca, disable static library

Begasus:


Begasus:
32bit buildmaster failed retreiving the source, the SOURCE_URI works fine locally?

PR #8345, opened 2023-04-12
mpd, revbump for libupnp

Begasus:


Begasus:
Needs to wait untill libupnp is build on 64bit buildmaster (offline for a while now).

PR #8361, opened 2023-04-13
libharu, bump version, disable static library

Begasus:


Begasus:
Patch taken from Alpine Linux: https://git.alpinelinux.org/aports/tree/community/libharu/soversion.patch

Begasus:
@korli In the previous patchset I saw you removed the bindings from the build, I left them in for now, but they could easily be removed with: `rm -rf $dataDir`?

korli:
Not sure what you mean with removed bindings.

Begasus:
> Not sure what you mean with removed bindings.

The DEVPAK was removed here: https://github.com/haikuports/haikuports/commit/0b0a03322bfd6d544899e68abdccb97fdfbb1720#diff-2d2e7b9d280475457b5616ee86ef6cdfc8d9268311288c70b96dab36b167004dR19

korli:
Can't remember sorry. Is DE PACK off by default?

Begasus:
> Can't remember sorry. Is DEVPACK off by default?

I think it's on by default, but probably could be disabled, just a question if it would be usefull (only one way to find out I guess) ;)
I'll move $dataDir to the devel package and see where this leads from there. (if that is OK with you) :)

PR #8379, opened 2023-04-14
liblxi, bump version, cleanup

Begasus:


Begasus:
@kallisti5 the package isn't enabled so far for 64bit (didn't check a build, but I think it should be fine there), not being used by any other package so far, so bumped version, could enable 64bit if there would be a need for it.

Issue #8025, opened 2023-03-20
Drop support for Python 3.7


OscarL:
~As beta4 uses Python 3.9 as default, and as Python versions 3.7 and 3.8 are in "Source-only security fix releases" mode (with even less than 3 months of that for 3.7), we should drop support for the older ones (adding support for Python 3.10 where possible).~ Edit: scope for this issue has changed following discussion on comments below.

As beta4 uses Python 3.9 as default, and as Python versions 3.7 is nearing EOL, we'll be dropping support for it (and also for Python 2.7 package that might still remain), while adding support for Python 3.10 where possible.

To help keep track of the progress on this regard, I'll add some tasks list:

## Regular packages (recipes not under `dev-python/`):

List of *. recipes that contain either of:

	- "cmd:python3.7"
	- "_python3$"
	- "libpython.*m"

(and for Python 2.x: `_python$`, `cmd:python$` and `libpython2`)

- [x] aquaskk (Python 2.x) #8780
- [x] atk (Python 2.x) #8768
- [x] aflplusplus #8730
- [x] automake #8517
- [x] bakefile (Python 2.x) #9214
- [x] BeCJK (Python 2) #8499
- [x] bliss060 (Python 2.x) #9210
- [x] brotli #8221
- [x] bzr (Python 2.x) #9172
- [x] caffe (Python support is disabled on the recipe already)
- [x] calibre
- [x] calligra (Python 2.x) #9226
- [x] cegui #8701
- [x] chocolate_doom (Python 2.x) #8772
- [x] command_not_found (Python 2.x) #8497
- [x] cram (Python 2.7 and 3.7) #8725
- [x] crawl
- [ ] csound (Python 2.x)
- [x] cudatext #8732
- [x] dblatex #8702
- [x] distcc #8519
- [ ] edk2 (Python 2.x)
- [x] fife (Python 2.x,) (see #8840 / #8839)
- [x] fontforge #8239
- [x] glances #8171
- [x] glib2 (Python 2.x) #9191
- [x] gmsh #8430
- [x] graphviz #8241
- [x] grpc (Python2.x and 3.x, only for TEST(), marked broken on 32 bits) #8752
- [x] gtest #8718
- [x] gyp (Python 2.x) #8779
- [ ] handbrake (Python 2.x)
- [x] hugen (Python 2.x) #8502
- [x] iniparse (Python 2.x) #9189
- [x] itstool (Python 2.x) #8605
- [x] json_glib #8505
[...]

PR #8384, opened 2023-04-14
libmypaint, switch python version

Begasus:


Begasus:
https://github.com/haikuports/haikuports/issues/8025

PR #8402, opened 2023-04-15
libsndfile, cleanup, move cmake config to dev package, fix libVersion

Begasus:
Libversion(Compat) was declared wrong for this one, needs further testing, guess quite some depending packages will need a revbump for this one.


Begasus:
Last version bump didn't produce problems, should be good to merge then? (last bumped changed libVersion from 1.0.31 to 1.1.0 (which is actually 1.0.34)).

Issue #8410, opened 2023-04-16
Restore Monsterz recipe


pulkomandy:
In https://github.com/haikuports/haikuports/pull/8256, monsterz was removed because it needs Python 2.

There is a Python 3 patch maintained by Debian: https://salsa.debian.org/games-team/monsterz/-/tree/master/debian/patches

We could use that to restore the recipe.

PR #8416, opened 2023-04-17
monsterz: restore recipe.

OscarL:
This addresses #8410 by:

- Reverting #8256.
- Applying Debian's patches, so it it works with Python 3.x.
- Other minor fixes to the recipe, and location of the settings file (by patching configdir on monsterz.py).

Added a vector icon file generated (vía autotracer.org) based on icon.png.

Tested on 64 and 32 bits.


Begasus:
Good to go?

pulkomandy:
You can't add an icon as a BResource to anything else than an ELF executable. You can however add it directly to the file attributes instead (using 'addattr' for example).

As for starting from DeskBar, probably this results in the app getting the wrong "current directory" at start (we have patched a lot of software for this: on Linux and Windows the default is starting in the place where the executable is located, but in our case we start in /boot/home).

pulkomandy:
> I also tried to use an .rdef file, but Icon-O-Matic was generating a file way too big for my taste (when exporting the icon).

Well, that's what you get from auto-tracing it from a PNG file. It results in way too complex SVG which in turns results in a very large HVIF file. It should be cleaned or re-drawn by hand.

Begasus:
Maybe keep the svg icon if someone wants to tackle it, and create an issue for this icon, ps, in previous version there wasn't an icon for it?

pulkomandy:
I made a smaller icon.

I started from graphics.svg in the source repository and extracted just the red monster using inkscape. Then I imported that in Icon O Matic and optimized it a bit to remove duplicate shapes, and some details. It's down to 1.6KB as HVIF now.

Making it even smaller would require realigning the control points to the grid, which would change the monster shape a bit, but should not really be noticeable on a small icon.

http://pulkomandy,tk/drop/monsterz.svg
http://pulkomandy,tk/drop/monsterz.iom
http://pulkomandy,tk/drop/monsterz.hvif

PR #8418, opened 2023-04-17
libwebm, don't install static library

Begasus:


Begasus:
This is quite a common used library, don't think the static library should be used, but better to be safe here.
If anything is blocking this from merging please add a comment (or if it should be fine too).

PR #8420, opened 2023-04-17
libxmp, cleanup

Begasus:


Begasus:
@diversys cleaned up some things here, shouldn't harm the base package, for the devel package some strings change.

PR #8439, opened 2023-04-19
paragui, drop reduntant, unmaintained and broken recipe

Begasus:


Begasus:
@pulkomandy you did the patchset in 2015, this isn't maintained anymore (last release (this one) was from 2004).
If you have no need for it this can be removed from the repository. (not in the depot anymore too)

pulkomandy:
The patchset is attributed to me because I converted it from an old .patch file in 2015, but I didn't write it. I have no need for this library, I did this only because there was a ticket about it on the old Trac bugtracker (which seems dead now).

PR #8485, opened 2023-04-22
youcompleteme: retire old recipe versions.

OscarL:
These require libpython3.6m, which we don't support anymore.


Begasus:
Doesn't this completely remove it from the depot?

pulkomandy:
I don't see any reason to remain on the old recipes. Let's start from the latest one and try to fix that.

You should be able to "fix" your terminal with the "reset" command after this. Or you can try testing in gvim, that may cause less problems with the terminal.

PR #8492, opened 2023-04-23
wavpack, fix libVersion

Begasus:


Begasus:
This is used by ffmpeg4 among others, libVersion changes from 1.2.0 (wrong) to 1.2.4 (correct), installing it locally doesn't break anything as the version is higher then previous (unlike libsndfile) so should be safe to merge.
But could use a review to make sure it is. :)

Issue #8514, opened 2023-04-25
manpages installed in the wrong directory


Begasus:
Upon checking package bind_utils I saw that it installed the manpages in $docDir (cp -v doc/man/{dig.1,host.1,nslookup.1} $docDir)
After that looking in $manDir I see faulty directories there too (needs cleanup).

augiedoggie:
Is this still an issue that needs to be tracked?  I haven't noticed any packages with misplaced man pages.

Begasus:
Will do another checkup on my installed system, I guess most issues are resolved atm, if not they could still be linked here if need be, thnx for the reminder.

Begasus:
- [ ] freerdp has manpages installed in $dataDir/man

Begasus:
Closing, if something pops up just link it here.

PR #8515, opened 2023-04-25
bind_utils, move manpages to the correct place

Begasus:


Begasus:
Issue for this is created at: https://github.com/haikuports/haikuports/issues/8514

PR #8520, opened 2023-04-25
libtool, disable static library, remove libtool file

Begasus:


waddlesplash:
Actually, I think this one is a case where we want to delete the shared library, not the static library.

Begasus:
Thanks will give it another spin.

korli:
> Actually, I think this one is a case where we want to delete the shared library, not the static library.

Why actually?

waddlesplash:
The shared library is a set of helper routines for dlopen only, I think? I'm not sure it really warrants being dynamically linked.

Begasus:
Just did a check on PR https://github.com/haikuports/haikuports/pull/8141

With the changes from here I only get a static library build in the source, with the current one from the depot (with the shared library from libtool) I'm getting the shared library for libtermkey ...

Begasus:
Reverted changes and now it disables the static library, will do some further testing, but so far looking good.

PR #8521, opened 2023-04-25
sparse, remove obsolete patch

Begasus:


Begasus:
@scottmc this patchset was added in 2010, any reason to keep it around?

PR #8558, opened 2023-04-30
qemu-8.0.0: new recipe

davidkaroly:
not sure about the old recipes, shall we remove them? maybe at least the ones older than 3.1.1.1?


waddlesplash:
Yes, probably.

PR #8591, opened 2023-05-05
tiny, new recipe

Begasus:


augiedoggie:
The recipe should probably be called `dev-lang/tiny/tiny-v0.x.x`.  The `first` part in the port name is just a git tag that was used instead of `v0.x.x` or something.

Begasus:
> The recipe should probably be called `dev-lang/tiny/tiny-0.x.x`. The `first` part in the port name is just a git tag that was used instead of `v0.x.x` or something.

Done :)

pentalive:
Hey All - Thanks for creating this for Tiny. I hope many have some fun with this.  I looked over the recipe, I don't know much about recipes but it looks excellent to me.  Thanks again!

pentalive:
Does this mean that tiny will show up in HaikuDepot under devel?

Begasus:
> Does this mean that tiny will show up in HaikuDepot under dev-lang?

Yes :)

PR #8618, opened 2023-05-09
gtk_doc, move to secondary arch for 32bit, use glib2-2.75.1 instead o…

Begasus:
…f old glib2-2.57.1


korli:
Actually we should split the docs of glib2, remove the dependency devel:glib2, patch in common.py GetModuleDocDir().

Begasus:
Guess this can be closed, fixed in https://github.com/haikuports/haikuports/commit/37c0ba44606bbfcc1523c74a3e8159530d47ca2e

korli:
Maybe my change is incomplete, the glib2 docs might be needed in REQUIRES. Well, let's see if this breaks some other recipe using gtk_doc.

Begasus:
It will pop up soon enough when that's the case. :)

Issue #8626, opened 2023-05-11
mpv: app signature required


danboid:
I would like to be able to play video files from Tracker using mpv by right-clicking on it and choosing 'Play with mpv' and have the option to configure mpv as the default video player for Tracker via the FileType tool but it seems it requires an app signature for this to work.

danboid:
Hi Oscar

Thanks for quickly fixing this on the ports side. I've added a comment to that bug to link to this issue so there is another reason registered to get it fixed.

I intend to read the Haiku developers guide soon so that I might be able to have a go at fixing such things.

PR #8630, opened 2023-05-12
mpv: update to 0.35.1, minor fixups.

OscarL:
- Added an .hvif file. Converted from "etc/mpv-gradient.svg".
- Added application resources (signature, icon, etc).
- Build/install man pages.

Prompted by https://github.com/haikuports/haikuports/issues/8626. Still not working with Tracker's "Open with", as our mpv needs further patching to handle that usage.

Thus, the "resource file_types message" in the ".rdef.in" file has been left commented out for now (to avoid misleading users into thinking it will work).

----

(previous PR description, obsolete, as we're not using B_ARGV_ONLY now, leaving it here so the conversation below makes sense)

- Added an .hvif file. Converted from "etc/mpv-gradient.svg".
- Added application resources (signature, argv-only, icon, etc).
- Build/install man pages.

Addresses https://github.com/haikuports/haikuports/issues/8626.

As there seems to be a bug/limitation in Tracker's "Open with" when using the B_ARGV_ONLY flag, the "resource file_types message" in the .rdef.in file has been left commented out for now (to avoid misleading users into thinking it will work).

(Tracker issue: https://dev.haiku-os.org/ticket/11893)


augiedoggie:
> mpv correctly appears in Tracker's "Open with" menu for the supported file types now, but due to some bug/limitation in Tracker, it doesn't actually open the files in mpv :-(

I don't think we should add this part until it actually works.

media-video/mpv/mpv-0.35.1.recipe
```diff
@@ -130,6 +133,22 @@ INSTALL()
...
+	sed \
+		-e "s|@MAJOR@|$MAJOR|" \
+		-e "s|@MIDDLE@|$MIDDLE|" \
+		-e "s|@MINOR@|$MINOR|" \
+		-e "s|@SHORT_INFO@|$SHORT_INFO|" \
+		-e "s|@LONG_INFO@|$LONG_INFO|" \
+		$portDir/additional-files/$portName.rdef.in > $portName.rdef
+
+	addResourcesToBinaries $portName.rdef $binDir/$portName
+	mimeset -f $binDir/$portName
```
augiedoggie, line 150: Have you tested this on 32 bit? `$portName` will be `mpv_x86` so I'm not sure some of these commands will work.  Specifically the `$portDir/additional-files/$portName.rdef.in` and  the two with `$binDir/$portName`.  I haven't had a chance to try it out though.

PR #8631, opened 2023-05-12
icu73, add recipe

Begasus:


Begasus:
For now still disabled, could use some testing (ran the testsuite for both icu66 and icu73, both were comparable)

jmairboeck:
See also #7665 and #8118.

PR #8639, opened 2023-05-14
chrpath: new recipe

OscarL:
Tested (*very* lightly) only on beta4 64 bits (now also lightly tested on 32 bits).

----

I assume we're supposed to use stuff as `CMAKE_SKIP_RPATH` instead if that's available, or disable `rpath` via some `sed` patching if the build system doesn't provides an easier mechanism.

I just had a "complex to build" project (still no proper .recipe for it), and I'll probably need either this `chrpath` thing, or some serious `sed` scripts :-D.

~Leaving it as Draft for now,~ as I'm not sure my .patchset is correct, or if this `chrpath` is something we actually want/need.

----

Drop Draft status:

If we have `cmd:patchelf`, I guess having this one also available can't hurt that much :-)


kallisti5:
Looks good to me :-)

PR #8657, opened 2023-05-17
cabextract, bump version

Begasus:


Begasus:
Build failure on 64bit (this one is for a "real" developer :) ):

```
./mspack/system.h:35:21: error: static declaration of 'memset' follows non-static declaration
  35 | static inline void *memset(void *s, int c, size_t n) {

...

./mspack/system.h:40:22: error: static declaration of 'strlen' follows non-static declaration
  40 | static inline size_t strlen(const char *s) {
```

https://build.haiku-os.org/buildmaster/master/x86_64/logviewer.html?buildruns/8876/builds/79857.log

PR #8663, opened 2023-05-17
Adds toot, a CLI client for Mastodon

FabianReidinger:
As this is my first commit, please check thoroughly. It worked locally.


Begasus:
@FabianReidinger with some small changes and a few requirements missing this could be added to the depot, if you open it again we could guide you through the process. Thanks for taking an interest into haikuports!

PR #8674, opened 2023-05-18
botan3, add recipe for version 3.0.0

Begasus:


Begasus:
Not sure if I should move the cmd to a tools package (will conflict with botan 2* version) and create a seperate package for the python binding?

korli:
Looks good.

PR #8675, opened 2023-05-18
hatchling, new python recipe

Begasus:


Begasus:
Could have a look into, not so sure about the REQUIRES vs PRE_REQUIRES :)
Pings @OscarL

PR #8683, opened 2023-05-19
hstr: cleanup recipe, update to version 3.1

OscarL:
Notice: people updating from 2.x versions will need to update their bash settings on the "~/config/settings/profile" according to the instructions given by `hstr -B`.

Users of zsh, should follow `hstr -Z` instead.


app-shells/hstr/hstr-3.1.recipe
```diff
@@ -1,23 +1,17 @@
...
-CHECKSUM_SHA256="085f8a087481bcdf33e75e2fa5aaa9289931782c0bee2db3e02425b9a7d83cdf"
+CHECKSUM_SHA256="e5293d4fe2502662f19c793bef416e05ac020490218e71c75a5e92919c466071"
 PATCHES="hstr-$portVersion.patchset"
 
 ARCHITECTURES="all !x86_gcc2"
-SECONDARY_ARCHITECTURES="x86"
-
-# On x86_gcc2 we don't want to install the commands in bin/<arch>/, but in bin/.
-commandBinDir=$binDir
-commandSuffix=$secondaryArchSuffix
```
Begasus, line 17: Any reason why to move this from $prefix/bin to $binDir?

PR #8692, opened 2023-05-20
media-sound/ocp: 0.2.105 version bump

mywave82:


Begasus:
Checked on both architectures?

mywave82:
> Checked on both architectures?

I not an experienced user, so only done build-test with the out-of-the-box haikuports.

The flags currently in recipe is based on earlier user experiences to my knowledge.

Begasus:
No problem, let's keep track on the buildmasters and eventual issues coming up after the merge.

PR #8705, opened 2023-05-21
btanks: fix build issues related to lua.

OscarL:
Ensure we're pulling lua 5.3 (wasn't the case for some reason on my machine).


games-action/btanks/btanks-0.9.8083.recipe
```diff
@@ -32,7 +32,7 @@ REQUIRES="
 	haiku$secondaryArchSuffix
 	lib:libexpat$secondaryArchSuffix
 	lib:libGLU$secondaryArchSuffix
-	lib:liblua$secondaryArchSuffix
+	lib:liblua$secondaryArchSuffix >= 5.3
```
Begasus, line 35: Doesn't hurt, but version string should only be needed for the devel requirement.

Issue #6926, opened 2022-05-15
Latest Cutter crashes  when accessing the Preferences in the menu


Begasus:
Will try to update here when I get around to collect more information.

XVilka:
@Begasus new versions of Rizin and Cutter are out, we fixed a few significant security problems in them:
-   https://repology.org/project/rizin/versions
-   https://github.com/rizinorg/rizin/releases/tag/v0.4.0
-   https://github.com/rizinorg/rizin/releases/tag/v0.4.1
-   https://github.com/rizinorg/cutter/releases/tag/v2.1.0
-   https://github.com/rizinorg/cutter/releases/tag/v2.1.1
-   https://github.com/rizinorg/cutter/releases/tag/v2.1.2

I recommend retrying again after bumping these versions. The amount of changes and fixes is quite significant.

XVilka:
Now it's 0.5.x and 2.2.x:
- https://github.com/rizinorg/rizin/releases/tag/v0.5.0
- https://github.com/rizinorg/rizin/releases/tag/v0.5.1
- https://github.com/rizinorg/cutter/releases/tag/v2.2.0

Begasus:
Crash still around with Cutter v2.2.1, not sure if it's a Haiku thing or something in our port that needs fixing.
Maybe mention this upstream (not sure if I already done that) @XVilka ?

XVilka:
@Begasus please share the stacktrace

Begasus:
Will try to grab some information on the crash. Rebuild with local changes in progress (as mentioned in the other post) atm.

Begasus:
Adding the crash report with:  
[FG6PU.txt](https://github.com/haikuports/haikuports/files/11541556/FG6PU.txt)

Begasus:
After switching to Qt6 ... ;)

![Cutter-v2-2-1-3](https://github.com/haikuports/haikuports/assets/16057090/243cb21d-e046-419b-a713-988f1d8d39b9)

PR #8713, opened 2023-05-23
cutter, bump versin, use older FindRizin.cmake to fix detection for r…

Begasus:
…z_core ...


Begasus:
Crash for opening the preferences in the menu still present: https://github.com/haikuports/haikuports/issues/6926

PR #8717, opened 2023-05-23
xcairo: fixed build

kenmays:
added gtkdocize.


kenmays:
Yes, we currently pack the generated HTML docs.

Begasus:
> I guess a more appropriate question is... shouldn't those docs be present only the _devel packages instead of on the base package?

Move that to the issue tracker? IMHO docs shouldn't be in a base package either (other then README/COPYRIGHT and things like those.

PR #8724, opened 2023-05-24
texstudio, bump version, use icon from original source

Begasus:


Begasus:
This pulls in quite some texlive packages, but they should provide enough to have a fully functional suite without worrying about missing parts. If things are still missing please report back here with an issue.

PR #8728, opened 2023-05-25
kdevelop, switch to boost169

Begasus:


Begasus:
@threedeyes made some small changes (some related to install failures on 32bit), some related to boost and python, not sure if you got an update planned for kdevelop, if so this can be closed.

PR #8741, opened 2023-05-27
xorgproto, bump version

Begasus:


Begasus:
@waddlesplash this is also used by xlibe, could you check?

x11-base/xorg-proto/xorgproto-2022.2.recipe
```diff
@@ -120,4 +119,13 @@ BUILD()
...
 	make install
+
+	# licenses
+	mkdir -p $dataDir/licenses/xorgproto
+	cp COPYING-* $dataDir/licenses/xorgproto
+
+	# cleanup
+	rm -f $dataDir/licenses/xorgproto/COPYING-{applewmproto,windowswmproto}
+	rm -f $dataDir/pkgconfig/applewmproto.pc
+	rm -f $includeDir/X11/extensions/apple*
```
waddlesplash, line 130: These shouldn't be removed.

Begasus: I added this after checking some recipes over at Repology.
Will remove it, thanks for checking.

Begasus: Fixed.

waddlesplash: Which recipes over at Repology? Does Gentoo keep or remove them?

Begasus: Gentoo keeps them, Arch removes them, Fedora removes them ...  :) (so it's a bit of a mix)

https://git.alpinelinux.org/aports/tree/main/xorgproto/APKBUILD
https://gitlab.archlinux.org/archlinux/packaging/packages/xorgproto/-/blob/main/PKGBUILD
https://src.fedoraproject.org/rpms/xorg-x11-proto-devel/blob/rawhide/f/xorg-x11-proto-devel.spec
https://gitweb.gentoo.org/repo/gentoo.git/tree/x11-base/xorg-proto/xorg-proto-2022.2.ebuild

waddlesplash: Let's stick with what Gentoo does.

waddlesplash:
looks ok.

PR #8745, opened 2023-05-28
scdoc: new recipe

hoanga:
the following recipe adds support for [scdoc](https://git.sr.ht/~sircmpwn/scdoc/)

### DIFF
--- a/app-text/scdoc/scdoc-1.11.2.recipe
+++ b/app-text/scdoc/scdoc-1.11.2.recipe
@@ -35,7 +35,5 @@ INSTALL()
 {
 	mkdir -p $manDir
 
-	make install PREFIX=$prefix
-	mv $prefix/share/man/* $manDir
-	rm -rf $prefix/share/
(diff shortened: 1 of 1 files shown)


app-text/scdoc/scdoc-1.11.2.recipe
```diff
@@ -0,0 +1,41 @@
...
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $manDir
+
+	make install PREFIX=$prefix
+	mv $prefix/share/man/* $manDir
+	rm -rf $prefix/share/
```
Begasus, line 40: From the Makefile:
```
PREFIX?=/usr/local
BINDIR?=$(PREFIX)/bin
MANDIR?=$(PREFIX)/share/man
PCDIR?=$(PREFIX)/share/pkgconfig
```

You can use these variables here, eg MANDIR=mandir ... saves the cleanup in the end

hoanga: thanks for the pointers! updated!

PR #8752, opened 2023-05-28
grpc: switch to Python 3.9, drop Python 2.x, fix provides/versions.

OscarL:
* Added a missing PROVIDES, and a couple to PROVIDES_devel (both match now).
* Fixed libVersion/libCppVersion/libVersionCppCompat variables.
~* Use -DgRPC_BUILD_TESTS=OFF, as it makes the build require far less time/storage.~

### DIFF
--- a/net-libs/grpc/grpc-1.53.0.recipe
+++ b/net-libs/grpc/grpc-1.53.0.recipe
@@ -6,7 +6,7 @@ connected systems."
 HOMEPAGE="https://grpc.io/"
 COPYRIGHT="2015-2023 The gRPC Authors"
 LICENSE="Apache v2"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/grpc/grpc/archive/refs/tags/v$portVersion.tar.gz"
 CHECKSUM_SHA256="9717ffc52120861136e478155c2ff3a9c21740e2244de52fa966f376d7471adf"
 SOURCE_FILENAME="grpc-$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


korli:
Please keep gRPC_BUILD_TESTS ON. They need to be run anyway.

net-libs/grpc/grpc-1.53.0.recipe
```diff
@@ -19,17 +19,18 @@ PATCHES="grpc-$portVersion.patchset"
 ARCHITECTURES="all !x86_gcc2"
 SECONDARY_ARCHITECTURES="!x86"
 
-libVersion="29.0.0"
+libVersion="30.0.0"
```
Begasus, line 22: This should be:
```diff
-libVersion="30.0.0"
+libVersion="30.1.0"
```

net-libs/grpc/grpc-1.53.0.recipe
```diff
@@ -19,17 +19,18 @@ PATCHES="grpc-$portVersion.patchset"
...
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
-libCppVersion="1.53"
-libCppVersionCompat="$libCppVersion compat >= ${libCppVersion%%.*}"
+libCppVersion="1.53.0"
```
Begasus, line 24: This probably can be done as (saves changes on next update):
```diff
-libCppVersion="1.53.0"
+libCppVersion="$portVersion"
```

Begasus (approved):

korli:
Any rebuilds needed? Bear?

Begasus:
Could be, needs to be checked once the build completes on the buildmaster, but I guess it should be OK. (only thing depending on this iirc)

PR #8758, opened 2023-05-30
setuptools(_scm), bump version

Begasus:


Begasus:
Needed to fix create the correct version for sphinxcontrib-applehelp (upcoming new recipe).

korli:
Builds with 3.8?

Begasus:
> Builds with 3.8?

Checked with babel which is still using python3.8, no problems there.

korli:
> > Builds with 3.8?
> 
> Checked with babel which is still using python3.8, no problems there.
> 
> **EDIT** babel needs a rebuild anyway for cldr (current babel misses `locale-data` in the package, required for creating tex files)

OK

> **EDIT 2** should we drop python3.8? (this one is still used quite a lot by other recipes using python3.8 I guess)

 maybe add python 3.11 at the same time? :)

Begasus:
> maybe add python 3.11 at the same time? :)

I would if we had it. :)

dev-python/setuptools_scm/setuptools_scm-7.1.0.recipe
```diff
@@ -25,27 +25,30 @@ BUILD_REQUIRES="
...
+		setuptools_$pythonPackage
+		"
+	BUILD_PREREQUIRES="$BUILD_PREREQUIRES
+		cmd:python$pythonVersion
+		"
 done
 
 INSTALL()
 {
-	export SETUPTOOLS_SCM_PRETEND_VERSION="$portVersion"
```
Begasus, line 48: This doesn't do or fix anything regarding depending recipes, they should need this variable.

dev-python/setuptools/setuptools-67.8.0.recipe
```diff
@@ -70,14 +54,13 @@ INSTALL()
 		installLocation=$prefix/lib/$python/vendor-packages/
 		export PYTHONPATH=$installLocation:$PYTHONPATH
 		mkdir -p $installLocation
-
-		mv build$pythonVersion build
+		rm -rf build
 		export SETUPTOOLS_USE_DISTUTILS=local
-		$python setup.py install \
+#		$python setup.py egg_info
```
Begasus, line 59: Think this can be removed, the outcome is the same with or without.

Begasus: Removed ...

Begasus: Needs checking then :)

Begasus: Switched to pythonhosted for the file, the egg file's version looks correct now, doesn't need the line though.
It also fixes some other packages in the pipeline where the egg into isn't really correct (using pythonhosted link).
Thanks!

PR #8759, opened 2023-05-30
unicode-cldr, bump version

Begasus:


Begasus:
This is used by babel (python), but not detected? (even previous version)

PR #8766, opened 2023-05-31
u_boot: switch to Python 3.9, add missing PATCHES.

OscarL:
* ~64 bits: no package provides: cmd:arm_none_eabi_gcc.~ builds OK now.
* 32 bits: errors related to:
  * "runtime_loader: Cannot open file libmpfr.so.4"
  * "Kconfig:66: syntax error" and " Kconfig:65: invalid option".

~Marking as broken until those get resolved.~


pulkomandy:
Well I can have a look at at least fixing arm-none-eabi-gcc for 64bit and see what happens...

pulkomandy:
> Seems like some arm_none_eabi_* packages might need a rev-bump to trigger a rebuild after the last change to arm_none_eabi_binutils?

Yes, I expected that haikuporter/buildmaster would find automatically that previously broken packages can now be built, but it seems that's not the case. I don't know if enabling the uboot recipe will fix that (forcing to retry the dependencies) or if all packages need a revision bump to force a rebuild.

pulkomandy:
Yes, let's try it and see what happens!

PR #8770, opened 2023-05-31
noteshrink: switch to Python 3.10, re-enable recipe.

OscarL:
Needs #10706 merged first.

### DIFF
--- a/media-gfx/noteshrink/noteshrink-0.1.1.recipe
+++ b/media-gfx/noteshrink/noteshrink-0.1.1.recipe
@@ -1,14 +1,13 @@
 SUMMARY="Convert scans of handwritten notes to beautiful, compact PDFs"
 DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes to compact PDFs."
-HOMEPAGE="https://github.com/mzucker/noteshrink"
+HOMEPAGE="https://mzucker.github.io/2016/09/20/noteshrink.html"
 COPYRIGHT="2016 Matt Zucker"
 LICENSE="MIT"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/mzucker/noteshrink/archive/$portVersion.tar.gz"
 CHECKSUM_SHA256="2a29c09768973e688b693b61337c6c49384e8123cf88824dba335cc8c4ed2ca8"
 
-ARCHITECTURES="!any"
-# Doesn't work on gcc2hybrid - missing dependencies.
+ARCHITECTURES="any"
 
 PROVIDES="
 	noteshrink = $portVersion
@@ -16,15 +15,38 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku
-	imagemagick
-	numpy_python >= 1.10
-	pillow_python
-	scipy_python
-	setuptools_python
+	numpy_python310
+	pillow_python310
+	scipy_python310
+	# These are optional:
+	# cmd:pngcrush
+	# cmd:pngquant
+	# cmd:optipng
+	cmd:convert # Deprecated in IMv7. imagemagick wants to be called "magick" now.
+	cmd:python3
+	"
+
+TEST_REQUIRES="
+	numpy_python310
+	pillow_python310
+	scipy_python310
+	cmd:convert
+	cmd:make
+	cmd:optipng
 	"
 
 INSTALL()
 {
 	mkdir -p $binDir
-	cp noteshrink.py $binDir/noteshrink
+	sed 's|#!/usr/bin/env python|#!/bin/python3|' noteshrink.py > $binDir/noteshrink
+}
+
+
+TEST()
+{
+	# As tests, let it convert the examples. Should create the following PDFs:
+	# "graph-paper-ink-only.pdf", "notesA.pdf", "notesB.pdf", and "tree.pdf",
+	# along with some left-over ".png"s.
+	rm -f example_output/*
+	make
 }


media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -3,28 +3,55 @@ DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes to
...
 
-ARCHITECTURES="!any"
-# Doesn't work on gcc2hybrid - missing dependencies.
+ARCHITECTURES="any"
+# The package is "any" because the result is only a Python script under /bin/.
+# However, on gcc2 systems, it will depend on Python modules only available for _x86.
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	hostArchSuffix="_x86"
+else
+	hostArchSuffix="$secondaryArchSuffix"
```
jmairboeck, line 16: when not using SECONDARY_ARCHITECTURES, $secondaryArchSuffix is always empty, so this could also be

```suggestion
	hostArchSuffix=""
```

jmairboeck: On the other hand, this is not good anyway for potentially reusing any-arch packages (see https://github.com/haikuports/haikuporter/issues/234).

Better would be if numpy, pillow and scipy added a suffix-less PROVIDE for x86_gcc2.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -3,28 +3,55 @@ DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes to
...
 	"
 REQUIRES="
-	haiku
-	imagemagick
-	numpy_python >= 1.10
-	pillow_python
-	scipy_python
-	setuptools_python
+	haiku${hostArchSuffix}
+	imagemagick${hostArchSuffix}
```
jmairboeck, line 25:
```suggestion
	cmd:convert
```
(probably, ImageMagick also provides other commands)

media-gfx/noteshrink/noteshrink-0.1.1.recipe
jmairboeck: It is a good thing that this is an "any" package, I'm not disputing that. Actually, anything that can be "any" should be.

But when using "any", it should be possible that, for example, you build the package on x86_64, take the resulting hpkg file and install it on x86_gcc2 (or riscv64 even), and it should work, or the other way round.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
korli: cmd:convert is definitely preferred.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
korli: Indeed hostArchSuffix should be dropped, if REQUIRES can be made arch agnostic.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -3,28 +3,55 @@ DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes to
...
 	cmd:noteshrink
 	"
 REQUIRES="
-	haiku
-	imagemagick
-	numpy_python >= 1.10
-	pillow_python
-	scipy_python
-	setuptools_python
+	haiku${hostArchSuffix}
```
korli, line 24: haiku is probably not needed.

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -3,28 +3,55 @@ DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes to
...
+	numpy${hostArchSuffix}_python39
```
korli, line 25: numpy_x86_python39 should also provide numpy_python39

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -3,28 +3,54 @@ DESCRIPTION="Noteshrink is an utility that is used to convert scans of notes to
...
-ARCHITECTURES="!any"
-# Doesn't work on gcc2hybrid - missing dependencies.
+ARCHITECTURES="any"
+# The package is "any" because the result is only a Python script under /bin/.
+# However, on gcc2 systems, it will depend on Python modules only available for _x86.
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	hostArchSuffix="_x86"
+else
+	hostArchSuffix=""
+fi
```
Begasus, line 17: Why insist on keeping this any? I don't think the 32bit package will run on 64bit (or does it)?

Begasus: So basically it can't be `any` then?

jmairboeck: As long as pillow and scipy aren't fixed, it can't be `any`. But they should be fixed, similar to numpy.

Begasus: Library dependencies will be hard to fix for pillow and numpy (for gcc2), so I guess we're stuck on secondary architecture for 32bit.

jmairboeck: No, they should just add a block like this to the `eval` block:

```sh
if [ $targetArchitecture = x86_gcc2 ]; then\
   PROVIDES_${pythonPackage}+=\"\n\
       scipy_${pythonPackage} = $portVersion\n\
       \"\n\
fi\n\
```

Begasus: I don't think even that will make it install-able on 64bit? As nothing will provide `lib:liblapack$secondaryArchSuffix` (numpy) for instance?

jmairboeck: It will still depend on scipy_x86_python39 on 32 bit, but that package will provide both `scipy_x86_python39` and `scipy_python39`, so that the `x86` can be dropped from the requirements of `any` packages.

Begasus: OK, starts to make sense, just wanted to grasp the changes, at least at it's current state it doesn't work. :)

Begasus:
@OscarL any progress on this?

media-gfx/noteshrink/noteshrink-0.1.1.recipe
```diff
@@ -17,14 +16,38 @@ PROVIDES="
...
+	pillow_python310
+	scipy_python310
+	cmd:make
+	cmd:optipng
 	"
 
 INSTALL()
 {
 	mkdir -p $binDir
+	sed -i 's|#!/usr/bin/env python|#!/bin/python3|' noteshrink.py
```
korli, line 41: This should be combined with the copy

korli:
LGTM

Begasus:
@OscarL I guess this will fail when merged because of scipy?

Begasus:
If nothing else open an issue to track the issue here. Thanks!

PR #8817, opened 2023-06-07
pyopenssl: switch to "any" arch, add Python 3.10

OscarL:
I guess this could use REPLACES, at least for 32 bits, but I didn't find anything in-tree actually requiring `pyopenssl`, so I think we can do without it.


dev-python/pyopenssl/pyopenssl-22.0.0.recipe
```diff
@@ -10,42 +10,43 @@ HOMEPAGE="https://github.com/pyca/pyopenssl
 COPYRIGHT="2008-2022 The pyOpenSSL developers"
 LICENSE="Apache v2
 	BSD (3-clause)"
-REVISION="3"
+REVISION="5"
```
Begasus, line 13: 5?

PR #8851, opened 2023-06-11
meson_python, new python recipe

Begasus:


dev-python/meson-python/meson_python-0.13.1.recipe
```diff
@@ -0,0 +1,75 @@
+SUMMARY="Short description of ProjectX"
```
jmairboeck, line 1: This looks wrong ;-)

dev-python/meson-python/meson_python-0.13.1.recipe
```diff
@@ -0,0 +1,75 @@
...
+DESCRIPTION="``meson-python`` is a Python build backend built on top of the Meson__
+build system. It enables to use Meson for the configuration and build
+steps of Python packages. Meson is an open source build system meant
+to be both extremely fast, and, even more importantly, as user
+friendly as possible. ``meson-python`` is best suited for building
+Python packages containing extension modules implemented in languages
+such as C, C++, Cython, Fortran, Pythran, or Rust. Consult the
+documentation__ for more details."
```
jmairboeck, line 9: Is Markdown formatting supported in the description? Otherwise, this should be cleaned up.

dev-python/meson-python/meson_python-0.13.1.recipe
Begasus: Yeah, was a bit too fast on this one I guess (same for the DESCRIPTION), will have a look again later. Thanks!

Begasus:
IIRC this was one of the reasons to include it (checking for bumping scipy), didn't want it to rot in my directory so added just in case it might be usefull, haven't found anything further that uses this package though. :)

eli-schwartz:
> Might be good to add TEST() for this one, as I'm running into some issues when trying to use meson-python to build a modern version of scipy (complains about gfortran not producing executable binaries, also has issues fiding `/bin/cython` or `/bin/cython3` as none of our packages provides that. had to add a symlink under `~/config/non-packaged/bin` on my [WIP recipe for scipy](https://github.com/haikuports/haikuports/compare/master...OscarL:haikuports:scipy-1.10.1)).

(Meson will just skip those tests in its internal testsuite, if it cannot find `cython`.)

Note that the name `cython3` is non-canonical and used exclusively on Debian because Debian broke it by default.

PR #8868, opened 2023-06-18
yt-dlp: switch to nightly builds

humdingerb:
The last 'official' release was in March 2023. Development of yt-dlp is still very busy, but they switched to providing nightly builds [1], while possibly also doing 'stable' releases from time to time in the future.

I think we should switch to their nightly builds as well, as long as they appear to be stable, and graba new one every month or so.

[1] https://github.com/yt-dlp/yt-dlp/issues/7175#issuecomment-1569442827


humdingerb:
Hmm, it may be that the version number in the recipe name isn't quite right.
I've installed yt_dlp-2023.06.17.224306 locally and SoftwareUpdater offers to update to the older 2023.03.04...

What's correct portVersion here? **2023.06.17~22430**6 ?

pulkomandy:
2023.06.* is clearly larger than 2023.03.* so there should be no problem here.

But SoftwareUpdater always does a full-sync and not an update, so it will downgrade local packages to an older version that's in the repository.

Begasus:
SoftwareUpdater doesn't know about the "new" version you got localy and will install the version it does know.
Merging this should update the repositories and make this available and removes the older one from being installed.

PR #8875, opened 2023-06-20
lua, bump lua5.4, make it default

Begasus:


Begasus:
lua5.4 uses `-DLUA_COMPAT_5_3` during build, I've set this as `compat >= 5.3` in the recipe for lua5.4, not changing that for lua5.3 (uses `-DLUA_COMPAT_5_2`) to break some dependencies there.

korli:
> lua5.4 uses `-DLUA_COMPAT_5_3` during build, I've set this as `compat >= 5.3` in the recipe for lua5.4, not changing that for lua5.3 (uses `-DLUA_COMPAT_5_2`) to break some dependencies there.

This is a bad idea, with this, one can't get the desired 5.4 as a requirement.

Begasus:
> This is a bad idea, with this, one can't get the desired 5.4 as a requirement.

Reverted

PR #8876, opened 2023-06-20
iaito, new recipe

Begasus:


Begasus:
Translations not included (crashes iaito when installed), needs to be started from Terminal, crashes when launched from Tracker/Deskbar (patches that are around here didn't fix the crash), reported upstream.

PR #8880, opened 2023-06-22
brltty, remove obsolete patch

Begasus:
Looking at the latest sources the patch has been upstreamed/integrated.


Begasus:
Created recipes for brltty and liblouis at: https://github.com/Begasus/haikuports/commit/de7f963e55628f59b09fae774343e58f0ad961ce
Not pushing as brltty isn't working as expected (no real support), just in case someone might want to pick them up.

PR #8883, opened 2023-06-23
abcl: bump version

hoanga:
this updates abcl to the latest release

note that an updated jdk (17+) is recommended per release notes

```
> abcl
Armed Bear Common Lisp 1.9.2
Java 17.0.7 N/A
OpenJDK 64-Bit Server VM
Low-level initialization completed in 0.308 seconds.
Startup completed in 1.714 seconds.
Type ":help" for a list of available commands.
CL-USER(1):
```


dev-java/abcl/abcl-1.9.2.recipe
```diff
@@ -0,0 +1,51 @@
...
+BUILD_REQUIRES="
+	"
+BUILD_PREREQUIRES="
+	java:environment == 17
+	"
+
+INSTALL()
+{
+	source /system/data/profile.d/openjdk17.sh
+	export JAVA_HOME=$JDK17_HOME
```
korli, line 36: Why is this needed?

dev-java/abcl/abcl-1.9.2.recipe
```diff
@@ -0,0 +1,51 @@
...
+	"
+REQUIRES="
+	haiku
+	java:environment >= 17
+	"
+
+BUILD_REQUIRES="
+	"
+BUILD_PREREQUIRES="
+	java:environment == 17
```
korli, line 30: Why is this needed?

Begasus:
Please also remove the old version while you're in the process.

dev-java/abcl/abcl-1.9.2.recipe
hoanga: i am not entirely sure. it came from the earlier port definitions where i might have ran into some circular conditions that don't appear to exist anymore.  have removed this. thanks!

dev-java/abcl/abcl-1.9.2.recipe
hoanga: similar to above and have had it removed

PR #8920, opened 2023-07-02
foobillardplus: fix build and add data directory workaround

### LINKED ISSUE #8891: FooBillard++ not launcching from Tracker/Deskbar
foobillardplus 3.42.0-2 haikuports-x86_64 is trying to launch but milliseconds after I see window borders, it crashes. I have installed some packages perhaps necessary to run it, SDL and libpng. Haiku OS current stable R4_Beta on Dell Latitude e7250 RAM 16GB

augiedoggie:
fixes #8891


augiedoggie:
> Is that "chrdir() to executable dir" behaviour on SDL2 documented anywhere? Seems that was different on SDL 1.2, at least when reading [this patch](https://github.com/haikuports/haikuports/blob/1cabecdf5b9ae80408bf3cebd9337fec49f524ef/games-arcade/alienblaster/patches/alienblaster-1.1.0.patch#L16) that calls `chdir()` _before_ `SDL_Init()`.
> 
> I've tried searching for where SDL documents this but failed miserably :-/

This behavior is specific to Haiku.  I'm not aware of any documentation about it other than reading the source code or the [patch file](https://github.com/haikuports/haikuports/blob/1cabecdf5b9ae80408bf3cebd9337fec49f524ef/media-libs/libsdl2/patches/libsdl2-2.26.4.patchset#L167)

augiedoggie:
Looks like it tries to `chdir` in a couple of [places](https://github.com/haikuports/haikuports/blob/1cabecdf5b9ae80408bf3cebd9337fec49f524ef/media-libs/libsdl2/patches/libsdl2-2.26.4.patchset#L455). This explains why I could also get the game to fail when I started it with `env -i`

PR #8925, opened 2023-07-02
fife, bump to current commit

Begasus:


Begasus:
unknown-horizons still crashes with this one (launced from source)
atm openGL  isn't disabled, doing some work there to see if it can be disabled and what it would bring in terms of launching unkown-horizons.

Begasus:
If someone wants to tackle it, keep this in mind (from AUR)
```
    # If compilation seems to stall at 99% for ages, do not abort!
    # While not perfectly convenient, that is to be expected from
    # SWIG. Eventually, you'll reach the holy triple-digit land.
```

PR #8933, opened 2023-07-03
diffutils, also provide non suffixed 32bit package

Begasus:


jmairboeck:
Why is this needed? Shouldn't other packages depend on the provided commands (e.g. `cmd:diff`) instead of the package name?

Even in texlive, for example, I used `cmd:kpsewhich` as a "placeholder" for `texlive_core` exactly for this reason.

korli:
Writing a reason for changes would indeed be nice. :)

Begasus:
Guess you've all read the IRC logs, with the fix done in previous revision automake and autoconf were broken for 32bit (first plan was to fix those 2 but then I saw autoconf also used coreutils and didn't complain about that one.
Looking into coreutils recipe I used the workaround used there to make this switch for diffutils.

@korli looks like you did that change for coreutils :) https://github.com/haikuports/haikuports/commit/3868d4b6e8a0996baf2ea5323319ace2584fdf12#diff-4916146522e55e1df576b3b14f6774f5a64247db7726a9dd988e7f5fa6124faaR145

korli:
I added a REPLACES, not a PROVIDES.

jmairboeck:
@korli you did add the PROVIDES also, in a later commit: https://github.com/haikuports/haikuports/commit/917fe03dc78731c71fe773a67e717ec242a50bb5#diff-4916146522e55e1df576b3b14f6774f5a64247db7726a9dd988e7f5fa6124faaR146

But I think this could still be cleaned up and replaced with dependencies on concrete commands instead.

Note that for python recipes, the situation is different because they don't usually provide any commands to depend on, so there is no other way to do that.

Begasus:
Whoops, my bad! :/ Will revert and rework the ones for automake and autoconf then?
That would mean stepping away from gcc2 for both.

jmairboeck:
> That would mean stepping away from gcc2 for both.
No, their REQUIRES can just be changed to `cmd:diff` instead, for example.

Begasus:
OK, will look into this, need to boot my 32bit machine for this

jmairboeck:
I created an issue to track this: #8969

Begasus:
Rebuild diffutils, also autoconf and automake, looks it's not a hard requirement for diffutils on both, no policy error when they are not listed in REQUIRES (diffutils nor coreutils).

Begasus:
PR to revert the changes: https://github.com/haikuports/haikuports/pull/8970

PR #8962, opened 2023-07-04
rtaudio: add recipe

hfsfox:
No mediakit backend at the moment, dummy output only

### DIFF
--- a/media-libs/rtaudio/patches/rtaudio-5.2.0.patchset
+++ b/media-libs/rtaudio/patches/rtaudio-5.2.0.patchset
@@ -1,0 +1,79 @@
+From dd9aa2e5fe6aeba2217439fa8d5e4de9f32ff0e7 Mon Sep 17 00:00:00 2001
+From: hfsfox <darkkitsunezx128k2309@gmail.com>
+Date: Thu, 26 Jan 2023 17:38:20 +0200
+Subject: Enable build on Haiku
+
+
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 21da42a..8b427b6 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -24,7 +24,7 @@ pkg_check_modules(pulse libpulse-simple)
+ if (CMAKE_SYSTEM_NAME MATCHES "kNetBSD.*|NetBSD.*")
+   message(STATUS "NetBSD detected, using OSS")
+   set(xBSD ON)
+-elseif(UNIX AND NOT APPLE)
++elseif(UNIX AND NOT APPLE AND NOT HAIKU)
+   set(LINUX ON)
+ endif()
+ 
+diff --git a/RtAudio.cpp b/RtAudio.cpp
+index 40fb6a9..0fa1e80 100644
+--- a/RtAudio.cpp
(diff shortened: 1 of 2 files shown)


korli:
Should be disabled if it doesn't work.

hfsfox:
@korli it is required to build other applications such as bambootracker

korli:
@hfsfox will bambootracker be usable without sound output?

hfsfox:
@korli yes, because midi via rtmidi works fine, and rtaudio is the only thing required for build now.

diversys:
> @korli yes, because midi via rtmidi works fine, and rtaudio is the only thing required for build now.

Makes sense. @korli seems like it's good to go?

media-libs/rtaudio/patches/rtaudio-5.2.0.patchset
```diff
@@ -0,0 +1,79 @@
...
+   list(APPEND LINKLIBS Threads::Threads)
+ endif()
+ 
+@@ -315,7 +319,11 @@ install(FILES RtAudio.h rtaudio_c.h
+ export(PACKAGE RtAudio)
+ 
+ # Set installation path for CMake files.
+-set(RTAUDIO_CMAKE_DESTINATION share/rtaudio)
++if (HAIKU)
++    set(RTAUDIO_CMAKE_DESTINATION lib/cmake/rtaudio)
```
Begasus, line 70: Maybe (haven't check, but cmake should end up in lib/x86 on 32bit)?
```diff
-+    set(RTAUDIO_CMAKE_DESTINATION lib/cmake/rtaudio)
++    set(RTAUDIO_CMAKE_DESTINATION ${CMAKE_INSTALL_LIBDIR}/cmake/rtaudio)
```

korli:
> > @korli yes, because midi via rtmidi works fine, and rtaudio is the only thing required for build now.
> 
> Makes sense. @korli seems like it's good to go?

IMO rtaudio shouldn't be required, if the software is usable without it.

Begasus:
https://github.com/BambooTracker/BambooTracker/releases/download/v0.6.2/BambooTracker-src-v0.6.2.tar.gz includes the sources for rtaudio and rtmidi, system rtmidi can be used as a cmake variable.
From upstream:
```
RtAudio / RtMidi: Use a precompiled RtAudio / RtMidi installation instead of our vendored one.
(Obsoletes the optional dependencies above)
Requires RtAudio (5.1.0 or higher) and/or RtMidi (4.0.0 or higher) + pkg-config.
```
So I'm guessing it's not a hard requirement (can build it without rtaudio).
Could you check this forr BoomTracker @hfsfox ?

Begasus:
BambooTracker builds fine with: 

`cmake -Bbuild -S. -DCMAKE_INSTALL_PREFIX=/boot/home/destdir/ -DRTMIDI_API_ALSA=OFF`

Can't seem to do anything else with it though. Doesn't even open the demo files.

Begasus:
@hfsfox could you mark the arch's as "?", when enabled it would lead to false ideas thinking this is fully functional and will flood the forum probably with question as why there is no output. :)
Once done this could be merged and one that needs it can build it themselves to go from here.

Begasus:
Merging disabled for now.

PR #8970, opened 2023-07-05
diffutils, revert previous commit to enable suffix-less PROVIDES

Begasus:


Begasus:
Thanks for the merge, this however means that autoconf and automake should be fixed, if anything else is needed please comment on the open PR's, this way on an update it will remove the packages for autoconf, automake and libtool.

PR #8971, opened 2023-07-05
coreutils, don't provide suffix-less PROVIDES, make depending package…

Begasus:
…s rely on cmd:* from the package


Begasus:
Note, this will have a huger impact then the one for diffutils ...
Quite a lot of `BUILD_PREREQUIRES="coreutils"` them out there

Begasus:
Looks like on 32bit Haiku depends on the non suffixed package.

korli:
Isn't coreutils a hard dependency of haikuporter?

Begasus:
When I wanted to install the suffix-less package it wanted to bring down the whole system, so my guess is that the base gcc2 hrev depends on it?

Begasus:
Then this can't be applied as long as it isn't fixed there, I guess coreutils can be replaced like the ones currently mentioned in the PR's for autoconf and automake.

PR #8975, opened 2023-07-06
hivelytracker, revbump for rebuild

Begasus:


Begasus:
Needs a revbump after the removal of non existing lib:libsdl (fixed in the meantime)

PR #8982, opened 2023-07-06
autoconf, cleanup, change dependency for diffutils/coreutils

Begasus:


Begasus:
Related PR for diffutils: https://github.com/haikuports/haikuports/pull/8970

Begasus:
Removed secondary architecture support for autoconf213 (it is the only one containing support for both compilers on 32bit)

Begasus:
Dependencies added as declared here: https://www.linuxfromscratch.org/museum/lfs-museum/3.2/LFS-BOOK-3.2-HTML/chapter06/autoconf.html

PR #8983, opened 2023-07-06
automake, cleanup, change  dependency for coreutils/diffutils

Begasus:


Begasus:
Related PR for diffutils: https://github.com/haikuports/haikuports/pull/8970
Removed obsolete recipes.

Begasus:
A bit older version, but added dependencies as declared here: https://www.linuxfromscratch.org/museum/lfs-museum/3.2/LFS-BOOK-3.2-HTML/chapter06/automake.html

korli:
Now I see it, maybe the dependency on coreutils and diffutils was for bootstrapping.

korli:
@Begasus basic coreutils would be some commands with no other requirements I think. As you said we could add another "coreutils" provides to group everything needed for bootstrapping, if ever needed.

Begasus:
> As you said we could add another "coreutils" provides to group everything needed for bootstrapping, if ever needed.

I'll leave that up to you guys :) For now I think most issues should be resolved haikuports related.

Issue #7248, opened 2022-10-03
Update minisign 0.9 -> 0.10


applecuckoo:
Mainly realised this since minisign is used by the Haiku project itself to verify Haiku images. Not an expert at this so maybe @waddlesplash could help? Here's the changelog:

> - Minisign can be compiled with Zig instead of cmake+make+a C toolchain
> - Minimal `VERIFY_ONLY` versions can be built again
> - Prehashing is now enabled by default, regardless of the input size. Support for non-prehashed signatures will eventually be removed
> - Legacy signatures can be rejected with the addition of the `-H` flag
> - Precompiled binaries for Win64, Linux and WebAssembly are now available.

waddlesplash:
I don't really know much about our use of minisign, @kallisti5 set that up.

Begasus:
0.11 still builds fine, broken on 32bit secondary architecture (installs in $prefix/bin), so do we really want to have both architectures for 32bit as primary builds fine (there is only 1 binary in the package)?

Begasus:
Fixed in: https://github.com/haikuports/haikuports/pull/9042

PR #9042, opened 2023-07-16
minisign, bump version

Begasus:


Begasus:
Fixes https://github.com/haikuports/haikuports/issues/7248
Not sure if we need the binary for primary and secondary architecture on 32bit?

Begasus:
Would have to boot up 32bit version to verify the build (has been a while since, maybe there is a newer version already upstream) :)
If this doesn't involve Haiku's build/check process I guess it would be OK to merge then.

Begasus:
> As this only provides a `cmd:`... I'd say just move it to x86 and drop the GCC2 one. [No mention of minisign](https://xref.landonf.org/source/search?q=minisign&defs=&refs=&path=&hist=&type=&project=haiku) on Haiku's sources, only [mentioned on Haiku's infrastructure repo](https://github.com/search?q=repo%3Ahaiku%2Finfrastructure+minisign&type=code), and that's using a linux docker image.

Done :)

PR #9049, opened 2023-07-18
avisynthplus: update to v3.7.3

qyot27:


media-libs/avisynthplus/avisynthplus-3.7.3.recipe
```diff
@@ -43,6 +43,7 @@ BUILD_PREREQUIRES="
 	cmd:cmake
 	cmd:gcc$secondaryArchSuffix
 	cmd:make
+	cmd:git
```
korli, line 46: shouldn't be needed

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
```diff
@@ -57,7 +58,6 @@ INSTALL()
 {
 	make -C build install
 
-	prepareInstalledDevelLib libavisynth
```
korli, line 60: Why is it removed?

qyot27: See the discussion when the recipe was first added, https://github.com/haikuports/haikuports/pull/5614#issuecomment-762809293

prepareInstalledDevelLib breaks client programs because most of them interface with AviSynth through dlopen.  Most prominently:
[FFmpeg](http://git.videolan.org/?p=ffmpeg.git;a=blob;f=libavformat/avisynth.c;h=b426ac343e67e6357c39cbd15b326be1d53add68;hb=2cd09fec9f8d3b16d03e762dec6001cc0b792659#l35)
[x264](https://code.videolan.org/videolan/x264/-/blob/master/input/avs.c#L29)
[avs2yuv](https://github.com/DJATOM/avs2yuv/blob/master/avs2yuv.c#L20)

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
qyot27: Without it, the build fails:

```
waiting for build package avisynthplus-3.7.3-1 to be activated
waiting for build package avisynthplus_devel-3.7.3-1 to be activated
Building ...
-- The CXX compiler identification is GNU 11.2.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Detected target processor as: x86_64
-- The C compiler identification is GNU 11.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Could NOT find Git (missing: GIT_EXECUTABLE) 
-- Found DevIL: /boot/system/lib/libIL.so  
Found DevIL library: /boot/system/lib/libIL.so
-- Configuring done
-- Generating done
-- Build files have been written to: /sources/AviSynthPlus-3.7.3/build
make: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[1]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Leaving directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Leaving directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Leaving directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Leaving directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Leaving directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
make[2]: Entering directory '/sources/AviSynthPlus-3.7.3/build'
CMake Error at /sources/AviSynthPlus-3.7.3/avs_core/Version.cmake:19 (string):
  string sub-command STRIP requires two arguments.


avs_core/CMakeFiles/VersionGen.dir/build.make:70: recipe for target 'avs_core/CMakeFiles/VersionGen' failed
make[2]: *** [avs_core/CMakeFiles/VersionGen] Error 1
make[2]: Leaving directory '/sources/AviSynthPlus-3.7.3/build'
CMakeFiles/Makefile2:255: recipe for target 'avs_core/CMakeFiles/VersionGen.dir/all' failed
make[1]: *** [avs_core/CMakeFiles/VersionGen.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
```
I have a feeling I know what's going sideways, and that'll be dealt with upstream.  But is adding the yet-to-be-written patch to resolve that to the recipe better/preferable to just declaring `cmd:git`?

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
pulkomandy: Maybe this deserves a comment in the recipe, otherwise we will keep adding it at every update. I suspect doing it this way results in some warnings from haikuporter strict mode?

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
pulkomandy: Maybe include a comment in the recipe that git is currently needed due to an upstream bug, and a link to the place where they are fixing it (bugreport, commit message, ...)

Also note that since we build in a generated git repository, there will be none of the tags from upstream and SHA1 of commits will not match, which will be confusing. It seems strange that it's not possible to build this outside of a git tree...

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
korli: prepareInstalledDevelLib is correct, because any client programs not using dlopen has its build broken.
Please add the desired symlink in $libDir manually.

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
korli: `if (EXISTS ${REPO}/.git)` should probably become `if (EXISTS ${REPO}/.git AND GIT_FOUND)`

adding cmd:git doesn't help, see [here](https://github.com/AviSynth/AviSynthPlus/blob/fc5b9bc41fd47001b7da39ea777d29c0ede2a2a7/avs_core/CMakeLists.txt#L205),  RELEASE_TARBALL isn't added anymore. should be `if (NOT EXISTS ${REPO}/.git OR NOT GIT_FOUND)`

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
korli: BTW I see no recipe at Haikuports which depends on lib:libavisynth. How do you test?

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
qyot27: > Also note that since we build in a generated git repository, there will be none of the tags from upstream and SHA1 of commits will not match, which will be confusing. It seems strange that it's not possible to build this outside of a git tree...

All the stuff in Version.cmake that relies on git is supposed to be skipped in the case of RELEASE_TARBALL being active* (and I just tested on Linux by grabbing the tarball and removing git - the configuration and build succeeded, RELEASE_TARBALL was added correctly, and the Version output was correct for that case).  It's actually something of the inverse problem here: because the build is happening in a generated git repository, and the check for whether to use RELEASE_TARBALL is based on the presence of the `.git` directory existing (which in the ordinary instance of grabbing the tarball from Github, it doesn't), the check fails and like @korli pointed out, RELEASE_TARBALL doesn't get added.  What confuses me is why it didn't fail for either of the 3.7.0 or 3.7.1 recipes, assuming that the generated git repository is something haikuporter was doing at that time as well.

*most of it is also supposed to be skipped for regular builds from the release branches: the difference between a normal git-present release build and RELEASE_TARBALL is the first line that appears in the output of Version():
`AviSynth+ 3.7.3 (r4003, 3.7, x86_64)` == git is present, release build was built from the '3.7' branch
`AviSynth+ 3.7.3 (x86_64)` == output of RELEASE_TARBALL
All the other tag and commit hash and date stuff in Version.cmake is for the case of a build from *after* the latest tag, so it can clearly be identified as a development build.  So for really any release build or RELEASE_TARBALL, that wouldn't be present either.

What the check didn't account for was that a `.git` directory would reappear despite the tarball and the chroot wouldn't have git available inside of it.

The fix would be this patch, with the changes @korli mentioned:
https://github.com/qyot27/AviSynthPlus/commit/59fb4c93d3afabde8b0573045082c0a3bd3023d9

I've added the patch and will force-push.

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
qyot27: Testing was done with already existing source builds of FFmpeg and avs2yuv.  avs2yuv doesn't exist as a recipe (it probably should, but that's a separate issue), and FFmpeg can only enable AviSynth support if `--enable-gpl` is used; if there was a decision to have an alternate, ffmpeg_gpl package, then it could go in there.

Of the existing ports, x264 is tethered to commit bfc87b7a330f75f5c9a21e56081e4b20344f139e, [which enables AviSynth+ on Haiku](https://code.videolan.org/videolan/x264/-/commit/bfc87b7a330f75f5c9a21e56081e4b20344f139e).  x264 auto-enables the avs input module since it carries its own copy of the headers with it and doesn't need a _devel package to enable it.

korli: OK I understand. For x264, it means one could add a `SUPPLEMENTS="x264$secondaryArchSuffix"` or `SUPPLEMENTS_bin="x264$secondaryArchSuffix"` in the avisynthplus recipe.

qyot27:
Added `SUPPLEMENTS_bin` (the avs input module is only part of the x264 CLI), squashed everything back together, and force-pushed.

media-libs/avisynthplus/avisynthplus-3.7.3.recipe
korli: thanks!

PR #9053, opened 2023-07-19
pygobject, new recipe

Begasus:


dev-python/pygobject/pygobject-3.44.1.recipe
```diff
@@ -0,0 +1,82 @@
...
+such as GTK, GStreamer, WebKitGTK, GLib, GIO and many more."
+HOMEPAGE="https://wiki.gnome.org/Projects/PyGObject"
+COPYRIGHT="2014 Developer name
+	2016 Another developer's name"
+LICENSE="GNU LGPL v2"
+REVISION="1"
+SOURCE_URI="https://download.gnome.org/sources/pygobject/3.44/pygobject-$portVersion.tar.xz"
+CHECKSUM_SHA256="3c6805d1321be90cc32e648215a562430e0d3d6edcda8f4c5e7a9daffcad5710"
+
+ARCHITECTURES="any"
```
jmairboeck, line 12: Is this really correct? It requires gcc to build, so likely contains some binaries, and it uses `$secondaryArchSuffix` which is not defined for `any` (which is a primary architecture), so this probably doesn't work on x86.

dev-python/pygobject/pygobject-3.44.1.recipe
```diff
@@ -0,0 +1,82 @@
+SUMMARY="Python bindings for GLib/GObject/GIO/GTK"
+DESCRIPTION="PyGObject is a Python package which provides bindings for GObject based libraries \
+such as GTK, GStreamer, WebKitGTK, GLib, GIO and many more."
+HOMEPAGE="https://wiki.gnome.org/Projects/PyGObject"
+COPYRIGHT="2014 Developer name
+	2016 Another developer's name"
```
jmairboeck, line 6: This looks wrong ;)

PR #9054, opened 2023-07-20
pass: Adding recipe

ambroff:
Adds a recipe for 1.7.4. I have been using this for a couple of days and it seems to work fine.


app-admin/pass/pass-1.7.4.recipe
```diff
@@ -0,0 +1,82 @@
+SUMMARY="The standard unix password manager"
+DESCRIPTION="Password management should be simple and follow Unix philosophy. \
+With pass, each password lives inside of a gpg encrypted file whose filename is \
+the title of the website or resource that requires the password. These \
+encrypted files may be organized into meaningful folder hierarchies, copied \
+from computer to computer, and, in general, manipulated using standard command \
+line file management utilities.\
```
diversys, line 7: I guess you don't need a slash at the end here if you want a new line? As well as in a few other places.

PR #9089, opened 2023-07-25
ksystemlog, new recipe

Begasus:


kde-apps/ksystemlog/ksystemlog-22.04.3.recipe
```diff
@@ -0,0 +1,136 @@
+SUMMARY="KDE SystemLog Application"
+DESCRIPTION="Long ProjectX description.
+Spanning several paragraphs, maybe
+ - containing
+ - bullet points etc.
+You can also have really long lines that do not fit in 80 characters and use \
+a backslash to wrap the line."
```
augiedoggie, line 7: Needs a real description.

kde-apps/ksystemlog/ksystemlog-22.04.3.recipe
```diff
@@ -0,0 +1,136 @@
...
+HOMEPAGE="https://apps.kde.org/nl/ksystemlog/"
```
augiedoggie, line 8: It would be better to use the english URL I think. https://apps.kde.org/ksystemlog/

Begasus: Both fixed in https://github.com/haikuports/haikuports/pull/9091 
Thanks for noticing it 👍

PR #9114, opened 2023-07-30
fltk, bump version

Begasus:
Checked with tuxpaint-config, behaviour is the same as previous version.

Scrolling locks on both versions.


Begasus:
Latest tag is from last year, I've gone like some others went seeing at Replology.
There is no crash, I have to shut it down with Ctrl-C in Terminal or with PC in the Systray. Not sure how to catch anything there?

waddlesplash:
I think we should stick with the latest release only.

You can attach Debugger to the running process, mark all threads as "Debugged", then Save Report from the menu.

Begasus:
OK, best I could do for now.
[tuxpaint-config-14233-debug-01-08-2023-10-30-30.report.txt](https://github.com/haikuports/haikuports/files/12226303/tuxpaint-config-14233-debug-01-08-2023-10-30-30.report.txt)

waddlesplash:
It's probably waiting for events that never come. I can try and debug it later.

I still think we shouldn't upgrade to a Git version with no particular reason, though.

Begasus:
Reason behind it was to move the documentation to the devel package.

Begasus:
Closing for now, will wait untill new release is out.

waddlesplash:
Scrolling lock fixed in Xlibe: https://github.com/waddlesplash/xlibe/commit/39a6c0de49647e90822690e1db44606fb96ccb09

PR #9142, opened 2023-08-03
unrar, bump version

Begasus:


Begasus:
Parse errors with gcc2, so marked x86_gcc2 as broken here. Binary is still installed in $prefix/bin.

PR #9146, opened 2023-08-04
tar, revbump for 32bit, use --disable-year2038

Begasus:
Shouldn't be needed for 64bit, but I guess it will be fixed by 2038? :)


korli:
Should only be disabled on x86

Begasus:
> Should only be disabled on x86

Done. :)

PR #9150, opened 2023-08-04
agg: bump version

kenmays:
Bumped to AGG 2.7-r138 (latest official source)
Tested successfully using Haiku hrev57193 x64 w/ GNU gnash 0.8.11-3.

Haiku_gnash_0 8 11-agg_2 7

Note: GNU gnash 0.8.11-3 playing Harley_Davidson.swf video file using AGG 2.7-r138.


Begasus:
There has been some discussions on the forum on AGG, let's wait what comes up there?

pulkomandy:
Version 2.7 of agg does not exist. Please don't invent version numbers.

kenmays:
@Begasus - That discussion is mainly about the kernel's AGG implementation. Haikuport's AGG port is only used by the gnash port.

@pulkomandy- The original source tarball reflects agg-2.7.0 as well as a specific versioning commit from AGG 2.6.0 to 2.7.0. Not a "smoke-in-mirror" invention.

pulkomandy:
I saw that, but there is no tag or release of this anywhere, nor any announcement of a release. So, to me it looks like this was just changed in the development version to differenciate it from the latest release of 2.6.

In that case, we should mark it as such, for example using a version like 2.7~r138 to clearly identify that this is not an official 2.7 release.

That does not prevent using the latest version if we need it, but we really have to track exactly what we are using. A commit changing a version number in a repository is not a release.

kenmays:
Change recipe to reflect 2.7.0~r138.

PR #9153, opened 2023-08-05
agar, bump version, enable recipe

Begasus:


media-libs/agar/agar-1.7.0.recipe
```diff
@@ -119,6 +121,8 @@ INSTALL()
 	sed -i "s|include|develop/headers$secondaryArchSubDir|" $commandBinDir/agar-sg-config
 	sed -i "s|include|develop/headers$secondaryArchSubDir|" $commandBinDir/agar-sk-config
 	sed -i "s|include|develop/headers$secondaryArchSubDir|" $commandBinDir/agar-vg-config
+	# libGL doesn't provide a gl.pc file, remove requirement for it in agar.pc
+	sed -i "s|freetype2  gl  |freetype2  |" $libDir/pkgconfig/agar.pc
```
Begasus, line 125: `pkg-config agar --libs` fails on this if `gl` is included in `agar.pc` (so do the other pkgconfig files as some depend on agar.pc.

PR #9167, opened 2023-08-09
Update toot-0.36.0.recipe

FabianReidinger:
Executing toot in Terminal gives an error at the beginning, that Soupieve is missing. It continued. I installed Soupsieve_python310, executed it again, and the error message at the beginning was gone. Thus, I added the line "soup sieve_...". Hope that works.

### DIFF
--- a/net-im/toot/toot-0.36.0.recipe
+++ b/net-im/toot/toot-0.36.0.recipe
@@ -20,6 +20,7 @@ REQUIRES="
 	haiku
 	cmd:python$pythonVersion
 	beautifulsoup4_$pythonPackage
+	soupsieve_$pythonPackage
 	requests_$pythonPackage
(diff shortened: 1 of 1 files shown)


augiedoggie (commented):
You will need to increment the REVISION to 2 so that a new package is built.

augiedoggie:
It looks like this error message is coming from the `beautifulsoup4` package.  I think maybe `soupseive` should be added to that recipe instead.

Begasus:
Thanks for the notice! As @augiedoggie mentioned, beautifulsoup4 was missing the dependency for soupsieve, this is fixed in: https://github.com/haikuports/haikuports/pull/9168
Try to update, things should be OK now.

FabianReidinger:
👍🏼

PR #9175, opened 2023-08-11
masterpiece, revbump for python, switch SOURCE_URI

Begasus:


Begasus:
Disabled for now, not all functionality is working as it should I think. :)

PR #9185, opened 2023-08-11
agg: perspective transformation update

kenmays:
- updated for detailed library description
- added perspective transformation linker patch (Zardshard)


Begasus:
Build failed (again) on both buildmasters (SOURCE_URI doesn't seem to find the archive?).

PR #9187, opened 2023-08-11
haikuporter: updated to latest 1.2.x-revision

kenmays:
- Updated to latest git-cac0903 - 26 Jul 2023.
- Updated for Python 3.10 transition.


Begasus:
In my opinion this should be done with a tagged release, maybe upstream can do so?

Begasus:
Tagged release wil be out soon, closing here.

PR #9200, opened 2023-08-13
llvm: remove older llvm recipes

kenmays:
Goal:
- Remove recipes LLVM < 9.


Begasus:
Why do this in 15 commits?
This should be carfully investigated, there are still a few other still using llvm9 for instance:
- [ ] aflplusplus-3.12c
- [ ] qt_creator-6.0.1
- [ ] clazy-1.6
- [ ] kdevelop-21.12.1
(from a quick search)

kenmays:
Actually, that is problematic for rebuilding, use case testing, and debugging. Best to migrate those ports to the default LLVM version for Haiku R1B4.

Begasus:
At this moment only 5 up to 8 are broken in the reports, no need to remove LLVM9 at this point, it's still valid without changes.

kenmays:
Restored LLVM9 recipe/patches.

Begasus:
Could you please squash your commits? Up to 17 now :/

PR #9202, opened 2023-08-13
perl, texinfo-7: provide a secondaryArch package

jmairboeck:
LilyPond requires a newer texinfo than the 6.1 that is available for x86_gcc2. Provide a secondary architecture version of texinfo 7 to fix that.

Perl (secondary) is required for texinfo.

@Begasus @pulkomandy This needs a thorough review please. I am not sure if this is a good solution.

### DIFF
--- a/dev-lang/perl/perl-5.32.1.recipe
+++ b/dev-lang/perl/perl-5.32.1.recipe
@@ -23,58 +23,59 @@ CHECKSUM_SHA256="03b693901cd8ae807231b1787798cf1f2e0b8a56218d07b7da44f784a7caeb2
 PATCHES="perl-$portVersion.patchset"
 
 ARCHITECTURES="all"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	# assume that any perl commands are backwards compatible to version 5.
-	perl = $portVersion compat >= 5
-	cmd:corelist = $portVersion compat >= 5
-	cmd:cpan = $portVersion compat >= 5
-	cmd:enc2xs = $portVersion compat >= 5
-	cmd:encguess = $portVersion compat >= 5
-	cmd:h2ph = $portVersion compat >= 5
-	cmd:h2xs = $portVersion compat >= 5
-	cmd:instmodsh = $portVersion compat >= 5
-	cmd:json_pp = $portVersion compat >= 5
-	cmd:libnetcfg = $portVersion compat >= 5
-	cmd:perl = $portVersion compat >= 5
-	cmd:perl$portVersion = $portVersion compat >= 5
-	cmd:perlbug = $portVersion compat >= 5
-	cmd:perldoc = $portVersion compat >= 5
-	cmd:perlivp = $portVersion compat >= 5
-	cmd:perlthanks = $portVersion compat >= 5
-	cmd:piconv = $portVersion compat >= 5
-	cmd:pl2pm = $portVersion compat >= 5
-	cmd:pod2html = $portVersion compat >= 5
-	cmd:pod2man = $portVersion compat >= 5
-	cmd:pod2text = $portVersion compat >= 5
-	cmd:pod2usage = $portVersion compat >= 5
-	cmd:podchecker = $portVersion compat >= 5
-	cmd:podselect = $portVersion compat >= 5
-	cmd:prove = $portVersion compat >= 5
-	cmd:ptar = $portVersion compat >= 5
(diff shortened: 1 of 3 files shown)


dev-lang/perl/perl-5.32.1.recipe
```diff
@@ -123,7 +126,8 @@ INSTALL()
 	chmod a+x $binDir/{perl,perlthanks}
 
 	# copy script into site_perl that takes care of massaging @INC into what we need.
-	cp sitecustomize.pl $prefix/non-packaged/lib/perl5/site_perl/$perlShortVersion/
+	sed "s|/lib/|/lib$secondaryArchSubDir/|g" sitecustomize.pl > $prefix/non-packaged/$relativeLibDir/perl5/site_perl/$perlShortVersion/sitecustomize.pl
+	# TODO: maybe generate the script dynamically in the recipe (using recipe variables) instead of statically in the patchset
```
jmairboeck, line 130: Just to clarify what I meant here:

Currently, the file `sitecustomize.pl` is added to the sources via a patch in the patchset (and again modified with a later patch). We could instead add a template `sitecustomize.pl.in` in additional-files and fill out the template with sed replacements, similar to how `*.rdef.in` files are used.

It would also simplify the file a bit, because it would not need both `/boot/system/` and `/boot/home/config/` prefixed paths, but could instead just use `$prefix` (i.e. `/packages/perl-...`) to cater for both cases.

sys-apps/texinfo/texinfo-7.0.3.recipe
```diff
@@ -9,44 +9,47 @@ SOURCE_URI="http://ftp.gnu.org/gnu/texinfo/texinfo-$portVersion.tar.gz"
 CHECKSUM_SHA256="3cc5706fb086b895e1dc2b407aade9f95a3a233ff856273e2b659b089f117683"
 PATCHES="texinfo-$portVersion.patchset"
 
-ARCHITECTURES="all !x86_gcc2"
+ARCHITECTURES="all"
```
Begasus, line 12: If it builds for gcc2 wouldn't it supersede the 6.* version?

jmairboeck: sorry, that was left-over from an experiment. I will fix it. (it doesn't build with gcc2)

sys-apps/texinfo/texinfo-7.0.3.recipe
```diff
@@ -1,52 +1,59 @@
...
+	lib:libiconv$secondaryArchSuffix
+	lib:libintl$secondaryArchSuffix
+	lib:libncurses$secondaryArchSuffix
+	lib:libperl$secondaryArchSuffix
 	"
+if [ -n "$secondaryArchSuffix" ]; then
+CONFLICTS="
+	texinfo
+	"
+fi
```
jmairboeck, line 39: As discussed on IRC with @Begasus:

The conflict is problematic because at least `automake` and `autoconf` require the primary arch texinfo.

How should we solve this?
 * ignore it and hope for the best? (not good)
 * use something like `--datadir=$dataDir$secondaryArchSubDir`, putting the data into `data/x86`?

The documentation can be removed from the secondary package.

Begasus: As mentioned there, I'm not a big fan on using a $secondaryArchSubDir in $dataDir (afaik it should be arch independent).
Checked rebuilds for autoconf and automake without the cycling dependency for texinfo (it wasn't there before so can be safely removed), rebuilds are fine and the conflict will only remove the primary texinfo when installing the secondary one.
My 2 cents for now, keep the conflict and let's see what others will say here.

jmairboeck: Using a different dataDir as mentioned above wouldn't work anyway because the locale files would end up wrong.

Another alternative would be to use a different package name in autoconf, so that the files end up in a different directory. It would be a bit hacky though.

jmairboeck: I resolved the conflict by renaming the package (in autoconf) as described before and removing the documentation from the secondary package.

If this is not wanted and the conflict should be kept instead, I can also revert the last commit again.

jmairboeck: I reverted the hack. We will keep the conflict, now that autoconf (#9223) and automake (#9224) are fixed.

pulkomandy:
Seems OK to me

PR #9195, opened 2023-08-12
llvm12, revbump for python changes

Begasus:

PR #9217, opened 2023-08-14
clang12: Add missing GNU headers include path

### LINKED ISSUE #9198: `clang` does not detect GNU extensions in Haiku
Using clang 12 on Haiku `hrev57210`, with this source file:

```c
#include <math.h>
#include <stdint.h>

int main()
{
	return (int)(intptr_t)(void*)&sincos;
}
```

Compilation using `gcc test.c -o test -D_GNU_SOURCE` works normally.

However, compilation using `clang` fails:

```
~/Desktop> clang test.c -o test -D_GNU_SOURCE
test.c:6:32: error: use of undeclared identifier 'sincos'
        return (int)(intptr_t)(void*)&sincos;
                                      ^
1 error generated.
```

trungnt2910:
Should fix #9198.

### DIFF
--- a/sys-devel/llvm/patches/clang-12.0.1.patchset
+++ b/sys-devel/llvm/patches/clang-12.0.1.patchset
@@ -194,3 +194,28 @@ index 50a20c2..ac65381 100644
 -- 
 2.37.3
 
+
+From 089a07be1f393fc06a8b67f015e40a718ba550dc Mon Sep 17 00:00:00 2001
+From: Trung Nguyen <57174311+trungnt2910@users.noreply.github.com>
+Date: Mon, 14 Aug 2023 14:58:56 +1000
+Subject: [PATCH] Haiku: Add missing GNU headers include path
+
+---
+ lib/Frontend/InitHeaderSearch.cpp | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/lib/Frontend/InitHeaderSearch.cpp b/lib/Frontend/InitHeaderSearch.cpp
+index b8ecee03..2d999898 100644
+--- a/lib/Frontend/InitHeaderSearch.cpp
++++ b/lib/Frontend/InitHeaderSearch.cpp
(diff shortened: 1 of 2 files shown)


trungnt2910:
Added the Python changes from #9195 as requested by @Begasus on IRC.

sys-devel/llvm/patches/clang-12.0.1.patchset
```diff
@@ -194,3 +194,28 @@ index 50a20c2..ac65381 100644
...
+--- a/lib/Frontend/InitHeaderSearch.cpp
++++ b/lib/Frontend/InitHeaderSearch.cpp
+@@ -331,6 +331,7 @@ void InitHeaderSearch::AddDefaultCIncludePaths(const llvm::Triple &triple,
+     AddPath("/boot/system/develop/headers/3rdparty", System, false);
+     AddPath("/boot/system/develop/headers/bsd", System, false);
+     AddPath("/boot/system/develop/headers/glibc", System, false);
++    AddPath("/boot/system/develop/headers/gnu", System, false);
+     AddPath("/boot/system/develop/headers/posix", System, false);
+ #ifdef HAIKU_HYBRID_SECONDARY
+     AddPath("/boot/system/develop/headers/" HAIKU_HYBRID_SECONDARY, System, false);
```
jmairboeck, line 218: Did you miss the same for `HAIKU_HYBRID_SECONDARY`? It is not visible from this patch, but I suppose that list is supposed to contain the same directories a second time for the secondary architecture.

trungnt2910: I don't think so. `headers/bsd` is only added in that one place, so `headers/gnu` should be treated similarly.

jmairboeck: You are right, these don't exist in the `x86` subdirectory. Sorry for the noise.

PR #9223, opened 2023-08-15
autoconf, revert some unneeded dependencies

Begasus:


sys-devel/autoconf/patches/autoconf264-2.64.patchset
```diff
@@ -28,5 +28,29 @@ index 4f046c0..215a5bf 100644
  
  @c Handling the indexes with Texinfo yields several different problems.
 -- 
-2.16.4
+2.37.3
+
+
+From f4a695d1483bcdf81c108ed9c8375106c270b34e Mon Sep 17 00:00:00 2001
```
Begasus, line 34: Not checked yet on 32bit, shouldn't be needed there as primary version for texinfo in still version 6.* (same for autoconf269)

Begasus: Checked on 32bit, running fine on both primary and secondary architecture.

PR #9226, opened 2023-08-15
calligra, remove python(2.7) requirement

Begasus:


Brunobasta:
Calligra does not start with Haiku 64bit nightly hrev57520, I think it stopped working some month ago!?
Where is the issues site for Calligra btw? Could not find it. (New issue!)
To send the bug report! Or is it a Haiku issue?

Begasus:
With a check on a clean 64bit R1B4 VM it worked fine after install, not sure what is going on here. :/

PR #9229, opened 2023-08-15
python3.7: remove EOL version.

OscarL:


Begasus:
Probaly some packages will be broken due to this, but those should be easy to resolve once they show up (most are already on par with python3.9 and python3.10 anyway). Thanks!

PR #9234, opened 2023-08-16
nodejs12, drop recipe, requires python2.7

Begasus:


Begasus:
npm and yarn use cmd:node on 64bit, but they are not bind to a specific version of nodejs.

Begasus:
Closing as irrelavant, python2.7 doesn't seem to be required at runtime.

PR #9238, opened 2023-08-16
glib_networking, remove python, not used/mentioned during build

Begasus:


Begasus:
This is for gcc2 only as python2.7 has been dropped, doesn't seem to be mentioned in dependencies at Gentoo.

Begasus:
Closing as irrelevant for now, on the next update it should be checked out.

PR #9239, opened 2023-08-16
libreoffice 32bit, try to fix build

Begasus:


Begasus:
Giving up, when time comes to it someone could tackle for a newer version. :)

PR #9248, opened 2023-08-18
runescape_launcher: add recipe

TheZeldakatze:
I wrote a simple launcher script for old school runescape, based on [https://gitlab.com/coringao/runescape/](https://gitlab.com/coringao/runescape/). It also has a check in it to kill the game when it's closed, as it otherwise will hang when it's quit (maybe something to do with the audio?). I hope I did everything correctly with the licensing.

Currently the game does not load when run outside of the terminal. On x86_64 it gets stuck on loading and x86_32 it cannot find the java executable. For what I could tell, the problem seems to have to do with differing environment variables when run from the terminal compared to otherwise. On x86_32 $JRE11_HOME is missing when run normally compared to the terminal. jre 11 is required as the game won't start with newer ones. I'll have a look at maybe fixing this shortly

### DIFF
--- a/games-rpg/runescape_launcher/runescape_launcher-0.9.0.recipe
+++ b/games-rpg/runescape_launcher/runescape_launcher-0.9.0.recipe
@@ -1,0 +1,49 @@
+SUMMARY="A launcher script for old-school Runescape"
+DESCRIPTION="This is a launcher script that downloads the \
+old school Runescape launcher and starts it. \
+Please note that runescape itself is closed-source and has a \
+proprietary license.\
+\
+This script is based on https://gitlab.com/coringao/runescape/"
+HOMEPAGE="https://github.com/TheZeldakatze/haiku-oldschoolrunescape-launcher"
+COPYRIGHT="2023, Maite Gamper \
+2016-2020, Carlos Donizete Froes"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/TheZeldakatze/haiku-oldschoolrunescape-launcher/archive/refs/tags/v0.9.0.tar.gz"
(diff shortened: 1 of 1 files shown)


TheZeldakatze:
Now it works properly on x86_64, I'll check on x86_32 tomorrow

TheZeldakatze:
now works on both x86_64 and x86_32

games-rpg/runescape_launcher/runescape_launcher-0.9.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+Please note that runescape itself is closed-source and has a \
+proprietary license.\
+\
+This script is based on https://gitlab.com/coringao/runescape/"
+HOMEPAGE="https://github.com/TheZeldakatze/haiku-oldschoolrunescape-launcher"
+COPYRIGHT="2023, Maite Gamper \
+2016-2020, Carlos Donizete Froes"
+LICENSE="BSD (2-clause)"
+REVISION="1"
+SOURCE_URI="https://github.com/TheZeldakatze/haiku-oldschoolrunescape-launcher/archive/refs/tags/v0.9.0.tar.gz"
```
Begasus, line 13: Looking at your changes there I  don't really see the need to not use upstream 0.8 tag instead?
The commit you made and the rdef file could be easaly adopted on the original source I guess (haven't tested).

TheZeldakatze: I don't know if that would make sense, as the original script has quite a few differences compared to the stuff I did (though I'd guess it would be possible to do all the changes based directly on coringao/runescape, though you'd need to do this overflow buffer check thing that closes the game, as otherwise you'll have an unresponive window and you'll have to force-close the game).

PR #9249, opened 2023-08-18
betterspades, remove not working SOURCE_URI's, add local zip for miss…

Begasus:
…ing files


Begasus:
Stupid mistakes on my account before. :)

PR #9260, opened 2023-08-21
openldap, bump version, make 2.6* default, disect/fix 2.4*

Begasus:


net-nds/openldap/openldap-2.6.6.recipe
```diff
@@ -41,9 +41,12 @@ REQUIRES="
 	lib:libcrypto$secondaryArchSuffix
 	lib:libssl$secondaryArchSuffix
 	"
+CONFLICTS="
```
jmairboeck, line 44: Shouldn't this be `REPLACES`?

PR #9268, opened 2023-08-23
libgcrypt, bump version

Begasus:


dev-libs/libgcrypt/patches/libgcrypt-1.10.2.patchset
```diff
@@ -69,5 +69,28 @@ index 75c49a0..794e72d 100644
...
+diff --git a/cipher/pubkey.c b/cipher/pubkey.c
+index 2341c86..2885c82 100644
+--- a/cipher/pubkey.c
++++ b/cipher/pubkey.c
+@@ -1275,7 +1275,7 @@ _gcry_pk_selftest (int algo, int extended, selftest_report_func_t report)
+ 
+ struct pk_random_override {
+   size_t len;
+-  unsigned char area[];
++  unsigned char area[1];
```
jmairboeck, line 90: This is a "flexible array member". I think the equivalent for gcc2 would be `unsigned char area[0];` (see https://gcc.gnu.org/onlinedocs/gcc/Zero-Length.html). If `sizeof(struct pk_random_override)` isn't used anywhere, when this probably doesn't matter in practice, but I would change the length to 0 anyway, just to be safe (probably even in a `#if __GNUC__ == 2` block).

Begasus: Need to boot into 32bit to check this, I looked over at upstream for the changes there and applied it here. If you got the time maybe you could have a look, otherwise I'll get back to it. Thanks for pointing out!

PR #9279, opened 2023-08-24
opendyslexic: new recipe

TheZeldakatze:
I decided to use the newer
https://github.com/antijingoist/opendyslexic
repository instead of
https://github.com/antijingoist/open-dyslexic
even though the gpo recipe uses the older repository,
as the newer one is linked to by the old repository and has a different license


Begasus:
Didn't check a build, but I guess you need to add the license to the PR?

TheZeldakatze:
> Didn't check a build, but I guess you need to add the license to the PR?

The license is in the recipe file, it's the sil Open Font v1.1 license

Begasus:
> > Didn't check a build, but I guess you need to add the license to the PR?
> 
> The license is in the recipe file, it's the sil Open Font v1.1 license

Haiku doesn't supply the license, so you need to add it to the PR :)

Issue #7252, opened 2022-10-05
[Python] Create/Update recipe for the upcomming 3.11.0 release


OscarL:
Python 3.11.0 has an expected released date of [2022-10-24](https://peps.python.org/pep-0664/).

I have working (as in: it builds, installs, and is able to run the REPL) recipe+patchset for the current 3.11.0rc2 version (based on the 3.10.7 recipe+patchset... thanks @korli et al!).

~~https://github.com/OscarL/haikuports/commit/d937592e839fb1a748194278e273a88bd5671476~~ 
Edit: watch this branch: https://github.com/OscarL/haikuports/tree/python311, until I open at least a draft PR.

It is a first draft, but I figure it might come in handy (or at least serve as a point of discussion). It certainly was/is a learning experience!

This is my first attempt at working with patchsets of this size/scope, so any tip/suggestions/fixes are more than welcome!

Some notes:

From that 3.10.7 patchset, only 6 diffs from the first patch (@korli's "initial Haiku patch") were not applying cleanly anymore. Some of those even appear to not be necessary any longer.

What I did was split the patchset, filter out those diffs from the first patch, and apply them in order (the rest applied cleanly).

Then I manually coerced the remaining diffs to play ball with the current Python code. I'm sure I messed up something somewhere, and I'm specially not sure if the changes I've made to `Lib/siteconfig.py` are correct.

I guess we'll find out.

PS: I've attempted to build with "--enable-optimizations", but it seemed to hung up (one core pegged to 100%, but memory usage was flat-lining, with zero I/O), so I killed it after half an hour of no changes.  At the time of the hang, there were some "libgcov error" messages on the Terminal window.

Edit: Tried building again with "--enable-optimizations", and the build was successful. The PYSTONE benchmark runs around a 25% faster than on 3.10.7 (latest version on HaikuDepot, without optimizations).

Not sure what can causes the builds to fail occasionally.

Begasus:
Did you clean the build before running a new build? (saw this also a few times for packages that build fine, cleaning and doing a new build solved that)

Begasus:
Well Python 3.11 is still in development stage (iirc), maybe you could get in contact with the developers there about open issues atm?

korli:
Well 3.12 might be even more work and regressions.

Begasus:
Merged in the meantime.

PR #9284, opened 2023-08-25
python3.11: new recipe for Python 3.11

OscarL:
Mayor selling point of this version: it brings several speed-related improvements over 3.10.

I'm adding it here just to make it available, gather some feedback, but with the hopes it is used only for testing purpose, and not to start adding support for it to our Python packages' recipes ("_python311" versions of packages).

Python 3.12.0 will be released on October. The 3.12.0 betas, so far, run fine on Haiku, so we might want to "skip" 3.11, and add 3.12 instead (while/after dropping 3.8), specially considering that 3.12 brings up more "breaking" changes (deprecated modules, changes in packaging, etc.) than previous versions.

Run `hp --test python3.11`, results are in-line with what we normally get on Haiku:

363 tests OK. 29 tests failed. 21 tests skipped. (and a bunch of test disable to avoid hangs on the tests-run).

Related: #7252

### DIFF
--- a/dev-lang/python/patches/python3.11-3.11.5.patchset
+++ b/dev-lang/python/patches/python3.11-3.11.5.patchset
@@ -1,0 +1,900 @@
+From 5d191a7773513212c1b4ca8e496a36f6776d75ad Mon Sep 17 00:00:00 2001
+From: Jerome Duval <jerome.duval@gmail.com>
+Date: Thu, 10 Apr 2014 16:03:33 +0000
+Subject: initial Haiku patch
+
+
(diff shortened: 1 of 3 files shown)


pulkomandy:
How long do we plan to keep Python 3.8 alive? (just to have an idea of the timeline and plans)

pulkomandy:
Yes, I agree. Last time we discussed it @korli asked for some extra delay (which is reasonable) but I don't think we discussed much about how long that delay should be.

Begasus:
It's been  about half a year ago, aside from us and a few others no reactions there, I vote for dropping python3.8 while in the process.

PR #9300, opened 2023-08-29
Adding OpenVPN patchset to work with Haiku

Swangeon:


Begasus:
Needs a revision bump to be rebuild on buildmasters. :)

Swangeon:
Went right over my head that the recipe had the revision variable. Thanks @OscarL and it should be okay now.

PR #9305, opened 2023-08-30
python3.8: remove recipe.

OscarL:
This will get EOL-ed in October (assuming it ever gets a .19 release).

We've anounced its demise on the forums in March, with a 6 months grace period. Now that we have 3.11 on the repos, time to let this one go.

While this only removes the Python 3.8 package, and not each individual "_python38" package we have, at least it will make it obvious we need to change those recipes at install/update/rebuild time.

Only one recipe (not under /dev-python/) uses `python3.8` directly, and that's for `calibre`'s TEST() (that recipe needs moving to 3.10 anyway).

So long Python 3.8!


Begasus:
Thanks!

PR #9309, opened 2023-08-30
col: new recipe

OscarL:
Ultra small package, as an alternative to having to install the bigger `util_linux` package, just because we need `cmd:col` (`zsh` and `xonch` recipes, for example).


Begasus:
As discussed on IRC, col can be pulled out of the util_linux package, same thing is already done there for cmd:cal, so maybe just go that direction?

Begasus:
> Sure thing! Waiting for that PR to get merged to close this one then! :-P

Done in: https://github.com/haikuports/haikuports/pull/9310

PR #9311, opened 2023-08-31
agg: fixed chksum issue

kenmays:
Chksum fix.


Begasus:
Error searching for the package now ... this could use a stable download uri ...

pulkomandy:
We should not pick random commits from an SVN repository and package them. There will be a stable URI when there is a release. I already complained about this in a previous version of this MR, but I was ignored.

Begasus:
Will try to keep up on previous comments made here.

Issue #9322, opened 2023-09-01
AGG no longer uses GPL license changed to MIT


bugcrazy:
AGG 2.4 uses modified BSD license, AGG 2.5 uses GPL license, AGG 2.7 uses MIT license:

"License
MIT License"

https://sourceforge.net/projects/agg/

Although the copying file uses the modified version of the BSD license, the README file says that the license is MIT:

https://sourceforge.net/p/agg/svn/HEAD/tree/agg-2.4/copying

https://sourceforge.net/p/agg/svn/HEAD/tree/agg-2.4/README

@kenmays it says in the readme and summary of the MIT license repository, but in the copying it says modified BSD license, in the Haiku container it is GPL, but GPL is only AGG version 2.5.

https://github.com/haikuports/haikuports/blob/a854cd0f039aa10b19cb57e7898620d037cfa037/x11-libs/agg/agg-2.7.0~r139.recipe#L20

kenmays:
As far as AGG r139 - GPLv3 license (i.e. refer to COPYING) is used. The MIT license was the original license under the author.

pulkomandy:
The current COPYING file disagrees with you @kenmays. The COPYING file was linked in the first comment here and it very clearly is not the GPL...

kenmays:
@pulkomandy - I agree with your historical account as I read the same. Also, I reviewed the source and found haikuporter added the COPYING file to my work directory - so, I admit the mistake here.

PR #9327, opened 2023-09-01
agg: updated license and fixed build

kenmays:
- Fixed src build issue for buildbots.
- Updated for dual licensing per "copying" file in src.
Ref: https://sourceforge.net/p/agg/svn/HEAD/tree/agg-2.4/copying
Closes: https://github.com/haikuports/haikuports/issues/9322

### DIFF
--- a/x11-libs/agg/agg-2.7.0~r139.recipe
+++ b/x11-libs/agg/agg-2.7.0~r139.recipe
@@ -15,12 +15,11 @@ do much more than that. The ideas and the philosophy of AGG are:
   * Reliability and stability (including numerical stability)."
 HOMEPAGE="https://agg.sourceforge.net/antigrain.com"
 COPYRIGHT="2002-2006 Maxim Shemanarev
-	2009-2021 John Horigan
-	1994-2021 Free Software Foundation"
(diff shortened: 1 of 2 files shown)


bugcrazy:
I think it would be better to put contributors, as the [README](https://sourceforge.net/p/agg/svn/HEAD/tree/agg-2.4/README)  file only mentions the original creator of the AGG, also remove the mention of the FSF from the line:

https://github.com/haikuports/haikuports/blob/a854cd0f039aa10b19cb57e7898620d037cfa037/x11-libs/agg/agg-2.7.0~r139.recipe#L18

https://github.com/haikuports/haikuports/blob/a854cd0f039aa10b19cb57e7898620d037cfa037/x11-libs/agg/agg-2.7.0~r139.recipe#L19

It would be better to switch to Anti-Grain Geometry contributors

korli:
> A suggestion, if I may: you could `git squash` all the commits into a single one, and use a commit message more like what you have now in this PR description text. That would be better than the default commit message we'll get by default if this gets merged as-is.

Nice suggestion, but it seems to me that kenmays doesn't use git to push to github :)

PR #9346, opened 2023-09-03
dos2unix, bump version, cleanup

Begasus:


Begasus:
Checked on both arch's, time to nuke the old ones in here :)

PR #9354, opened 2023-09-04
OpenVPN patchset update to find where Haiku publishes the driver

Swangeon:
Reflects change to where I publish the driver from [here](https://review.haiku-os.org/c/haiku/+/6608/31/src/add-ons/kernel/drivers/network/tun/driver.cpp#24)


net-vpn/openvpn/openvpn-2.6.4.recipe
```diff
@@ -7,10 +7,13 @@ HOMEPAGE="https://openvpn.net"
 COPYRIGHT="2002-2023 OpenVPN Inc"
 LICENSE="OpenVPN
 	GNU GPL v2"
-REVISION="1"
+REVISION="3"
```
Begasus, line 10: Revision should go up one number, in this case it would need to be "2"

Begasus:
Conflicting files. Did you create a new branch on a clean master or used your previous branch here?

net-vpn/openvpn/openvpn-2.6.4.recipe
Begasus: Ah my bad, didn't look any further. :)

Swangeon:
sorry about that, didn't realize that doing it on the same branch would be such a problem :p. Also thought I had a clan master but didn't. Thanks as always @OscarL & @Begasus!

Issue #9333, opened 2023-09-02
llvm16: `llvm16_clang` requires `llvm12_libs` instead of the correct `llvm16_libs`.


trungnt2910:
On a Haiku machine with `llvm16_libs` somehow already pre-installed:

```
~/Desktop> pkgman install llvm16_clang
100% repochecksum-1 [65 bytes]
Validating checksum for Haiku...done.
100% repochecksum-1 [64 bytes]
Validating checksum for HaikuPorts...done.
The following changes will be made:
  in system:
    install package llvm12_libs-12.0.1-8 from repository HaikuPorts
    install package llvm16_clang-16.0.6-1 from repository HaikuPorts
Continue? [yes/no] (yes) : 
```

## How it used to be in `llvm12`

https://github.com/haikuports/haikuports/blob/a848303ac7fd4058cde7b822e7ca36b9b13a6d29/sys-devel/llvm/llvm12-12.0.1.recipe#L435-L441

https://github.com/haikuports/haikuports/blob/a848303ac7fd4058cde7b822e7ca36b9b13a6d29/sys-devel/llvm/llvm12-12.0.1.recipe#L493-L500

`llvm12_libs` provides `libclang`, and `llvm12_clang` consumes that library.

## How it is now in `llvm16`

https://github.com/haikuports/haikuports/blob/a848303ac7fd4058cde7b822e7ca36b9b13a6d29/sys-devel/llvm/llvm16-16.0.6.recipe#L400-L406

https://github.com/haikuports/haikuports/blob/a848303ac7fd4058cde7b822e7ca36b9b13a6d29/sys-devel/llvm/llvm16-16.0.6.recipe#L459-L465

`llvm16_libs` does not provide `libclang` anymore, yet `llvm16_clang` is still said to require `libclang`. `pkgman` therefore pulls `llvm12_libs` for that library to fulfill the declared dependencies.

trungnt2910:
`llvm16_libs` should indeed not provide `libclang`:
```
~/Desktop> ls /boot/system/lib | grep clang
clang
libclang-cpp.so.12
libclang-cpp.so.16
libclang.so.12
```

The `clang` driver also does not seem to need the library (only `libclang-cpp.so.16` is opened):
```
~/Desktop> strace -l -e open -t -T clang -v test.c -o test -fuse-ld=lld 2>&1 | grep clang
clang version 16.0.6
[...]

Issue #9324, opened 2023-09-01
llvm16: 32-bit build failed


davidkaroly:
Not to be forgotten... spillover from #9141 - the 32-bit buildmaster reports a failed build, cannot find `libunwind.so.*` (probably it's looking at the wrong path)
https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=5040&viewMode=expanded

I suspect the following part of the recipe it not entirely right, i.e. `$effectiveTargetMachineTriple/libunwind` is not correct on a 32-bit system:

```
	prepareInstalledDevelLibs \
		libfindAllSymbols \
		libLLVM* \
		libLTO	\
		libclang* \
		liblld* \
		libRemarks \
		$effectiveTargetMachineTriple/libunwind
```

davidkaroly:
yep you're right, it's installed in `lib/i586-pc-haiku/` while the recipe is looking for it in `lib/x86/i586-pc-haiku/`

```
-- Installing: /packages/llvm16_x86-16.0.6-1/.self/lib/i586-pc-haiku/libunwind.so.1.0
-- Installing: /packages/llvm16_x86-16.0.6-1/.self/lib/i586-pc-haiku/libunwind.so.1
-- Installing: /packages/llvm16_x86-16.0.6-1/.self/lib/i586-pc-haiku/libunwind.so
```

```
mv: cannot stat '/packages/llvm16_x86-16.0.6-1/.self/lib/x86/i586-pc-haiku/libunwind.*': No such file or directory
```

davidkaroly:
Got a successful build with the referred PR.

davidkaroly:
`llvm16_x86` available in HaikuDepot.

PR #9357, opened 2023-09-04
llvm16: fix build on 32-bit; fix deps for clang package

davidkaroly:
Trying tackle issues #9324 and #9333
* adjust path for `libunwind` so that we always look for it in `$prefix/lib` (also in case of x86 secondary architecture)
* expand the list of `libclang...` when invoking `prepareInstalledDevelLibs` so that the libs land in the correct folders
* move `libclang` and `libclang-cpp` to the `llvm16-clang` package
* adjust provides and requires accordingly

### DIFF
--- a/sys-devel/llvm/llvm16-16.0.6.recipe
+++ b/sys-devel/llvm/llvm16-16.0.6.recipe
@@ -30,7 +30,7 @@ other than the ones listed above.
 HOMEPAGE="https://www.llvm.org/"
 COPYRIGHT="2003-2023 University of Illinois at Urbana-Champaign"
 LICENSE="Apache v2 with LLVM Exception"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/llvm/llvm-project/releases/download/llvmorg-$portVersion/llvm-project-$portVersion.src.tar.xz"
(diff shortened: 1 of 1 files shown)


sys-devel/llvm/llvm16-16.0.6.recipe
```diff
@@ -602,13 +601,106 @@ INSTALL()
 	# not all arches contain the same libraries. The inventory for each arch
 	# also changes between releases... so lets KISS.
```
korli, line 602: the comment doesn't match anymore.

davidkaroly: I understand this comment refers to `libLLVM*` which is still wildcarded... so it's still valid.

korli: Ok

augiedoggie:
There are several empty directories in the main LLVM16 package that should be cleaned up.  A `man` directory, the `lib/x86_64-unknown-haiku` directory leftover from the libunwind subpackage,  a `macho-o` directory, etc...

It seems like there are also some commands/scripts that should be moved to a different subpackage, like `run_clang_tidy` should be moved to the clang subpackage.

davidkaroly:
> 

yes that's a good suggestion. I've created PR #9405 to address these.

PR #9363, opened 2023-09-05
tqdm: update to version 4.66.1

OscarL:
- Switched build to use build/installer.
- Switched default version to Python 3.10.
- Only install man page / bash completions for the default Python version.
- Changed SOURCE_URI to avoid:
  > LookupError: setuptools-scm was unable to detect version
  for /sources/tqdm-4.66.1.

  (that error message suggested using PyPI's tarballs instead of Github ones)


Begasus:
Build failure on both buildmasters: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/5061/builds/81817.log
Probably suffers from the same as the other day and needs the build dir removed between the versions?

PR #9368, opened 2023-09-05
constantly: style cleanups, fix build.

OscarL:


dev-python/constantly/constantly-15.1.0~git.recipe
```diff
@@ -28,39 +28,49 @@ BUILD_REQUIRES="
...
+		"
 done
 
 INSTALL()
 {
+	# Patch this call, otherwise we get:
+	# "setuptools.extern.packaging.version.InvalidVersion: [...]: 'refs-gh-pull-20-rebase'"
+	sed -i -e 's|version=versioneer.get_version()|version="15.1.0dev"|'  \
+		$sourceDir/setup.py
+
```
jmairboeck, line 55: This should probably be done in the `PATCH()` function (or maybe `BUILD()`). `INSTALL()` is a bit late for patching, especially because this is only done once and doesn't depend on the python version.

PR #9377, opened 2023-09-07
qemu: update to 8.0.4

davidkaroly:


korli (approved):

davidkaroly:
something not quite right with the sdl display driver

korli:
x86: we get "no renderer found!" then it crashes with a segmentation fault

Is it a regression?

davidkaroly:
it seems to be a build environment issue...
* if I install pre-build qemu-8.0.3 packages on 32-bit Haiku, it works fine
* if I build qemu-8.0.4, I get the error
* if I rebuild qemu-8.0.3 from sources, I also get the same error. Something changed recently in SDL or Mesa packages?

davidkaroly:
re-testing shows that the issue was there on qemu-8.0.3 already therefore it's not a regression- so I think we can merge this PR as it won't make things worse.

The issue does not show up consistently: sometimes qemu starts up, sometimes it throws an exception and the 'no renderer' error message. Might be an issue in SDL initialization (CreateWindow or CreateRenderer) - I'll create a separate ticket for that.

PR #9384, opened 2023-09-08
locale_maketext_extract, add perl module

Begasus:


dev-perl/locale_maketext_extract/locale_maketext_extract-1.00.recipe
jmairboeck, line 1: I think this should be named `locale_maketext_lexicon` as that seems to be the main package and also one of the modules that this contains. Maybe adding `locale_maketext_extract` as an extra PROVIDES entry.

Should I rename the recipe? (I am currently updating all dependencies of Frozen-Bubble for secondaryArch support.)

jmairboeck: I'm pretty new to perl packagaging as well and not really an expert either ;) I just look at the details (as always) and note strange things I notice.

PR #9385, opened 2023-09-08
compress_bzip2, add perl module

Begasus:


dev-perl/compress_bzip2/compress_bzip2-2.28.recipe
```diff
@@ -0,0 +1,41 @@
...
+	haiku
+	vendor_perl
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libbz2
+	"
+BUILD_PREREQUIRES="
+	cmd:gcc
```
jmairboeck, line 27: If this requires gcc, doesn't this contain binary code then? It shouldn't use "any" as architecture in this case, but rather "all" and SECONDARY_ARCHITECTURES="x86" (to provide both versions for x86_gcc2).

Begasus: Had been tinkering about that too yesterday, maybe it was getting late? :)
Without it it fails:
```
make[1]: Entering directory '/sources/Compress-Bzip2-2.28/bzlib-src'
gcc -c   -fwrapv -fno-strict-aliasing -pipe -D_FORTIFY_SOURCE=2 -O   -DVERSION=\"\" -DXS_VERSION=\"\" -fpic "-I/packages/perl-5.32.1-1/.self/lib/perl5/5.32.1/x86_64-haiku/CORE"   blocksort.c
/bin/sh: line 1: gcc: command not found
```
I'll change this (and pathtools) to use "all" instead of "any" 👍

Issue #9394, opened 2023-09-09
[RFC] secondaryArch perl packages


jmairboeck:
Perl is now available as a secondaryArch package on x86_gcc2 (since #9202). Does it make sense to provide perl module packages (in `dev-perl`) twice for x86_gcc2?

For packages that use compiled parts (i.e. use `ARCHITECTURES="all"`), that would just mean adding the usual `SECONDARY_ARCHITECTURES="x86"` and `$secondaryArchSuffix`.

For `any` arch packages (which is the majority, I presume), we could still add `SECONDARY_ARCHITECTURES="x86"` and thus create a second package with a `_x86` suffix on x86_gcc2.

Is this a good idea or am I on the wrong track here? We might need some secondaryArch perl packages if we encounter some that use compiled parts which don't build with gcc2. Also, should `perl_x86` be treated as  a "first-class citicen" on x86_gcc2? Right now, it seems that it isn't really (but `perl_x86` is also still rather new).

----

Alternatively, we could maybe also drop the gcc2 version of perl and still provide only one perl version (i.e. `perl_x86`, and make it provide `cmd:perl` etc.) also on x86_gcc2, if this doesn't cause problems elsewhere, like with gawk.

----

## Summary

We have multiple different solutions here:

1. keep both versions as is, using `$secondaryArchSuffix` on everything and have `perl_x86` install its stuff in `$prefix/lib/x86/perl`.
   - this requires using `SECONDARY_ARCHITECTURES="x86"` and `$secondaryArchSuffix` also for perl module packages:
     - #9403
     - #9436
     - #9438
   - and some fixes for perl itself: #9437
2. drop gcc2 perl and make `perl_x86` the only perl and using `$prefix/lib/perl5`.
   - this requires fixing the Haiku build to not require old texinfo versions for x86_gcc2 (and other architectures), i.e. [Haiku ticket #18587](https://dev.haiku-os.org/ticket/18587) (or remove the perl dependency on those) and the appropriate changes to the perl recipe: #9439
3. drop `perl_x86` again and make perl only available on the primary architecture.
   - this requires removing the dependency on perl for texinfo_x86-7 (see #9461)
   - it also blocks having packages on x86 with binary perl extensions which require a modern compiler on x86 (like sdl_perl and frozen-bubble).
4. the middle-ground solution of keeping both but conflicting each other would probably not work because `cmd:perl` wouldn't be unique any more.

pulkomandy:
[...]

PR #9401, opened 2023-09-09
sdl_perl, add perl module

Begasus:


dev-perl/sdl_perl/sdl_perl-2.548.recipe
```diff
@@ -0,0 +1,77 @@
...
+	lib:libSDL_ttf_2.0$secondaryArchSuffix
+	lib:libSDL_gfx$secondaryArchSuffix
+	lib:libSDL_Pango$secondaryArchSuffix
+	lib:libsmpeg$secondaryArchSuffix
+	lib:libtiff$secondaryArchSuffix
+	alien_sdl
+	capture_tiny
+	file_sharedir
+	tie_simple
+	vendor_perl
```
jmairboeck, line 40:
```suggestion
	vendor_perl$secondaryArchSuffix
```

This probably requires $secondaryArchSuffix on the perl packages to work on x86 (see #9394).

Begasus: By the looks of it we would need the separate modules to be arch depending (as shown in IRC earlier)

Begasus:
Merging for now, 32bit fixes will follow once perl issues there are sortened out.

PR #9403, opened 2023-09-09
archive_extract: cleanup, add tests

jmairboeck:
This is a test for #9394. If this works and is accepted, the same pattern could be used for other perl packages.

### DIFF
--- a/dev-perl/archive_extract/archive_extract-0.88.recipe
+++ b/dev-perl/archive_extract/archive_extract-0.88.recipe
@@ -6,7 +6,7 @@ commandline tools"
 HOMEPAGE="https://metacpan.org/pod/Archive::Extract"
 COPYRIGHT="2006-2021 Andy Lester"
 LICENSE="Artistic"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://cpan.metacpan.org/authors/id/B/BI/BINGOS/Archive-Extract-$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


jmairboeck:
I added some cleanup in a second commit to remove some architecture specific files. Something similar should probably be done for all "any" arch perl packages.

Begasus:
LGTM, will try  this with the other ones and see if we can get frozen-bubble up and running on 32bit :)

dev-perl/archive_extract/archive_extract-0.88.recipe
```diff
@@ -38,5 +39,10 @@ BUILD()
...
 INSTALL()
 {
-	make install DESTDIR="${DESTDIR}"
+	make install
+
+	# remove architecture-specific files
+	cd $prefix
+	rm -r $(perl '-V:archlib|vendorarch' | cut -d\' -f2 | cut -d/ -f5-)
+		# cut extracts the quoted string and strips the prefix (which is perl's and not ours)
 }
```
Begasus, line 48: This is a problem with compress_bzip2

jmairboeck: This is only meant for "any" arch recipes, not for architecture specific ones.

jmairboeck:
I reverted the secondaryArch stuff now that this is finally settled and just kept the cleanup and test changes in.

PR #9405, opened 2023-09-10
llvm16: clean up recipe

davidkaroly:
Following up on the suggestions from @augiedoggie 
* move scan-build-py scripts to clang_analysis package
* move other clang scripts to clang package
* remove empty folders from llvm16 base package

successful build on 32-bit&64-bit


augiedoggie (approved):
I haven't tried a build but it looks good.  Thanks for cleaning it up a bit.

PR #9411, opened 2023-09-13
amule, new recipe

Begasus:


Begasus:
Creates Incomming and Temp directories in ~/.aMule (with some other files like settings,preferences ...), Haven't found out how to change those paths.
Languages are not installed (I can use autotools which builds the translations but then aMule still doesn't pick them up, so left them out for now.

PR #9426, opened 2023-09-15
Another World (NEO-RAW): new recipe

Luca1991:
This is the recipe for Another World (based on NEO-RAW).
The original version of this game is NOT opensource, NEO-RAW is a re-implementation of the original game VM.
Due to this, you have to provide your game files (BANK* and MEMLIST.BIN).

Haiku-OS forum thread (as reference): https://discuss.haiku-os.org/t/another-world-raw-ported-to-haiku/

@Begasus, show no mercy for my scripts! 😇

### DIFF
--- a/games-action/another_world/additional-files/another_world.sh
+++ b/games-action/another_world/additional-files/another_world.sh
@@ -1,9 +1,14 @@
 #!/bin/bash
 
-DATA_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world
+DATA_PATH=`findpaths -e B_FIND_PATH_DATA_DIRECTORY another_world | head -n 1`
 SAVE_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world/save
 
-mkdir -p $DATA_PATH # create data directory if it does not exist
+if [ -z $DATA_PATH ];
+then
+	DATA_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world
+	mkdir -p $DATA_PATH # create data directory if it does not exist
+fi
+
 mkdir -p $SAVE_PATH # create save directory if it does not exist
 
 # Check if all files are present
@@ -22,9 +27,10 @@ if [ ! -f $DATA_PATH/BANK01 ] \
 || [ ! -f $DATA_PATH/BANK0D ] \
 || [ ! -f $DATA_PATH/MEMLIST.BIN ];
 then
-    if [ "$(alert --warning "Please copy original Another World files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open $DATA_PATH" "Cancel")" == "Open $DATA_PATH" ]; then
+    if [ "$(alert --warning "Please copy original 'Another World' files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open data folder" "Cancel")" == "Open data folder" ];
+    then
         open $DATA_PATH
     fi
 else # all files are present, launch the game
-   `/bin/finddir B_SYSTEM_APPS_DIRECTORY`/AnotherWorld/raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
+   neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
 fi
--- a/games-action/another_world/additional-files/another_world.sh
+++ b/games-action/another_world/additional-files/another_world.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+REQ_FILES=('BANK01' 'BANK02' 'BANK03' 'BANK04' 'BANK05' 'BANK06' 'BANK07'
+ 'BANK08' 'BANK09' 'BANK0A' 'BANK0B' 'BANK0C' 'BANK0D' 'MEMLIST.BIN')
 DATA_PATH=`findpaths -e B_FIND_PATH_DATA_DIRECTORY another_world | head -n 1`
 SAVE_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world/save
 
@@ -12,25 +14,22 @@ fi
 mkdir -p $SAVE_PATH # create save directory if it does not exist
 
 # Check if all files are present
-if [ ! -f $DATA_PATH/BANK01 ] \
-|| [ ! -f $DATA_PATH/BANK02 ] \
-|| [ ! -f $DATA_PATH/BANK03 ] \
-|| [ ! -f $DATA_PATH/BANK04 ] \
-|| [ ! -f $DATA_PATH/BANK05 ] \
-|| [ ! -f $DATA_PATH/BANK06 ] \
-|| [ ! -f $DATA_PATH/BANK07 ] \
-|| [ ! -f $DATA_PATH/BANK08 ] \
-|| [ ! -f $DATA_PATH/BANK09 ] \
-|| [ ! -f $DATA_PATH/BANK0A ] \
-|| [ ! -f $DATA_PATH/BANK0B ] \
-|| [ ! -f $DATA_PATH/BANK0C ] \
-|| [ ! -f $DATA_PATH/BANK0D ] \
-|| [ ! -f $DATA_PATH/MEMLIST.BIN ];
-then
-    if [ "$(alert --warning "Please copy original 'Another World' files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open data folder" "Cancel")" == "Open data folder" ];
+for file in ${REQ_FILES[@]};
+do
+    if [ ! -f $DATA_PATH/$file ];
     then
-        open $DATA_PATH
+        if [ $TERM == "dumb" ];
+        then
+            if [ "$(alert --warning "Missing file: $file. Please copy original 'Another World' files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open data folder" "Cancel")" == "Open data folder" ];
+            then
+                open $DATA_PATH
+            fi
+        else
+            echo "Missing file: $file. Please copy original 'Another World' files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)."
+        fi
+        exit
     fi
-else # all files are present, launch the game
-   neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
-fi
+done
+
+# all files are present, launch the game
+neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
--- a/games-action/another_world/additional-files/another_world.sh
+++ b/games-action/another_world/additional-files/another_world.sh
@@ -32,4 +32,4 @@ do
 done
 
 # all files are present, launch the game
-neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
+exec neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
--- a/games-action/another_world/additional-files/another_world.sh
+++ b/games-action/another_world/additional-files/another_world.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+DATA_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world
+SAVE_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world/save
+
+mkdir -p $DATA_PATH # create data directory if it does not exist
+mkdir -p $SAVE_PATH # create save directory if it does not exist
+
+# Check if all files are present
+if [ ! -f $DATA_PATH/BANK01 ] \
+|| [ ! -f $DATA_PATH/BANK02 ] \
+|| [ ! -f $DATA_PATH/BANK03 ] \
+|| [ ! -f $DATA_PATH/BANK04 ] \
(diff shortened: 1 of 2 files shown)


games-action/another_world/additional-files/another_world.sh
```diff
@@ -0,0 +1,30 @@
...
+|| [ ! -f $DATA_PATH/BANK07 ] \
+|| [ ! -f $DATA_PATH/BANK08 ] \
+|| [ ! -f $DATA_PATH/BANK09 ] \
+|| [ ! -f $DATA_PATH/BANK0A ] \
+|| [ ! -f $DATA_PATH/BANK0B ] \
+|| [ ! -f $DATA_PATH/BANK0C ] \
+|| [ ! -f $DATA_PATH/BANK0D ] \
+|| [ ! -f $DATA_PATH/MEMLIST.BIN ];
+then
+    if [ "$(alert --warning "Please copy original Another World files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open $DATA_PATH" "Cancel")" == "Open $DATA_PATH" ]; then
```
humdingerb, line 25: As "$DATA_PATG" is quite long, this would be a fairly wide button...
How about "Open data folder"?

humdingerb, line 25: Putting 'Another World' in single quotes makes for easier reading.

games-action/another_world/another_world-0.0.1.recipe
```diff
@@ -0,0 +1,53 @@
+SUMMARY="Another World (NEO-RAW)"
+DESCRIPTION="RAW is a re-implementation of the engine used \
+in the game Another World. This game, released \
```
humdingerb, line 3:
```suggestion
in the game 'Another World'. This game, released \
```

games-action/another_world/another_world-0.0.1.recipe
```diff
@@ -0,0 +1,53 @@
...
+under the name Out Of This World in non-European \
```
humdingerb, line 4:
```suggestion
under the name 'Out Of This World' in non-European \
```

games-action/another_world/another_world-0.0.1.recipe
```diff
@@ -0,0 +1,53 @@
...
+countries, was written by Eric Chahi at the beginning of the '90s. \
```
humdingerb, line 5: Remove "\" to create a linebreak. In fact, I'd insert a newline, too, so it stands out evenmore.
```suggestion
countries, was written by Eric Chahi at the beginning of the '90s. 


```

humdingerb (changes requested):

games-action/another_world/another_world-0.0.1.recipe
```diff
@@ -0,0 +1,53 @@
...
+YOU NEED ORIGINAL GAME FILES (DOS VERSION)."
```
humdingerb, line 6: I wonder if it wouldn't be a good idea to add the info where to copy the original files here, too. A user may wonder where to put those, and getting easily frustrated, never even starts the game and never sees the alert...

```suggestion
YOU NEED ORIGINAL GAME FILES!
Put the DOS version of 'BANK*' and 'MEMLIST.BIN' in ~/config/non-packaged/data."

Or use finddir and a variable. Or leave it as is, it's just a suggestion. :)
```

games-action/another_world/another_world-0.0.1.recipe
```diff
@@ -0,0 +1,53 @@
...
+BUILD()
+{
+	cmake -DCMAKE_BUILD_TYPE=Release .
+	make
+}
+
+INSTALL()
+{
+	mkdir -p $appsDir/AnotherWorld/
+	cp raw $appsDir/AnotherWorld/raw
```
Begasus, line 49: For me I would just put the binary in $binDir (or $prefix/bin), that way the user can always control the binary itself even without using the script.

games-action/another_world/additional-files/another_world.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+DATA_PATH=`/bin/finddir B_USER_NONPACKAGED_DATA_DIRECTORY`/another_world
```
humdingerb, line 3: I wonder, if using `findpaths B_FIND_PATH_DATA_DIRECTORY` instead of `finddir`, and later iterate over all returned locations when checking for the existence of the game files, would be better.
That way, if someone created a HPKG with those files, it'd be found as well, because then they would not be in B_USER_NONPACKAGED_DATA_DIRECTORY, but in a regular /system/data or ~/config/data.

When prompting the user with an alert, we should still suggest B_USER_NONPACKAGED_DATA_DIRECTORY, I think. Though B_SYSTEM_NONPACKAGED_DATA_DIRECTORY would work as well (and if Haiku ever goes multiuser, the game data would only needed to be kept once in "system", and not possibly multiple times for each user).

humdingerb: ETA:  `findpath -r another_world B_FIND_PATH_DATA_DIRECTORY` will return the game data path if it exists, or an error message.

games-action/another_world/another_world-0.0.1.recipe
Luca1991: Should I create "AnotherWorld" directory in $binDir? I ask because:
1) "raw" is a rather common name, it may conflict with other binaries
2) should "another_world.sh" (the launch script) be put there too?

games-action/another_world/additional-files/another_world.sh
Luca1991: Uhm, in my case `findpaths -r another_world B_FIND_PATH_DATA_DIRECTORY` returns "/boot/system/data" and I don't have the game data directory there.

Anyway I think that the idea to support a multiuser env is good, but:
1) we need a reliable way to dynamically retrieve both $DATA_PATH (and $SAVE_PATH) while checking for files and when launching the game
2) I think that $SAVE_PATH should still point to ~/config/non-packaged/data/another_world in order to avoid save-games corruption in multiuser systems (when Haiku will goes multiuser).

Please let me know what should I do :)

humdingerb: > Uhm, in my case `findpaths -r another_world B_FIND_PATH_DATA_DIRECTORY` returns "/boot/system/data" and I don't have the game data directory there.

I think I have misunderstood the -r parameter. Try with `findpaths -e B_FIND_PATH_DATA_DIRECTORY another_world`. Best pipe through `head -n 1` so you only get the highest prio location if there are more than one (e.g. installed in ~/config/non-packaged/data and the system hierarchy).
The above should reliably return the location of the "another_world" data folder, if it exists.

I agree, $SAVE_PATH should always point to ~/config/non-packaged/data/another_world, so each future user has their own save point.

games-action/another_world/another_world-0.0.1.recipe
Begasus: Got a point here, same goes however in $appsDir, renaming it would be weird too, so no strong opinion atm, maybe some others can chime in here?

Begasus: Oh, and no subdirectories in $binDir please :)

humdingerb: Question is, as a game, will it often be started from the commandline? Will it benefit from being in the PATH in /bin?
If not, putting it in $appsDir feels more correct.

Luca1991: @Begasus and @humdingerb, my opinion is that, since you always have to specify at least the --datapath parameter, not many people will benefit from launching the "raw" executable directly. At that point, I think most users will write their own launch script.

But this is just my opinion, if having the bin in $binDir is something you want, I'm ok to move it there ;)

humdingerb: Maybe this question should be posed to a wider audience, but my feeling is that the /bin folder is where the CLI apps live. Having a GUI app or game there feels strange to me, but that may be personal preferece or unfamiliarity with how other OS handle it.

pulkomandy: Other OS don't make that distinction.

Linux has only "bin" and Windows has only "Program Files" (which is more like apps/)

As often, we go with the "why not both?" approach.

I think a subdirectory in appsDir makes sense here.

Luca1991: @pulkomandy Linux also has the "opt" directory, which is more similar to "Program Files" in Windows, and I think it is like "appsDir" in Haiku. That's why I think it is ok to use a subdirectory there, where to place both the binary and the startup script. But again, this is just personal preference :) 

Another solution that can make everyone happy: 
1) rename the game binary from "raw" to "neo-raw" (to avoid any name clashes)
2) copy the new "neo-raw" binary to $binDir
3) create the "AnotherWorld" directory in $appDir, then copy the startup script and symlink "neo-raw" here

This way the user has the neo-raw binary in $binDir and can launch it from terminal if he wants, and at the same time the game is also bootable from the startup script.

humdingerb: Or go with your 1) and 2), but put the startup script as "Another World" directly in $appDir. No need for a sub-folder with just the one file.

Begasus: Agreed with @humdingerb  :)

Luca1991: Perfect, I agree with @humdingerb too. I'll patch the script later tonight :)

Luca1991:
Hello @humdingerb and @Begasus

All proposed fixes have been applied! The game is bootable with data placed in ~/config/non-packaged/data/another_world/ or /boot/system/non-packaged/data/another_world/.

Please double-check and let me know, thanks :)

games-action/another_world/additional-files/another_world.sh
```diff
@@ -22,9 +27,10 @@ if [ ! -f $DATA_PATH/BANK01 ] \
 || [ ! -f $DATA_PATH/BANK0D ] \
 || [ ! -f $DATA_PATH/MEMLIST.BIN ];
 then
-    if [ "$(alert --warning "Please copy original Another World files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open $DATA_PATH" "Cancel")" == "Open $DATA_PATH" ]; then
+    if [ "$(alert --warning "Please copy original 'Another World' files (DOS version) to $DATA_PATH (BANK* and MEMLIST.BIN)." "Open data folder" "Cancel")" == "Open data folder" ];
```
korli, line 30: Please use a for loop:

```
for file in BANK01 BANK02; do
   if [ ! -f $DATA_PATH/$file ]; then
      ...
      exit 1
   fi
done

neo-raw ...
```

Also calling a script in a Terminal shouldn't popup a graphical alert...

Luca1991: hi @korli 
I agree the using a for loop is much cleaner and I'll fix this later today.

> Also calling a script in a Terminal shouldn't popup a graphical alert...

While I can certainly understand what you mean, I've used that popup in a script for the following reasons:
1) It was suggested in the forum thread [https://discuss.haiku-os.org/t/another-world-raw-ported-to-haiku/13962/9](url).
2) It was already used in (among others) JA2 recipe [https://github.com/haikuports/haikuports/blob/master/games-strategy/ja2_stracciatella/additional-files/ja2_stracciatella.sh](url) Indeed my startup script is greatly inspired from this one.
3) It is surely a great QoL for newbie (and/or lazy) users.
4) There is no other way to notify the user about the missing game files.

That said, please let me know if this is something unacceptable and I'll remove it.

pulkomandy: It would be possible to test for the TERM variable to be set: if it is set, you're running from Terminal, you can use echo. If it is not set, you're running from Tracker, use alerts.

But maybe we should keep this simple and stop over-engineering it?

korli: I thought there was something else than alert, which correctly uses a popup or the Terminal input. Maybe not.

Luca1991: I like what suggested by @pulkomandy. I'll use the TERM var to choose between an echo or an alert.

Tonight I'll implement these two things. Thank you.

PS: is even possible to boot haiku in text-only mode? I wasn't aware of that.

pulkomandy: Yes, I never remember how it is between Haiku and BeOS (one sets it to "dumb", the other does not set it at all if I remember correctly)

Luca1991:
Hi all @humdingerb , @pulkomandy , @korli  and @Begasus 

The last commit should make everyone happy. I think this PR is now mergeable.

Thanks,
Luca

humdingerb:
The PROVIDES section is missing a `cmd:neo_raw`.
Otherwise looks good. The alert pops up. I couldn't test further, because I lack the game files.

games-action/another_world/additional-files/another_world.sh
```diff
@@ -12,25 +14,22 @@ fi
...
+        exit
     fi
-else # all files are present, launch the game
-   neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
-fi
\ No newline at end of file
+done
+
+# all files are present, launch the game
+neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
```
korli, line 35:
```suggestion
exec neo-raw --datapath=$DATA_PATH --savepath=$SAVE_PATH $@
```

Luca1991:
Hi @humdingerb  and @korli, your suggested changes have been applied with my last commit.

korli:
LGTM

PR #9437, opened 2023-09-16
perl: fix 'perlpath' config variable on secondaryArch

jmairboeck:
This is another case of wrongly using $prefix/bin instead of $binDir.

It fixes a few broken tests, but doesn't fix sdl_perl (and thus frozen-bubble) yet.

I would suggest to not merge this yet. Maybe more fixes are necessary to fix sdl_perl.


jmairboeck:
Superseded by #9439

PR #9439, opened 2023-09-17
perl: drop gcc2, making perl_x86 the default

jmairboeck:
This is an alternative solution to #9394

### DIFF
--- a/dev-lang/perl/perl-5.32.1.recipe
+++ b/dev-lang/perl/perl-5.32.1.recipe
@@ -16,58 +16,64 @@ HOMEPAGE="https://www.perl.org/"
 COPYRIGHT="1993-2019 Larry Wall and others"
 LICENSE="GNU GPL v1
 	Artistic"
-REVISION="1"
+REVISION="2"
 perlShortVersion="${portVersion%.*}"
 SOURCE_URI="http://www.cpan.org/src/perl-$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


dev-lang/perl/perl-5.32.1.recipe
```diff
@@ -16,59 +16,65 @@ HOMEPAGE="https://www.perl.org/"
...
+	cmd:splain = $portVersion compat >= 5
+	cmd:streamzip = $portVersion compat >= 5
+	cmd:xsubpp = $portVersion compat >= 5
+	cmd:zipdetails = $portVersion compat >= 5
 	# vendor_perl refers to the path to packaged perl modules, which includes
 	# the short perl version, so it is backwards compatible up this one.
-	vendor_perl$secondaryArchSuffix = $perlShortVersion compat >= $perlShortVersion
-	lib:libperl$secondaryArchSuffix = $portVersion compat >= $perlShortVersion
+	vendor_perl = $perlShortVersion compat >= $perlShortVersion
+	lib:libperl = $portVersion compat >= $perlShortVersion
```
jmairboeck, line 66: This one is a bit of a special case of a library, as it is a library built with gcc 13 but without $secondaryArchSuffix because it isn't in $libDir. But there is also no devel counterpart for it, so it is a bit special anyway.

jmairboeck: fixed in 14c4afa

PR #9441, opened 2023-09-17
poppler, cleanup old recipes

Begasus:
From what I could find only openscenegraphs uses poppler076, this can be updated to use a more recent version, other packages use the default poppler version (22.01.0), in the works is poppler23 (version 23.09.0).


Begasus:
The old versions won't build anymore on the newer gcc versions (seems this is an issue since gcc11 from searching the web for the error), also 2 of them still use python2.7 (deprecated), so I think it's time to drop these version and fix what comes allong.
```
/sources/poppler-0.76.1/glib/poppler-document.cc: In function 'GType poppler_layers_iter_get_type()':
/sources/poppler-0.76.1/glib/poppler-private.h:155:13: error: argument 2 of '__atomic_load' must not be a pointer to a 'volatile' type
  155 |         if (g_once_init_enter (&g_define_type_id__volatile)) {                        \
      |             ^~~~~~~~~~~~~~~~~
/sources/poppler-0.76.1/glib/poppler-document.cc:2985:1: note: in expansion of macro 'POPPLER_DEFINE_BOXED_TYPE'
 2985 | POPPLER_DEFINE_BOXED_TYPE (PopplerLayersIter, poppler_layers_iter,
```

PR #9461, opened 2023-09-19
texinfo: move commands to $prefix/bin and update for secondaryArch perl

korli:
to drop the dependency on libperl_x86.
installs binaries in $prefix/bin on x86.

### DIFF
--- a/sys-apps/texinfo/texinfo-7.0.3.recipe
+++ b/sys-apps/texinfo/texinfo-7.0.3.recipe
@@ -38,12 +38,8 @@ REQUIRES="
 	lib:libiconv$secondaryArchSuffix
 	lib:libintl$secondaryArchSuffix
 	lib:libncurses$secondaryArchSuffix
+	lib:libperl
 	"
-if [ -z "$secondaryArchSuffix" ]; then
-	REQUIRES+="
-		lib:libperl
-		"
-fi
 if [ "$targetArchitecture" = x86_gcc2 ]; then
 	CONFLICTS="texinfo"
 fi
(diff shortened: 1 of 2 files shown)


jmairboeck:
So you want to keep perl on gcc2 only on x86 and remove perl_x86 again? That would also be solution for #9394, but would also exclude frozen-bubble from there, and maybe other things in the future.

korli:
Actually I have no preferences on perl on x86. AFAICT newer perl versions continue to build on x86_gcc2. perl_x86 only is an option also.

Begasus:
Good to merge?

jmairboeck:
It looks like we are moving perl to gcc 13 only, dropping gcc 2 (see #9439), so the "moving the binaries" part of this is good, the "disable perl xs" part is then not needed any more.

sys-apps/texinfo/texinfo-7.0.3.recipe
```diff
@@ -49,24 +59,31 @@ BUILD_PREREQUIRES="
...
 	make $jobArgs
 }
 
 INSTALL()
 {
 	make install
 
-	# remove libtool files
-	rm $libDir/texinfo/*.la
+	if [ -z "$secondaryArchSuffix" ]; then
```
jmairboeck, line 84: why is this if needed?

jmairboeck:
> Good to merge?

Now I think it is ;)

jmairboeck:
I also included a small fix for texinfo-6 (revbump and removed dependency) to rebuild it without perl XS, now that gcc2 perl is gone.

jmairboeck:
I added another small fix for texinfo 7, adding back the $secondaryArchSuffix for libperl (see 14c4afa for the explanation).

PR #9462, opened 2023-09-20
tzlocal: remove broken, unused recipe.

OscarL:
And... we're dropping it. Original PR description follows:

----

This package was actually broken for Python 3.9, and it was (and is) not used for anything on-tree.

Update to latest version, and support only Python 3.10 for now. Still not really useful, as it needs fixes on our Pythons' recipes (correctly setting `--with-tzpath=` during configure, and requiring `timezone_data`).

This is still an improvement, as it defaults to UTC for now, instead of just erroring out like before.

---

Just as a clarification: I expect it to become fully functional once we adjust our main Python packages.

In any case, it will still remain installable via pip if we decide to just drop it.


korli:
We might drop the recipe.

PR #9464, opened 2023-09-20
sip: update to version 6.7.11, support Python 3.10.

OscarL:
Recipe cleanups/fixes.


korli:
Did you test on x86?

korli:
The use of $binDir but no suffix on cmd: looks problematic...

PR #9465, opened 2023-09-20
node_semver: drop unused recipe.

OscarL:
Another one bites the dust... (original PR description follows)

---

Minor cleanup/fixes.

Not used by anything on-tree. Might want to just drop this one next time. (still will be available via `pip install`).


korli:
About the importance of explaining why a python module is added in the commit message... https://github.com/haikuports/haikuports/commit/b3556a2ed3c8c4f1c2c52c8e7fb3686054302d6f

Begasus:
> About the importance of explaining why a python module is added in the commit message... [b3556a2](https://github.com/haikuports/haikuports/commit/b3556a2ed3c8c4f1c2c52c8e7fb3686054302d6f)

My bad, should remember this ;)

@OscarL no idea what I needed it for back then, probably added it to not let it rot here, can be removed as nothing is using it. :)

Begasus:
@OscarL could very well be, in that case both can be dropped, if the need arizes they could be revived, thanks for spotting this :)

PR #9466, opened 2023-09-20
blake3, new recipe

Begasus:


Begasus:
Downloaded while building rizin, though it seems it doesn't provide the option to actualy use the system version.
If clutter can be removed in  a later stage, could be usefull to others in the future maybe? :)

PR #9475, opened 2023-09-21
rdflib: switch recipe to Python 3.10.

OscarL:
Instead of dropping it, just move it to Python 3.10.

Original PR description follows:

---
Nothing uses it on-tree. Still available via `pip install`.


korli:
It provides commands

PR #9488, opened 2023-09-21
xion, 0.17.0, move binaries to tools package

Begasus:
0.18.1 new version, add conflicts for tools package Both versions can be co-installed aside from the binaries


Begasus:
mdds2.1 needs to switch to secondary architecture, requires devel:libboost_system for the tests. (mmds2.1 build broken on buildmaster for 32bit primary arch)

PR #9491, opened 2023-09-22
weechat: update to version 4.0.4, move to Python 3.10.

OscarL:
Lightly tested on beta4 64 bits (was able to connect and chat over the OFTC's `#haiku` channel).

### DIFF
--- a/net-irc/weechat/weechat-4.0.4.recipe
+++ b/net-irc/weechat/weechat-4.0.4.recipe
@@ -7,7 +7,7 @@ COPYRIGHT="2003-2023 Sébastien Helleu"
 LICENSE="GNU GPL v3"
 REVISION="1"
 SOURCE_URI="https://weechat.org/files/src/weechat-$portVersion.tar.xz"
-CHECKSUM_SHA256="f7cb65c200f8c090c56f2cf98c0b184051e516e5f7099a4308cacf86f174bf28"
+CHECKSUM_SHA256="ae5f4979b5ada0339b84e741d5f7e481ee91e3fecd40a09907b64751829eb6f6"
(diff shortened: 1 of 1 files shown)


net-irc/weechat/weechat-4.0.4.recipe
```diff
@@ -54,7 +54,7 @@ BUILD_REQUIRES="
 	devel:libintl$secondaryArchSuffix
 	devel:liblua$secondaryArchSuffix >= 5.3
 	devel:libncurses$secondaryArchSuffix
-	devel:libpython3.9$secondaryArchSuffix
+	devel:libpython3.10$secondaryArchSuffix
```
augiedoggie, line 57: Seems like it should also have a `lib:libpython3.10$secondaryArchSuffix` in the REQUIRES too.  Does it actually link against `libpython`?

augiedoggie: I see it listed as an optional dependency for Arch Linux and I don't see libpython in the `ldd` output so I guess it's ok.  Haiku/Haikuporter could use a RECOMMENDS variable.

Begasus: My guess is that it will use what is available at runtime, same as for Geany.

augiedoggie: Right, but some of the optional dependencies are listed, like lua, but python isn't.

Begasus: Maybe not in their docs, in Arch they are listed as optional: https://archlinux.org/packages/extra/x86_64/weechat/

PR #9518, opened 2023-09-24
kvazaar, bump version, needed for updated crypto++

Begasus:


Begasus:
Maybe wrong wording, should be: `needed for updated crypto++` :)

PR #9519, opened 2023-09-24
timezone_data: update to version 2023c

OscarL:
Set make flags also on BUILD(), otherwise tzselect ended up with the wrong location for TZDIR.

Previous version failed with:
`/bin/tzselect: time zone files are not set up correctly`

Now it finds the files ok, but trying to use it further causes /bin/date to crash around 7 times in a row (?!).

`> zdump -v America/Buenos_Aires` also works now, while previously failed with:

`America/Buenos_Aires: No such file or directory`,

or with:

```
> ln -s /boot/system/data/zoneinfo/America/Buenos_Aires /etc/localtime
> zdump -v /etc/localtime
/etc/localtime: No error
```

Now the latter also works OK.

### DIFF
--- a/sys-libs/timezone_data/patches/timezone_data-2023c.patchset
+++ b/sys-libs/timezone_data/patches/timezone_data-2023c.patchset
@@ -1,14 +1,14 @@
-From 8cf4e37523a8ec3fc4267c54f5b7c1a83d3227a0 Mon Sep 17 00:00:00 2001
+From 45d6570d90bb9b04a55f59e92db9cf53c9ee4f93 Mon Sep 17 00:00:00 2001
 From: Jerome Duval <jerome.duval@gmail.com>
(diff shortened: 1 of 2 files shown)


korli:
/bin/date is from coreutils, what's the relation? libtz.a is also provided without headers by arch linux. Why is there a lib:libtz PROVIDES btw?

korli:
What's left with this PR?

Begasus:
Build failure on 32bit: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/5306/builds/82206.log

korli:
Seems, for C89, a macro needs to be defined:
https://github.com/eggert/tz/blob/dd83b3a2b5afb4b1c296aede092785830487adc9/NEWS#L118

PR #9528, opened 2023-09-25
stress_ng: update to version 0.16.05

OscarL:
Built and very lightly tested on both 64 and 32 bits (beta4).

Had to drop half of the patchset for "_x86" (`stress-icache.c` has received [some heavy changes](https://github.com/ColinIanKing/stress-ng/commit/cd815f23accabc9e90086c0a3afefea8d9c75c2a#diff-b2df50226485121d3c09dac892044d98667b0ab2c3aaf8e065467716b87bd0b7)).

Tried that part on 32 bits with `stress-ng -M -t 30 --icache 2`, it didn't explode, but maybe @davidkaroly could review, just in case?


davidkaroly:
As I remember the icache patch was needed to get an executable on 32-bit that can actually loaded... by default there were some sections with too large alignment constraints which were rejected by runtime_loader for some reason. If you get a binary file that starts up properly without any runtime_loader errors, then the icache patch is not needed any more.

PR #9535, opened 2023-09-28
djvu, bump version (move djvulibre to djvu according to Gentoo)

Begasus:


Begasus:
Large part of the patchset are patches imported from Gentoo and Fedora (addresses issues reported as vulnerable).

Begasus:
Need to check 32bit first. (works fine with the djvu translator on 64bit - checked)

Begasus:
I've got 2 scripts in $binDir (in the tools package), this fails atm with the created list in `defineDebugInfoPackage`, is there a way to `exclude` these 2 in the `debugList`?
```
debugList=(
	$libDir/libdjvulibre.so.$libVersion
	)
	for i in `echo "$PROVIDES_tools" | sed -n -e "s/^\s*cmd://p"`; do
		debugList+=("${binDir/$portName/${portName}_tools}"/$i)
	done
fi
defineDebugInfoPackage $portName "${debugList[@]}"
```

PR #9561, opened 2023-10-03
python3.11: update to version 3.11.6

OscarL:


dev-lang/python/python3.11-3.11.6.recipe
```diff
@@ -92,6 +93,10 @@ BUILD_PREREQUIRES="
 	cmd:pkg_config$secondaryArchSuffix
 	"
 
+TEST_REQUIRES="
+	timezone_data
```
korli, line 97: isn't it already with REQUIRES covered?

korli: Correct

PR #9563, opened 2023-10-03
db, bump version

Begasus:


Begasus:
First time I fiddled a bit with java so disabled the options in the recipe, I could get 2 .jar files which were installed in $libDir, but couldn't get them to launch. Maybe something to tackle for someone more familiar :)
Also enabling java probably brakes the gcc2 build (didn't check there yet).

jmairboeck:
This could be done similar to how I did it with portmidi, i.e. conditionally disable java for the gcc2 version.

dev-db/db/db18-18.1.40.recipe
```diff
@@ -0,0 +1,132 @@
...
+#	lib:libdb_java_18.1_g$secondaryArchSuffix = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libcrypto$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	"
+
+if [ -z "$secondaryArchSuffix" ]; then
+	SUMMARY_tools="The TIFF tools"
```
jmairboeck, line 37: This summary is wrong.

Begasus: Whoops! Left over from copy/paste :) (will fix)

Begasus:
> This could be done similar to how I did it with portmidi, i.e. conditionally disable java for the gcc2 version.

Probably, but since it's not really in use by others I didn't enable it, if need be gcc2 could be dropped either way :)
If you think it would be useful I can enable it and nuke gcc2 build for it.

Issue #9219, opened 2023-08-15
LibreOffice 6.4.7.2-10 stoped working (32 bit)


unspacyar:
Hello. After recent update, LibreOffice stopped working. When trying to start the application from Terminal, I get the following error:

```
/boot/system/apps/LibreOffice> ./'LibreOffice Calc' 
runtime_loader: /boot/system/add-ons/x86/locale/catalogs/IFF: Could not resolve symbol '_Z15convert_to_utf8mPKcPlPcS1_S1_c'
resolve symbol "_Z15convert_to_utf8mPKcPlPcS1_S1_c" returned: -2147478780
runtime_loader: /boot/system/add-ons/x86/locale/catalogs/IFF: Troubles relocating: Symbol not found
/boot/system/apps/LibreOffice> warn:sal.osl:1026:1:sal/osl/unx/thread.cxx:1026: RTL_TEXTENCODING_DONTKNOW -> _ASCII_US
warn:configmgr:1026:1:configmgr/source/xcuparser.cxx:289: unknown component "org.openoffice.Office.UI.DbReportWindowState" in "file:///boot/system/apps/LibreOffice/program/../share/registry/res/registry_en-US.xcd"
warn:configmgr:1026:1:configmgr/source/xcuparser.cxx:289: unknown component "org.openoffice.Office.UI.ReportCommands" in "file:///boot/system/apps/LibreOffice/program/../share/registry/res/registry_en-US.xcd"
warn:configmgr:1026:1:configmgr/source/xcuparser.cxx:901: ignoring modify of unknown set member node "StarOffice XML (Base) Report" in "file:///boot/system/apps/LibreOffice/program/../share/registry/res/fcfg_langpack_en-US.xcd"
warn:configmgr:1026:1:configmgr/source/xcuparser.cxx:901: ignoring modify of unknown set member node "StarOffice XML (Base) Report Chart" in "file:///boot/system/apps/LibreOffice/program/../share/registry/res/fcfg_langpack_en-US.xcd"
warn:i18nlangtag:1026:1:i18nlangtag/source/languagetag/languagetag.cxx:1568: LanguageTag::getLanguageFromLangtag: pLangT==NULL for 'UTF-8'
warn:i18nlangtag:1026:1:i18nlangtag/source/languagetag/languagetag.cxx:1626: LanguageTag::getRegionFromLangtag: pRegionT==NULL for 'UTF-8'
[...]

PR #9566, opened 2023-10-03
 libreoffice: enabled x86 build

kenmays:
Tested on hrev57305 x86.
- Passed config.
- Builds without any major issues. Stopped on a missing Qt6 declaration (no issue on x64).
- Review further on buildbot x86.

### DIFF
--- a/app-office/libreoffice/libreoffice-7.3.7.2.recipe
+++ b/app-office/libreoffice/libreoffice-7.3.7.2.recipe
@@ -49,7 +49,7 @@ ADDITIONAL_FILES="
 
 POST_INSTALL_SCRIPTS="$relativePostInstallDir/create_buildid.sh"
 ARCHITECTURES="all !x86_gcc2"
-SECONDARY_ARCHITECTURES="!x86"
+SECONDARY_ARCHITECTURES="x86"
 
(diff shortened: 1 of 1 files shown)


lonemadmax:
Did it build for you? When I tried, it failed due to looking for `moc` in a wrong place. It would find the qt5 version and then error out. `/system/lib/Qt6` is where you find it in x86_64, not in the 32 bit hybrid. Changing that in `configure.ac` to `$qt6_libdir/Qt6` (as the libexec defined in the qt6-base recipe) worked for both.

I would have sent my own PR with that, but I guess there's no point in building and providing a new version until #9219 is fixed?

kenmays:
@lonemadmax - Check with @begasus as he is testing a build of LO 7.x soon on the x86 R1B4 buildbot.

Begasus:
I said I would check a local build, not see what the buildbot will do.
I recently checked a newer version on 64bit, managed to build it but it didn't start, so not pushing untill I (or someone else) has checked it's working as expected (could take some time).

Begasus:
LO on 64bit is still fine, there isn't/wasn't an issue there. For 32bit the issue is known and reported already a while back.

kenmays:
Going to custom build LO fox x86.

PR #9569, opened 2023-10-04
git: add recipe for version 2.42.0

OscarL:
Include bash-completion and prompt scripts.

Also: remove old 2.30.0 recipe.


Begasus:
Keeping a hold on this untill things are clear here.

waddlesplash:
The packages used at build time are stored in a separate repository from the main HaikuPorts one, created by copying/hardlinking packages to "pin" them. Changes can be made at HaikuPorts to remove packages used there without affecting Haiku builds. So that's not a concern.

PR #9579, opened 2023-10-05
ghostscript_gpl, bump version

Begasus:


Begasus:
Would be nice if someone with supported printers could check this, I knew I could connect over wifi, but forgot to write down how I did it. :)

PR #9583, opened 2023-10-06
gnuplot: update to version 5.4.10

OscarL:
Cleaned up unneded REQUIRES.

~Still no GUI output.~ edit: enabled `--with-x`. Bit buggy, but good enough for now.

### DIFF
--- a/sci-visualization/gnuplot/gnuplot-5.4.10.recipe
+++ b/sci-visualization/gnuplot/gnuplot-5.4.10.recipe
@@ -4,11 +4,13 @@ visualize mathematical functions and data interactively, but has grown to \
 support many non-interactive uses such as web scripting. It is also used \
 as a plotting engine by third-party applications like Octave."
 HOMEPAGE="http://www.gnuplot.info/"
-COPYRIGHT="1986-1993, 1998, 2004-2019 Thomas Williams, Colin Kelley"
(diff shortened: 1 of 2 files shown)


waddlesplash:
> and to give time (and something to test) to @waddlesplash and his XLibe :-D

Display fixed in Xlibe master. After fixing that, I saw Cmd+wheel zooming was broken, and fixed that too.

However, click-drag-zoom doesn't work right (draws white boxes and text instead of black so it can't be seen.) Not sure why, I spent a bit of time debugging and didn't figure out the cause.

waddlesplash:
I tested with some GNUplot scripts that seemed to work fine. I didn't try with anything so simple as that, so perhaps I missed something.

waddlesplash:
> running > gnuplot test.gnuplot open and immediately closes a window (too fast to see if something gets actually rendered 

Runing those commands in a gnuplot session here works fine, window draws normally and sticks around.

waddlesplash:
> Opening as draft, while I play with the included code for the be terminal and corresponding gnuplot_be binary

I took a glance at this. Did you get it to work at all? It looks like it might be missing components when they moved the source tree around... I would look for an old tarball back from an early version where this worked and see if there's things missing vs now. Or maybe there's just been too many refactors...

waddlesplash:
I didn't manage to get the `_be` terminal to run, but I did manage to compile it without editing the source code (only the Makefile) under x86_64.

PR #9584, opened 2023-10-06
libdca, disable static library, remove broken link libdts.a

Begasus:


Begasus:
libdts.a linked to libdca.a which isn't around (wasn''t earlier too), so ended up as broken link in the devel package.

PR #9585, opened 2023-10-06
pdflib, bump version

Begasus:


Begasus:
This one is used by the add-on pdfwriter, however I can't seem to print to PDF file (either with current version or this version)?

Begasus:
This version dates back to 2006, newest one I could find is version 7.0.5p3 (link at slackware), dates in there still go back to 2012, so seems like we need to step away from this (if things are broken for pdfwriter), there is already an issue to replace this with other packages providing the same functionality.

Begasus:
Closing for now. Should be addressed with another library.

PR #9587, opened 2023-10-06
mtools, bump version, rename cmd:mmd

Begasus:


Begasus:
Tried renaming the binary to mmd-dos, but launching that failed, so it's not really an option.
Also noticed that conflict is wrong here :) multimarkdown should conflict with this (as on 32bit there is no primary multimarkdown package).

Begasus:
@OscarL renamed the scripts locally to use the .sh extension, issue is that like in bescreencapture you would need to rename the cmd for mmd also to mmd.sh.

PR #9588, opened 2023-10-06
multimarkdown, bump version

Begasus:


app-text/multimarkdown/patches/multimarkdown-6.7.0.patchset
```diff
@@ -0,0 +1,50 @@
...
+-			DESTINATION share/doc/MultiMarkdown
++			DESTINATION ${CMAKE_INSTALL_DOCDIR}/MultiMarkdown
+ 		)
+ 		set (CPACK_COMPONENT_DOCS_DISPLAY_NAME "Documentation")
+ 		set (CPACK_COMPONENT_DOCS_DESCRIPTION "Install README and LICENSE.")
+ 
+ 		# Install LaTeX support files
+ 		install (FILES ${latex}
+-			DESTINATION share/texmf/tex/latex/mmd6
++			DESTINATION ${CMAKE_INSTALL_DATADIR}/texmf/tex/latex/mmd6
```
jmairboeck, line 24: Our texlive packages use `$dataDir/texlive/texmf/...` as a base directory. Maybe this needs an actual additional `/texlive/` in there to actually work.

(I didn't test it though.)

Begasus: Didn't check that part too, I just adjusted the installation path rather then moving the installed (share) to data in the recipe.
But yeah sounds about right. :)

Begasus: Working on it, set the path now to `$dataDir/texlive/texmf-dist/...` (which should reflect the layout for the other .sty files ending up in latex subdirectory).

PR #9597, opened 2023-10-08
hydrogen: fix for correct removing unneeded statiс library

hfsfox:


augiedoggie:
There are still a number of problems with this recipe.  The PROVIDES should have `hydrogen$secondaryArchSuffix`, there are empty directories(like `lib/x86`), there are other directories and files that are not useful on Haiku(like `data/applications`, `data/icons`, `data/metainfo`), and I don't think it needs to have a subfolder inside of $appsDir for just one executable.

PR #9605, opened 2023-10-09
Added pe-bear recipe

Luca1991:
Hi,
this is the recipe for PE-bear a tool for initial analysis of PE files.


app-forensics/pebear/pe_bear-0.6.5.2.recipe
```diff
@@ -0,0 +1,57 @@
+SUMMARY="PE-bear reverse engineering tool"
+DESCRIPTION="Portable Executable reversing tool with a friendly GUI"
+HOMEPAGE="https://hshrzd.wordpress.com/pe-bear"
+COPYRIGHT="2022-2023 hasherezade"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://github.com/hasherezade/pe-bear/archive/refs/tags/v$portVersion.zip"
+CHECKSUM_SHA256="bbd46e4af231ae5257538e8157a077b2fe5843a5dc18beaaaabbae928bfa1025"
+SOURCE_DIR="pe-bear-$portVersion"
+srcGitSubHash1="e2c08098410c85e5546b77a2785444bbc9e0e21b"
```
korli, line 10:
```suggestion
srcGitRev_2="e2c08098410c85e5546b77a2785444bbc9e0e21b"
```

app-forensics/pebear/pe_bear-0.6.5.2.recipe
```diff
@@ -0,0 +1,57 @@
...
+SOURCE_URI_2="https://github.com/hasherezade/bearparser/archive/$srcGitSubHash1.zip"
+CHECKSUM_SHA256_2="4d85d8207285ae5efa7c016aa62606ca155894fee789f028822f57a33117e590"
+SOURCE_FILENAME_2="bearparser-$srcGitSubHash1.zip"
+srcGitSubHash2="61cdc56159986e6d77be571b04c570cf2c378d13"
```
korli, line 14:
```suggestion
srcGitRev_3="61cdc56159986e6d77be571b04c570cf2c378d13"
```

app-forensics/pebear/pe_bear-0.6.5.2.recipe
```diff
@@ -0,0 +1,57 @@
...
+	rm -rf $sourceDir/bearparser $sourceDir/capstone
+	ln -srf $sourceDir2/bearparser-$srcGitSubHash1 $sourceDir/bearparser
+	ln -srf $sourceDir3/capstone-$srcGitSubHash2 $sourceDir/capstone
+	cmake -Bbuild -S. -DCMAKE_BUILD_TYPE=Release
+	make -C build
+}
+
+INSTALL()
+{
+	mkdir -p $binDir $appsDir
```
korli, line 54: why $binDir ?

app-forensics/pebear/pe_bear-0.6.5.2.recipe
```diff
@@ -0,0 +1,57 @@
...
+CHECKSUM_SHA256_3="821ba752993719617ca4c8b546a79dec876c06eedf1e393e86ee4c6899e00099"
+SOURCE_FILENAME_3="capstone-$srcGitSubHash2.zip"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	pe_bear$secondaryArchSuffix = $portVersion
+	app:PE_bear = $portVersion
+	cmd:pe_bear
```
korli, line 25: why cmd:pe_bear?

app-forensics/pebear/pe_bear-0.6.5.2.recipe
Luca1991: Because it can also be launched from terminal (this is how I use it on Linux). Maybe I should just place it in $appsDir?

app-forensics/pebear/pe_bear-0.6.5.2.recipe
Luca1991: Ok, I'll update the recipe within some hours. Thanks.

app-forensics/pebear/pe_bear-0.6.5.2.recipe
Luca1991: @korli sorry I was in a rush this morning and I misunderstood your comment. You are actually right, $binDir is out of place here. My initial idea was to put the binary there too so it can be launched from terminal, but there is really no need. I'll remove it later today. Thanks a lot and sorry for this.

Luca1991:
Hi @korli 
I've implemented all the proposed fixed and I've also added a rdef file.

augiedoggie:
Folder should probably have an underscore in it `pe_bear`.  Also, it seems like this would fit better inside of the `dev-util` directory alongside things like cutter/rizin/patchelf/...

augiedoggie:
> As for the directory: this tool is also used in forensic analysis of malware, but if you think that `dev-util` directory is a better choice, I can move it there. Please let me know. Thanks a lot.

I guess it could go either way.  It wasn't a big issue for me so unless someone else would like it moved then it can stay put in `app-forensics`.

Luca1991:
@augiedoggie I've renamed the directory. 
I'll leave it in `app-forensics` for now, but feel free to change it at anytime.

Begasus:
Gentoo doesn't mention it (at least I can't find it on repology), so for now let's keep it as is. Thanks @Luca1991 !

PR #9606, opened 2023-10-10
graphene, new recipe, dependency for gtk4

Begasus:


Begasus:
Nothing there yet for gtk4, but solves one dependency :)

PR #9611, opened 2023-10-11
harfbuzz, bump version, fix build for pango

Begasus:
… support

### DIFF
--- a/media-libs/harfbuzz/harfbuzz-8.2.1.recipe
+++ b/media-libs/harfbuzz/harfbuzz-8.2.1.recipe
@@ -18,7 +18,7 @@ COPYRIGHT="2004-2018 Behdad Esfahbod
 LICENSE="MIT"
 REVISION="1"
 SOURCE_URI="https://github.com/harfbuzz/harfbuzz/releases/download/$portVersion/harfbuzz-$portVersion.tar.xz"
-CHECKSUM_SHA256="20770789749ac9ba846df33983dbda22db836c70d9f5d050cb9aa5347094a8fb"
+CHECKSUM_SHA256="0fec78f98c9c8faf228957a201c8846f809452c20f8445eb092a1ba6f22dbea5"
 PATCHES="harfbuzz-$portVersion.patchset"
 
 ARCHITECTURES="all !x86_gcc2"
@@ -31,7 +31,7 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
 	commandBinDir=$prefix/bin
 fi
 
-libVersion="0.60730.0"
+libVersion="0.60821.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
 
 PROVIDES="
@@ -42,14 +42,9 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:libbz2$secondaryArchSuffix
 	lib:libfreetype$secondaryArchSuffix
 	lib:libgraphite2$secondaryArchSuffix
-	lib:libiconv$secondaryArchSuffix
 	lib:libicuuc$secondaryArchSuffix
-	lib:libintl$secondaryArchSuffix
-	lib:libpng16$secondaryArchSuffix
-	lib:libz$secondaryArchSuffix
 	"
 
 SUMMARY_glib="$SUMMARY (glib)"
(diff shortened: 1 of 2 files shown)


korli:
So two packages providing devel:libharfbuzz, haikuporter isn't good at interpreting conflicts. So both could end up in the chroot.

Begasus:
This is one solution (and yes thought about the devel:harfbuzz conflict), another solution I started on was to provide a _glib package without the devel:harfbuzz and thus requiring harfbuzz for install, harfbuzz could be installed without the _glib package then and the _glib one would be an "addition" to the harfbuzz one.

Begasus:
OK, cleaned up the work I had prior to this one for the _glib package, still a wip.
At the moment both devel packages provide mostly the same headers, only in the first one the ft and icu headers are included and in the other the gobject headers, maybe only provide the gobject headers in the _glib package and make that devel package also depend on harfbuzz_devel then?

media-libs/harfbuzz/harfbuzz_glib-8.2.1.recipe
```diff
@@ -0,0 +1,121 @@
...
+	rm -rf "$libDir"/libharfbuzz-subset.*
+	rm -rf "$developLibDir"/pkgconfig/harfbuzz.pc
+	rm -rf "$developLibDir"/pkgconfig/harfbuzz-subset.pc
+
+	# devel package
+	packageEntries devel \
+		"$dataDir" \
+		"$developDir" \
+		"$libDir"/cmake \
+		"$libDir"/girepository-1.0
```
Begasus, line 115: Not sure on this one, should the typelib files be kept in the base or devel package?

media-libs/harfbuzz/harfbuzz_glib-8.2.1.recipe
```diff
@@ -0,0 +1,160 @@
...
+		libharfbuzz-gobject
+	fixPkgconfig
+
+	# remove libraries provided in base harfbuzz package
+	rm -rf "$libDir"/libharfbuzz.*
+	rm -rf "$libDir"/libharfbuzz-subset.*
+	rm -rf "$developLibDir"/pkgconfig/harfbuzz.pc
+	rm -rf "$developLibDir"/pkgconfig/harfbuzz-subset.pc
+
+	for i in \
```
Begasus, line 110: This clears $includeDir for the headers provided in the base harfbuzz_devel package, maybe there is a way to list only the ones needed to keep instead of this list?
Headers that need to be kept:
```
hb-cairo.h
hb-gobject.h
hb-gobject-structs.h
hb-gobject-enums.h
hb-glib.h
```

media-libs/harfbuzz/harfbuzz_glib-8.2.1.recipe
Begasus: Keeping it in base, no other package so far moves this to the devel package.

media-libs/harfbuzz/harfbuzz_glib-8.2.1.recipe
```diff
@@ -0,0 +1,160 @@
...
+	# remove libtool library files
+	rm -f "$libDir"/libharfbuzz*.la
+
+	prepareInstalledDevelLibs \
+		libharfbuzz-cairo \
+		libharfbuzz-gobject
+	fixPkgconfig
+
+	# remove libraries provided in base harfbuzz package
+	rm -rf "$libDir"/libharfbuzz.*
```
korli, line 106: But you said libharfbuzz with glib enabled depends on glib, so they are not the same.

Begasus: Another option could be not moving the harfbuzz(icu+subset) libraries to the devel package so we could leave out their devel:* out in the _devel package?

Begasus: Never used it, does it work kinda like CONFLICTS?

Begasus: Changing CONFLICTS to FRESHENS, I see that it pulls in both the base and devel package for harfbuzz (when building pango), if this is what it is suppose to do I could push this change?

korli: FRESHENS is like a REQUIRES for the package manager, I think.

korli:
> I'm tempted to drop the whole idea here to enable glib and go with our initial solution to build pango, that was to move the devel:libharfbuzz_gobject to the subpackage.

Or you can build two times with and without glib, provides all libs one more time in _glib subpackage with FRESHENS.

korli:
> > Or you can build two times with and without glib, provides all libs one more time in _glib subpackage with FRESHENS.
> 
> Still no idea how to do that, nothing has been using this so far? The only donwside would be the missing symbol for pango (that only revealed itself for a check on gtk4).

PROVIDES_glib="
	harfbuzz_glib$secondaryArchSuffix = $portVersion
	lib:libharfbuzz_cairo$secondaryArchSuffix = $libVersionCompat
	lib:libharfbuzz_gobject$secondaryArchSuffix = $libVersionCompat
	"
REQUIRES="
	haiku$secondaryArchSuffix
	harfbuzz$secondaryArchSuffix == $portVersion
	lib:libcairo$secondaryArchSuffix
	lib:libgobject_2.0$secondaryArchSuffix
	"
FRESHENS="
	lib:libharfbuzz$secondaryArchSuffix
	"

same for _glib_devel.

Begasus:
OK, going nuts here :) Another try, this time with 2 builds/installs. Still not sure about FRESHENS (no glib_devel package created).

Begasus:
Still no-go, gtk4 is still getting an error for: `undefined reference to 'hb_glib_script_to_script'`

Begasus:
Will do a quick "fix" for now untill we get this sorted, too many variants for this recipe already and none with a good solution so far.
Let's fix this for the pango build (I'll create a seperate issue for this later).

PR #9651, opened 2023-10-20
frogatto, add icon, Deskbar link

Begasus:


Begasus:
As this package only contains data it should be "any" architecture, however it depends on anura which can't be build with gcc2, so maybe this needs to be changed to "all !x86_gcc2" and enable SECONDARY_ARCHITECTURES="x86"?

PR #9658, opened 2023-10-20
AnotherWorld: added icon

Luca1991:
Hello,
this PR adds an icon to Another World (launcher script).


humdingerb:
Would it have made more sense to create an rdef with the icon and short/long description and application signature. And add the Icon-O-Matic file in the additional-files folder. Makes editing the icon easier if needed.

Begasus:
I've seen it passing by with using HVIF files, but yes, maybe better for future references. :)

humdingerb:
Contrary to HVIF a IOM file keeps the names for shapes/paths/styles. Having the full resources from an rdef has only advantages.

Luca1991:
Hi @humdingerb 
What do you mean by "icon-o-matic file"? Do you mean the original HVIF file?

humdingerb:
In IOM you can "Export" as PNG or HVIF etc. or "Save" in IOM's own format, which isn't as space-efficient as HVIF, but contains meta data like the names of objects (shapes/paths/styles) which makes editing it easier.

PR #9660, opened 2023-10-21
[DO NOT MERGE - BROKEN] Thimbleweed Park (engge2): Added recipe 

Luca1991:
Hi this is a recipe for Thimbleweed Park, based on engge2 (open source reimplementation).
You need original game data.

This is currently broken :( 
I can build the game without any issues outside of haikuporter, but if I try to build this recipe it fails at BUILD stage.
It seems that nimbler can't find nim libs.
I don't know if this is an error on my side (very likely) or if nimbler is bugger when launched from haikuporter.
DUE TO THIS ISSUE, SINCE IT CAN'T PASS THE BUILD PHASE, I CAN'T GUARANTEE THAT THE REST OF THE RECIPE IS OK!!!

So, if it is not working why I'm sending this PR? In the hope that someone more knowlegeable than me with nim/nibler can spot the issue and point me to the right direction.

This also includes an icon in IOM format (as suggested this morning).

Luca


Luca1991:
@Begasus this is an open source reimplementation of thimbleweed park. If this should go to games-engines then also AnotherWorld, HeartOfDarkness and Flashback should go to games-engines.

Begasus:
From both github pages they are announced as engines, so they should be used as such (see anura "engine" needed for frogatto).
This engine is called engge2 and should reflect it's name.
The one for Flashback is called reminiscene and should reflect that name.
IMHO it's misleading as the game itself is called in differently from the engine in both cases.

Luca1991:
@Begasus so we should rename and move all the other (to games-engines)? Is this possibile without causing issues to HaikuDepot?

For example "Another World" should be moved to "NEO-RAW", "Heart of Darkness" to "hode" and so on.

Please let me know, thanks.

Begasus:
Let's leave it for now, we'll wait on what other might have to say here.

hoanga:
> 
> This is currently broken :( I can build the game without any issues outside of haikuporter, but if I try to build this recipe it fails at BUILD stage. It seems that nimbler can't find nim libs. I don't know if this is an error on my side (very likely) or if nimbler is bugger when launched from haikuporter.

just some random thoughts from someone who's worked a little bit with nim in haiku. if i had to make a guess it's probably closer to the latter as using something like the nimble build tool within a haikuporter environment comes with gotchas.

from what i understand of the nimble build tool, it also has the ability to retrieve declared dependencies similar to build tools in other programming language ecosystems such as pip for python or cargo for rust. what this means is there's extra dependencies specified within the .nimble file for any non-trivial nim project that could be dynamically retrieved (ex. from git repositories or directly from something like https://nimble.directory/) when one uses _nimble build_. while this makes it rather convenient if you use the build tool in its typical use cases, i think this might be a bit trickier when used inside of a haikuporter environment.

it's not clear to me if:

1. those dependencies should also be specified in the recipe file 
2. OR if it is expected that it is enough to just include nimble (and maybe nim) commands to build a nim package within a haikuporter environment

my guess would be that it is closer to _1_ some possible venues to check that might be helpful are

* check if there's other haiku recipes that use nimble or nim as a dependency (i could not find any when i checked)
* check how a rust-based recipe or python-based recipe (or other) in haikuporter handle this 
* see if nimble has a way of retrieving dependencies separate from the actual build step itself

Begasus:
Closing for now untill someone is able to fix building with nimble for it, will open an issue for that.

PR #9664, opened 2023-10-22
catch2, bump version

Begasus:


Begasus:
> Is there a package (in-tree) that only requires the `lib:` but not the `devel:` provides from this package?

Not at the moment (from inrecipe nothing is using it either way), did a check on [zug](https://sinusoid.es/zug/) that failed to find catch2 with the included cmake files (pointing to the .a file in previous version). with the shared library and changes it's detected and used.

PR #9671, opened 2023-10-22
deark: new recipe

PocketNerdIO:
Compiled with no patches. Did some tests with some old Psion files and it works fine.

I have only one concern. The developer describes the license as "MIT-style," and then mentions that some files might be under other licenses.

> Starting with version 1.4.x, Deark is distributed under an MIT-style license. See the [COPYING](https://github.com/jsummers/deark/blob/master/COPYING) file for the license text.
>
> The main Deark license does not necessarily apply to the code in the "foreign" subdirectory. Each file there may have its own licensing terms.

I've set `"MIT"` as the `LICENSE` in the recipe as that was the closest option. I've also seen a Gentoo overlay that has also used `"MIT"`. Please let me know if there is a more appropriate option.

### DIFF
--- a/app-misc/deark/deark-1.6.5.recipe
+++ b/app-misc/deark/deark-1.6.5.recipe
@@ -1,0 +1,49 @@
+SUMMARY="Utility to extract data from various file formats"
+DESCRIPTION="Deark is a command-line utility that can decode certain types of files, and either:
+
+1. convert them to a more-modern or more-readable format; or
+2. extract embedded files from them"
+HOMEPAGE="https://entropymine.com/deark/
+	https://github.com/jsummers/deark"
+COPYRIGHT="2023 Jason Summers"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/jsummers/deark/archive/v$portVersion.tar.gz"
+CHECKSUM_SHA256="2aa29675c0b326c5242da2c5e4b1d7746f2aad7bcc9bfefa73448e580e9fae2c"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	deark$secondaryArchSuffix = $portVersion
+	cmd:deark = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	help2man$secondaryArchSuffix
+	"
+BUILD_PREREQUIRES="
+	cmd:help2man
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make$secondaryArchSuffix
+	"
+
+BUILD()
+{
+	make
+	make man
+}
+
+INSTALL()
+{
+	mkdir -p $prefix/bin
+	mkdir -p $manDir/man1
+
+	make DEARK_INSTALLDIR=$prefix/bin install
+	cp deark.1 $manDir/man1
+}


app-misc/deark/deark-1.6.5.recipe
```diff
@@ -0,0 +1,47 @@
+SUMMARY="Utility to extract data from various file formats"
+DESCRIPTION="Deark is a command-line utility that can decode certain types of files, and either: \
+\
+1. convert them to a more-modern or more-readable format; or \
```
Begasus, line 4: No backslashes at the end of the lines here

app-misc/deark/deark-1.6.5.recipe
```diff
@@ -0,0 +1,47 @@
...
+2. extract embedded files from them"
+HOMEPAGE="https://entropymine.com/deark/ https://github.com/jsummers/deark"
```
Begasus, line 6: Start the second link on a new line starting with a tab.

app-misc/deark/deark-1.6.5.recipe
```diff
@@ -0,0 +1,47 @@
...
+
+BUILD()
+{
+	make
+	make man
+}
+
+INSTALL()
+{
+	mkdir -p $binDir
```
Begasus, line 43: Better install in $prefix/bin (otherwise you need `cmd:deark$secondaryArchSuffix` in PROVIDES).

Begasus: Looking at the Makefile, there is an option to set this?:
```
	make DEARK_INSTALLDIR=$prefix/bin
```
https://github.com/jsummers/deark/blob/master/Makefile#L143

PocketNerdIO: Is it good practice to use `$secondaryArchSuffix`? Or does it not matter, and INSTALL just needs to be changed to suit?

Begasus: For packages `not building with gcc2` we prefer to use $prefix/bin (you can find examples when looking for `commandBinDir` here) for the binaries on 32bit, this resolves to have to use $secondaryArchSuffix for binaries, libraries should have it `though` as that path is important for them.

app-misc/deark/deark-1.6.5.recipe
```diff
@@ -0,0 +1,49 @@
...
+
+BUILD()
+{
+	make
+	make man
+}
+
+INSTALL()
+{
+	mkdir -p $prefix/bin
```
Begasus, line 44: Are you sure you still need to manually create $prefix/bin (this should be done with line 47)?

PocketNerdIO: Yeah, I tried it without and it failed.

```
Collecting files to be packaged ...
install -s deark /packages/deark-1.6.5-1/.self/bin
Warning: [Errno -2147459067] Not a directory: '/packaging/deark/bin'
cleaning 'chroot/boot' folder
keeping chroot folder /boot/home/haikuports/app-misc/deark/work-1.6.5 intact for inspection
```

Begasus:
Thanks!

Begasus:
Build failure on 32bit buildmaster :) 
resolving build dependencies failed: build-requires "help2man_x86" of package "deark_x86-1.6.5" could not be resolved

app-misc/deark/deark-1.6.5.recipe
```diff
@@ -0,0 +1,49 @@
...
+	deark$secondaryArchSuffix = $portVersion
+	cmd:deark = $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	help2man$secondaryArchSuffix
```
Begasus, line 27: Not needed:
```diff
-	help2man$secondaryArchSuffix
```

app-misc/deark/deark-1.6.5.recipe
```diff
@@ -0,0 +1,49 @@
...
+	"
+BUILD_PREREQUIRES="
+	cmd:help2man
+	cmd:gcc$secondaryArchSuffix
+	cmd:ld$secondaryArchSuffix
+	cmd:make$secondaryArchSuffix
```
Begasus, line 33: make is not secondary arch:
```diff
-	cmd:make$secondaryArchSuffix
+	cmd:make
```

PocketNerdIO:
OK, fixed those. Should I do a new pull request?

Begasus:
Yes please (clean branch) :) No need for a revbump as we just want to launch a build on 32bit.

PR #9675, opened 2023-10-23
deark: fix for 32-bit

PocketNerdIO:


Begasus:
Little tip in case you bump into this again, when searching if a cmd needs $secondaryArchSuffix or not:
https://github.com/haikuports/haikuporter/blob/master/haikuports-sample.conf#L11

If you have that in your profile you could do a search in your recipes with:` inrecipe cmd:make`

PR #9692, opened 2023-10-28
hydrogen, revbump, 32bit fixes

Begasus:


Begasus:
For libtar, lrdf and rubberband when the build starts there is a summary mentioning: available but not desired
For rubberband it seems only the binary is usefull at runtime. So maybe disable these libraries in REQUIRES/BUILD_REQUIRES?

PR #9695, opened 2023-10-28
npm, bump version

Begasus:

PR #9700, opened 2023-10-28
nodejs, bump to latest LTS version

Begasus:


Begasus:
I'm probably missing something here, building nim works fine, also the build for npm with this version is OK but for npm I get a lot or warnings like this:
```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 close listeners added to [TLSSocket]. Use emitter.setMaxListeners() to increase limit
```

waddlesplash:
Are we on an old version of npm?

Begasus:
> Are we on an old version of npm?

npm has been updated to latest version here: https://github.com/haikuports/haikuports/pull/9695
A comment about nodejs was done there.

waddlesplash:
I think the warning is an internal Node/Npm thing, I don't know that I'd worry about it. We might as well merge.

PR #9702, opened 2023-10-28
netbeans_bin, bump version

Begasus:


dev-util/netbeans/netbeans_bin-19.recipe
```diff
@@ -0,0 +1,76 @@
...
+	cd netbeans
+	rm -rf profiler {cnd,dlight,ide}/bin \
+		platform/modules/lib/{amd64,i386,x86}
+	find . -name "*.dll" -o -name "*.exe" -o -name "*.cmd" \
+		-o -name "*.bat" -o -name ".lastModified" | xargs rm
+
+	sed -i 's,\${HOME}/.netbeans,\${HOME}/config/settings/netbeans,g' bin/netbeans
+	sed -i 's,\${HOME}/.cache/netbeans,\${HOME}/config/settings/cache/netbeans,g' bin/netbeans
+
+	cat << EOF > $appsDir/netbeans/netbeans.sh
```
Begasus, line 49: This was here from previous version, netbeans can be launched from Tracker/Deskbar, so not sure we need to add the launchscript also?

jmairboeck: This sets the `PATH` so that the openjdk tools are part of it. I guess it makes sense to have these. Alternatively, one could install `openjdk17_default` to get the same effect with the default `PATH`.

Begasus: Let's keep it around then :)

PR #9703, opened 2023-10-29
pycharm_community_bin, bump version

Begasus:


Begasus:
Installed the "any" package created on 64bit and installed on 32bit, there it fails to launch:
```
/boot/system/apps/pycharm/bin> pycharm.sh
--
  | Error occurred during initialization of VM
  | Could not reserve enough space for 1536000KB object heap

```
Don't think it's possible to disable `x86_gcc2` together with `any` in `ARCHITECTURES`? So maybe only enable this for `x86_64`? (NetBeans crashes on 32bit, so probably needs the same then).

dev-util/pycharm-community/pycharm_community_bin-2023.2.3.recipe
```diff
@@ -1,14 +1,17 @@
...
 REVISION="1"
 SOURCE_URI="https://download.jetbrains.com/python/pycharm-community-$portVersion.tar.gz#noarchive"
-CHECKSUM_SHA256="10331302819324794e70843a73bfedbc6023778e4374f57f80608983be4e7a11"
+CHECKSUM_SHA256="d59dd88c1eb51cdd756433d415588c573ca944ebf6f08844b8ac8cd2e3d9937b"
 ADDITIONAL_FILES="PyCharm_Logo.hvif"
 
 ARCHITECTURES="any"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	ARCHITECTURES="?any"
+fi
```
Begasus, line 14: This doesn't prevent from being able to check a package from the x86_64 buildmaster, but holds off on creating one for 32bit.

korli: pycharm is unsupported upstream on 32-bit since 2019.

Begasus: Then I guess it's better to use "?all x86_64" for now here (and for NetBeans). 
Come to think of it, I could launch PyCharm without problems, but I couldn't get it to connect to the interpreter, so let's wait a while on this one.

Begasus:
OK, got the interpreter in use inside pycharm, with some quick checks it seems to be working well enough for a noob like me. :)

PyCharm_CE-2023 2 3

PR #9705, opened 2023-10-29
picard: bump version to 2.10

phw:
This time it took me a bit longer for the update, but here is the new Picard 2.10 release. Build tested locally.


Begasus:
Thanks!

PR #9706, opened 2023-10-29
janet, disable cksum, crashes on 32bit

Begasus:


Begasus:
debug report from the 32bit build (crash):
[janet_cksum_debug.txt](https://github.com/haikuports/haikuports/files/13198484/janet_cksum_debug.txt)

PR #9709, opened 2023-10-30
live555, bump version

Begasus:


Begasus:
Cleaned up old (2016! :) ) recipes/patchset, moved to media-plugins (Gentoo layout)
(devel) package only used in VLC (BUILD_REQUIRES) (that should be changed to use `live555${secondaryArchSuffix}_devel` on the next update.
Gentoo and some other (linuxes) provide shared libraries for this, not sure if we want to do the same (given it hasn't been used so far)?

Begasus:
Got a local build with shared libraries, if desired that could be added later.

PR #9711, opened 2023-10-30
steghide, bump version

Begasus:


Begasus:
Not used by others, checked testcase worked out OK without patching (on 64bit), builds fine on 32bit.

PR #9733, opened 2023-11-05
npm, update, make depend on latest nodejs

Begasus:


Begasus:
For the record:
```
ERROR: npm v10.2.1 is known not to run on Node.js v12.22.7.  This version of npm supports the following node versions: `^18.17.0 || >=20.5.0`. You can find the latest version at https://nodejs.org/.
```

Issue #9642, opened 2023-10-18
llvm17: static analyzer scripts don't work


augiedoggie:
The `scan-build` script is looking for `ccc-analyzer` and `c++-analyzer` in /boot/system/bin because of the sed command on [line 693](https://github.com/haikuports/haikuports/blob/master/sys-devel/llvm/llvm17-17.0.1.recipe#L693) but those commands are [installed in `$libExecDir`](https://github.com/haikuports/haikuports/blob/master/sys-devel/llvm/llvm17-17.0.1.recipe#L726) which is /boot/system/lib.  Either the scripts should be moved from /lib to /bin or the sed command should be adjusted.

The `scan-view` script uses a shebang of `#!/usr/bin/env python` but Haiku doesn't provide a `python` command.  It needs a patch or adjustment with sed to use `python3`.

augiedoggie:
Fixed by #9734

PR #9734, opened 2023-11-06
llvm17: fix analyzer scripts

davidkaroly:
this should help with #9642


augiedoggie:
Thanks, I'll check it out later and then close #9642

Issue #9737, opened 2023-11-07
openblas: variable "prefix" not defined in `openblas.pc`


OscarL:
While trying to update the `scipy` recipe, that has a dependency on OpenBLAS (among other things), the meson (build system back end used by scipy) tries to find openblas and fails with the following log:

```
Called: `/bin/pkg-config --modversion openblas` -> 1
stderr:
Variable 'prefix' not defined in '/boot/system/develop/lib/pkgconfig/openblas.pc'
```

Current content of the openblas.pc file:

```
libdir=${prefix}/develop/lib
libsuffix=
includedir=${prefix}/develop/headers
openblas_config= USE_64BITINT= DYNAMIC_ARCH=1 DYNAMIC_OLDER= NO_CBLAS= NO_LAPACK=1 NO_LAPACKE=1 NO_AFFINITY=1 USE_OPENMP=1 PRESCOTT MAX_THREADS=64
version=0.3.23
extralib=-lpthread -lpthread -lgomp
Name: openblas
Description: OpenBLAS is an optimized BLAS library based on GotoBLAS2 1.13 BSD version
Version: ${version}
URL: https://github.com/xianyi/OpenBLAS
Libs: -L${libdir} -lopenblas${libsuffix}
Libs.private: ${extralib}
Cflags: -I${includedir}
```

To compare with other package that does defines "prefix":

```
prefix=/packages/fribidi-1.0.13-1/.self
includedir=${prefix}/develop/headers
libdir=${prefix}/develop/lib

Name: GNU FriBidi
Description: Unicode Bidirectional Algorithm Library
Version: 1.0.13
Libs: -L${libdir} -lfribidi
Cflags: -I${includedir}/fribidi
```

I don't have experience with `pkg-config` and .pc files, so not sure what else besides calling [fixPkgConfig](https://github.com/haikuports/haikuports/blob/2013dcc952759e9740df851f7e75d870c8be091c/sci-libs/openblas/openblas-0.3.23.recipe#L94) is needed to fix this.

Begasus:
Is this what you are looking for? :)
```
~> pkg-config --modversion openblas
0.3.23
```

PR #9738, opened 2023-11-08
Fix missing prefix in pkgconfig files

Begasus:


Begasus:
Should fix: https://github.com/haikuports/haikuports/issues/9737

PR #9743, opened 2023-11-10
icu74, fix checksum, remove SOURCE_URI_3, this was wrong in their int…

Begasus:
…entional release


Begasus:
In their initial release the LICENSE file was a link and hence was broken on download (SOURCE_URI_3).

PR #9749, opened 2023-11-12
qmplay2: bump to 23.10.22

kenmays:
Tested on Haiku R1B4 x64.


Begasus:
Thanks, no crash now when accessing the volume slider. :)

Issue #9741, opened 2023-11-09
Git package includes mksh incompatible shell includes in the default profile


ghost:
When starting a new shell with mksh with git installed this happens:
`Welcome to the Haiku shell.

/bin/mksh: /etc/profile.d/git-completion.sh[251]: syntax error: unexpected '(('
/bin/mksh: /etc/profile.d/git-prompt.sh[166]: syntax error: unexpected '(('
~ `

I don't know when this happened. those files clearly mention beeing for bash/zsh only so should not be added to the deafult profile.d which is for posix compatible shells.

Begasus:
On a sidenote, git bash completion is working fine. :)

pulkomandy:
> this would be overwritten on the next update presumeably?

No, the only implemented policy for global-writable-files conflict resolution is "keep-old", which keeps the old version and never replaces it with a package update. As documented here: https://www.haiku-os.org/docs/develop/packages/BuildingPackages.html

This allows users to modify or replace these scripts. If they were in the git package, it would be impossible to disable them without uninstalling said package, and impossible to customize them as well.

Begasus:
@OscarL I updated git local here to version 2.42.1 without the files in profile.d, are there any benifits from having them in there? Arch moves them to bash-completions? https://gitlab.archlinux.org/archlinux/packaging/packages/git/-/blob/main/PKGBUILD?ref_type=heads#L95

augiedoggie:
> Feel free to move them to the `bash-completions` package as you see it fit (either by downloading them as part of the `bash-completions`, adding them as `additional-files`, or in a patchset, even? :-)
> 
> I planned to just remove these offending files from the next git recipe update (at least to address the breakage I caused), but I'll be glad if you beat me to it :-)

I don't think they need to move to a different package, just a different folder so they don't get sourced by every shell at startup.  There are `bash-completion` dirs where the files should be put instead of the `profile.d` directory.  Like `/system/data/bash-completion/completions`

Begasus:
Gentoo even does it diffferent:
```
	newbashcomp contrib/completion/git-completion.bash ${PN}
	bashcomp_alias git gitk
	# Not really a bash-completion file (bug #477920)
	# but still needed uncompressed (bug #507480)
	insinto /usr/share/${PN}
	doins contrib/completion/git-prompt.sh
```
So far even without both files I haven't seen any downside.

Begasus:
[...]

PR #9762, opened 2023-11-17
git, bump version

Begasus:


Begasus:
Disabled cp'n the 2 files mentioned on the issue now.
https://github.com/haikuports/haikuports/issues/9741

Begasus:
> I think you should just revert the change in the patchset.

Thought about it too yesterday, but was too late for that, will remove that part of the patchset today. :)

Begasus:
Removed obsolete part in the patchset, left in the comment in case someone else is updating it.

PR #9772, opened 2023-11-18
ffmpeg: added ffmpeg 6.1 recipe

kenmays:
Tested on Haiku R1B4 (hrev56768+93) x86_64.


media-video/ffmpeg/ffmpeg61-6.1.recipe
```diff
@@ -0,0 +1,348 @@
...
+	libswscale \
+	"
+# Do not sort alphabetically
+libavutilVersion="58.29.100"
+libavcodecVersion="60.31.102"
+libavformatVersion="60.16.100"
+libavdeviceVersion="60.3.100"
+libavfilterVersion="9.12.100"
+libswscaleVersion="7.5.100"
+libswresampleVersion="4.12.100"
```
jmairboeck, line 35: It looks like these versions are compatible with ffmpeg 6.0, so shouldn't this just update the existing `ffmpeg6` recipe instead?

Besides, if it is going to be kept separately, it should be called `ffmpeg6.1` instead of `ffmpeg61`.

media-video/ffmpeg/ffmpeg61-6.1.recipe
```diff
@@ -0,0 +1,348 @@
...
+and stream audio and video. It includes libavcodec - the leading audio/video \
+codec library."
+HOMEPAGE="https://ffmpeg.org/"
+COPYRIGHT="2000-2003 Fabrice Bellard
+	2003-2023 the FFmpeg developers"
+LICENSE="GNU LGPL v3"
+REVISION="1"
+SOURCE_URI="https://ffmpeg.org/releases/ffmpeg-$portVersion.tar.xz"
+CHECKSUM_SHA256="488c76e57dd9b3bee901f71d5c95eaf1db4a5a31fe46a28654e837144207c270"
+PATCHES="ffmpeg-$portVersion.patchset"
```
jmairboeck, line 12: the patchset is missing

media-video/ffmpeg/ffmpeg61-6.1.recipe
Begasus: Agreed here @jmairboeck there should only be one ffmpeg6 package, no need to sub-version them here.

PR #9775, opened 2023-11-20
nheko, use bundled mtxclient for now (static), undefined references w…

Begasus:
…ith new nlohmann_json

### DIFF
--- a/net-im/nheko/nheko-0.11.3.recipe
+++ b/net-im/nheko/nheko-0.11.3.recipe
@@ -5,7 +5,7 @@ etc.) and less like an IRC client. It was developed using the Qt5 framework."
 HOMEPAGE="https://nheko.im/nheko-reborn/nheko"
 COPYRIGHT="2017-2023 Nheko Authors"
 LICENSE="GNU GPL v3"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://github.com/Nheko-Reborn/nheko/archive/v$portVersion.tar.gz"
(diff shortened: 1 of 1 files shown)


Begasus:
New nlohmann_json is needed for newer (upcomming) nheko so not reverting that, using the bundled mtxclient fixes the build, should be revised on the new version release.

net-im/nheko/nheko-0.11.3.recipe
```diff
@@ -113,4 +119,6 @@ INSTALL()
 
 	addResourcesToBinaries nheko.rdef $appsDir/Nheko
 	addAppDeskbarSymlink $appsDir/Nheko
+
+	prepareInstalledDevelLib libmatrix_client
```
korli, line 123: What's the benefit? The static libs shouldn't be needed at runtime.

Begasus: It comes included as bundled mtxclient to fix the build, should be fine to remove on a version bump for Nheko

Begasus: I could run another check to see if it launches fine building with the bundled one and using our shared library in the mtxclient package.

korli: I mean, the headers and the libs shouldn't be in the mheko package. Please remove them.

Begasus: Was looking into that when checking earlier, should be resolved in https://github.com/haikuports/haikuports/pull/9776

PR #9782, opened 2023-11-22
move prison to kde-frameworks

Begasus:


Begasus:
No version changes, only move according to gentoo layout where it should be.

PR #9787, opened 2023-11-24
gcompris, bump version

Begasus:


Begasus:
Checked on 32bit and 64bit, works fine (couldn't get the background music on 64bit, on 32bit downloading inline worked ok, also voices).

PR #9794, opened 2023-11-28
knewstuff, add missing library

Begasus:


Begasus:
@threedeyes could you run a check, the missing library gave me errors when building libalkimia.

PR #9804, opened 2023-11-29
qca, add build type for cmake

Begasus:


Begasus:
Looks like it's not ready for gcc13 :/ might need an version bump.

PR #9813, opened 2023-12-01
pyqt5, fix references on moving the api

Begasus:


Begasus:
Doing one more check build on 64bit, should be good now :)

PR #9829, opened 2023-12-05
dillo, new recipe

Begasus:
Basic webbrowser, this is mostly to show our FLTK port is up to building it. :)


Begasus:
Kudos to @coolcoder613eb for the initial idea/patch!

PR #9830, opened 2023-12-05
scons, update and cleanup

Begasus:


Begasus:
Not checked on 32bit yet.

PR #9847, opened 2023-12-07
yaml_cpp, bump version, keep older library around

Begasus:


dev-cpp/yaml_cpp/yaml_cpp0.7-0.7.0.recipe
```diff
@@ -17,19 +17,14 @@ libVersion="$portVersion"
...
 	"
-
-PROVIDES_devel="
-	yaml_cpp${secondaryArchSuffix}_devel = $portVersion
-	devel:libyaml_cpp$secondaryArchSuffix = $libVersionCompat
-	"
-REQUIRES_devel="
-	yaml_cpp$secondaryArchSuffix == $portVersion base
+REPLACES="
+	yaml_cpp$secondaryArchSuffix
```
korli, line 27: This should trigger the removal of the yaml_cpp_devel package. If you could confirm, then that's OK

Begasus: In Dutch, but should show it's working as expected :)
"Installeer" = "Install"
"verwijder" = "remove"
yaml_cpp

PR #9852, opened 2023-12-08
pygments, build fix

Begasus:


dev-python/pygments/pygments-2.14.0.recipe
```diff
@@ -34,7 +34,7 @@ BUILD_REQUIRES="
 
 PYTHON_PACKAGES=(python39 python310)
 PYTHON_VERSIONS=(3.9 3.10)
-commandSuffixes=(3.8 "" 3.10)
+commandSuffixes=(3.9 "" 3.10)
```
jmairboeck, line 37: These arrays have different lengths. It should be either `commandSuffixes=(3.9 3.10)` or `commandSuffixes=("" 3.10)` (most likely the latter, but I'm not sure).

jmairboeck: Ah, right, because python3.10 is now the default. So the correct one should be `commansSuffixes=(3.9 "")`.

PR #9855, opened 2023-12-08
pygments: recipe clean ups, fix build.

OscarL:


Begasus:
Importing pygments works fine now with python3.9 and python3.10, thanks!
gtk_doc however still fails to find it: 
```
Building ...
The Meson build system
Version: 1.2.1
Source dir: /sources/gtk-doc-1.33.2
Build dir: /sources/gtk-doc-1.33.2/build
Build type: native build
Project name: gtk-doc
Project version: 1.33.1
C compiler for the host machine: cc (gcc 13.2.0 "cc (2023_08_10) 13.2.0")
C linker for the host machine: cc ld.bfd 2.41
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program python3 found: YES (/bin/python3)
WARNING: You should add the boolean check kwarg to the run_command call.
         It currently defaults to false,
         but it will default to true in future releases of meson.
         See also: https://github.com/mesonbuild/meson/issues/9300

meson.build:45:4: ERROR: Problem encountered: Required Python3 module 'pygments' not found

A full log can be found at /sources/gtk-doc-1.33.2/build/meson-logs/meson-log.txt
Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 1.
cleaning 'chroot/boot' folder
keeping chroot folder /Opslag/haikuports/dev-util/gtk_doc/work-1.33.2 intact for inspection
```
Tried with both python versions. It doesn't need a rebuild as it's still working fine, but still ...

Begasus:
> Edit: or at least change `cmd:python3` to `cmd:python3.9` in the meantime :-)

Already tried that (as mentioned above), in Terminal it's OK, so let's merge this.

PR #9873, opened 2023-12-10
wxqt: update to 3.2.4

davidkaroly:
successful build (incl. demos and samples) on 64-bit and 32-bit
demos look fine

not sure if we have any significant apps in haikuports that relies on this?


Begasus:
> not sure if we have any significant apps in haikuports that relies on this?

https://github.com/search?q=repo%3Ahaikuports%2Fhaikuports%20lib%3Alibwx_&type=code

Atleast I'd like to check the scummvm-tools with this one :)

davidkaroly:
> Atleast I'd like to check the scummvm-tools with this one :)
> 
> **EDIT** strike that, those are using the gtk variant :)

yes that would be exactly the question... do we have any packages that's supposed to work with the Qt-flavor of wxWidgets rather than the GTK-flavor.

Of course it would be cool to run audacity and tenacity on wxQt but it does not seem feasible at this point (unless I do the debugging and contribute it upstream, at least for the case of tenacity, as they already have an almost-working wxqt branch)

pulkomandy:
I think WxQt was ported before we got GTK running. It is not very well supported in the Wx world (despite the promises of WxWidgets to make things independant from the underlying API, large apps will usually need some custom widgets and these can't really be done in a cross platform way).

I think now that we have GTK, it would be OK to retire this if no one plans to work on it?

Begasus:
+1 for me, I checked the same version here for wxgtk, works localy without new patches and scummvm-tools is still fine with it. IIRC @extrowerk did the initial port here?

PR #9876, opened 2023-12-11
libedit: update to version 20230828, remove older recipes.

OscarL:
Opening as Draft, as I need to test this further. At this point, only tested it on beta4 32 bits, and with some programs that use the `_x86` variant (like Python >= 3.10).

---

Edit 1:

Tested it on beta4 64 bits with:

- pcre2test
- python3.10/11/12
- sftp

Those were built with the 20210910 version of libedit, and seem to work fine with this new one.

Edit 2:

The following two work, but line editing is wonky. They might actually need to use `libreadline.so` instead.

- kc
- xorriso (not sure if I tested it right, used: `xorriso -use_readline on -prompt "> "`).

Non-working line-editing:

- sqlite3's .recipe requires libedit, but /bin/sqlite3 doesn't seems to need it (no mention of it in `readelf -d` or listimage when sqlite3 is running). It might be missing some config flags at build time.

Edit 3:

Used the newest version of libedit_devel to build Python 3.12. It build ok, and seems to work just fine. Setting this ready for review.


Begasus:
Thanks for the heads-up, was waiting for a green light here. :)

PR #9881, opened 2023-12-12
mac, add recipe for version 10.29

Begasus:


Begasus:
@enzo1982 added the latest version here, this is only used by fre-ac, I moved the binary to a seperate tool package as from what I can tell only the library is used by fre-ac?
I left the older one alive for now, if it's ok to move completely I can drop the old recipe, and revbump fre-ac for this.

Begasus:
A rebuild for freac didn't seem to pick up libMAC? (doesn't show up in the preferences)

Begasus:
Let's close this for now, no hurry in this. I'll keep the branch alive though.

PR #9884, opened 2023-12-13
hydrogen, revbump for portmidi changes

Begasus:


Begasus:
Some packages are reported broken due to the libVersion change/fix for portmidi, this is one of them.

PR #9885, opened 2023-12-13
audacity, revbump for portmidi changes

Begasus:


Begasus:
Same thing for libVersion patch/fix for portmidi, Audacity (and Tenacity) crash when trying to open the Open dialog.
So not sure if I should leave it enabled.
@davidkaroly maybe you could have a look here?

PR #9893, opened 2023-12-15
xmake, new recipe

cmeerw:
A cross-platform build utility based on Lua

### DIFF
--- a/dev-util/xmake/patches/xmake-2.8.5.patchset
+++ b/dev-util/xmake/patches/xmake-2.8.5.patchset
@@ -1,0 +1,55 @@
+From 5ce1f116e81b981daac2e58d5c3c07557b574214 Mon Sep 17 00:00:00 2001
+From: Christof Meerwald <cmeerw@cmeerw.org>
+Date: Fri, 15 Dec 2023 14:56:01 +0000
+Subject: [PATCH] adjust fallback for Haiku OS
+
+---
+ core/src/xmake/engine.c | 5 +++++
+ 1 file changed, 5 insertions(+)
+
+diff --git a/core/src/xmake/engine.c b/core/src/xmake/engine.c
+index f1b0b381b..7469e7878 100644
+--- a/core/src/xmake/engine.c
++++ b/core/src/xmake/engine.c
+@@ -682,8 +682,13 @@ static tb_size_t xm_engine_get_program_file(xm_engine_t* engine, tb_char_t* path
+         static tb_char_t const* s_paths[] =
+         {
+             "~/.local/bin/xmake",
++#if defined(TB_CONFIG_OS_HAIKU)
++            "~/config/bin/xmake",
++            "/boot/system/bin/xmake"
++#else
+             "/usr/local/bin/xmake",
+             "/usr/bin/xmake"
++#endif
+         };
+         for (tb_size_t i = 0; i < tb_arrayn(s_paths); i++)
+         {
+-- 
+2.42.0
+
+From 6666b40bd8952d198b510dad0c24a912ce66350e Mon Sep 17 00:00:00 2001
+From: Christof Meerwald <cmeerw@cmeerw.org>
+Date: Fri, 15 Dec 2023 21:32:27 +0000
+Subject: [PATCH] disable root user check for Haiku
+
+---
+ xmake/core/main.lua | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/xmake/core/main.lua b/xmake/core/main.lua
+index db284c431..f3412d1d1 100644
+--- a/xmake/core/main.lua
++++ b/xmake/core/main.lua
+@@ -266,7 +266,7 @@ function main.entry()
+     end
+ 
+     -- check run command as root
+-    if not option.get("root") and os.getenv("XMAKE_ROOT") ~= 'y' then
++    if not option.get("root") and os.getenv("XMAKE_ROOT") ~= 'y' and os.host() ~= 'haiku' then
+         if os.isroot() then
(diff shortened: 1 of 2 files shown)


dev-util/xmake/patches/xmake-2.8.5.patchset
```diff
@@ -0,0 +1,54 @@
...
+index f1b0b381b..a24058591 100644
+--- a/core/src/xmake/engine.c
++++ b/core/src/xmake/engine.c
+@@ -682,8 +682,12 @@ static tb_size_t xm_engine_get_program_file(xm_engine_t* engine, tb_char_t* path
+         static tb_char_t const* s_paths[] =
+         {
+             "~/.local/bin/xmake",
++#if defined(TB_CONFIG_OS_HAIKU)
++            "/Haiku/system/bin/xmake"
++#else
```
augiedoggie, line 20: This should be `/boot/system/...`.  The disk isn't guaranteed to be named `Haiku`.

A better option would be to use the `FindDirectory` or `PathFinder` API to search for `xmake` in any bin directory, not just `/system/bin`

dev-util/xmake/xmake-2.8.5.recipe
```diff
@@ -0,0 +1,50 @@
+SUMMARY="A cross-platform build utility based on Lua"
+DESCRIPTION="What is Xmake? \
+1. Xmake is a cross-platform build utility based on the Lua scripting language. \
+2. Xmake is very lightweight and has no dependencies outside of the standard library. \
+3. Uses the \`xmake.lua\` file to maintain project builds with a simple and readable syntax. \
+Xmake can be used to directly build source code (like with Make or Ninja), or it can generate project source files like CMake or Meson. It also has a built-in package management system to help users integrate C/C++ dependencies."
```
augiedoggie, line 6: The first line and the list items don't need the line continuation(i.e. backslash).  The long line after that should be split at or before 100 characters with a continuation.

dev-util/xmake/xmake-2.8.5.recipe
```diff
@@ -0,0 +1,50 @@
...
+HOMEPAGE="https://xmake.io/"
+COPYRIGHT="2015-present TBOOX Open Source Group"
```
augiedoggie, line 8: I don't think having `present` in a copyright is terribly useful.  Should probably be just the start year of `2015` or the full date range of `2015-2023`

dev-util/xmake/xmake-2.8.5.recipe
```diff
@@ -0,0 +1,50 @@
...
+	"
+REQUIRES="
+	haiku
+	lib:libatomic
+	lib:libncurses >= 6
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libncurses >= 6
```
augiedoggie, line 30: This should probably have a `devel:libatomic` build requirement to match the `lib:libatomic` even though the package will probably be active anyhow.

augiedoggie (changes requested):

dev-util/xmake/patches/xmake-2.8.5.patchset
cmeerw: Also got a suggestion on discuss.haiku-os.org for other options here, but that's for another day

dev-util/xmake/xmake-2.8.5.recipe
cmeerw: Changed it to 2023 for now - the present is what the package actually uses (and what's on their homepage)

dev-util/xmake/patches/xmake-2.8.5.patchset
augiedoggie: Yeah, that was also me on the forums :stuck_out_tongue: 

I didn't look close enough at the code to see that it only needed the current exe path.  In that case `get_next_image_info()` is probably the way to go to do it properly.

Depending on what app you're porting you can sometimes get away with reckoning the path using `argv[0]`.  A lot of the SDL games do `chdir(dirname(argv[0]))` so they can find their data directory.

dev-util/xmake/patches/xmake-2.8.5.patchset
```diff
@@ -0,0 +1,54 @@
...
+index f1b0b381b..e1d7a92e7 100644
+--- a/core/src/xmake/engine.c
++++ b/core/src/xmake/engine.c
+@@ -682,8 +682,12 @@ static tb_size_t xm_engine_get_program_file(xm_engine_t* engine, tb_char_t* path
+         static tb_char_t const* s_paths[] =
+         {
+             "~/.local/bin/xmake",
++#if defined(TB_CONFIG_OS_HAIKU)
++            "/boot/system/bin/xmake"
++#else
```
augiedoggie, line 20: As long as you can specify multiple paths to check then you should add `/boot/home/config/bin/xmake`.  Haiku packages can also be installed to the user home directory at `~/config/packages/` which changes the paths. Once that is added I think the recipe is probably ready to go.

cmeerw: added `~/config/bin/xmake` and tested that it works for both cases

augiedoggie:
Thanks, it will probably be a little while before it shows up in HaikuDepot.  The [package builders](https://build.haiku-os.org/) are currently busy rebuilding Qt 5 & 6.

dev-util/xmake/xmake-2.8.5.recipe
augiedoggie: That's just so they don't have to remember to keep updating it every year.  Copyrights are generally assumed to be valid from whatever year they started and extend into the future.

In any case, I don't feel strongly about it, I just think `present` is silly and I don't recall any other recipes doing that.

PR #9898, opened 2023-12-17
libatomic_ops: bump to 7.8.2

ivmai:
Update libatomic_ops recipe to v7.8.2.


Begasus:
boehm_gc and ecl still good with this version?

ivmai:
> boehm_gc and ecl still good with this version? (checked, no breakage)

Yes, boehm_gc is not affected. I don't know about ecl but should not.

Begasus:
Thanks! 👍

PR #9905, opened 2023-12-20
kuserfeedback, new kde framework

Begasus:


Begasus:
Still need to check 32bit, on 64bit it works fine with Kate (checking the general settings there).

Begasus:
Sidenote, it defaults to not sharing any information/feedback, so user still needs to adjust the level of feedback.

korli:
> Sidenote, it defaults to not sharing any information/feedback, so user still needs to adjust the level of feedback.

Must be a KDE default?

Begasus:
> > Sidenote, it defaults to not sharing any information/feedback, so user still needs to adjust the level of feedback.
> 
> Must be a KDE default?

Probably (and good too), checked with KWrite on 64bit, shows the same.

PR #9912, opened 2023-12-21
update luajit 2.1 package

cmeerw:
luajit has switched to a rolling release model using the Unix timestamp of the latest commit as the version number

### DIFF
--- a/dev-lang/luajit/luajit-2.1.1702296283.recipe
+++ b/dev-lang/luajit/luajit-2.1.1702296283.recipe
@@ -3,14 +3,17 @@ DESCRIPTION="Lua is a powerful, dynamic and light-weight programming \
 language. It may be embedded or used as a general-purpose, stand-alone \
 language. LuaJIT is a Just-In-Time Compiler (JIT) for it."
 HOMEPAGE="https://luajit.org/luajit.html"
-COPYRIGHT="2005-2015 Mike Pall"
+COPYRIGHT="2005-2023 Mike Pall"
 LICENSE="MIT"
-REVISION="2"
-SOURCE_URI="https://luajit.org/download/LuaJIT-2.1.0-beta3.tar.gz"
-CHECKSUM_SHA256="1ad2e34b111c802f9d0cdf019e986909123237a28c746b21295b63c9e785d9c3"
-SOURCE_DIR="LuaJIT-2.1.0-beta3"
+REVISION="1"
+# Note: last component of the version is the Unix timestamp of the git commit
+srcGitRev="29b0b282f59ac533313199f4f7be79490b7eee51"
+SOURCE_URI="https://github.com/LuaJIT/LuaJIT/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="e322ab3d39d722fa06076258a495dee8cd11b9fdfbff7608d19b60b6e378e206"
+SOURCE_DIR="LuaJIT-$srcGitRev"
+PATCHES="luajit-${portVersion%.*}.patchset"
 
-ARCHITECTURES="!all x86"
+ARCHITECTURES="all !x86_gcc2"
 SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
@@ -48,17 +51,17 @@ BUILD()
 INSTALL()
 {
 	make install PREFIX=$prefix INSTALL_LIB="$libDir" INSTALL_BIN="$binDir" \
-		INSTALL_INC="$includeDir" INSTALL_SHARE="$dataDir" \
-		INSTALL_MAN="$manDir/man1" INSTALL_LMOD="$dataDir/lua/5.1"
+		INSTALL_INC="$includeDir/luajit-2.1" INSTALL_SHARE="$dataDir" \
+		INSTALL_MAN="$manDir/man1" INSTALL_PKGCONFIG=$developLibDir/pkgconfig \
+		GIT_RELVER=${portVersion##*.}
 
-	ln -sr $libDir/libluajit-5.1.so.2.1.0 $libDir/libluajit-5.1.so.2
+	# remove static library
+	rm $libDir/libluajit-5.1.a
 
-	# the pkg-config is for 5.1 compat
-	mkdir -p $libDir/pkgconfig
-	cp etc/luajit.pc $libDir/pkgconfig/luajit.pc
+	# fix pkgconfig libdir
+	sed -i -e "s,^libdir=.*,libdir=$developLibDir," $developLibDir/pkgconfig/luajit.pc
 
 	prepareInstalledDevelLib libluajit-5.1
-	fixPkgconfig
 
 	packageEntries devel $developDir
 }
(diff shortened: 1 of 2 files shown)


dev-lang/luajit/luajit-2.1.1702296283.recipe
```diff
@@ -3,14 +3,15 @@ DESCRIPTION="Lua is a powerful, dynamic and light-weight programming \
...
-CHECKSUM_SHA256="1ad2e34b111c802f9d0cdf019e986909123237a28c746b21295b63c9e785d9c3"
-SOURCE_DIR="LuaJIT-2.1.0-beta3"
+REVISION="1"
+SOURCE_URI="git+https://luajit.org/git/luajit.git#29b0b282f59ac533313199f4f7be79490b7eee51"
+# version is the Unix timestamp of the git commit
+SOURCE_DIR="luajit-2.1.1702296283"
+PATCHES="luajit-${portVersion%.*}.patchset"
 
-ARCHITECTURES="!all x86"
+ARCHITECTURES="x86_64 x86 arm arm64"
```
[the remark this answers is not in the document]
cmeerw: Wouldn't "all" also include RISC-V, which I don't think is supported by luajit yet?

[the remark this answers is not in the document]
Begasus: Suggested change is fine for now, if things are broken for riscv it will pop up once(?) there is a buildmaster for that (I presume a lot of packages would fail as almost none report them broken for riscv atm). 👍

dev-lang/luajit/luajit-2.1.1702296283.recipe
```diff
@@ -3,14 +3,15 @@ DESCRIPTION="Lua is a powerful, dynamic and light-weight programming \
...
 HOMEPAGE="https://luajit.org/luajit.html"
-COPYRIGHT="2005-2015 Mike Pall"
+COPYRIGHT="2005-2023 Mike Pall"
 LICENSE="MIT"
-REVISION="2"
-SOURCE_URI="https://luajit.org/download/LuaJIT-2.1.0-beta3.tar.gz"
-CHECKSUM_SHA256="1ad2e34b111c802f9d0cdf019e986909123237a28c746b21295b63c9e785d9c3"
-SOURCE_DIR="LuaJIT-2.1.0-beta3"
+REVISION="1"
+SOURCE_URI="git+https://luajit.org/git/luajit.git#29b0b282f59ac533313199f4f7be79490b7eee51"
```
Begasus, line 9: I saw there is a release also at their github page, maybe that could be used instead (or a srcGitRev tag)?

Begasus: Reasoning? unsafe sources are disabled by default so buildmaster wont pick this up for a build:
`Error: Downloading from unsafe sources is disabled in haikuports.conf!`

Begasus: This works (commit is the same):
```
srcGitRev="29b0b282f59ac533313199f4f7be79490b7eee51"
SOURCE_URI="https://github.com/LuaJIT/LuaJIT/archive/$srcGitRev.tar.gz"
CHECKSUM_SHA256="e322ab3d39d722fa06076258a495dee8cd11b9fdfbff7608d19b60b6e378e206"
```

dev-lang/luajit/luajit-2.1.1702296283.recipe
```diff
@@ -48,10 +49,12 @@ BUILD()
 INSTALL()
 {
 	make install PREFIX=$prefix INSTALL_LIB="$libDir" INSTALL_BIN="$binDir" \
-		INSTALL_INC="$includeDir" INSTALL_SHARE="$dataDir" \
-		INSTALL_MAN="$manDir/man1" INSTALL_LMOD="$dataDir/lua/5.1"
+		INSTALL_INC="$includeDir/luajit-2.1" INSTALL_SHARE="$dataDir" \
+		INSTALL_MAN="$manDir/man1" INSTALL_LMOD="$dataDir/lua/5.1" \
+		GIT_RELVER=${portVersion##*.}
```
Begasus, line 54: This fixes the location for the pkgconfig file and the lines 56 and 60 can be removed then:
```diff
-		GIT_RELVER=${portVersion##*.}
+		INSTALL_PKGCONFIG=$libDir/pkgconfig GIT_RELVER=${portVersion##*.}
```

dev-lang/luajit/luajit-2.1.1702296283.recipe
```diff
@@ -3,14 +3,15 @@ DESCRIPTION="Lua is a powerful, dynamic and light-weight programming \
...
+COPYRIGHT="2005-2023 Mike Pall"
 LICENSE="MIT"
-REVISION="2"
-SOURCE_URI="https://luajit.org/download/LuaJIT-2.1.0-beta3.tar.gz"
-CHECKSUM_SHA256="1ad2e34b111c802f9d0cdf019e986909123237a28c746b21295b63c9e785d9c3"
-SOURCE_DIR="LuaJIT-2.1.0-beta3"
+REVISION="1"
+SOURCE_URI="git+https://luajit.org/git/luajit.git#29b0b282f59ac533313199f4f7be79490b7eee51"
+# version is the Unix timestamp of the git commit
+SOURCE_DIR="luajit-2.1.1702296283"
```
Begasus, line 11: Almost forgot, with above suggesting SOURCE_DIR changes to:
```
SOURCE_DIR="LuaJIT-$srcGitRev"
```

dev-lang/luajit/luajit-2.1.1702296283.recipe
cmeerw: I am assuming you meant `$develLibDir` here?

Begasus: $developLibDir, can be used also, would obsolete fixPkgconfig (but you'd need to be sure everything is correct in the pkgconfig file then.

cmeerw: I think it is correct now. `fixPkgconfig` actually breaks the file as it removes the subdirectory from the includeDirectory again (and I think we really want to install the headers into the subdirectory - this way it's also consistent with other lua packages)

cmeerw: Actually, let me check again...

cmeerw: Hmm... so libdir should also point to devel/lib, I guess? It doesn't for lua54 either (so that might be an issue with that package)

Begasus: This is probably where fixPkgconfig steps in, there are a few recipes around that "sed" back the subdirectory after fixPkgconfig.

cmeerw: I am just fixing the libdir myself now (instead of letting fixPkgconfig fix it and then fixing the includedir breakage)

Begasus:
Thanks!

PR #9927, opened 2023-12-29
zxing_cpp, revbump, change compat version

Begasus:


Begasus:
@korli SONAME for this gives "3", but if I leave it at that (libVersion = 2.2.1) I'm having a problem installing another library that needs this, changed it back so it now shows compat >= 2 instead of 3.
Should I handle this differntly?
Terminal info:
```
~> readelf -d /system/lib/libZXing.so.2.2.1
 0x000000000000000e (SONAME)             Library soname: [libZXing.so.3]
```

PR #9954, opened 2024-01-06
sqlite: bump to 3.44.2

kenmays:
Tested on Haiku R1B4 (+94) x64.


Begasus:
Needs to be checked, already got the recipe for that here, iirc some issues from preventing me to push it.

PR #9955, opened 2024-01-06
gnuplot: bumped for rebuild

kenmays:
Bumped for libvpx 1.11.0. Confirmed x86 build.
NOTE: Added gnuplot v6.0.0 recipe - requires libvpx 1.13.1 to install,, can build with libvpx 1.11.0.


Begasus:
Better would be to check a build for "gd" and "gnuplot" (as gnuplot also needs gd). :)
And thus a revbump for both would be needed. That way we can clear the way to drop 1.11.0 in the end.

kenmays:
GD fixed: https://github.com/haikuports/haikuports/pull/9956

sci-visualization/gnuplot/gnuplot-6.0.0.recipe
```diff
@@ -73,7 +87,10 @@ BUILD()
 	aclocal
 	automake
 	runConfigure --omit-dirs binDir ./configure --bindir="$commandBinDir" \
+		--with-bitmap-terminals \
 		--without-cairo \
+		--with-lua=no \
+		--with-qt=qt6
```
Begasus, line 93: qt6 here and qt5 in the requirements?

kenmays: Allowed. Falls back to Qt5 if Qt6 not present during configure.

Begasus: Doesn't explain why you use Qt6 here.

kenmays: Fixed, in case the fallback gets removed in newer src revisions.

PR #9958, opened 2024-01-06
less: bump version

kenmays:
Build tested on: Haiku R1B4 x64


Begasus:
Should be checked on 32bit (gcc2).

kenmays:
Ok. I'll push this for x64 only for now.

PR #9978, opened 2024-01-11
paladin; bump revision

mbrumbelow:
Version bump for Paladin.


mbrumbelow:
I would have no issues with modern GCC use for 32 bit. I don't think GCC2 is still needed.

Begasus:
Isn't Paladin also used for older (read BeOS apps) sources that require gcc2?

mbrumbelow:
@Begasus do we know for a fact that Paladin is used to build GCC2 BeOS applications?

Begasus:
> @Begasus do we know for a fact that Paladin is used to build GCC2 BeOS applications?

As mentioned on IRC, I'm not familiar enough on that subject, just don't want to see it breaking anything there, for what I have used it was to (try) to open some .proj files from back in the days, and/or create Makefile(s) where there weren't.
If that doesn't inflict with those I presume it's OK to go ahead with the switch.

PR #9980, opened 2024-01-11
less: bump version

kenmays:
Migrated to less v643 for Haiku x86/x64. (GCC13).
- Disabled GCC2 build. 
- Added changes for x86 build.


sys-apps/less/less-643.recipe
```diff
@@ -6,15 +6,15 @@ large input files it starts up faster than text editors like 'vi'.
...
-REVISION="2"
+REVISION="1"
 SOURCE_URI="http://www.greenwoodsoftware.com/less/less-$portVersion.tar.gz"
-CHECKSUM_SHA256="f3dc8455cb0b2b66e0c6b816c00197a71bf6d1787078adeee0bcf2aea4b12706"
+CHECKSUM_SHA256="2911b5432c836fa084c8a2e68f6cd6312372c026a58faaa98862731c8b6052e8"
 PATCHES="less-$portVersion.patchset"
 ADDITIONAL_FILES="sysless.in"
 
-ARCHITECTURES="all"
+ARCHITECTURES="all !x86_gcc2"
```
jmairboeck, line 17: missing `SECONDARY_ARCHITECTURES="x86"`

sys-apps/less/less-643.recipe
```diff
@@ -48,18 +48,23 @@ BUILD()
 {
 	autoconf
 	runConfigure ./configure --with-editor=nano \
```
jmairboeck, line 50:
```suggestion
	runConfigure --omit-dirs binDir ./configure --bindir $prefix/bin --with-editor=nano \
```

sys-apps/less/less-643.recipe
```diff
@@ -48,18 +48,23 @@ BUILD()
...
-	make install
+	make install-strip
 
 	syslessDir=$settingsDir/etc
-	mkdir $syslessDir
+	mkdir "$syslessDir"
 	cp "$portDir/additional-files/sysless.in" \
 		"$syslessDir/sysless.src"
-	$binDir/lesskey -o "$syslessDir/sysless" \
+	"$binDir"/lesskey -o "$syslessDir/sysless" \
```
jmairboeck, line 63:
```suggestion
	"$prefix"/bin/lesskey -o "$syslessDir/sysless" \
```

jmairboeck (changes requested):
This needs a few more changes, otherwise it wouldn't be available any more for 32 bit x86.

sys-apps/less/less-643.recipe
Begasus: Nitpicking here ....
We introduced (don't know who started it) :) commandBinDir for these cases, if we stop using these we might as well drop them in the future ...

jmairboeck: I think `$commandBinDir` was intended for more complicated cases than just secondaryArch on x86_gcc2 (maybe for future 32-bit support on x86_64? I don't know).

This was just a "quick and dirty" suggestion with the minimal changes to get it working correctly on x86. Using `$commandBinDir` is probably cleaner.

Begasus: The situation for 64bit secondary architecture would be the same as for 32bit I guess, so I don't think that would explain it (I know I got corrected when not using it in the past) :)

sys-apps/less/less-643.recipe
jmairboeck, line 1: I think, adding a REPLACES for x86_gcc2 would be good too, given that the package name changes from "less" to "less_x86", i.e.:

```
if [ "$targetArchitecture" = "x86_gcc2" ]; then
	REPLACES="less"
fi
```

I'm not sure exactly where that goes. Between REQUIRES and BUILD_REQUIRES?

Begasus:
Missing patches for filename.c and lesskey.c from previous versions?

sys-apps/less/less-643.recipe
Begasus: A REPLACES or CONFLICTS is needed here yes, can't have `/bin/less` for both arches at the same time ...

Begasus:
Only 1 parse error with gcc2, should be an easy fix and obsoletes the need to build this for secondary architecture. (in command.c), this is not including the missing patches from previous comment.

PR #9982, opened 2024-01-13
gawk, fix installation path for shell scripts

Begasus:


sys-apps/gawk/gawk-5.2.2.recipe
```diff
@@ -69,8 +69,9 @@ defineDebugInfoPackage gawk$secondaryArchSuffix \
 BUILD()
 {
 	AUTOCONF=: AUTOMAKE=: \
-		runConfigure --omit-dirs binDir ./configure \
-		--disable-rpath --with-gnu-ld --bindir=$commandBinDir
+		runConfigure --omit-dirs "binDir sysconfDir" ./configure \
+		--disable-rpath --with-gnu-ld --bindir=$commandBinDir \
+		--sysconfdir=$sysconfDir/etc
```
Begasus, line 74: Not sure if I should use `$syconfDir` here or `$settingsDir`, both act the same, thoughts?

jmairboeck: `--sysconfdir=$sysconfDir/subdir` looks like a bug for me. Maybe `$sysconfDir` should actually be `$settingsDir/etc`? (i.e. this is actually a haikuporter bug)

Begasus: Crossed my mind also, but wouldn't that break other packages, not everything is installed in /system/settings/etc ...

Begasus: In this case I don't think it's an error on sysconfdir, from their NEWS.1 file:
```
9. `make install' now installs shell startup files
   $sysconfdir/profile.d/gawk.{csh,sh} containing shell functions to
   manipulate the AWKPATH and AWKLIBPATH environment variables.  On a Fedora
   system, these files belong in /etc/profile.d, but the appropriate location
   may be different on other platforms.
```

PR #9984, opened 2024-01-13
barrier_haiku: new recipe.

OscarL:
Opening as draft for now. I was trying to update `synergy_haiku`, but couldn't make it work with a barrier server running on Windows... then... found @kallisti5's [barrier-haiku](https://gitlab.com/kallisti5/barrier-haiku), and it worked (almost) right away.

Leaving it as draft for now (in case @kallisti5 has some feedback on the recipe), and because I want to see if I can address an issue I found with the settings file monitoring code.

Still..., pretty handy!

Tested (for now) only on nightly (hrev57509) 64 bits (connecting to a Barrier server version 2.4.0-release-330d758b running on Windows 10).


kallisti5:
Barrier seems dead to be honest.  The Synergy codebase has been a roaming project for decades.

https://github.com/input-leap/input-leap is the latest incarnation of it.

PR #9991, opened 2024-01-15
libpcre2: bump to v10.42

kenmays:
Tested on Haiku R1B4 x64.


Begasus:
Would be nice if this was checked with gcc2 too.

waddlesplash:
This change deleted the only working recipe for GCC2 libpcre. Please provide a fix, or else it will be reverted.

kenmays:
See: https://github.com/haikuports/haikuports/pull/10118

PR #9992, opened 2024-01-15
intel_wifi_firmwares: Bump version

kenmays:


waddlesplash:
All changes to these packages usually correspond to changes against the drivers in OpenBSD/FreeBSD which need to be merged into Haiku. Updating them without corresponding driver updates is not likely to fix anything, and may well break things; or, if there are no changes, then there's no need to update them.

I don't see any such changes at present, so bumping these versions is unnecessary.

PR #10012, opened 2024-01-22
nghttp2: bump version

kenmays:
Built on Haiku r1B4 x64.


Begasus:
On my system this will try to uninstall poppler23 and curl, maybe keep the old one without devel package around to avoid this?

korli:
> On my system this will try to uninstall poppler23 and curl, maybe keep the old one without devel package around to avoid this? EDIT: I see it has a common $dataDir, so that won't work, if nothing else depending packages need a revbump

The old recipe provides the same SONAME, so it should work as is.

Begasus:
My bad, it's the devel packages for the others.

PR #10014, opened 2024-01-24
minipro: new recipe

MaximeChretien:
Minipro is an open source software used to control Minipro TL866xxx MCU programmers/logic testers/... (Over 13000 supported chips).

Considering that it is a rather small program, I looked like a nice exercise to make my first recipe. I did my best to follow the documentation and it seems like files are at the right places after install, at least for my understanding.

I tested the basic functionalities of the program, especially the ones using the local data files and it seems to work fine so far.

There's a few things I'm not sure about:
 - Is there any way to support the bash completion files ? I didn't see anything that allows me to put them somewhere relevant.
 - I put `all !x86_gcc2` for the architectures. I believe this program should work everywhere since it only depends on `libusb` although I can only test it on x86_64 so far. As for `x86_gcc2`, I understood that it's only relevant for old BeOS apps, right ?

As I mentionned before, this is my first recipe and I'm also fairly new to Haiku so any feedback is welcome !

### DIFF
--- a/sci-electronics/minipro/minipro-0.6.recipe
+++ b/sci-electronics/minipro/minipro-0.6.recipe
@@ -1,0 +1,61 @@
+SUMMARY="An open source program for controlling the MiniPRO TL866xx programmers"
+DESCRIPTION="Minipro is an Open Source tool intended to become a complete cross-platform
+replacement for the proprietary utility from Autoelectric. Currently it
+supports more than 13000 of target devices - including AVRs, PICs as
+well as a huge number of other microcontrollers and various memory
+chips.
+"
+HOMEPAGE="https://gitlab.com/DavidGriffith/minipro"
+COPYRIGHT="2014-2024 David Griffith"
+LICENSE="GNU GPL v3"
+REVISION="1"
+SOURCE_URI="https://gitlab.com/DavidGriffith/minipro/-/archive/$portVersion/minipro-$portVersion.tar.gz"
+CHECKSUM_SHA256="16b4220b5fc07dddc4d1d49cc181a2c6a735c833cc27f24ab73eac2572c9304a"
+PATCHES="minipro-$portVersion.patchset"
+
+ARCHITECTURES="all !x86_gcc2"
+
+PROVIDES="
+	minipro = $portVersion
+	cmd:minipro = $portVersion
+	"
+REQUIRES="
+	haiku
+	lib:libusb_1.0
+	"
+
+PROVIDES_devel="
+	minipro_devel = $portVersion
+	devel:libminipro = $portVersion
+"
+REQUIRES_devel="$REQUIRES"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libusb_1.0
+	"
+
+BUILD_PREREQUIRES="
+	cmd:ar
+	cmd:cp
+	cmd:gcc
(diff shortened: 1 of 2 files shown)


dev-embedded/minipro/minipro-0.6.recipe
```diff
@@ -0,0 +1,64 @@
...
+PATCHES="minipro-$portVersion.patchset"
+
+ARCHITECTURES="all !x86_gcc2"
+
+PROVIDES="
+	minipro = $portVersion
+	cmd:minipro = $portVersion
+	"
+
+REQUIRES="
```
korli, line 23:
```suggestion
REQUIRES="
```

dev-embedded/minipro/minipro-0.6.recipe
```diff
@@ -0,0 +1,64 @@
...
+	haiku
+	lib:libusb_1.0
+	"
+
+PROVIDES_devel="
+	minipro_devel = $portVersion
+	devel:libminipro = $portVersion
+"
+
+REQUIRES_devel=$REQUIRES
```
korli, line 33:
```suggestion
REQUIRES_devel="$REQUIRES"
```

dev-embedded/minipro/minipro-0.6.recipe
```diff
@@ -0,0 +1,64 @@
...
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libusb_1.0
+	"
+
+BUILD_PREREQUIRES="
+	cmd:ar
+	cmd:cp
+	cmd:gcc
+	cmd:git
```
korli, line 44:
```suggestion
```

korli:
Gentoo suggests sci-electronics instead of dev-embedded.

dev-embedded/minipro/minipro-0.6.recipe
MaximeChretien: `git` is used by the Makefile to generate a version file so I believe it's better to include it in the recipe.
See https://gitlab.com/DavidGriffith/minipro/-/blob/master/Makefile?ref_type=heads#L33.

dev-embedded/minipro/minipro-0.6.recipe
MaximeChretien: Done.

dev-embedded/minipro/minipro-0.6.recipe
MaximeChretien: Done.

MaximeChretien:
> Gentoo suggests sci-electronics instead of dev-embedded.

Alright. I didn't find `minipro` in the gentoo repo so I was not sure about where to put it. I moved it to `sci-elecronics`.

dev-embedded/minipro/minipro-0.6.recipe
pulkomandy: We create a temporary git repository for the build, which does not include the original commit and tags from minipro. So, the generated version will be meaningless and confusing. It's better to not pretend it comes from Minipro's own git repository, and generate the version otherwise (which the Makefile seems to know how to do already).

MaximeChretien: I see, that make sense. I removed it then.

PR #10023, opened 2024-01-28
mc: bump version

hfsfox:


app-misc/mc/patches/mc-4.8.31.patchset
```diff
@@ -0,0 +1,202 @@
+From dfff00db8aa2921ea022f53bb601b2c645311de2 Mon Sep 17 00:00:00 2001
+From: hfsfox <darkkitsunezx128k2309@gmail.com>
+Date: Sun, 28 Jan 2024 15:48:00 +0200
+Subject: add pathces from previous release
```
augiedoggie, line 4: The patch looks the same as before.  You should just rename/move the previous one that lists the authors of the patch and the reason for the changes.

augiedoggie (changes requested):
It's easier to see the changes from each version bump if you move the old recipe and patchset to the new name.

You can delete the old recipe and patchset and push a new change to the pull request and it should be detected as a renamed file.

PR #10024, opened 2024-01-29
utfcpp, new dependency for taglib2.0

Begasus:


Begasus:
This is a headers only library, so I left it as "all" in architectures, it need (doesn't use it afaik) gcc at "build" time, if that's OK I can leave it as is or set it to use secondary architecture.

jmairboeck:
Does secondaryArch gcc have the non-secondaryArch `develop/headers` directory in the default include path? If not, this might need a secondaryArch to install the headers correctly.

dev-libs/utfcpp/utfcpp-4.0.5.recipe
```diff
@@ -0,0 +1,54 @@
...
+	cmd:cmake
+	cmd:gcc
+	cmd:make
+	"
+
+BUILD()
+{
+	cmake -Bbuild -S. $cmakeDirArgs \
+		-DCMAKE_BUILD_TYPE=Release
+	make -Cbuild $jobArgs
```
jmairboeck, line 43: Upstream says that the [CMake file is deprecated](https://github.com/nemtrif/utfcpp/blob/master/CMakeLists.txt) and instead [suggests just copying the header files directly](https://github.com/nemtrif/utfcpp/blob/master/README.md#installation).

Maybe this should just be removed and the header files installed directly with `cp`? Or are the installed CMake files needed downstream?

jmairboeck:
> Does secondaryArch gcc have the non-secondaryArch `develop/headers` directory in the default include path? If not, this might need a secondaryArch to install the headers correctly.

If this does work however, this could even be an "any" package if it contains only header files.

dev-libs/utfcpp/utfcpp-4.0.5.recipe
Begasus: Tried that, just cp the files to $includeDir/... taglib doesn't pick it up then.

Begasus:
> If this does work however, this could even be an "any" package if it contains only header files.

Will do a check on 32bit tomorrow also with taglib.

Begasus:
Changed to "any" architecture, build (and checked on 64bit) and did a check on 32bit with this 64bit 'any' package.

PR #10029, opened 2024-01-31
openssh, bump version

Begasus:


Begasus:
Not checked on 32bit yet, the gcc2 part of the patch doesn't apply anymore so need to test this first.

Begasus:
Turns out the gcc2 patch isn't needed anymore, quick check on 32bit worked out OK.

net-misc/openssh/patches/openssh-9.6p1.patchset
```diff
@@ -131,72 +129,30 @@ index 4f2ad8f..8946d87 100644
...
- is the most recent file that matches:
--.Pa ~/.ssh/id*.pub ,
-+.Pa ~/config/settings/settings/ssh/id*.pub ,
- (excluding those that match
--.Pa ~/.ssh/*-cert.pub )
-+.Pa ~/config/settings/settings/ssh/*-cert.pub )
- so if you create a key that is not the one you want
- .Nm
- to use, just use
+diff --git a/contrib/ssh-copy-id b/contrib/ssh-copy-id
```
korli, line 141: this doesn't look correct. The patch should be applied to contrib/ssh-copy-id.1

Begasus: jikes, thanks for spotting!

Begasus: Fixed, in ssh-copy-id there is still `TARGET_PATH=".ssh/authorized_keys"` shouldn't that be changed too?

jmairboeck: > Fixed, in ssh-copy-id there is still `TARGET_PATH=".ssh/authorized_keys"` shouldn't that be changed too?

The target of ssh-copy-id can be anything, not just Haiku, and for other unix-like systems, this is correct. Ideally, it would have to detect what the target system is and adjust the path accordingly.

This is also a problem when trying to use it from Linux for a Haiku system, so that problem can't be fixed easily, unfortunately.

korli: Thanks, looks good

PR #9778, opened 2023-11-20
openssl3: new recipe

davidkaroly:
I guess this one is long overdue... the recipe might be not the most beautiful and it has seen only limited testing so far but at least it's a start.

Successful build on x86_64, x86_gcc and x86 secondary arch.

There are a few failed test cases from "make test" - I'll try to add more details on that soon.

So far I marked the recipe as untested which more or less covers the current situation, and this way it should not interfere with existing openssl-1.1.1w recipe.

PR #10035, opened 2024-02-01
openssl: added openssl 3.2.1 recipe

kenmays:
Tested and built on Haiku R1B4 x64.

NOTE: Disabled. Await for Haiku R1B5 review. Await full Haikuports integration review (await for Haiku R1B5/OpenSSL 1.1.1 compatibility checks/review). Feel free to help on any further Haiku R1B5 integration patching. 

OpenSSL 3.2.1 is backward compatible with OpenSSL 3.2.0 and OpenSSL 3.1.0, but not with OpenSSL 1.1.1. OpenSSL versions before 1.1.1 are out of support and no longer receiving updates.

OpenSSL 3.2.1 support ends on Nov 2025. OpenSSL 1.1.1 support ended as of Sept 2023.

See: https://dev.haiku-os.org/ticket/18581
See: https://www.openssl.org/docs/man3.1/man7/migration_guide.html


waddlesplash:
There is already #9778, for an older version but it contains the necessary patches (and is called `openssl3` as is best.)

PR #10059, opened 2024-02-06
libpagemaker, revbump for boost version

Begasus:


Begasus:
These revbumps were needed to reduce the required boost version for LO (3 versions got installed).

PR #10087, opened 2024-02-09
grcompiler, drop recipe, no updates after icu66

Begasus:
Can't be build with newer ICU version, no activity upstream for 3 years, repology doesn't give lot's of hits for it anymore.
If no one is using it I guess it's safe to drop it.


korli:
or disable the recipe

Begasus:
> or disable the recipe

Crossed my mind also, still leaves a trail then hunting down recipes using libicuuc >= 66 (but no big deal there). :)

Begasus:
Changed back and disabled.

PR #10093, opened 2024-02-09
postgresql12, revbump, use icu74

Begasus:


dev-db/postgresql/postgresql12-12.0.recipe
```diff
@@ -221,6 +221,6 @@ INSTALL()
 
 TEST()
 {
-	LIBRARY_PATH="$sourceDir/tmp_install/packages/postgresql-$portVersion-$REVISION/.self/lib${LIBRARY_PATH:+:$LIBRARY_PATH}"
+	LIBRARY_PATH="$sourceDir/src/interfaces/libpq${LIBRARY_PATH:+:$LIBRARY_PATH}"
```
Begasus, line 224: Test suite didn't run without this change (although path is correct afaik).

Begasus:
Could use a second look. :)

PR #10097, opened 2024-02-09
noto: add recipe for version 24.2.1

korli:
The default display font in Haiku, the Noto Sans Display variant, is no more.

As a consequence, when testing, the default font switched to Noto Sans Thai.


waddlesplash:
Some details here: https://github.com/notofonts/notofonts.github.io/issues/16

Not sure what to do about this. We could probably just use the old versions; most fonts don't get upgrades every few years anyway...

Begasus:
Guess I can close my branche for this then :)
https://github.com/Begasus/haikuports/commit/db7898a05b8af53f714ec9cc38c44f3828c17338

korli:
I submitted https://dev.haiku-os.org/ticket/18793 to ease the upgrade.

korli:
I updated the recipe to include an old version of NotoSansDisplay. No change required on the Haiku side.

PR #10102, opened 2024-02-12
mumble: new recipe

TheZeldakatze:
This is a pretty hacky recipe I had laying around for a while, I had problems with portaudio, as mumble assumes that portaudio does some resampling, which it doesn't do on Haiku. This patch just passes on the audio regardless of the frequency, which could cause some audio "artifacts" but it does work. It would probably be better though to do some resampling at the portaudio or Haiku level.

Currently does not build (anymore) since the recipe is using the release tarball instead of the git repo clone I tested it on, I'll fix that up when I'm at home


TheZeldakatze:
could someone also test if mumble works on their system? The computer I'm currently using does not have a working microphone input (Haiku has trouble with the Intel HDA and the old SoundBlaster PCI128 I had laying around) and I currently don't have accesss to the computer I tested it with a few days ago.

net-voip/mumble/mumble-1.5.517.recipe
```diff
@@ -0,0 +1,101 @@
...
+	app:Mumble
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libboost_system$secondaryArchSuffix >= 1.83.0
+	lib:libexpat$secondaryArchSuffix
+	lib:libogg$secondaryArchSuffix
+	lib:libopus$secondaryArchSuffix
+	lib:libpcre2_posix$secondaryArchSuffix
+	poco$secondaryArchSuffix
```
Begasus, line 31: Replace with required lib:libraries, in this case it needs:
* libPocoFoundation
* libPocoXML
* libPocoZip

net-voip/mumble/mumble-1.5.517.recipe
```diff
@@ -0,0 +1,101 @@
...
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	mumble$secondaryArchSuffix
+	app:Mumble
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libboost_system$secondaryArchSuffix >= 1.83.0
```
Begasus, line 26: No need for versioning here:
```diff
-	lib:libboost_system$secondaryArchSuffix >= 1.83.0
+	lib:libboost_system$secondaryArchSuffix
```

Begasus:
Don't have any microphone here to test it with. :)

net-voip/mumble/mumble-1.5.517.recipe
Begasus: poco package needs some changes for this to work, there is no $libVersion/$libVersionCompat for this ($libVersion = 94), so using lib:lib* won't work for this.

net-voip/mumble/mumble-1.5.517.recipe
```diff
@@ -0,0 +1,101 @@
...
+	cmd:make
+	cmd:pkg_config$secondaryArchSuffix
+	cmd:python3
+	"
+
+BUILD() {
+	 cmake -Bbuild -Dbundled-opus=OFF -Dbundled-rnnoise=OFF -Dbundled-speex=OFF \
+	 	-Dsymbols=ON -Dserver=OFF -Dclient=ON -Ddbus=OFF -Dzeroconf=OFF -Dtracy=OFF \
+	 	-DTRACY_ENABLE=OFF -Doverlay=OFF -Dtests=OFF -Dplugins=OFF \
+	 	-DCMAKE_BUILD_TYPE=RelWithDebInfo -Dwarnings-as-errors=OFF
```
Begasus, line 75: Could you sort this one line per option, this is heard to read now :)
Also (sugestion), use:
```
cmake -Bbuild -S.
...
make -Cbuild
```

Begasus:
Played around a bit with it, launches fine, had some issue with adding a server (windows not responding), finaly got around to give the IP for localhost etc to the server fields, audio test when launched first time seem OK.
But then I had it minimized in the Deskbar and left clicking on it made it go wild! It multiplied in the Deskbar, screen went flickering ... eventual had to reboot to gain access to the system again, not something for prime-time ... my 2 cents.

net-voip/mumble/mumble-1.5.517.recipe
TheZeldakatze: I had it that way, but I could not install the package as it did not find poco. I'll have a look at the poco recipe

Begasus: Revbumped poco earlier today, you can now use lib:lib* and devel:lib* :)

Begasus: > I could not install the package as it did not find poco.

I had that earlier too, hence the revbump for poco.

TheZeldakatze:
> Played around a bit with it, launches fine, had some issue with adding a server (windows not responding), finaly got around to give the IP for localhost etc to the server fields, audio test when launched first time seem OK. But then I had it minimized in the Deskbar and left clicking on it made it go wild! It multiplied in the Deskbar, screen went flickering ... eventual had to reboot to gain access to the system again, not something for prime-time ... my 2 cents.

Sorry, I have only now noticed it, that's some horrible flickering and I don't know how I missed that. Interrestingly deminimizing it over the context menu, it seems to work fine, I'll have a look at that

About the windows not responding, for some reason, the add server dialog appears behind the server selection screen sometimes, I'll look if that's changeable.

One more thing I noticed is that when mumble had opened the input stream, sound recorder could not access the microphone anymore, but that might be a problem with portaudio

TheZeldakatze:
> But then I had it minimized in the Deskbar and left clicking on it made it go wild! It multiplied in the Deskbar, screen went flickering ... eventual had to reboot to gain access to the system again, not something for prime-time ... my 2 cents.


The flickering _should_ be gone now, I hope. I'm really sorry that I missed that, well knowing that such flickering can cause severe harm especially for photosensitive people. I'll be more careful about such things in the future:


I still have to fix the ordering of the server connection dialouges

net-voip/mumble/mumble-1.5.517.recipe
TheZeldakatze: thank you

Begasus:
Thanks, will give this a spin again shortly. :)

PR #10112, opened 2024-02-16
jasper: bump version

kenmays:
Tested on Haiku R1B4 x64


Begasus:
Checked with gcc2?

kenmays:
Not yet. But, I don't expect any major issue.

PR #10116, opened 2024-02-18
gn, switch to cmd:python3

Begasus:


Begasus:
@korli as the sources are hosted on your repository could you check? The default one runs the tests fine, with the older one there are some failing tests.

Begasus:
Ran a build/test run with python2.7, there the tests pass OK with the older version, some failures in the newer version. So I guess it should be OK to merge. :)

PR #10127, opened 2024-02-20
nodejs20: add unicode support

halamix2:
This PR adds unicode support to nodejs20.
Tested by running `npm test` on `camelcase` package,w hich uses unicode regexp. Works under Haiku master (master haiku and haikuports depots). It does


net-libs/nodejs/nodejs20-20.9.0.recipe
```diff
@@ -56,11 +62,12 @@ BUILD_PREREQUIRES="
 	cmd:ninja
 	cmd:python3
 	cmd:which
+	cmd:pkg_config
```
augiedoggie, line 65: Entries should be sorted alphabetically. Also, `cmd:pkg_config` will need to be changed to `cmd:pkg_config$secondaryArchSuffix` in order to work on 32 bit.  There are two pkg_config packages on 32 bit and nodejs20 needs to have the secondary one active.

augiedoggie (changes requested):

net-libs/nodejs/nodejs20-20.9.0.recipe
halamix2: Done

halamix2:
I might want to compile this under x86 and report back if there are any issues (I've only checked x86_64)

korli:
> I probably should compile this under x86 and report back if there are any issues (I've only checked x86_64)

Don't bother. This isn't enabled on this recipe, and probably unsupported upstream.

PR #10129, opened 2024-02-21
subversion: update to version 1.14.3.

OscarL:
Tested on beta4, 64 bits.

Edit: and built/smoke-tested on 32 bits.


Begasus:
LGTM, checked on both arch's.

PR #10130, opened 2024-02-21
git: update to version 2.43.2.

OscarL:
Switch dependencies for git-svn from alien_svn (obsolete), to subversion.

Built and smoke-tested (bot `git` and `git svn`) on beta4, 64 bits.

Edit: same on 32 bits now.


Begasus:
LGTM, checked on both arch's.

PR #10131, opened 2024-02-21
jsoncpp, revbump disable not used cmd:python

Begasus:


Begasus:
@korli you kept the jsoncpp24 for compat in the last update, dropped jsoncpp22 in this one (not seeing anything on disk for this on both arch's), also disable static library for jsconcpp24.
In case jsoncpp24 isn't needed anymore this could be dropped too?

korli:
@Begasus I can't find commits for rebuild after latest jsoncpp upgrade. But I could find rebuilds after jsoncpp22.

Begasus:
@korli jsoncpp_devel has been disabled 4 years ago, I guess it's safe to drop it by now :)
https://github.com/haikuports/haikuports/commit/726aba35558f267a627f232fb2ea2cae5530e535

korli:
> @korli jsoncpp_devel has been disabled 4 years ago, I guess it's safe to drop it by now :) [726aba3](https://github.com/haikuports/haikuports/commit/726aba35558f267a627f232fb2ea2cae5530e535)

Like I wrote, dropping jsoncpp22 is OK. What do you mean?

Begasus:
Let's keep jsoncpp24 around for a bit. That can wait.

PR #10142, opened 2024-02-24
opencv3, revbump, fixes build

Begasus:


Begasus:
This is still the old version, lapack and ffmpeg errors have been fixed in newer versions, so not patching to bump these to the new version, use newer opencv for that.

PR #10144, opened 2024-02-24
fpc: fix script shebang

### LINKED ISSUE #10140: fpc: bad script shebang during creation
`fpcbuild-3.2.2/fpcsrc/compiler/cscript.pas` in `dev-lang/fpc/work-3.2.2/sources` in line 400 contains
```pascal
{$ifdef BEOS}
   AddStart('#!/boot/beos/bin/sh');
{$else}
  AddStart('#!/bin/sh');
{$endif}
```

My guess is that the if should be changed to `{$ifdef BEOS and not defined(HAIKU)}`

halamix2:
- fixes #10140
- tested on x86_64 master 

Edit: it's still broken

### DIFF
--- a/dev-lang/fpc/fpc-3.2.2.recipe
+++ b/dev-lang/fpc/fpc-3.2.2.recipe
@@ -11,7 +11,7 @@ HOMEPAGE="https://freepascal.org/"
 COPYRIGHT="1993-2021 Florian Klaempfl and others"
 LICENSE="GNU LGPL v3
 	GNU GPL v3"
-REVISION="2"
+REVISION="3"
 SOURCE_URI="https://downloads.sourceforge.net/freepascal/fpcbuild-$portVersion.tar.gz"
 CHECKSUM_SHA256="85ef993043bb83f999e2212f1bca766eb71f6f973d362e2290475dbaaf50161f"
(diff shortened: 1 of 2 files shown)


Begasus:
Checked with hedgewars, detection was fine, in Terminal I was able to build that one, just not yet with haikuporter.

hfsfox:
In haiku libGLU.so does not exist, that libGLU.so.1, programs with  uses GLU broken, so patch need to revert or create symlink in haiku

Begasus:
libGLU.so exists in $developLibDir, can you provide me an example where it fails? I'm happy to put it back in, just haven't found any downside so far (neither working with Lazarus).

Begasus:
> In haiku libGLU.so does not exist

Isn't this the same for libiconv? there is no libiconv.so in $libDir, that's also in $developLibDir.

hfsfox:
> libGLU.so exists in $developLibDir, can you provide me an example where it fails? I'm happy to put it back in, just haven't found any downside so far (neither working with Lazarus).

in /system/lib only libGLU.so.1

If add GLU to modules and try to run program:
`qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
Exception at 00000201F5B4C26B: Exception:
Could not load library: libGLU.so.`

Same situation with libiconv

Begasus:
OK, I'll have another run tomorrow on this and if things work out fine I'll revert the changes for libiconv and libGLU, thanks for stepping in.

PR #10158, opened 2024-02-28
tig: update to version 2.5.8.

OscarL:


pulkomandy:
Color fonts are not supported at all by app_server. So at best you can use Noto Emoji (the monochrome version) which should already be in the font fallback path.

It seems strange in the logs that some, but not all, emojis are working. why are they different?

Begasus:
Current version had the same problem on the emoticon tests and didn't even succesfully passed all tests (no SUMMARY).

PR #10165, opened 2024-02-29
openexr32, add new version

Begasus:


Begasus:
@korli this should be fine, doesn't conflict in base packages and leaves the 2.4* version as default. Good to go?

korli:
Looks good.

PR #10174, opened 2024-03-02
kquickcharts, remove xsltproc dependency

Begasus:


Begasus:
This triggers the build on the buildmasters as previous push missed the package for glslang.

PR #10187, opened 2024-03-04
exempi, new recipe

Begasus:


Begasus:
Building as shared library, using the static library resulted in the same undefined error when building tellico, using the shared library for tellico is fine.

Begasus:
Note, on next bump disable gcc2 build :)

PR #10195, opened 2024-03-07
kdenlive, new recipe

Begasus:


Begasus:
Leave disabled for now untill further testing can be done by someone else.

PR #10199, opened 2024-03-08
pnglite: add a symlink on the _devel package.

OscarL:
A user over IRC pointed out the missing "libpnglite.so".

For reference, this is what Fedora does:

https://src.fedoraproject.org/rpms/pnglite/blob/rawhide/f/pnglite.spec

---

Tested on 32 bits (gcc2 and x86), and 64 bits.

### DIFF
--- a/media-libs/pnglite/pnglite-0.1.17.recipe
+++ b/media-libs/pnglite/pnglite-0.1.17.recipe
@@ -4,13 +4,13 @@ It was created as a substitute for libpng in situations when libpng is more than
 HOMEPAGE="https://sourceforge.net/projects/pnglite"
 COPYRIGHT="2007 Daniel Karling"
 LICENSE="Zlib"
-REVISION="3"
+REVISION="4"
 SOURCE_URI="http://prdownloads.sourceforge.net/pnglite/pnglite-$portVersion.zip"
 CHECKSUM_SHA256="6444b13b9ec5b6f9de8f72513a00870325779e3b05bfcf554edb1ab0c90f5962"
 SOURCE_DIR=""
 
 ARCHITECTURES="all"
-SECONDARY_ARCHITECTURES="x86_gcc2 x86"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	pnglite$secondaryArchSuffix = $portVersion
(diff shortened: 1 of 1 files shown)


pulkomandy:
This seems not correct. The non-suffixed file doesn't need to be in $libDir, only in $developLibDir. And prepareInstalledDevelLibs should have already put it there.

Here is how it works:

When you compile and link something, you use -lpnglite. The linker searches in developLibDir and finds libpnglite.so

The library should have a soname of libpnglite.so.0. So, that's what the linker stores in the executable it is linking.

At runtime, the runtime_loader looks in libDir, and find the file with a matching name (with the .0) and loads the library.

Why do we do this?

It allows to link the library without worrying about a specific version (just add -lpnglite to the linker flags). But then, at runtime, if there is a newer version of the library with a different ABI, it will get a new soname (libpnglite.so.1 for example). So, the two runtime libraries can happily live side by side, and each executable will load the right one.

I see you looked at Linux, but Linux works completely differently: they don't separate $libDir and $developLibdir, and they put everything in the same directory. Also, their equivalent of runtime_loader does not work entirely based on filenames, instead they have a database of which libraries are installed (updated by ldconfig). The two systems are sufficiently different that maybe it's not a great idea to look at how Linux does things, if you don't want to be confused.

media-libs/pnglite/pnglite-0.1.17.recipe
```diff
@@ -54,6 +54,7 @@ INSTALL()
 	mkdir -p $includeDir $libDir
 	cp pnglite.h $includeDir/
 	cp libpnglite.so.0 $libDir/
+	ln -sr $libDir/libpnglite.so.0 $libDir/libpnglite.so
```
Begasus, line 57: Shouldn't this be: `ln -sr $libDir/libpnglite.so.0 $developLibDir/libpnglite.so` ?

Begasus: Strike that, looks OK :)

pulkomandy:
Yes, you're right for dlopen (in its current implementation) you will need the .so without suffix, or patch the code.

It may be a good idea to add a comment in the Python recipes explaining that some software opens the library with dlopen and so the symlink is needed. So, someone doesn't remove it by accident not understanding why it's there.

And maybe on Haiku side, we can fix our dlopen() implementation to try finding matching libraries with a suffix? The documentation from POSIX allows us to do so: "The file argument is used to construct a pathname to the executable object file. If file contains a <slash> character, the file argument is used as the pathname for the file. Otherwise, file is used in an implementation-defined manner to yield a pathname.". So, we can try to add a suffix to the given name in that case.

PR #10200, opened 2024-03-08
mame: bump version

PocketNerdIO:
Just reapplied the same patches. No other changes needed.


Begasus:
@PocketNerdIO maybe disable the recipe on the next bump for 32bit? build failure with exhausted memory on buildmaster, IIRC that was the same problem previous time.

PR #10207, opened 2024-03-11
ECM, bump version

Begasus:


kde-frameworks/extra-cmake-modules/extra_cmake_modules-6.0.0.recipe
```diff
@@ -4,14 +4,14 @@ modules provided by CMake, including ones used by find_package() to find \
...
-HOMEPAGE="https://projects.kde.org/projects/kdesupport/extra-cmake-modules/"
+HOMEPAGE="https://invent.kde.org/frameworks/extra-cmake-modules"
 COPYRIGHT="2010-2024 KDE Organisation"
 LICENSE="BSD (2-clause)"
 REVISION="1"
-SOURCE_URI="https://download.kde.org/stable/frameworks/${portVersion%.*}/extra-cmake-modules-${portVersion}.tar.xz"
-CHECKSUM_SHA256="ee3e35f6a257526b8995a086dd190528a8ef4b3854b1e457b8122701b0ce45ee"
-SOURCE_DIR="extra-cmake-modules-${portVersion}"
-PATCHES="extra_cmake_modules-${portVersion}.patchset"
+SOURCE_URI="$HOMEPAGE/-/archive/v$portVersion/extra-cmake-modules-v$portVersion.tar.bz2"
```
jmairboeck, line 11: Any reason why you switched to a Git archive instead of the release tarball? (i.e. https://download.kde.org/stable/frameworks/6.0/extra-cmake-modules-6.0.0.tar.xz)

Begasus: No specific reason, it's been my main source for checking all the KDE things out. Nothing is written into stone for now, but haven't seen any downside on using these so far. :)

jmairboeck: Archives created from git may not contain all the needed files, especially generated ones, like the `configure` script for autotools-based software. (that isn't a problem for CMake of course). If git submodules are used, they also wouldn't be part of the git archive.

In this case, it probably doesn't matter much, but using a release tarball still feels nicer and more "official".

Begasus: Updated :) Thanks, knew already about sometimes missing configure script, didn't know about the submodules! 👍

Begasus:
Checked with several KF5 recipes, 2 major ones kdevelop and cantor still fine with this version.

PR #10215, opened 2024-03-14
Add gnome-backgrounds receipe

yann64:
gnome-background 42.0 receipe

### DIFF
--- a/x11-themes/gnome-backgrounds/gnome_backgrounds-42.0.recipe
+++ b/x11-themes/gnome-backgrounds/gnome_backgrounds-42.0.recipe
@@ -4,7 +4,6 @@ HOMEPAGE="https://gitlab.gnome.org/GNOME/gnome-backgrounds/"
 COPYRIGHT="Jakub Steiner "
 LICENSE="CCBYSA"
 REVISION="1"
-
 SOURCE_URI="https://gitlab.gnome.org/GNOME/gnome-backgrounds/-/archive/$portVersion/gnome-backgrounds-$portVersion.tar.gz"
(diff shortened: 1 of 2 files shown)


yann64:
I do not understand the above error :
- haikuporter builds the package correctly on my side
- unless I am missing something, there are no trailing spaces int he receipe

Also, the lint-new-receipe.sh is does not seem to be included in haikuporter, shouldn't it be part of haikuporter? That would help me to understand the error.

waddlesplash:
> Granted, there's already two such wallpaper packs on HaikuPorts, but at least one of those contains work by Haiku users (I would just drop the other :-D).

I tend to agree.

Begasus:
@waddlesplash I think you can close without merge too? :) -1 for me also, although I got some KDE wallpapers around, I'm not going to create a recipe for it, these wallpapers can be found on the net.

yann64:
> You have trailing spaces on line 17 of the recipe (right above REQUIRES).
> 
> Most editors allow to show "invisible/special" chars, or also provide a "trim trailing spaces" feature you could use.
> 
> Besides that, your SUMMARY reads: "GNOME default icon theme", not matching the DESCRIPTION, or content. You also have an extra space in before closing double quote on the COPYRIGHT line.
> 
> FWIW (and this is just my opinion as a regular user, mind you), I do not think HaikuPorts should host this. Unless this affects functionality of ported apps... people just wanting to use a particular wallpaper, can just download them from elsewhere.
> 
> Granted, there's already two such wallpaper packs on HaikuPorts, but at least one of those contains work by Haiku users (I would just drop the other :-D).

Asked on IRC if that would make sense to merge this receipe as I suspected adding wallpapers was not an obvious thing, so not surprised by the outcome.
Thanks for the feedback on the receipe anyway, it will be valuable whenever I commit future receipes (not wallpapers thought :-))

PR #10216, opened 2024-03-14
kimageannotator, new recipe

Begasus:


Begasus:
Patch isn't really required, but having dependency for libX11 isn't also. :)

PR #10219, opened 2024-03-15
bacon : new receipe

yann64:
Receipe for BaCon : the Basic to C Converter.


dev-lang/bacon/bacon-4.7.2.recipe
```diff
@@ -0,0 +1,49 @@
...
+different platforms (both 32bit and 64bit), while trying to revive \
+the days of the good old BASIC."
+HOMEPAGE="https://chiselapp.com/user/bacon/repository/bacon/home"
+COPYRIGHT="2009-2023 Peter van Eerten"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://chiselapp.com/user/bacon/repository/bacon/attachdownload/bacon-$portVersion.tar.gz?page=Downloads&file=bacon-$portVersion.tar.gz"
+CHECKSUM_SHA256="619b721c820e1b11314cd7b88da39db3e76d9063605ee98aa35fb3fe635ab06c"
+
+ARCHITECTURES="all"
```
Begasus, line 15: This probably wont work on 32bit, libfltk is secondary arch only there, so needs secondaryArchSuffix and the likes here :)

yann64: Ok, will check over the weekend as I am not yet familiar with that (and will amend the receipe accordingly).

Also got your messages on IRC:

- New version got published today, that is why I missed it. I was just unlucky with the release timing.
- Good point for the menu entry. Will also check on that over the weekend.

yann64:
Went the easy way and ignored the GUI (for now) as I only use cli anyway.

dev-lang/bacon/bacon-4.8.recipe
```diff
@@ -0,0 +1,46 @@
...
+	cmd:make
+	cmd:awk
+	cmd:flex
+	cmd:indent
+	cmd:gcc
+	"
+
+BUILD()
+{
+	runConfigure ./configure --prefix=$binDir --datadir=$dataDir
```
Begasus, line 39: prefix and datadir shouldn't be neededd if you use runConfigure

Begasus:
Build failure for both arch's, source uri changed?

jmairboeck:
The error message suggests that their server certificate is expired or something.

PR #10220, opened 2024-03-16
jam: bump git revision

augiedoggie:


augiedoggie:
It's been a few years since jam has been updated.  The current version doesn't include the compile_commands.json support among other fixes.

sys-devel/jam/jam-2.5_2018_11_21.recipe
jmairboeck, line 1: Shouldn't the date part of the version be adjusted?

augiedoggie: I'm not opposed to changing the version but it's been like this for a long time(6 years) with other updates to it since then, as you can see by the REVISION.

augiedoggie: I'll probably go ahead and change it since `jam -v` reports a different date, although the old version didn't match the `jam -v` output anyhow.

augiedoggie:
Funny, I never noticed the permissions were wrong before, but now I see that it's like that even on the copy of `jam` that I use on Linux.  Maybe someone should add a `chmod` to the Jamfile.

korli:
I noticed when trying the build with a secondary user.

PR #10226, opened 2024-03-16
tcl: bump version to 8.6.14

jmairboeck:
Add gcc2 fixes to the new version, remove the old version 8.6.11

Move tclooConfig.sh to the devel package, similar to tclConfig.sh


jmairboeck:
Test report: same as before (see #7721), some tests are hanging and need to be killed manually.

dev-lang/tcl/tcl-8.6.14.recipe
```diff
@@ -9,24 +9,17 @@ COPYRIGHT="Regents of the University of California, Sun Microsystems, Inc., Scri
...
-if [ "$targetArchitecture" = x86_gcc2 ]; then
-	commandSuffix=
-	commandBinDir=$prefix/bin
-fi
-
 PROVIDES="
 	tcl$secondaryArchSuffix = $portVersion compat >= 8.6
-	cmd:tclsh$commandSuffix = $portVersion compat >= 8.6
-	cmd:tclsh8.6$commandSuffix = $portVersion compat >= 8.6
+	cmd:tclsh$secondaryArchSuffix = $portVersion compat >= 8.6
```
jmairboeck, line 21: Should one of the packages (x86_gcc2, x86) not provide tclsh, i.e. remove it, so that we don't have both?

Begasus: Is there any advantige on having a gcc2 package? Otherwise mark it as "?" ?

jmairboeck: There was one before (version 8.6.11) and it wasn't that much effort to fix, as you can see in the patchset. I was just wondering what to do with tclsh. Another possibility would be to split it and make a separate `_tools` package.

Begasus: Imho I would stick with only providing the binaries for secondary architecture as you suggested in the first comment and remove them for gcc2 (if they aren't required for anything else). But that's just me :)

jmairboeck:
> Does the test runner provides a way of excluding those? (if not... would `rm problematic_tests` be viable without needing to patch things, for example?).

I haven't really investigated how to improve the test situation yet, maybe I should do that ...

Thanks for the suggestions!

PR #10228, opened 2024-03-17
IntelliJ: BUGFIX: Fix launcher script execution

cyrusbuilt:
Observation: When launching IntelliJ, "idea.sh" appears in the DeskBar. The 'bash' process executes and immediately pegs a core at 100%, and remains that way indefinitely. The JVM never executes and therefore, IntelliJ never launches.

Manually executing the launcher script ('idea.sh') from the Terminal succeeds. It appears (as discussed with @augiedoggie) the launcher script doesn't like being executed via a symlink in Tracker/DeskBar.

This PR modifies the execution to launch background which seems to work fine. This same issue also applies to PyCharm (I intend to open a separate PR for that). This is not an ideal solution, but is at least one that seems to work fine. Certainly open to suggestions if anyone has a better idea.


korli:
This rings a bell, but can't remember where/when.
I see in ancestris the [following](https://github.com/haikuports/haikuports/blob/53a5526dd9055b5bdcaac2f84dcbc07f5bafc667/app-misc/ancestris/ancestris-11.recipe#L39) is used:
`exec \$(dirname \$(readlink -f \$0))/bin/ancestris`

cyrusbuilt:
@korli I'll try that

korli:
I've no better idea than your fix ATM.

cyrusbuilt:
@korli Fair enough. I discarded my changes and I'll leave this PR as-is. This same fix is working for PyCharm too, so I'll get a PR opened for that shortly. Thanks!

Begasus:
Newer version is available according to Repology, 2023.3.6, maybe bump it while you are at it? :)

cyrusbuilt:
@Begasus I'll certainly give it a shot!

cyrusbuilt:
@Begasus Done. Works well so far!

PR #10230, opened 2024-03-19
fpc, revbump, re-add fix for libGLU

Begasus:


Begasus:
@hfsfox could you run some checks on this? Did local checks with builds for hedgewars and lazarus, still working fine.

hfsfox:
with correct library name works fine, for example try to add to start project that modules:https://github.com/Den3D/SrcExLazarusOpenGL/blob/63614778c3d000fad0d051d886df6ce09fa26f6f/LazGL_15_Game%20Snake2D/LazGL.lpr#L6

Begasus:
Tried building it in Lazarus but running into errors, I don't think it's related to libGLU, can you check on your side that these changes work OK for you?
error

PS, I got latest CudaText build fine with Lazarus, so that shouldn't be an error.

Begasus:
> P.S.:And for lazarus may be better build with Qt6 binding library

Thanks for checking, I already did a local check with Qt6 that works fine, I can push this for the wip PR on Lazarus.

PR #10232, opened 2024-03-19
PyCharm: BUGFIX: Fix launcher execution from DeskBar

cyrusbuilt:
Same issue as PR #10228 

This allows the PyCharm launcher script to properly execute from DeskBar.


Begasus:
Thanks, can confirm launching from Deskbar works fine, on a local build however launching it from the HaikuDepot "app" it crashes the first time, after that opening it with the HaikuDepot "app" it launches fine.

Begasus:
Newer 23023.3.5 available upstream also. Maybe bump version in one go?

cyrusbuilt:
@Begasus yeah I'll give this one a shot too. Maybe I'll get lucky and it will fix some of the intermittent crashes I've been seeing on startup (which seems to stem from the Java Swing library in use)

cyrusbuilt:
@Begasus Done. This also seems to fix the intermittent crashing I was getting on startup. I was able to launch 5 times so far without crash (first launch asked me to import prior settings, then quit and I had to relaunch). Whereas, it used  to crash at least one out of 3 times.

Begasus:
Thanks @cyrusbuilt , my guess would be that since it's a pretty quickly update after prior release it was a bugfix release. :)

PR #10235, opened 2024-03-19
ghostscript: bump version

jmairboeck:


jmairboeck:
no, just 64 bit, but I don't expect any problems there either.

PR #10238, opened 2024-03-22
tig: fix build.

OscarL:
`make install-doc-man` is just too unreliable, for some reason.

### DIFF
--- a/dev-vcs/tig/tig-2.5.8.recipe
+++ b/dev-vcs/tig/tig-2.5.8.recipe
@@ -63,7 +63,12 @@ BUILD()
 INSTALL()
 {
 	make install
-	make install-doc-man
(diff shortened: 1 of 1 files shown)


dev-vcs/tig/tig-2.5.8.recipe
```diff
@@ -63,7 +63,12 @@ BUILD()
 INSTALL()
 {
 	make install
-	make install-doc-man
+
+	# temporary solution because `make install-doc-man` fails/works randomly.
```
Begasus, line 67: Works with cmd:xmlto and cmd:asciidoc as mentioned on IRC, also mentioned it doesn't bring anything new, so let's go this way. :)

PR #10251, opened 2024-03-24
autoconf: update to version 2.72

jmairboeck:
retain version 2.71 for compatibility, similar to the existing older versions.

Questions for the experts:
 - do we need to keep version 2.71?
 - the current, version-less version (2.71 before, 2.72 now) uses `compat >= 2.60`. Should that be changed?

### DIFF
--- a/dev-build/autoconf/autoconf-2.72.recipe
+++ b/dev-build/autoconf/autoconf-2.72.recipe
@@ -1,0 +1,78 @@
+SUMMARY="A tool to automatically configure source code"
+DESCRIPTION="GNU Autoconf is a tool for configuring source code and \
+makefiles. Using autoconf, programmers can create portable and configurable \
+packages, because the person building the package is allowed to specify \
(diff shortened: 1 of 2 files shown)


jmairboeck:
Test report:

autoconf 2.71 (the renamed version) test results:

ERROR: 537 tests were run,
197 failed (4 expected failures).
62 tests were skipped.

autoconf 2.72 test results:

ERROR: 542 tests were run,
199 failed (4 expected failures).
67 tests were skipped.

Begasus:
LGTM

jmairboeck:
There may indeed be some incompatibilities in autoconf 2.72, see for example https://developers.redhat.com/blog/2023/02/23/autoconf-272-broke-my-build-how-do-i-fix-it

However, if some problems arise in the future, the affected packages can be switched to autoconf_2.71 instead, which is still available.

The question about compat versions is still open, but I'll still merge this now.

PR #10253, opened 2024-03-25
fpc, revbump for 32bit

Begasus:


Begasus:
Let's try this again, checked on 2 R1B4 32bit installations, both build the package fine, not sure why it failed (should fail) on the buildmaster.

PR #10255, opened 2024-03-25
qt6_location, new recipe

Begasus:
Required for latest NeoChat.


dev-qt/qt6-location/qt6_location-6.6.0.recipe
```diff
@@ -0,0 +1,102 @@
+SUMMARY="A family of player software for 2d vector graphics animations"
```
jmairboeck, line 1: This doesn't look quite right.

Begasus: Hmm, not sure where I got that :) will  have to run some searches for it (they don't always supply a README with sufficiant information)

PR #10260, opened 2024-03-26
Added batchrename-0.2.0 recipe

axeld:
It's a Tracker add-on, and stand-alone application that I wrote a couple of years back, as I wasn't happy with the existing offerings. It's actual name is just 'rename', but that one is already taken by bbjimmy's add-on.

### DIFF
--- a/haiku-apps/batchrename/batchrename-0.2.0.recipe
+++ b/haiku-apps/batchrename/batchrename-0.2.0.recipe
@@ -1,0 +1,48 @@
+SUMMARY="An application and Tracker add-on to rename multiple files at once"
+DESCRIPTION="batchrename is a standalone application, and Tracker add-on for Haiku. \
+You can rename multiple files in multiple directories at once, and also shows \
+a preview. It offers many renaming options, and can also rename files by contents \
+of their attributes, or the result of shell scripts in case you need that."
+HOMEPAGE="https://github.com/axeld/rename"
+COPYRIGHT="2019-2024 pinc Software"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/axeld/rename/archive/39a87c5d5127544f831e89fce91c2b33193cc103.tar.gz"
+CHECKSUM_SHA256="085bd2ceb244b6bd44e19903fe5efc08567bb05d42aa0e1dd88f554b756ec58a"
+SOURCE_DIR="rename-39a87c5d5127544f831e89fce91c2b33193cc103"
+
+ARCHITECTURES="all"
+
+USER_SETTINGS_FILES="\"settings/pinc.rename settings\""
+
+PROVIDES="
+	batchrename
+	cmd:batchrename= $portVersion
+	app:batchrename= $portVersion
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
+	haiku_devel
+	devel:libicuuc
+	"
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:g++
+	cmd:make
+	"
+
+BUILD()
+{
+	make $jobArgs BUILDHOME=`finddir B_SYSTEM_DEVELOP_DIRECTORY`
+}
+
+INSTALL()
+{
+	mkdir -p $binDir
+	cp batchrename $binDir/batchrename
+	mkdir -p $addOnsDir/Tracker
+	ln -sf $binDir/batchrename $addOnsDir/Tracker/Batch\ rename
+}


haiku-apps/batchrename/batchrename-0.2.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+HOMEPAGE="https://github.com/axeld/rename"
+COPYRIGHT="2019-2024 pinc Software"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/axeld/rename/archive/b768c8525dbaf15b055ffd0b303b8c79a17d8c9a.tar.gz"
+CHECKSUM_SHA256="d94bd208ca03f46f58ed4daef4a044810630a86df60cf1431912375116a5f229"
+SOURCE_DIR="rename-b768c8525dbaf15b055ffd0b303b8c79a17d8c9a"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
```
jmairboeck, line 15: Do Tracker add-ons even work for a secondary architecture?

haiku-apps/batchrename/batchrename-0.2.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+	cmd:rename= $portVersion
+	app:rename= $portVersion
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	icu74${secondaryArchSuffix}_devel
```
jmairboeck, line 30: Usually, references to library names are used instead of package names, so something like `devel:libicuuc$secondaryArchSuffix >= 74`. But this is just a cosmetic change in this case.

haiku-apps/batchrename/batchrename-0.2.0.recipe
```diff
@@ -0,0 +1,49 @@
...
+	"
+
+BUILD()
+{
+	make $jobArgs BUILDHOME=`finddir B_SYSTEM_DEVELOP_DIRECTORY`
+}
+
+INSTALL()
+{
+	mkdir -p $binDir
```
jmairboeck, line 45: `$binDir` has the secondary architecture appended, so it is `$prefix/bin/x86` on x86, and binaries are symlinked to `$prefix/bin` with `-x86` appended.

You should be using `$prefix/bin` here instead, or append `$secondaryArchSuffix` to the `cmd:rename` PROVIDES above (the former is preferred).

haiku-apps/batchrename/batchrename-0.2.0.recipe
axeld: Oops, good point. I actually was too lazy to look into why it wouldn't build anymore with GCC2. Please hold off of the recipe until I fixed that, I completely forgot about this.

haiku-apps/batchrename/batchrename-0.2.0.recipe
axeld: I actually only need the headers, but if your version work, then it would definitely be preferred, as I'm pretty agnostic about the actual version. Thanks for the suggestion!

haiku-apps/batchrename/batchrename-0.2.0.recipe
axeld: Sorry, I don't get this. As it is now (as far as I understood), the binary "rename" would end up in bin/x86, and could therefore be started with just "rename" from a Terminal, right? That's at least what I wanted to achieve here.

axeld:
Are you, by any chance, have an idea why USER_SETTINGS_FILES would not work? When I left it in, the created .PackageInfo could not be parsed, and it would put quotes around the wrong part.

augiedoggie:
> Are you, by any chance, have an idea why USER_SETTINGS_FILES would not work? When I left it in, the created .PackageInfo could not be parsed, and it would put quotes around the wrong part.

The space in the path probably needs to be escaped.  `USER_SETTINGS_FILES="settings/pinc.rename\ settings"`

haiku-apps/batchrename/batchrename-0.2.0.recipe
augiedoggie: If the command is in `$binDir` then the command would be `rename-x86` unless you used the `setarch x86` command and then you would be able to just use `rename`.  If you always want it to be available as `rename`, regardless of `setarch` status, then it needs to be in `$prefix/bin`

augiedoggie: Of course, if you're going to fix gcc2 support then you won't be needing the secondary arch and using `$binDir` is fine in that case.

axeld:
I actually tried that, but with the input string of "settings/pinc.rename\ settings", it just put the '\' in the final string like this: {"settings/pinc-rename\" settings}

axeld:
Great, thanks, I will try that!

haiku-apps/batchrename/batchrename-0.2.0.recipe
jmairboeck: If the version doesn't matter, you can also leave out the version constraint then. For gcc2, there is only icu-57 anyway (the package name has no version suffix). With the `devel:` syntax, you don't need to differentiate that then.

haiku-apps/batchrename/batchrename-0.2.0.recipe
pulkomandy: Axel is right here, the binary will end up in bin/x86 and be accessible in the path. There is no problem here.

However, for programs that don't build with gcc2, we have started to put the gcc13 version directly in bin/ anyways, as it is simpler and there is no reason to use the x86/ subdirectory in that case.

jmairboeck: @pulkomandy Thanks for the clarification. I wasn't aware that `bin/x86` is in the PATH by default (I don't use 32 bit much). But the extra suffixed symlink would still be unneeded.

augiedoggie: > @pulkomandy Thanks for the clarification. I wasn't aware that `bin/x86` is in the PATH by default (I don't use 32 bit much). But the extra suffixed symlink would still be unneeded.

It's not in my $PATH on either of my 32 bit VMs

axeld:
I've updated the code base to compile with GCC 2 again, and fixed the remaining issues in the recipe, too. At least I hope so :-) It should be ready to go now.

augiedoggie:
I got a build failure when trying to build on x86_64 nightly.

```
RenameWindow.cpp: In constructor 'RenameWindow::RenameWindow(RenameSettings&)':
RenameWindow.cpp:141:48: error: call of overloaded 'SetVisibleItem(long int)' is ambiguous
  141 |         fCardView->CardLayout()->SetVisibleItem(0L);
      |         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
In file included from /boot/system/develop/headers/os/interface/LayoutBuilder.h:11,
                 from RenameWindow.cpp:23:
/boot/system/develop/headers/os/interface/CardLayout.h:19:57: note: candidate: 'void BCardLayout::SetVisibleItem(int32)'
   19 |                         void                            SetVisibleItem(int32 index);
      |                                                         ^~~~~~~~~~~~~~
/boot/system/develop/headers/os/interface/CardLayout.h:20:57: note: candidate: 'void BCardLayout::SetVisibleItem(BLayoutItem*)'
   20 |                         void                            SetVisibleItem(BLayoutItem* item);
      |                                                         ^~~~~~~~~~~~~~

```

Also, I don't think the old BeOS method of appending a shortcut like `-R` to the Tracker add-on filename works any longer.  Perhaps you could switch it to the "BEOS:default_shortcut" resource. When I built it on 32 bit, Tracker always showed  "Batch rename-R" with no shortcut.

augiedoggie:
Just another side note, the `util_linux` package also provides a `rename` command.

axeld:
I've renamed the executable to 'batchrename', too, and fixed the compiling issue on x86-64. However, I haven't got the time to really test those changes, and I haven't tried building it on x86-64 yet, either. I will probably find the time to do this tomorrow, though.

Begasus:
Could use a nice icon :)

haiku-apps/batchrename/batchrename-0.2.0.recipe
```diff
@@ -0,0 +1,48 @@
...
+You can rename multiple files in multiple directories at once, and also shows \
+a preview. It offers many renaming options, and can also rename files by contents \
+of their attributes, or the result of shell scripts in case you need that."
+HOMEPAGE="https://github.com/axeld/rename"
+COPYRIGHT="2019-2024 pinc Software"
+LICENSE="MIT"
+REVISION="1"
+SOURCE_URI="https://github.com/axeld/rename/archive/39a87c5d5127544f831e89fce91c2b33193cc103.tar.gz"
+CHECKSUM_SHA256="085bd2ceb244b6bd44e19903fe5efc08567bb05d42aa0e1dd88f554b756ec58a"
+SOURCE_DIR="rename-39a87c5d5127544f831e89fce91c2b33193cc103"
```
Begasus, line 12: Suggestion, as this is a commit checkout we mostly use the variable srcGitRev to define the tag and rename the recipe to reflect that it's a git checkout: eg `batchrename-0.2.0~git.recipe`
```diff
-SOURCE_URI="https://github.com/axeld/rename/archive/39a87c5d5127544f831e89fce91c2b33193cc103.tar.gz"
-CHECKSUM_SHA256="085bd2ceb244b6bd44e19903fe5efc08567bb05d42aa0e1dd88f554b756ec58a"
-SOURCE_DIR="rename-39a87c5d5127544f831e89fce91c2b33193cc103"
+srcGitRev="39a87c5d5127544f831e89fce91c2b33193cc103"
+SOURCE_URI="https://github.com/axeld/rename/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="085bd2ceb244b6bd44e19903fe5efc08567bb05d42aa0e1dd88f554b756ec58a"
+SOURCE_FILENAME="batchrename-$portVersion-$srcGitRev.tar.gz"
+SOURCE_DIR="rename-$srcGitRev"
```

Begasus: On a future update only the srcGitRev needs to be updated then.

Begasus:
Build and Tracker addon working OK on 32bit and 64bit.

haiku-apps/batchrename/batchrename-0.2.0.recipe
axeld: Thanks for testing, and the suggestions! That makes sense!

PR #10261, opened 2024-03-26
texlive_core & texlive: update to version 2024

jmairboeck:
`texlive_core` is done.

I just started the first build of the big `texlive` package.


app-text/texlive/texlive-2024.recipe
```diff
@@ -196,8 +195,6 @@ GLOBAL_WRITABLE_FILES="
 
 PATCH() {
 	sed -i "s|\`kpsewhich -var-value=TEXMFROOT\`|\"$dataDir/texlive\"|" $sourceDir/tlpkg/TeXLive/TLUtils.pm
-	sed -i "s|\`kpsewhich -var-value=FC_CACHEDIR\`|\"$sharedStateDir/texlive/fonts/cache\"|" $sourceDir3/tlpkg/tlpostcode/xetex.pl
-	sed -i "s|\`kpsewhich -var-value=FONTCONFIG_PATH\`|\"$sharedStateDir/texlive/fonts/conf\"|" $sourceDir3/tlpkg/tlpostcode/xetex.pl
```
jmairboeck, line 200: This sed change doesn't apply any more, because they changed the involved variable names, see https://tug.org/svn/texlive?revision=70088&view=revision

It doesn't seem to be needed any more though (the package builds successfully), presumably because of exactly that change.

note: I did not actually test XeTeX.

Begasus:
Perhaps better to split this up into 2 seperate PR's, there is no way to garentee that texlive_core will be build first (as mentioned yesterday at IRC).

PR #10264, opened 2024-03-28
texstudio, bump version

Begasus:


jmairboeck (approved):
This will probably also work fine with the old version of texlive, so it can also be merged before texlive is fixed, if you want.

PR #10276, opened 2024-03-30
mame: Move patchset to correct location for 0.264; bump revision

PocketNerdIO:
Moved `mame-0.264.patchset` from `games-emulation/mame/` to `games-emulation/mame/patches/`.


Begasus:
One of the reasons I stay on R1B4 (beside it's stable) is that  I know what builds here also builds on buildmaster, as the opposite isn't always true when I would run a nightly system and push to buildmasters (those still using R1B4).

PR #10278, opened 2024-03-31
squashfs_tools: add mime-type, and Expander rules.

OscarL:
Allows to just double click on ".sqfs" files to list or extract its content.

Icon is Haiku's "data/artwork/icons/Device_Volume". Could use a custom one.


sys-fs/squashfs_tools/squashfs_tools-4.6.1.recipe
```diff
@@ -73,8 +77,24 @@ BUILD()
...
+	rm x-squashfs.rsrc
+
+	# Create expander rules
+	mkdir -p $dataDir/expander/rules
+	echo "\"application/x-squashfs\"	.sqfs	\"unsquashfs -d '' -lls %s\"	\"unsquashfs -d . %s\"" \
+		> $dataDir/expander/rules/squashfs
+	echo "\"application/x-squashfs\"	.sfs	\"unsquashfs -d '' -lls %s\"	\"unsquashfs -d . %s\"" \
+		>> $dataDir/expander/rules/squashfs
+	echo "\"application/x-squashfs\"	.squashfs	\"unsquashfs -d '' -lls %s\"	\"unsquashfs -d . %s\"" \
+		>> $dataDir/expander/rules/squashfs
```
jmairboeck, line 99: This doesn't seem to contain anything that depends on a recipe variable. Why not add another file to additional-files and just copy it?

PR #10296, opened 2024-04-03
u_boot, bump version

Begasus:
previously one marked as vulnerable at repology


Begasus:
@OscarL as you previously enabled it, could you have a look here too? (not checked on 32bit yet)

Begasus:
32bit build ok too.

PR #10307, opened 2024-04-06
binwalk, bump version

Begasus:


Begasus:
2 sites providing, this is the one that is also used by Gentoo (both seem to have the same author also).
I never used/checked this so would be nice if someone that needs this could check it out.

1. https://github.com/ReFirmLabs/binwalk (providing 2.3.4)
2. https://github.com/OSPG/binwalk (this one, providing 2.4.0)

korli:
https://github.com/ReFirmLabs/binwalk didn't move since the fork, so it's safe to switch IMO.

PR #10308, opened 2024-04-07
ncurses6: update to version 6.4.

OscarL:
Also, removed older recipe/patch.


korli:
It seems 6.4 is still relevant for CVE-2023-29491 ...  
Arch uses https://invisible-mirror.net/archives/ncurses/current/ncurses-6.4-20230520.tgz along with --disable-root-access
    --disable-root-environ

korli:
Should be now https://invisible-mirror.net/archives/ncurses/6.4/ncurses-6.4-20230520.tgz

jmairboeck:
If you want to use something a bit more official: the author also provides the snapshots on GitHub, which as them all tagged: https://github.com/ThomasDickey/ncurses-snapshots/tags. The one in question is available here: https://github.com/ThomasDickey/ncurses-snapshots/archive/refs/tags/v6_4_20230520.tar.gz

korli:
> @korli any reason for going for that particular date "20230520", over just using the more recent "20240330"?

Who else uses these newer snapshots?

korli:
Then 20230520 should be OK?

korli:
> Still not sure why the need to follow Arch here instead of other distros, but... fine... I'll use that.

If Arch didn't see the need to update, then it must work good enough. With rolling distros it can be adventurous to be in the firsts to updates. IMO being cutting-edge needs more workforce, which we don't have.

korli:
Looks good. Just wondering about the version naming 6.4.20230520. Maybe 6.4_20230520 would be better.

sys-libs/ncurses/ncurses6-6.4_20230520.recipe
```diff
@@ -3,28 +3,31 @@ DESCRIPTION="The Ncurses (new curses) library is a free software emulation of \
...
+	1998-2017,2018 Free Software Foundation, Inc."
 LICENSE="MIT"
-REVISION="2"
-SOURCE_URI="https://ftpmirror.gnu.org/ncurses/ncurses-$portVersion.tar.gz
-	https://ftp.gnu.org/gnu/ncurses/ncurses-$portVersion.tar.gz"
-CHECKSUM_SHA256="97fc51ac2b085d4cde31ef4d2c3122c21abc217e9090a43a30fc5ec21684e059"
-SOURCE_DIR="ncurses-$portVersion"
-PATCHES="ncurses-$portVersion.patchset"
+REVISION="1"
+srcVersion=${portVersion//./_}
```
korli, line 21: is it still needed?

korli: OK, I wasn't sure.

korli:
LGTM

PR #10310, opened 2024-04-08
Gottet: Update to 1.2.3 and Qt 6

win8linux:
This PR updates Gottet to version `1.2.3` and to use Qt 6.

Confirmed working on `x86_64`:
screenshot1
screenshot2


games-puzzle/gottet/gottet-1.2.3.recipe
```diff
@@ -21,37 +21,49 @@ PROVIDES="
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
+	qt6_tools$secondaryArchSuffix
```
augiedoggie, line 24: qt6_tools aren't usually required at run time, just build time. Specifying the qt6_tools_devel package in BUILD_REQUIRES like you've done below should be enough since it will pull in the main qt6_tools package as well.
```suggestion
```

PR #10324, opened 2024-04-10
kgraphviewer, a graphviz dot graph file viewer

Begasus:


Begasus:
Using a default KDE icon as I can't find a svg one to convert into rdef. This is a srcGit checkout mainly because I wanted to try this out on KF6/Qt6, but works fine with KF5/Qt5 also, so kept the checkout.

PR #10327, opened 2024-04-12
less: bump version, remove old version

jmairboeck:
Fixes CVE-2022-46663

Build checked on x86_gcc2 and x86_64.

Tests hang (and don't build with gcc2, which could be patched but it doesn't make much sense to do so). Running the installed version seems to work fine.

Could someone (like @korli) please have a look at the patchset if everything is correct and makes sense?


jmairboeck:
> Considering this only provides `cmd:`s... couldn't we just switch it to `x86`?

Maybe we could do that, but this is one of Haiku's build packages, so I won't be able to decide that. The gcc2 fix for the new version was rather trivial (except for the tests, but they don't work on x86_64 either).

korli:
Looks fine to me (compared the patchsets).

jmairboeck:
Thanks!

PR #10331, opened 2024-04-13
nasm: Update to 2.15.02, add patch for path separators

robxnano:


Begasus:
Builds OK, test run compared to previous version is the same (64bit).

PR #10340, opened 2024-04-16
xmlsec, update SOURCE_URI

Begasus:


Begasus:
@threedeyes this is a dependency for libreoffice, while you are looking into that, could you bump this to the latest release 1.3.4? (build checked on 64bit and is fine)

PR #10345, opened 2024-04-16
kompare, bump version

Begasus:


Begasus:
libkomparediff2 as is can be build for KF6, but seeing KDevelop isn't there yet keep that one and this one on KF5.

PR #10357, opened 2024-04-19
clazy, bump version

Begasus:


Begasus:
Latest release (about 2 years old) with a few upstream patches and a fix created by the Arch people (that's where I found it) to build this with LLVM 17, shows up correctly in KDevelop.

PR #10367, opened 2024-04-22
abaddon: new recipe

TheZeldakatze:
I've finally did some work on the abaddon port, which is an
alternative discord client written in c++. Currently there is
no sound support, however I'd guess that it could be already of
use for people that want to use the text.

To get sound working, you'd need to patch miniaudio, which I don't know if I'm able to do that, but I'll try


TheZeldakatze:
I did not manage to port miniaudio, so I've turned off sound. As Discord also is used for chatting, I guess that it is still of use for some people maybe

Begasus:
Looks like you forgot the rdef file?

Begasus:
Crashes when quiting the app, I don't do discord, so I'm not able to check this to see if it's functional. :)

Begasus:
Looks like you included globulation2 changes in here?

TheZeldakatze:
> Looks like you included globulation2 changes in here?

I removed them again, I'll put them into another PR.
I was in a hurry to catch a train and just hastily pushed everything, forgetting that I did some  changes to globulation aswell.

> Crashes when quiting the app, I don't do discord, so I'm not able to check this to see if it's functional. :)

Interesting, it only happens when you have not yet logged in. I'll have a look at that

Begasus:
> Interesting, it only happens when you have not yet logged in. I'll have a look at that

I wasn't logged in, so that makes sense (don't have an account for it). If it's OK for you just give a notice :)

Begasus:
Merging for now, if you find anything in need of fixing please open a new PR, thanks!

PR #10368, opened 2024-04-23
gpgme, bump version

Begasus:


app-crypt/gpgme/gpgme-1.23.2.recipe
```diff
@@ -6,87 +6,77 @@ Currently it uses GnuPG as its backend but the API isn't restricted to \
...
-	cmd:tar
-	cmd:xargs
 	"
 
 BUILD()
 {
 	autoreconf -fi
 	runConfigure ./configure
 	make $jobArgs
+
```
Begasus, line 79: Not sure about the next phrase, it runs fine the first build but fails on a second build (figures as gpg-agent is already nuked) ...

app-crypt/gpgme/gpgme-1.23.2.recipe
```diff
@@ -6,87 +6,77 @@ Currently it uses GnuPG as its backend but the API isn't restricted to \
...
-	cmd:bison
-	cmd:gcc${secondaryArchSuffix}
-	cmd:git
+	cmd:autoreconf
+	cmd:gcc$secondaryArchSuffix
 	cmd:gpg$secondaryArchSuffix
 	cmd:gpgconf$secondaryArchSuffix
 	cmd:gpg_error$secondaryArchSuffix
-	cmd:grep
-	cmd:kill
```
jmairboeck, line 71: These are used below in the BUILD function, together with `ps` and `xargs` below. Shouldn't these be kept in therefore?

app-crypt/gpgme/gpgme-1.23.2.recipe
```diff
@@ -6,87 +6,77 @@ Currently it uses GnuPG as its backend but the API isn't restricted to \
...
 	"
 
 BUILD()
 {
 	autoreconf -fi
 	runConfigure ./configure
 	make $jobArgs
+
 	# kill gpg-agent processes
 	ps -o Id Team | grep 'gpg-agent' | grep -v "grep" | awk '{print $1}' | xargs kill -9 $1
```
jmairboeck, line 81: A little shell script nit-pick: the `$1` at the end is superfluous and can be removed. In this context it would expand to the first argument of the BUILD function (which has none), so this will always be empty anyway.

Maybe if this fails sometimes, it would help to suppress the error by adding `|| true` at the end? (untested).

app-crypt/gpgme/gpgme-1.23.2.recipe
Begasus: It works with or without them in the recipe, can put them back in.

Begasus: Added back in.

app-crypt/gpgme/gpgme-1.23.2.recipe
Begasus: Removed $1, no issues detected. :)

Begasus: I seems that on buildmaster this failed (probably no gpg-agent running there):
```
kill: no process ID specified
Try 'kill --help' for more information.
```

PR #10379, opened 2024-04-26
xapian, bump version

Begasus:
link *.so back into libDir , fixes build for khelpcenter(still not functiona)

### DIFF
--- a/dev-libs/xapian/xapian-1.4.25.recipe
+++ b/dev-libs/xapian/xapian-1.4.25.recipe
@@ -10,7 +10,7 @@ COPYRIGHT="1999-2018 The Xapian Authors"
 LICENSE="GNU GPL v2"
 REVISION="1"
 SOURCE_URI="https://oligarchy.co.uk/xapian/$portVersion/xapian-core-$portVersion.tar.xz"
-CHECKSUM_SHA256="05884af00b06702ce486057d62a3bfbe6606cf965ada0f5ea570b328a2fa1ea8"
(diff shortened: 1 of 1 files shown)


dev-libs/xapian/xapian-1.4.25.recipe
```diff
@@ -109,6 +111,8 @@ INSTALL()
 	prepareInstalledDevelLib libxapian
 	fixPkgconfig
 
+	ln -sf $libDir/libxapian.so.$libVersion $libDir/libxapian.so
```
Begasus, line 114: Maybe this could be done with a sed command, but haven't figured that out, libxapian.so is referenced in $libDir/cmake/xapian/xapian-config.cmake with:
```
SET(XAPIAN_SHARED_LIBRARY "/packages/xapian-1.4.25-2/.self/lib/libxapian.so" CACHE FILEPATH "Shared Xapian library")
```

jmairboeck: The cmake file should be patched to use `$developLibDir` instead. This is used to find the library to link to.

Begasus: That doesn't do the trick, as it will mix up $libDir and $developLibDir for finding libraries/headers (tried to my best knowlidge, hence I added the comment here). :) IIRC it mixed up header path versus library path when checking up on khelpcenter (ps, this wasn't/isn't an issue for the other ports using this).

PR #10397, opened 2024-05-01
kpimtextedit, revome unneeded version for ECM, no rebuild

Begasus:


Begasus:
Since ECM 6.0.0 this is backwords compatible, so version isn't needed anymore (only removing when (rev)bumping recipes (or when I forget it like in this case) :)

PR #10416, opened 2024-05-04
kio_extras, revbump for kdsoap_ws_discovery_client

Begasus:


Begasus:
Upstream doesn't change libName in kdsoap_ws_discovery_client for Qt5/Qt6, hence the revbump, rely on the Qt5 version for now.

PR #10436, opened 2024-05-11
trax: new recipe

### LINKED ISSUE #10407: Create a recipe for TraX
https://github.com/HaikuArchives/TraX

jmairboeck:
Fixes #10407

### DIFF
--- a/haiku-apps/trax/trax-1.1.1.recipe
+++ b/haiku-apps/trax/trax-1.1.1.recipe
@@ -1,0 +1,59 @@
+SUMMARY="Find files by name AND location, unlike Haiku's Find"
+DESCRIPTION="TraX is not a \"query\" tool. It does not keep a database of files and attributes. \
+It is a traditional find tool, which searches the disk every time, looking for files that meet \
+your criteria. The find-criteria are not Haiku specific, and it does not (currently) search \
+attributes.
+
+TraX comes in two varieties.
+
+ * FastTraX is similar to the Windows find tool. Windows users should find its features very \
+familiar.
+ * TraX is a cross between the UNIX \"find\" command the the MacOS \"Find Files\" tool. It has \
+much of the power of \"find\" with a Mac-like interface (a Mac on steroids that is)."
+HOMEPAGE="https://github.com/HaikuArchives/TraX/"
+COPYRIGHT="1999-2000, Tim Vernum"
+LICENSE="GNU GPL v2"
+REVISION="1"
+srcGitRev="ed09bd45e6cd9faa365c8c1f542f1636aa19cd63"
+SOURCE_URI="https://github.com/HaikuArchives/TraX/archive/$srcGitRev.tar.gz"
+CHECKSUM_SHA256="8a4261a787fa861ee40b78448e73e00e50800d2b660945c813a14ca5df31c8b4"
+SOURCE_DIR="TraX-$srcGitRev"
+
+ARCHITECTURES="all"
+
+PROVIDES="
+	trax = $portVersion
+	app:FastTraX = $portVersion
+	app:TraX = $portVersion
+	lib:libfind
+	lib:libTraX
+	"
+REQUIRES="
+	haiku
+	"
+
+BUILD_REQUIRES="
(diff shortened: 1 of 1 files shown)


haiku-apps/trax/trax-1.1.1.recipe
```diff
@@ -0,0 +1,59 @@
...
+
+INSTALL()
+{
+	# make install doesn't really do anything, it is done by the default target already
+	# but keep it here if this is fixed later.
+	make install
+
+	mkdir -p $appsDir $libDir
+	cp -a bin/FastTraX bin/TraX $appsDir
+	cp -a bin/lib/*.so $libDir
```
jmairboeck, line 56: Is installing the libraries to `$libDir` correct? Upstream installs them in a `lib` subdirectory, but I think having them in `$libDir` is nicer than in a subdirectory of `$appsDir`. Or should everything be put in a subdirectory of `$appsDir`, given there are two versions of the app?

haiku-apps/trax/trax-1.1.1.recipe
```diff
@@ -0,0 +1,59 @@
...
+SOURCE_DIR="TraX-$srcGitRev"
+
+ARCHITECTURES="all"
+
+PROVIDES="
+	trax = $portVersion
+	app:FastTraX = $portVersion
+	app:TraX = $portVersion
+	lib:libfind = $portVersion
+	lib:libTraX = $portVersion
```
jmairboeck, line 29: Should the libraries be versioned at all? They don't really have a version.

If they aren't installed in `$libDir`, they also don't need to be listed here.

haiku-apps/trax/trax-1.1.1.recipe
augiedoggie: > Upstream installs them in a `lib` subdirectory

A lot of old BeOS apps did that so they could be distributed as a zip file.  You could just extract and run the app right away because `%A/lib` is in `$LIBRARY_PATH`.  This isn't necessary when installing from a package, so I always move them to `$libDir`.

haiku-apps/trax/trax-1.1.1.recipe
augiedoggie: If you version them then you should start the `$libVersion` at 0.0.0.  There are a number of recipes that do this.

jmairboeck: The HaikuArchives repo contains a resource file for libfind.so with the version number 0.0.0, but it isn't actually used. So effectively, both these libraries have no version at all.

jmairboeck: I removed the version numbers.

pulkomandy:
> Is installing the libraries to $libDir correct? Upstream installs them in a lib subdirectory, but I think having them in $libDir is nicer than in a subdirectory of $appsDir. Or should everything be put in a subdirectory of $appsDir, given there are two versions of the app?

Are these libraries used by multiple binaries in TraX? If so, it may make sense that this is shared code just for these binaries and not for anyone else.

Otherwise, if it's private code for just one app it makes no sense to have it as a library, include it directly in the binary. And if it's meant to be used more generally, put it in $libDir and $developLibDir, and provide a _devel package, so everyone can use it in their own apps.

jmairboeck:
> Are these libraries used by multiple binaries in TraX?

They are both used by both versions of the app, according to haikuporter's policy checks.

> And if it's meant to be used more generally, put it in $libDir and $developLibDir, and provide a _devel package, so everyone can use it in their own apps.

For libfind.so, that may make sense indeed, that has a kind-of "public" header in the repository (but it would need to be installed manually in the recipe). libTraX.so seems to be just for sharing code between the two app binaries.

jmairboeck:
I created a [Pull request](https://github.com/HaikuArchives/TraX/pull/8) to fix the existing standalone makefile for libfind.so. Then it could be packaged separately as e.g. "trax_findlib" (also with a secondaryArch package) and TraX made dependendant on that without complicating this recipe too much.

PR #10440, opened 2024-05-14
cudatext, revbump, fix 32bit binary

Begasus:


Begasus:
Previous tag for 32bit binary wasn't executable. :)

PR #10455, opened 2024-05-20
fs_uae, fs_uae_launcher: Update FS-UAE to 3.1.66, patch Launcher for better exe finding

johnbintz:
* The current FS-UAE Launcher did not find the new 3.1.66 `fs-uae` binary correctly so I had to modify the binary search source to allow a Haiku-specific search, which isn't unexpected in that part of the code.
* Compiling FS-UAE with `--with-glew` was complaining about a missing `gl.pc` file. From what I've read on other porting discussions this is intentional? To get compilation to work I had to provide all of the necessary flags as env variables. It does seem to run without `--with-glew` but I didn't want to change the original recipe too much if I could help it.
* FS-UAE seemed to need an `autoreconf` before I could build the new source.

Screenshot from 2024-05-20 06-43-55


Begasus:
Even without --with-glew I'm getting build error for fs-uae on 64bit (R1B4)
```
In file included from src/dosbox/dosbox.h:92,
                 from src/x86.cpp:43:
./src/dosbox/setup.h:88:35: error: ISO C++17 does not allow dynamic exception specifications
   88 |         Value& operator= (Hex in) throw(WrongType)                { return copy(Value(in));}
      |                                   ^~~~~
./src/dosbox/setup.h:89:35: error: ISO C++17 does not allow dynamic exception specifications
   89 |         Value& operator= (int in) throw(WrongType)                { return copy(Value(in));}
      |                                   ^~~~~
./src/dosbox/setup.h:90:36: error: ISO C++17 does not allow dynamic exception specifications
   90 |         Value& operator= (bool in) throw(WrongType)               { return copy(Value(in));}
      |                                    ^~~~~
...
```

Cosmetic changes needed also as there are spaces used instead of tabs in the front of the lines of both recipes.

johnbintz:
I'll see if I can reproduce the build error locally from a clean install.

Begasus:
My bad, build correctly on 64bit, thanks, now I need something to check this with ;)

johnbintz:
This doesn't require an Amiga Kickstart ROM to start up, you can use the built-in open source ROM. It is not very exciting but it works. :) https://retro-commodore.eu/files/downloads/Amiga/Games/Public%20Domain/C/Crossword%20Creator,%20The%20v1.1%20(1990)(Polyglot%20Software)%5bconstruction%20kit%5d.zip

Unzip it and mount the resulting ADF file like this in FS-UAE Launcher, then click Start:

Screenshot from 2024-05-20 08-53-12

Begasus:
Thanks! Indeed not very exiting (but then again, not that familiar with this, and it booted the file). :)

PR #10458, opened 2024-05-22
latex2html, bump version

Begasus:


Begasus:
@jmairboeck I checked in 64bit (where I "do" have texlive-2024 installed), not sure it is a hard dependency for the texlive version?

jmairboeck:
I think it will probably work with any texlive, according to https://github.com/latex2html/latex2html/blob/master/README.md#requirements. I have never used this myself, so I don't know for sure.

Begasus:
IIRC I tested it back then with BeTEX? ;)

PR #10466, opened 2024-05-24
Add recipe for haiku-format 18.1.6

owenca:


pulkomandy:
Can we delete the previous versions of the recipe?

owenca:
> Can we delete the previous versions of the recipe?

LLVM 18.1.6 was the planned final release, but it seems that there will be another point release due to llvm/llvm-project#92991. I'm not sure if I need to update haiku-format to 18.1.7, so we probably want to keep the 17.0.6 recipe for now.

PR #10467, opened 2024-05-24
sqlite: quick fix to re-enable libedit

augiedoggie:


augiedoggie:
Needs a proper patch but dealing with autoconf gives me a headache.

PR #10480, opened 2024-05-27
php,revbump, use ICU74

Begasus:


Begasus:
Same as the others, could use some extra pair of eyes :)

PR #10486, opened 2024-05-28
Add KF6 frameworks, karchive6

Begasus:


kde-frameworks/karchive/karchive6-6.2.0.recipe
```diff
@@ -0,0 +1,88 @@
...
+INSTALL()
+{
+	make -Cbuild install
+
+	prepareInstalledDevelLib \
+		libKF6Archive
+
+	packageEntries devel \
+		$developDir \
+		$dataDir/doc \
```
jmairboeck, line 80: Shouldn't these files go somewhere in `$documentationDir` (or maybe `$developDocDir`). `$dataDir/doc` looks like a "half-Linuxism" for me (where `share/doc` is the equivalent of `$documentationDir`).

I suspect that the CMake doesn't use `${CMAKE_INSTALL_DOCDIR}` but something like `${CMAKE_INSTALL_DATADIR}/doc`. Could this be patched?

Begasus: I left it as is yes, could be patched, but then we would have to do this for all frameworks that use this which (imho) would lead to more noise at the moment?
I'm not moving them "at this moment" while pushing them, can have a look later though. :)

Begasus: As it looks like it's using datarootdir for the paths (qtqchdir), fidling a bit with it with 6.3.0 to sort this out.
Using: `-DKDE_INSTALL_QTQCHDIR=$documentationDir/packages/KF6` at the moment as it concerns the KF6 frameworks, Qt6 Assistant doesn't pick it up (even if I use the qt6-base directory in there), so you have to manually add it in there to use it, at least I think it would be easier to have them all in one place then in: -DKDE_INSTALL_QTQCHDIR=$documentationDir/packages/ki18n (and the likes)?

jmairboeck: I'm no expert in this and haven't tested anything myself. So I trust your judgement here fully :)

Begasus: Thanks :) Just hope no one steps up after x version further. :)

Begasus: From Arch: usr/share/doc/qt6/KF6I18n.qch ...

PR #10537, opened 2024-05-29
Add KF6 frameworks, kimageformats6

Begasus:


Begasus:
For jxl it could use an update to add the plugin here. For now leave disabled.

PR #10564, opened 2024-06-02
Enable Openssl3, bump recipes that depend on it and are used by Haiku

pulkomandy:


pulkomandy:
This may need some coordination with Haiku. https://review.haiku-os.org/c/haiku/+/7718

dev-lang/python/python3.9-3.9.18.recipe
```diff
@@ -53,7 +53,7 @@ REQUIRES="
 	lib:liblzma$secondaryArchSuffix
 	lib:libncurses$secondaryArchSuffix
 	lib:libsqlite3$secondaryArchSuffix
-	lib:libssl$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix >= 3
```
korli, line 56: this is normally not needed.

korli: same for other recipes.

dev-vcs/git/git-2.45.1.recipe
```diff
@@ -92,7 +92,7 @@ REQUIRES="
 	lib:libintl$secondaryArchSuffix
 	lib:libpcre2_8$secondaryArchSuffix
 	lib:libnghttp2$secondaryArchSuffix
-	lib:libssl$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffi >= 3
```
jmairboeck, line 95: Typo:

```suggestion
	lib:libssl$secondaryArchSuffix >= 3
```

pulkomandy:
It's needed if we want to merge this all in one go. Otherwise we have to do it in two steps, first enable the new openssl, and then bump the revision of everything that depends on it. I'm also not sure it will work if the older version of OpenSSL is always pulled in by the Haiku package (then there's no reason the dependency resolution would also include the new version).

And it's also useful to track what software was already rebuilt with the new OpenSSL version.

So, I think it's not a bad idea to have this done explicitly.

korli:
> It's needed if we want to merge this all in one go. Otherwise we have to do it in two steps, first enable the new openssl, and then bump the revision of everything that depends on it. I'm also not sure it will work if the older version of OpenSSL is always pulled in by the Haiku package (then there's no reason the dependency resolution would also include the new version). And it's also useful to track what software was already rebuilt with the new OpenSSL version. So, I think it's not a bad idea to have this done explicitly.

No, that's only needed on the devel:libssl.

dev-vcs/git/git-2.45.1.recipe
pulkomandy: Done.

dev-lang/python/python3.9-3.9.18.recipe
pulkomandy: Ok, fixed by removing the constraint on lib: and keeping it only on devel:

dev-libs/openssl/openssl3-3.0.13.recipe
```diff
@@ -21,7 +21,7 @@ CHECKSUM_SHA256="88525753f79d3bec27d2fa7c66aa0b92b3aa9498dafd93d7cfa4b3780cdae31
 SOURCE_DIR="openssl-$portVersion"
 PATCHES="openssl3-$portVersion.patchset"
 
-ARCHITECTURES="?all"
+ARCHITECTURES="all"
 SECONDARY_ARCHITECTURES="?x86"
```
korli, line 25: Why not enabling for x86?

pulkomandy: I enabled only what I could test. I did not boot my 32bit system this weekend. I can make sure everything works there as well if no one else does it before me.

korli:
Just to understand: if this gets built on master, it should also be tested on r1beta4, right?

pulkomandy:
For now I only tested building a nightly build with these updates (and I did not succeed yet, see the comments on the corresponding Haiku changes).

I think the steps are:

- Check Haiku build with OpenSSL 3 builds and works (fix the problem I hit with file_data there, try booting the image and see if everything still works: WebPositive, pkgman, etc)
- Check that the packages can also be installed on beta4 (I think openssl3 and openssl 1.1 can be installed side by side so this should be fine). In this case Haiku code will still depend on and use OpenSSL 1.1.
- Once merged, migrate all other things that use openssl 1.1 to openssl3, to prepare for the beta5 release
- After beta5 is released and all things migrated: remove openssl 1.1 package

This should also be checked on all architectures.

Begasus:
> * Check that the packages can also be installed on beta4 (I think openssl3 and openssl 1.1 can be installed side by side so this should be fine). In this case Haiku code will still depend on and use OpenSSL 1.1.

Quick check on building the package so far on 64bit R1B4, looking at package content I see that $dataDir/ssl contains the same files for openssl1 and openssl3?

korli:
Also keep in mind, that loading libssl.so.1 and libssl.so.3 in the same application might cause problems.

Begasus:
Haiku image build on 64bit boots OK in qemu (WebPositive + pkgman working fine)

Begasus:
Getting a warning when launching a non related recipe now:
```
runtime_loader: Cannot open file libproviders.so (needed by /boot/system/lib/libcrypto.so.1.1): No such file or directory
```

pulkomandy:
It looks like this is caused by reusing the same config directory for openssl 1.1 and openssl 3. So we need to configure it to use a different directory so they can live side by side peacefully: ***@***.***/msg91213.html

dev-libs/openssl/openssl3-3.0.13.recipe
```diff
@@ -106,7 +106,7 @@ defineDebugInfoPackage openssl3$secondaryArchSuffix \
 BUILD()
 {
 	./config --prefix=$prefix --libdir=$relativeLibDir \
-		--openssldir=$dataRootDir/ssl \
+		--openssldir=$dataRootDir/ssl3 \
```
korli, line 109: probably a bad idea, because this is where the root certificates are located. Does it work nonetheless?

Begasus: Seems to fixed the earlier warning on libproviders.so:
```
/Opslag/haikuports> pkgman install qt_creator
100% repochecksum-1 [65 bytes]
Validating checksum for Haiku...done.
100% repochecksum-1 [64 bytes]
Validating checksum for HaikuPorts...done.
  in system:
    install package qt_creator-13.0.0-1 from repository HaikuPorts
Continue? [yes/no] (yes) : 
Re-using download '/boot/system/packages/administrative/state_2024-05-02_14:24:55/qt_creator-13.0.0-1-x86_64.hpkg' from previous transaction
Validating checksum for https://eu.hpkg.haiku-os.org/haikuports/r1beta4/x86_64/current/packages/qt_creator-13.0.0-1-x86_64.hpkg...done.
[system] Applying changes ...
```
just checked on 64bit R1B4 on a clean build/install for openssl3

Begasus:
Uploaded some prebuild packages (for now only 64bit), build on R1B4 for those not wanting to build each package related here (to test out), at https://github.com/Begasus/Haiku_Testdrive/tree/master/openssl/64bit

dev-libs/openssl/openssl3-3.0.13.recipe
korli: I mean: does openssl3 work?

Begasus: If I knew how to check this I could answer that :) Just did a build for a 64bit anyboot iso with these installed, as far as it should use ssl_devel it can only pick up openssl3 as openssl_devel conflicts/isn't installed here atm.
Build went fine and image boots fine in qemu.
At this moment all I could do for now is build/install the packages, see if nothing is broken.
If there is a quick way to check it I'm all ears. :)

Begasus: OK, not sure if it's correct, but getting this output in Terminal:
```
~/Desktop> openssl_3 s_client -connect example.com:443
CONNECTED(00000004)
depth=1 C = US, O = DigiCert Inc, CN = DigiCert Global G2 TLS RSA SHA256 2020 CA1
verify error:num=20:unable to get local issuer certificate
verify return:1
depth=0 C = US, ST = California, L = Los Angeles, O = Internet\C2\A0Corporation\C2\A0for\C2\A0Assigned\C2\A0Names\C2\A0and\C2\A0Numbers, CN = www.example.org
verify return:1
---
Certificate chain
 0 s:C = US, ST = California, L = Los Angeles, O = Internet\C2\A0Corporation\C2\A0for\C2\A0Assigned\C2\A0Names\C2\A0and\C2\A0Numbers, CN = www.example.org
   i:C = US, O = DigiCert Inc, CN = DigiCert Global G2 TLS RSA SHA256 2020 CA1
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Jan 30 00:00:00 2024 GMT; NotAfter: Mar  1 23:59:59 2025 GMT
 1 s:C = US, O = DigiCert Inc, CN = DigiCert Global G2 TLS RSA SHA256 2020 CA1
   i:C = US, O = DigiCert Inc, OU = www.digicert.com, CN = DigiCert Global Root G2
   a:PKEY: rsaEncryption, 2048 (bit); sigalg: RSA-SHA256
   v:NotBefore: Mar 30 00:00:00 2021 GMT; NotAfter: Mar 29 23:59:59 2031 GMT
---
Server certificate
-----BEGIN CERTIFICATE-----
MIIHbjCCBlagAwIBAgIQB1vO8waJyK3fE+Ua9K/hhzANBgkqhkiG9w0BAQsFADBZ
MQswCQYDVQQGEwJVUzEVMBMGA1UEChMMRGlnaUNlcnQgSW5jMTMwMQYDVQQDEypE
aWdpQ2VydCBHbG9iYWwgRzIgVExTIFJTQSBTSEEyNTYgMjAyMCBDQTEwHhcNMjQw
MTMwMDAwMDAwWhcNMjUwMzAxMjM1OTU5WjCBljELMAkGA1UEBhMCVVMxEzARBgNV
BAgTCkNhbGlmb3JuaWExFDASBgNVBAcTC0xvcyBBbmdlbGVzMUIwQAYDVQQKDDlJ
bnRlcm5ldMKgQ29ycG9yYXRpb27CoGZvcsKgQXNzaWduZWTCoE5hbWVzwqBhbmTC
oE51bWJlcnMxGDAWBgNVBAMTD3d3dy5leGFtcGxlLm9yZzCCASIwDQYJKoZIhvcN
AQEBBQADggEPADCCAQoCggEBAIaFD7sO+cpf2fXgCjIsM9mqDgcpqC8IrXi9wga/
9y0rpqcnPVOmTMNLsid3INbBVEm4CNr5cKlh9rJJnWlX2vttJDRyLkfwBD+dsVvi
vGYxWTLmqX6/1LDUZPVrynv/cltemtg/1Aay88jcj2ZaRoRmqBgVeacIzgU8+zmJ
7236TnFSe7fkoKSclsBhPaQKcE3Djs1uszJs8sdECQTdoFX9I6UgeLKFXtg7rRf/
hcW5dI0zubhXbrW8aWXbCzySVZn0c7RkJMpnTCiZzNxnPXnHFpwr5quqqjVyN/aB
KkjoP04Zmr+eRqoyk/+lslq0sS8eaYSSHbC5ja/yMWyVhvMCAwEAAaOCA/IwggPu
MB8GA1UdIwQYMBaAFHSFgMBmx9833s+9KTeqAx2+7c0XMB0GA1UdDgQWBBRM/tAS
TS4hz2v68vK4TEkCHTGRijCBgQYDVR0RBHoweIIPd3d3LmV4YW1wbGUub3Jnggtl
eGFtcGxlLm5ldIILZXhhbXBsZS5lZHWCC2V4YW1wbGUuY29tggtleGFtcGxlLm9y
Z4IPd3d3LmV4YW1wbGUuY29tgg93d3cuZXhhbXBsZS5lZHWCD3d3dy5leGFtcGxl
Lm5ldDA+BgNVHSAENzA1MDMGBmeBDAECAjApMCcGCCsGAQUFBwIBFhtodHRwOi8v
d3d3LmRpZ2ljZXJ0LmNvbS9DUFMwDgYDVR0PAQH/BAQDAgWgMB0GA1UdJQQWMBQG
CCsGAQUFBwMBBggrBgEFBQcDAjCBnwYDVR0fBIGXMIGUMEigRqBEhkJodHRwOi8v
Y3JsMy5kaWdpY2VydC5jb20vRGlnaUNlcnRHbG9iYWxHMlRMU1JTQVNIQTI1NjIw
MjBDQTEtMS5jcmwwSKBGoESGQmh0dHA6Ly9jcmw0LmRpZ2ljZXJ0LmNvbS9EaWdp
Q2VydEdsb2JhbEcyVExTUlNBU0hBMjU2MjAyMENBMS0xLmNybDCBhwYIKwYBBQUH
AQEEezB5MCQGCCsGAQUFBzABhhhodHRwOi8vb2NzcC5kaWdpY2VydC5jb20wUQYI
KwYBBQUHMAKGRWh0dHA6Ly9jYWNlcnRzLmRpZ2ljZXJ0LmNvbS9EaWdpQ2VydEds
b2JhbEcyVExTUlNBU0hBMjU2MjAyMENBMS0xLmNydDAMBgNVHRMBAf8EAjAAMIIB
fQYKKwYBBAHWeQIEAgSCAW0EggFpAWcAdABOdaMnXJoQwzhbbNTfP1LrHfDgjhuN
acCx+mSxYpo53wAAAY1b0vxkAAAEAwBFMEMCH0BRCgxPbBBVxhcWZ26a8JCe83P1
JZ6wmv56GsVcyMACIDgpMbEo5HJITTRPnoyT4mG8cLrWjEvhchUdEcWUuk1TAHYA
fVkeEuF4KnscYWd8Xv340IdcFKBOlZ65Ay/ZDowuebgAAAGNW9L8MAAABAMARzBF
AiBdv5Z3pZFbfgoM3tGpCTM3ZxBMQsxBRSdTS6d8d2NAcwIhALLoCT9mTMN9OyFz
IBV5MkXVLyuTf2OAzAOa7d8x2H6XAHcA5tIxY0B3jMEQQQbXcbnOwdJA9paEhvu6
hzId/R43jlAAAAGNW9L8XwAABAMASDBGAiEA4Koh/VizdQU1tjZ2E2VGgWSXXkwn
QmiYhmAeKcVLHeACIQD7JIGFsdGol7kss2pe4lYrCgPVc+iGZkuqnj26hqhr0TAN
BgkqhkiG9w0BAQsFAAOCAQEABOFuAj4N4yNG9OOWNQWTNSICC4Rd4nOG1HRP/Bsn
rz7KrcPORtb6D+Jx+Q0amhO31QhIvVBYs14gY4Ypyj7MzHgm4VmPXcqLvEkxb2G9
Qv9hYuEiNSQmm1fr5QAN/0AzbEbCM3cImLJ69kP5bUjfv/76KB57is8tYf9sh5ik
LGKauxCM/zRIcGa3bXLDafk5S2g5Vr2hs230d/NGW1wZrE+zdGuMxfGJzJP+DAFv
iBfcQnFg4+1zMEKcqS87oniOyG+60RMM0MdejBD7AS43m9us96Gsun/4kufLQUTI
FfnzxLutUV++3seshgefQOy5C/ayi8y1VTNmujPCxPCi6Q==
-----END CERTIFICATE-----
subject=C = US, ST = California, L = Los Angeles, O = Internet\C2\A0Corporation\C2\A0for\C2\A0Assigned\C2\A0Names\C2\A0and\C2\A0Numbers, CN = www.example.org
issuer=C = US, O = DigiCert Inc, CN = DigiCert Global G2 TLS RSA SHA256 2020 CA1
---
No client certificate CA names sent
Peer signing digest: SHA256
Peer signature type: RSA-PSS
Server Temp Key: ECDH, prime256v1, 256 bits
---
SSL handshake has read 3821 bytes and written 747 bytes
Verification error: unable to get local issuer certificate
---
New, TLSv1.3, Cipher is TLS_AES_256_GCM_SHA384
Server public key is 2048 bit
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
No ALPN negotiated
Early data was not sent
Verify return code: 20 (unable to get local issuer certificate)
---
---
Post-Handshake New Session Ticket arrived:
SSL-Session:
    Protocol  : TLSv1.3
    Cipher    : TLS_AES_256_GCM_SHA384
    Session-ID: F4F1D56AF391AA4DB99E52DBCC7566695A9B4C682B6CE8C4734B8694E00E01A0
    Session-ID-ctx: 
    Resumption PSK: 5FCEFDC3C04F67A4751637A3EF4A68657E256720E6F3D0812EC822CE367A311663FF117551833C610D855FA66EC21A4A
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    TLS session ticket lifetime hint: 7200 (seconds)
    TLS session ticket:
    0000 - 6e e8 7c 10 e1 ec 9d fc-ba e9 28 42 c0 ad c3 5a   n.|.......(B...Z
    0010 - 9a a9 62 72 69 cb 58 6f-3b cd b5 11 ea 84 bf b7   ..bri.Xo;.......
    0020 - eb 7b b6 11 23 bf 4e c8-4d 9e 78 b7 05 97 34 df   .{..#.N.M.x...4.
    0030 - ed 87 1a 32 f8 4c 83 90-79 b9 9e 2d 11 7a e0 05   ...2.L..y..-.z..
    0040 - 93 55 3f 85 b3 6e 8a c7-79 cc 8d 60 7b 86 2f 1d   .U?..n..y..`{./.
    0050 - 0e 72 8e 1e 9a 5c 1f e2-70 e1 58 74 d0 ce 56 46   .r...\..p.Xt..VF
    0060 - fc cf 31 ae 09 2c 28 4c-96 66 e2 ca 58 50 2d 42   ..1..,(L.f..XP-B
    0070 - 35 87 76 f3 11 6c ef be-52 8e 00 2f c2 16 4b 08   5.v..l..R../..K.
    0080 - 2d 0d ee bb 8e 2c 4a 8a-bc d1 6b b3 08 91 07 99   -....,J...k.....
    0090 - e2 09 67 88 cb cf 25 6f-c6 c3 6a eb c4 45 7d f6   ..g...%o..j..E}.
    00a0 - 4b 30 90 80 cf 07 44 03-56 94 7a 27 2f ec c1 4f   K0....D.V.z'/..O
    00b0 - eb c3 28 8e c9 b5 7b 09-89 c0 bc 9f 8c 8b 34 26   ..(...{.......4&

    Start Time: 1717503349
    Timeout   : 7200 (sec)
    Verify return code: 20 (unable to get local issuer certificate)
    Extended master secret: no
    Max Early Data: 0
---
read R BLOCK
---
Post-Handshake New Session Ticket arrived:
SSL-Session:
    Protocol  : TLSv1.3
    Cipher    : TLS_AES_256_GCM_SHA384
    Session-ID: 64BA67430DBE8B5D716C57E0828CA708A4CAD441313ED370D0821FC5CE2603AD
    Session-ID-ctx: 
    Resumption PSK: 6A568BBCBB3AC5677209E6FB0E475E28E8B8CD5E8B3B7B9F81EF98DA648E6778BB8D8950604B37EE801A37BA1502B0C7
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    TLS session ticket lifetime hint: 7200 (seconds)
    TLS session ticket:
    0000 - 6e e8 7c 10 e1 ec 9d fc-ba e9 28 42 c0 ad c3 5a   n.|.......(B...Z
    0010 - cf 81 e4 49 8f 51 3c 1c-b3 a7 09 11 9a ac ad d4   ...I.Q<.........
    0020 - 2a 47 06 44 9b f3 7f e6-be ee d4 39 fa 2c 83 7f   *G.D.......9.,..
    0030 - ba 38 63 f2 16 28 69 9e-a5 2e 5e 86 79 86 5f 99   .8c..(i...^.y._.
    0040 - 6b 7e 38 43 bc 34 d0 46-c8 4a 60 99 81 0c 38 53   k~8C.4.F.J`...8S
    0050 - 3c fb 03 fd 82 0e 34 a9-d0 fb 9d 7a 9b f7 41 89   <.....4....z..A.
    0060 - 44 01 3c e0 c2 b4 f7 a2-d4 97 93 80 2d 4a e2 48   D.<.........-J.H
    0070 - 96 61 98 35 ae a5 e2 0d-13 6b d9 61 9d a2 62 70   .a.5.....k.a..bp
    0080 - 88 0a 4a 25 8a 29 cd 0e-fb a8 79 47 47 50 63 d7   ..J%.)....yGGPc.
    0090 - 5f cc 54 bb 62 ea 8b 44-c1 5e 5c 6e 76 0e b7 c1   _.T.b..D.^\nv...
    00a0 - 44 ec 73 29 32 e5 6a e3-f6 1e 25 75 41 c4 33 0e   D.s)2.j...%uA.3.
    00b0 - d4 75 dc 31 71 f7 f4 de-81 aa cd 53 8e 9c 33 5b   .u.1q......S..3[

    Start Time: 1717503349
    Timeout   : 7200 (sec)
    Verify return code: 20 (unable to get local issuer certificate)
    Extended master secret: no
    Max Early Data: 0
---
read R BLOCK
bad select -2147483638
~/Desktop> 
```

pulkomandy: Ah, I see my previous reply got mangled because there was an @ in an URL, great. Here is a link in a different archive of the same mailing list... https://www.spinics.net/lists/openssl-users/msg15473.html

I changed this to fix the mentionned issue about libproviders.so.

> verify error:num=20:unable to get local issuer certificate

This means it didn't find the certificates. So, yes, that will be a problem as more things start using OpenSSL 3. Good catch :)

Maybe we can add a link to the certificates in the other directory, or change something in the configuration file, so that the certificates are still found in the previous location. But we need a different directory for other configuration files because they are not compatible (unless we give up on having both versions co-exist).

korli: "unable to get local issuer certificate" doesn't sound good

korli:
I see that in Arch the data directory isn't present for openssl 1.1, only for 3.2. Maybe openssl can depend on openssl3

dev-libs/openssl/openssl3-3.0.13.recipe
Begasus: Not sure if it's the right way (thinking out loud), https://sources.debian.org/patches/openssl/3.3.0-1/Remove-the-provider-section.patch/
With this and --openssldir=$dataRootDir/ssl I now get:
```
Post-Handshake New Session Ticket arrived:
SSL-Session:
    Protocol  : TLSv1.3
    Cipher    : TLS_AES_256_GCM_SHA384
    Session-ID: C21940C1EACFD36488CE472AB1A068B502B05AF4799C76F6D83A80B62EA958C6
    Session-ID-ctx: 
    Resumption PSK: 8A7F9CEBCD68C55770E3131D4C597FD344D30ABAC7BABD081C8997CBCB2730D3A56CCBCCE80C8804C67A4EB635314517
    PSK identity: None
    PSK identity hint: None
    SRP username: None
    TLS session ticket lifetime hint: 7200 (seconds)
    TLS session ticket:
    0000 - 47 e9 ed c1 e0 7c 51 24-3d ca 45 ba 10 75 b2 d8   G....|Q$=.E..u..
    0010 - 39 92 80 2d 03 e4 5a 8c-24 f5 99 8f 76 42 86 2a   9..-..Z.$...vB.*
    0020 - 97 c4 66 e8 bd f5 40 62-51 3b f3 e1 ce 80 42 e9   ..f...@bQ;....B.
    0030 - 32 17 08 b0 81 7a 9f ec-73 cc 8e ff c5 ec d8 fb   2....z..s.......
    0040 - 45 3e 61 3f 16 c9 ff c9-1b 33 4d 1b ab 19 50 a6   E>a?.....3M...P.
    0050 - d8 38 e9 6a 06 bc 39 51-c4 17 7d ce e9 1f 44 e0   .8.j..9Q..}...D.
    0060 - 81 19 79 d6 f6 19 ef 35-02 9f 65 7c 04 4b 00 68   ..y....5..e|.K.h
    0070 - 18 4f 67 fa 93 ed 15 5d-55 62 c5 f2 ee 88 ec 9c   .Og....]Ub......
    0080 - ef 4b 89 a9 e1 f4 81 20-13 5d 20 01 02 a2 59 8c   .K..... .] ...Y.
    0090 - 66 93 5d fd cb c2 01 8a-b1 6e c7 18 75 d8 c0 1b   f.]......n..u...
    00a0 - 6b 4b 04 fa d4 62 8e ba-92 b0 ac 3d 68 f5 fe cf   kK...b.....=h...
    00b0 - 7f f6 c3 7a 8b 8a aa 9a-c1 1b 1e 24 c2 82 90 8d   ...z.......$....
    00c0 - 84 f3 6e 81 42 60 36 af-16 78 d7 68 1f 33 a1 2c   ..n.B`6..x.h.3.,

    Start Time: 1717506496
    Timeout   : 7200 (sec)
    Verify return code: 0 (ok)
    Extended master secret: no
    Max Early Data: 0
---
read R BLOCK
qHTTP/1.0 501 Not Implemented
Content-Type: text/html
Content-Length: 357
Connection: close
Date: Tue, 04 Jun 2024 13:08:55 GMT
Server: ECLF (dcd/7D84)
```

pulkomandy: Yes, I think that would work, no need for a separate directory then. We can share the same ssl/ directory between the two versions

Begasus:
At Fedora they also disable installing the html man/documentation, maybe something to consider seeing there is already a seperate man package? (or move the html files also to the man package)

Begasus:
Rebooted 64bit R1B4 with the openssl3 (and related) packages, now WebPositive crashes on launch (pkgman still works fine).

~> tail -f /var/log/syslog 
2024-06-04 17:46:02 KERN: package_daemon: [1519580032:   762] CommitTransactionHandler::_ChangePackageActivation(): activating 0, deactivating 1 packages
2024-06-04 17:46:02 KERN: unknown: [1519580657:   762] Volume::_ChangeActivation(): 0 new packages, 1 old packages
2024-06-04 17:46:02 KERN: unknown: [1519580758:   762] package "kqtquickcharts24_devel-24.02.2-2-x86_64.hpkg" deactivated
2024-06-04 17:46:03 KERN: package_daemon: [1520080115:   762] Volume::_PackagesEntryRemoved("kqtquickcharts24_devel-24.02.2-2-x86_64.hpkg")
2024-06-04 17:47:13 KERN: slab memory manager: created area 0xffffffffea801000 (330022)
2024-06-04 17:47:13 KERN: slab memory manager: created area 0xffffffffeb001000 (330023)
2024-06-04 17:47:13 KERN: slab memory manager: created area 0xffffffffeb801000 (330024)
2024-06-04 17:47:13 KERN: slab memory manager: created area 0xffffffffec001000 (330025)
2024-06-04 17:47:13 KERN: slab memory manager: created area 0xffffffffec801000 (330485)
2024-06-04 17:49:20 KERN: slab memory manager: created area 0xffffffffed001000 (341347)
2024-06-04 17:50:55 KERN: vm_page_fault: vm_soft_fault returned error 'Bad address' on fault at 0x0, ip 0xddb2bc59b0, write 0, user 1, exec 0, thread 0x2c70
Last message repeated 1 time
2024-06-04 17:50:55 KERN: debug_server: Thread 11376 entered the debugger: Segment violation
2024-06-04 17:50:55 KERN: stack trace, current PC 0xddb2bc59b0  </boot/system/lib/libroot.so> pthread_rwlock_wrlock + 0x10:
2024-06-04 17:50:55 KERN:   (0x7fa439ef0dd0)  0xd910993ed9  </boot/system/lib/libcrypto.so.3> CRYPTO_THREAD_write_lock + 0x9
2024-06-04 17:50:55 KERN:   (0x7fa439ef0de0)  0xd910a0a653  </boot/system/lib/libcrypto.so.3> x509_store_add + 0x53
2024-06-04 17:50:55 KERN:   (0x7fa439ef0e10)  0xd910a0a6f5  </boot/system/lib/libcrypto.so.3> X509_STORE_add_cert + 0x15
2024-06-04 17:50:55 KERN:   (0x7fa439ef0e30)  0xd9109eea0f  </boot/system/lib/libcrypto.so.3> X509_load_cert_crl_file_ex + 0xaf
2024-06-04 17:50:55 KERN:   (0x7fa439ef0e70)  0xd9109eeb65  </boot/system/lib/libcrypto.so.3> by_file_ctrl_ex + 0x55
2024-06-04 17:50:55 KERN:   (0x7fa439ef0ea0)  0xd910a090f4  </boot/system/lib/libcrypto.so.3> X509_STORE_load_file_ex + 0x54
2024-06-04 17:50:55 KERN:   (0x7fa439ef0ee0)  0xfd0796ba7c  </boot/system/lib/libcurl.so.4.8.0> Curl_ssl_setup_x509_store + 0x51c
2024-06-04 17:50:55 KERN:   (0x7fa439ef0f70)  0xfd0796c3e8  </boot/system/lib/libcurl.so.4.8.0> ossl_connect_step1 + 0x598
2024-06-04 17:50:55 KERN:   (0x7fa439ef1100)  0xfd0796cf3f  </boot/system/lib/libcurl.so.4.8.0> ossl_connect_common + 0x11f
2024-06-04 17:50:55 KERN:   (0x7fa439ef1150)  0xfd0796fb5f  </boot/system/lib/libcurl.so.4.8.0> ssl_cf_connect + 0x1cf
2024-06-04 17:50:55 KERN:   (0x7fa439ef11a0)  0xfd0790743b  </boot/system/lib/libcurl.so.4.8.0> cf_setup_connect + 0x22b
2024-06-04 17:50:55 KERN:   (0x7fa439ef1210)  0xfd07901965  </boot/system/lib/libcurl.so.4.8.0> cf_hc_connect + 0x135
2024-06-04 17:50:55 KERN:   (0x7fa439ef1270)  0xfd07905952  </boot/system/lib/libcurl.so.4.8.0> Curl_conn_connect + 0x42
2024-06-04 17:50:55 KERN:   (0x7fa439ef12a0)  0xfd0793e6d0  </boot/system/lib/libcurl.so.4.8.0> multi_runsingle + 0x570
2024-06-04 17:50:55 KERN:   (0x7fa439ef1310)  0xfd0793f824  </boot/system/lib/libcurl.so.4.8.0> curl_multi_perform + 0xc4
2024-06-04 17:50:55 KERN:   (0x7fa439ef13b0)  0xa9f65b5913  </boot/system/lib/libWebKitLegacy.so.1.9.11> _ZN7WebCore20CurlRequestScheduler12workerThreadEv + 0x123
2024-06-04 17:50:55 KERN:   (0x7fa439ef1420)  0xa9f85b00f9  </boot/system/lib/libWebKitLegacy.so.1.9.11> _ZN3WTF6Thread10entryPointEPNS0_16NewThreadContextE + 0x59
2024-06-04 17:50:55 KERN:   (0x7fa439ef1450)  0xa9f860e6f9  </boot/system/lib/libWebKitLegacy.so.1.9.11> _ZN3WTFL19wtfThreadEntryPointEPv + 0x9
2024-06-04 17:50:55 KERN:   (0x7fa439ef1460)  0xddb2bc3ae8  </boot/system/lib/libroot.so> _ZL20pthread_thread_entryPvS_ + 0x18
2024-06-04 17:50:57 KERN: debug_server: Killing team 11358 (/boot/system/apps/WebPositive)
2024-06-04 17:50:57 KERN: debug_server: TeamDebugHandler::Init(): Failed to get info for team 11358: Operation on invalid team
2024-06-04 17:50:57 KERN: debug_server: KillTeam(): Error getting info for team 11358: Operation on invalid team
2024-06-04 17:50:57 KERN: debug_server: Killing team 11358 ()

korli:
@Begasus I enabled symbol versioning in 2b79a5a87fb11694b575f6202fc9228a9b964456. Because it hasn't been enabled for 1.1, it might help or not.

korli:
> @korli I saw you pushed this to master while the changes here are done in a seperate branch, wont those conflict/superseed somewhere? (checking build atm though).

this branch can be rebased on master anytime.

dev-libs/openssl/openssl3-3.0.13.recipe
korli: I tested the x86 and x86_gcc2 builds, they seem OK with the master patchset.

Begasus: Builds went fine here for 32bit too, even building the haiku anyboot image went well on 32bit (need to recheck with latest changes), in qemu everything seems to be working fine.

Begasus:
Converted to draft untill we can figure this out. If unneeded it can be undone.

Begasus:
Looks like there is another one in need for a revbump here:
```
Error:  problem 1: package libevent_devel-2.1.12-2 requires devel:libcrypto>=1.1, but none of the providers can be installed
```

korli:
> Looks like there is another one in need for a revbump here:
> 
> ```
> Error:  problem 1: package libevent_devel-2.1.12-2 requires devel:libcrypto>=1.1, but none of the providers can be installed
> ```

BTW there is an openssl3 patch for libevent 2.1.12 at arch linux.

Begasus:
> BTW there is an openssl3 patch for libevent 2.1.12 at arch linux.

Didn't check, but a simple revbump (with reference >= 3 for openssl) was enough (for now) to get an updated package.

korli:
> Didn't check, but a simple revbump (with reference >= 3 for openssl) was enough (for now) to get an updated package.

Yeah, it seems to be about correctness, refcounting for instance.

Begasus:
> Yeah, it seems to be about correctness, refcounting for instance.

Checked a build for libevent with and without the patch, test case compariable for both, but if we need to revbump maybe just include that too then.

dev-lang/python/python3.9-3.9.18.recipe
```diff
@@ -9,7 +9,7 @@ OSI-approved open source license."
 HOMEPAGE="https://www.python.org"
 LICENSE="Python"
 COPYRIGHT="1990-2023 Python Software Foundation"
-REVISION="1"
+REVISION="2"
```
Begasus, line 12: Any reason why python3.9 is used? I thought python3.10 was/is the default.

pulkomandy:
5 juin 2024 à 09:40 "Schrijvers Luc" ***@***.***> a écrit:



> 
> Next run ... cleaned the Haiku generated and started building a clean anyboot iso with local updated packages:
> 
>  * image boots up fine in Terminal
> 
>  * WebPositive runs fine
> 
> Install build openssl3 package from the github page (packages are updated on changes here)
> 
>  * WebPositive still works fine
> 
> Install build curl package (new)
> 
>  * after reboot in qemu, WebPositive crashes with the same errors as shown earlier
> 
> So it seems curl is the problem here?

To fix this crash, the symbol versioning from Korli may help. So we need to rebase this branch on master and rebuild openssl with all the changes (plus the other things that were discussed here).

Unfortunately I will not have time to look into it this week.

pulkomandy:
5 juin 2024 à 16:39 "Schrijvers Luc" ***@***.***> a écrit:



> 
> ***@***.***** commented on this pull request.
> 
> ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
> 
> In dev-lang/python/python3.9-3.9.18.recipe https://github.com/haikuports/haikuports/pull/10564#discussion_r1627918702 :
> 
> > @@ -9,7 +9,7 @@ OSI-approved open source license."
>  HOMEPAGE="https://www.python.org"
>  LICENSE="Python"
>  COPYRIGHT="1990-2023 Python Software Foundation"
> -REVISION="1"
> +REVISION="2"
> 
> Any reason why python3.9 is used? I thought python3.10 was/is the default.

I kept what was used in Haiku (currently still Python 3.9).

If we decide to ship Python 3.10 in the next beta, we have to update that on Haiku side.
I was trying to do it separately to avoid a huge snowball of updates to Haiku repositories.

korli:
> So it seems curl is the problem here?

WebPositive links against libbnetapi.so, which needs libssl.so.1 and libcrypto.so.1. libWebKitLegacy loads libcurl, which in turn loads libssl.so.3. One probably needs a libbnetapi.so linked against libssl.so.3.

Begasus:
> I kept what was used in Haiku (currently still Python 3.9).

OK, was just wondering, maybe good to not try to do all at once.

Begasus:
> > So it seems curl is the problem here?
> 
> WebPositive links against libbnetapi.so, which needs libssl.so.1 and libcrypto.so.1. libWebKitLegacy loads libcurl, which in turn loads libssl.so.3. One probably needs a libbnetapi.so linked against libssl.so.3.

Not sure how to "fix" this, I guess libbnetapi.so is linked inside the source with libcrypto.so.1, this would mean a rewrite for that?

pulkomandy:
5 juin 2024 à 18:37 "Jérôme Duval" ***@***.***> a écrit:



> 
> > 
> > So it seems curl is the problem here?
> > 
> 
> WebPositive links against libbnetapi.so, which needs libssl.so.1 and libcrypto.so.1. libWebKitLegacy loads libcurl, which in turn loads libssl.so.3. One probably needs a libbnetapi.so linked against libssl.so.3.

Maybe we can simply remove the linking to libnetapi. I think it's only used to fetch favicons, and that could be done with curl.

However, it will be interesting to see if your symbol versioning changes may fix this first.

korli:
> Maybe we can simply remove the linking to libnetapi. I think it's only used to fetch favicons, and that could be done with curl. However, it will be interesting to see if your symbol versioning changes may fix this first.

In my tests the symbol versioning changes didn't help.

waddlesplash:
That seems odd to me; why didn't the symbol versioning changes help? What's the error or crash exactly?

waddlesplash:
(Would rebuilding OpenSSL 1.1.1 with symbol versioning break ABI?)

korli:
> That seems odd to me; why didn't the symbol versioning changes help? What's the error or crash exactly?

It's the one at https://github.com/haikuports/haikuports/pull/10564#issuecomment-2147875790

pulkomandy:
We'd have to test again after merging my changes to Haiku, I removed a function to set up a custom lock handling that is not needed anymore. It's possible that openssl3 is confused by us calling that function for openssl 1.1 somehow (meaning their symbol versioning is not perfect, not a surprise given how the legacy openssl APIs tend to expose the internals of openssl).

korli:
> (Would rebuilding OpenSSL 1.1.1 with symbol versioning break ABI?)

didn't try this.

korli:
Just asking: Is having a branch for r1beta4 at haikuports an option?

Begasus:
> Just asking: Is having a branch for r1beta4 at haikuports an option?

There had been some talks on IRC about something simular, it's a thing for buildmasters too I guess, as they should be aware of it, also pkgman/HaikuDepot should know which one to use?

korli:
What I tested on r1beta4: 
* I added an rpath to libWebKitLegacy where it could find a libcurl linked with libssl.so.1.1:  `patchelf --force-rpath --set-rpath /system/non-packaged/lib/haikuwebkit /system/non-packaged/lib/libWebKitLegacy.so.1.9.1`
* I copied a current libcurl.so.4.8.0 and libcurl.so.4 to /system/non-packaged/lib/haikuwebkit and patched the hardcoded path to the CARootCertificates with `perl -0777 -pe 's/curl-7.85.0-2/substr q{curl-8.7.1-2\/}."\0"x length$&,0,length$&/e or die "pattern not found"' -i /system/non-packaged/lib/haikuwebkit/libcurl.so.4.8.0`
* I installed the new packages openssl3 and curl packages from Begasus.
* I repackaged the openssl3 package to apply this patch  https://salsa.debian.org/debian/openssl/-/blob/baaa04f2b358e06258b3733c8deef58f09011ad7/debian/patches/Remove-the-provider-section.patch
* WebPositive continues to work with https websites.

My proposal is to patch the haikuwebkit package to include a private libcurl.so still using libssl.so.1.1. This makes sure WebPositive in r1beta4 works, independently of the installed curl package. 
Alternatively, a second curl package can be produced, which installs somewhere in /system/lib/curl and have haikuwebkit uses it (dependency + patchelf rpath).
There is a list of recipes which depend on both libcurl and libssl, which would need to be rebuilt, just after enabling curl built with openssl3.

pulkomandy:
I will try to have a look at removing libnetapi usage from HaikuWebKit/WebPositive, that should remove any need for linking openssl 1.1, right?

korli:
> I will try to have a look at removing libnetapi usage from HaikuWebKit/WebPositive, that should remove any need for linking openssl 1.1, right?

it means haikuwebkit und curl have to be switched to openssl3 together at Haikuports. And then it's only possible if libnetapi usage is also removed in r1beta4, as it's the same repo, *OR* switching haikuwebkit to openssl3 can then be done after the r1beta5 release, if the libnetapi usage is removed for the release (I didn't test, whether libnetapi with openssl3 works with a haikuwebkit linked against openssl 1.1).
I must admit, I get lost in the possible upgrade paths.

waddlesplash:
It may make the most sense to do the upgrade only after the beta branch, I suppose. That way new packages won't get back to beta4 or older nightly builds.

korli:
> It may make the most sense to do the upgrade only after the beta branch, I suppose. That way new packages won't get back to beta4 or older nightly builds.

Indeed this a good point.

pulkomandy:
So, from haikuwebkit side, the plan would be:

- Remove haikuwebkit dependency on libnetapi
- After that, move haikuwebkit and curl to openssl 3

If that works, libnetapi can remain linked with openssl 1.1 without problems until a new Haiku release, right?

Unless it's WebPositive and not HaikuWebkit that links to libnetapi, in which case extra steps will be needed.

pulkomandy:
I had a look at haikuwebkit and why it uses libbnetapi. It is not for favicons as I thought, instead it is for implementing websockets, it uses BSecureSocket.

This could be replaced by direct use of openssl if needed, but I'd rather avoid that. There are temporary solutions: the one Korli suggested of a temporary libcurl.so, or I could copy BSecureSocket code into haikuwebkit codebase until the migration is complete.

korli:
The plan to upgrade after the release branch (no beta4 packages anymore) looks preferred to me.

PR #10572, opened 2024-06-06
ktextaddons6, add KF6 package

Begasus:


Begasus:
When I started out with the KF6 frameworks some libraries got the extention "6" to the recipe name, I did the same for the libraries for gear 24 (seeing the libraries are called 24.*.* that makes sense), now with libraries like these which can be build for KF5 and KF6 it's not that clear.
Maybe I should call those kf6_* instead of *6 (like what is done with the qt6 packages eg qt6_* instead of *6)?

PR #10576, opened 2024-06-08
inter font: switch from ttf ot otf

humdingerb:
Apparently it's rendered more accurately, see
https://discuss.haiku-os.org/t/what-font-you-are-using/9878/14

### DIFF
--- a/media-fonts/inter/inter-4.0.recipe
+++ b/media-fonts/inter/inter-4.0.recipe
@@ -6,7 +6,7 @@ to marketing and signage."
 HOMEPAGE="https://rsms.me/inter/"
 COPYRIGHT="2016 Inter Project Authors"
 LICENSE="SIL Open Font License v1.1"
-REVISION="1"
+REVISION="2"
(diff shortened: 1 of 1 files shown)


Begasus:
No strong opinion here, n00b question maybe, can't you have both (otf +  ttf)?

jmairboeck:
> No strong opinion here, n00b question maybe, can't you have both (otf + ttf)?

I'm not sure if they could be distinguished easily, or how app_server would even react. I don't think having both versions has any advantages.

Begasus:
@humdingerb can do the honours then :)

humdingerb:
I'll give it until tomorrow, in case anyone got objections.

If this rendering difference were true for all ttf/otf fonts, should all ttf fonts be replaced with their otf counterpart (if available)? Let me point out that "should" doesn't mean we find a volunteer to go through all fonts to make that change... :)

humdingerb:
Let's merge then!

PR #10579, opened 2024-06-08
miniupnpc, disable gcc2 build

Begasus:


Begasus:
No need for rebuild, afaict no use for it on gcc2.

PR #10590, opened 2024-06-09
Libresprite: resource fix

sabcha1:


Begasus:
Files conflict, could you have a look here?

jmairboeck:
It looks like your branch needs to be rebased on the upstream master.

media-gfx/libresprite/additional-files/setup-data.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+mkdir -r /boot/home/config/settings/libresprite/data
+cp -r /boot/system/data/libresprite/* /boot/home/config/settings/libresprite/data/
```
Begasus, line 4: Better place to store these would be `/boot/home/config/non-packaged/data`

PR #10597, opened 2024-06-12
love, bump version

Begasus:


Begasus:
Checked some downloaded games, launch is fine, aside from some that require an older version (older then 11.*) so the 11.2  doesn't work for those either.
Tried building the older 0.10.2 but build failed on that one.

PR #10599, opened 2024-06-13
Genio: Depend on cmd::clang instead of a specific library version

jackburton79:
Removed dependency to specific llvm version

### DIFF
--- a/haiku-apps/genio/genio-2.1.0.recipe
+++ b/haiku-apps/genio/genio-2.1.0.recipe
@@ -13,7 +13,7 @@ HOMEPAGE="https://github.com/Genio-The-Haiku-IDE/Genio/releases"
 COPYRIGHT="2022-2024 The Genio Team"
 LICENSE="BSD (3-clause)
 	MIT"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://github.com/Genio-The-Haiku-IDE/Genio/archive/v2.1.tar.gz"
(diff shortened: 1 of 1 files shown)


Begasus:
Thanks! 👍

haiku-apps/genio/genio-2.1.0.recipe
```diff
@@ -29,7 +29,7 @@ PROVIDES="
 REQUIRES="
 	haiku${secondaryArchSuffix}
 	gcc${secondaryArchSuffix}_syslibs_devel
-	llvm17${secondaryArchSuffix}_clang
+	cmd:clang$secondaryArchSuffix
```
korli, line 32: should be cmd:clang

Begasus: My bad, @jackburton79 could you do the honours?

jackburton79: Would this work on 32 bit for the secondary arch ?

Begasus: Resolved, thanks! :)

Begasus: > Would this work on 32 bit for the secondary arch ?

Genio is already on secondary arch for 32bit, cmd:clang also, but installed to $prefix/bin instead of $binDir ($prefix/bin/x86).

PR #10600, opened 2024-06-13
clazy, 32bit build fix

Begasus:
Add upstream patches (prepares for building with LLVM18)

### DIFF
--- a/dev-util/clazy/patches/clazy-1.11.patchset
+++ b/dev-util/clazy/patches/clazy-1.11.patchset
@@ -1,4 +1,4 @@
-From 25069b81e93503e892ff6157ee99e8d3f75f18a4 Mon Sep 17 00:00:00 2001
+From 782b36b80abbe8ebe947883aa0c4dca1179268da Mon Sep 17 00:00:00 2001
 From: Cristian Adam <cristian.adam@qt.io>
 Date: Tue, 6 Sep 2022 16:30:02 +0200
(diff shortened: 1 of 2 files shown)


Begasus:
Added LLVM18 (upstream) patches, these don't collide when building with LLVM17, all tests passed on checking 64bit with both clang versions.

dev-util/clazy/patches/clazy-1.11.patchset
```diff
@@ -285,37 +285,179 @@ index b0da926..1ba1126 100644
...
+   add_executable(clazy-standalone ${CLAZY_STANDALONE_SRCS})
+ 
+-- 
+2.45.1
+
+
+From 6fd688c72eda9a1ccf76ac697cc8735ebb7b70fd Mon Sep 17 00:00:00 2001
+From: Schrijvers Luc <begasus@gmail.com>
+Date: Thu, 13 Jun 2024 13:50:31 +0200
+Subject: Disable PCH for 32bit
```
Begasus, line 441: @korli this isn't stricktly needed for 64bit, if so I could move this patch out and only apply it on 32bit only.

Begasus: Could be done later if need be. :)

Begasus:
Checked build on 32bit (R1B4), KDevelop picks it up after installation and with path to the documention set it detects 94 checkes (same as on 64bit (R1B4).

korli:
Thanks!

PR #10606, opened 2024-06-16
opensubdiv3.4, revive lib package for blender 2.*

Begasus:


Begasus:
Without this I'm getting the next:
```
~/Desktop> /boot/system/apps/Blender/Blender
runtime_loader: Cannot open file libosdGPU.so.3.4.3 (needed by /boot/system/apps/Blender/Blender): No such file or directory
```

PR #10611, opened 2024-06-19
minizip-ng, new recipe

Begasus:


Begasus:
Looking at Gentoo's packages I can't find regular minizip in there anymore, maybe they dropped it?
Anyway, this one is needed for newer OpenColorIO.

sys-libs/minizip-ng/minizip_ng-4.0.7.recipe
```diff
@@ -0,0 +1,82 @@
...
+		libminizip-ng
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir
+}
+
+TEST()
+{
+	make check
```
jmairboeck, line 81: This is probably wrong when using CMake with Ninja to build it. I would expect something like `ninja -Cbuild test`.

Begasus: No test done in there, this was fixed in: https://github.com/haikuports/haikuports/pull/10613 :)

PR #10614, opened 2024-06-19
openimageio2.3, rename recipe, drop obselete ones

Begasus:


Begasus:
I know I pushed just a change for this version, but since there are 2 versions in haikuports (third on the way 2.5.12.0) I think it's better to use simular naming scheme for all recipes in here.

Also from what I could find, the older ones are not in use anywhere?

Begasus:
libdeflate is a newer dependency due to https://github.com/haikuports/haikuports/pull/10612 (should be merged first).

PR #10615, opened 2024-06-19
opencolorio2.3, add new version

Begasus:


media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
```diff
@@ -0,0 +1,31 @@
...
+diff --git a/src/utils/NumberUtils.h b/src/utils/NumberUtils.h
+index 08a9ba0..8b0ec8d 100644
+--- a/src/utils/NumberUtils.h
++++ b/src/utils/NumberUtils.h
+@@ -65,7 +65,7 @@ really_inline from_chars_result from_chars(const char *first, const char *last,
+ #ifdef _WIN32
+     tempval = _strtod_l(first, &endptr, loc.local);
+ #else
+-    tempval = ::strtod_l(first, &endptr, loc.local);
++    tempval = ::strtod(first, &endptr);
```
Begasus, line 16: Not sure it's correct, been strugling with it and couldn't find a descent sollution so far. Help wanted. :)

jmairboeck: strtod_l is available under `__USE_GNU`, it seems: https://xref.landonf.org/source/xref/haiku/src/system/libroot/posix/glibc/stdlib/stdlib.h#249

I don't remember exactly how to get that though. Shouldn't that be the default? Does the CMake file set the C++ standard explicitly to override that maybe?

jmairboeck: My bad, this is an internal header in libroot, so it probably isn't available.

media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
```diff
@@ -0,0 +1,31 @@
...
+ #endif
+ 
+     if (errno != 0 && errno != EINVAL)
+@@ -109,7 +109,7 @@ really_inline from_chars_result from_chars(const char *first, const char *last,
+     // On OSX, strtod_l is for some reason drastically faster than strtof_l.
+     tempval = static_cast<float>(::strtod_l(first, &endptr, loc.local));
+ #else
+-    tempval = ::strtof_l(first, &endptr, loc.local);
++    tempval = ::strtod(first, &endptr);
```
jmairboeck, line 25: I think this should be `::strtof` (from the context).

media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
Begasus: Have been searching the function in the headers ... couldn't find it (not even in the source for Haiku).

media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
Begasus: Good call, will check!

Begasus: Fixed. Thanks!

media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
Begasus: Updated patchset to be a bit more inline with OS checks.

media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
```diff
@@ -0,0 +1,33 @@
...
+ #else
+     tempval = ::strtod_l(first, &endptr, loc.local);
+ #endif
+@@ -108,6 +111,9 @@ really_inline from_chars_result from_chars(const char *first, const char *last,
+ #elif __APPLE__
+     // On OSX, strtod_l is for some reason drastically faster than strtof_l.
+     tempval = static_cast<float>(::strtod_l(first, &endptr, loc.local));
++#endif
++#ifdef __HAIKU__
```
jmairboeck, line 26: I think this would be more correct in this case:

```suggestion
+#elif defined(__HAIKU__)
```

for the Haiku case it wouldn't matter, but for the non-Haiku case ;)

Begasus: Need to check, I think I tried that approach ... :)

media-libs/opencolorio/patches/opencolorio-2.3.2.patchset
Begasus: Updated here to be inline with the other one.

PR #10616, opened 2024-06-20
ccache: bump version

Zardshard:
Tested on compiling haikuwebkit. It's much faster and I've encountered no crashes so far, so, looks like a success!


dev-util/ccache/ccache-4.10.recipe
```diff
@@ -43,6 +43,7 @@ BUILD_REQUIRES="
 	devel:libz$secondaryArchSuffix
 	devel:libzstd$secondaryArchSuffix
 	"
+
```
korli, line 46:
```suggestion
```

dev-util/ccache/patches/ccache-4.10.patchset
```diff
@@ -61,5 +61,28 @@ index fcc2134..11e6259 100644
...
+diff --git a/src/ccache/Config.cpp b/src/ccache/Config.cpp
+index b4974e1..8a6836d 100644
+--- a/src/ccache/Config.cpp
++++ b/src/ccache/Config.cpp
+@@ -1194,7 +1194,7 @@ std::string
+ Config::default_temporary_dir() const
+ {
+   static const std::string run_user_tmp_dir = [] {
+-#ifndef _WIN32
++#if !defined(_WIN32) && !defined(HAIKU)
```
jmairboeck, line 82:
```suggestion
+#if !defined(_WIN32) && !defined(__HAIKU__)
```

PR #10622, opened 2024-06-22
raylib: new recipe

hoanga:
the following recipe adds support for [raylib](https://github.com/raysan5/raylib)

sample run

```
> uname -a
Haiku shredder 1 hrev57768 Jun 20 2024 06:26:28 x86_64 x86_64 Haiku

> cat main.c
#include "raylib.h"

int main(void)
{
    InitWindow(800, 450, "raylib [core] example - basic window");

    while (!WindowShouldClose())
    {
        BeginDrawing();
            ClearBackground(RAYWHITE);
            DrawText("Congrats! You created your first window!", 190, 200, 20, LIGHTGRAY);
        EndDrawing();
    }

    CloseWindow();

    return 0;
}

> cc -o helloraylib main.c -lglfw -lraylib

> ls helloraylib
helloraylib
```

### DIFF
--- a/media-libs/raylib/raylib-5.0.recipe
+++ b/media-libs/raylib/raylib-5.0.recipe
@@ -69,7 +69,7 @@ INSTALL()
 	cd build
 	make install
 
-	prepareInstalledDevelLibs \
+	prepareInstalledDevelLib \
 		libraylib
 
 	fixPkgconfig
--- a/media-libs/raylib/raylib-5.0.recipe
+++ b/media-libs/raylib/raylib-5.0.recipe
@@ -38,8 +38,7 @@ BUILD()
 	cd build
 
 	cmake .. -DUSE_EXTERNAL_GLFW=ON \
-		-DCMAKE_BUILD_TYPE=Release \
-		-DCMAKE_INSTALL_PREFIX:PATH="$prefix"
+		-DCMAKE_BUILD_TYPE=Release $cmakeDirArgs
 	make
 }
 
@@ -51,10 +50,4 @@ INSTALL()
 
 	prepareInstalledDevelLibs \
 		libraylib
-
-	oldLibDir=$libDir
-	libDir=$prefix/lib
-
-	mv $prefix/include/* $includeDir/
-	rmdir $prefix/include
 }
--- a/media-libs/raylib/raylib-5.0.recipe
+++ b/media-libs/raylib/raylib-5.0.recipe
@@ -29,7 +29,7 @@ REQUIRES="
 
 PROVIDES_devel="
 	raylib${secondaryArchSuffix}_devel = $portVersion
-	devel:libraylib${secondaryArchSuffix} = $libVersionCompat
+	devel:libraylib$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES_devel="
 	raylib$secondaryArchSuffix == $portVersion base
@@ -42,7 +42,7 @@ BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
 	devel:libGL$secondaryArchSuffix
 	devel:libGLU$secondaryArchSuffix
-	devel:libglfw${secondaryArchSuffix}
+	devel:libglfw$secondaryArchSuffix
 	"
 BUILD_PREREQUIRES="
 	cmd:gcc$secondaryArchSuffix
@@ -72,5 +72,9 @@ INSTALL()
 	prepareInstalledDevelLibs \
 		libraylib
 
-	packageEntries devel $developDir
+	fixPkgconfig
+
+	packageEntries devel \
+		$developDir \
+		$libDir/cmake
 }
--- a/media-libs/raylib/raylib-5.0.recipe
+++ b/media-libs/raylib/raylib-5.0.recipe
@@ -13,17 +13,35 @@ SOURCE_FILENAME="$portVersion.tar.gz"
 ARCHITECTURES="all !x86_gcc2"
 SECONDARY_ARCHITECTURES="x86"
 
+libVersion="4.5.0"
+libVersionCompat="$libVersion compat >= ${libVersion%.**}"
+
 PROVIDES="
 	raylib$secondaryArchSuffix = $portVersion
-	devel:libraylib$secondaryArchSuffix = $portVersion
+	lib:libraylib$secondaryArchSuffix = $libVersionCompat
 	"
 REQUIRES="
 	haiku$secondaryArchSuffix
-	lib:glfw$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libGLU$secondaryArchSuffix
+	lib:libglfw$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	raylib${secondaryArchSuffix}_devel = $portVersion
+	devel:libraylib${secondaryArchSuffix} = $libVersionCompat
+	"
+REQUIRES_devel="
+	raylib$secondaryArchSuffix == $portVersion base
+	devel:libGL$secondaryArchSuffix
+	devel:libGLU$secondaryArchSuffix
+	devel:libglfw$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
 	haiku${secondaryArchSuffix}_devel
+	devel:libGL$secondaryArchSuffix
+	devel:libGLU$secondaryArchSuffix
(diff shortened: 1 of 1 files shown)


media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+SOURCE_URI="https://github.com/raysan5/raylib/archive/refs/tags/$portVersion.tar.gz"
+CHECKSUM_SHA256="98f049b9ea2a9c40a14e4e543eeea1a7ec3090ebdcd329c4ca2cf98bc9793482"
+SOURCE_FILENAME="$portVersion.tar.gz"
+
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	raylib = $portVersion
+	devel:libraylib = $portVersion
```
jmairboeck, line 18: these are both missing `$secondaryArchSuffix`

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+	"
+REQUIRES="
+	haiku$secondaryArchSuffix
+	glfw$secondaryArchSuffix
```
jmairboeck, line 22:
```suggestion
	lib:libglfw$secondaryArchSuffix
```

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	glfw${secondaryArchSuffix}_devel
```
jmairboeck, line 27:
```suggestion
	devel:libglfw$secondaryArchSuffix
```

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,60 @@
...
+{
+	cd build
+	mkdir -p $libDir $includeDir
+	make install
+
+	prepareInstalledDevelLibs \
+		libraylib
+
+	oldLibDir=$libDir
+	libDir=$prefix/lib
```
jmairboeck, line 56: `$libDir` is a haikuporter-specific variable, so these assignments do nothing and can be removed.

Moreover, libraries should generally be installed to `$libDir` and not `$prefix/lib`, which I suppose this is doing. These are different, when a secondary architecture (i.e. x86) is used. Can the library installation path be specified via a CMake option? If the project uses [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), this can be done via `$cmakeDirArgs`. If not, the library could be moved manually like you did with the headers below (before calling `prepareInstalledDevelLib`).

jmairboeck (changes requested):

media-libs/raylib/raylib-5.0.recipe
jmairboeck: Additionally, this is missing `lib:libraylib$secondaryArchSuffix = $portVersion`. The `devel:` entry should be moved to a `_devel` subpackage. Please see existing library packages as examples how to do this.

media-libs/raylib/raylib-5.0.recipe
jmairboeck: Having looked at the CMake configuration files, I think using `$cmakeDirArgs` should work. Then the part moving the headers below can be removed as well.

media-libs/raylib/raylib-5.0.recipe
hoanga: updated

media-libs/raylib/raylib-5.0.recipe
hoanga: updated

media-libs/raylib/raylib-5.0.recipe
hoanga: was able to apply the `$cmakeDirArgs` suggestion. thanks!

media-libs/raylib/raylib-5.0.recipe
hoanga: by default it configures a static library. there is an option to enable shared libs (_BUILD_SHARED_LIBS_). i've applied that option and tried to restructure per suggestion.

jmairboeck (approved):

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+REQUIRES="
+	haiku$secondaryArchSuffix
+	lib:libGL$secondaryArchSuffix
+	lib:libGLU$secondaryArchSuffix
+	lib:libglfw$secondaryArchSuffix
+	"
+
+PROVIDES_devel="
+	raylib${secondaryArchSuffix}_devel = $portVersion
+	devel:libraylib${secondaryArchSuffix} = $libVersionCompat
```
Begasus, line 32:
```diff
-	devel:libraylib${secondaryArchSuffix} = $libVersionCompat
+	devel:libraylib$secondaryArchSuffix = $libVersionCompat
```

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+	devel:libGL$secondaryArchSuffix
+	devel:libGLU$secondaryArchSuffix
+	devel:libglfw$secondaryArchSuffix
+	"
+
+BUILD_REQUIRES="
+	haiku${secondaryArchSuffix}_devel
+	devel:libGL$secondaryArchSuffix
+	devel:libGLU$secondaryArchSuffix
+	devel:libglfw${secondaryArchSuffix}
```
Begasus, line 45:
```diff
-	devel:libglfw${secondaryArchSuffix}
+	devel:libglfw$secondaryArchSuffix
```

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+}
+
+INSTALL()
+{
+	cd build
+	make install
+
+	prepareInstalledDevelLibs \
+		libraylib
+
```
jmairboeck, line 74: Looks good!

Could you add

```sh
fixPkgconfig
```

so that it is moved to the devel package?

Begasus (changes requested):
No brackets needed here

media-libs/raylib/raylib-5.0.recipe
```diff
@@ -0,0 +1,76 @@
...
+	packageEntries devel $developDir
```
jmairboeck, line 75:
```suggestion
	packageEntries devel \
		$developDir \
		$libDir/cmake
```

The cmake files should also go into the devel package (`fixCmake` is still broken in the released version of haikuporter).

media-libs/raylib/raylib-5.0.recipe
Begasus: Hope we don't scare you :) Another small thing, only one library so:
```diff
-	prepareInstalledDevelLibs \
+	prepareInstalledDevelLib \
```

media-libs/raylib/raylib-5.0.recipe
hoanga: updated

media-libs/raylib/raylib-5.0.recipe
hoanga: updated

media-libs/raylib/raylib-5.0.recipe
hoanga: updated to include `fixPkgConfig` and `lib`

i don't understand cmake conventions very well (if at all), so the feedback is helpful!

media-libs/raylib/raylib-5.0.recipe
hoanga: updated, and good to know about `fixCmake`

media-libs/raylib/raylib-5.0.recipe
Begasus: For the above comments cmake is not involved, in case you are wondering about how $cmakeDirArgs work, this takes care of all the default installation paths, so you don't need to declare them all in a recipe, eg not have to worry about -DCMAKE_INSTALL_PREFIX/-DCMAKE_INSTALL_BINDIR etc.

Begasus (approved):

Begasus:
Checked the examples, works nice, thanks!

PR #10626, opened 2024-06-23
perl: bump version, enable threads, don't install sitecustomize.pl

### LINKED ISSUE #10624: perl sitecustomize.pl script
Why do we install a custom `sitecustomize.pl` script in our perl package?

It is intended to change `/packages/`-symlink based paths in `@INC` (the path where perl searches for installed modules) to concrete `/boot/system` and `/boot/home/config` based paths. But what is the problem using the symlink ones? The only difference I see is to use both system paths and user paths, regardless of where the perl package is actually installed, but is that actually needed?

The current version of the file doesn't do anything on x86 because it searches `/packages/perl-$portVersion/.self/` instead of `perl_x86` and therefore doesn't actually change the `@INC` paths.

Also, fixing the installed version of the script via an update is difficult because it is marked as `keep-old` in the recipe (which is correct IMHO).

I propose to remove the `sitecustomize.pl` script from the perl recipe and patchset altogether.

Note: fixing `sitecustomize.pl` doesn't help for `sdl_perl_x86` ...

jmairboeck:
See https://perldoc.perl.org/perl#History for details of changes. Fixes CVE-2023-47038. (the other security fix affects only Windows)

Because this is a major update of perl, this requires rebuilding all perl packages.
 - including git, among others! -- which is why this is still a draft.

Enable threads, which are enabled by most Linux distros as well, apparently. This required some additional patches.
 - maybe this helps with sdl_perl on x86 (we will see)

Don't install the `sitecustomize.pl` script any more. I don't think it is actually needed. If it is retained though, it would need some changes to support secondaryArch. The patches which create the script are still included for now, but they could be removed if this change is accepted.
Fixes #10624.

### DIFF
--- a/dev-lang/perl/patches/perl-5.40.0.patchset
+++ b/dev-lang/perl/patches/perl-5.40.0.patchset
@@ -19,7 +19,7 @@ index 570f25a..81baf48 100644
    # fall-through if we can't unlink now
    _deferred_unlink($fh, $path, 0);
 -- 
-2.45.1
+2.45.2
 
 
 From 9ecf37f2989d220763e11553ce938d0c9095b4f8 Mon Sep 17 00:00:00 2001
@@ -43,7 +43,7 @@ index 7bcd491..85260b8 100644
    } else {
      plan tests => 4;
 -- 
-2.45.1
+2.45.2
 
 
 From 7f9232cdcd444ae11980276a1df2c4be8644ad18 Mon Sep 17 00:00:00 2001
@@ -105,7 +105,7 @@ index 0ec7479..0f09f53 100644
 -
 +# haiku sets all its specifics via Configure
 -- 
-2.45.1
+2.45.2
 
 
 From 58103414df3ca4fa4ba920538f7879ae34a9aeaa Mon Sep 17 00:00:00 2001
@@ -131,7 +131,7 @@ index 3c8af53..1bbbf5b 100755
  
  
 -- 
-2.45.1
+2.45.2
 
 
 From 5b1ca829779068d5c29c456909c3598dc8676206 Mon Sep 17 00:00:00 2001
@@ -182,7 +182,7 @@ index ce3270e..cab9a79 100644
    }
  
 -- 
-2.45.1
+2.45.2
 
 
 From 89ddca4242c23b959888f837199c223f32304a21 Mon Sep 17 00:00:00 2001
@@ -286,7 +286,7 @@ index 0000000..81e5f99
 +__END__
 +
 -- 
-2.45.1
+2.45.2
 
 
 From 8b1508e136aa3dfe3dd032c5d8e17178ada2c4d5 Mon Sep 17 00:00:00 2001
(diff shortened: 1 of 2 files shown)


dev-lang/perl/patches/perl-5.40.0.patchset
```diff
@@ -90,6 +90,8 @@ index fa8ebe5..0f09f53 100644
 -cc="gcc"
 -ld='gcc'
 -
+-ccflags="$ccflags -fno-stack-protector"
+-
```
jmairboeck, line 94: This is cleared out together with the rest of this file as before, and later added back in (see below). Should this be merged together?

dev-lang/perl/patches/perl-5.40.0.patchset
```diff
@@ -513,32 +492,29 @@ index 25ace13..8a04ead 100644
...
- 			eval $checkccflag
+diff --git a/hints/haiku.sh b/hints/haiku.sh
+index 0f09f53..b76c7c1 100644
+--- a/hints/haiku.sh
++++ b/hints/haiku.sh
+@@ -1 +1,3 @@
+-# haiku sets all its specifics via Configure
++# haiku sets nearly all its specifics via Configure
++
++ccflags="$ccflags -fno-stack-protector"
```
jmairboeck, line 512: This is now set via the hints file instead of directly in the Configure script because this was refactored upstream and this is probably the easiest way to do it now (see https://github.com/Perl/perl5/pull/17860).

dev-lang/perl/patches/perl-5.40.0.patchset
```diff
@@ -355,10 +334,10 @@ index 81e5f99..25ace13 100644
  __END__
  
 -- 
-2.21.0
+2.45.1
 
 
-From 57fa1985f6edd826cdb17e40da3c6cce93b40ada Mon Sep 17 00:00:00 2001
+From 888bb0055430fe31da4a8bbf35f8fe2f482273a5 Mon Sep 17 00:00:00 2001
```
jmairboeck, line 340: This is the first of 2 commits that add the `sitecustomize.pl` script.

dev-lang/perl/patches/perl-5.40.0.patchset
```diff
@@ -513,32 +492,29 @@ index 25ace13..8a04ead 100644
...
++# haiku sets nearly all its specifics via Configure
++
++ccflags="$ccflags -fno-stack-protector"
 -- 
-2.21.0
+2.45.1
 
 
-From 7393131191509a0d0c57c7776027b7d61d8d1e63 Mon Sep 17 00:00:00 2001
+From 1f79db86954bb75e4212034e8dde0b6eda989245 Mon Sep 17 00:00:00 2001
```
jmairboeck, line 517: This is the second one.

dev-lang/perl/patches/perl-5.40.0.patchset
```diff
@@ -575,10 +551,10 @@ index a321e51..76c0e46 100644
  );
  my @newINC;
 -- 
-2.21.0
+2.45.1
 
 
-From b500365b5a960c8ad1ba719f3e4d64ce635c3c35 Mon Sep 17 00:00:00 2001
+From a3940547651bbdd2a863be98ba3794bf132c2edf Mon Sep 17 00:00:00 2001
```
jmairboeck, line 557: This one can probably be dropped now that we moved away from gcc2.

dev-vcs/git/git-2.45.2.recipe
```diff
@@ -133,7 +133,7 @@ REQUIRES_scalar="
 	"
 
 # The `subversion` package includes perl bindings that this tool needs.
-# Note: `subversion` still built with gcc2 on 32 bits (matching perl there, at least).
+# Note: `subversion` still built with gcc2 on 32 bits.
```
jmairboeck, line 136: This needs to be checked if this still works (perl is now using gcc 13 on x86).

dev-vcs/git/git-2.45.2.recipe
jmairboeck, line 1: Why does this even depend on `vendor_perl`? The perl files it contains are in `data/perl5`. (How does it work then?)

jmairboeck: We can't get rid of (versioned) perl dependencies anyway, because this also depends on `cmd:perl`, so I went the other way round to install the perl modules into the vendor_perl directory (where they actually belong, IMHO).

korli:
@jmairboeck you might already know: if both git and perl updates are merged together, the buildbot will still possibly build git against the old perl version.

jmairboeck:
> @jmairboeck you might already know: if both git and perl updates are merged together, the buildbot will still possibly build git against the old perl version.

Thanks for the info. Does this also apply when they are separate commits, i.e. when using rebase merging, or must they be pushed completely separately?

jmairboeck:
It may make sense to split the perl parts of git into its own subpackage and let git_cvs, git_email and git_svn depend on that. That would remove the perl dependency at least from the main git package.

dev-vcs/git/git-2.45.2.recipe
Begasus: subversion failed in the INSTALL part for the bindings libsvn_swig_perl on 32bit

jmairboeck: Thanks. I suspected that. I think we should switch subversion to secondaryArch then.

Begasus: I don't know who invented "portPackageLinksDir", but I'm never getting this right ;)
Seems cm:apr_1_config$secondaryArchSuffix should do it (still checking)

korli:
> Thanks for the info. Does this also apply when they are separate commits, i.e. when using rebase merging, or must they be pushed completely separately?

The buildbot master updates to the latest commit, and checks changed recipes. Dependency resolution is done with existing dependencies when they're no constraint. Indeed if a constraint makes the build of a package required, the build master will take care. But for dependencies without constraint, recipes are built out of order.

jmairboeck:
Ah, so that could be fixed by a version constraint in BUILD_REQUIRES in the git recipe.

dev-vcs/git/git-2.45.2.recipe
Begasus: Test result on 64bit are the same as in the subversion recipe (no supprise as there isn't anything really changed), but still. :)

Begasus: @jmairboeck PR for subversion: https://github.com/haikuports/haikuports/pull/10628
Maybe you could check it out? (testcase on 32bit still running)

jmairboeck: I won't have much time to test anything before the weekend, probably.

Begasus: Test results from 32bit: same as on 64bit, so I guess the PR should be good to go. :)

jmairboeck:
> @jmairboeck you might already know: if both git and perl updates are merged together, the buildbot will still possibly build git against the old perl version.

I split the git changes into its own PR (#10649) and separated the perl modules and gitweb into their own subpackages to remove the perl dependency of the main git package.

PR #10628, opened 2024-06-24
subversion, revbump for perl_x86 package on 32bit

Begasus:


dev-vcs/subversion/subversion-1.14.3.recipe
```diff
@@ -31,82 +32,83 @@ PROVIDES="
...
+	devel:libsvn_ra_svn_1$secondaryArchSuffix = 0.0.0 compat >= 0
+	devel:libsvn_repos_1$secondaryArchSuffix = 0.0.0 compat >= 0
+	devel:libsvn_subr_1$secondaryArchSuffix = 0.0.0 compat >= 0
+	devel:libsvn_swig_perl_1$secondaryArchSuffix = 0.0.0 compat >= 0
+	devel:libsvn_wc_1$secondaryArchSuffix = 0.0.0 compat >= 0
 	"
 REQUIRES_devel="
-	subversion == $portVersion base
+	subversion$secondaryArchSuffix == $portVersion base
 	"
```
jmairboeck, line 87: `REPLACES="subversion"` should be added for x86_gcc2

dev-vcs/subversion/subversion-1.14.3.recipe
```diff
@@ -31,82 +32,88 @@ PROVIDES="
...
+	lib:libsqlite3$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
 	"
+if [ "$targetArchitecture" = x86_gcc2 ]; then
+	REPLACES="
+		subversion
+		"
+fi
```
Begasus, line 69: @jmairboeck think this should do it, haven't checked on 32bit yet though

jmairboeck:
Do the perl bindings of subversion depend on the perl version? I suppose they do, so this should be merged after the perl PR.

Begasus:
Needs: https://github.com/haikuports/haikuports/pull/10626 to be merged first. I didn't check against the new version @jmairboeck but I guess it should be ok, could do a build on 64bit for perl and check if this is still fine though.

dev-vcs/subversion/subversion-1.14.3.recipe
```diff
@@ -31,82 +32,88 @@ PROVIDES="
...
+	lib:libaprutil_1$secondaryArchSuffix
+	lib:libcrypto$secondaryArchSuffix
+	lib:libexpat$secondaryArchSuffix
+	lib:libiconv$secondaryArchSuffix
+	lib:liblz4$secondaryArchSuffix
+	lib:libserf_1$secondaryArchSuffix
+	lib:libsqlite3$secondaryArchSuffix
+	lib:libssl$secondaryArchSuffix
+	lib:libxml2$secondaryArchSuffix
+	lib:libz$secondaryArchSuffix
```
jmairboeck, line 63: As this contains perl modules in the vendor_perl directory, should we maybe add `vendor_perl` to the requirements here?

Although then we should probably split the perl bindings into a subpackage (similar to git) to avoid a hard perl dependency of the main package.

jmairboeck (approved):
I noticed that the perl bindings also contain a `perllocal.pod` file which should be avoided for packages because it doesn't make sense to package this file.

Maybe we can fix this by using a "manual" installation of the perl bindings as described [here](https://svn.apache.org/viewvc/subversion/tags/1.14.3/subversion/bindings/swig/INSTALL?view=markup#l235).

But I think all these changes can be done after merging this.

jmairboeck:
> I noticed that the perl bindings also contain a `perllocal.pod` file which should be avoided for packages because it doesn't make sense to package this file. (It contains "documentation" of the currently installed package and is amended by the `make install` process, but it would contain only subversion stuff in this package, and all perl packages would overwrite the file from all other packages.)
> 
> Maybe we can fix this by using a "manual" installation of the perl bindings as described [here](https://svn.apache.org/viewvc/subversion/tags/1.14.3/subversion/bindings/swig/INSTALL?view=markup#l235) and use `make pure_install`.
> 
> But I think all these changes can be done after merging this.

I did this using a small patch instead, similar to what I did with irssi: https://github.com/haikuports/haikuports/commit/d50f57363a8b3b621148562793d1060f0b4787aa#diff-a49a9e8ab59a84d046263803b7075f1fec823ffee4be356300e5206c83440022R25

PR #10629, opened 2024-06-25
EXRTranslator for Haiku

Begasus:


haiku-apps/exrtranslator/exrtranslator-0.1.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+BUILD_PREREQUIRES="
+	makefile_engine
+	cmd:gcc$secondaryArchSuffix
+	cmd:make
+	"
+
+BUILD()
+{
+	if [ -n "$secondaryArchSuffix" ]; then
+		sed -i 's|develop/headers/|develop/headers/x86/|g' Makefile
```
jmairboeck, line 72: We currently have only one kind of secondaryArch, so it doesn't matter, but a generic replacement would be better I think:

```suggestion
		sed -i "s|develop/headers/|develop/headers$secondaryArchSubDir/|g" Makefile
```

Begasus: Can't commit here, doing a check on 32bit and will do a revbump then. 
Pokes @korli in case you are watching this.

korli: secondary arches aren't "all" but only x86, so in this case it doesn't matter.

Begasus: OK, then we can leave it as is. Thanks for responding!

haiku-apps/exrtranslator/exrtranslator-0.1.0.recipe
```diff
@@ -0,0 +1,81 @@
...
+DESCRIPTION="EXR Translator is a translator add-on for displaying EXR images.
+
+Using this add-on will allow any application using Haiku Data Translations \
+to view the EXR file format files."
+HOMEPAGE="https://github.com/HaikuArchives/EXRTranslator"
+COPYRIGHT="2008 Haiku Inc."
+LICENSE="MIT"
+REVISION="1"
+if [ "$effectiveTargetArchitecture" = x86_gcc2 ]; then
+	srcGitRev="6cdbb0b924cbcdeb85c6d55c95b244cc0353c312"
```
korli, line 11: Note: this is commit 62856f5b5b797d00ff1356519ab5ce1c428ea660 + 32-bit changes.

PR #10649, opened 2024-06-29
git: bump version, split perl modules and web into subpackages

jmairboeck:
Install the perl modules into the vendor_perl directory.

This should avoid the perl dependency of the main git package, simplifying perl upgrades.

Note: this is split off from the perl PR (#10626). I suggest merging this before the perl upgrade and revbump git again afterwards, together with the other perl packages.


dev-vcs/git/git-2.45.2.recipe
```diff
@@ -66,6 +66,14 @@ PROVIDES_gui="
...
 	cmd:gitk = $portVersionCompat
 	"
+PROVIDES_perl="
+	git${secondaryArchSuffix}_perl = $portVersionCompat
+	"
+if [ -n "$secondaryArchSuffix" ]; then
+	PROVIDES_perl+="
+		git_perl = $portVersionCompat
+		"
+fi
```
jmairboeck, line 76: Reason for this block: `git_perl` is effectively a perl module package, similar to the ones in dev-perl. This could then be used as a dependency of another any-arch perl package.

waddlesplash:
Aren't there still parts of core Git written in perl, or have those all been replaced?

jmairboeck:
> Aren't there still parts of core Git written in perl, or have those all been replaced?

The git-core package in openSUSE doesn't depend on perl: https://build.opensuse.org/projects/devel:tools:scm/packages/git/files/git.spec?expand=1

According to `grep` in /boot/system/lib/git-core, perl is invoked by two shell scripts: `git-filter-branch` and `git-request-pull`. I think this is okay to not have a hard dependency?

waddlesplash:
Yeah, that's probably fine then.

dev-vcs/git/git-2.45.2.recipe
```diff
@@ -133,12 +151,17 @@ REQUIRES_scalar="
 	"
 
 # The `subversion` package includes perl bindings that this tool needs.
-# Note: `subversion` still built with gcc2 on 32 bits (matching perl there, at least).
 REQUIRES_svn="
 	haiku$secondaryArchSuffix
 	git$secondaryArchSuffix == $portVersion base
+	git${secondaryArchSuffix}_perl == $portVersion
 	subversion
```
jmairboeck, line 158: I forgot adding `$secondaryArchSuffix` here after subversion was switched to secondaryArch. Will fix this when revbumping for the new perl.

jmairboeck:
Apparently, this broke `git help` (as reported by @Begasus on IRC) because the `git-web--browse` shouldn't have been moved to the web subpackage. I will fix this in the next update.

PR #10641, opened 2024-06-27
man_pages_posix: bump to 2017.a 

kenmays:
Bump to 2017.a

The Open Group Base Specifications Issue 7, 2018 edition

IEEE Std 1003.1-2017

PR #10656, opened 2024-06-30
Revert "man_pages_posix: bump to 2017.a "

lonemadmax:
Reverts haikuports/haikuports#10641

The PATCH produces incorrect files for this version, and there's a license in the source removing the paragraph where explicit permission to modify and redistribute was granted.


lonemadmax:
Something I don't know is if this is enough to remove the package from the depot.

Begasus:
If not a revbump should be enough to re-enable the old one, thanks!

Issue #10650, opened 2024-06-29
icu74: bump to 74.2


kenmays:
2023-12-13: ICU 74.2 released with date/time formatting bug fixes.

Begasus:
Please provide a link to the changelog or relevant changes.

Begasus:
From: https://icu.unicode.org/download/74

```
ICU 74.2 maintenance release
Release date: 2023-12-13

ICU 74.2 updates to [CLDR 44.1](https://cldr.unicode.org/index/downloads/cldr-44#h.nvqx283jwsx) locale data. These are maintenance releases for ICU 74 and CLDR 44, with limited sets of bug fixes and no API or structural changes.

The CLDR bug fix relevant for ICU is for some formatting patterns that erroneously had two adjacent space characters. These are coalesced into one. ([CLDR-17233](https://unicode-org.atlassian.net/browse/CLDR-17233))

[List of tickets fixed in ICU 74.2](https://unicode-org.atlassian.net/projects/ICU/versions/10336/tab/release-report-all-issues)

Important: DateFormat.getInstanceForSkeleton() and the DateTimePatternGenerator sometimes used the wrong patterns because they failed to use/inherit certain data. ([ICU-22575](https://unicode-org.atlassian.net/browse/ICU-22575) — CLDR 44 had removed some redundant data that ICU relied on.)
```

Build checked for Haiku 64bit and ICU74.2 still good, but I let others decide on this.

Begasus:
Just did a quick check with installing ICU75 (this needs to be kept in mind also for 74.1 vs 74.2):

```
problem 1: package icu75_devel-75.1-1 conflicts with icu74_devel provided by icu74_devel-74.1-3
  solution 1:
    - allow deinstallation of icu74_devel-74.1-3
    - allow deinstallation of libical_devel-3.0.16-2
    - allow deinstallation of harfbuzz_devel-8.3.0-2
    - allow deinstallation of libvisio_devel-0.1.7-6
    - allow deinstallation of sdl2_ttf_devel-2.20.2-2
    - allow deinstallation of librsvg_devel-2.50.7-4
    - allow deinstallation of freetype_devel-2.13.2-1
    - allow deinstallation of pango_devel-1.50.14-2
    - allow deinstallation of libass_devel-0.17.1-1
    - allow deinstallation of libkdcraw24_devel-24.05.0-1
    - allow deinstallation of qt6_httpserver_devel-6.7.0-1
[...]

PR #10654, opened 2024-06-29
ICU: bump version

kenmays:
Bumped ICU74 to v74.2
Closes: https://github.com/haikuports/haikuports/issues/10650

Ref: https://icu.unicode.org/download/74

"ICU 74.2 maintenance release
Release date: 2023-12-13

ICU 74.2 updates to [CLDR 44.1](https://cldr.unicode.org/index/downloads/cldr-44#h.nvqx283jwsx) locale data. These are maintenance releases for ICU 74 and CLDR 44, with limited sets of bug fixes and no API or structural changes.

The CLDR bug fix relevant for ICU is for some formatting patterns that erroneously had two adjacent space characters. These are coalesced into one. ([CLDR-17233](https://unicode-org.atlassian.net/browse/CLDR-17233))

[List of tickets fixed in ICU 74.2](https://unicode-org.atlassian.net/projects/ICU/versions/10336/tab/release-report-all-issues)

Important: DateFormat.getInstanceForSkeleton() and the DateTimePatternGenerator sometimes used the wrong patterns because they failed to use/inherit certain data. ([ICU-22575](https://unicode-org.atlassian.net/browse/ICU-22575) — CLDR 44 had removed some redundant data that ICU relied on.)"

PR #10658, opened 2024-07-01
Icu: bump to 74.2

kenmays:
Ref: https://icu.unicode.org/download/74
Ref: https://github.com/haikuports/haikuports/pull/10654

Closes: https://github.com/haikuports/haikuports/issues/10650


Begasus:
As mentioned in earlier replies to previous PR, ICU74 is used for Haiku's build infra, will let others decide on this, but atm I think it doesn't make sense on pushing this.
Thoughts from others?

kenmays:
@begasus - Agree, since we now have the PR for ICU 75.1.

Begasus:
A swtich to ICU75 wouldn't be that easy, multiple libraries involved there. ;) Thanks!

Issue #9842, opened 2023-12-07
nheko crashes with an Invalid opcode exception


TheZeldakatze:
I've installed Nheko on a nightly build of Haiku (hrev57439) x86_64. It crashes on startup with an 'invalid Opcode exception'. When looking at the Debugger, the exception happens in __x86_rdseed which seems to come from libstdc++. I've attached the report file.

[Nheko-54541-debug-07-12-2023-00-13-36.report.txt](https://github.com/haikuports/haikuports/files/13593097/Nheko-54541-debug-07-12-2023-00-13-36.report.txt)

kallisti5:
Same issue here.  I tried bumping the revision for a rebuild.. but ran into the same issue.

Begasus:
Still running fine here. (just did a fresh install on 32bit)

Begasus:
R1B4 64bit still OK, maybe a beta vs nightly thing?

TheZeldakatze:
> R1B4 64bit still OK, maybe a beta vs nightly thing?

I've tried both stock and updated beta4 x64, as well as the nightlies, it always crashes with a segment violation. After installing dbus though, it works without a problem

I've not been able to replicate the invalid opcode exception after a system update, so I guess that has been fixed.

I'll have a look if you can build nheko without the dbus requirement, else I'd
make a PR for requiring dbus

Begasus:
Just ran a check on 32bit R1B4, still good there, no crash.

Begasus:
Confirmed that after install dbus there is no crash (checked clean VM 64bit nightly from today).

PR #10660, opened 2024-07-03
nheko, revbump, nheko crashes on clean install when libdbus is not in…

Begasus:
…stalled


Begasus:
Closes: https://github.com/haikuports/haikuports/issues/9842

Qt5 has a dependency here it seems: 

`runtime_loader: Cannot open file libdbus-1.so.3 (needed by /boot/system/lib/libQt6Core.so.6.7.0): No such file or directory`

PR #10666, opened 2024-07-06
texinfo-7: bump version

jmairboeck:
Seems like a straightforward upgrade, but there are two minor issues:

 * there is now a build dependency on gperf. This may have an impact on bootstrapping (I don't know).
 * during configure, conftest crashes during one check, see https://dev.haiku-os.org/ticket/18940


korli:
LGTM

jmairboeck:
How do the builders react to the crash (if it also happens on beta 4)? Are they configured to kill crashing teams automatically?

Begasus:
Got a hunch this broke texinfo? Trying some recipes that use it end up in error, first I thought it was related to changes in gettext. To be sure I launched a build for xorriso on 32bit with no changes to gettext.
First build went fine (using the older 6.2 that was still installed), uninstalled that one and installed new (this), and build for xorriso failed. (first one I tried was maxima on 64bit that failed too, simular error).

jmairboeck:
I haven't really tested building anything else with the new version. It could well be that there are incompatible between 7.0 and 7.1. Should we restore the old version?

Begasus:
Could you make a PR so I could run some checks with it?

Begasus:
Same thing, maybe not related? (paste external) https://bpa.st/KSOA

PR #10669, opened 2024-07-06
tinyxml, add missing pkgconfig file, style changes

Begasus:


Begasus:
This was missing for sourcetrail: https://github.com/petermost/Sourcetrail

dev-libs/tinyxml/tinyxml-2.6.2.recipe
```diff
@@ -57,7 +57,24 @@ INSTALL()
...
+includedir=$includeDir
+
+Name: TinyXML
+Description: A simple, small, C++ XML parser
+Version: $portVersion
+Libs: -L\${libdir} -ltinyxml
+Cflags: -I\${includedir}
+EOF
+
+	fixPkgconfig
```
jmairboeck, line 77: Why is `fixPkgconfig` needed if you are writing it directly to $developLibDir?

Begasus: Left-over from previous attempts I think ;) Will remove (after some testing)

PR #10679, opened 2024-07-11
gnuplot: fix recipe issue preventing build on x86 32 bits.

OscarL:


jmairboeck:
`$libExecDir` is a very nice idea!

PR #10682, opened 2024-07-12
iup: re-enable patch, keep recipe disabled.

OscarL:
Build still broken. Fails while trying to link led.c (code generated by yacc) due to several undefined references.

Commented out lua52 stuff, following changes for "cd" on https://github.com/haikuports/haikuports/pull/10684.

Whoever ends up fixing this build can either uncomment, or remove, them as necessary.

### DIFF
--- a/x11-libs/iup/iup-3.8.recipe
+++ b/x11-libs/iup/iup-3.8.recipe
@@ -17,10 +17,10 @@ REVISION="2"
 SOURCE_URI="http://downloads.sourceforge.net/project/iup/3.8/Docs%20and%20Sources/iup-3.8_Sources.zip" # The tar.gz has permission problems.
 CHECKSUM_SHA256="8030c4f35c3a3f096e6552cecb600d8ee95c9c379c76368d99fa6ffc6ebc8b5c"
 SOURCE_DIR="iup"
-#PATCHES="iup-3.8.patchset"
+PATCHES="iup-3.8.patchset"
 
-ARCHITECTURES="x86_gcc2"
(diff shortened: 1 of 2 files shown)


x11-libs/iup/iup-3.8.recipe
```diff
@@ -17,10 +17,10 @@ REVISION="2"
 SOURCE_URI="http://downloads.sourceforge.net/project/iup/3.8/Docs%20and%20Sources/iup-3.8_Sources.zip" # The tar.gz has permission problems.
 CHECKSUM_SHA256="8030c4f35c3a3f096e6552cecb600d8ee95c9c379c76368d99fa6ffc6ebc8b5c"
 SOURCE_DIR="iup"
-#PATCHES="iup-3.8.patchset"
+PATCHES="iup-3.8.patchset"
 
-ARCHITECTURES="x86_gcc2"
-SECONDARY_ARCHITECTURES="x86"
+ARCHITECTURES="?all ?x86_gcc2"
+SECONDARY_ARCHITECTURES="?x86"
```
Begasus, line 23: This disabled the package, users that don't already have it can't install it, is that intentional?

jmairboeck: The `?x86_gcc2` part is redundant, it is included in `?all`. (Architectures need only be specified if their type (`!`, `?` or nothing) is different from `all`).

Begasus: Well ... maybe it needs to be re-added if "all" will be enabled at one point. :)

PR #10683, opened 2024-07-12
luarocks, revbump, add more finetuning for mutltiple lua versions

Begasus:


dev-lua/luarocks/luarocks-3.11.1.recipe
```diff
@@ -47,18 +48,22 @@ BUILD_PREREQUIRES="
 
 BUILD()
 {
+	local sysIncDir="$(finddir B_SYSTEM_HEADERS_DIRECTORY)$secondaryArchSubDir"
 	./configure --prefix=$prefix \
-		--with-lua-include="$(finddir B_SYSTEM_HEADERS_DIRECTORY)$secondaryArchSubDir/lua54" \
+		--with-lua-include="$sysIncDir/lua54" \
```
Begasus, line 53: @FractalU while this $sysIncDir is perfectly fine, why use it here? Does it change the behaviour as done with the previous version?

FractalU: No, it doesn't change behavior. The `$sysIncDir` variable is a replacement for the old code and has the exact same value. I set `$sysIncDir` because I use it multiple times and to reduce boilerplate. I don't know whether `$includeDir` can be used instead.
 
`--with-lua-include="$sysIncDir/lua54"` is used to set `LUA_INCDIR` in `config.unix`, a config file generated by `./configure`, which is then being used to set `LUA_INCDIR` in `config-5.4.lua`. `config-5.4.lua` is automatically being generated by `make build`.

Begasus: OK, no problem, sounds good enough (just that I only seen it used here in a few recipes).
We can't use $includeDir (mentioned in the issue) because it points to `./self/develop/headers` and we need `/boot/system/headers` for this.

FractalU:
@Begasus Actually, `make build` can be replaced with `make`. It's also more conformant with `make install`. I checked the `GNUmakefile`.

Please, wait with merging. There is more coming.

Begasus:
Merging for now, please report back on later/new issues. Thanks so far!

PR #10727, opened 2024-07-22
gdb: Add recipe for GDB 15.1

trungnt2910:
This is a comprehensive port built from the ground up.

x86 support is dropped, but can be added in a later commit if absolutely required.

### DIFF
--- a/dev-util/gdb/gdb-15.1.recipe
+++ b/dev-util/gdb/gdb-15.1.recipe
@@ -1,0 +1,120 @@
+SUMMARY="The GNU debugger"
+DESCRIPTION="GDB is the GNU debugger. It allows you to see what is going on \
+'inside' another program while it executes -- or what another program was \
+doing at the moment it crashed.
+
+GDB can do four main kinds of things (plus other things in support of \
+these) to help you catch bugs in the act:
+
+ * Start your program, specifying anything that might affect its behavior.
+ * Make your program stop on specified conditions.
+ * Examine what has happened, when your program has stopped.
+ * Change things in your program, so you can experiment with correcting \
+the effects of one bug and go on to learn about another.
+
+The program being debugged can be written in Ada, C, C++, Objective-C, \
+Pascal (and many other languages). Those programs might be executing \
+on the same machine as GDB (native) or on another machine (remote). \
+GDB can run on most popular UNIX and Microsoft Windows variants."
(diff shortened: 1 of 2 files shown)


Begasus:
From comment on the forum this needs nightly, so as it is atm it can't be build on the buildmasters.

dev-util/gdb/gdb-15.1.recipe
```diff
@@ -0,0 +1,116 @@
...
+
+TEST_REQUIRES="
+	cmd:runtest
+	"
+
+BUILD()
+{
+	runConfigure --omit-dirs "docDir dataRootDir" ./configure \
+		--disable-nls --enable-gdb --enable-gdbserver \
+		--with-system-zlib --with-system-readline
```
waddlesplash, line 91: I think you need a --with-python here (to ensure GDB actually picks up the python, if unspecified it may silently miss it.)

waddlesplash: It looks like we specifically need `--with-python=python3`.

waddlesplash: And apparently `export HAIKU_USE_VENDOR_DIRECTORIES=1` to tell Python to export the correct directories so GDB can find the headers.

trungnt2910: Resolved in new recipe.

dev-util/gdb/patches/gdb-15.1.patchset
```diff
@@ -0,0 +1,7181 @@
+From 9c9cd5cf820ea4f83b72409dcaaa797f351d2ea1 Mon Sep 17 00:00:00 2001
+From: Trung Nguyen <me@trungnt2910.com>
+Date: Mon, 29 Jul 2024 22:55:42 +1000
+Subject: [PATCH 1/3] gdb/remote: fix assertion failure during startup
```
waddlesplash, line 4: What's the link for the upstream report for this patch?

waddlesplash:
Buildmasters now on beta5 branch base packages, so we can merge this.

dev-util/gdb/patches/gdb-15.1.patchset
trungnt2910: https://sourceware.org/pipermail/gdb-patches/2024-July/210771.html

PR #10733, opened 2024-07-23
libuv, bump version

Begasus:


Begasus:
Not really required, but in preperation for newer LST nodejs.

PR #10737, opened 2024-07-24
borgbackup: recipe/dependencies cleanup.

OscarL:
Used `readelf -d | grep NEEDED` on the compiled modules to make sure we only require what is actually needed at runtime.

Double checked documented runtime dependencies on Python modules, grepped the sources for the "dependencies" I removed, and smoke tested the provided `cmd:`s after install.

Cleaned up build requirementes while at it. In particular, `cmd:pip` wasn't actually required, (and we should avoid using it on recipes, in any case, IMO, at least in a way we don't control, and using `--no-deps` while at it).


Begasus:
Missing `cmd:pkg_config$secondaryArchSuffix`

Begasus:
Fixed in: https://github.com/haikuports/haikuports/pull/10741 :)

Begasus:
> Will fix.

That would be an even better fix! Thanks! :)

PR #10744, opened 2024-07-24
python_pkgconfig: recipe cleanup/fixes.

OscarL:
While this module could be "any" in theory, doing so means the "x86_gcc2" version of "pkg-config" gets called (that causes problems when packages that use this module to build stuff, get the wrong info on 32 bits, no sane python3 package would be asking for, or expecting, GCC2 stuff).

Cleaned up a bit the recipe while at it, and droped 3.9, as the only user on-tree of this module is Python 3.10 only already (borgbackup).

---

Tested builds of this and `borgbackup` (after reverting #10741) on both 32 and 64 bits. If this gets merged, will open a PR removing from `borgbackup` the transitive/indirect requirement for `cmd:pkg_config$secondaryArchSuffix`.

### DIFF
--- a/dev-python/pkgconfig/python_pkgconfig-1.5.5.recipe
+++ b/dev-python/pkgconfig/python_pkgconfig-1.5.5.recipe
@@ -4,58 +4,74 @@ Python 3.3+."
 HOMEPAGE="https://pypi.org/project/pkgconfig/"
 COPYRIGHT="201-2021 Mathias Vogelgesang"
 LICENSE="MIT"
-REVISION="3"
-SOURCE_URI="https://pypi.python.org/packages/source/p/pkgconfig/pkgconfig-$portVersion.tar.gz"
+REVISION="4"
+SOURCE_URI="https://files.pythonhosted.org/packages/source/p/pkgconfig/pkgconfig-$portVersion.tar.gz"
 CHECKSUM_SHA256="deb4163ef11f75b520d822d9505c1f462761b4309b1bb713d08689759ea8b899"
 SOURCE_DIR="pkgconfig-$portVersion"
 
-ARCHITECTURES="any"
+# Ideally, this should be "any", but we really want "pkg-config-x86" here, not the gcc2 version.
+# Python modules using this one have no concept of Haiku's x86_gcc2 vs x86, and we can pretty
+# much assume they only care for the non-gcc2 version of libs.
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	$portName = $portVersion
 	"
 REQUIRES="
-	haiku
+	haiku$secondaryArchSuffix
 	"
 
 BUILD_REQUIRES="
-	haiku_devel
+	haiku${secondaryArchSuffix}_devel
 	"
 
-PYTHON_PACKAGES=(python39 python310)
-PYTHON_VERSIONS=(3.9 3.10)
-for i in "${!PYTHON_PACKAGES[@]}"; do
-pythonPackage=${PYTHON_PACKAGES[i]}
-pythonVersion=${PYTHON_VERSIONS[$i]}
-eval "PROVIDES_${pythonPackage}=\"\
-	${portName}_$pythonPackage = $portVersion\
-	\"; \
-REQUIRES_$pythonPackage=\"\
-	haiku\n\
-	cmd:pkg_config\n\
-	cmd:python$pythonVersion\
-	\""
-BUILD_REQUIRES="$BUILD_REQUIRES
-	setuptools_$pythonPackage"
-BUILD_PREREQUIRES="$BUILD_PREREQUIRES
-	cmd:python$pythonVersion"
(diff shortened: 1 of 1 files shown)


dev-python/pkgconfig/python_pkgconfig-1.5.5.recipe
```diff
@@ -4,58 +4,74 @@ Python 3.3+."
...
 
-ARCHITECTURES="any"
+# Ideally, this should be "any", but we really want "pkg-config-x86" here, not the gcc2 version.
+# Python modules using this one have no concept of Haiku's x86_gcc2 vs x86, and we can pretty
+# much assume they only care for the non-gcc2 version of libs.
+ARCHITECTURES="all !x86_gcc2"
+SECONDARY_ARCHITECTURES="x86"
 
 PROVIDES="
 	$portName = $portVersion
```
jmairboeck, line 19: I think, this needs to be changed to `python_pkgconfig$secondaryArchSuffix = $portVersion`. And maybe a "suffix-less" provide added conditionally for x86.

Begasus: I tend to agree with @jmairboeck here, this one is a bit of a weird one in the python recipes, so far only used by borgbackup. But it does search cross compiler (gcc2/gcc13) for library headers.
`$portName` on 32bit already refers to `python_pkgconfig_x86`, adding a suffix-less provides would make it handier to leave out the `$secondaryArchSuffix` on this?

korli: Personally I'd stay with "any", then check at runtime in python for the right executable at: https://github.com/matze/pkgconfig/blob/master/pkgconfig/pkgconfig.py#L108 
https://github.com/matze/pkgconfig/blob/master/pkgconfig/pkgconfig.py#L123

sys.platform.startswith("haiku") and platform.machine() == 'BePC'  then 'pkg-config-x86'

korli: IMO #10741 is correct, the build depends on this right tool.

jmairboeck: > We don't do that for other `ARCHITECTURES="all !x86_gcc2"` `dev-python/` recipes, AFAICS.
> 
> (see for example the recipes for `cffi`, `argon2-cffi-bindings`, `numpy`, `contourpy`).
> 
> So if that change is needed, then the same would apply for all the rest of recipes. I'd suggest opening an issue for that, so it is easier to keep track of progress.
> 
> FWIW, I haven't seen yet a "non _x86" `PROVIDES` conditionally added for the "main" package (only for `_python3x`). Do we ever add on "package A" a `REQUIRES` for "package_b" instead of a "package_b_python3x" version specific one? (the latter should all already provide the "non x86" names).

Ah, I missed the fact that this is just an empty base package and the real contents are in version-specific subpackages. The main package doesn't need a non-suffixed PROVIDE then. Sorry for the confusion.

pulkomandy: There may be a diffferent way to fix the problem here. It's possible to set the PKG_CONFIG_LIBDIR (if I remember the variable name correctly) to tell pkgconfig to search for .pc file in the lib/x86 directory. If you do this, you can then use the x86_gcc2 version of pkg-config to build software for the x86 architecture.

I'm not sure why we are not doing that generally (compiling pkg-config with x86 only, and using this variable to point it to the right place), but, if that's the only problem, maybe it is simpler to just set this environment variable in the recipes that use python_pkgconfig, and then the package itself will require no special case?

Begasus: > It's possible to set the PKG_CONFIG_LIBDIR (if I remember the variable name correctly

This should probably be: `PKG_CONFIG_PATH`, I use it sometimes in a cmake script to check out new things like:

`export PKG_CONFIG_PATH+=/boot/home/config/non-packaged/lib/pkgconfig/`

korli:
How do you know which `pkg-config` executable is being used? I see in https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/6565/builds/83828.log both package are available:
```
/boot/home/haikuports/packages/pkgconfig-0.29.2-4-x86_gcc2.hpkg
/boot/home/haikuports/packages/pkgconfig_x86-0.29.2-4-x86_gcc2.hpkg
```

jmairboeck:
> Making pkg-config "x86" only, and using PKG_CONFIG_PATH where/if needed sounds interesting.

Maybe that could be even set by `setarch` or whatever haikuporter is using to change the environment when building secondaryArch packages.

dev-python/pkgconfig/python_pkgconfig-1.5.5.recipe
korli: Well it's just a python wrapper. Furthermore "pip install pkgconfig" won't install pkg-config for you.

PR #10746, opened 2024-07-25
labplot, bump version

Begasus:


Begasus:
Build failure on 32bit, it seems to pull in Qt6 there (poppler?) need to investigate.

PR #10749, opened 2024-07-25
gsplus, revbump, fix installation on 32bit

Begasus:


Begasus:
Maybe disable this for 32bit, reported at the forum, and local checks confirmed on 64bit it's  working fine, on 32bit (nightly) it gives an error on launch.

PR #10750, opened 2024-07-26
python3.10: patch set re-work.

### LINKED ISSUE #10211: python3 packages: at least some packages can't be used if installed via `pkgman install -H`
Writing this one down before I forget.

At least some packages can't be used properly if installed via `pkgman install -H`.

Example:

`pkgman install -H ./coverage_python310-7.4.3-1-x86_64.hpkg`, then:

```
> python3 -c "import coverage"
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'coverage'
```

and trying to run `coverage`:

```
> coverage
  File "/boot/home/config/bin/coverage", line 33, in <module>
    sys.exit(load_entry_point('coverage==7.4.3', 'console_scripts', 'coverage')())
  File "/boot/home/config/bin/coverage", line 22, in importlib_load_entry_point
    for entry_point in distribution(dist_name).entry_points
  File "/packages/python3.10-3.10.13-3/.self/lib/python3.10/importlib/metadata/__init__.py", line 969, in distribution
    return Distribution.from_name(distribution_name)
  File "/packages/python3.10-3.10.13-3/.self/lib/python3.10/importlib/metadata/__init__.py", line 548, in from_name
    raise PackageNotFoundError(name)
importlib.metadata.PackageNotFoundError: No package metadata was found for coverage
```

----

For reference:

```python
>>> import sysconfig, pprint
>>> sysconfig.get_scheme_names()
('haiku', 'haiku_home', 'haiku_user', 'haiku_vendor', 'nt', 'nt_user', 'osx_framework_user', 'posix_home', 'posix_prefix', 'posix_user')
>>> pprint.pprint(sysconfig.get_paths('haiku'))
{'data': '/packages/python3.10-3.10.13-3/.self/non-packaged',
 'include': '/packages/python3.10-3.10.13-3/.self/non-packaged/develop/headers/python3.10',
 'platlib': '/packages/python3.10-3.10.13-3/.self/non-packaged/lib/python3.10/site-packages',
 'platstdlib': '/packages/python3.10-3.10.13-3/.self/lib/python3.10',
 'purelib': '/packages/python3.10-3.10.13-3/.self/non-packaged/lib/python3.10/site-packages',
 'scripts': '/packages/python3.10-3.10.13-3/.self/non-packaged/bin',
 'stdlib': '/packages/python3.10-3.10.13-3/.self/lib/python3.10'}
>>> pprint.pprint(sysconfig.get_paths('haiku_home'))
{'data': '/packages/python3.10-3.10.13-3/.self',
 'include': '/packages/python3.10-3.10.13-3/.self/develop/headers/python',
 'platinclude': '/packages/python3.10-3.10.13-3/.self/develop/headers/python',
 'platlib': '/packages/python3.10-3.10.13-3/.self/lib/python',
 'platstdlib': '/packages/python3.10-3.10.13-3/.self/lib/python',
 'purelib': '/packages/python3.10-3.10.13-3/.self/lib/python',
 'scripts': '/packages/python3.10-3.10.13-3/.self/bin',
 'stdlib': '/packages/python3.10-3.10.13-3/.self/lib/python'}
>>> pprint.pprint(sysconfig.get_paths('haiku_user'))
{'data': '/boot/home/config/non-packaged',
 'include': '/boot/home/config/non-packaged/develop/headers/python3.10',
 'platlib': '/boot/home/config/non-packaged/lib/python3.10/site-packages',
 'platstdlib': '/boot/home/config/non-packaged/lib/python3.10',
 'purelib': '/boot/home/config/non-packaged/lib/python3.10/site-packages',
 'scripts': '/boot/home/config/non-packaged/bin',
 'stdlib': '/boot/home/config/non-packaged/lib/python3.10'}
>>> pprint.pprint(sysconfig.get_paths('haiku_vendor'))
{'data': '/packages/python3.10-3.10.13-3/.self',
 'include': '/packages/python3.10-3.10.13-3/.self/develop/headers/python3.10',
 'platlib': '/packages/python3.10-3.10.13-3/.self/lib/python3.10/vendor-packages',
 'platstdlib': '/packages/python3.10-3.10.13-3/.self/lib/python3.10',
 'purelib': '/packages/python3.10-3.10.13-3/.self/lib/python3.10/vendor-packages',
 'scripts': '/packages/python3.10-3.10.13-3/.self/bin',
 'stdlib': '/packages/python3.10-3.10.13-3/.self/lib/python3.10'}
>>> 
```

OscarL:
Re-worked the patchset, mostly to avoid having scattered changes to the same areas, trying to keep each "commit" on a narrower focus/topic.

Fixes #10211, and attempts to fix some remaining _INSTALL_SCHEMES issues. (makes `platinclude` point to the system headers for the default scheme, which should avoid having to set HAIKU_USE_VENDOR_DIRECTORIES=1 when bulding gdb).

---

Built and tested (hp --test python310) on 64 bits. Built and smoke-tested on 32 bits too.

### DIFF
--- a/dev-lang/python/patches/python3.10-3.10.14.patchset
+++ b/dev-lang/python/patches/python3.10-3.10.14.patchset
@@ -1,14 +1,18 @@
-From 5be29419ed996bdc3b6545c3f7c5d4cf345e43ca Mon Sep 17 00:00:00 2001
+From 8f72108a053605954500f69c86fa62714553cd73 Mon Sep 17 00:00:00 2001
 From: Oscar Lesta <oscar.lesta@gmail.com>
-Date: Sun, 7 Apr 2024 00:24:39 -0300
-Subject: Initial Haiku patch.
+Date: Thu, 25 Jul 2024 17:26:26 -0300
+Subject: Initial Haiku patch (re-worked a bit).
 MIME-Version: 1.0
 Content-Type: text/plain; charset=UTF-8
 Content-Transfer-Encoding: 8bit
 
 This combines:
-- Original "initial Haiku patch" from Jérôme Duval.
-- A variation of Adrien Destugues changes in setup.py from his x86 patch.
+
+* The original "initial Haiku patch" from Jérôme Duval.
+* A variation of Adrien Destugues changes in setup.py from his x86 patch.
+
+Ideally, we can continue moving changes from here into "per topic" commits,
+leaving here mostly only make/config related changes, for example.
 
 diff --git a/Include/pyport.h b/Include/pyport.h
 index 6ab0ae4..ad8c45c 100644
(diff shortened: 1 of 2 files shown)


dev-lang/python/patches/python3.10-3.10.14.patchset
```diff
@@ -750,7 +532,7 @@ index ec3cece..6a29d2a 100755
 2.45.2
 
 
-From 390b9112392e602714714c8bb0275beee85914fe Mon Sep 17 00:00:00 2001
+From a5f44dabeeae79d50ad890d398e3a531917bb3c6 Mon Sep 17 00:00:00 2001
 From: Jerome Duval <jerome.duval@gmail.com>
 Date: Fri, 4 Oct 2019 22:02:35 +0200
 Subject: since 3.8, don't reinit locks on fork.
```
jmairboeck, line 538: Completely unrelated to this change, but:

I haven't looked at the corresponding code exactly, but this boosts my confidence that my "scary" patch for perl (https://github.com/haikuports/haikuports/pull/10626/files#diff-0870d8eb3e20e6a6321e6e8e4c72df07fcf42e600ddd6bd53491fbcd1ceb904dR536-R568) is actually correct on Haiku (judging from that commit message).

waddlesplash: Mh, this doesn't sound right to me. Why doesn't it work just the same as other POSIX OSes? We acquire all the heap locks in libroot and unlock them after fork.

waddlesplash: The EPERM would probably be because the child has a new thread ID and is trying to unlock a mutex that is locked by the parent thread. I think it should reinit the locks instead, but it looks like that's what this code does...?

The perl code is more interesting. Does it really just try to unlock the locks in the child? I'm not sure how that would work on Linux either.

jmairboeck: I'm not sure if these two problems (python and perl) are related at all.

If I decoded the error number correctly, I'm pretty sure I got EPERM when building perl.

waddlesplash: From what I am seeing, perl on `musl` would have the same problem -- unless it uses `PTHREAD_MUTEX_NORMAL`, it appears? So maybe we need to adjust Haiku's behavior for that.

jmairboeck: miniperl apparently calls the unlock function in both the parent and the child after fork. See https://github.com/Perl/perl5/blob/d8935409c94992d3e665cdaa1c5411eb55a0e9e2/miniperlmain.c#L87-L100

waddlesplash: It looks like Perl is using `pthread_mutexattr_default`, which means we get `PTHREAD_MUTEX_DEFAULT`. The specification indicates that implementations are allowed to map this to one of the other types, and Haiku basically has it behave like `PTHREAD_MUTEX_ERRORCHECK`, so we get this behavior.

waddlesplash: Ultimately your patch probably isn't safe though, the right thing to do here is to re-initialize the locks in the child. This probably qualifies as a bug in Perl because it is relying on behavior that the POSIX specification explicitly states is undefined (see https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_unlock.html). I would report it upstream, honestly.

waddlesplash: This code in Python looks correct to me however. @korli, any recollection of why you added this patch?

waddlesplash: https://github.com/haikuports/haikuports/commit/2993b575fc087a4c91c087fa4f00b299473a1eb4

So, we should retest with `muon test` I suppose and see what happens on a nightly build.

korli: You mean "meson test", right?

waddlesplash: Yes. May be possible that the hangs were related to bugs in Haiku's locks that have since been fixed.

jmairboeck: > Ultimately your patch probably isn't safe though, the right thing to do here is to re-initialize the locks in the child. This probably qualifies as a bug in Perl because it is relying on behavior that the POSIX specification explicitly states is undefined (see https://pubs.opengroup.org/onlinepubs/9699919799/functions/pthread_mutex_lock.html). I would report it upstream, honestly.

It looks like there is already a ~10 year old issue about this in perl: https://github.com/Perl/perl5/issues/13237 with an old Debian patch for that. Although it seems that Debian (kFreeBSD) doesn't seem to be affected by it any more.

PR #10755, opened 2024-07-27
texstudio, bump version

Begasus:


app-office/texstudio/texstudio-4.8.1.recipe
```diff
@@ -43,41 +43,42 @@ REQUIRES="
...
+#	texlive_fontsextra
+#	texlive_fontsrecommended
+#	texlive_formatsextra
+#	texlive_games
+#	texlive_humanities
+#	texlive_latexrecommended
+#	texlive_latexextra
+#	texlive_pictures
+#	texlive_mathscience
+#	texlive_music
```
Begasus, line 63: @jmairboeck any thoughts on this, I noticed when installing without these packages around it's really hard to tell if HaikuDepot is still running or not. :)

jmairboeck: I think you are right. The texlive packages shouldn't be a hard dependency.

Begasus: Thanks for confirming. :)

PR #10762, opened 2024-07-28
ccache, revbump, move static library and create devel package

Begasus:


Begasus:
Doing another quick check for static libraries in $libDir revealed this, looking at the current package it also shows develop files included all in the base package (where they shouldn't be needed).

Zardshard:
Tested this change on building HaikuWebKit mostly from cache. Looks good!

Begasus:
Thanks, there shouldn't be any functional change, needed to know that. :)

PR #10769, opened 2024-07-29
psutil, bump version, enable builds

Begasus:


Begasus:
Ran some checks with psutil installed, Vorta and Spyder are running fine with it (seems to be a dependency for it, fails to launch also when not installed).

Vorta

Spyder-5 4 5

Both Vorta and Spyder installed through pip.

PR #10779, opened 2024-07-31
indent, revbump, don't depend on base gettext package

Begasus:


Begasus:
Same as with glib2 PR, don't depend on base package, require _devel package.

Begasus:
Related issue: https://github.com/haikuports/haikuports/issues/10780

PR #10786, opened 2024-08-03
gertty: Recipe cleanups. Build only once.

OscarL:
This is a stand-alone utility. No need to build it for multiple Python versions (switched to "default" Python 3.10).

Removed unneeded or indirect depedencies.

For example, `mako` is really a dependency of `alembic_py`. That package should be fixed instead of listing deps here (not sure what, if any, actually requires `typing-extensions`).

---

Draft for now, until `alembic_py` gets fixed, and I manage to test gertty against some real server.

### DIFF
--- a/dev-util/gertty/gertty-1.6.0.recipe
+++ b/dev-util/gertty/gertty-1.6.0.recipe
@@ -19,9 +19,8 @@ changes."
 HOMEPAGE="https://pypi.org/project/gertty/"
 COPYRIGHT="2022 The TTY Group"
 LICENSE="Apache v2"
-REVISION="7"
-pypiVersion="13/fa/67165ece7a08f7142bcfda2e5cee145dbda78e003c11924098bfc6efaf0e"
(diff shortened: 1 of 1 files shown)


dev-util/gertty/gertty-1.6.0.recipe
```diff
@@ -38,58 +37,49 @@ BUILD_REQUIRES="
...
-			--root=/ --prefix=$prefix
+	mkdir -p $installLocation
 
-		packageEntries  $pythonPackage \
-			$prefix/lib/python* \
-			$prefix/bin \
-			$docDir
-	done
+	$python setup.py build install \
+		--root=/ --prefix=$prefix
```
jmairboeck, line 84: Now that this is built only once, maybe move the `build` part to a `BUILD()` function? The `python` variable declaration could be moved out of the function into global scope to not repeat it.

jmairboeck: It was just a suggestion. You can also leave it as is if this works better.

Begasus:
Let's see how this goes. Thanks!

PR #10793, opened 2024-08-06
python3.13: new recipe.

OscarL:
~Draft, as this is still just a release candidate~ (final version expected for October). Edit: using official 3.13.0 relase now.

The REPL improvements alone are nice, so maybe some one wants to give it an early spin.

~(I haven't tried yet builds with the experimental `--disable-gil`)~ Edit: GIL-free/free-threaded builds do not currently work on Haiku.


jmairboeck:
I think the recipe should be renamed to `python3.13-3.13.0~rc1.recipe`. Preview versions are denoted by a `~` (like haiku-r1~beta4). Then the final version will be recognized as newer, when the package is updated later.

korli:
LGTM

PR #10794, opened 2024-08-06
chocolate-doom: Fix copyright line

fragglet:
Currently only the last name in the list is displayed:

Screenshot 2024-08-06 at 10 40 06 AM

### DIFF
--- a/games-fps/chocolate-doom/chocolate_doom-3.0.1.recipe
+++ b/games-fps/chocolate-doom/chocolate_doom-3.0.1.recipe
@@ -3,10 +3,7 @@ DESCRIPTION="Chocolate Doom is a source port of the game Doom, by id Software. \
 The purpose of Chocolate Doom is to be as compatible as possible with the \
 original DOS version of Doom."
 HOMEPAGE="https://www.chocolate-doom.org/"
-COPYRIGHT="Id Software Inc. 1993-1996
-	Simon Howard
-	James Haley
-	Samuel Villarreal"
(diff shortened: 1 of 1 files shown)


jmairboeck:
I think this is more an issue of HaikuDepot that it doesn't display all the copyright entries, but only the last one.

fragglet:
> I think this is more an issue of HaikuDepot that it doesn't display all the copyright entries, but only the last one.

Agreed, although in the meantime it would be nice to have this merged as a workaround

jmairboeck:
If you want the package to actually get rebuilt, you have to increase the REVISION variable (if it built successfully before).

fragglet:
> If you want the package to actually get rebuilt, you have to increase the REVISION variable (if it built successfully before).

Done - thanks for the tip.

korli:
@andponlin should this be fixed in Haiku Depot?

Begasus:
This should rather be fixed in HaikuDepot application, don't think this PR is the way to go to use COPYRIGHT (some have pretty long lists so using this as a reference would be a bad idea (my 2 cents)).

Begasus:
Closing as invalid for now, multiple recipes suffer from this and so should be handled by HaikuDepot app.

andponlin:
See [here](https://review.haiku-os.org/c/haiku/+/8012) for the HaikuDepot change covering this. The UI display will require some further work, but if you view the package a "tool tip" is available which will show the full text.

Begasus:
Thanks for working on this @andponlin !

PR #10796, opened 2024-08-10
iaito, bump version, switch python version

Begasus:


Begasus:
Needs radare2 to be build on buildmasters before merging.

PR #10798, opened 2024-08-10
guile 3.0: new version of guile

jmairboeck:
This can be installed separately from guile 2.2 (and guile 1).

I added a "-3.0" suffix to the commands to avoid an otherwise unnecessary conflict on the _tools and _devel packages. The libraries are designed to allow side-by-side installation and contain that suffix anyway.

Or should we make this new version the "default" instead and add a suffix or conflicts to the tools in guile 2.2 instead?

The next version of lilypond (probably 2.26) will switch to guile 3.0 exclusively. The current version (2.24) could be built with guile 3.0, but 2.2 is still the recommended version.

### DIFF
--- a/dev-scheme/guile/guile3.0-3.0.10.recipe
+++ b/dev-scheme/guile/guile3.0-3.0.10.recipe
@@ -1,0 +1,151 @@
+SUMMARY="Libraries for GNU Ubiquitous Intelligent Language for Extensions"
+SUMMARY_tools="Binaries for GNU Ubiquitous Intelligent Language for Extensions"
+DESCRIPTION="Guile is a library designed to help programmers create flexible \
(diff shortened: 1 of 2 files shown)


jmairboeck:
Tests don't build with missing symbols (linker errors) in some tests, which I don't understand. It is missing `rpl_free` (a wrapper around `free` which preserves `errno`), which is defined in lib/libgnu.a, which should be linked by the relevant tests.

It can be worked around by `#undef free` in these tests, but I'm not sure if we should do that.

jmairboeck:
I now switched the defaults around, making 3.0 the default without command suffixes and adding it to version 2.2 instead.

Also did some cleanups while at it, e.g. removed the `$portVers` variable, which is named weirdly.

PR #10799, opened 2024-08-11
gdb: Improve GDB 15 stability

trungnt2910:
Fix issues, including random hangs, attach issues in multithreaded programs, and memory read failures.

Full list of fixes is available [here](https://github.com/trungnt2910/gdb-haiku/commits/b74dc3ba/).


trungnt2910:
If nobody says otherwise, then OK.
I've been attaching this to various programs, including GDB itself, and it seems good enough.

However, if you want to avoid having a new PR 3 days later, then maybe wait until next week when the GSoC project officially ends.

Begasus:
Fair enough, can wait untill GSoC ends :)

PR #10802, opened 2024-08-11
nodejs: retire older recipes.

OscarL:
Recipes on-tree require cmd:node >= 16 or 20 already.


Begasus:
qtwebengine builds are disabled on buildmasters (hence the _bin packages), so it doesn't interfere with it here.
On a revbump for qtwebengine I guess a revision for ICU would be handy too.

PR #10804, opened 2024-08-11
Fix referencing to projectx (template leftovers)

Begasus:


Begasus:
Not revbumping some here, let's have the packages around that exist untill a next version bump is needed.

PR #10812, opened 2024-08-12
Added port for Seer (seergdb)

ilzu:
Now that there is a working gdb port here is a QT frontend for it. This seems to have quite nice visualisers for inspecting memory.


Begasus:
Trailing whitespace in the recipe, we'll hold off untill PR for gdb is merged and the whitespace is fixed. Thanks!

ilzu:
Removed that trailing whitespace. This should work without #10799 but of course having gdb itself working better makes this work better too, so I'd say wait for that to be merged first.

Begasus:
Almost forgot this one here. :) Thanks!

Issue #10815, opened 2024-08-12
Update gtkwebkit to less outdated depedencies


pulkomandy:
Hello,

Just some notes on my attempts to clean up old versions of libraries on my system. This is all for x86 secondary architecture.

- gtkwebkit is currently built against ICU 66 instead of a newer 74 or even 75
- I tried to rebuild it, but it needs libevent, which depends on openssl 1.1.1
- I tried to rebuild libevent, but that failed due to trying to use kqueue but using parts of it that we don't implement yet (EVFILT_SIGNAL)

- gtkwebkit, qt5 and fluidsynth2 also depend on flac13 instead of the newer flac1.4.
- I tried to rebuild fluidsynth2, but it fails fo find header files from libnsdispatch.

Of course, help is welcome to fix some of these issues :)

pulkomandy:
Yes, forcing libevent to not enable kqueue support until this is improved on Haiku side should be fine. I think disabling just parts of the kqueue implementation would be a bad idea, it would then only half-work

Begasus:
> * gtkwebkit is currently built against ICU 66 instead of a newer 74 or even 75

It even uses LLVM12

Begasus:
webkit-gtk needs cmd:xgettext atleast in BUILD_PREREQUIRES, but build errors atm (link will stay up for a week): https://bpa.st/LGMA

waddlesplash:
> I think disabling just parts of the kqueue implementation would be a bad idea, it would then only half-work

It turns out the signal handling is basically separate from the event handling and can be safely disabled (and libevent can then use its standard signal handling routines.) I sent @OscarL a patch to test, seems to work here after a fix to kqueue in Haiku, and should be much more performant (and not prone to strange races.)

pulkomandy:
Thanks, I was able to rebuild it, should be fine now!

Begasus:
@pulkomandy so the fluidsynth PR can be closed?

pulkomandy:
I'm not sure what the status is with flac13 on x86, didn't check that yet (and I will not be in front of my 32bit machine for a few days).

Begasus:
Build wise flac13 is fine, I only started to look into it because it was mentioned here, if there are no problems anymore maybe it's pointless medling with the existing packages atm? (given that there are no newer versions involved)

pulkomandy:
[...]

PR #10821, opened 2024-08-12
libevent: (kqueue backend) Allow building without EVFILT_SIGNAL.

OscarL:
Patchset thanks to waddlesplash.

(reference: https://github.com/haikuports/haikuports/issues/10815)

While we're at it, remove an unused .patchset for an older version.

----

Tested building on hrev57953, 64 bits. Tested runtime by using `links` for quite a while. No more 100% CPU usage! (with this and after `libbsd.so` changes from [hrev57959](https://cgit.haiku-os.org/haiku/tag/?h=hrev57959)).


korli:
Disabling directly in the recipe should be more correct, havekqueue=no running ./configure

jmairboeck:
According to the configure script, the correct variable to set is `ac_cv_func_kqueue=no`. (autoconf is weird ...)

PR #10823, opened 2024-08-12
links: update to version 2.30.

OscarL:
Tested on hrev57953, 64 bits.

As before, mouse wheel works on text mode, but not on graphical (`links -g`) mode. :-/


Begasus:
Could you re-create (hp -e) the patchset?

PR #10790, opened 2024-08-05
ffmpeg: bumped to 6.1.2

kenmays:
FFmpeg 6.1.2:
- [x] - Haiku R1B4
- [x] - Haiku R1B4 hrev57915

PR #10832, opened 2024-08-14
FFmpeg: 6.1.2 bump

kenmays:
Haiku R1B5: Removed FFmpeg 4.2.9.
Ref: https://github.com/haikuports/haikuports/pull/10790
- [x] - Haiku R1B5 hrev57966


jmairboeck:
ffmpeg 4 is still used by some (non-system) recipes, like OpenCASCADE.

I'm not against removing it (see https://github.com/haikuports/haikuports/commit/3ea257a7f928ecf28177b9c9b3e2ee67ae4f5dfb), but we should at least check affected packages and upgrade or disable ffmpeg usage for them.

What about ffmpeg 5?

kenmays:
Keep them all for now.

PR #10847, opened 2024-08-17
libpostproc, depend on ffmpeg5_devel for now

Begasus:
Although old, seems to be still in use, so use ffmpeg5_devel for now.


Begasus:
Needs patching when building for ffmpeg6.

PR #10853, opened 2024-08-19
Noto Sans CJK: Use variable width font

waddlesplash:
We probably should add a new package with the fixed-width fonts for compatibility with applications that don't support variable-width fonts yet. (Haiku itself and Qt 6 should already, though.)

### DIFF
--- a/media-fonts/noto_sans_cjk/noto_sans_cjk-2.004.recipe
+++ b/media-fonts/noto_sans_cjk/noto_sans_cjk-2.004.recipe
@@ -6,38 +6,15 @@ especially when multiple styles and weights and even languages are mixed on \
 a page.
 
 This package supports Japanese, Korean and simple/traditional Chinese.
-Included are various weights and two 'monospaced' typefaces ('regular' and \
-'bold')."
+Included are the Sans and Monospace fonts as variable-width typefaces."
 HOMEPAGE="https://www.google.com/get/noto/"
-COPYRIGHT="2012 Google Inc."
+COPYRIGHT="2012-2021 Google Inc."
 LICENSE="SIL Open Font License v1.1"
-REVISION="1"
+REVISION="2"
 # Japanese
-SOURCE_URI="https://github.com/googlefonts/noto-cjk/releases/download/Sans$portVersion/06_NotoSansCJKjp.zip"
-CHECKSUM_SHA256="bf5b26feb83ee02533d4eb4a127bd90ed9d42bdd7ca7cb6e4030b2f21ef55cc5"
+SOURCE_URI="https://github.com/notofonts/noto-cjk/releases/download/Sans$portVersion/01_NotoSansCJK-OTF-VF.zip"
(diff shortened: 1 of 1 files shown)


Begasus:
@waddlesplash anything to be done here?

waddlesplash:
I think we need to test with the various applications still on Qt 5 and see what happens (or patch them, or add a secondary package.)

kenz-gelsoft:
I can test this as a Japanese user.

Testing Qt5 apps and setting fonts to 
Noto Sans CJK is enough?

waddlesplash:
Yes, I think so.

kenz-gelsoft:
CuteMarkEd

cutemarked_vf
- NotoSans CJK works with bold

---

Calligra

Appearance_CJK
- Set NotoSans CJK to Haiku Appearance Plain Font
- Thin fonts for Labels other than Menus

ThinFontOtherThanMenus

- Bold text works with NotoSans CJK for document font

CalligraDocFont

---

Notepadqq

NotePadQQ

ThinOnlyCJK

NotoSansHasManyWeight

waddlesplash:
> Weight other than thin is not available for Qt5 apps. 

Yeah, that's the expected behavior, since Qt5 doesn't support variable-width fonts.

IIRC, @lonemadmax mentioned something about a patch for Qt5 that adds such support? But if the current behavior is acceptable as-is then maybe we don't need to bother.

kenz-gelsoft:
> But if the current behavior is acceptable as-is then maybe we don't need to bother.

It is better if Regular weight will be used as current Thin weight feels thin ;-)

But it is useable and if it is limited to Qt5 apps, it should be acceptable. I assume major Qt5 apps will be upgraded to Qt6.

Release and get feedback!

lonemadmax:
I'm the worst. I get distracted too easily and forgot I had that lying around. Let's see if I can go back to it, clean up and publish what I have before the year ends.

Without the patch, Qt5 will only know about the first style in a variable font, which given those tests seems to be Thin in this case. An app may get synthesized bold and italic when it asks for them, but the only style listed will be that first one. I guess in most fonts it's either regular or the thinnest weight with the rest of axis in "normal" values, but if it happened to be be a bold one, that's what you would get, and I think there's no synthesized non-bold. In theory it could even be something like creepy-upside_down-condensed-heavy-italic in one of those fancy fonts.

> Thin fonts for Labels other than Menus

You'll get the same thin style in Qt5 menus if you set Noto Sans CJK as menu font in Appearance. My old eyes can't tell for sure, but it seems you are also getting thin CJK substitutions in notepadqq's menus.

It was also mentioned in the [maillist](https://www.freelists.org/post/haiku-development/About-variable-width-fonts-and-beta5) that Qt6 also suffered the first style only issue for menu font.

kenz-gelsoft:
> > Thin fonts for Labels other than Menus
> 
> You'll get the same thin style in Qt5 menus if you set Noto Sans CJK as menu font in Appearance. My old eyes can't tell for sure, but it seems you are also getting thin CJK substitutions in notepadqq's menus.

Oh, you're right. I changed only Plain font.

PR #10856, opened 2024-08-21
proj, revbump due to tiff changes

Begasus:


Begasus:
Also added cmd:diff to TEST_REQUIRES, fixes failing tests.

Issue #10818, opened 2024-08-12
Broken QuiteRSS


kim1963:
After update Qt5 .hpkg to version 5.15.14-2

waddlesplash:
Please give more information about what is broken.

kim1963:
RSS download not work.

kim1963:
After revert Qt5 .hpkg
news reader work fine/

lonemadmax:
Can be closed after Qt5 fix.

PR #10858, opened 2024-08-21
Qt5: look for SSL library with version 3 suffix

### LINKED ISSUE #10836: Broken Otter browser
screenshot313

hrev57937+36 64 bit

### LINKED ISSUE #10837: yt_dlp wanting to install ffmpeg-4.2.9 etc.
I was just about to install yt-dlp on the beta5-test-image and wondered why it wants to install ffmpeg-4.2.9, ffmpeg_avdevice-4.2.9 and ffmpeg_tools-4.2.9.  Why doesn't it go for the ffmpeg6-6.1.1_* versions? ffmpeg6-6.1.1-3-x86_64.hpkg is already installed with the beta5-test-image...

lonemadmax:
Fixes #10836 and #10818

Only built and tested on a recent x86_64 nightly with openssl3.

Just below the changed lines, the typical unix lib paths are added, maybe those should also be changed to finddir calls for Haiku.

### DIFF
--- a/dev-qt/qt5/patches/qtbase-5.15.14.patchset
+++ b/dev-qt/qt5/patches/qtbase-5.15.14.patchset
@@ -1,5 +1,4 @@
-
-From 5d39eaa4c2578632be296d2cebbc4b830c77ccff Mon Sep 17 00:00:00 2001
+From 162eebac0d16049549b138401c9ba2a76bf3dd79 Mon Sep 17 00:00:00 2001
 From: =?UTF-8?q?Fran=C3=A7ois=20Revol?= <revol@free.fr>
 Date: Mon, 11 Feb 2019 15:53:49 +1000
 Subject: Make sure libs are searched in the develop/ dir by CMake
@@ -7,7 +6,7 @@ Subject: Make sure libs are searched in the develop/ dir by CMake
 Static libs at least are not in the lib[/x86] dirs...
 
 diff --git a/mkspecs/features/create_cmake.prf b/mkspecs/features/create_cmake.prf
-index 24ed125..52f637e 100644
+index 64639f1..42ad698 100644
 --- a/mkspecs/features/create_cmake.prf
 +++ b/mkspecs/features/create_cmake.prf
 @@ -112,6 +112,9 @@ win32:!static:!staticlib {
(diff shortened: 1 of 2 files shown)


waddlesplash:
Makes sense to me. Anything this is waiting for in particular?

lonemadmax:
A 32-bit check. I can't currently build it even without changes, it looks like some issue on my side: sometimes make crashes, sometimes it doesn't find something.

Until proven wrong, I'm guessing it wouldn't work, as LIBRARY_PATH contains lib, not lib/x86.

waddlesplash:
I think we search the path automatically on `dlopen` if a non-absolute path is specified. So, can't we just patch Qt to do that?

lonemadmax:
> I think we search the path automatically on `dlopen` if a non-absolute path is specified. So, can't we just patch Qt to do that?

It seems that's what the function is trying to do here. Without the details for different OSes, it builds a list of paths from LD_LIBRARY_PATH and then adds /lib, /usr/lib and company.

The new version just takes B_{USER,SYSTEM}{NONPACKAGED,}_LIB_DIRECTORY for Haiku.


> I guess this fixes #10836 instead of #10837?

Right, thank you.

waddlesplash:
> Without the details for different OSes, it builds a list of paths from LD_LIBRARY_PATH and then adds /lib, /usr/lib and company.

No, I mean that on Haiku, `dlopen("libcrypto.so.3")` (for example) should automatically search through the given library paths. Plus this should automatically respect "Disable user addons" and secondary architecture subdirectories, which manually searching the paths won't. So if you can make the code do that (and it works) that would be ideal.

lonemadmax:
> `dlopen("libcrypto.so.3")` should automatically [work]. So if you can make the code do that (and it works) that would be ideal.

In the end it was already doing that. Almost.

Manually looking in the lib dirs for libcrypto.so.* until finding one that loads is the *third* thing it tries, the first one being what it is supposed to do: `dlopen("libcrypto.so.<version>")`, with version taken from SHLIB_VERSION_NUMBER, which is removed in OpenSSL 3. Using OPENSSL_SHLIB_VERSION solves the issue.

waddlesplash (approved):

PR #10861, opened 2024-08-22
lxml: update to version 5.3.0.

OscarL:
Fix crash on "import lxml.etree" related to calling xmlInitParser() too late. See: https://github.com/lxml/lxml/pull/370

---

Tested `lxml_python39` with `Sigil` (loaded even an almost 20 MB .epub), and `lxml_python310` with `hp --test gtk_doc`.

The latter has one test failing, but that's the same case with the older lxml 4.9.1.

Didn't tested with Calibre because it has way too many depedencies for my taste.


waddlesplash (approved):

Begasus:
Thanks!

PR #10863, opened 2024-08-23
qrcode, re-instate, in use by external application, bump version

Begasus:


Begasus:
@OscarL didn't clean the recipe fully, maybe other python versions will be added.

Begasus:
> Just for reference's sake, this was previously dropped in #10805.
> 
> Can't the external application just pip install this (or ask their users to do so, or even include them as part of that app?).
> 
It can, there is however a possibility that the application will end up in haikuports tree also (it's one that uses Haiku-PyAPI).

> I'm started to be against packaging pip-installable things not needed on-tree. Of course, my opinion is just that in the end.
> 
Agreed, hence I merged the removal at first. :)

> Edit: depending on what that app needs, [segno](https://pypi.org/project/segno/) might be an option, or qrcode without PIL (that seems to be an "extra" for qrcode, making it even more easily pip-installable).

Begasus:
> Mmm, do we _really_ need the Python 3.9 version too? (that I'm a bit more against :-D)

Saw that comming :D Haiku-PyAPI is "at the moment" also for both python versions, hence the addition. :)

PR #10866, opened 2024-08-24
Qt6, bump for openssl3

Begasus:


Begasus:
This wont work like this, wanted to bump all in one, but seeing the packages already exist they wouldn't follow the correct sequence needed. (would only be helpfull on a version bump), will look again later.

Begasus:
Closing .. will do seperate PR's for this.

PR #10868, opened 2024-08-24
qt6_shadertools, revbump for openssl3 changes to qt6_base

Begasus:


Begasus:
Needs to wait untill qt6-base is done.

PR #10869, opened 2024-08-24
argparse-manpage: new python recipe

jmairboeck:
This is needed for the new version of [psutils](https://github.com/rrthomas/psutils).

Versions < 3 of psutils were written in C and Perl, for version 3 they were rewritten in Python. Currently they are shipped as part of texlive_core (in version 1.23) but I would like to "unbundle" them.

As I'm not that skilled with python recipes yet, could someone please review this? @OscarL

### DIFF
--- a/dev-python/argparse-manpage/argparse_manpage-4.6.recipe
+++ b/dev-python/argparse-manpage/argparse_manpage-4.6.recipe
@@ -1,0 +1,108 @@
+SUMMARY="Automatically build man-pages for your Python project"
+DESCRIPTION="Avoid documenting your Python script arguments on two places! This is typically done \
+in an argparse.ArgumentParser help configuration (help=, description=, etc.), and also in a \
+manually crafted manual page.
+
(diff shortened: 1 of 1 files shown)


jmairboeck:
Yes, psutils requires cmd:paper at runtime, and a few other things to build first. I'm not quite sure if we should really add all these packages as recipes.

This PR is just for argparse-manpage for now, because that requires a patch to install the manpages in the correct location (also for downstream packages then).

jmairboeck:
I added libpaper 2 today, which provides cmd:paper, because of this. See https://github.com/haikuports/haikuports/commit/71dcb7728b042770ec88ac052d731162637de47b

jmairboeck:
cmd:paper is needed for psutils. This is just one of its requirements. psutils as a separate package are coming later :)

jmairboeck:
Ah, sorry, I missed that. It did build, but I suppose it would fail at runtime for other pyproject.toml-based packages then. I will add that dependency.

PR #10872, opened 2024-08-25
qt6_declarative, revbump for openssl3

Begasus:


dev-qt/qt6-declarative/qt6_declarative-6.7.0.recipe
```diff
@@ -187,6 +222,8 @@ INSTALL()
 		$developLibDir/pkgconfig/*.pc
 	sed -i 's,${_IMPORT_PREFIX}/lib',$developDir/lib, \
 		$libDir/cmake/Qt6PacketProtocolPrivate/Qt6PacketProtocolPrivateTargets-release.cmake
+	sed -i 's,${_IMPORT_PREFIX}/lib',$developDir/lib, \
+		$libDir/cmake/Qt6QmlBuiltins/Qt6QmlBuiltinsTargets-release.cmake
```
jmairboeck, line 226: Could these commands (all of them) be replaced with `fixCmake` maybe? (it was probably still broken when this was written)

Begasus: For some reason not all cmake files in $developLibDir need this patch, tried and got errors on not finding requirements, this change here is only for a static (standalone) library that gave errors not being found.

Begasus: PS, must admit that I "forgot" that fixCmake was "fixed" in latest changes. :D

PR #10890, opened 2024-08-25
qt6_positioning, revbump for openssl3

Begasus:


Begasus:
Wait for qt6_serialport to be launched on buildmasters.

PR #10892, opened 2024-08-25
pypdf: new python recipe

jmairboeck:


dev-python/pypdf/pypdf-4.3.1.recipe
```diff
@@ -0,0 +1,69 @@
...
+PYTHON_VERSIONS=(3.10)
+for i in "${!PYTHON_VERSIONS[@]}"; do
+	pythonVersion=${PYTHON_VERSIONS[$i]}
+	pythonPackage=python${pythonVersion//.}
+	eval "PROVIDES_$pythonPackage=\"
+		${portName}_$pythonPackage = $portVersion
+		\"
+	REQUIRES_$pythonPackage=\"
+		haiku
+		cmd:python$pythonVersion
```
jmairboeck, line 38: This has some optional dependencies:
 * cryptography for supporting encrypted PDFs
 * Pillow for extracting images

I suppose I should leave them out and not require them? @OscarL

jmairboeck: Thanks, then I'll leave them out. The old version of psutils didn't even support PDF at all, so that is already an improvement.

I guess that this is needed for supporting PDF. For psutils, it is a hard requirement.

PR #10901, opened 2024-08-25
wand: new python recipe

jmairboeck:
The last one of the dependencies of psutils.

Could you please have a look again @OscarL ?

### DIFF
--- a/dev-python/wand/wand-0.6.13.recipe
+++ b/dev-python/wand/wand-0.6.13.recipe
@@ -1,0 +1,68 @@
+SUMMARY="The ctypes-based simple ImageMagick binding for Python"
+DESCRIPTION="Wand is a ctypes-based simple ImageMagick binding for Python, supporting 2.7, 3.3+, \
+and PyPy. All functionalities of MagickWand API are implemented in Wand."
+HOMEPAGE="https://wand-py.org/"
+COPYRIGHT="2011-2018 by Hong Minhee
+	2019-2023 by E. McConville"
+LICENSE="MIT"
+REVISION="1"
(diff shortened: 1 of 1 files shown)


jmairboeck:
We could keep this as "any" by cheating and using `cmd:magick` (or some of the other ones provided by ImageMagick) instead of `lib:libImageMagick$secondaryArchSuffix`.

dev-python/wand/wand-0.6.13.recipe
```diff
@@ -0,0 +1,68 @@
+SUMMARY="The ctypes-based simple ImageMagick binding for Python"
+DESCRIPTION="Wand is a ctypes-based simple ImageMagick binding for Python, supporting 2.7, 3.3+, \
+and PyPy. All functionalities of MagickWand API are implemented in Wand."
+HOMEPAGE="https://wand-py.org/"
+COPYRIGHT="2011-2018 by Hong Minhee
+	2019-2023 by E. McConville"
+LICENSE="MIT"
+REVISION="1"
+pypiVersion="59/d5/1bdd7c9662d5e9078e25ba0eb69bdb122859295746d40ab8dfef3a7b4d42"
```
jmairboeck, line 9: BTW: Is there a reason to keep this separately from SOURCE_URI when it is only used there? Is the idea that the actual SOURCE_URI doesn't change on upgrades?

jmairboeck:
> FWIW, recipe looks good to me. I think just using .whl's makes the most sense for things like this, that we only add due to chained dependencies (no real need to deal with, in this case, the gigantic >11 MiB tarball).
> 
> Commenting this because I _think_ not all other devs might agree (at least not in all cases).

I did something similar also for java recipes before, like the antlr ones. These also just repackage the .jar download.

PR #10904, opened 2024-08-25
libass: upgrade to 0.17.3 and enable libunibreak

TheOneric:
I’m not sure if revision should be bumped if the change is part of the same PR as a version bump; for now i included a bump so both commits also work in isolation

### DIFF
--- a/media-libs/libass/libass-0.17.3.recipe
+++ b/media-libs/libass/libass-0.17.3.recipe
@@ -2,22 +2,32 @@ SUMMARY="Library for SSA/ASS subtitles rendering"
 DESCRIPTION="Libass is a portable subtitle renderer for the ASS/SSA (Advanced \
 Substation Alpha/Substation Alpha) subtitle format."
 HOMEPAGE="https://github.com/libass/libass"
-COPYRIGHT="2006-2021 libass contributors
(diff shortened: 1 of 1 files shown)


media-libs/libass/libass-0.17.3.recipe
```diff
@@ -2,22 +2,32 @@ SUMMARY="Library for SSA/ASS subtitles rendering"
...
+	1994 Sun Microsystems Inc
+	1988-1993 The Regents of the University of California
+	2015-2018 Janne Grunau
+	2015 Martin Storsjo
+	2018 Two Orioles LLC
+	2018 VideoLAN and dav1d authors
+	2021 王一 Wang Yi"
 LICENSE="ISC"
-REVISION="1"
+REVISION="2"
```
jmairboeck, line 23: REVISION should stay as 1. It is supposed to be increased when the recipe is updated for the same version as before so that the package is rebuilt and available as an update. When upgrading to a new version, it is reset to 1.

TheOneric: fixed

TheOneric:
friendly ping

Begasus:
Thanks!

PR #10913, opened 2024-08-26
texlive_core: bump version (bug fix update), unbundle external packages

jmairboeck:
Tex Live has many external tools which it bundles, but the upstream is different for them. I thought that it might make sense to move them to separate recipes instead.

~~I'll probably also try to update the `createSubpackageInfos` script in texlive so that it generates `cmd:...` REQUIRES for the texlive packages so that these external packages are proper dependencies then. Let's see if this works out.~~

This also contains a minor bug fix update for texlive_core itself (version 2024.2). This could be pushed separately, but I didn't yet to not unnecessarily rebuild it. The kpathsea version is unchanged (still 6.4.0), so nothing changes on the texlive side for this. (but that one would require a revbump anyway when the extra REQUIRES are added, so we can retry building it on the buildmasters then ...)


dev-tex/gsftopk/gsftopk-1.19.2.recipe
jmairboeck, line 1: I'm not really sure about this one. The "upstream" doesn't really seem to be maintained any more (last update in 2000!) and I even backported a change from texlive to support our modern ghostscript (see the patchset).

dev-tex/dvi2tty/dvi2tty-6.0.2.recipe
```diff
@@ -0,0 +1,78 @@
...
+	cmd:diff
+	"
+
+defineDebugInfoPackage dvi2tty$secondaryArchSuffix \
+	$prefix/bin/disdvi \
+	$prefix/bin/dvi2tty
+
+BUILD()
+{
+	cd source/texk/dvi2tty
```
jmairboeck, line 60: This one uses the texlive build system (autotools) and not its "native" Makefile because that seems broken. The repository of this is a partial snapshot/fork of the texlive one, so I'm also not sure if it makes that much sense to split off this one.

jmairboeck (commented):
I haven't included the ones from the "Japanese TeX Development Community" (yet) (see https://tug.org/svn/texlive/trunk/Build/source/texk/README?revision=71501&view=markup). Their repository is also a [partial "fork"](https://github.com/texjporg/tex-jp-build) of the upstream texlive one, so I'm not sure if it would make sense to split these off? (in a common "tex_jp" package)

jmairboeck:
> I'll probably also try to update the `createSubpackageInfos` script in texlive so that it generates `cmd:...` REQUIRES for the texlive packages so that these external packages are proper dependencies then. Let's see if this works out.

That is probably a bad idea, adding hard requirements on quite obscure tools to the collection packages. We would need to split them up further, but as there are over 4000 single packages in TeX Live, that's not managable.

For example, xdvi is part of texlive_basic, and we don't even have xdvi currently, so that wouldn't even work.

PR #10914, opened 2024-08-26
Add HaiQR recipe

tmtfx:
Add HaiQR app recipe


Begasus:
Could you join IRC for a bit? Less noise here. :)

Begasus:
Didn't want to mess with squashing the commits, qrcode reverted to old version, works ok now, thanks!

PR #10921, opened 2024-08-26
qca, bump version

Begasus:


Begasus:
Leaving older qca around for now as some recipes have the wrong library required (lib:libqca doesn't exist).

PR #10939, opened 2024-08-27
breeze_icons, keep 6.2.0 around for 32bit (for now)

Begasus:


Begasus:
Since 6.3.0 it seems the build for this is broken on 32bit with the same error as on last push: https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/6724/builds/84157.log

PR #10941, opened 2024-08-28
breeze_icons, fix build for 32bit

Begasus:


Begasus:
Fixes build error "virtual memory exhausted: Out of memory" from previous run on buildmaster.

PR #10975, opened 2024-08-29
amd_microcode: bump to 20240828

kenmays:
Ref: (https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/?id=091bd5adf19c7ab01214c64689952acb4833b21d)

Note: 
(https://www.amd.com/en/resources/product-security/bulletin/amd-sb-7014.html) - rev. 08/20/2024

* Update AMD cpu microcode for processor family 17h - 2024-07-10
* Update AMD cpu microcode for processor family 19h - 2024-07-10


korli:
I often follow dates from Arch Linux. In this case 20240809 would be the date of the public report of the vulnerability, see https://www.amd.com/en/resources/product-security/bulletin/amd-sb-7014.html

mtl1979:
> As mentioned on a [previous recipe update](https://github.com/haikuports/haikuports/commit/59a6549b50bda33c9cedb0f8809ce26cace764fd#commitcomment-139499788). Are the changes in SOURCE_URI necessary for the files that did not get updated? (I guess is harmless, as long as checksum is the same, just wondering about korli's previous feedback on that comment).

As far as I know, the changing part of the URL is the Git commit hash, which means as long as the original/old commit is still in the git repository, updating the SOURCE_URI isn't necessary if the file hasn't changed between the two commits...

kenmays:
After the IRC, I didn't test this - just slapped postinstall together. But, I'll pull this PR as I don't know the new outcome and don't have the time to review it on various AMD systems yet.

korli:
Testing is overrated :p

mtl1979:
> Testing is overrated :p

If something doesn't work, just wait for someone to notice and ask them to fix it... Problem solved... That's the whole essence of peer review. It's not like it's guaranteed that if it works locally, it also works on buildmaster and every other hrev there has been or will be. There's so many things we can break before feature freeze of R1B6 that we also have so much time to choose what to fix in mean time.

kenmays:
mtr1979 - True. If the lighted path here is to provide a tested post install script on various AMD platforms to upload the AMD microcode to them - so be it. Let that get done first before further commentary and review.

PR #10983, opened 2024-08-29
New receipe  : cppfront

yann64:
Add cppfront to haikuports
Hopefully will lower the entry bar to use the Haiku API for those not familiar with plain C++


dev-lang/cppfront/cppfront-0.7.4.recipe
```diff
@@ -0,0 +1,41 @@
...
+
+BUILD()
+{
+	g++ source/cppfront.cpp -std=c++20 -o cppfront
+}
+
+INSTALL()
+{
+	mkdir -p $binDir
+	cp cppfront $binDir
```
Begasus, line 40: This won't work with what PROVIDES declares, this way the binary ends up in /bin/x86 on 32bit resulting in:
```
	cmd:cppfront$secondaryArchSuffix = $portVersion
```
Better would be to install directly to /bin ($prefix/bin)
```diff
-	mkdir -p $binDir
-	cp cppfront $binDir
+	mkdir -p $prefix/bin
+	cp cppfront $refix/bin
```

yann64: Being on X86_64 I will create a x86 VM otherwise I will keep not seeing this type of issues.
Will update/correct in the coming days.

Begasus: Thanks!

Begasus:
Thanks! :)

PR #10984, opened 2024-08-29
podofo, bump version (also openssl3 issue)

Begasus:


Begasus:
Checked with (a rebuild) scribus and calibre (no changes), @korli the previous patchset doesn't apply anymore, changes are done in those files upstream, so far check on 64bit seems to be good.

Issue #10864, opened 2024-08-24
webkit_gtk did not find freetype


pulkomandy:
https://build.haiku-os.org/buildmaster/master/x86_gcc2/logviewer.html?buildruns/6674/builds/84036.log

```
-- Found Cairo: /packages/xcairo1.18_x86-1.18.0-1/.self/develop/lib/x86/libcairo.so (found suitable version "1.18.0", minimum required is "1.14.0")
-- Found Fontconfig: /packages/fontconfig_x86-2.13.96-2/.self/develop/lib/x86/libfontconfig.so (found suitable version "2.13.96", minimum required is "2.8.0")
CMake Error at /boot/system/data/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
 Could NOT find Freetype (missing: FREETYPE_LIBRARY FREETYPE_INCLUDE_DIRS)
 (Required is at least version "2.4.2")
Call Stack (most recent call first):
 /boot/system/data/cmake/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
 /boot/system/data/cmake/Modules/FindFreetype.cmake:162 (find_package_handle_standard_args)
 Source/cmake/WebKitFindPackage.cmake:44 (_find_package)
 Source/cmake/OptionsGTK.cmake:18 (find_package)
 Source/cmake/WebKitCommon.cmake:229 (include)
 CMakeLists.txt:21 (include)


-- Configuring incomplete, errors occurred!
Warning: Command '['bash', '-c', '. /wrapper-script']' returned non-zero exit status 1.
cleaning 'chroot/boot' folder
keeping chroot folder /boot/home/haikuports/net-libs/webkit-gtk/work-x86-2.40.0 intact for inspection
Error: Build has failed - stopping.
command exit status: 1
purging port with command: cd "/boot/home/haikuports" && "/boot/home/haikuporter/haikuporter" --config="/boot/home/haikuports/builder.conf" --no-package-obsoletion --ignore-messages --purge "webkit_gtk_x86-2.40.0"
Checking if any dependency-infos need to be updated ...
Looking for stale dependency-infos ...
Removing sources of webkit_gtk_x86-2.40.0 ...
Removing source /boot/home/haikuports/net-libs/webkit-gtk/download/webkitgtk-2.40.0.tar.xz ...
Cleaning work directory of webkit_gtk_x86-2.40.0 ...
build failed: build failure
```

The packages are activated:

[...]

Issue #10234, opened 2024-03-19
cmake doesn't support anything except x86 architectures


kallisti5:
Looking at Source/kwsys/SystemInformation.cxx, the cmake system information assumes Haiku is always x86 (via static cpu_info references to x86-specific cpuid fields)

Example on riscv64:
```
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx: In member function 'bool cmsys::SystemInformationImplementation::QueryHaikuInfo()':
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5002:3: error: 'cpuid_info' was not declared in this scope; did you mean 'cpu_info'?
 5002 |   cpuid_info cpu_info;
      |   ^~~~~~~~~~
      |   cpu_info
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5003:22: error: expected primary-expression before ',' token
 5003 |   get_cpuid(&cpu_info, 0, 0);
      |                      ^
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5003:3: error: 'get_cpuid' was not declared in this scope; did you mean 'getuid'?
 5003 |   get_cpuid(&cpu_info, 0, 0);
      |   ^~~~~~~~~
      |   getuid
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5008:25: error: expected primary-expression before '.' token
 5008 |   strncpy(vbuf, cpu_info.eax_0.vendor_id, 12);
      |                         ^
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5015:22: error: expected primary-expression before ',' token
 5015 |   get_cpuid(&cpu_info, 1, 0);
      |                      ^
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5017:38: error: expected primary-expression before '.' token
 5017 |   this->NumberOfLogicalCPU = cpu_info.eax_1.logical_cpus;
      |                                      ^
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5020:31: error: expected primary-expression before '.' token
 5020 |   this->ChipID.Type = cpu_info.eax_1.type;
      |                               ^
/sources/cmake-3.28.3/Source/kwsys/SystemInformation.cxx:5023:33: error: expected primary-expression before '.' token
 5023 |   this->ChipID.Family = cpu_info.eax_1.family;
      |                                 ^
[...]

PR #10986, opened 2024-08-29
cmake: fix detection of secondary arch

lonemadmax:
/boot/system/lib/x86 may not be the last path in the list, as is the case with clang after packaged version 12.

Related: #10864

Doesn't build due to an unrelated problem, though. `arc4random_buf` is not detected, but while compiling Utilities/cmlibarchive/libarchive/archive_random.c stdlib.h does include it, leading to a redefinition. Don't know where and how to add _DEFAULT_SOURCE and libbsd, all that I've tried has failed. But then I don't know cmake.

### DIFF
--- a/dev-build/cmake/cmake-3.28.3.recipe
+++ b/dev-build/cmake/cmake-3.28.3.recipe
@@ -91,6 +91,8 @@ defineDebugInfoPackage cmake$secondaryArchSuffix \
 
 BUILD()
 {
+	export LDFLAGS="-lbsd"
+
 	# not an autotools configure
 	./configure --prefix=$prefix \
 		--datadir=/$relativeDataDir/cmake \
--- a/dev-build/cmake/cmake-3.28.3.recipe
+++ b/dev-build/cmake/cmake-3.28.3.recipe
@@ -7,7 +7,7 @@ of your choice."
 HOMEPAGE="https://cmake.org/"
 COPYRIGHT="2002-2023 Kitware, Inc., Insight Consortium"
 LICENSE="CMake"
-REVISION="1"
+REVISION="2"
 SOURCE_URI="https://cmake.org/files/v${portVersion%.*}/cmake-$portVersion.tar.gz"
 CHECKSUM_SHA256="72b7570e5c8593de6ac4ab433b73eab18c5fb328880460c86ce32608141ad5c1"
 PATCHES="cmake-$portVersion.patchset"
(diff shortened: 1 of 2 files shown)


Begasus:
Haven't found a way to patch this in but the next works for getting a full build, adding this just before `./configure`
```
	export CPPFLAGS="_GNU_SOURCE"
	export LDFLAGS="-lbsd"
```

For the record, I was/am building cmake without GUI so it won't take in the extra dependencies when checking here.

Begasus:
@lonemadmax I've pushed my and your changes to: https://github.com/Begasus/haikuports/commit/21b23f3de2d9103f0e8ac54dbf0839312c9eb0f6
could you check also, and maybe import this here also (build on 32bit almost done here).
could also push the changes to the recipe here if need be :)

lonemadmax:
> adding this just before `./configure`
> 

I feel stupid now. I had passed it to make, but never to configure. All that's needed is the library, so that the test sees the symbol and libarchive happily compiles and links.

> 
> For the record, I was/am building cmake without GUI so it won't take in the extra dependencies when checking here.

Do you mean you were going to remove cmake-gui or that you did it to cut down on compilation while trying to find a fix for my problem? In the first case I'd prefer we closed this and you went on.

Begasus:
> > adding this just before `./configure`
> 
> I feel stupid now. I had passed it to make, but never to configure. All that's needed is the library, so that the test sees the symbol and libarchive happily compiles and links.
> 
> > For the record, I was/am building cmake without GUI so it won't take in the extra dependencies when checking here.
> 
> Do you mean you were going to remove cmake-gui or that you did it to cut down on compilation while trying to find a fix for my problem? In the first case I'd prefer we closed this and you went on.

Remove cmake-gui (to get rid of the extra dependencies), no need to close this, if checks are done I can just push the "recipe" alone to this PR.

@X547 mentioned if the next could be included for riscv: https://github.com/haikuports/haikuports/issues/10234#issuecomment-2041416537

Begasus:
Went along and pushed the recipe here, should be on par (checking never hurts) :)

lonemadmax:
RISCV untested.

testUVProcessChain crashes for me in 32bits.

Couldn't run the tests for x86_64.

Begasus:
Same crash on 64bit
Think this should be good if things are fine on 32bit (few testruns went fine here) to merge?

X547:
Recipe still have cyclic build dependency: cmake -> libcpppdap -> cmake.

Begasus:
> Recipe still have cyclic build dependency: cmake -> libcpppdap -> cmake.

Didn't know that one untill now.

pulkomandy:
I confirm that this fixes the problem with webkit_gtk. Merging, and the other problems can be fixed separately.

X547:
> RISCV untested.

riscv64 build confirmed (with disabled libcppdap package for now).

PR #11018, opened 2024-09-01
ccls: fix llvm18 requires

augiedoggie:


dev-util/ccls/ccls-0.20240202.recipe
```diff
@@ -32,14 +32,14 @@ PROVIDES="
 REQUIRES="
 	haiku$secondaryArchSuffix
 	lib:libclang_cpp$secondaryArchSuffix >= 18
-	lib:libLLVM$secondaryArchSuffix
+	lib:libLLVM$secondaryArchSuffix >= 18
```
korli, line 35: only BUILD_REQUIRES constraints should be needed.

PR #11034, opened 2024-09-01
harfbuzz: bump to 9.0.0

ElectrodeYT:
Harfbuzz 9.0.0 seems to build some tools; I split them out into a seperate `harfbuzz_tools` package.
Also, would like advice on how to handle the glib dependency: I enabled the `glib` option as I am (or at least, was, have stumbled into some further issues that honestly are a little too complex for me :sweat_smile:) trying to port gtk4 and it needed it enabled here.


media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -79,6 +80,24 @@ REQUIRES_devel="
 	devel:libicuuc$secondaryArchSuffix
 	"
 
+PROVIDES_tools="
+	harfbuzz${secondaryArchSuffix}_tools = $portVersion compat >= 0.9
```
jmairboeck, line 84: is this a typo? (should that be 9.0 and not 0.9)

Begasus:
gtk4 "can" be build, minibrowser crashes etc last time I tried, not really something to work on iirc being told back then. :)

ElectrodeYT:
> gtk4 "can" be build, minibrowser crashes etc last time I tried, not really something to work on iirc being told back then. :)

I did get it to compile, what's not working is cursor themes (which I sorta patched?), Window decorations (expected), it rendering more than a single frame and it not crashing when you do, well, really anything (crashes inside of our Wayland code, as far as I can tell, but I haven't really debugged it that much yet)

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -31,7 +31,7 @@ if [ "$targetArchitecture" = x86_gcc2 ]; then
 	commandBinDir=$prefix/bin
 fi
 
-libVersion="0.60830.0"
+libVersion="0.60900.0"
 libVersionCompat="$libVersion compat >= ${libVersion%%.*}"
```
jmairboeck, line 35: General question: is this backwards compatible with the old version? or does other software that uses the libraries need to be rebuilt?

According to this it should be compatible, but the major version increase may be an indicator otherwise.

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
ElectrodeYT: That was me copying the other provides lines

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
ElectrodeYT: I haven't instantly noticed things crashing or failing to link, how would I best check this?

jmairboeck: Ok, should be good then I think.

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
jmairboeck: Sorry, I didn't notice the unchanged part above. Maybe factor it out in a separate `portVersionCompat` variable to not repeat this twice.

Not all recipes do this, but I think it would be clearer that way.

ElectrodeYT: OK, done

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -18,7 +18,7 @@ COPYRIGHT="2004-2018 Behdad Esfahbod
 LICENSE="MIT"
 REVISION="2"
```
jmairboeck, line 19: reset the revision to 1 on updates

ElectrodeYT: Woops, done

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -79,6 +82,24 @@ REQUIRES_devel="
...
 
+PROVIDES_tools="
+	harfbuzz${secondaryArchSuffix}_tools = $portVersion compat >= $portVersionCompat
+	cmd:hb_info
+	cmd:hb_ot_shape_closure
+	cmd:hb_shape
+	cmd:hb_subset
+	"
+
+REQUIRES_tools="
```
korli, line 93:
```suggestion
REQUIRES_tools="
```

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -79,6 +82,24 @@ REQUIRES_devel="
...
+	haiku$secondaryArchSuffix
+	lib:libharfbuzz$secondaryArchSuffix
+	lib:libharfbuzz_icu$secondaryArchSuffix
+	lib:libharfbuzz_subset$secondaryArchSuffix
+	lib:libharfbuzz_gobject$secondaryArchSuffix
```
korli, line 98:
```suggestion
	harfbuzz$secondaryArchSuffix == $portVersion base
	haiku$secondaryArchSuffix
	lib:libharfbuzz_gobject$secondaryArchSuffix
```

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -161,6 +182,10 @@ INSTALL()
 		"$documentationDir" \
 		"$developDir" \
 		"$libDir"/cmake
+
+	# tools package
+	packageEntries tools \
+		"$binDir"
```
korli, line 188:
```suggestion
		"$commandBinDir"
```

media-libs/harfbuzz/harfbuzz-9.0.0.recipe
```diff
@@ -117,7 +138,7 @@ BUILD()
 		--datadir=$dataDir \
 		--bindir="$commandBinDir" \
 		-Dcairo=disabled \
-		-Dglib=disabled \
+		-Dglib=enabled \
```
korli, line 141: this was disabled in 24e31ffce283420266b7a9af3d47dd9e19e966cd for some reason

Begasus: IIRC it was disabled for a reason (eg not have a dependeny on glib)?

korli: yeah the commit message often isn't really helpful...

ElectrodeYT: I enabled it here because I was trying to port GTK4 and it needs a Harfbuzz with this option set; personally given that glib is a very standard package to have I can't see why having a dependency on it here would be a problem

pulkomandy: Harfbuzz is used in app_server, and so, all dependencies have to be shipped with Haiku, and it creates some complications to update them. So we try to keep dependencies to a minimum in these packages.

Also, glib provides things largely redundant with our existing standard libraries, so including it in Haiku base install would be a bit against the spirit of the whole project, I guess (to try to provide an unified system with a single API to do each thing). But different people may have different visions of what Haiku is and wether this is acceptable.

korli: https://docs.gtk.org/gtk4/building.html doesn't list harfbuzz as a dependency. https://www.linuxfromscratch.org/blfs/view/12.2/x/gtk4.html doesn't either.

Begasus: Looking at the old package I have here it seems to list it in packagelist (lost the recipe down the road it seems)
```
	lib:libgraphene_1.0 >= 0.1000.8
	lib:libharfbuzz >= 0.60821.0
	lib:libharfbuzz_gobject >= 0.60821.0
```

ElectrodeYT: > https://docs.gtk.org/gtk4/building.html doesn't list harfbuzz as a dependency. https://www.linuxfromscratch.org/blfs/view/12.2/x/gtk4.html doesn't either.

Weird; when I was compiling it it wanted to include `hb_glib.h`, which only gets installed when you set glib=true

korli: All this for https://github.com/GNOME/gtk/commit/1f7783224b7bc6bd8ca4b0f906e3ee542cc249db

Begasus:
Closing in favor for other (updated) version.

PR #11035, opened 2024-09-01
make: bump to 4.4.1

ElectrodeYT:


ElectrodeYT:
I'm not aware of there having been any breaking changes with make recently (although I am also not really the most knowledge about that); have remove the old recipes

About the issues: I haven't had any problems yet compiling things with it (`-j8` on a 4-core/4-thread machine), honestly wouldn't be surprised if the issues have been fixed, but I also have no idea exactly what the issue was to begin with; will try and investigate to see if I can reproduce any of the issues with 4.4.1.

ElectrodeYT:
Haven't been able to reproduce it yet, but I'm also currently struggling to get Haiku to survive compiling `gcc` (going to assume the fact I'm doing this on real hardware isn't helping), does anyone have any examples on what caused the previous issues?

pulkomandy:
The commit message for disabling it says:

    our signal handling is not good enough yet
    make[1]: *** duping jobs pipe: Bad file descriptor.  Stop.

The issue with 4.2 was the same.

> going to assume the fact I'm doing this on real hardware isn't helping

That shouldn't be a problem, in fact real hardware usually works faster and more reliably than VMs. What errors do you get?

ElectrodeYT:
> That shouldn't be a problem, in fact real hardware usually works faster and more reliably than VMs. What errors do you get

None; Haiku crashes/hangs so hard that I don't even get into KDL; as far as I can tell this isn't related to using a newer version of Make

waddlesplash:
Hangs that don't get to KDL should be reported on the bugtracker, if you haven't yet.

(What CPU are you running on? Does disabling SMP "solve" the hangs?)

ElectrodeYT:
> Hangs that don't get to KDL should be reported on the bugtracker, if you haven't yet.
> 
> (What CPU are you running on? Does disabling SMP "solve" the hangs?)

Was planning on reporting it to the bugtracker when I can get serial to work to try and see if I can get some more debug logs (forgot I needed a null-modem cable lol)
(i5 6500T, haven't tried disabling SMP yet, will try in a bit)

ElectrodeYT:
> Hangs that don't get to KDL should be reported on the bugtracker, if you haven't yet.
> 
> (What CPU are you running on? Does disabling SMP "solve" the hangs?)

Have opened up a ticket for this: https://dev.haiku-os.org/ticket/19069

ElectrodeYT:
Sorry for having been gone for a bit; have been busy with other things and lacking a little motivation lol

I have tested this a bit, and I can't get it to crash anymore, so as far as I can tell, this works

korli:
build fails on x86_gcc2.

waddlesplash:
Apparently make 4.3 and above use posix_spawn instead of fork, that sounds like a big advantage. Even if it uses file_actions, we should just create a real "spawn" syscall here and get the performance boost of not needing fork.

waddlesplash:
The error message ultimately comes from a failed dup here: https://github.com/mirror/make/blob/c63a5bc6a2881d515bb3020ed477fcba08fb2f3d/src/posixos.c#L110

I'm not sure how this is signal-related?

ElectrodeYT:
> Nope, same problem still happens as before:
> 
> ```
> make[3]: *** duping jobs pipe: Bad file descriptor.  Stop.
> make[3]: *** Waiting for unfinished jobs....
> ```
> 
> It got relatively far into the build of GCC, though, before this happened. And unfortunately the problem is intermittent, I didn't get it a second time with -j1.

As much as I haven't been doing too much Haiku stuff lately (have mostly been working on other projects), I have never gotten this myself; I wonder if its something super threading-related and I've just always gotten lucky (using -j4 on a 4 core CPU) or if there is something else going on?

waddlesplash:
I was using -j3 in a VM when this happened.

ElectrodeYT:
I've only ever been using real hardware, can't see how that should really affect it though

waddlesplash:
Whatever the problem was with `pselect`, it seems to be gone; I re-enabled that in the recipe and the GCC build completed successfully. (That option seems to avoids using those extra FDs entirely.)

waddlesplash:
Going to merge this but leave it disabled for the moment, and enable after some more testing.

waddlesplash:
I don't see that here. Are you running on beta5 + RAMFS? If so that's expected.

waddlesplash:
In that case, we can probably enable this "make" by default then, since the buildmasters don't use RAMFS of course.

PR #11041, opened 2024-09-02
abaddon, revbump for openssl3

Begasus:


Begasus:
@pulkomandy I had no problems with a static library for fmt when checking this?

Issue #10865, opened 2024-08-24
Packages to rebuild for openssl 3


pulkomandy:
List of recipes that use devel:libcrypto without a ">= 3" specifier (generated using `git grep "devel:libcrypto[^ p]*$"`):

- [ ] app-emulation/boxedwine/boxedwine-21.0.1.recipe (not compatible with openssl 3)
- [ ] dev-db/mysql/mysql-5.0.83.recipe (recipe is marked as broken on all architectures, untouched since being converted from bep file)
- [ ] dev-games/godot/godot-3.0.6.recipe (recipe marked as broken, made some "blind" changes)
- [x] dev-lang/php/php-7.4.30.recipe (not compatible with openssl 3)
- [ ] games-kids/tuxmath/tuxmath-1.7.2.recipe (Can't exec "autopoint": No such file or directory at /packages/autoconf-2.72-1/.self/data/autoconf/Autom4te/FileUtils.pm line 318.) segfaults
- [ ] net-im/kotatogram-desktop/kotatogram_desktop-1.4.9.recipe (CMake Error at cmake/options.cmake:88 (message): Unknown platform type)
- [x] www-client/otter-browser/otter_browser-1.0.03.recipe
- [x] www-servers/apache/apache-2.4.57.recipe
- [x] www-servers/lighttpd/lighttpd-1.4.53.recipe
- [x] webkit_gtk: fails to build on 32bit, does not find freetype

I am rebuilding them one by one after adding the >= 3 as needed, help welcome :)

Begasus:
Already did the ones for Qt6 here, packages relying on Qt6 and using openssl should be done after Qt6 has been done (my 2 cents). Already got a "fixed" version for nexcloud-client here, will push when done.
@threedeyes I only saw one reference for openssl in the patches for Qt6, so I guess that should still be fine?

jmairboeck:
I just built a new version of libpaper for my "project" of updating and unbundling texlive_core's dependencies. I'll take a look at R if it works with the new version and can also do openssl at the same time.

lonemadmax:
Should the ones that declare devel dependency on libssl and not libcrypto be also rebuilt? Falkon is one such case reported in the forums. It has both libraries in requires, but only libssl is declared in build requires.

pulkomandy:
Yes, I made the list in this ticket from devel:libcrypto, but devel:libssl should also be checked

Begasus:
[...]

PR #11059, opened 2024-09-05
php, drop php7, not compatible with openssl3

Begasus:


Begasus:
Related: https://github.com/haikuports/haikuports/issues/10865

PR #11071, opened 2024-09-06
kcolorpicker, re-instate package

Begasus:


Begasus:
Keep around for when it's used.

PR #11077, opened 2024-09-09
discid: build for Python 3.11/3.12

phw:
Another Python package to build for newer Python versions.


Begasus:
If there is no real reason to include them, then I tend to agree with @OscarL no need for python3.11/3.12 packages if they aren't used.

phw:
Makes sense, if packages are only wanted for dependencies of existing packages. Python 3.10 is the current default, right? And applications based on Python should stick to this, or would it be ok to e.g. upgrade the Picard package to use Python 3.12?

phw:
@OscarL Thanks a lot for the details, this makes it much clearer to me on how to handle the packaging.

PR #11092, opened 2024-09-10
itinerary, add KF6 package

Begasus:


Begasus:
Still an issue with kirigami, so best to launch it from Terminal.

PR #11094, opened 2024-09-11
asio, bump version

Begasus:


Begasus:
Old openssl dropped, doesn't seem to detect newer version, so disable it in the recipe.

PR #11097, opened 2024-09-11
lcov, revbump, rely on cmd:perl

Begasus:


Begasus:
@jmairboeck as we can't rely on the package "perl" I changed the requirement to "cmd:perl", this doesn't brake the build, not sure with perl changes if this also effects this (as it's "any" architecture).

jmairboeck (approved):
Yes, I think `cmd:perl` is correct in this case. This contains scripts that are executed with perl. It isn't a perl module package, those should use `vendor_perl` instead.

Begasus:
@jmairboeck there is one more in the reports in this regard :) https://eu.hpkg.haiku-os.org/haikuports/master/x86_gcc2/current/report.txt

PR #11107, opened 2024-09-12
libaccounts_qt, qmake is secondaryArchSuffix on 32bit

Begasus:


jmairboeck:
Is there a reason why qmake uses $secondaryArchSuffix? There is no gcc2 version of Qt.

Begasus:
Because no one moved it to $prefix/bin or used $commandBinDir? ;) Anyway, probably will disable this for 32bit, build failures on both Qt versions: https://build.haiku-os.org/buildmaster/master/x86_gcc2/?buildrunDir=6856&viewMode=expanded

PR #11109, opened 2024-09-12
kaccounts_integration, new recipes

Begasus:


Begasus:
For now disabled on 32bit (and packages that will rely on this) untill signond build is fixed.

PR #11153, opened 2024-09-19
Recipe for Dune Legacy

Peppersawce:
Dune Legacy is an engine to run Dune II. Yes, we already got OpenDUNE but this one has more features (and the port of OpenDUNE has issues playing music). The game installs and seems to run fine but once a game starts the window doesn't refresh properly, at least on my end.

What could cause something like that? I recall compiling and running this port without any issue on Haiku a bunch of years ago


games-strategy/dunelegacy/dunelegacy-0.96.4.recipe
```diff
@@ -0,0 +1,72 @@
...
+COPYRIGHT="2016 Dune Legacy"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="http://sourceforge.net/projects/dunelegacy/files/dunelegacy/$portVersion/dunelegacy-$portVersion-src.tar.bz2"
+CHECKSUM_SHA256="4740df9e94712b34f5d6fa1ca3c24658c4808b40431aa156ed66c318be9f4de8"
+SOURCE_FILENAME="dunelegacy-$portVersion-src.tar.bz2"
+PATCHES="dunelegacy-$portVersion.patchset"
+ADDITIONAL_FILES="dunelegacy.rdef.in"
+
+ARCHITECTURES="all ?x86_gcc2 ?x86"
```
Begasus, line 17: x86 isn't supported as primary architecture, no need to support gcc2 on this I guess.
```suggestion
ARCHITECTURES="all !x86_gcc2"
```

games-strategy/dunelegacy/dunelegacy-0.96.4.recipe
```diff
@@ -0,0 +1,72 @@
...
+SECONDARY_ARCHITECTURES="x86"
+
+PROVIDES="
+	dunelegacy$secondaryArchSuffix = $portVersion
+	app:dunelegacy = $portVersion
```
Begasus, line 22: I'd prefer (my interpretations) to have the binary installed in $commandBinDir (try to look up some examples on how to do this).

games-strategy/dunelegacy/dunelegacy-0.96.4.recipe
```diff
@@ -0,0 +1,72 @@
...
+	"
+
+BUILD()
+{
+	autoreconf --install
+
+	configure \
+		--build=$targetArchitecture-pc-haiku \
+		--bindir=$appsDir \
+		--datarootdir=$dataDir
```
Begasus, line 50: You can use runConfigure here, no need to set the paths (maybe just for commandBinDir).
If you decide for that, the lines mentioning the binary for the rdef should be changed also.

Peppersawce:
Any tips on how to debug the screen flicker issue btw?

Begasus:
No idea on the refres/flicker issue, maybe related to SDL2?

lonemadmax:
[Off-topic]

> and the port of OpenDUNE has issues playing music

Seems to work for me. Notice music is played through MIDI, and I think the nightlies don't include a MIDI soundfont by default, so if the issue is that no music is played you may need to install one.

Peppersawce:
I did some digging and found out the screen flicker issue is common for this version of the source:
[Flickering issue on Linux](https://github.com/henricj/dunelegacy/issues/2)
I applied the patch described there but the issue still affects the Mentat screen.

I could downgrade to 0.96.3  as that version doesn't seem to have the bug.
I am more tempted to switch to this version https://github.com/henricj/dunelegacy as it is more current. How would I go around setting a variant package? Would it be better if it was a separate package instead?

Peppersawce:
So, I noticed the entire Multiplayer menu was busted and crashed at almost any action, recompiled with more flags and it crashes less frequently (but it will eventually crash somewhere in there)

Peppersawce:
Added a recipe for a newer source where the flicker is completely fixed (after patching) among some improvements.

games-strategy/dunelegacy/dunelegacy-0.96.4_22.recipe
```diff
@@ -0,0 +1,75 @@
...
+DESCRIPTION="Dune Legacy is an effort by a handful of developers to revitalize \
+the first-ever real-time strategy game. \
+It tries to be as similar as possible to the original gameplay but to integrate \
+user interface features most modern realtime-strategy games have like selecting \
+multiple units"
+HOMEPAGE="https://dunelegacy.sourceforge.net/website/"
+COPYRIGHT="2016-2022 Dune Legacy"
+LICENSE="GNU GPL v2"
+REVISION="1"
+SOURCE_URI="https://sourceforge.net/code-snapshots/git/d/du/dunelegacy/code.git/dunelegacy-code-6ea9ac96854daa8c75ba429e78dc6716b147e106.zip"
```
Begasus, line 11: You now have 2 different recipes, do they need to co-exist?
Also, I think this is a snapshot/commit checkout? In that case `srcGitRev` is preferable.

Peppersawce: Not really, the new version seems better in every regard ngl

Peppersawce: Actually, I'm encountering an issue while downloading this version, the download link has to be generated first so it doesn't actually work in the recipe :(

Peppersawce: Fixed

Peppersawce:
Forgot to mention, even with the new version online multiplayer doesn't work and makes the game crash, if anybody's got an idea on how to solve that I'm all ears

Begasus:
Good to merge? Build went fine on 64bit, but I can't check the game itself as I don't have any data. :)

Peppersawce:
Imo yes, but I don't know if it compiles/runs on 32bits, it probably does I guess but some feedback on that would be nice

